PR #242, opened 2014-11-27
Walkthrough demonstrating weave 1.0 features

squaremo:
@rade says:

One way to elucidate all the features we'd like to be part of Weave 1.0, is to imagine what a demo transcript would look like.

Headings denote the new features we are demonstrating in that part of the transcript.

The demo highlights features with minimal effort; for a really compelling demo it may be advisable to re-cast everything in terms of an imaginary system somebody is putting together.

We focus on new features. All these are additive, i.e. all existing functionality is still present but may require slightly different CLI commands than present. Nothing should rely on new features or changes to Docker.

Key theme: ultimate simplicity for the common case

### DIFF
--- a/site/walkthrough.md
+++ b/site/walkthrough.md
@@ -1,0 +1,70 @@
+---
+title: weave Walkthrough
+---
+
+## launch DNS by default, IP allocation and naming, both with sensible defaults, rendezvous peering, auto-attach
+
+````bash
+root@a:~# weave launch
+root@a:~# weave run --name=pong -d -ti ubuntu
+````
+Weave automatically picked an IP range that isn't used locally (this range can also be specified explicitly, as an argument to `weave launch`), and allocated the container an IP in that range. The container is registered in weaveDNS as `pong.weave.local`. weaveDNS was launched automatically since no `--no-dns` option was given on `weave launch`.
+
+````bash
+root@b:~# weave launch
+root@b:~# weave run -ti ubuntu
+root@c0c8b4a7d709:/# ping pong
+ping pong.weave.local (10.0.1.1) 56(84) bytes of data.
+64 bytes from pong.weave.local (10.0.1.1): icmp_seq=1 ttl=64 time=0.036 ms
+64 bytes from pong.weave.local (10.0.1.1): icmp_seq=2 ttl=64 time=0.031 ms
+^C
+--- pong.weave.local ping statistics ---
+2 packets transmitted, 2 received, 0% packet loss, time 999ms
+rtt min/avg/max/mdev = 0.031/0.033/0.036/0.006 ms
+````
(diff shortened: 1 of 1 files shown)


site/walkthrough.md
```diff
@@ -0,0 +1,109 @@
...
+
+## launch DNS by default, IP allocation and naming, both with sensible defaults, rendezvous peering, auto-attach
+
+````bash
+root@a:~# weave launch
+root@a:~# weave run --name=pong -d -ti ubuntu
+````
+
+Weave automatically picked an IP range that isn't used locally (this
+range can also be specified explicitly, as an argument to `weave
```
bboreham: It isn't sufficient to pick a range that isn't used _locally_; it needs to be free on every host you're planning to add to the Weave network.  Which is quite a hard thing for the program to figure out on its own...

site/walkthrough.md
```diff
@@ -0,0 +1,109 @@
...
+^C
+--- pong.weave.local ping statistics ---
+2 packets transmitted, 2 received, 0% packet loss, time 999ms
+rtt min/avg/max/mdev = 0.031/0.033/0.036/0.006 ms
+````
+
+The container we started on `c` has the same name as the container we
+started on `a`. As a result, name resolution will return the addresses
+of both containers in random order.
+
```
bboreham: Just to note that, while starting two containers called 'pong' on different machines is allowed, Docker won't let you run two containers on the same machine with the same name, so you can't load-balance in this way across multiple processes without caring where they run.

Maybe in the context of swarm, Docker would even disallow the multi-machine case.

So, it may be sub-optimal to recommend this as a primary mechanism.

squaremo: We should probably use a label for this, in that case.

site/walkthrough.md
rade: I was well aware of that issue when I wrote this. The automatic picking is a convenience that is meant to work in most situations but not always. When two weave networks join up they could check whether they picked the same ranges. Though quite what they should do when that check fails I do not know. Hmm. Is there any reasonable approach here that *doesn't* require the user to specify a range?

site/walkthrough.md
rade: I don't see what labels get you. Might as well specify the DNS name, as is currently (and should still) be possible.

The whole point of using the name *by default* is that it is simpler. Also, afaik, docker is going the same direction with name based lookup.

Re swarm... I am pretty sure swarm uses prefixing or other mechanisms to uniquify container IDs across hosts.

site/walkthrough.md
rade: #743 is our current take on this.

bboreham:
Should we close this, having now released 1.0?

rade:
No, we need to go through it and pick up on any features we have missed (and still deem worthwhile).

squaremo:
> No, we need to go through it and pick up on any features we have missed (and still deem worthwhile).

I am in favour of closing it to take it off the radar. If you want to pick its bones, you can still do that after it's closed.

rade:
All remaining features are now covered by issues. -> closing.

PR #298, opened 2014-12-25
Remove both portnumber of connection and standard port

viirya:
When adding targets for peers in `router/connection_maker.go`, both portnumber of connection and standard port are tried. So I think it might be better to remove both addresses when any one of them is connected? Otherwise, another target will always be re-trying.

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -153,9 +153,13 @@ func (cm *ConnectionMaker) checkStateAndAttemptConnections() time.Duration {
 	now := time.Now() // make sure we catch items just added
 	after := MaxDuration
 	for address, target := range cm.targets {
-		if our_connected_targets[address] {
-			delete(cm.targets, address)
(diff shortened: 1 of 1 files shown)


bboreham:
What happens when the multiple ports on the same IP address are actually multiple hosts behind a NAT router?

(I suspect in this scenario no inward connections will get past the NAT router, but wondered if you had considered it)

rade:
`ConnectionMaker` tries to connect to any ip:port combo it gets its hands on. It could be smarter, but dropping ip:port from the target list when we have a connection to ip:weave_port risks missing endpoints behind which other peers might be lurking.

viirya:
I didn't consider the possibility when there are multiple hosts at the same IP but different ports. I am thinking if we can add another variable to record that a target is additionally added besides another target with non-standard port? So only if it is, we consider the two targets are the same and remove them both.

Issue #293, opened 2014-12-23
"Cannot open network namespace" error when running weave from systemd unit on CentOS 7


errordeveloper:
This was initially spotted by @yaronr.

Running `sudo weave launch` from CLI manually works, but when trying to run it from a systemd unit (like one shown below), the container fails to start and error displayed is about `eth0` missing.

```
# /etc/systemd/system/weave.service 
[Unit]
Description=Weave Network
Documentation=http://zettio.github.io/weave/
Requires=docker.service
After=docker.service
[Service]
ExecStartPre=/usr/bin/weave launch -password 122333
ExecStartPre=/usr/bin/weave launch-dns 10.10.2.1/16
ExecStart=/usr/bin/docker logs -f weave
SuccessExitStatus=2
ExecStop=/usr/bin/weave stop
ExecStop=/usr/bin/weave stop-dns
```
With a slight modification to the weave script adding call to `ip netns exec $NETNS ip link show eth0` without redirect above  line [215](https://github.com/zettio/weave/blob/34db7044e43e1dfd46c4eca9e242c319f77e246e/weave#L215); you can see the real error message:
```
Cannot open network namespace "2524": Permission denied
```

Firstly, I think we shouldn't hide what's happening there, i.e. capture output and display it on error or something like that...

Secondly, the cause of this might be with systemd+selinux, I'm not too sure.

errordeveloper:
@yaronr FYI, this is not an issue on Fedora 21 Base, not yet on Atomic. I have tested with official images for Base (`ami-66c76c11` @ `eu-west-1`). I have ran `yum upgrade && reboot` prior to testing it, which installed latest  `selinux-policy-3.13.1-103.fc21`.

rade:
#310 will document how to get around this issue.

There is nothing else to do here. -> closing.

wanghailiangx:
Maybe you could try to run "ip netns add 2524" at first.

PR #309, opened 2015-01-01
eliminate dependency on ethtool and conntrack on host

### LINKED ISSUE #101: eliminate dependency on ethtool and conntrack on host
I wonder if it is possible to have `ethtool` and `conntrack` inside the weave container instead of relying on the host for this? Reason for asking is because I use CoreOS but it does not have `conntrack` installed.

rade:
We package ethtool and conntrack into a `weavetools` image and run that as a short-lived container with `--privileged` and `--net=host` (or, in one case `--net=container:<id>`).

We use a plain ubuntu container as a build environment, so a) we don't require onerous tooling on the build host, and b) don't pollute the build host with build artefacts. The latter arise because building conntrack requires 'global' installation of its dependencies.

Closes #101.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,4 +29,5 @@ _testmain.go
 # Project specific
 weaver/weaver
 weavedns/weavedns
+tools/bin
 releases
--- a/Makefile
+++ b/Makefile
@@ -7,12 +7,15 @@ SUDO=sudo
 WEAVE_VERSION=git-$(shell git rev-parse --short=12 HEAD)
 WEAVER_EXE=weaver/weaver
 WEAVEDNS_EXE=weavedns/weavedns
+WEAVETOOLS_EXES=tools/bin
 WEAVER_IMAGE=zettio/weave
 WEAVEDNS_IMAGE=zettio/weavedns
+WEAVETOOLS_IMAGE=zettio/weavetools
 WEAVER_EXPORT=/var/tmp/weave.tar
 WEAVEDNS_EXPORT=/var/tmp/weavedns.tar
+WEAVETOOLS_EXPORT=/var/tmp/weavetools.tar
 
-all: $(WEAVER_EXPORT) $(WEAVEDNS_EXPORT)
+all: $(WEAVER_EXPORT) $(WEAVEDNS_EXPORT) $(WEAVETOOLS_EXPORT)
 
 $(WEAVER_EXE) $(WEAVEDNS_EXE):
 	go get -tags netgo ./$(shell dirname $@)
@@ -29,6 +32,9 @@ $(WEAVER_EXE) $(WEAVEDNS_EXE):
 $(WEAVER_EXE): router/*.go weaver/main.go
 $(WEAVEDNS_EXE): nameserver/*.go weavedns/main.go
 
+$(WEAVETOOLS_EXES): tools/build.sh
+	$(SUDO) docker run --rm -v $(realpath $(<D)):/home/weave ubuntu sh /home/weave/build.sh
+
 $(WEAVER_EXPORT): weaver/Dockerfile $(WEAVER_EXE)
 	$(SUDO) docker build -t $(WEAVER_IMAGE) weaver
 	$(SUDO) docker save $(WEAVER_IMAGE):latest > $@
@@ -37,10 +43,15 @@ $(WEAVEDNS_EXPORT): weavedns/Dockerfile $(WEAVEDNS_EXE)
 	$(SUDO) docker build -t $(WEAVEDNS_IMAGE) weavedns
 	$(SUDO) docker save $(WEAVEDNS_IMAGE):latest > $@
 
+$(WEAVETOOLS_EXPORT): tools/Dockerfile $(WEAVETOOLS_EXES)
+	$(SUDO) docker build -t $(WEAVETOOLS_IMAGE) tools
+	$(SUDO) docker save $(WEAVETOOLS_IMAGE):latest > $@
+
 # Add more directories in here as more tests are created
 tests:
 	cd nameserver; go test -tags netgo
 
 clean:
-	-$(SUDO) docker rmi $(WEAVER_IMAGE) $(WEAVEDNS_IMAGE)
-	rm -f $(WEAVER_EXE) $(WEAVEDNS_EXE) $(WEAVER_EXPORT) $(WEAVEDNS_EXPORT)
+	-$(SUDO) docker rmi $(WEAVER_IMAGE) $(WEAVEDNS_IMAGE) $(WEAVETOOLS_IMAGE)
+	rm -f $(WEAVER_EXE) $(WEAVEDNS_EXE) $(WEAVER_EXPORT) $(WEAVEDNS_EXPORT) $(WEAVETOOLS_EXPORT)
+	$(SUDO) rm -rf $(WEAVETOOLS_EXES)
--- a/README.md
+++ b/README.md
@@ -27,26 +27,12 @@ capabilities, so these can continue to be used by containers.
 
 ## Installation
 
-To run weave on a host, you need to install...
+Ensure you are running Linux (kernel 3.5 or later) and have Docker
+(version 0.9.1 or later) installed. Then install weave with
 
-1. Linux and Docker. We've tested with Docker versions 0.9.1 through
-   1.3.1, but other versions should work too. Linux kernels after 3.5
-   are known to work; the newer the better.
-2. weave. Install this with
-
-        sudo wget -O /usr/local/bin/weave \
-          https://github.com/zettio/weave/releases/download/latest_release/weave
-        sudo chmod a+x /usr/local/bin/weave
-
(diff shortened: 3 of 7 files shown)


rade:
@dpw @squaremo I would appreciate your feedback on the following...

## building the weavetools image
The main problem with the current approach is that the weavetools build is quite time-consuming - about 2m30s vs less than 1s for weave+weavedns.

That price is usually only paid once though. Unless one does a lot of 'clean' builds. In which case moving the weavetools arefacts from the 'clean' target to a different target ('distclean'?) would be the easiest fix.

There are some alternatives...

The weavetools build is dominated by the setting up of the build environment. We could pay that price just once, in one of the following ways...

Firstly, we could use the host as the build environment. The problem with that is twofold. Firstly, it creates some more build dependencies than we currently have. Nothing too onerous, but that may change if we add more stuff to weavetools. Secondly, I can only get conntrack to build by installing its dependencies 'globally', which is a showstopper. The conntrack build uses pkgconfig to locate dependencies. In theory it should be possible to perform a completely localised build even then, but poking that hideously complicated build tool chain for a few hours proved fruitless.

Secondly, we could publish a weavetools-build image containing the build tool chain. We would then perform builds in that image rather than the current plain ubuntu image. We could have a separate github repo with the Dockerfile for that image and use Dockerhub's [automatic builds](http://docs.docker.com/docker-hub/builds/) to get it onto Dockerhub. The main issue I see with this approach is risk of the weavetools-build image getting out of sync with what the weavetool build needs, especially when attempting to rebuild an older weave release. I reckon the risk is quite low - changes we make to weavetools-build would likely be backward compatible - but it is non-zero. Also, any additions to weavetools that have new build dependencies would require commits across two repositories.

However, even with the tool chain in place, the fetching & building of ethtool & conntrack, and their dependencies, takes an appreciable amount of time too - about 50s on my machine.

To eliminate that, we could extend the 2nd approach above and publish a weavetools-pre image instead (or as an extension) of weavetools-build. The fetching & building of ethtool & conntrack and their dependencies would happen as part of the weavetools-pre image build. The weavetools image build then has to do very little - just fire up a weavetools-pre container in order to copy the tools into a volume-mounted dir, and then build the weavetools image from that. The main downside of this approach is that the aforementioned issues around backward compatibility and multi-repository commits would be significantly compounded. However, I'd expect the weavetools-pre image to evolve rather slowly (though faster than weavetools-build).

We could go further still and move the entire production of the weavetools image to a separate repo, effectively treating it as an external dependency. It really needs to be independently versioned too at that point, and IMO that just adds unnecessary complication.

## versioning
Right now weavetools does not report a version. It gets tagged an uploaded to the docker hub as part of a release, but it is not included in `weave version`. Should it be? If so, how?

1. simply display the image version. Easy, but different from what we
   do elsewhere. OTOH, there really is no other version besides the
   image version; the executables inside it have their own version but
   these are not *weave* versions (and I do not think we should
   display them; the contents of the weavetools image ought to remain
   opaque to users).
2. include a `version` file in the image and somehow read that. This
   is quite involved, but possible. Also, while it produces the same
   result as what we do for the other images, the way it is achieved
   is quite different.
3. include a mini executable in the image that reports the
   version. This would make it easy to obtain the version - just run a
   container with that executable as the entry point. The downside is
   that the most straightforward way of adding such an executable,
   based on what we do elsewhere, i.e. a mini go program, also bloats
   the image a fair bit. We could write a small C program. Yuck. Also,
   this approach still isn't quite the same as what we do for other
   images...
4. As 3 but also set the default entry point of the image to the mini
   exe, and make that respond to `--version`. This would then make
   version reporting identical to what we do for the other images. The
   downside is that it makes invocation of the tools inside the images
   awkward; instead of simply `docker run ... zettio/weavetools /bin/tool arg1...`
   we'd have to `docker run ... --entrypoint=/bin/tool zettio/weavetools arg1 ...`.
   That is no big deal, since we run the image in just one place, but it does
   feel a little wrong.

## duplication in Makefile

There is some annoying duplication in the Makefile. The rules for building the three images are *nearly* identical. The main obstacle to refactoring them into one rule is that the image name cannot be derived from the directory name and vice versa, i.e. weave != weaver. We can't really change that. I am sure we could still remove the duplication, but it may not actually yield much a reduction in code.

dpw:
> The main problem with the current approach is that the weavetools build is quite time-consuming - about 2m30s vs less than 1s for weave+weavedns.
> 
> That price is usually only paid once though. Unless one does a lot of 'clean' builds. In which case moving the weavetools arefacts from the 'clean' target to a different target ('distclean'?) would be the easiest fix.

Does anyone do 'make clean' often enough to make this a problem?  'distclean' is probably not worth it unless/until someone screams.

> There are some alternatives...
> 
> The weavetools build is dominated by the setting up of the build environment. We could pay that price just once, in one of the following ways...
> 
> Firstly, we could use the host as the build environment. The problem with that is twofold. Firstly, it creates some more build dependencies than we currently have. Nothing too onerous, but that may change if we add more stuff to weavetools. Secondly, I can only get conntrack to build by installing its dependencies 'globally', which is a showstopper. The conntrack build uses pkgconfig to locate dependencies. In theory it should be possible to perform a completely localised build even then, but poking that hideously complicated build tool chain for a few hours proved fruitless.

Building weave on fedora is awkward, and I probably wouldn't recommend it to anyone else, but I do it.  So I would not welcome anything that makes it harder than it already is.

> Secondly, we could publish a weavetools-build image containing the build tool chain. We would then perform builds in that image rather than the current plain ubuntu image. We could have a separate github repo with the Dockerfile for that image and use Dockerhub's automatic builds to get it onto Dockerhub. The main issue I see with this approach is risk of the weavetools-build image getting out of sync with what the weavetool build needs, especially when attempting to rebuild an older weave release. I reckon the risk is quite low - changes we make to weavetools-build would likely be backward compatible - but it is non-zero. Also, any additions to weavetools that have new build dependencies would require commits across two repositories.

Seems like overkill, at least for now.  If we were to go down this route,  maybe we could think about containerizing the whole weave build (of interest to me due to the issues building on fedora, but it would also help with build reproducibility).

> ## duplication in Makefile
> 
> There is some annoying duplication in the Makefile. The rules for building the three images are nearly identical. The main obstacle to refactoring them into one rule is that the image name cannot be derived from the directory name and vice versa, i.e. weave != weaver. We can't really change that. I am sure we could still remove the duplication, but it may not actually yield much a reduction in code.

It will be easier to use define/endef and $(eval $(call ...)) than to contort those into one rule.

rade:
@dpw

> It will be easier to use define/endef and $(eval $(call ...)) than to contort those into one rule.

Ok. I don't think the duplication is bad enough to justify that. If there were a couple more rules or they were more complicated then yes. Let's leave it for now.

errordeveloper:
I think this might bypass the #293 as well, should I test that?

rade:
> I think this might bypass the #293 as well

I doubt it. `ip netns` is still being used in a few places.

errordeveloper:
> I doubt it. ip netns is still being used in a few places.

What would it take to ship `ip` command in the tools container as well?

rade:
> What would it take to ship `ip` command in the tools container as well?

A separate issue ;) And you might as well throw in iptables too. But be warned; if you file such an issue I'll be obliged to explain why it's a bad idea to do this.

squaremo:
* Building the image

I agree with @dpw, I think going beyond what is there now is overkill for the minute. A minutes-long build once in a while is far from terrible.

As a halfway point, perhaps the makefile could commit a weavetools-build image locally and use it if it's up to date. I don't know how hard that would be to get right.

 * Versioning

Out of these, I prefer a tiny C program that is called explicitly (`docker run weavetools /bin/print-version`). I don't care about consistency in how a version numbers are solicited as much as I care about keeping things simple. A file would do too, if it were possible to output the contents of it in one go.

rade:
> As a halfway point, perhaps the makefile could commit a weavetools-build image locally and use it if it's up to date. I don't know how hard that would be to get right.

That would be 90% of the way to the auto-built image in dockerhub, so I don't think it's worth doing.

> I prefer a tiny C program that is called explicitly

Hmm. That requires a fair bit of machinery to get into the weavetools image.

What is wrong with option 1, i.e. displaying the image version?

squaremo:
> What is wrong with option 1, i.e. displaying the image version?

I didn't really consider that one because you'd gone on to suggest other things, but there's nothing at all wrong with it.

dpw:
> I prefer a tiny C program that is called explicitly
> 
> Hmm. That requires a fair bit of machinery to get into the weavetools image.

Why? It doesn't sound complicated relative to building ethtool and conntrack.

squaremo:
> Does anyone do 'make clean' often enough to make this a problem? 'distclean' is probably not worth it unless/until someone screams.

Having just spent some time working on the build system side of things: scream.

squaremo:
>> Does anyone do 'make clean' often enough to make this a problem? 'distclean' is probably not worth it unless/until someone screams.
>>

> Having just spent some time working on the build system side of things: scream.

Having just spent some time testing the release process: SCREAM

PR #313, opened 2015-01-05
Gossip layer to communicate topology between peers

bboreham:
The old ProtocolUpdate and FetchAll requests disappear, replaced with a more generic Gossip implementation.
In addition to the usual send-to-some mode, this Gossip layer implements Broadcast-to-all and Unicast requests, although they are not used to communicate topology.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -48,7 +48,6 @@ type LocalConnection struct {
 	SessionKey        *[32]byte
 	heartbeatFrame    *ForwardedFrame
 	heartbeat         *time.Ticker
-	fetchAll          *time.Ticker
 	fragTest          *time.Ticker
 	forwardChan       chan<- *ForwardedFrame
 	forwardChanDF     chan<- *ForwardedFrame
@@ -285,8 +284,6 @@ func (conn *LocalConnection) queryLoop(queryChan <-chan *ConnectionInteraction) 
 			}
 		case <-tickerChan(conn.heartbeat):
 			conn.Forward(true, conn.heartbeatFrame, nil)
-		case <-tickerChan(conn.fetchAll):
-			err = conn.handleSendSimpleProtocolMsg(ProtocolFetchAll)
 		case <-tickerChan(conn.fragTest):
 			conn.setStackFrag(false)
 			err = conn.handleSendSimpleProtocolMsg(ProtocolStartFragmentationTest)
@@ -343,13 +340,9 @@ func (conn *LocalConnection) handleSetEstablished() error {
 		frame:   PMTUDiscovery},
 		nil)
 	conn.heartbeat = time.NewTicker(SlowHeartbeat)
-	conn.fetchAll = time.NewTicker(FetchAllInterval)
 	conn.fragTest = time.NewTicker(FragTestInterval)
 	// avoid initial waits for timers to fire
 	conn.Forward(true, conn.heartbeatFrame, nil)
-	if err := conn.handleSendSimpleProtocolMsg(ProtocolFetchAll); err != nil {
-		return err
-	}
 	conn.setStackFrag(false)
 	if err := conn.handleSendSimpleProtocolMsg(ProtocolStartFragmentationTest); err != nil {
 		return err
@@ -376,7 +369,6 @@ func (conn *LocalConnection) handleShutdown() {
 	}
 
 	stopTicker(conn.heartbeat)
-	stopTicker(conn.fetchAll)
 	stopTicker(conn.fragTest)
 
 	// blank out the forwardChan so that the router processes don't
@@ -572,35 +564,6 @@ func (conn *LocalConnection) handleProtocolMsg(tag ProtocolTag, payload []byte) 
 			return fmt.Errorf("unexpected nonce on unencrypted connection")
 		}
 		conn.Decryptor.ReceiveNonce(payload)
-	case ProtocolFetchAll:
-		// There are exactly two messages that relate to topology
-		// updates.
-		//
-		// 1. FetchAll. This carries no payload. The receiver responds
-		// with the entire topology model as the receiver has it.
-		//
-		// 2. Update. This carries a topology payload. The receiver
-		// merges it with its own topology model. If the payload is a
-		// subset of the receiver's topology, no further action is
-		// taken. Otherwise, the receiver sends out to all its
-		// connections an "improved" update.
-		conn.SendProtocolMsg(ProtocolMsg{ProtocolUpdate, conn.Router.Peers.EncodeAllPeers()})
-	case ProtocolUpdate:
-		newUpdate, err := conn.Router.Peers.ApplyUpdate(payload)
-		if _, ok := err.(UnknownPeersError); err != nil && ok {
-			// That update contained a peer we didn't know about;
-			// request full update
-			conn.SendProtocolMsg(ProtocolMsg{ProtocolFetchAll, nil})
-			break
-		}
-		if err != nil {
-			return err
-		}
-		if len(newUpdate) != 0 {
-			conn.Router.ConnectionMaker.Refresh()
-			conn.Router.Routes.Recalculate()
-			conn.Router.Ourself.SendProtocolMsg(ProtocolMsg{ProtocolUpdate, newUpdate})
-		}
 	case ProtocolPMTUVerified:
 		conn.verifyPMTU <- int(binary.BigEndian.Uint16(payload))
 	case ProtocolGossipUnicast:
--- a/router/connection.go
+++ b/router/connection.go
@@ -603,6 +603,12 @@ func (conn *LocalConnection) handleProtocolMsg(tag ProtocolTag, payload []byte) 
 		}
 	case ProtocolPMTUVerified:
 		conn.verifyPMTU <- int(binary.BigEndian.Uint16(payload))
+	case ProtocolGossipUnicast:
+		return conn.Router.handleGossip(payload, deliverGossipUnicast)
+	case ProtocolGossipBroadcast:
+		return conn.Router.handleGossip(payload, deliverGossipBroadcast)
+	case ProtocolGossip:
+		return conn.Router.handleGossip(payload, deliverGossip)
 	default:
 		conn.log("ignoring unknown protocol tag:", tag)
 	}
--- a/router/consts.go
+++ b/router/consts.go
@@ -18,9 +18,8 @@ const (
 	PMTUDiscoverySize  = 60000
 	FastHeartbeat      = 500 * time.Millisecond
 	SlowHeartbeat      = 10 * time.Second
-	FetchAllInterval   = 30 * time.Second
 	FragTestInterval   = 5 * time.Minute
-	ReadTimeout        = 2 * FetchAllInterval
+	ReadTimeout        = 1 * time.Minute
 	PMTUVerifyAttempts = 8
 	PMTUVerifyTimeout  = 10 * time.Millisecond // gets doubled with every attempt
 	MaxDuration        = time.Duration(math.MaxInt64)
--- a/router/gossip.go
+++ b/router/gossip.go
@@ -1,0 +1,162 @@
+package router
+
+import (
+	"bytes"
+	"encoding/gob"
+	"fmt"
+	"log"
+	"time"
+)
+
+const GossipInterval = 30 * time.Second
+
+type Gossip interface {
+	// specific message from one peer to another
+	// intermediate peers relay it using unicast topology.
+	GossipUnicast(dstPeerName PeerName, buf []byte) error
+	// send a message to every peer, relayed using broadcast topology.
+	GossipBroadcast(buf []byte) error
+}
+
+type Gossiper interface {
+	OnGossipUnicast(sender PeerName, msg []byte) error
+	OnGossipBroadcast(msg []byte) error
+	// Return state of everything we know; gets called periodically
+	Gossip() []byte
+	// merge in state and return "everything new I've just learnt",
+	// or nil if nothing in the received message was new
+	OnGossip(buf []byte) ([]byte, error)
+}
+
+type GossipChannel struct {
+	ourself  *LocalPeer
+	name     string
+	hash     uint32
+	gossiper Gossiper
+}
+
+func (router *Router) NewGossip(channelName string, g Gossiper) Gossip {
+	channelHash := hash(channelName)
+	channel := &GossipChannel{router.Ourself, channelName, channelHash, g}
+	router.GossipChannels[channelHash] = channel
+	return channel
+}
+
+func (router *Router) SendAllGossip() {
+	for _, channel := range router.GossipChannels {
+		channel.SendGossipMsg(channel.gossiper.Gossip())
+	}
+}
+
+func (router *Router) SendAllGossipDown(conn Connection) {
+	for _, channel := range router.GossipChannels {
+		protocolMsg := channel.gossipMsg(channel.gossiper.Gossip())
+		conn.(ProtocolSender).SendProtocolMsg(protocolMsg)
+	}
+}
+
+func (router *Router) handleGossip(payload []byte, onok func(*GossipChannel, PeerName, []byte, *gob.Decoder) error) error {
+	decoder := gob.NewDecoder(bytes.NewReader(payload))
+	var channelHash uint32
+	if err := decoder.Decode(&channelHash); err != nil {
+		return err
+	}
+	channel, found := router.GossipChannels[channelHash]
+	if !found {
+		return fmt.Errorf("[gossip] received unknown channel with hash %v", channelHash)
+	}
+	var srcName PeerName
+	if err := decoder.Decode(&srcName); err != nil {
+		return err
+	}
+	if err := onok(channel, srcName, payload, decoder); err != nil {
+		return err
+	}
+	return nil
+}
+
+func deliverGossipUnicast(channel *GossipChannel, srcName PeerName, origPayload []byte, dec *gob.Decoder) error {
+	var destName PeerName
+	if err := dec.Decode(&destName); err != nil {
+		return err
+	}
+	if channel.ourself.Name == destName {
+		var payload []byte
+		if err := dec.Decode(&payload); err != nil {
+			return err
+		}
+		return channel.gossiper.OnGossipUnicast(srcName, payload)
+	} else {
+		return channel.relayGossipUnicast(destName, origPayload)
+	}
+}
+
+func deliverGossipBroadcast(channel *GossipChannel, srcName PeerName, origPayload []byte, dec *gob.Decoder) error {
+	var payload []byte
+	if err := dec.Decode(&payload); err != nil {
+		return err
+	}
+	if err := channel.gossiper.OnGossipBroadcast(payload); err != nil {
+		return err
+	}
+	return channel.relayGossipBroadcast(srcName, origPayload)
+}
+
+func deliverGossip(channel *GossipChannel, srcName PeerName, _ []byte, dec *gob.Decoder) error {
+	var payload []byte
+	if err := dec.Decode(&payload); err != nil {
+		return err
+	}
+	if newBuf, err := channel.gossiper.OnGossip(payload); err != nil {
+		return err
+	} else if newBuf != nil {
+		channel.SendGossipMsg(newBuf)
+	}
+	return nil
+}
+
+func (c *GossipChannel) SendGossipMsg(buf []byte) {
+	protocolMsg := c.gossipMsg(buf)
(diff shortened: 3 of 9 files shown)


rade:
a couple of cosmetics...

1. there area few places where go's `if blah, ok := mumble(); ok {} else {}` syntax should be used instead of separate assignment & conditional.

2. the gossip-related log messages could do with some uniformity, e.g. "Gossip: ..." (and try to not mention "gossip" in the message itself).

router/connection.go
```diff
@@ -472,46 +467,52 @@ func (conn *LocalConnection) receiveTCP(decoder *gob.Decoder, usingPassword bool
...
+			channelHash, msg := decodeGossipChannel(msg[1:])
+			channel, found := conn.Router.GossipChannels[channelHash]
+			if !found {
+				conn.log("received unknown gossip channel:\n", channelHash)
+			} else {
+				_, _, msg := decodePeerName(msg)
+				newBuf := channel.gossiper.OnGossip(msg)
+				if newBuf != nil {
+					channel.GossipMsg(newBuf)
+				}
```
rade: The above three gossip-related branches are crying out for a a helper invoked as `conn.handleGossip(msg[1:], func (channel, msg) {...})` that does the channel lookup, moans if it is not found, and otherwise invokes the function.

router/consts.go
```diff
@@ -33,9 +33,10 @@ const (
 	ProtocolFragmentationReceived  = iota
 	ProtocolStartFragmentationTest = iota
 	ProtocolNonce                  = iota
-	ProtocolFetchAll               = iota
-	ProtocolUpdate                 = iota
 	ProtocolPMTUVerified           = iota
+	ProtocolGossip
```
rade: please just ditch the `iota` everywhere it is not needed. Straight on master.

router/gossip.go
```diff
@@ -0,0 +1,105 @@
+package router
+
+import (
+	"log"
+)
+
+type Gossip interface {
+	// specific message from one peer to another
+	// intermediate peers should relay it using unicast topology.
```
rade: s/should/

router/gossip.go
```diff
@@ -0,0 +1,105 @@
...
+	GossipUnicast(dstPeerName PeerName, buf []byte) error
+	// intended for a state change that everyone should hear about quickly
```
rade: There's an inconsistency here with the above description of unicast; there we talk about what it is, here about the purpose.

router/gossip.go
```diff
@@ -0,0 +1,105 @@
...
+	// relayed using broadcast topology.
+	GossipBroadcast(buf []byte) error
+}
+
+type Gossiper interface {
+	OnGossipBroadcast(msg []byte)
+	OnGossipUnicast(sender PeerName, msg []byte)
+	// Return state of everything we know; intended to be called periodically
```
rade: s/intended to be/gets

router/gossip.go
```diff
@@ -0,0 +1,105 @@
...
+	h := hash(channelName)
+	channel := &GossipChannel{router.Ourself, channelName, h, g}
+	router.GossipChannels[h] = channel
+	return channel
+}
+
+// contains state for everyone that sending peer knows
+// done on an interval; sent by one peer down [all/random subset of] connections
+// peers that receive it should examine the info, and if it is broadcast
+func (router *Router) SendAllGossip() {
```
rade: I cannot make any sense of any of the above comment.

router/gossip.go
```diff
@@ -0,0 +1,105 @@
...
+			nameLenByte := []byte{byte(len(peerName))}
+			msg := Concat([]byte{ProtocolGossip}, uint32slice(c.hash), nameLenByte, peerName, buf)
+			conn.(*LocalConnection).SendTCP(msg)
+		}
+	})
+}
+
+// intended for state from sending peer only
+// done when there is a change that everyone should hear about quickly
+// peers that receive it should relay it using broadcast topology.
```
rade: The comment is a duplicate of that on the interface. remove.

router/gossip.go
```diff
@@ -0,0 +1,105 @@
...
+		for _, conn := range peer.NextBroadcastHops(srcPeer) {
+			conn.SendTCP(msg)
+		}
+	} else {
+		log.Println("Unable to relay gossip from unknown peer", srcName)
+	}
+}
+
+// specific message from one peer to another
+// intermediate peers should relay it using unicast topology.
```
rade: The comment is a duplicate of that on the interface. remove.

rade:
another cosmetic...

- "unicast" should come before "broadcast" everywhere (interface defs, method defs, handling code, etc, etc)

router/types.go
```diff
@@ -23,6 +23,7 @@ type Router struct {
 	ConnLimit       int
 	BufSz           int
 	LogFrame        func(string, []byte, *layers.Ethernet)
+	GossipChannels  map[uint32]*GossipChannel // does not change after startup - access not locked
```
rade: I'd ditch the comment.

Also, the new field should be moved up. Below `Routes`, I reckon.

router/connection.go
bboreham: It's not quite as simple as that - we can relay a unicast message without being able to understand the contents of the message.

router/peers.go
```diff
@@ -57,6 +57,15 @@ func (peers *Peers) ForEach(fun func(PeerName, *Peer)) {
 	}
 }
 
+// Create an "improved" update from an incoming update:
```
rade: That only half-describes what this function does, i.e. it describes what goes in the result but not what happens to the state.

router/connection.go
```diff
@@ -426,6 +422,17 @@ func checkHandshakeStringField(fieldName string, expectedValue string, handshake
...
+func (conn *LocalConnection) handleGossip(msg []byte, onok func(channel *GossipChannel, srcName PeerName, payload []byte)) {
+	channelHash, msg := decodeGossipChannel(msg[1:])
+	channel, found := conn.Router.GossipChannels[channelHash]
+	if !found {
+		conn.log("received unknown gossip channel:\n", channelHash)
+	} else {
+		srcName, payload := decodePeerName(msg)
+		onok(channel, srcName, payload)
+	}
+}
```
rade: cosmetic: that 'if' should be combined with the preceding assignment.

rade:
> the gossip-related log messages could do with some uniformity, e.g. "Gossip: ..." (and try to not mention "gossip" in the message itself).

This hasn't been done. Elsewhere we use brackets to denote sub-system, so perhaps that is what we should do here, e.g. "[gossip]".

rade:
So the major area of concern here, besides the complexity/readability of the tests, is the change from what gets communicated when. It would be helpful if you summarised here the differences between the old and new implementation.

testing/util.go
```diff
@@ -8,8 +8,8 @@ import (
 	"testing"
 )
 
-func CallSite() string {
-	_, file, line, ok := runtime.Caller(2)
+func CallSite(level int) string {
```
rade: This is rather brittle. Always was but now even more so.

I reckon we should simply display the entire stack trace instead.

router/peers_test.go
```diff
@@ -0,0 +1,170 @@
...
+	// Now try adding some connections
+	r1.AddTestConnection(r2)
+	r2.AddTestConnection(r1)
+	checkEncoding(t, r1.Peers.EncodeAllPeers(), rs(r1, r2), ca(cs(r2), nil))
+	checkEncoding(t, r2.Peers.EncodeAllPeers(), rs(r1, r2), ca(nil, cs(r1)))
+	// Currently, the connection from 2 to 3 is one-way only
+	r2.AddTestConnection(r3)
+	checkEncoding(t, r2.Peers.EncodeAllPeers(), rs(r1, r2, r3), ca(nil, cs(r1, r3), nil))
+	checkEncoding(t, r3.Peers.EncodeAllPeers(), rs(r3), ca(nil))
+}
```
rade: It should be possible (and would be preferable) to test the codec opaquely. I.e. feed the output of EncodeAllPeers into ApplyUpdate on a different router.

router/peers_test.go
```diff
@@ -0,0 +1,170 @@
...
+	r2.AddTestConnection(r1)
+	r2.AddTestConnection(r3)
+	r3.AddTestConnection(r1)
+	r1.AddTestConnection(r3)
+	r2.AddTestRemoteConnection(r1, r3)
+	r2.AddTestRemoteConnection(r3, r1)
+
+	// Drop the connection from 2 to 3, and 3 isn't garbage-collected because 1 has a connection to 3
+	r2.DeleteTestConnection(r3)
+	peersRemoved := r2.Peers.GarbageCollect()
```
rade: instead of instrumenting the gc with a different callback, a better test would to be check whether Peers contain exactly the peers we expect.

router/mocks_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (r1 *Router) DeleteTestConnection(r2 *Router) {
+	toName := r2.Ourself.Peer.Name
+	toPeer, _ := r1.Peers.Fetch(toName)
+	toPeer.DecrementLocalRefCount()
+	delete(r1.Ourself.Peer.connections, toName)
+	r1.Ourself.Peer.version += 1
+}
+
```
rade: I've [just added](https://github.com/zettio/weave/commit/1a31748e4092b3519a58e8b87ee0996a5f4509f5) some helper functions to `LocalPeer` which can be called instead of directly modifying `connections` and `versions` in all of the above.

router/peers_test.go
bboreham: That would check that the code is self-consistent, but would not check that the encoding matches the documentation or the previous release.  Since what you suggest is effectively done inside `TestGossipTopology(),` this test can either stand as an additional warning if the protocol is accidentally broken, or deleted if maintaining it is too much of a burden.

router/peers_test.go
bboreham: There is no different callback here; that is in `TestGossipTopology()`, which also checks whether Peers contain exactly the peers we expect.  This test is  checking whether the return value of the function is as expected.

router/gossip_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+// Create a remote Peer object plus all of its connections, based on the name and UIDs of existing routers
+func tp(r *Router, routers ...*Router) *Peer {
+	peer := NewPeer(r.Ourself.Peer.Name, r.Ourself.Peer.UID, r.Ourself.Peer.version)
+	for _, r2 := range routers {
+		p2 := NewPeer(r2.Ourself.Peer.Name, r2.Ourself.Peer.UID, r2.Ourself.Peer.version)
+		peer.connections[r2.Ourself.Peer.Name] = &mockConnection{p2, ""}
+	}
+	return peer
+}
```
rade: this should use `peer.SetVersionAndConnections`.

router/gossip.go
```diff
@@ -0,0 +1,137 @@
...
+	OnGossipBroadcast(msg []byte)
+	// Return state of everything we know; gets called periodically
+	Gossip() []byte
+	// merge in state and return "everything new I've just learnt",
+	// or nil if nothing in the received message was new
+	OnGossip(buf []byte) []byte
+}
+
+type GossipChannel struct {
+	localPeer *LocalPeer
```
rade: we call this 'ourself' everywhere else.

router/gossip.go
```diff
@@ -0,0 +1,137 @@
...
+			peerName := c.localPeer.Name.Bin()
+			nameLenByte := []byte{byte(len(peerName))}
+			msg := Concat([]byte{ProtocolGossip}, uint32slice(c.hash), nameLenByte, peerName, buf)
+			conn.(ConnectionSender).SendTCP(msg)
+		}
+	})
+}
+
+func handleGossip(conn *LocalConnection, msg []byte, onok func(channel *GossipChannel, srcName PeerName, origMsg, payload []byte)) {
+	if len(msg) < 10 {
```
rade: constant, please.

router/gossip.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (c *GossipChannel) GossipUnicast(dstPeerName PeerName, buf []byte) error {
+	srcPeerByte := c.localPeer.Name.Bin()
+	nameLenByte := []byte{byte(len(srcPeerByte))}
+	dstPeerByte := dstPeerName.Bin()
+	dstNameLenByte := []byte{byte(len(dstPeerByte))}
+	msg := Concat([]byte{ProtocolGossipUnicast}, uint32slice(c.hash), nameLenByte, srcPeerByte, dstNameLenByte, dstPeerByte, buf)
+	return c.RelayGossipTo(dstPeerName, msg)
+}
```
rade: We do use the `gob` codec in the rest of the weave protocol, so why not here?

bboreham: It takes 17 bytes to encode the basic gossip header, as opposed to 12 with this implementation.  We do use non-gob encodings, e.g. [here](https://github.com/zettio/weave/blob/1b229fcc30c45fb4c5e06c6a85d0e8e914feed6a/router/crypto.go#L133) and [here](https://github.com/zettio/weave/blob/1b3d65476c23de7f6242f32b924cd142bdb0110d/router/connection.go#L180).

That said, it would make a better argument if this code was similar to either of those, possibly using encoding/binary.

router/connection.go
```diff
@@ -11,6 +11,10 @@ import (
 	"time"
 )
 
+type ConnectionSender interface {
+	SendTCP(msg []byte)
+}
+
```
rade: I wouldn't bother with an interface with that; just cast to LocalConnection in the one call site.

bboreham: It's implemented by mockChannelConnection too, which delivers the info locally.

rade: yes, sorry. Realised that just after typing it.

router/gossip.go
```diff
@@ -0,0 +1,149 @@
...
+	channel := &GossipChannel{router.Ourself, channelName, h, g}
+	router.GossipChannels[h] = channel
+	return channel
+}
+
+func (router *Router) SendAllGossip() {
+	for _, c := range router.GossipChannels {
+		c.GossipMsg(c.gossiper.Gossip())
+	}
+}
```
rade: s/c/channel, for consistency.

router/gossip.go
```diff
@@ -0,0 +1,149 @@
...
+		channel.GossipMsg(msg)
+	}
+}
+
+func (router *Router) SendAllGossipDown(conn Connection) {
+	for _, c := range router.GossipChannels {
+		gossip := c.gossiper.Gossip()
+		c.send(gossip, conn)
+	}
+}
```
rade: s/c/channel, for consistency.

router/connection.go
```diff
@@ -502,38 +499,14 @@ func (conn *LocalConnection) receiveTCP(decoder *gob.Decoder) {
...
-				conn.Router.Ourself.BroadcastTCP(Concat(ProtocolUpdateByte, newUpdate))
-			}
 		case ProtocolPMTUVerified:
 			conn.verifyPMTU <- int(binary.BigEndian.Uint16(msg[1:]))
+		case ProtocolGossipUnicast:
+			handleGossip(conn, msg, deliverGossipUnicast)
+		case ProtocolGossipBroadcast:
+			handleGossip(conn, msg, deliverGossipBroadcast)
+		case ProtocolGossip:
+			handleGossip(conn, msg, deliverGossip)
```
rade: I reckon the above should all take `msg[1:]`.

bboreham: Would make it harder to do the relaying of unicast and broadcast messages.

router/connection.go
rade: See #355.

router/connection.go
rade: #355 provides an abstraction enforcing that.

router/mocks_test.go
```diff
@@ -0,0 +1,113 @@
...
+func NewTestRouter(t *testing.T, name PeerName, queue *peerQueue) *Router {
+	onMacExpiry := func(mac net.HardwareAddr, peer *Peer) {}
+	onPeerGC := func(peer *Peer) {
+		//t.Log("Removing unreachable", peer)
+		if queue != nil {
+			queue.peers = append(queue.peers, peer)
+		}
+	}
+	router := NewRouter(nil, name, nil, 10, 1024, nil)
+	router.Macs.onExpiry = onMacExpiry
```
rade: You don't really need to override `Macs.onExpiry`, do you? The default only does some logging.

router/gossip.go
```diff
@@ -0,0 +1,149 @@
...
+func (router *Router) SendAllGossip() {
+	for _, c := range router.GossipChannels {
+		c.GossipMsg(c.gossiper.Gossip())
+	}
+}
+
+func (router *Router) SendGossip(channelName string, msg []byte) {
+	channelHash := hash(channelName)
+	if channel, found := router.GossipChannels[channelHash]; !found {
+		log.Println("[gossip] attempt to send on unknown channel:", channelName)
```
rade: I suggest introducing a function that can be called as `logGossip("attempt to send ...", channelName)`. There are potential call sites.

router/peers.go
```diff
@@ -132,6 +132,30 @@ func (peers *Peers) String() string {
...
+			}
+			firstConn = false
+			buf.WriteString(fmt.Sprintf("\n    { \"remoteName\": \"%s\", \"tcpAddr\": \"%s\", \"established\": \"%t\" }", remoteName, conn.RemoteTCPAddr(), conn.Established()))
+		})
+		buf.WriteString(" ] }")
+	})
+	buf.WriteString("\n]}")
+	return buf.String()
+}
+
```
rade: This is horrific. Why does implementing `MarshalJSON` on the various types not work?

router/mocks_test.go
```diff
@@ -0,0 +1,113 @@
...
+// Construct a Router object with a mock interface to check peer
+// garbage-collection, and without firing up any ancilliary goroutines
+func NewTestRouter(t *testing.T, name PeerName, queue *peerQueue) *Router {
+	onMacExpiry := func(mac net.HardwareAddr, peer *Peer) {}
+	onPeerGC := func(peer *Peer) {
+		//t.Log("Removing unreachable", peer)
+		if queue != nil {
+			queue.peers = append(queue.peers, peer)
+		}
+	}
```
rade: Instead of instrumenting the gc with a callback that collects gc'ed peers, a better test would to be check whether Peers contain exactly the peers we expect.

weaver/main.go
```diff
@@ -145,6 +145,11 @@ func handleHttp(router *weave.Router) {
 		io.WriteString(w, fmt.Sprintln("Encryption", encryption))
 		io.WriteString(w, router.Status())
 	})
+	http.HandleFunc("/topo-json", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
+		json, _ := router.Peers.MarshalJSON()
+		w.Write(json)
+	})
```
rade: please move the json encoding stuff to a separate PR

router/mocks_test.go
```diff
@@ -0,0 +1,95 @@
...
+	toPeer.DecrementLocalRefCount()
+	conn, _ := r1.Ourself.Peer.ConnectionTo(toName)
+	r1.Ourself.deleteConnection(conn)
+}
+
+type mockConnection struct {
+	local         *Peer
+	remote        *Peer
+	remoteTCPAddr string // we are not currently checking the TCP address
+}
```
rade: I suggest you define this simply as an alias for / trivial extension of RemoteConnection. That way you won't need to define any of the methods below.

router/mocks_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (conn *mockConnection) Local() *Peer          { return conn.local }
+func (conn *mockConnection) Remote() *Peer         { return conn.remote }
+func (conn *mockConnection) RemoteTCPAddr() string { return "" }
+func (conn *mockConnection) Shutdown(error)        {}
+func (conn *mockConnection) Established() bool     { return true }
+
+func AssertEmpty(t *testing.T, array []*Peer, desc string) {
```
rade: That is too generic a name.

router/gossip.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (c *GossipChannel) send(buf []byte, conn Connection) {
+	msg := ProtocolMsg{ProtocolGossip, GobEncode(c.hash, c.ourself.Name, buf)}
+	conn.(ProtocolSender).SendProtocolMsg(msg)
+}
+
+func (c *GossipChannel) GossipMsg(buf []byte) {
+	c.ourself.ForEachConnection(func(_ PeerName, conn Connection) {
+		if conn.Established() {
```
rade: Gossip should be sent on *all* connections, not just 'Established' ones.

Issue #289, opened 2014-12-23
weavedns usage includes `test.*` options


squaremo:

rade:
@squaremo I believe you have banged your head against this one for a while. Mind assigning it to yourself?

PR #314, opened 2015-01-05
Move testing utils to its own package

squaremo:
So as not to pollute command-line flags with test.* options.

It appears that importing "testing" (from stdlib) will add flags. To
avoid this, we need code to not depend on testing. But, since
testing_utils was previously in the same package as the rest of the
code, it dragged testing in.

Renaming testing_util.go to e.g., nameserver/util_for_test.go might be enough to
avoid it being compiled with the regular code. We may want it to be
available to tests for other packages -- after all it is utilities for
testing -- so it goes in its own package.


squaremo:
> This is meant to fix #289, right?

Yus

squaremo:
NB you may well have to `rm  $GOPATH/pkg/linux_amd64/github.com/zettio/weave/*` and `make clean` for this to have an effect. Annoyingly.

PR #324, opened 2015-01-07
weave-224: zeroconf auto-discovery

inercia:
This branch adds support for rendezvous services and, in particular, for a mDNS-based rendezvous mechanism. Once the router is started, `weave join mdns://` will join any other router in the same network, or you can specify a particular group with `weave join mdns:///somegroup`. By running `weave leave` with the same group name, the rendezvous service will stop announcing and discovering on that group (but will keep connections discovered so far).

The rendezvous service could be extended in the future for supporting other services like zookeeper, etcd, AWS's route53...


Makefile
```diff
@@ -8,23 +8,27 @@ WEAVE_VERSION=git-$(shell git rev-parse --short=12 HEAD)
...
 
 update:
 	go get -u -f -v -tags -netgo ./$(dir $(WEAVER_EXE)) ./$(dir $(WEAVEDNS_EXE))
 
-$(WEAVER_EXE) $(WEAVEDNS_EXE):
+$(WEAVER_EXE) $(WEAVEDNS_EXE) $(WEAVERENDEZ_EXE):
 	go get -tags netgo ./$(@D)
 	go build -ldflags "-extldflags \"-static\" -X main.version $(WEAVE_VERSION)" -tags netgo -o $@ ./$(shell dirname $@)
-	@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
+	@ldd $@ 2>/dev/null | grep "not a dynamic executable" >/dev/null || { \
```
rade: This change is wrong and suggests it hasn't been tested.

Even w/o netgo we do get a statically linked executable. But a statically linked glibc will still attempt to load name resolvers dynamically.

cmd/rendezvous/main.go
```diff
@@ -0,0 +1,136 @@
...
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGQUIT)
+	buf := make([]byte, 1<<20)
+	for {
+		sig := <-sigs
+		switch sig {
+		case syscall.SIGQUIT:
+			runtime.Stack(buf, true)
+			Info.Printf("=== received SIGQUIT ===\n*** goroutine dump...\n%s\n*** end\n", buf)
+			rm.Stop()
```
rade: SIGQUIT handling is meant to follow the fine Java tradition of producing a thread dump. It shouldn't do anything else, so you don't want to `rm.Stop()` here.

router/connection_maker.go
```diff
@@ -48,7 +48,8 @@ func StartConnectionMaker(ourself *LocalPeer, peers *Peers) *ConnectionMaker {
 		peers:          peers,
 		queryChan:      queryChan,
 		cmdLineAddress: make(map[string]bool),
-		targets:        make(map[string]*Target)}
+		targets:        make(map[string]*Target),
+	}
```
rade: This is a gratuitous change.

router/connection_maker.go
```diff
@@ -111,6 +112,7 @@ func (cm *ConnectionMaker) queryLoop(queryChan <-chan *ConnectionMakerInteractio
 	}
 }
 
+
```
rade: This is a gratuitous change.

weaver/main.go
```diff
@@ -34,16 +33,16 @@ func main() {
...
+		justVersion		bool
+		ifaceName  		string
+		routerName 		string
+		password   		string
+		wait       		int
+		debug      		bool
+		prof       		string
+		peers      		[]string
+		connLimit  		int
+		bufSz      		int
```
rade: This is a gratuitous change. And it won't survive gofmt.

weaver/main.go
```diff
@@ -124,6 +123,7 @@ func main() {
 	router := weave.NewRouter(iface, ourName, []byte(password), connLimit, bufSz*1024*1024, logFrame)
 	router.Start()
 	for _, peer := range peers {
+		// the peer id must be just a regular IP address...
```
rade: remove

weave
```diff
@@ -69,6 +76,10 @@ command_exists () {
     command -v $1 >/dev/null 2>&1
 }
 
+log()   { echo "$1"; }
+warn()  { log "WARNING: $1"; }
+fatal() { log "ERROR: $1"; exit 1; }
+
```
rade: The majority of log messages should (and, in the original, used to) go to stderr, not stdout.

nameserver/zone.go
```diff
@@ -49,15 +49,20 @@ func (zone *ZoneDb) indexOf(match func(Record) bool) int {
...
-			return r.IP, nil
+			res = append(res, r.IP)
 		}
 	}
-	return nil, LookupError(name)
+
+	if len(res) == 0 {
+		err = LookupError(name)
+	}
+	return
```
rade: I have raised #338 for returning all matching local records.

squaremo:
OK I've had a fair look at the PR in its entirety. (Aside: there are a lot of incidental changes, and commits pulled in by merging, rather than rebasing onto, master -- it would be nice to have a cleaner PR to merge, at some point)

Overall I like the structure. To recap for observers: a weaverendezvous process runs in a container using the host network, and gets told by the user via the weave script about URLs to join or leave. The scheme of such a URL gives the method of rendezvous to use, e.g. mnds, and the rest of the URL names a group to join. The rendezvous container then starts a worker to check that group for peers via that rendezvous method, and when it finds any, tells the weave container. The mDNS worker checks for peers by occasionally sending an mDNS query asking for the group it's been told (as a domain name).

Now for some commentary.

I appreciate the desire to reuse bits (e.g., the mdns_client and mdns_server) from elsewhere in weave. I'm not sure it leads in the right direction though.

I would expect an mDNS rendezvous to use something more like (perhaps exactly like) DNS-SD and ask for say, "_weave._tcp.<group>.local.". Reusing the mdns_server means zones have to be extended to handle giving multiple answers to queries -- something that will have to be done (see #338 as Matthias mentioned), but which I'd rather wasn't done as a side-effect of adding oblique functionality.

The guessing external interfaces bit makes me uneasy, though I can see why it exists. Can we not come up with a better solution -- e.g., assume the common case eth0, and allow it to be supplied as an argument? I wouldn't think one would want to advertise all IPs by all available methods (so in fact, it might even be better to talk about IPs or networks rather than interfaces). But you will have thought this through @inercia -- what was your line of reasoning?

Makefile
inercia: This change was made because the previous condition didn't seem right either: even after recompiling with the "netgo" flag set, the `weaverendezvous` binary had a `"cgo_stub"` string. I think there are no dependencies that link with cgo generated libraries either... So any suggestion?

cmd/rendezvous/main.go
inercia: I agree. Moved to the right signal, SIGINT.

router/connection_maker.go
inercia: Changes, but I disagree. I think `}` at an end-of-line is not a good idea as it produces conflicts for small changes... in fact, I thought the comma in structs was compulsory for this reason...

router/connection_maker.go
inercia: Done

weaver/main.go
inercia: Fixed

weaver/main.go
inercia: Fixed

weave
inercia: I agree.

nameserver/zone.go
inercia: I agree. Not solved here though...

inercia:
@squaremo, sorry for the "merge" commits: each team has its own way of working, and I didn't realize you rebase instead of merging... I don't know if it can be fixed now (maybe by rebasing interactively?)...

Regarding the use of the mDNS code, I absolutely agree that a DNS-SD implementation would be much better, but it would also require more work, and there is nothing in the mDNS specification that forbids to return [multiple IP addresses](http://tools.ietf.org/html/rfc6762#page-7) for a DNS name query. This rendezvous service is more of a proof of concept that could evolve with time...

And regarding the way we obtain the external IP addresses, it can look somehow bizarre but I cannot see any other solution while keeping it simple, and it is in fact not so far from [recommended solutions](https://tools.ietf.org/html/rfc5245) where peers gather all their possible external IPs and ports and *offer* them to connecting peers (DNS-SD would also be of help here as we could specify IPs priorities, ports). And I think Weave could benefit from something like ICE in the future...

Makefile
rade: Perhaps there were some build outputs left in your $GOPATH/pkg dir. Try removing that.

inercia:
By the way, we could get some ideas/code for more rendezvous services from Docker's [Swarm](https://github.com/docker/swarm/tree/master/discovery) (in fact, I think that code would make sense as a library of its own)

squaremo:
> By the way, we could get some ideas/code for more rendezvous services from Docker's Swarm (in fact, I think that code would make sense as a library of its own)

Possibly -- if Docker ever split it out we could look into contributing to it and using it.

squaremo:
> And regarding the way we obtain the external IP addresses, it can look somehow bizarre but I cannot see any other solution while keeping it simple

I think it tries too hard to be simple! Some large proportion of the time, the desired interface will be eth0; less often, it will be one other interface (maybe eth1); somewhat less often the host will be multi-homed and will want to advertise more than one interface. So defaulting to eth0, and otherwise allowing one or more interface to be supplied as an argument, makes more sense to me as a simple solution.

> This rendezvous service is more of a proof of concept that could evolve with time...

Yes fair enough. There may be other reasons to build in DNS-SD support (including long-running query support), so it may be that this becomes easier down the line.

errordeveloper:
> So defaulting to eth0, and otherwise allowing one or more interface to be supplied as an argument, makes more sense to me as a simple solution.

With most recent versions of udev the naming of Ethernet interfaces has changed. What appeared as `eth0` before, is now `enp0s3` (it is the case on CoreOS as well as CentOS 7, most likely every other distro as well). So hardcoding it as `eth0` is no longer a safe choice. It's best to find a method of getting first non-loopback interface, and use that as a default.

awh:
Out of interest - what is the status of this PR?

inercia:
@awh I could bring it back to life if there is enough interest, it depends on you guys...

inercia:
Superseded by #1097

Issue #326, opened 2015-01-08
[dns] reverse lookups for weave IPs on other peers go unanswered


rade:
From [the code](https://github.com/zettio/weave/blob/71d96eed7d6d6833286226b3fdf47328cd994376/nameserver/server.go#L89)

````
					Debug.Printf("Failed lookup for %s; sending mDNS query", q.Name)
					// We don't know the answer; see if someone else does
					// TODO
````

So we don't actually do what we say in the log. Even responding with a lookup failure would be better than that, since then at least clients won't be sitting around waiting, as [observed](https://github.com/zettio/weave/issues/322#issuecomment-69129981) by one weave user.

Issue #327, opened 2015-01-08
[dns] reverse lookups for non-weave IPs go unanswered


rade:
Similar to #326, but for non-weave IPs.

bboreham:
Closed because they no longer go unanswered - they now get an NXDOMAIN response as a result of #328. 
Giving a useful response is the subject of #346

PR #328, opened 2015-01-08
Refactor, Reimplement, Repair (m)DNS

squaremo:
* The Zone interface is split in two, DB operations and another interface Lookup
 * The (regular) handlers are given a list of Lookup implementations to try in order
 * The mDNSy server uses a dns.ServeMux to serve only the .weave.local and .in-addr.arpa domains
 * The mDNSy server now answers PTR queries if it can

### DIFF
--- a/nameserver/mdns_server.go
+++ b/nameserver/mdns_server.go
@@ -50,23 +50,51 @@ func (s *MDNSServer) Start(ifi *net.Interface) error {
 		return err
 	}
 
-	handleMDNS := func(w dns.ResponseWriter, r *dns.Msg) {
-		// Ignore answers to other questions
-		if len(r.Answer) == 0 && len(r.Question) > 0 {
-			q := r.Question[0]
-			if ip, err := s.zone.LookupLocal(q.Name); err == nil {
-				m := makeAddressReply(r, &q, []net.IP{ip})
-				if err = s.sendResponse(m); err != nil {
-					Warning.Printf("Error writing to %s", w)
+	handler := func(qtype uint16, lookup func(*dns.Msg, *dns.Question) *dns.Msg) dns.HandlerFunc {
+		return func(w dns.ResponseWriter, r *dns.Msg) {
+			// Handle only questions, ignore answers. We might also
+			// ignore questions that arise locally (i.e., that come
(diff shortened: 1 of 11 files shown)


nameserver/mdns_server.go
```diff
@@ -50,23 +50,52 @@ func (s *MDNSServer) Start(ifi *net.Interface) error {
...
+
+	handleA := handler(dns.TypeA, func(r *dns.Msg, q *dns.Question) *dns.Msg {
+		if ip, err := s.zone.LookupLocal(q.Name); err == nil {
+			return makeAddressReply(r, q, []net.IP{ip})
+		} else {
+			return nil
+		}
+	})
+
+	handlePTR := handler(dns.TypePTR, func(r *dns.Msg, q *dns.Question) *dns.Msg {
```
bboreham: I think this should be named more about  reverse lookups and less about PTR.  There could be other PTR requests in future

squaremo: Fair cop

nameserver/mdns_server.go
```diff
@@ -50,23 +50,52 @@ func (s *MDNSServer) Start(ifi *net.Interface) error {
...
 
-	handleMDNS := func(w dns.ResponseWriter, r *dns.Msg) {
-		// Ignore answers to other questions
-		if len(r.Answer) == 0 && len(r.Question) > 0 {
-			q := r.Question[0]
-			if ip, err := s.zone.LookupLocal(q.Name); err == nil {
-				m := makeAddressReply(r, &q, []net.IP{ip})
-				if err = s.sendResponse(m); err != nil {
-					Warning.Printf("Error writing to %s", w)
+	handler := func(qtype uint16, lookup func(*dns.Msg, *dns.Question) *dns.Msg) dns.HandlerFunc {
```
rade: I suggest you lift this to a method. Would improve readability, imo.

rade:
I believe this is meant to fix #326 and #327. Correct?

squaremo:
> I believe this is meant to fix #326 and #327. Correct?

Nope, just the first (answer reverse DNS for weave addresses). The latter is partly covered by https://github.com/zettio/weave/issues/318, which I'm working on separately.

PR #332, opened 2015-01-08
Elminate the need for `-passwordfile` and simplify the script a little

### LINKED ISSUE #330: allow password to be supplied in env
@errordeveloper discovered that contrary to our thinking in #276 it is in fact possible to pass a `WEAVE_PASSWORD` env var to the weave container w/o it showing up in `ps`... by invoking docker with `-e WEAVE_PASSWORD` instead of `-e WEAVE_PASSWORD=...`.

We should do that, and get rid of the `-passwordfile` option, which was introduced solely because we couldn't find a better way to solve that problem.

errordeveloper:
This should close #330.

### DIFF
--- a/site/features.md
+++ b/site/features.md
@@ -131,23 +131,23 @@ sharing it between applications:
 ### <a name="security"></a>Security
 
 In order to connect containers across untrusted networks, weave peers
-can be told to encrypt traffic by supplying a `-password` or
-`-passwordfile` option when launching weave, e.g.
+can be told to encrypt traffic by supplying a `-password` option or
+`WEAVE_PASSWORD` environment variable when launching weave, e.g.
 
     host1# weave launch -password wEaVe
 
 or
 
-    host1# weave launch -passwordfile /path/to/file
+    host1# export WEAVE_PASSWORD=wEaVe
+    host1# weave launch
 
-where the file must contain a single line like this:
-
-    WEAVE_PASSWORD=wEaVe
+_NOTE: The command line option takes precedence over the environment
+variable._
 
 The same password must be specified for all weave peers; it is a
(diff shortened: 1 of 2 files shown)


site/features.md
```diff
@@ -131,19 +131,17 @@ sharing it between applications:
...
 
-    host1# weave launch -passwordfile /path/to/file
-
-where the file must contain a single line like this:
-
-    WEAVE_PASSWORD=wEaVe
+    host1# export WEAVE_PASSWORD=wEaVe
+    host1# weave launch
 
+_NOTE: command line option overrides the environment variable._
```
rade: "The command line option takes precedence over the environment variable."

weave
```diff
@@ -471,11 +471,8 @@ case "$COMMAND" in
         fi
         if [ "$1" = "-password" ] ; then
             [ $# -gt 1 ] || usage
-            PASSWORD=$2
-            shift 2
-        elif [ "$1" = "-passwordfile" ] ; then
-            [ $# -gt 1 ] || usage
-            PASSWORDFILE=$2
+            WEAVE_PASSWORD="$2"
+            export WEAVE_PASSWORD
```
rade: exporting env vars from scripts is generally bad. in this particular case it will affect any subsequent `weave launch`.

`-password` should continue to be implemented as before; yes it is less secure than the env var route, but that is why the latter exists.

errordeveloper: @rade, would you mind to explain how exactly a subcequent `weave lauch` would be affected? I am not sure I can follow...

rade: Ignore me. For some reason I thought `export` exported to the parent env.

errordeveloper: Well, it would when you `source $(which weave) launch`, maybe that's how
you mixed it up ;)

PR #333, opened 2015-01-08
re-instate 'make publish'

### LINKED ISSUE #331: make it possible (and easy) to release new images outside of official releases
We moved the `make publish` image publishing task into the release script as part of #263, since at the time the weave script would always pull the latest version of the images from the docker hub, so it was important to ensure that no "out of band" releases replaced that latest-tagged image.

The downside is that there is now no easy way for us to get images out *outside* a release. This is especially annoying since often there are updates to the weave script which go hand in hand with changes to the image, so as soon as one such change has been merged, the weave script on master no longer works with the published images.

The introduction of [version alignment](#306) has removed the rationale for preventing out-of-band releases. Official releases now always pull images of a specific version. By contrast, the weave script on master works with 'latest'.

Therefore re-introducing `make publish` would allow us to release new images whenever we like.

The release script should invoke `make publish`, so we avoid duplicating the code.

rade:
and call it from the release script

Closes #331.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
+PUBLISH=publish_weave publish_weavedns publish_weavetools
+
 .DEFAULT: all
-.PHONY: all update tests clean
+.PHONY: all update tests publish $(PUBLISH) clean
 
 # If you can use docker without being root, you can do "make SUDO="
 SUDO=sudo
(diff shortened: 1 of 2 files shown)


rade:
There are a few commits on master leading up to this.

@squaremo I would appreciate some help in testing.

1. do something
2. follow release process
3. do something else

What exactly do I need to do in 1+3 such that the above leaves no trace behind anywhere? Not on my machine, not in my github account/repos, not in my dockerhub account, and certainly not in the zettio accounts.

I know I can override GITHUB_USER and DOCKERHUB_USER, but that is not quite enough to accomplish the above, is it?

rade:
> What exactly do I need to do \[...\] such that \[testing the release process\] leaves no trace behind anywhere?

I suggest we document that in the wiki, since this won't be the last time we are making changes that affect the release process.

squaremo:
https://github.com/zettio/weave/wiki/TestingReleaseProcess

squaremo:
Waiting for weavetools image to build.

squaremo:
LGTM.

PR #334, opened 2015-01-08
cope with weavetools delays

### LINKED ISSUE #321: `weave launch` breakage due to excessive execution time of weavetools container
A user attempting to get weave to work on an Azure VM encountered failures on `weave launch` due to the weave router terminating after exceeding the 5s waiting period for the ethwe interface to become available. Further investigation revealed that docker container startup times on that VM were in the seconds. Since we are [now](#309) invoking ethtool via the weavetools container, and this needs to happen before marking the interface as up, the excessive container execution time would exceed the wait threshold.

We should fix that by increasing the threshold. I suggest 20s.

A related issue is that if this is the first invocation of the weavetools container then that may entail downloading the required image, which can take a very long time indeed. So we ought to make sure we have the image in place before entering the time-critical section between starting a container and marking its interface as up.

rade:
- ensure that we have the weavetools image downloaded before entering any time-critical sections, in particular post-container-startup network configuration
- in the weave router and dns containers, wait for longer for the ethwe interface to become available

Fixes #321.

### DIFF
--- a/weave
+++ b/weave
@@ -132,8 +132,6 @@ create_bridge() {
         ip link add name v${CONTAINER_IFNAME}du mtu $MTU type dummy
         ip link set dev v${CONTAINER_IFNAME}du master $BRIDGE
         ip link del dev v${CONTAINER_IFNAME}du
-        # disable offloading
-        run_tool host ethtool -K $BRIDGE tx off >/dev/null
         # Work around the situation where there are no rules allowing traffic
         # across our bridge. E.g. ufw
         add_iptables_rule filter FORWARD -i $BRIDGE -o $BRIDGE -j ACCEPT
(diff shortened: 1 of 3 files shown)


squaremo:
Trying this out:

```
$ git checkout master
$ # edit weaver/Dockerfile to reduce wait to 2 seconds
$ make
$ sudo ./weave stop
$ docker rmi zettio/weavetools
$ sudo ./weave launch
Container 54c733e428e6876b15b64a6b4d1bde950b89264fefd4c4fbacf3c9a718c52dce died
```
Oh noes!

then

```
$ git checkout pr334
$ # edit weaver/Dockerfile to reduce wait to 2 seconds
$ make
$ sudo ./weave stop
$ docker rmi zettio/weavetools
$ sudo ./weave launch
df90aed7b1f77bc9c1551ad93f15f5aac2e06cac36411fdd992e4dd57e7b3beb
```
:+1:

Issue #311, opened 2015-01-02
[dns] failed lookups not logged intelligibly at default log level


errordeveloper:
With `-debug` I can see this:
```
INFO: 2015/01/02 19:09:41.729556 Adding spark-worker-1.weave.local. -> 10.10.1.11
DEBUG: 2015/01/02 19:09:42.404113 Local query: {Name:spark-worker-1.weave.local. Qtype:1 Qclass:1}
DEBUG: 2015/01/02 19:09:42.404158 Local query: {Name:spark-worker-1.weave.local. Qtype:28 Qclass:1}
WARNING: 2015/01/02 19:09:42.404161 Local query not handled: {Name:spark-worker-1.weave.local. Qtype:28 Qclass:1}
DEBUG: 2015/01/02 19:09:42.407608 Local rdns query: {Name:11.1.10.10.in-addr.arpa. Qtype:12 Qclass:1}
DEBUG: 2015/01/02 19:09:42.407619 Looking for address: [10 10 1 11]
DEBUG: 2015/01/02 19:09:42.407623 Found name: spark-worker-1.weave.local.
DEBUG: 2015/01/02 19:09:43.689734 Local rdns query: {Name:0.0.0.0.in-addr.arpa. Qtype:12 Qclass:1}
DEBUG: 2015/01/02 19:09:43.689734 Looking for address: [0 0 0 0]
DEBUG: 2015/01/02 19:09:43.689734 Failed lookup for 0.0.0.0.in-addr.arpa.; sending mDNS query
DEBUG: 2015/01/02 19:09:48.692856 Local rdns query: {Name:0.0.0.0.in-addr.arpa. Qtype:12 Qclass:1}
DEBUG: 2015/01/02 19:09:48.692896 Looking for address: [0 0 0 0]
DEBUG: 2015/01/02 19:09:48.692903 Failed lookup for 0.0.0.0.in-addr.arpa.; sending mDNS query
DEBUG: 2015/01/02 19:09:53.844650 Local query: {Name:spark-masterz.weave.local. Qtype:1 Qclass:1}
DEBUG: 2015/01/02 19:09:53.844650 Failed lookup for spark-masterz.weave.local.; sending mDNS query
DEBUG: 2015/01/02 19:09:53.844650 Local query: {Name:spark-masterz.weave.local. Qtype:28 Qclass:1}
[...]

PR #336, opened 2015-01-09
Reform logging in weaveDNS

squaremo:
* Put prefixes (including msg IDs) in front of log messages to
   distinguish them
 * Failed lookups get logged under Info as per issue #311
 * More accurate messages as per issue #311 

(and switch debug logging on during tests, just to be helpful)

### DIFF
--- a/nameserver/mdns_lookup.go
+++ b/nameserver/mdns_lookup.go
@@ -10,10 +10,13 @@ func mdnsLookup(client *MDNSClient, name string, qtype uint16) (*Response, error
 	channel := make(chan *Response)
 	client.SendQuery(name, qtype, channel)
 	for resp := range channel {
-		Debug.Printf("Got response name %s addr %s", resp.Name, resp.Addr)
 		if err := resp.Err; err != nil {
+			Debug.Printf("[mdns] Error for query type %s name %d: %s",
+				dns.TypeToString[qtype], name, err)
 			return nil, err
 		} else {
+			Debug.Printf("[mdns] Got response name for query type %s name %s: name %s, addr %s",
+				dns.TypeToString[qtype], name, resp.Name, resp.Addr)
 			return resp, nil
 		}
 	}
--- a/nameserver/mdns_server.go
+++ b/nameserver/mdns_server.go
@@ -91,12 +91,15 @@ func (s *MDNSServer) makeHandler(qtype uint16, lookup LookupFunc) dns.HandlerFun
 			q := &r.Question[0]
 			if q.Qtype == qtype {
 				if m := lookup(s.zone, r, q); m != nil {
-					Debug.Printf("Found local answer to mDNS query %s", q.Name)
+					Debug.Printf("[mdns msgid %d] Found local answer to mDNS query %s",
+						r.MsgHdr.Id, q.Name)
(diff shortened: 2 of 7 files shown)


nameserver/mdns_server_test.go
```diff
@@ -34,7 +34,7 @@ func sendQuery(name string, querytype uint16) error {
 
 func TestServerSimpleQuery(t *testing.T) {
 	// The ff can be handy for debugging (obvs)
-	//wt.InitDefaultLogging(true)
+	wt.InitDefaultLogging(true)
```
rade: is this meant to be on? If so, the comment should go.

rade:
What is the purpose of logging r.MsgHdr.Id?

nameserver/mdns_server_test.go
squaremo: Oh I don't know, it still explains something.

squaremo:
> What is the purpose of logging r.MsgHdr.Id?

To be able to correlate events to individual messages. Most use in mDNS.

rade:
I figured it was for correlation, but am unclear on the scope. Is it local only, i.e. it allows you to correlate multiple log messages spat out by the same weavedns? Or does it enable correlation between mdns client and server(s)?

squaremo:
> Is it local only, i.e. it allows you to correlate multiple log messages spat out by the same weavedns? Or does it enable correlation between mdns client and server(s)?

For regular DNS it's just the local weavedns. For mdns it will be the same ID to all multicast recipients.

nameserver/mdns_lookup.go
```diff
@@ -10,10 +10,13 @@ func mdnsLookup(client *MDNSClient, name string, qtype uint16) (*Response, error
 	channel := make(chan *Response)
 	client.SendQuery(name, qtype, channel)
 	for resp := range channel {
-		Debug.Printf("Got response name %s addr %s", resp.Name, resp.Addr)
 		if err := resp.Err; err != nil {
+			Debug.Printf("[mdns] Error for query type %s name %d: %s",
```
bboreham: `go vet` says: arg name for printf verb %d of wrong type: string

nameserver/mdns_server.go
```diff
@@ -91,12 +91,15 @@ func (s *MDNSServer) makeHandler(qtype uint16, lookup LookupFunc) dns.HandlerFun
 			q := &r.Question[0]
 			if q.Qtype == qtype {
 				if m := lookup(s.zone, r, q); m != nil {
-					Debug.Printf("Found local answer to mDNS query %s", q.Name)
+					Debug.Printf("[mdns msgid %d] Found local answer to mDNS query %s",
+						r.MsgHdr.Id, q.Name)
 					if err := s.sendResponse(m); err != nil {
-						Warning.Printf("Error writing to %s", s.sendconn)
+						Warning.Printf("[mdns msgid %d] Error writing to %s",
```
bboreham: `go vet` says: arg s.sendconn for printf verb %s of wrong type: *net.UDPConn
(although this seems to have been what it did before)

nameserver/mdns_lookup.go
squaremo: It's a fair cop. Fixing on master.

PR #343, opened 2015-01-13
improve connection logging

### LINKED ISSUE #340: Improve error reporting when TCP is working but UDP is blocked between peers
A user reported issues which manifested like this:

````
weave 2015/01/12 12:20:57.900126 Attempting connection to 92.222.68.183:6783
weave 2015/01/12 12:20:58.826744 Discovered local MAC 8a:00:aa:79:ad:0d
weave 2015/01/12 12:21:57.952029 ->[7a:78:e1:97:d0:30]: error: read tcp4 92.222.68.183:6783: i/o timeout
weave 2015/01/12 12:21:57.952216 ->[7a:78:e1:97:d0:30]: connection shutting down
````

The fact that it manages to report the other peer's name (7a:78:e1:97:d0:30) shows that the TCP handshake completed, but the lack of an "established active connection" log message suggests that the heartbeat frames aren't making it through on the udp channel.  Further investigation using tcpdump showed that no udp traffic was getting through.

For ease of diagnosis, it would be better if weave reported that no UDP traffic was received.

### PR DESCRIPTION
**Pull request by rade**
1) log more lifecycle events
  - acceptance
  - handshake completion
  - added & removed (from our connection map)
The latter is a bit verbose but may help us track down the cause for
#341. If we had different log levels, which is not something I am
going to introduce here, then this would fall in the 'debug' category.

2) better error logging
  - log error as part of "shutting down" log event rather than separately
  - indicate "blocked UDP connectivity" as a possible cause when
    getting a timeout on a not-fully-established connection.
The rationale for the latter is that inbound connections get
marked as established when first receiving a heartbeat frame over
UDP, which are the only frames that get transmitted prior to the
connection being fully established. Note that outbound
connections get marked as established when receiving the
ConnectionEstablished message over TCP, which is sent by the peer
when it first receives a heartbeat. See #342.

Closes #340.

### DIFF
--- a/router/local_peer.go
+++ b/router/local_peer.go
@@ -210,6 +210,7 @@ func (peer *LocalPeer) handleAddConnection(conn *LocalConnection) {
 	peer.Lock()
 	peer.connections[toName] = conn
 	peer.Unlock()
+	conn.log("connection added")
 }
 
 func (peer *LocalPeer) handleDeleteConnection(conn *LocalConnection) {
@@ -226,6 +227,7 @@ func (peer *LocalPeer) handleDeleteConnection(conn *LocalConnection) {
 	peer.Lock()
 	delete(peer.connections, toName)
 	peer.Unlock()
+	conn.log("connection deleted")
 	broadcast := false
 	if conn.Established() {
 		peer.Lock()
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: router/local_peer.go [hunk 1]
```diff
@@ -252,7 +254,7 @@ func (peer *LocalPeer) handleConnectionEstablished(conn *LocalConnection) {
 	peer.Lock()
 	peer.version += 1
 	peer.Unlock()
-	log.Println("Peer", peer.Name, "established active connection to remote peer", conn.Remote().Name, "at", conn.RemoteTCPAddr())
+	conn.log("connection fully established")
```
bboreham: We've lost the logging of the remote IP address here, which could be of interest.  Is the reasoning that we will have logged it earlier in Router.acceptTCP() ?

rade: correct.

bboreham: How does it work if we initiated the connection?

rade: you may want to take a look at the logs :)

"Attempting connection to ..."

Also, both ends log the IP/port at the end of handshake.

Issue #318, opened 2015-01-06
[dns] handle all query types


rade:
weavedns currently only handles A and AAAA queries. That is enough for most apps but not all, which is one reason preventing is from configuring container name resolution via weavedns by default.

So we should improve our nameserver implementation to handle all query types.

This in turn requires a departure from the current model where we handle A/AAAA queries by a simple lookup with `net.LookupIP`, since that won't work for any other query types. Instead we need to do proper query forwarding. Which [isn't hard](https://github.com/kenshinx/godns/blob/a3f7507c4a853ea9021ce596005831b668808d3b/resolver.go#L24) but requires that we figure out what nameserver to talk to. Ideally we'd read resolv.conf. Which [go has code for](https://golang.org/src/net/dnsconfig_unix.go?h=dnsReadConfig#L23) but annoyingly hides behind higher level APIs that are of no use to us.

bboreham:
More code to read resolv.conf [here](https://github.com/miekg/dns/blob/master/clientconfig.go#L22)

squaremo:
https://github.com/squaremo/weave/compare/zettio:master...squaremo:issue318_proxy_all_dns?expand=1

squaremo:
Reverse lookups are a problem: if we don't have the address locally, how do we know whether to ask our peers or to ask the internet? (weaveDNS doesn't presently attempt anything beyond checking the local zone database).

A not-completely-terrible solution might be to ask mDNS, and only ask the internet after it has timed out. A refinement of that might be to have an heuristic for which order to ask {internet, mDNS}, based on which addresses are in the DB. I'd have to think about how this affects correctness.

rade:
> A not-completely-terrible solution might be to ask mDNS, and only ask the internet after it has timed out.

Good enough for now, I think.

bboreham:
Closed because the bulk of this issue is done; remainder in new issue at #346

PR #344, opened 2015-01-13
Act as a proxy for non-weave DNS queries

squaremo:
weaveDNS now uses the servers in /etc/resolv.conf as fallbacks. It will ask the fallback server(s) if the query is for a name not under its authority; i.e., in .weave.local. It doesn't filter by query type, so as the smoke test shows, it will work for other kinds of question e.g., MX records.

As part of this, I've changed the names of the Lookup methods to get rid of the needless and inaccurate "Local" qualifier. Another factoring would be to have one method, pass through the whole question (or at least both the type and the name), and let the zone or whatever figure out how to answer it; but this will do for now.

### DIFF
--- a/nameserver/http_test.go
+++ b/nameserver/http_test.go
@@ -43,7 +43,7 @@ func TestHttp(t *testing.T) {
 	wt.AssertStatus(t, resp.StatusCode, http.StatusOK, "http response")
 
 	// Check that the address is now there.
-	foundIP, err := zone.LookupLocal(successTestName)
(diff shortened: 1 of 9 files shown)


rade:
Surely this closes #318. No?

rade:
Also, [apparently](https://github.com/zettio/weave/pull/328#issuecomment-69328870) it is also meant to 'partially' solve #327.

bboreham:
It mostly resolves #318 but reverse lookups still don't work for non-weave addresses, as noted in #327.

PR #345, opened 2015-01-13
Add tests for resolving names across hosts

squaremo:
It's more precise to use dig rather than ping. The stock ubuntu image
doesn't come with dig (in package dnsutils), but luckily Aanand made
an image which is exactly ubuntu+dnsutils.

### DIFF
--- a/test/200_dns_test.sh
+++ b/test/200_dns_test.sh
@@ -5,7 +5,7 @@
 C1=10.2.0.78
 C2=10.2.0.34
 
-start_suite "Resolve names"
+start_suite "Resolve names on a single host"
 
 run_on $HOST1 sudo $WEAVE stop || true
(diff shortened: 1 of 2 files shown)


rade:
I am concerned that we now have no tests that use the resolver library.

bboreham:
I was quite puzzled that this image you are using doesn't have `ping`.  It seems to be quite old, and the standard 'ubuntu' image has grown since then.

rade:
Perhaps one of the other 4782 ubuntu images on the docker hub is the one we want.

squaremo:
> I was quite puzzled that this image you are using doesn't have ping. It seems to be quite old, and the standard 'ubuntu' image has grown since then.

It has `/bin/sh` and `dig`, which are what I need. For ping I can just use the stock Ubuntu image.

Issue #346, opened 2015-01-13
[dns] reverse lookups for non-weave IP addresses give error response


bboreham:
For example, using an IP address that should be resolvable:

````
$ docker exec -i c1 sh -c "dig -x 8.8.8.8"
; <<>> DiG 9.8.1-P1 <<>> -x 8.8.8.8
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 47457
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;8.8.8.8.in-addr.arpa.		IN	PTR

;; Query time: 502 msec
;; SERVER: 172.17.42.1#53(172.17.42.1)
;; WHEN: Wed Jan 14 10:39:58 2015
;; MSG SIZE  rcvd: 38
````

See comments at https://github.com/zettio/weave/issues/318#issuecomment-68879397

PR #348, opened 2015-01-14
Use fallback nameserver for rdns

squaremo:
(Incidental change: turn debug logging on during smoke tests.)


rade:
This is meant to fix #346.

PR #349, opened 2015-01-15
better handling of undirectional udp connectivity

### LINKED ISSUE #342: connections marked as 'established' with unidirectional UDP connectivity
Inbound connections get marked as established when first receiving a heartbeat frame over UDP, which are the only frames that get transmitted prior to the connection being fully established. By contrast, outbound connections get marked as established when receiving the `ConnectionEstablished` message over TCP, which is sent by the peer when it first receives a heartbeat.

Consequently, when UDP is blocked in the opposite direction of TCP connection establishment, then the connection still gets marked as 'established' by both peers.

rade:
Currently we cannot distinguish bi-directional and uni-directional connectivity between peers. This is what currently happens for a connection from A to B...

A: remoteUDPAddr=X
B: remoteUDPAddr=nil
A: send fast heartbeats over UDP
B: receive heartbeat; remoteUDPAddr is nil
   remoteUDPAddr=Y
   send 'established' over TCP
   established=true
   send slow heartbeats over UDP
A: receive 'established' over TCP
   established=true
   send slow heartbeats over UDP

If udp packets make it through from A to B but not B to A then both ends mark the connection as 'established'. Conversely, if udp packets make it through from B to A but not A to B then the connection is not marked as 'established' at either end.

In the new scheme, this is what happens...

A: receivedHeartbeat=false, remoteUDPAddr=X
B: receivedHeartbeat=false, remoteUDPAddr=nil
A: send fast heartbeats over UDP
B: receive heartbeat; receivedHeartbeat is false
   remoteUDPAddr=Y
   receivedHeartbeat=true
   send 'established' over TCP
   send fast heartbeats over UDP
A: receive 'established' over TCP (*)
   established=true
   send slow heartbeats over UDP
A: receive heartbeat; receivedHeartbeat is false (*)
   receivedHeartbeat=true
   send 'established' over TCP
B: receive 'established' over TCP
   established=true
   send slow heartbeats over UDP

(*) either of these may occur first

Now, if udp packets make it through from A to B not not B to A then A marks the connection as 'established' but B doesn't. Conversely, if
udp packets make it through from B to A but not A to B then B marks the connection as 'established' but A doesn't.

Also, the output of `weave status` now indicates when a connection is unestablished.

Fixes #342.

### DIFF
--- a/router/types.go
+++ b/router/types.go
@@ -42,25 +42,26 @@ type RemoteConnection struct {
 type LocalConnection struct {
 	sync.RWMutex
 	RemoteConnection
-	TCPConn       *net.TCPConn
-	tcpSender     TCPSender
-	remoteUDPAddr *net.UDPAddr
-	established   bool
-	stackFrag     bool
-	effectivePMTU int
-	SessionKey    *[32]byte
-	heartbeat     *time.Ticker
-	fetchAll      *time.Ticker
-	fragTest      *time.Ticker
-	forwardChan   chan<- *ForwardedFrame
-	forwardChanDF chan<- *ForwardedFrame
-	stopForward   chan<- interface{}
-	stopForwardDF chan<- interface{}
-	verifyPMTU    chan<- int
-	Decryptor     Decryptor
-	Router        *Router
-	UID           uint64
-	queryChan     chan<- *ConnectionInteraction
+	TCPConn           *net.TCPConn
+	tcpSender         TCPSender
+	remoteUDPAddr     *net.UDPAddr
+	established       bool
+	receivedHeartbeat bool
+	stackFrag         bool
+	effectivePMTU     int
+	SessionKey        *[32]byte
+	heartbeat         *time.Ticker
+	fetchAll          *time.Ticker
+	fragTest          *time.Ticker
+	forwardChan       chan<- *ForwardedFrame
+	forwardChanDF     chan<- *ForwardedFrame
+	stopForward       chan<- interface{}
+	stopForwardDF     chan<- interface{}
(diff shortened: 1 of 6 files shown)


router/types.go
```diff
@@ -42,25 +42,26 @@ type RemoteConnection struct {
...
+	forwardChan       chan<- *ForwardedFrame
+	forwardChanDF     chan<- *ForwardedFrame
+	stopForward       chan<- interface{}
+	stopForwardDF     chan<- interface{}
+	verifyPMTU        chan<- int
+	Decryptor         Decryptor
+	Router            *Router
+	UID               uint64
+	queryChan         chan<- *ConnectionInteraction
 }
```
squaremo: THANKS ROB PIKE

rade:
Note this actually changes how/when topology gets communicated. I've only just realised that; oops.

The reason for the change is that when a peer receives ConnectionEstablished, it sends a FetchAll.

Previously, for a new connection only the remote peer would send a ConnectionEstablished message, and it did so after marking its own connection as established. Hence, on receiving ConnectionEstablished the initiating peer would send FetchAll to the remote peer and get back a topology that includes the new connection in the remote peer's connection list.

Now, both peers send ConnectionEstablished, and only mark their own connection as established when receiving such a message. Consequently FetchAll is sent by both peers to each other, and the answer may not include the new connection. This is ok logically since - previously and now - marking a connection as established results in the broadcast of an Update, so the two peers learn about each other's half of the new connection that way. It is possibly less efficient, but note that in the old version peers would typically receive an Update containing unknown peers and hence have to issue a FetchAll anyway.

PR #355, opened 2015-01-20
better abstraction for protocol messages

rade:
- Instead of representing protocol messages as []byte, represent them as instances of a ProtocolMsg struct, containing a ProtocolTag identifying the message type, and a []byte for the payload.
- Introduce ProtocolSender interface, containing a SendProtocolMsg method. This is implemented by LocalConnection and LocalPeer, replacing the former SendTCP/BroadcastTCP methods. In the former case the ProtocolMsg is sent to the peer at the other end of the connection, in the latter case it is sent to all peers the local peer is connected to.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -200,10 +200,10 @@ func (conn *LocalConnection) SetEstablished() {
 }
 
 // Async
-func (conn *LocalConnection) SendProtocolMsg(tag ProtocolMsg, msg []byte) {
+func (conn *LocalConnection) SendProtocolMsg(m ProtocolMsg) {
 	conn.queryChan <- &ConnectionInteraction{
 		Interaction: Interaction{code: CSendProtocolMsg},
-		payload:     TaggedProtocolMsg{tag, msg}}
+		payload:     m}
 }
 
 // ACTOR server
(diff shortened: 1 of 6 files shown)


rade:
Possible points of contention:

1. I decided to make LocalPeer implement ProtocolSender. Arguably, SendProtocolMsg is less descriptive than BroadcastTCP, but IMO the uniformity is preferable and there is little scope for confusion.

2. I am passing ProtocolMsg by value instead of reference. Doing so avoids some type noise. It's a tiny struct, so some copying shouldn't cost a lot, and in any case the tcp communication is not on the performance critical path.

3. I didn't go the whole hog and switch to representing each message as a separate type. Doesn't seem worth it and would have been a bigger change, including a bump in the protocol version.

4. There's a lot of `Protocol` in some code fragments now, e.g. `ne.conn.SendProtocolMsg(ProtocolMsg{ProtocolNonce, encodedNonce})`. I can't think of a way around that; leaving out `Protocol` in any of the places there would be prone to clashes, since we'd be left with very generic terms.

bboreham:
"Protocol" doesn't seem a very helpful word; almost everything we deal with is a protocol - UDP, TCP, DNS, etc.  It is the _Weave_ protocol, but adding that word will make things more unwieldy.

How about 'Control' - ControlMsg, SendControlMsg?  However I don't like changing the tags to ControlNonce, ControlUpdate, etc.  Maybe just take the prefix off them - just FragmentationReceived, PMTUVerified, etc.

rade:
I don't think there is any confusion what protocol we are talking about here.

As for dropping the prefix; the remainder would be too generic and prone to clashes. Think 'Nonce', 'Update', for example.

PR #357, opened 2015-01-21
Add create-bridge command

errordeveloper:
This will should allow one to use an unmodified version of the script to create the bridge so it can be used with `docker --bridge=weave`, if required.

### DIFF
--- a/weave
+++ b/weave
@@ -139,11 +139,13 @@ create_bridge() {
         run_iptables -t nat -N WEAVE >/dev/null 2>&1 || true
         add_iptables_rule nat POSTROUTING -j WEAVE
     }
-    # disable offloading - we do this even when $BRIDGE already exists
-    # simply because it has the desirable side-effect of ensuring we
-    # have the $TOOLS_IMAGE and thus do not incur downloading-induced
-    # delays in more time-sensitive sections elsewhere.
-    run_tool host ethtool -K $BRIDGE tx off >/dev/null
+    if [ ! "$1" = "--without-ethtool" ] ; then
+        # disable offloading - we do this even when $BRIDGE already exists
+        # simply because it has the desirable side-effect of ensuring we
+        # have the $TOOLS_IMAGE and thus do not incur downloading-induced
+        # delays in more time-sensitive sections elsewhere.
+        run_tool host ethtool -K $BRIDGE tx off >/dev/null
+    fi
 
     ip link set dev $BRIDGE up
 }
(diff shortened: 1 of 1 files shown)


sttts:
:+1:

rade:
What's the use case?

errordeveloper:
> What's the use case?

@rade see updated description.

rade:
Right, but, as you discovered yourself, create_bridge requires a running docker in order to invoke ethtool via the weavetools container. So I don't see how this PR helps in the described use case.

sttts:
Is there a reason to call ethtool through docker? I mean other than that it might be not available on the host machine and via the weavetools container and --net=host we can still run it?

rade:
The whole point of the weavetools container is to reduce dependencies on what is installed on the host.

errordeveloper:
> Right, but, as you discovered yourself, create_bridge requires a running docker in order to invoke ethtool via the weavetools container. So I don't see how this PR helps in the described use case.

Sure, but this is not intending to provide complete functionality, user still need to set take care of changing docker options and restarting it. It's no even a documented options, so it's for quite advanced users.

One thing to consider would be to spilt the tail of `cereate_bridge()` function.

Actually, we do this right now:
```
run_tool host ethtool -K $BRIDGE tx off >/dev/null
ip link set dev $BRIDGE up
```

I don't think `ethtool` has too run before passing the bridge to docker, so it would be okay to call it later. I would need to test whether `ip link ... up` would be fine to move as well, probably yes...

So I'd suggest to move the main part of  `create_bridge()` into a another function and call it from `create_bridge()` as well as the `create-bridge` subcommand.

rade:
> I would need to test whether `ip link ... up`  would be fine to move as well

That strikes me as wrong.

> One thing to consider would be to spilt the tail of `create_bridge()` function.

I can't think of a sensible name for that function, so instead suggest adding an optional param to `create_bridge`, which, if present, suppresses the ethtool invocation.

errordeveloper:
>I can't think of a sensible name for that function, so instead suggest adding an optional param to create_bridge, which, if present, suppresses the ethtool invocation.

That was my second thought actually.

errordeveloper:
@rade I updated the commit as we discussed.

tomwganem:
I'm trying to use `weave create-bridge` before running docker. It doesn't seem to work, but this pull request seems to indicate that it should. Could someone clarify this for me?

```
==> weave create-bridge
2015/06/17 14:17:37 Cannot connect to the Docker daemon. Is 'docker -d' running on this host?
```

bboreham:
Nowadays you need to say `weave --local create-bridge` otherwise it will try to fetch its dependencies via Docker.

However, please re-assess why you want to tell Docker to use the weave bridge. If it is for IP allocation, weave has that built-in now.

tomwganem:
I've been using http://sttts.github.io/docker/weave/mesos/2015/01/22/weave.html as a guide for setting up a mesos cluster running marathon that creates docker containers that need to communicate with each other. I am using consul and `progrium/registrator` for service discovery and dns.

If I don't have to use the weave bridge, that's less setup for me to worry to about.

PR #359, opened 2015-01-21
Obtain status info in JSON format

bboreham:
Json generation is done by creating dummy structs and arrays that match the format we want, copying the information into them, then getting Go's standard json encoder to do the work.

All the implementation is pulled into one file json.go rather than scattering it across all the relevant implementations; at time of writing this seems easier to grasp.

### DIFF
--- a/weaver/main.go
+++ b/weaver/main.go
@@ -145,6 +145,11 @@ func handleHttp(router *weave.Router) {
 		io.WriteString(w, fmt.Sprintln("Encryption", encryption))
 		io.WriteString(w, router.Status())
 	})
+	http.HandleFunc("/status-json", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
+		json, _ := router.MarshalJSON()
+		w.Write(json)
+	})
 	http.HandleFunc("/connect", func(w http.ResponseWriter, r *http.Request) {
 		peer := r.FormValue("peer")
 		if addr, err := net.ResolveTCPAddr("tcp4", weave.NormalisePeerAddr(peer)); err == nil {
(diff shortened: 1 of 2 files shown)


rade:
@squaremo mind taking a look at this?

weaver/main.go
```diff
@@ -145,6 +145,11 @@ func handleHttp(router *weave.Router) {
 		io.WriteString(w, fmt.Sprintln("Encryption", encryption))
 		io.WriteString(w, router.Status())
 	})
+	http.HandleFunc("/status-json", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
```
squaremo: This looks dodgy. It means any site can request this document (by guessing, e.g., a docker IP) and thereby get information on the weave network. Presently the HTTP endpoint is only exposed to the host network, but this way lies vulnerabilities.

squaremo:
What's the argument for using specialise methods rather than giving the struct definitions tags -- just that the data aren't quite organised in the right way?

bboreham:
Taking the structs one at a time:
`Router` - need to specify the way the Interface is encoded; can't directly override what net.Interface does.
`MacCache` - the `table` is keyed on a `uint64` encoding of the mac address; this would look ugly.  Also it is not exported so go's json encoder would ignore it.
`Peers` - table is not exported, and want to render the `*Peer` references as names only.
`Routes` - again, fields are not exported, and while the default rendering would contain the right information it will be in a hard-to-understand format, just 'peername : peername' for each entry.
`Peers` - changing `connections` from a map into an array
`RemoteConnection` - fields are not exported, and want to render the `*Peer` references as names only.

bboreham:
I notice that each Peer's UID and version number is not included in the output, whereas it is in `weave status`

bboreham:
@squaremo I added top-level version and encryption info

Issue #360, opened 2015-01-23
[dns] "buffer size too small" error


rade:
````
--- FAIL: TestDNSServer (1.53s)
	util.go:69: Unexpected error: dns: buffer size too small
		goroutine 74 [running]:
		github.com/zettio/weave/testing.stackTrace(0xa23700, 0x0, 0x0)
			/home/matthias/go/src/github.com/zettio/weave/testing/util.go:78 +0x88
		github.com/zettio/weave/testing.StackTrace(0x0, 0x0)
			/home/matthias/go/src/github.com/zettio/weave/testing/util.go:73 +0x35
		github.com/zettio/weave/testing.Fatalf(0xc20806cd80, 0x863650, 0x14, 0xc2080c3de8, 0x1, 0x1)
			/home/matthias/go/src/github.com/zettio/weave/testing/util.go:69 +0x27
		github.com/zettio/weave/testing.AssertNoErr(0xc20806cd80, 0x2aafcb1cf0f8, 0xc20800a340)
			/home/matthias/go/src/github.com/zettio/weave/testing/util.go:12 +0xea
		github.com/zettio/weave/nameserver.TestDNSServer(0xc20806cd80)
			/home/matthias/go/src/github.com/zettio/weave/nameserver/server_test.go:75 +0xb79
		testing.tRunner(0xc20806cd80, 0xa20338)
			/usr/local/go/src/testing/testing.go:447 +0xbf
		created by testing.RunTests
			/usr/local/go/src/testing/testing.go:555 +0xa8b
DEBUG: 2015/01/23 11:25:53.989234 [zonedb] Looking for address: [10 0 2 1]
FAIL
exit status 1
FAIL	github.com/zettio/weave/nameserver	2.847s
````

This doesn't fail consistently, so clearly it depends on environmental conditions.

rade:
@bboreham mentioned something about our dns server not implementing TCP fallback. Does that mean the above failure may be encountered in ordinary deployments and cause problems for our users?

bboreham:
Weavedns is using `dns.Exchange` which applies a buffer size of 512 bytes.  We could call via a `Client` object to increase this number; as long as we stay under 1400 bytes it should work in most circumstances.

However, it is still technically possible to have a ginormous DNS reply that requires TCP fallback.

rade:
This should have been closed when #361 was merged. Doing so now.

PR #361, opened 2015-01-23
Use a larger buffer than the default, so that more queries will succeed

bboreham:
Addresses #360.  A complete fix would involve implementing TCP fallback, but raising the limit from 512 bytes to 1400 should accommodate a large fraction of real-world responses.

The 'unit' test makes the assumption that type ANY on cisco.com will give a large response.  Empirically you also have to add the EDNS0 record in order to get a long response - experiment with commands like `dig ANY cisco.com. +qr +noedns`

### DIFF
--- a/nameserver/server_test.go
+++ b/nameserver/server_test.go
@@ -10,6 +10,12 @@ import (
 	"time"
 )
 
+const (
+	testRDNSsuccess  = "1.2.0.10.in-addr.arpa."
+	testRDNSfail     = "4.3.2.1.in-addr.arpa."
+	testRDNSnonlocal = "8.8.8.8.in-addr.arpa."
+)
+
 func TestDNSServer(t *testing.T) {
 	const (
 		port            = 17625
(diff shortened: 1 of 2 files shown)


rade:
This is intended to fix #360, right?

Presumably we should open a separate issue to implement TCP fallback?

bboreham:
Edited description and added #362.

nameserver/server_test.go
```diff
@@ -77,6 +78,15 @@ func TestDNSServer(t *testing.T) {
 	if !(len(r.Answer) > 0) {
 		t.Fatal("Number of answers > 0")
 	}
+	// Now ask a query that we expect to return a lot of data.
+	m.SetQuestion("cisco.com.", dns.TypeANY)
+	m.SetEdns0(UDPBufSize, false) // Special magic to coax the other end into complying
```
bboreham: note that this actually defeats the point of the test, because the dns library will allocate a larger buffer for queries with an EDNS0 record requesting it.

squaremo: Why have it then?

bboreham: It was my best attempt at forcing a large reply.  I subsequently discovered that doing it this way invalidates the test, and made the above comment.

rade:
@bboreham what is happening with this?

Issue #373, opened 2015-01-28
detect udp connectivity breakage


rade:
Weave is able to detect when there is no udp connectivity at the time of connection establishment. Modulo #372. But we do not detect when it breaks *subsequently*. There are heartbeats going across udp which should help prevent it from breaking, but we have no logic in place to *detect* such breakages and act on them (e.g. by tearing down the connection). This can leave us with permanently broken connectivity, for example in the event of a firewall forgetting a udp NAT path.

PR #376, opened 2015-01-29
tear down connections that have no (initial) udp connectivity

### LINKED ISSUE #372: udp connectivity failure detection no longer working
Following #313, the [udp connectivity failure detection](https://github.com/zettio/weave/blob/5dc9b085a018e639cfc05a6d63694517784b3c07/router/connection.go#L257) is no longer working.

Previously, a tcp connection would not see regular communication until it had been marked as 'established', because in the topology communication scheme would periodically send 'FetchAll' only after that. So in the absence of udp connectivity we'd time out. The timeout could of course occur for other reasons, namely disrupted *tcp* connectivity, but the *combination* of a timeout and 'not established' points at lack of udp connectivity as a likely cause.

In the new gossip-based topology communication, however, gossip (of anything, not just topology) occurs as soon as we have tcp connectivity. Therefore we will no longer encounter a timeout when *just* udp connectivity is disrupted. Which makes our test invalid.

rade:
Prior to the introduction of gossip, a tcp connection would not see
regular communication until it had been marked as 'established',
because the topology communication scheme would periodically send
'FetchAll' only after that. So in the absence of udp connectivity we'd
time out. The timeout could of course occur for other reasons, namely
disrupted tcp connectivity, but the combination of a timeout and 'not
established' pointed to a lack of udp connectivity as a likely
cause. So that's what we checked for.

In the new gossip-based topology communication, however, gossip (of
anything, not just topology) occurs as soon as we have tcp
connectivity. Therefore we will no longer encounter a timeout when
just udp connectivity is disrupted. Which makes our check invalid.

So instead we introduce a one-shot timer, created on handshake
completion and cancelled when we mark the connection as
'established'. When the timer fires, and the connection still hasn't
been established, we shut down the connection with an error indicating
lack of udp connectivity.

Fixes #372.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -38,26 +38,27 @@ type RemoteConnection struct {
 type LocalConnection struct {
 	sync.RWMutex
 	RemoteConnection
-	TCPConn           *net.TCPConn
-	tcpSender         TCPSender
-	remoteUDPAddr     *net.UDPAddr
-	established       bool
-	receivedHeartbeat bool
-	stackFrag         bool
-	effectivePMTU     int
-	SessionKey        *[32]byte
-	heartbeatFrame    *ForwardedFrame
-	heartbeat         *time.Ticker
-	fragTest          *time.Ticker
-	forwardChan       chan<- *ForwardedFrame
-	forwardChanDF     chan<- *ForwardedFrame
-	stopForward       chan<- interface{}
-	stopForwardDF     chan<- interface{}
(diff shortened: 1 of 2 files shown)


dpw:
Out of interest, what is supposed to happen if we lose UDP connectivity after the connection is considered established?

rade:
> Out of interest, what is supposed to happen if we lose UDP connectivity after the connection is considered established

That is covered by #373.

PR #377, opened 2015-01-29
Also avoid calling ethtool in expose and hide

sttts:
Those two commands are used in the pre-docker phase when you setup your
weave bridge using create-bridge, e.g. inside /etc/network/interfaces.

This extends #357.


rade:
Makes sense. Thanks. Merged.

sttts:
Cool! Thanks!

PR #386, opened 2015-02-03
exposing command from usage. see https://github.com/zettio/weave/pull/35...

Lawouach:
I guess that was meant to be exposed.

### DIFF
--- a/weave
+++ b/weave
@@ -6,6 +6,7 @@ SCRIPT_VERSION="(unreleased version)"
 usage() {
     echo "Usage:"
     echo "weave setup"
+    echo "weave create-bridge"
     echo "weave launch     [-password <password>] <peer> ..."
     echo "weave launch-dns <cidr>"
     echo "weave connect    <peer>"
--- a/weave
+++ b/weave
@@ -7,6 +7,7 @@ usage() {
     echo "Usage:"
     echo "weave setup"
     echo "weave create-bridge"
+    echo "weave destroy-bridge"
     echo "weave launch     [-password <password>] <peer> ..."
     echo "weave launch-dns <cidr>"
     echo "weave connect    <peer>"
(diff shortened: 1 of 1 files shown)


Lawouach:
sigh. I forgot to create branches to generate several PR. sorry for the cherry picking.

rade:
These commands are intentionally undocumented.

Lawouach:
Well since they've been blogged about, it seems rather strange not to publicly expose them. But fair enough. Maybe a comment in the script?

rade:
> Maybe a comment in the script?

Maybe.

Why did you add `destroy-bridge`? Just for symmetry?

Lawouach:
Indeed. It only stands to reason that you'd want to be able to script both ways.

rade:
>  It only stands to reason that you'd want to be able to script both ways.

We already have `weave reset`.

Lawouach:
True, although it does a little bit more than simply destroying the bridge.

I'd have actually called it "weave shutdown (or terminate)" since we initialize weave through "weave launch". Or I'd renamed "weave launch" to "weave init". 

But anyway, my point is about cognitive perspective. I usually expect to have a command to undo an action (probably an old testing habit ;)).

rade:
`shutdown` and `terminate` are too close to `stop`.

Why don't we rename all commands? I am sure our users would be delighted ;)

I get the symmetry/undo argument, but every additional command, even one half of a symmetric pair, increases cognitive load too (with the possible exception of having a completely symmetric command set). So to add `destroy-bridge` there has to be a reason for the command to exist *other than* symmetry. And I cannot think of one.

Lawouach:
Fair enough.

rade:
> Maybe a comment in the script

https://github.com/zettio/weave/commit/f01d9ce09225f683c4c572199730dc7bbef6a787

PR #387, opened 2015-02-04
Containerized weave build

dpw:
This adds a few tweaks and conveniences to support building weave in a container, and documents the process.   The documentation covers how to build under a Fedora docker host, and so might satisfy the users who wanted to build under Fedora.


rade:
Would it make sense to volume-mount the checked out repository in the build container? Currently it's always building from a git-cloned repo rather than building the checked out code, which is kinda surprising.

dpw:
> Would it make sense to volume-mount the checked out repository in the build container?

Yes.  If only it worked under Fedora/RHEL (with SElinux enabled, volume mounts still "work", but most of the time you can't access anything under the mount point).

I guess supporting both approaches is possible.  E.g. the image can check whether $GOPATH/src/github.com/zettio/weave exists, and only clone if it doesn't.  Would that be better?

rade:
> with SElinux enabled, volume mounts still "work", but most of the time you can't access anything under the mount point

I see. It might be worth sticking a note in the code about that.

> I guess supporting both approaches is possible. E.g. the image can check whether $GOPATH/src/github.com/zettio/weave exists, and only clone if it doesn't. Would that be better?

Yes, though it's possibly a bit too magic and it would be easy to accidentally build the wrong thing, e.g. when mistyping the destination in the volume mount. Add a flag, perhaps?

dpw:
> > I guess supporting both approaches is possible. E.g. the image can check whether $GOPATH/src/github.com/zettio/weave exists, and only clone if it doesn't. Would that be better?
> 
> Yes, though it's possibly a bit too magic and it would be easy to accidentally build the wrong thing, e.g. when mistyping the destination in the volume mount. Add a flag, perhaps

Ok.  How about forwarding any options to git-clone?  That will remain the need for the explicit WEAVE_REPO and WEAVE_BRANCH environment settings.  And if no options are provided, it can expect the bind mount.

rade:
> Ok. How about forwarding any options to git-clone? That will remain the need for the explicit WEAVE_REPO and WEAVE_BRANCH environment settings. And if no options are provided, it can expect the bind mount.

Sounds good.

dpw:
Ok, I have respun the branch due to significant changes.  It's ready for another look.

dpw:
As requested, this now supports bind-mounting your gopath into the container.  And intermediate files are left around (a no-changes rebuild takes 15 seconds for me).

tools/build.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
-set -e
+set -x -e
```
rade: -x?

build/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu
+# After installing the prerequisite we clear out sources.list so that
+# 'apt-get update' doesn't do anything.
```
rade: why is that necessary/desirable? Please explain that in the comment.

tools/build.sh
dpw: It's nice to have some indication of where that script got to, should it fail.  Given the amount of output it already  produces when it succeeds, printing a little more seems acceptable.

rade: ok. I was wondering whether this had been left over from debugging. perhaps add a comment?

dpw:
Ok, one comment added, and one comment elaborated.

rade:
I've made some tweak to the docs. See https://github.com/rade/weave/tree/dpw-build_in_container

One problem I ran into is that `make prequisites` installs the docker.io package, which replaced my lxc-docker 1.5.0 with an ancient 1.0.1. The incantations to use lxc-docker are in our Vagrantfile.

dpw:
> One problem I ran into is that make prequisites installs the docker.io package, which replaced my lxc-docker 1.5.0 with an ancient 1.0.1.

So, that is clearly a problem.

But switching the containerized build to use lxc-docker 1.5.0 also presents a problem, because in the container we only use docker as a client, and if the docker daemon on the host is using an earlier version, the client reports a version mismatch.  In contrast, using the ancient 1.0.1 version of docker.io in the container is fine: It has all the client functionality we need.

So build/Dockerfile should remain as it was, installing the ancient docker.io.  But perhaps 'make prerequisites' should be modified to install lxc-docker 1.5.0.  For that to work, the 'prerequisites' makefile target can't be a dependency of 'build'.  So 'make prerequisites' ends up being superfluous to this PR (but it was kind of superfluous already).

But what about the possibility that someone does 'make prerequisites' to build weave on their host, but already has an older version of docker installed, and is unpleasantly surprised when if gets upgraded?  Perhaps docker should be removed from the prerequisite package list entirely, and we should change the docs to tell people to take care of it themselves.

If I'd reached this point on my own, I'd simply give up on 'make prerequisites' and remove it.  I don't have much use for it myself. If you think it is useful, tell me what it should do, and I'll make it do that.

dpw:
Need to fix markdown syntax errors.

dpw:
> Need to fix markdown syntax errors.

Done.

dpw:
> Need to fix markdown syntax errors.

Done.

bboreham:
Please can you also fix the smoke tests to cope with the tarfiles having moved

Issue #45, opened 2014-09-11
support systems with older versions of `iptables` and `ip`


kenshin54:
I got error `RTNETLINK answers: Operation not supported` while running `weave launch 10.0.0.1/16`. Weave script use ip and iptables command to create bridge and nat rule. It seems ip command doesn't support bridge type on CentOS 6.5. And iptables doesn't support `-C` flag also.

Maybe we can use brctl as a backup method to create a bridge and check nat rule with delete then append to mitigate the issue.

bboreham:
Docker now requires at least kernel 3.10; closing.

PR #388, opened 2015-02-05
Move weave script into tools image with its deps

### LINKED ISSUE #312: run weave on remote docker
Say I want to run weave on two remote hosts that I'm connected to via there docker daemon, could I invoke the weave router on them via docker (without running commands via SSH)

something like 
DOCKER_HOST=tcp://someremote:2375 docker run weaveimg weave launch 
DOCKER_HOST=tcp://otherremote:2375 docker run weaveimg weave launch someremote

I assume if it can work at all, i would have to user --privileged and -host and maybe other things ..
In others words can weave itself be run within a container ?

Is that possible at all ? is that documented anywhere ?

Thank you.

squaremo:
This rearrangement means that the `weave` command can be used
remotely; and, as a side-effect, the script that does all the work is
using executables in its own filesystem, rather than invoking them via
docker.

A choice here is whether to base the tools image on scratch and add
all the deps, or base it on something else. I've used alpine linux,
which itself is based on busybox. This gives us the usual tools (sed
etc.) while keeping the image size small.

Adding the docker binary adds substantially to the image size, but it
cannot always be e.g., bind-mounted. To mitigate this image bloat, we
might build or otherwise get a client-only docker executable, or write
a special-purpose client, or indeed rewrite the driver script and use
a client library.

Closes #312, improves on #321


rade:
- The definition of TOOLS_IMAGE has been removed from the driver script, but it is still being used - by `setup` and `version`. This also raises the question of what these two commands should do now. `setup` is meant to download all required images for the version corresponding to the script. But the driver script is part of an image. So do we need a version on the new `weave` script, and run the setup command there?

- Surely the release script needs updating, to 'sed' the version number into the 'driver' script. Or we should get rid of that code and only have the tools image version.

- We need to update the docs to explain that the weave script and docker daemon do not need to be co-located (our installation instructions do kinda imply that) when DOCKER_HOST is set.

- I can think of one thing that will break for sure: the kind of scenario for which we introduced `create-bridge` and made `hide` and `expose` not depend on docker. See #357 and #377. It seems to me that the only way to solve that would be to have alternate mode of operation where the the weave script executes everything locally, as before. Or at least those three commands (w/o duplicating the code).

errordeveloper:
> I can think of one thing that will break for sure: the kind of scenario for which we introduced create-bridge and made hide and expose not depend on docker. See #357 and #377. It seems to me that the only way to solve that would be to have alternate mode of operation where the the weave script executes everything locally, as before. Or at least those three commands (w/o duplicating the code).

In terms of `create-bridge`, one can install the driver script on the host...

Looking the diff between weave script in master and driver script here:

```diff
47d46
< BASE_TOOLS_IMAGE=zettio/weavetools
50d48
< TOOLS_IMAGE=$BASE_TOOLS_IMAGE:$IMAGE_VERSION
72,78d69
< run_tool() {
<     TOOL_NET=$1
<     TOOL_COMMAND=$2
<     shift 2
<     docker run --rm --privileged --net=$TOOL_NET $TOOLS_IMAGE /bin/$TOOL_COMMAND "$@"
< }
< 
143,147c134
<         # disable offloading - we do this even when $BRIDGE already exists
<         # simply because it has the desirable side-effect of ensuring we
<         # have the $TOOLS_IMAGE and thus do not incur downloading-induced
<         # delays in more time-sensitive sections elsewhere.
<         run_tool host ethtool -K $BRIDGE tx off >/dev/null
---
>         ethtool -K $BRIDGE tx off >/dev/null
228c215
<     if run_tool host ethtool -K $GUEST_IFNAME tx off >/dev/null &&
---
>     if ethtool -K $GUEST_IFNAME tx off >/dev/null &&
247c234
<         run_tool container:$CONTAINER ethtool -K eth0 tx off >/dev/null &&
---
>         ip netns exec $NETNS ethtool -K eth0 tx off >/dev/null &&
327c314
<     run_tool host curl -s -X $http_verb "$@" http://$ip:$port$url
---
>     curl -s -X $http_verb "$@" http://$ip:$port$url
622c609
<         run_tool host conntrack -D -p udp --dport $PORT >/dev/null 2>&1 || true
---
>         conntrack -D -p udp --dport $PORT >/dev/null 2>&1 || true
637c624
<         run_tool host conntrack -D -p udp --dport $PORT >/dev/null 2>&1 || true
---
>         conntrack -D -p udp --dport $PORT >/dev/null 2>&1 || true

```

All I can see is `run_tool` being removed. So dual-mode driver script can be done by means of re-introducing `run_tool` it with a bypass mode. The only somewhat tricky part seems to be:
```
247c234
<         run_tool container:$CONTAINER ethtool -K eth0 tx off >/dev/null &&
---
>         ip netns exec $NETNS ethtool -K eth0 tx off >/dev/null &&
```

rade:
Another thing that breaks: env vars, specifically any env vars that the current weave script pays attention to. Which are:

- VERSION
- WEAVE_DOCKER_ARGS
- DOCKER_BRIDGE
- PROC_FS

squaremo:
> The definition of TOOLS_IMAGE has been removed from the driver script, but it is still being used - by setup and version. This also raises the question of what these two commands should do now. setup is meant to download all required images for the version corresponding to the script. But the driver script is part of an image. So do we need a version on the new weave script, and run the setup command there?

The `weave` script is the entry point, so I think that is what has to be versioned ultimately. Then the images are locked to that version -- i.e., what happens now. The version is then in two places: in the driver script, so it asks for the correct images, and in the weave script. The weave script simply asks for the versioned tools image, which has its version built-in.

We *could* alter things so that the outer script knew about versioning, so that you could do e.g., `weave setup 0.9.0` and have it install the correct images. But that's a bunch more work and I'm not sure we want to second-guess packaging.

> Surely the release script needs updating, to 'sed' the version number into the 'driver' script. Or we should get rid of that code and only have the tools image version.

Yes, it needs updating.

> We need to update the docs to explain that the weave script and docker daemon do not need to be co-located (our installation instructions do kinda imply that) when DOCKER_HOST is set.

Yes.

> I can think of one thing that will break for sure: the kind of scenario for which we introduced create-bridge and made hide and expose not depend on docker. See #357 and #377. It seems to me that the only way to solve that would be to have alternate mode of operation where the the weave script executes everything locally, as before. Or at least those three commands (w/o duplicating the code).

I have to admit I don't see how that can work without some contortions (e.g., put those commands in the weave script and include it in the image too, then call it from the driver when necessary).

squaremo:
> Another thing that breaks: env vars, specifically any env vars that the current weave script pays attention to. Which are:
> 
>     VERSION
>     WEAVE_DOCKER_ARGS
>     WEAVE_PASSWORD
>     DOCKER_BRIDGE
>     PROC_FS

These would need to be passed as `-e` arguments if present. The PROC_FS already gets passed, since it's tied to how the tools container is started. I'll have to look at the rationale for the password being in the environment -- it may not be acceptable to pass it over to docker as a command line arg.

squaremo:
> It seems to me that the only way to solve that would be to have alternate mode of operation where the the weave script executes everything locally, as before.

The logic in `weave` is to always forward commands to a docker container, but it could check if there's a  docker URL and assume it can effect things locally if it's absent or a unix domain socket (or have a special `--local` option).

If these must be special-cased anyway, it's then not such a big deal to include the weave script in the tools image and use it for these particular commands.

squaremo:
> WEAVE_PASSWORD

The aim of moving this into an environment variable is to give people a way to supply it without putting it on the command line. To achieve the same thing with the remote-actuation weave script, the password option parsing should be moved from the driver script to the weave script, and the driver script should receive it in an environment variable which is passes on to the container.

rade:
>  it could check if there's a docker URL and assume it can effect things locally if it's absent

That's what I was thinking. Doesn't mean it's the best approach though ;)

squaremo:
> it could check if there's a docker URL and assume it can effect things locally if it's absent
> 
> That's what I was thinking. Doesn't mean it's the best approach though ;)

I favour the `--local` alternative, since these are niche uses.

squaremo:
> These would need to be passed as -e arguments if present.

<var>VERSION</var>, <var>WEAVE_DOCKER_ARGS</var>, <var>DOCKER_BRIDGE</var>, and <var>WEAVE_PASSWORD</var> can all get passed directly along.

In the case of <var>WEAVE_PASSWORD</var>, there's no need to do the arg parsing first as described above, since it will pass along any arguments *as well as* the environment variable if present.

> The PROC_FS already gets passed, since it's tied to how the tools container is started.

An elaboration: when acting as a remote, <var>PROCFS</var> refers to the host filesystem that should be mapped as a volume to `/hostproc`; when acting locally, it refers to the proc filesystem with namespaces etc. So when running the tools container remotely, the host proc filesystem is mounted as `/hostproc`, and `/hostproc` is passed as the value of <var>PROCFS</var>.

squaremo:
<var>VERSION</var> is a tricky one.

If it determines the weavetools image to run when acting remotely, it will mostly fail since earlier versions aren't set up to be called remotely -- for one thing they don't have the script in the container image.

If the *latest* weavetools image is always invoked remotely, it seems against the spirit of using VERSION in the first place, although it will do the right thing mostly, which is to use weave and weavedns images of the given version.

bboreham:
How does a user run this?  Like `docker run zettio/weave weave run ...` ?

squaremo:
> How does a user run this? Like docker run zettio/weave weave run ... ?

As before, the user is expected to download the weave script and call `weave whatever`. The difference is that it will forward the call to the weavetools image on a remote docker instance, if a DOCKER_HOST is given.

(There's also the option of driving weavetools directly with the docker CLI, but that needs you to get a bunch of docker options exactly right)

squaremo:
> This also raises the question of what these two commands should do now. setup is meant to download all required images for the version corresponding to the script. But the driver script is part of an image. So do we need a version on the new weave script, and run the setup command there?

I am trying to figure out what we *want* to happen, and I think it's this:

If the invoked weave script has a version, or is given a version, it should try to call that remotely; this means

  * That version of weavetools will be started (downloaded, possibly) to run the command
  * It will setup or start that version of weave and weavedns

A problem is that if you want to remote control a version 0.8 weave, it'll fail because weavetools v0.8 doesn't have the weave script and therefore cannot be called remotely.

rade:
Making remoting work with old weaves is a nice-to-have at best.

squaremo:
This is ready for another go-round. I'm going to test the release procedure, since I've changed the image name and so on.

squaremo:
> This is ready for another go-round.

Oh except now you merged #387, I'll have to rebase.

squaremo:
> Oh except now you merged #387, I'll have to rebase.

Rebased. Ready for another look.

weave
```diff
@@ -27,27 +36,44 @@ usage() {
...
+
+BASE_IMAGE=$DOCKERHUB_USER/weave
+BASE_DNS_IMAGE=$DOCKERHUB_USER/weavedns
+BASE_EXEC_IMAGE=$DOCKERHUB_USER/weaveexec
+IMAGE=$BASE_IMAGE:$IMAGE_VERSION
+DNS_IMAGE=$BASE_DNS_IMAGE:$IMAGE_VERSION
+EXEC_IMAGE=$BASE_EXEC_IMAGE:$IMAGE_VERSION
+
+exec_remote() {
+    docker run --rm \
```
rade: This probably should have a `-t` so ctrl-c works.

weave
```diff
@@ -27,27 +36,44 @@ usage() {
...
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v "$PROCFS":/hostproc \
+        -e VERSION="$VERSION" \
+        -e WEAVE_DOCKER_ARGS="$WEAVE_DOCKER_ARGS" \
+        -e WEAVE_PASSWORD="$WEAVE_PASSWORD" \
```
rade: Don't pass in the value. See #332.

rade:
- site/building.md needs updating - it references weavetools.
- I'm curious whether we still need `--without-ethtool`. Could we ditch it? Replace with a "run if exists"?

rade:
> - I'm curious whether we still need `--without-ethtool`. Could we ditch it? Replace with a "run if exists"?

Based on the rationale given in #357, I think the answer is yes, we can ditch it.

rade:
> Based on the rationale given in #357, I think the answer is yes, we can ditch it.

Hmm. There may be more to that. I wonder whether there's an assumption that `weave --local create-bridge` works w/o having ethtool installed. I can see how that might be useful. So perhaps leave as is.

rade:
docker versioning is an issue. the weavexec image will contain a docker binary based on whatever version happens to be current in alpine linux at the time. Docker daemons are backward compatible. But not forward compatible. So if the version in weaveexec is more recent than the docker daemon we are trying to talk to then things will break.

So we really want weavexec to include as old a version of docker as possible. At a bare minimum, the docker included in weavexec must support weave's usage of docker. But there is a tension here... weave generically wraps some docker commands. So, in particular, `weave run` wraps `docker run`. Since command line parsing is performed by the docker client, and there has been some change to `docker run` parameters over recent versions, we really want to include as new a version of docker as possible. So we want want both the oldest possible (for compatibility with the most docker daemons), an newest possible version (for compatibility with the most applications).

rade:
[version pinning in apk](http://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Holding_a_specific_package_back)

squaremo:
> So we want want both the oldest possible (for compatibility with the most docker daemons), an newest possible version (for compatibility with the most applications). Which is somewhat problematic.

It's OK if we special-case `weave run` to use the user's docker instead of the weaveexec one. This special casing is necessary in the proxy branch anyway (since run has to go through the proxy).

squaremo:
> It's OK if we special-case weave run to use the user's docker instead of the weaveexec one.

There is some stuff that has to happen on the host, though; specifically, attaching the weave IP, and finding the IP address to give the container to use for DNS. The former is just weave attach, run remotely if necessary. The latter would need a new command (say `./weave dns-config`).

rade:
> The latter would need a new command (say `weave dns-config`).

i.e. in order to return the DNS server IP, so it can be fed into the `docker run` invocation?

squaremo:
> i.e. in order to return the DNS server IP, so it can be fed into the docker run invocation?

Yus

rade:
It occurs to me that we could remove all the existence and versions checks for `ip` and `docker`. Technically they are still useful for invocation with `--local`, but that is advanced (and undocumented) usage of weave and thus can get away with less handholding.

rade:
I believe this also closes #45. @squaremo do you agree?

rade:
As a follow-up, we should look into [packaging a docker client](https://github.com/binocarlos/powerstrip-weave/issues/8#issuecomment-78093704) instead of installing the full-blown docker exe in the weavexec image.

squaremo:
> As a follow-up, we should look into packaging a docker client instead of installing the full-blown docker exe in the weavexec image.

I may have to do this anyway, since it looks like the earliest docker in `apk` is 1.4. How annoying.

rade:
Technically it's the Docker Remote API version that matters, not the docker version. There doesn't appear to be a published mapping of the latter to the former. Great.

>  special-case `weave run` to use the user's docker instead of the weaveexec one

That still needs doing, right?

> site/building.md needs updating - it references weavetools.

Ditto.

I think that's it; other possible improvements, such as removing the version checks, can be deferred.

squaremo:
* Now uses a docker 1.3.1 from get.docker.com
 * weavetools -> weaveexec in .gitignore and site/building.md
 * rebased to master

The way I've changed the script to use the "local" docker for `weave run` needs looking at.

test/setup.sh
```diff
@@ -13,11 +13,11 @@ echo Copying weave images and script to hosts
 for HOST in $HOSTS; do
     docker_on $HOST load -i ../weave.tar
     docker_on $HOST load -i ../weavedns.tar
-    docker_on $HOST load -i ../weavetools.tar
-    run_on $HOST mkdir -p `dirname $WEAVE`
-    cat ../weave         | run_on $HOST sh -c "cat > $WEAVE"
+    docker_on $HOST load -i ../weaveexec.tar
+    cp ../weave ./
```
rade: What is that line for? I can't figure it out.

weave
```diff
@@ -572,49 +613,42 @@ case "$COMMAND" in
...
-                        ;;
-                esac;
-            done
+        if [ "$LOCAL" = "--local" ]; then
+            dns_args "$@"
+        else
+            DNS_ARGS=$(exec_remote dns-args "$@")
+            [ "$1" = "--with-dns" ] && shift 1
         fi
+        validate_cidr $1
```
rade: Is there a reason you moved this? Or is it just a merge gone wrong? Everywhere else we first set CIDR and then invoke validate_cidr.

weave
```diff
@@ -27,27 +36,47 @@ usage() {
...
+        -e DOCKER_BRIDGE \
+        -e PROCFS=/hostproc --privileged --net=host \
+        $EXEC_IMAGE --local "$@"
+}
+
+if [ "$1" = "--local" ]; then
+    LOCAL="$1"
+    shift 1
+# Run is a special case because we want to use the 'local' docker to
+# run the container, rather than the docker in $EXEC_IMAGE.
```
rade: "use the 'local' docker to run the container" is rather ambiguous. How about "bypass the docker in $EXEC_IMAGE"?

Also, this really could do with a sentence or two explaining *why* that is necessary.

weaveexec/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM gliderlabs/alpine
+MAINTAINER Weaveworks Inc <help@weave.works>
+WORKDIR /home/weave
+RUN ["apk", "add", "--update", "ethtool", "conntrack-tools", "curl", "iptables", "iproute2"]
+RUN ["sh", "-c", "rm -rf /var/cache/apk/*"]
+ADD ./weave /home/weave/
+ADD ./docker.tgz /
```
rade: wow, I didn't know docker was uncompressing .tgz in an 'ADD'. How gross. And `/` is on the path for the weave script execution, presumably?

rade:
rest looks good.

test/setup.sh
squaremo: Oh. Probably because I wanted to invoke weave from the current directory; but `WEAVE=../weave` after all, so it's not needed.

weaveexec/Dockerfile
squaremo: The tarball has the structure /usr/local/bin/docker, so the executable ends up in /usr/local/bin/.

weave
squaremo: Nope, got moved while I was rearranging `run)`. I'll move it back.

weave
```diff
@@ -27,27 +36,47 @@ usage() {
...
+elif [ "$1" != "run" ]; then
+    exec_remote "$@"
+    exit $?
+fi
+
+[ $(id -u) = 0 ] || [ "$1" = "run" ] || {
```
squaremo: Actually this might be wrong: the idea was that if you're using `run`, the things you need root for are done by weaveexec; but, if you've said `--local`, it'll try to do those directly.

PR #390, opened 2015-02-10
IP Allocation Management

### LINKED ISSUE #22: automatic IP assigment
It is a chore to have to specify a CIDR whenever starting a container with weave. DHCP to the rescue. Or something like that. Arguably DHCP is the wrong way round. We want to go from a name/domain to an IP in a particular subnet.

Note that as well as for application containers, we also need IP addresses for the weave container and weavedns, though we could defer that to a separate issue.

bboreham:
Allow the user to run a container without having to specify an IP address; maintains unique addresses across a weave network while peer nodes join and leave.

This PR is for review; it will be re-done with fewer commits when we are finished reviewing and amending.

The implementation largely follows https://github.com/zettio/weave/wiki/IP-allocation-design; some additional notes are in the file ipam/ipam.md.

Closes #22

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Folders
 _obj
 _test
+.vagrant
 
 # Architecture specific extensions/prefixes
 *.[568vq]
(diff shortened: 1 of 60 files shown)


squaremo:
In the finest tradition of github, :+1:

bboreham:
Closing because we had a re-think about the design

rade:
superseded by #563

Issue #362, opened 2015-01-23
[dns] support TCP fall-back for very large DNS responses


bboreham:
Per [RFC1123 section 6.1.3.2](https://tools.ietf.org/html/rfc1123#page-75)

> DNS resolvers and recursive servers MUST support UDP, and SHOULD support TCP

This could be necessary if a user of weave mapped lots of IP addresses to the same DNS name, or, more likely for a DNS request outside of the weave network which goes via fall-back to the host's DNS server.

inercia:
Not sure if this is the best solution for large DNS responses. I think TCP support in DNS servers is only there for clients with limited connectivity, where UDP connection cannot get through a firewall. In fact, "bind" is configured by default for accepting just a few TCP connections (about 50). Should we look for an alternative solution?

rade:
> I think TCP support in DNS servers is only there for clients with limited connectivity, where UDP connection cannot get through a firewall.

I am far from an expert on the subject, but the RFC quoted above states that "If the Answer section of the response is truncated and if the requester supports TCP, it SHOULD try the query again using TCP." So the purpose of TCP connectivity is to deal with large responses. As per the subject of this issue. And, I believe @bboreham also confirmed this in experiments.

inercia:
You are right, it seems that TCP is also used for long responses. Then I could try to implement this...

bboreham:
Just to note a conversation with @rade where we felt, for fall-back, if the query came in to us over UDP we should always use UDP to the upstream DNS, and if it came in over TCP then always use TCP.  If a UDP response came back truncated we should pass it on truncated to the caller.  This way we will maintain the functionality required by the RFC without having to implement any complex logic.

bboreham:
Also note that it would be very hard to generate a weave.local response that is too big for UDP, until we implement #338 and/or #226

inercia:
@bboreham You are right when you say that it would be very difficult to generate such long responses until we solve those other issues. Anyway, it has served me for knowing more about the DNS code...

Regarding truncated responses coming from the upstream DNS server, I'm concerned about weaveDNS becoming a TCP proxy: if multiple clients ask about a _long name_ that must be resolved with the upstream server, those clients would fallback to our TCP port, and then we should establish TCP connections to the upstream server. I'm not sure whether this is a good idea, I think DNS servers should not do this kind of thing. At least we should limit the number of outgoing TCP connections: otherwise a malicious client could DoS our server quite easily...

bboreham:
[...]

PR #392, opened 2015-02-11
weave-362: support TCP fall-back for very large DNS responses

inercia:
Listen for DNS requests on a TCP port in addition to the regular UDP port. This allows clients to resend queries when long responses are truncated with UDP.


nameserver/server_test.go
```diff
@@ -56,6 +63,16 @@ func TestDNSServer(t *testing.T) {
...
 
+	// Retry the query with the TCP client
+	tr, _, err := tc.Exchange(m, dnsAddr)
+	wt.AssertNoErr(t, err)
+	wt.AssertStatus(t, tr.Rcode, dns.RcodeSuccess, "DNS response code (TCP)")
+	wt.AssertEqualInt(t, len(tr.Answer), 1, "Number of answers (TCP)")
+	wt.AssertType(t, tr.Answer[0], (*dns.A)(nil), "DNS record (TCP)")
+	wt.AssertEqualString(t, tr.Answer[0].(*dns.A).A.String(), testAddr1, "IP address (TCP)")
+	// TODO: look for some way of testing the TCP fallback for truncated responses: it seems there is
+	// TODO: no client lib in Go that can do that...
```
bboreham: I think you should test a large response, say 8K, then you can try it over UDP and expect truncation, then try it over TCP.

weavedns/main.go
```diff
@@ -18,7 +18,8 @@ func main() {
 		justVersion bool
 		ifaceName   string
 		apiPath     string
-		dnsPort     int
+		udpPort     int
+		tcpPort     int
```
bboreham: Why would the TCP port number be different from the UDP port number?

nameserver/server_test.go
inercia: Yes, I can do that... I'm afraid there is no client library in Go that makes that TCP fallback automatically for long DNS responses... but I can do that kind of "manual" test

nameserver/server.go
```diff
@@ -72,17 +146,16 @@ func rdnsHandler(config *dns.ClientConfig, lookups []Lookup) dns.HandlerFunc {
 	}
 }
 
-/* When we receive a request for a name outside of our '.weave.local.'
-   domain, ask the configured DNS server as a fallback.
-*/
-func notUsHandler(config *dns.ClientConfig) dns.HandlerFunc {
+// When we receive a request for a name outside of our '.weave.local.'
+// domain, ask the configured DNS server as a fallback.
+func (s *DNSServer) notUsHandler() dns.HandlerFunc {
```
bboreham: This seems like a gratuitous change; nothing in this function needs anything from DNSServer, so the code has become less general for no gain.

weavedns/main.go
inercia: I assumed they can be different, but if that will never be possible, I should change it... are you sure they will never be different?

nameserver/server.go
```diff
@@ -32,7 +20,93 @@ func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
...
+	if err := s.tcpSrv.Shutdown(); err != nil {
+		return err
+	}
+	if err := s.udpSrv.Shutdown(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *DNSServer) queryHandler(lookups []Lookup) dns.HandlerFunc {
```
bboreham: This seems like a gratuitous change; nothing in this function needs anything from DNSServer, so the code has become less general for no gain.

bboreham:
Without implementing the other side of the conversation - the fall-back query to an outside DNS server - over TCP, it seems like this feature would be very little-used as it is hard to generate a long DNS response inside the weave.local domain.

nameserver/server.go
inercia: Do you mean moving the queryHandler as a member of DNSServer?

bboreham:
Overall this is fairly well-done; I have made some comments but don't want you to think I'm all negative.

nameserver/server.go
inercia: I did it for coherence with the creation of the `DNSServer` class: once you have the class, all the initialization parameters should be provided in the `New*()` methods, and that means that you store the config as an attribute in the instance... then it didn't make sense to pass it around to handlers anymore...

weavedns/main.go
bboreham: According to the RFCs it is always 53.  We allow it to be changed for testing purposes, but one number should suffice for that.  I can't think of any other reason to allow two different numbers.

bboreham:
Should note that this is intended to address #362

weavedns/main.go
inercia: Good, I'll fix that

nameserver/server.go
```diff
@@ -32,7 +20,94 @@ func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
...
+
+	mdnsServer, err := NewMDNSServer(s.zone)
+	checkFatal(err)
+
+	err = mdnsServer.Start(s.iface)
+	checkFatal(err)
+
+	address := fmt.Sprintf(":%d", s.port)
+	s.udpSrv = &dns.Server{Addr: address, Net: "udp", Handler: mux(&dns.Client{Net: "udp", UDPSize: UDPBufSize})}
+	s.tcpSrv = &dns.Server{Addr: address, Net: "tcp", Handler: mux(&dns.Client{Net: "tcp"})}
```
inercia: All of this initialization stuff (ie, `mux()`, `s.udpSrv`, `s.tcpSrv`) has been moved to `NewDNSServer()` in zettio/weave#394. The `Start()` method should just do that: start the server.

nameserver/server_test.go
```diff
@@ -112,6 +116,76 @@ func TestDNSServer(t *testing.T) {
...
+	}
+
+	Debug.Println("Using TCP for the same query")
+	tr, _, err := tc.Exchange(m, dnsAddr)
+	wt.AssertNoErr(t, err)
+	wt.AssertStatus(t, tr.Rcode, dns.RcodeSuccess, "DNS response code (TCP)")
+	if len(r.Answer) > 1 { // TODO: remove this when we have proper support for multiple responses (zettio/weave#338)
+		wt.AssertEqualInt(t, len(tr.Answer), ansCount, "Number of answers (TCP)")
+	}
+	wt.AssertType(t, tr.Answer[0], (*dns.A)(nil), "DNS record (TCP)") // assume they are all the same...
```
inercia: This is a bit of dirty code: we cannot test the TCP-fallback properly until we can send multiple answers for a zone... so this is some sort of transition code...

nameserver/server.go
```diff
@@ -72,17 +147,14 @@ func rdnsHandler(config *dns.ClientConfig, lookups []Lookup) dns.HandlerFunc {
...
+// When we receive a request for a name outside of our '.weave.local.'
+// domain, ask the configured DNS server as a fallback.
+func (s *DNSServer) notUsHandler(dnsClient *dns.Client) dns.HandlerFunc {
 	return func(w dns.ResponseWriter, r *dns.Msg) {
 		q := r.Question[0]
 		Debug.Printf("[dns msgid %d] Fallback query: %+v", r.MsgHdr.Id, q)
-		for _, server := range config.Servers {
-			reply, _, err := dnsClient.Exchange(r, fmt.Sprintf("%s:%s", server, config.Port))
+		for _, server := range s.config.Servers {
+			reply, _, err := dnsClient.Exchange(r, fmt.Sprintf("%s:%s", server, s.config.Port))
```
inercia: Each _notUs_ DNS query results in an outbound DNS request, but we should have some sort of caching-resolver, an actor that is queried about names, mnages a cache, etc... so this simple solution that will be improved when zettio/weave#225 is fixed.

nameserver/server_test.go
```diff
@@ -27,7 +27,7 @@ func TestDNSServer(t *testing.T) {
 	dnsAddr := fmt.Sprintf("localhost:%d", port)
 	testCIDR1 := testAddr1 + "/24"
 
-	common.InitDefaultLogging(true)
+	InitDefaultLogging(true)
```
bboreham: What's this change doing here?

nameserver/server_test.go
bboreham: Why not use a fallback query as your test case, then you can return whatever you like.

nameserver/server_test.go
inercia: I have removed the `common` prefix, so I can use `Debug.Printf(...)` instead of the verbose `common.Debug.Printf(...)`

nameserver/util.go
```diff
@@ -0,0 +1,17 @@
...
+	if e != nil {
+		Error.Fatal(e)
+	}
+}
+
+func checkWarn(e error) {
+	if e != nil {
+		Warning.Println(e)
+	}
+}
```
rade: Surely all the above should go into `common`.

nameserver/server_test.go
```diff
@@ -41,7 +41,10 @@ func TestDNSServer(t *testing.T) {
 	wt.AssertNoErr(t, err)
 
 	config := &dns.ClientConfig{Servers: []string{"127.0.0.1"}, Port: fallbackPort}
-	go startServerWithConfig(config, zone, nil, port, 0)
+	srv, err := NewDNSServerWithConfig(config, zone, nil, port)
+	wt.AssertNoErr(t, err)
+	defer srv.Stop()
+	go srv.Start()
```
rade: I don't understand why this is a goroutine. `Start` does itself spawn goroutines, and, afaict, does terminate.

nameserver/server_test.go
inercia: Hmmm, not sure about that... the point is to check that, after a truncated response with UDP, you can receive the full response with TCP, so testing that you can receive whatever you send with TCP does not make much sense (that would only test that DNS with TCP is reliable, something we already know/assume)

nameserver/server_test.go
```diff
@@ -112,6 +116,76 @@ func TestDNSServer(t *testing.T) {
...
+	defer s.Shutdown()
+
+	_, fallbackPort, err := net.SplitHostPort(fallbackAddr)
+	wt.AssertNoErr(t, err)
+
+	config := &dns.ClientConfig{Servers: []string{"127.0.0.1"}, Port: fallbackPort}
+	srv, err := NewDNSServerWithConfig(config, zone, nil, port)
+	wt.AssertNoErr(t, err)
+	defer srv.Stop()
+	go srv.Start()
```
rade: I don't understand why this is a goroutine. `Start` does itself spawn goroutines, and, afaict, does terminate.

nameserver/server_test.go
inercia: You are right. I'll fix this

nameserver/server.go
```diff
@@ -32,7 +20,94 @@ func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
...
+	return &DNSServer{
+		config: config,
+		zone:   zone,
+		iface:  iface,
+		port:   port,
+	}, nil
+}
+
+// Creates a new DNS server
+func NewDNSServer(zone Zone, iface *net.Interface, port int) (*DNSServer, error) {
```
rade: I don't think we need this function. It does very little and has just one call site. (NB: when removing the function, `NewDNSServerWithConfig` should obviously be renamed to `NewDNSServer`)

nameserver/server.go
```diff
@@ -32,7 +20,94 @@ func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
...
+	config *dns.ClientConfig
+	zone   Zone
+	iface  *net.Interface
+	port   int
+	udpSrv *dns.Server
+	tcpSrv *dns.Server
+}
+
+// Creates a new DNS server with a given config
+func NewDNSServerWithConfig(config *dns.ClientConfig, zone Zone, iface *net.Interface, port int) (*DNSServer, error) {
```
rade: I don't see the point of the error return. If at some future point we really want to return an error, it's easy enough to change the API.

nameserver/server.go
```diff
@@ -32,7 +20,94 @@ func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
...
+func NewDNSServer(zone Zone, iface *net.Interface, port int) (*DNSServer, error) {
+	config, err := dns.ClientConfigFromFile("/etc/resolv.conf")
+	if err != nil {
+		return nil, err
+	}
+	return NewDNSServerWithConfig(config, zone, iface, port)
+}
+
+// Start the DNS server
+func (s *DNSServer) Start() error {
```
rade: This function never actually appears to return an error. Instead every error is fatal. Which is horrific. (yes, I know some of this is a remnant of the previous code).

weavedns/main.go
```diff
@@ -64,7 +64,12 @@ func main() {
 	}
 
 	go weavedns.ListenHttp(weavedns.LOCAL_DOMAIN, zone, httpPort)
-	err := weavedns.StartServer(zone, iface, dnsPort, wait)
+	srv, err := weavedns.NewDNSServer(zone, iface, dnsPort)
+	if err != nil {
+		Error.Fatal("Failed to start server", err)
+	}
+
+	srv.Start()
```
rade: presumably you meant to write `err := srv.Start()`, given the code that follows.

nameserver/server.go
inercia: There are two usage cases in this PR: we create a concrete `dns.ClientConfig` and the invoke `NewDNSServerWithConfig` in unit tests, but the common case is to use `NewDNSServer`. Passing a _low-level_ parameter like `dns.ClientConfig` by default is not very elegant/useful, so that is the reason why `DNSServer` is properly initialized with a [DNSServerConfig](https://github.com/zettio/weave/pull/394/files#diff-7d7cca87c8b1140efab81f1f58b9c0edR25) in the [next PR](https://github.com/zettio/weave/pull/394). I can change it here anyway...

nameserver/server.go
inercia: Ok, I'll fix it.

weavedns/main.go
inercia: True

nameserver/server.go
inercia: Changed

inercia:
This PR has been tested in the following way:

1. run a BIND server with `docker run --name='bind' -d -p 2053:53/udp -p 2053:53/tcp -v /opt/bind:/data ctrld/bind:latest`
2. enter the container with `sudo docker-enter bind`
3. add, at the end of `/data/bind/etc/named.conf.default-zones`, a new zone definition:
`zone "example.com" { type master; file "/etc/bind/db.example.com"; };`
4. define the zone in a new file `/etc/bind/db.example.com` (notice some IP addresses have been skipped here):
```
$ORIGIN example.com.     ; designates the start of this zone file in the namespace
$TTL 1h                  ; default expiration time of all resource records without their own TTL value
example.com.  IN  SOA   ns.example.com. username.example.com. ( 2007120710 1d 2h 4w 1h )
example.com.  IN  NS    ns                    ; ns.example.com is a nameserver for example.com
example.com.  IN  NS    ns.somewhere.example. ; ns.somewhere.example is a backup nameserver for example.com
example.com.  IN  MX    10 mail.example.com.  ; mail.example.com is the mailserver for example.com
@             IN  MX    20 mail2.example.com. ; equivalent to above line, "@" represents zone origin
@             IN  MX    50 mail3              ; equivalent to above line, but using a relative host name
example.com.  IN  A     192.0.2.1             ; IPv4 address for example.com

              IN  A     192.0.9.1
              IN  A     192.0.9.2
              IN  A     192.0.9.3
              IN  A     192.0.9.4
              IN  A     192.0.9.5
              IN  A     192.0.9.6
              IN  A     192.0.9.7
              IN  A     192.0.9.8
              IN  A     192.0.9.9
              IN  A     192.0.9.10
              IN  A     192.0.9.11
              IN  A     192.0.9.12
              IN  A     192.0.9.13
              IN  A     192.0.9.14
              IN  A     192.0.9.15
              IN  A     192.0.9.16
              IN  A     192.0.9.17
              IN  A     192.0.9.18
              IN  A     192.0.9.19
              ...
              IN  A     192.0.10.243
              IN  A     192.0.10.244
              IN  A     192.0.10.245
              IN  A     192.0.10.246
              IN  A     192.0.10.247
              IN  A     192.0.10.248
              IN  A     192.0.10.249
              IN  A     192.0.10.250
              IN  A     192.0.10.251
              IN  A     192.0.10.252
              IN  A     192.0.10.253
              IN  A     192.0.10.254
              IN  A     192.0.10.255
              IN  AAAA  2001:db8:10::1        ; IPv6 address for example.com
ns            IN  A     192.0.2.2             ; IPv4 address for ns.example.com
              IN  AAAA  2001:db8:10::2        ; IPv6 address for ns.example.com
www           IN  CNAME example.com.          ; www.example.com is an alias for example.com
wwwtest       IN  CNAME www                   ; wwwtest.example.com is another alias for www.example.com
mail          IN  A     192.0.2.3             ; IPv4 address for mail.example.com
mail2         IN  A     192.0.2.4             ; IPv4 address for mail2.example.com
mail3         IN  A     192.0.2.5             ; IPv4 address for mail3.example.com
```
5. force the reload of data in BIND with `killall -HUP named`
6. without leaving the container, check you can get a valid response for `example.com` with `dig @127.0.0.1 example.com`. You will see something like:
```
# dig @127.0.0.1  example.com
;; Truncated, retrying in TCP mode.

; <<>> DiG 9.9.5-3ubuntu0.1-Ubuntu <<>> @127.0.0.1 example.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 5292
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 511, AUTHORITY: 2, ADDITIONAL: 3

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;example.com.			IN	A

;; ANSWER SECTION:
example.com.		3600	IN	A	192.0.10.175
example.com.		3600	IN	A	192.0.10.63
example.com.		3600	IN	A	192.0.10.56
example.com.		3600	IN	A	192.0.10.237
example.com.		3600	IN	A	192.0.10.26
example.com.		3600	IN	A	192.0.9.144
example.com.		3600	IN	A	192.0.10.104
example.com.		3600	IN	A	192.0.10.17
example.com.		3600	IN	A	192.0.9.3
...
```
and notice the `Truncated, retrying in TCP mode. ` line.
6. launch the WeaveDNS server from a container or directly. Set the fallback DNS server to the BIND container IP address. It can be done with `sudo $GOPATH/src/github.com/zettio/weave/weavedns/weavedns -fallback 127.0.0.1:2053 -debug -dnsport 1053`
7. query the WeaveDNS server with
```
 $ dig @127.0.0.1   -p 1053  example.com
;; Truncated, retrying in TCP mode.

; <<>> DiG 9.9.5-4.3ubuntu0.2-Ubuntu <<>> @127.0.0.1 -p 1053 example.com
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 59632
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 511, AUTHORITY: 2, ADDITIONAL: 3

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;example.com.			IN	A

;; ANSWER SECTION:
example.com.		3600	IN	A	192.0.9.33
example.com.		3600	IN	A	192.0.10.86
example.com.		3600	IN	A	192.0.9.7
example.com.		3600	IN	A	192.0.10.128
example.com.		3600	IN	A	192.0.10.52
example.com.		3600	IN	A	192.0.9.164
...
```
You should see the `Truncated, retrying in TCP mode. ` line again, and the WeaveDNS server logs should show tow consecutive queries:
```
...
DEBUG: 2015/02/23 11:39:57.051183 [dns msgid 10663] Fallback query: {Name:example.com. Qtype:1 Qclass:1}
DEBUG: 2015/02/23 11:39:57.052631 [dns msgid 10663] Given answer by 172.17.0.3 for query example.com.
DEBUG: 2015/02/23 11:39:57.053288 [dns msgid 59632] Fallback query: {Name:example.com. Qtype:1 Qclass:1}
DEBUG: 2015/02/23 11:39:57.057351 [dns msgid 59632] Given answer by 172.17.0.3 for query example.com.
...
```

inercia:
Minor comment: for some unknown reason, some commits from master have been incorporated in this branch after rebasing master... sorry about that.

rade:
> for some unknown reason, some commits from master have been incorporated in this branch after rebasing master... sorry about that.

Please figure out what went wrong there and fix it since otherwise this is pretty much impossible to review.

inercia:
@rade apologies for these extra commits, but they are just minor, small commits (ie, a copyright update, a flag for coverage...) and I think they will not affect the review in a significant way...

rade:
Perhaps, but we need to have confidence in our vcs and understand *exactly* what it's doing.

rade:
copying our resident git(hub) experts: @dpw @squaremo

inercia:
@rade it was my fault by using "Rebase my github fork" in IntelliJ...

dpw:
Well, it's hard to speculate how the branch got into that state.  The redundant commits in the branch are not considered identical to the originals due to the different 'Committer' field.  But it should be easy to fix: do `git rebase -i master` and remove the redundant commits.

BTW, I suggest removing the links to the issue in the commit titles.  Even if there is a good argument for it, it's not what the rest of the project does currently, so the resulting commit logs will be conspicuously inconsistent.

inercia:
Done. As @dpw said, a simple `git rebase -i master` did it.

Regarding the links to the issue, I think it is a good idea in general as it increases traceability. But I agree that it does not make much sense if I'm the only one doing it... what what do you guys think?

rade:
> Regarding the links to the issue

Let's not have a workflow discussion in this issue. Start an email thread.

nameserver/server.go
squaremo: Yes I would think so, there's no reason for it to depend on the server.

squaremo: Oh it is by analogy with the other *Handler functions. I think there is a better arrangement in there somewhere, but this is OK for now.

squaremo:
LGTM, nice one @inercia

inercia:
Thanks @squaremo

PR #394, opened 2015-02-13
weave-366: add the ability to specify domain name, other than .weave.local

inercia:
This PR improved the DNS server configuration mechanism. The new configuration system includes the ability to specify a different domain name other than ".weave.local".

Note: depends on zettio/weave#392


squaremo:
@inercia Might be worth rebasing this now that the other is both rebased and merged.

inercia:
@squaremo I've just rebased master.

nameserver/server.go
```diff
@@ -21,59 +23,95 @@ func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
...
-	Info.Printf("WeaveDNS server starting...")
+// Creates a new DNS server
+func NewDNSServer(config DNSServerConfig) (s *DNSServer, err error) {
+	// fill empty parameters with defaults...
+	if config.Port == 0 {
+		config.Port = DEFAULT_SERVER_PORT
+	}
+	if len(config.LocalDomain) == 0 {
+		config.LocalDomain = DEFAULT_LOCAL_DOMAIN
+	}
```
inercia: I assume there are no restrictions in the local domain name. Should we check that the `LocalDomain` follows some rules (ie, it has `"local."` at the end)?

squaremo: I think I would prefer it to be validated -- do we want users to be able to do daft things like name their domain "google.com"?

squaremo: Having said that, @rade points out that this is all about letting users have vanity domains, so we may as well let them have whatever they want.

squaremo:
Sorry to be picky @inercia -- can you separate out PRs for the basic change (allow local domain to be passed as an argument) from the wider changes (put lots of things in a config struct). I am not so sure about the latter, and I don't think the former depends on it.

nameserver/server.go
inercia: Ok, let's users dig their own grave...

inercia:
@squaremo don't you think that will make `NewDNSServer` a tedious function? IMO, users should be able to get a server with default values with something like `srv := NewDNSServer(zone, iface)`, but something like `NewDNSServer(config *dns.ClientConfig, localDomain string, zone Zone, iface *net.Interface, port int)` seems too long/verbose to me, specially when most of those parameters are optional. I thought this was the preferred way to deal with this kind of problem in Go, to use a `*Config` struct...

Anyway, I can move that code to another PR, but the number of optional parameters will grow even more [in the PR](https://github.com/inercia/weave/compare/weave-225) that addresses issue zettio/weave#225...

squaremo:
@inercia That's a fair point about constructor arguments; though I'd say the more common pattern is to have a struct literal rather than a constructor. Most of the work in the constructor is defaulting arguments. Can the rest be put in Start?

inercia:
@squaremo IMO a constructor should initialize the object and, for a struct like this, initialize all the attributes that are part of the object, so it should do things like setting default values, checking values provided by the caller (and return in case of errors) and, in this particular example, creating things like muxes, handlers, etc... Users could also initialize the struct by themselves, but I think it make things worse for complex structs: better to use constructors. IMO, constructors _should not_ start anything and _should not_ have any side effects, in the same way as any primitive type constructor does (ie, `a:=1` does not have side effects), so returning after `NewDNSServer()` should have the same side effects as returning after `a:=1`...

Then things should start _moving_ when we invoke `Start()` (I recognize that `ListenAndServe()` would be a better name, but I did it for coherence with the mDNS server code). We should then do things like starting listening at sockets, spawing goroutines, etc...

Maybe the `dns.Server` initializations could be considered part of the _action_ and not the _initialization_, but they basically do a couple of `m := NewServeMux(...); m.HandleFunc()`, and those should be just initializations...

So what do you think? Have I convinced you or do you feel uncomfortable with that code there? ;)

squaremo:
The thing that makes me uncomfortable is that the config struct is used to have optional arguments (ok), but also to hold some mutable state (eeuuuwww). So if you stuck to using it for optional arguments, I would be happy. Make sense?

inercia:
@squaremo you are right: the config struct should be just for immutable arguments. I'll fix that

squaremo:
Quick Q: Is the config put in a field in the server struct for (hypothetical) later inspection?

inercia:
@squaremo it doesn't make sense to keep it anymore... I'll remove it

weavedns/main.go
```diff
@@ -31,10 +32,11 @@ func main() {
 	flag.BoolVar(&justVersion, "version", false, "print version and exit")
 	flag.StringVar(&ifaceName, "iface", "", "name of interface to use for multicast")
 	flag.StringVar(&apiPath, "api", "unix:///var/run/docker.sock", "Path to Docker API socket")
+	flag.StringVar(&localDomain, "localDomain", "", "local domain (ie, 'weave.local.')")
```
bboreham: Suggest you default this to ".weave.local"

Issue #307, opened 2014-12-31
weave script is becoming unmaintainable and limiting


rade:
The weave script is doing quite a lot of work and keeps growing. It's getting to the point where it becomes quite hard to maintain and extend. Also, `sh` is a rather limited environment for complex programming, and access to any kernel features is always mediated by tools, which introduces dependencies, versioning issues etc.

On a positive note, the script has served us well as being the single thing that needs to be downloaded and installed - no need for packaging an OS specifics. Also, it has proved very hackable - we've received quite a few PRs for the script; far more than the go code.

Replacing the script with a statically linked exe would retain the "single download, no fuss install" of the current script.

errordeveloper:
I think that we can always bring back hackability by providing simple exec overrides to each of the main steps or even a some kind of interpreter (e.g. Lua or MicroPython). This is probably not of importance in the short term, but I just thought I'd make sure to mention what I've been thinking about this subject. I believe that moving from shell script to binary tool will eventually arise requests for configuration files, and I think we should discus hooks/scripting interfaces then. I think it would be much easier to provide a hook to override an entire step rather then provide configuration parameters for every single call within that step. This of course depends on what exactly we get asked for.

errordeveloper:
I have [forked][1] the script once for Kubernetes, it has a few things that are pretty much same in upstream version and I can go through it in more detail if we want to, but it might be an extreme case of modification.

[1]: https://github.com/errordeveloper/weave-demos/blob/master/poseidon/weave

dpw:
Bear in mind that the role of the weave script (or whatever supersedes it) could change significantly when the new docker networking stuff happens.  That might still be some time away, but knowing that it will happen, perhaps we should try to live with the current weave script, rather than making big plans to replace it or enhance it in ways that might not make sense after that transition.

If we can identify specific maintainability issues in the weave script, maybe there is something we can do to mitigate them?

paulbellamy:
If we do end up moving to a binary CLI, the [heroku toolbelt](https://toolbelt.heroku.com) and [cli](https://github.com/heroku/toolbelt) would probably be a good model.

squaremo:
While writing the plugin, I started moving implementations of common weave operations into a separate module. So far I've been able to replicate the important bits without needing the script at all, which is nice.

As an extrapolation, we could

 * move all the individual ("assign an IP") and composite ("give this container an assigned IP") operations into a library
 * port the proxy to use the library where necessary
 * make an executable with the same commands as the script
[...]

Issue #369, opened 2015-01-27
Be able to name weave peers


bboreham:
When studying logs and `weave status` output, it can get tiresome to keep mapping from the identity you know ("this is the node I run on EC2") to the hex peer-name used internally by Weave.

I propose that weave fetches the hostname of its container and displays it alongside the hex peer-name.  It would also need to include this information in gossip messages so it is included in `weave status` on other peers.

The user would set the hostname by adding something like `--hostname=ec2_node` on the end of the `weave launch` command.

bboreham:
Turns out this is not as simple as I thought, because `--hostname` will be passed to Weave as an argument, not to Docker.

PR #407, opened 2015-02-19
369 name weave peers

awh:
Implement `-hostname` flag to weave router for inclusion in `weave status`:

    root@weave01:~# weave status
    weave router git-170724472555
    Encryption off
    Our name is 7a:85:0d:3f:4b:8b (weave01)
    Sniffing traffic on &{25 65535 ethwe 26:47:3f:b1:9f:35 up|broadcast|multicast}
    MACs:
    e6:eb:3d:c2:f2:c0 -> 7a:85:0d:3f:4b:8b (2015-02-19 14:36:43.698599704 +0000 UTC)
    ea:9b:f6:9a:a8:94 -> 7a:ee:54:05:6d:19 (2015-02-19 14:36:45.467878022 +0000 UTC)
    02:c5:dc:fa:4f:19 -> 7a:c4:7d:e3:0f:9b (2015-02-19 14:36:50.594692416 +0000 UTC)
    Peers:
    Peer 7a:c4:7d:e3:0f:9b (weave03) (v37) (UID 12377675045837118970)
       -> 7a:ee:54:05:6d:19 (weave02) [178.62.77.81:6783]
       -> 7a:85:0d:3f:4b:8b (weave01) [178.62.57.35:6783]
    Peer 7a:85:0d:3f:4b:8b (weave01) (v8) (UID 8327482320931219319)
       -> 7a:c4:7d:e3:0f:9b (weave03) [178.62.106.102:59473]
       -> 7a:ee:54:05:6d:19 (weave02) [178.62.77.81:48078]
    Peer 7a:ee:54:05:6d:19 (weave02) (v4) (UID 13020696009711836879)
       -> 7a:85:0d:3f:4b:8b (weave01) [178.62.57.35:6783]
       -> 7a:c4:7d:e3:0f:9b (weave03) [178.62.106.102:53570]

Notes:

* `weave launch` does not accept user supplied `-hostname` as this would require introducing `getopt` or similar into the bash script and given [#307](https://github.com/zettio/weave/issues/307) is in the offing it didn't seem prudent; script supplies kernel hostname directly to weave router for now. Please advise if you would like this addressed before merging
* Protocol version was increased due to binary incompatibility introduced by inclusion of hostname in the gossip protocol messages

### DIFF
--- a/router/gossip_test.go
+++ b/router/gossip_test.go
@@ -19,7 +19,7 @@ type mockChannelConnection struct {
 // We need to create some dummy channels otherwise tests hang on nil
 // channels when Router.OnGossip() calls async methods.
 func NewTestRouter(name PeerName) *Router {
-	router := NewRouter(nil, name, nil, 10, 1024, nil)
+	router := NewRouter(nil, name, "hostname", nil, 10, 1024, nil)
 	router.ConnectionMaker.queryChan = make(chan *ConnectionMakerInteraction, ChannelSize)
 	router.Routes.queryChan = make(chan *Interaction, ChannelSize)
 	return router
@@ -35,8 +35,8 @@ func (router *Router) AddTestChannelConnection(r *Router) {
 	fromName := router.Ourself.Peer.Name
 	toName := r.Ourself.Peer.Name
 
-	fromPeer := NewPeer(fromName, router.Ourself.Peer.UID, 0)
-	toPeer := NewPeer(toName, r.Ourself.Peer.UID, 0)
+	fromPeer := NewPeer(fromName, "hostname", router.Ourself.Peer.UID, 0)
+	toPeer := NewPeer(toName, "hostname", r.Ourself.Peer.UID, 0)
 
 	r.Peers.FetchWithDefault(fromPeer)    // Has side-effect of incrementing refcount
 	router.Peers.FetchWithDefault(toPeer) //
@@ -70,10 +70,10 @@ func TestGossipTopology(t *testing.T) {
 // the routers supplied as arguments, carrying across all UID and
 // version information.
 func (router *Router) tp(routers ...*Router) *Peer {
-	peer := NewPeer(router.Ourself.Peer.Name, router.Ourself.Peer.UID, 0)
+	peer := NewPeer(router.Ourself.Peer.Name, "hostname", router.Ourself.Peer.UID, 0)
 	connections := make(map[PeerName]Connection)
 	for _, r := range routers {
-		p := NewPeer(r.Ourself.Peer.Name, r.Ourself.Peer.UID, r.Ourself.Peer.version)
+		p := NewPeer(r.Ourself.Peer.Name, "hostname", r.Ourself.Peer.UID, r.Ourself.Peer.version)
 		connections[r.Ourself.Peer.Name] = newMockConnection(peer, p)
 	}
 	peer.SetVersionAndConnections(router.Ourself.Peer.version, connections)
--- a/router/gossip_test.go
+++ b/router/gossip_test.go
@@ -19,7 +19,7 @@ type mockChannelConnection struct {
 // We need to create some dummy channels otherwise tests hang on nil
 // channels when Router.OnGossip() calls async methods.
 func NewTestRouter(name PeerName) *Router {
-	router := NewRouter(nil, name, "hostname", nil, 10, 1024, nil)
+	router := NewRouter(nil, name, "", nil, 10, 1024, nil)
 	router.ConnectionMaker.queryChan = make(chan *ConnectionMakerInteraction, ChannelSize)
 	router.Routes.queryChan = make(chan *Interaction, ChannelSize)
 	return router
@@ -35,8 +35,8 @@ func (router *Router) AddTestChannelConnection(r *Router) {
 	fromName := router.Ourself.Peer.Name
 	toName := r.Ourself.Peer.Name
 
-	fromPeer := NewPeer(fromName, "hostname", router.Ourself.Peer.UID, 0)
-	toPeer := NewPeer(toName, "hostname", r.Ourself.Peer.UID, 0)
+	fromPeer := NewPeer(fromName, "", router.Ourself.Peer.UID, 0)
+	toPeer := NewPeer(toName, "", r.Ourself.Peer.UID, 0)
 
 	r.Peers.FetchWithDefault(fromPeer)    // Has side-effect of incrementing refcount
 	router.Peers.FetchWithDefault(toPeer) //
@@ -70,10 +70,10 @@ func TestGossipTopology(t *testing.T) {
 // the routers supplied as arguments, carrying across all UID and
 // version information.
 func (router *Router) tp(routers ...*Router) *Peer {
-	peer := NewPeer(router.Ourself.Peer.Name, "hostname", router.Ourself.Peer.UID, 0)
+	peer := NewPeer(router.Ourself.Peer.Name, "", router.Ourself.Peer.UID, 0)
 	connections := make(map[PeerName]Connection)
 	for _, r := range routers {
-		p := NewPeer(r.Ourself.Peer.Name, "hostname", r.Ourself.Peer.UID, r.Ourself.Peer.version)
+		p := NewPeer(r.Ourself.Peer.Name, "", r.Ourself.Peer.UID, r.Ourself.Peer.version)
 		connections[r.Ourself.Peer.Name] = newMockConnection(peer, p)
 	}
 	peer.SetVersionAndConnections(router.Ourself.Peer.version, connections)
--- a/site/troubleshooting.md
+++ b/site/troubleshooting.md
@@ -86,11 +86,10 @@ The terms used here are explained further at
 [how it works](how-it-works.html).
 
 The 'Our name' line identifies the local weave router as a peer in the
-weave network. It displays both the peer name and the peer's nickname
-in parenthesis; the peer name defaults to a MAC address, whilst the
-nickname defaults to the name of the host on which the weave container
-was launched. If desired the nickname can be overriden by supplying
-the `-nickname` argument to `weave launch`.
+weave network, displaying the peer name followed by the peer's nickname
+in parenthesis. The nickname defaults to the name of the host on which
+the weave container was launched; if desired it can be overriden by
+supplying the `-nickname` argument to `weave launch`.
 
 The 'Sniffing traffic' line shows details of the virtual ethernet
 interface that weave is using to receive packets on the local
--- a/site/troubleshooting.md
+++ b/site/troubleshooting.md
@@ -58,7 +58,7 @@ This produces output like:
 ````
 weave router 0.7.0
 Encryption off
-Our name is 7a:f4:56:87:76:3b
+Our name is 7a:f4:56:87:76:3b (weave01)
 Sniffing traffic on &{39 65535 ethwe ae:e3:07:9c:8c:d4 up|broadcast|multicast}
 MACs:
 ba:8c:b9:dc:e1:c9 -> 7a:f4:56:87:76:3b (2014-10-23 16:39:19.482338935 +0000 UTC)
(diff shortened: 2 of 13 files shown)


rade:
Just for reference, this is meant to address #369.

rade:
Referring to this new piece of data as 'hostname' is wrong; it could be anything, and in many cases will *not* be the hostname. Yes, I know #369 calls it 'hostname', but I didn't file that issue ;)

And yes, we really, really must be able to pass that thing into `weave launch`.

awh:
Understood. Do you have a preference on what you'd like it to be called?

rade:
> Do you have a preference on what you'd like it to be called?

I am sure I do, but I don't know what it is :)

rade:
'displayname'? Not great, so happy to consider alternatives.

weaver/main.go
```diff
@@ -87,6 +89,18 @@ func main() {
 		routerName = iface.HardwareAddr.String()
 	}
 
+	if hostName == "" {
+		osHostName, err := os.Hostname()
+		if err != nil {
+			log.Fatal(err)
+		}
+		if osHostName == "" {
+			hostName = "<unknown>"
```
rade: I reckon that will look rather odd in the various places where that gets displayed. Set it to "", perhaps?

awh:
I was just about to suggest that or friendlyname, informalname....

weaver/main.go
```diff
@@ -121,8 +135,8 @@ func main() {
 		defer profile.Start(&p).Stop()
 	}
 
-	router := weave.NewRouter(iface, ourName, []byte(password), connLimit, bufSz*1024*1024, logFrame)
-	log.Println("Our name is", router.Ourself.Name)
+	router := weave.NewRouter(iface, ourName, hostName, []byte(password), connLimit, bufSz*1024*1024, logFrame)
+	log.Println("Our name is", router.Ourself.Name, "(" + hostName + ")")
```
rade: IMO it would be "more correct" and consistent to get extract the info from router.Ourself.

rade:
'friendlyname' is quite nice. let's invite @bboreham to help paint this particular bike shed :)

weaver/main.go
awh: Actually this logging is duplicated in router.Status() - how about log.Println(router.Status()) instead?

dpw:
If I may join the bike-shedding, what about "nickname"?

weaver/main.go
rade: That's by coincidence. And it also prints a lot more, which isn't particularly meaningful at this stage.

rade:
I do like 'nickname'.

weaver/main.go
awh: Hmm that won't look great - router.Status() returns a single string with embedded newlines...

awh:
Addressed comments to date!

rade:
The new option needs to added to the docs. Somewhere.

awh:
Just thinking about the default value of the nickname - would `$(hostname -f)` be better than `$(hostname)`?

rade:
Ouch, `hostname` isn't even in POSIX! I guess we don't care, but still surprising.

awh:
That is surprising. Bash has a `HOSTNAME` environment variable, however the script depends on `/bin/sh`. Unfortunately we can't solely depend on `os.Hostname()` from Go because it obtains the hostname of the container and not of the underlying host kernel :(

With respect to the other issue (short hostname vs FQDN) I suspect the latter is going to be useful if folks have naming schemes like `app01.dc01.example.com`, `app01.dc02.example.com`...

rade:
Can anybody think of a *downside* to using FQDNs? @bboreham

bboreham:
When I thought of this originally, I meant the "hostname" of the container.  So we didn't have to implement any flag-parsing, because `--hostname` is already a Docker option.  I'm not clear why you went down this other route.

rade:
The hostname of the weave container is not exactly meaningful.

bboreham:
It defaults to the container ID, which I usually know when testing, and I can set to anything else I feel like.  So it is meaningful to me.

rade:
setting it to anything else is not exactly straightforward

bboreham:
I seem to be missing this part.  Extra args in `weave launch` are handed to Docker; I can add `--hostname` to those args.  Where does it get difficult?

rade:
extra args to `weave launch` are handed to weave.

bboreham:
Aha.  I think that belongs as a comment to #369.

rade:
you'd have to use `WEAVE_DOCKER_ARGS`.

awh:
The only downside I can think of to FQDNs is that it might cause the output of `weave status` to wrap if you have a long domain...

bboreham:
I just took a look at `hostname -f` on one of the machines I use for testing, and it is "bb-ub14-test2.bb-ub14-test2.f3.internal.cloudapp.net".  I think I would hate that more than the hex numbers I have to work with today.

rade:
and the alternative is what?

bboreham:
My idea was for the user (me) to specify the name with a command-line flag.

rade:
Which you can do. The debate re fqdn vs unqualified concerns what should happen when you *don't* specify a name.

bboreham:
From my sampling of two Cloud VMs, `hostname` seems to be less worse than `hostname -f`.  The other example is "ip-172-31-20-70.eu-west-1.compute.internal"

Given the option to supply on the command line, anyone that cares to can use `hostname -f` to supply that, right?

rade:
I don't understand why you feel it's better to default to the short names instead of the the FQDNs. The latter at least give you a clue what cloud the host is in.

bboreham:
Based on those two samples, I can tell what cloud the host is in from the short name.

rade:
> Based on those two samples, I can tell what cloud the host is in from the short name.

Perhaps, but in case of EC2, the FQDN also tells you the region, which is useful.

What's your objection to using the FQDN by default? That it can be quite long and thus clutter the output? I can see how that could be annoying, but IMO the default should favour uniqueness & more info over compactness. As you say, it's easy to override by supplying `-hostname $(hostname)`.

site/troubleshooting.md
```diff
@@ -86,7 +86,11 @@ The terms used here are explained further at
 [how it works](how-it-works.html).
 
 The 'Our name' line identifies the local weave router as a peer in the
-weave network.
+weave network. It displays both the peer name and the peer's nickname
+in parenthesis; the peer name defaults to a MAC address, whilst the
```
rade: We don't provide an option to override the peer name, so saying that it "defaults to" a MAC is wrong. Also "a MAC"? *what* MAC? Yes, how-it-works (which, incidentally, needs updating to mention the nickname) contains similar wording but the context there is the name *format*, i.e. "defaults to a MAC" means "the default format of a name is MAC" (I'm not exactly delighted with that phrasing either though).

weaver/main.go
```diff
@@ -87,6 +89,14 @@ func main() {
...
 	}
 
+	if nickName == "" {
+		osHostname, err := os.Hostname()
+		if err != nil {
+			log.Fatal(err)
+		}
+    	nickName = osHostname
+	}
+
```
rade: The indentation has gone rather wrong in the above. Also, might as well assign straight to nickName.

weaver/main.go
```diff
@@ -49,6 +50,7 @@ func main() {
 	flag.BoolVar(&justVersion, "version", false, "print version and exit")
 	flag.StringVar(&ifaceName, "iface", "", "name of interface to read from")
 	flag.StringVar(&routerName, "name", "", "name of router (defaults to MAC)")
+	flag.StringVar(&nickName, "nickname", "", "nickname of host (defaults to container hostname)")
```
rade: You can't mention 'container' here; this code has no idea it's running as a container, and indeed might not.

rade:
IF we are going to go with `hostname -f` as the default, then the go-level default should also be the FQDN. How hard is that?

awh:
Go question - which is preferred?
As the code is now:
````
	if nickName == "" {
		osHostname, err := os.Hostname()
		if err != nil {
			log.Fatal(err)
		}
		nickName = osHostname
	}
````
Assigning directly to nickName (NB using '=' instead of ':='); now we're depending upon and overwriting an earlier 'err' declaration:
````
	if nickName == "" {
		nickName, err = os.Hostname()
		if err != nil {
			log.Fatal(err)
		}
	}
````
Explicitly shadow err:
````
	if nickName == "" {
		err := nil
		nickName, err = os.Hostname()
		if err != nil {
			log.Fatal(err)
		}
	}
````
Unfortunately this doesn't work because the inner `nickName` shadows the outer:
````
	if nickName == "" {
		nickName, err := os.Hostname()
		if err != nil {
			log.Fatal(err)
		}
	}
````

**Comment by awh** [comment]
Getting the FQDN in pure Go is doable:

````
func main() {
	hostname, err := os.Hostname()
	if err == nil {
		addrs, err := net.LookupHost(hostname)

		if err == nil {
			names, err := net.LookupAddr(addrs[0])

			if err == nil {
				fmt.Printf("%v", names)
			}
		}
	}
}
````
(Obviously both `LookupHost` and `LookupAddr` can return zero to many items, which the above code doesn't handle). In my Vagrant VM the above code prints this:
````
vagrant@ubuntu-14:~$ ./main
[ubuntu-14.04-amd64-vbox ubuntu-14]
````
Which looks to me like the `LookupAddr` function isn't parsing out the alias from the `hosts` file:
````
vagrant@ubuntu-14:~$ cat /etc/hosts
127.0.0.1	localhost
127.0.1.1	ubuntu-14.04-amd64-vbox	ubuntu-14
````

**Comment by rade** [comment]
> Go question - which is preferred?

The 2nd one. Nothing wrong with re-using `err`.

> Getting the FQDN in pure Go is doable.

That's horrific. going via reverse lookup. Yuck.

**Comment by awh** [comment]
:) As far as I am aware, this is exactly what `hostname -f` is doing underneath, albeit via `libnsl` to parse `nsswitch.conf`, `resolv.conf` and `hosts`...

**Comment by awh** [comment]
Shelling out to `hostname -f` from Go is an option I suppose :)

**Comment by rade** [comment]
> Shelling out to `hostname -f` from Go is an option I suppose :)

Sure. If we want to undo all the work we've done to avoid using the native resolver libraries. Oh and to producer an image containing just a single executable.

I'm afraid `hostname`/`os.Hostname()` is doing to have to be the default.

**Comment by awh** [comment]
> Sure. If we want to undo all the work we've done to avoid using the native resolver libraries. Oh and to producer an image containing just a single executable.

I was half joking, but learned something useful! Is the desire to avoid the resolver related to the warnings in #5?

> I'm afraid hostname/os.Hostname() is doing to have to be the default.

In that case as far as I am aware the currently implemented behaviour matches the now finalised requirements...

**Comment by rade** [comment]
>  Is the desire to avoid the resolver related to the warnings in #5?

Yes. But off-topic.

File: router/gossip_test.go [hunk 6]
```diff
@@ -19,7 +19,7 @@ type mockChannelConnection struct {
 // We need to create some dummy channels otherwise tests hang on nil
 // channels when Router.OnGossip() calls async methods.
 func NewTestRouter(name PeerName) *Router {
-	router := NewRouter(nil, name, nil, 10, 1024, nil)
+	router := NewRouter(nil, name, "hostname", nil, 10, 1024, nil)
```
rade: I reckon "" makes a better test nickname here and elsewhere.

weave
```diff
@@ -475,18 +475,31 @@ case "$COMMAND" in
...
+                    ;;
+            esac
+        done
         # Set WEAVE_DOCKER_ARGS in the environment in order to supply
         # additional parameters, such as resource limits, to docker
         # when launching the weave container.
         CONTAINER=$(docker run --privileged -d --name=$CONTAINER_NAME \
             -p $PORT:$PORT/tcp -p $PORT:$PORT/udp -e WEAVE_PASSWORD \
-            $WEAVE_DOCKER_ARGS $IMAGE -name $MACADDR -iface $CONTAINER_IFNAME "$@")
+            $WEAVE_DOCKER_ARGS $IMAGE -name $MACADDR -iface $CONTAINER_IFNAME -nickname "$NICKNAME" "$@")
```
rade: Nitpick: The arg order here should be: iface,name,nickname. To match the order of the arg parsing declaration and process in main.go, and the order of args in NewRouter.

awh:
Above commits address comments to date.

rade:
You missed a fair few places where "hostname" is used as a nickname in tests.

awh:
Apologies, I neglected to look further a field than the specific file you commented on. Fixed.

Issue #229, opened 2014-11-24
automated system tests


rade:
We should have some integration / full-system tests. A few things to cover, vaguely in order of priority
1. basic `weave *` command execution. Designed to catch silly mistakes we may make in the weave script. Which has happened. Single node suffices.
2. weave network testing. This does require multiple nodes. Most basic scenario is firing up two containers across hosts and making sure they can ping each other. But ultimately we should have tests for all weave features. We also need to include some qualitative performance tests, to catch things like forgetting to run ethtool and hence ending up with a working but abysmally performing network.
3. weavedns testing. ditto.
4. kernel versions. Would be good to test all the above across different kernel versions so we get alerted when we inadvertently introduce functionality that depends on specific kernels. We've been caught out by this already.
5. docker versions. ditto.
6. library versions. We should have tests that check our code works with the latest version of the various libs we depend on. We have been caught out by this. See also #228.

I see this specific issue as an umbrella for ideas; testing tends to be an open-ended issue where we can always come up with ideas for improvement. So let's record them here and then split out "actionable" pieces into their own issues.

Also, note that this issue is _not_ about CI.

rade:
It occurs to me that we could test all kinds of weave functionality by running a bunch of weave router containers on the same machine. This is lightweight - allowing us to simulate large(ish) networks and iterate quickly through different configurations.

Things like connection handling, topology, gossip can be tested that way. Basically anything that doesn't actually require capturing/injecting packets. We could test most of the [IP assignment](#22) too that way though would have to simulate the request/release.

Local iptable rules could simulate various network topologies and failures. We could have a generator for overlay topologies, i.e. a bunch of peers and the possible connectivity between them (tcp, udp one way, udp the other). Translate that into iptable rules.

dpw:
My experience of testing distributed systems is that most of the work is in testing failure conditions (and that any recovery happens as expected).  These tend to involve timeouts.  So to test with adequate coverage, you need to be able to test with simulated time.  (Bryan did mention that he has already done something like this in the tests for the new gossip protocol.)

If we are doing thorough testing, I'd strongly recommend investing effort in that direction.  I.e., simulate the network and time, and run multiple instances of weave (or some subset of its components) within a single process.

At the other end of the spectrum, we have the smoke tests.  We should continue to enhance these, because they are a full system test, including the shell script.  But I expect they will always be "wide but shallow".

rade:
> Also, note that this issue is *not* about CI.

#397 is.

PR #413, opened 2015-02-23
Tear down connections on prolonged loss of UDP heartbeat

awh:
Addresses #373 - detect udp connectivity breakage.

Once a connection moves to the established state indicating that the remote peer has received one of our heartbeats a timer is started. If we do not receive a UDP heartbeat from the remote peer within this time (default is three times the slow heartbeat interval) the connection is terminated:

    connection shutting down due to error: timed out waiting for UDP heartbeat

At this point the existing connection resumption mechanism takes over. In normal operation, the timer is reset each time we receive a heartbeat.

Implementation notes:

* It would be possible to merge `establishedTimeout` and `heartbeatTimeout` into a single timer; we would simply adjust duration and error messages depending on `conn.established` state.
* Alternatively, the heartbeat timeout could run at the same time as the established timeout rather than being kicked off once we know our heartbeat has got through to the other side.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -37,27 +37,26 @@ type RemoteConnection struct {
 type LocalConnection struct {
 	sync.RWMutex
 	RemoteConnection
-	TCPConn            *net.TCPConn
-	tcpSender          TCPSender
-	remoteUDPAddr      *net.UDPAddr
-	receivedHeartbeat  bool
-	stackFrag          bool
-	effectivePMTU      int
-	SessionKey         *[32]byte
-	establishedTimeout *time.Timer
-	heartbeatTimeout   *time.Timer
-	heartbeatFrame     *ForwardedFrame
-	heartbeat          *time.Ticker
-	fragTest           *time.Ticker
-	forwardChan        chan<- *ForwardedFrame
-	forwardChanDF      chan<- *ForwardedFrame
-	stopForward        chan<- interface{}
-	stopForwardDF      chan<- interface{}
-	verifyPMTU         chan<- int
-	Decryptor          Decryptor
-	Router             *Router
-	uid                uint64
-	queryChan          chan<- *ConnectionInteraction
+	TCPConn           *net.TCPConn
+	tcpSender         TCPSender
+	remoteUDPAddr     *net.UDPAddr
+	receivedHeartbeat bool
+	stackFrag         bool
+	effectivePMTU     int
+	SessionKey        *[32]byte
+	heartbeatTimeout  *time.Timer
+	heartbeatFrame    *ForwardedFrame
+	heartbeat         *time.Ticker
+	fragTest          *time.Ticker
+	forwardChan       chan<- *ForwardedFrame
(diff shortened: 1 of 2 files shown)


router/connection.go
```diff
@@ -329,6 +332,9 @@ func (conn *LocalConnection) handleReceivedHeartbeat(remoteUDPAddr *net.UDPAddr)
 	conn.remoteUDPAddr = remoteUDPAddr
 	conn.receivedHeartbeat = true
 	conn.Unlock()
+	if conn.established {
```
rade: Wouldn't it be more obvious to check for conn.heartbeatTimeout != nil?

rade:
What is the rationale for starting the heartbeat timer when becoming 'established'? The latter tells us that *our* heartbeats did get through, which says nothing about the other direction, which is what the timer is watching for.

Wouldn't it be better to instead start the timer after sending 'established' to the other side?

rade:
> Wouldn't it be better to instead start the timer after sending 'established' to the other side?

Or, why not start the timer straight away? That would simplify the code since then a) the timer would always exist, and b) we won't have to check for a condition to start it.

I think that would also allow us to get rid of the 'established' timeout, i.e. we could rely on the heartbeat timeout to tear down the connection instead.

awh:
I was thinking of the case where we are originating from behind NAT; there's no way we can receive a heartbeat from the remote peer until our outbound connection is established, so I deferred starting the timer until that had happened.

I like the idea of simplifying down to a single timer very much, although it would have the effect of slightly reducing the rigour of the test (e.g. timely delivery of the established message over the TCP channel would no longer be checked) - are you happy with that? If not, I'd suggest treating the two cases as completely orthogonal and start both timers in `run` so that we can eliminate the conditional checks...

rade:
> there's no way we can receive a heartbeat from the remote peer until our outbound connection is established

"almost established", i.e. our UDP packet must have made it across, but the Established TCP message may not have turned up yet.

So technically the initiating side should have a slightly longer timeout, but I can't imagine it's worth the hassle.

> timely delivery of the established message over the TCP channel would no longer be checked

The ReadTimeout puts an upper bound on that.

awh:
Above commits address comments to date.

rade:
Great. Does it work? How have you tested this?

awh:
Confirmed with manual testing on two digital ocean VMs by using

    iptables -I FORWARD 1 -p udp --dport 6783 -j DROP

to prevent weave related UDP traffic from being forwarded to the `docker0` bridge. Tried the following cases:

1. UDP traffic is blocked prior to launch; ensure that connection establishment times out but continues retrying
2. Remove block; ensure connection establishes
3. Restore block; ensure weave notices within configured timeout, tears down connection but continues retrying
4. Remove block; ensure connection restablishes
5. Left running for half an hour to observe stability

I used a combination of `docker logs weave` and `weave status` to observe the state on both peers at each step.

Would feel a lot more comfortable with #229 in place! It could use soak/torture testing...

Issue #411, opened 2015-02-23
"Too many open files" messages


bboreham:
A user writes: 

We are getting weave routers stuck in a loop printing "weave 2015/02/23 06:51:31.081132 accept tcp4 0.0.0.0:6783: too many open files” in the logs. It’s filling up the disk space because it never stops logging that line again and again.

In this case, it’s using all 1024 available FDs:

````
    $ lsof -a -p 9821

    COMMAND  PID USER   FD   TYPE DEVICE SIZE/OFF   NODE NAME
    weaver  9821 root  cwd    DIR   0,40     4096     12 /home/weave
    weaver  9821 root  rtd    DIR   0,40     4096      2 /
    weaver  9821 root  txt    REG   0,40 11348166     20 /home/weave/weaver
    weaver  9821 root  mem    REG    0,7           18729 socket:[18729] (stat: No such file or directory)
    weaver  9821 root  mem    REG    0,7           18727 socket:[18727] (stat: No such file or directory)
    weaver  9821 root    0r   CHR    1,3      0t0  17540 /dev/null
    weaver  9821 root    1w  FIFO    0,8      0t0  17291 pipe
    weaver  9821 root    2w  FIFO    0,8      0t0  17292 pipe
    weaver  9821 root    3r   CHR    1,9      0t0  17671 /dev/urandom
    weaver  9821 root    4u  sock    0,7      0t0  18727 can't identify protocol
    weaver  9821 root    5u  sock    0,7      0t0  18729 can't identify protocol
    weaver  9821 root    6u  sock    0,7      0t0  18730 can't identify protocol
    weaver  9821 root    7u  0000    0,9        0   5258 anon_inode
    weaver  9821 root    8u  sock    0,7      0t0  18731 can't identify protocol
    weaver  9821 root    9u  sock    0,7      0t0  18733 can't identify protocol
    weaver  9821 root   10u  sock    0,7      0t0  20147 can't identify protocol
    [...]
    weaver  9821 root   50u  sock    0,7      0t0 901856 can't identify protocol
    weaver  9821 root   51u  sock    0,7      0t0 901884 can't identify protocol
    weaver  9821 root   52u  sock    0,7      0t0 902019 can't identify protocol
    weaver  9821 root   53u  sock    0,7      0t0 902049 can't identify protocol
    weaver  9821 root   54u  sock    0,7      0t0 902100 can't identify protocol
    […]
````

bboreham:
Logs from the weave process:

````
[...]

PR #414, opened 2015-02-23
Add 'finished' chan to stop callers blocking on closed connection

bboreham:
Addresses a source of deadlock in weave arising from LocalConnection's habit of exiting its queryLoop() routine while callers are trying to talk to it.  After there are 16 messages in the channel, the sends block, and some of them can be from places where locks are held.

This PR introduces a second 'finished' channel, which is closed on exit from queryLoop().  All callers select on the 'finished' channel as well as the 'query' channel, and since a closed channel is always ready to read, they do not block.

This issue was found while investigating #411, but without stack traces I cannot be absolutely certain that this is the cause of that incident.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -57,6 +57,7 @@ type LocalConnection struct {
 	Router             *Router
 	uid                uint64
 	queryChan          chan<- *ConnectionInteraction
+	finished           <-chan struct{} // closed to signal that queryLoop has finished
 }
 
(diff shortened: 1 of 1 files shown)


rade:
I think you want a `sendQuery(code, payload)` helper.

bboreham:
Re-pushed with `sendQuery()` helper

Issue #415, opened 2015-02-24
Another deadlock


bboreham:
I created this situation by instantiating 22 peers and connecting them all up at once; this resulted in a number of connections being added and then removed; each operation results in gossip to all peers.

Excerpt from logs showing rapid sequence of events:

````
weave 2015/02/24 14:38:38.008336 ->[1a:9a:a4:30:2f:2c]: connection shutting down due to error: write tcp4 172.17.2.10:46783: connection reset by peer
weave 2015/02/24 14:38:38.008396 ->[02:31:50:de:53:4f]: connection shutting down due to error: write tcp4 172.17.2.6:41884: connection reset by peer
weave 2015/02/24 14:38:38.008865 ->[b2:1e:e1:32:d8:2c]: connection shutting down due to error: Connection limit reached (10)
weave 2015/02/24 14:38:38.008883 ->[1e:97:15:ae:b3:63]: connection shutting down due to error: Connection limit reached (10)
weave 2015/02/24 14:38:38.008901 ->[6a:5e:7b:a5:0b:e9]: connection shutting down due to error: Connection limit reached (10)
weave 2015/02/24 14:38:38.009185 ->[f2:63:b4:c6:86:05]: connection shutting down due to error: Connection limit reached (10)
weave 2015/02/24 14:38:38.009424 ->[2a:b5:e7:be:cd:b8]: connection shutting down due to error: Connection limit reached (10)
weave 2015/02/24 14:38:38.009977 ->[ca:a6:00:31:4a:59]: connection shutting down due to error: Connection limit reached (10)
````

The deadlock is that LocalPeer queryLoop is blocked on sending a topology update gossip message to a connection:

````
goroutine 21 [select, 32 minutes]:
github.com/zettio/weave/router.(*LocalConnection).sendQuery(0xc2093905a0, 0x0, 0x82d560, 0xc20a161440)
	zettio/weave/router/connection.go:196 +0xdf
github.com/zettio/weave/router.(*LocalConnection).SendProtocolMsg(0xc2093905a0, 0x5, 0xc20a168380, 0x2cc, 0x33a)
	zettio/weave/router/connection.go:228 +0x70
github.com/zettio/weave/router.func·008(0x3e1d5fea9d09, 0x7f00c85ba478, 0xc2093905a0)
	zettio/weave/router/gossip.go:121 +0x73
github.com/zettio/weave/router.(*Peer).ForEachConnection(0xc208004300, 0x7f00ca0e8c88)
[...]

PR #423, opened 2015-02-25
Avoid some deadlocks on full channel buffer

bboreham:
We have seen in #415 how deadlocks can occur when channel buffers fill up; raising the capacity from 16 to 1024 makes this less likely.

It is not a final fix for the problem, since outgoing TCP delays or other reasons can still cause a channel to back up.

This problem was discovered while researching #411.


bboreham:
#428 is a better fix for the problems I saw, so closing this PR.

PR #425, opened 2015-02-26
Add some basic integration tests

achanda:
Added some basic tests for launching a container and CLI help

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,10 @@ tests:
 	cd router; go test -cover -tags netgo
 	cd nameserver; go test -cover -tags netgo
 
+test-integration:
+	$(SUDO) cp $(CURDIR)/weave /usr/local/bin/weave
+	cd integration-cli; go test -cover
+
 $(PUBLISH): publish_%:
 	$(SUDO) docker tag -f $(DOCKERHUB_USER)/$* $(DOCKERHUB_USER)/$*:$(WEAVE_VERSION)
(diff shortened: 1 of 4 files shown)


achanda:
This sets up a basic outline for integration tests. More tests can be added once everyone agrees with this one. Also, I'm pretty new to Go and the Docker ecosystem, so please let me know if anything looks bad.

squaremo:
Did you see that there are smoke tests -- https://github.com/zettio/weave/tree/master/test

I'm not sure how this improves on those: did you have some particular advantage in mind?

achanda:
This was meant to be integration tests for the CLI and not end to end tests for full functionalities. While functional tests do test CLI indirectly, it will be nicer if CLI bugs can be caught earlier. So, the order of tests for each commit will be:
1. Unit tests
2. CLI integration tests
3. Other integration tests

If step 2 fails, the CI system will report back error. Having separate integration tests for the CLI can save some time in running full integration tests in case of a bug. I was looking at number 1 in #229 . Please let me know if I misunderstood.

rade:
We are testing more and more of the CLI in the integration tests. While I can see the value of separate CLI tests, I don't think the cost is justified.

achanda:
@rade these were meant to be integration tests, not CLI tests. I added some basic CLI checks as a starting point. The current integration tests are shell scripts and should be moved to a proper testing framework IMHO. This patch was starting to do just that.

Issue #225, opened 2014-11-23
[dns] caching for local domain


rade:
At present, weavedns first attempts to resolve names against the records from local containers. If that fails then it asks other peers and returns the first answer. This has some significant shortcomings...

1. every query for a name that belongs to a remote container requires a broadcast, network round trip, which introduces latency and increases network and CPU load, and doesn't scale.

2. the same occurs for a name that doesn't resolve at all

3. if the name is recorded more than once, we are entirely at the mercy of underlying network conditions etc, as to what answer we return, or even if we return an answer at all

The obvious solution is to introduce caching.

In its most basic form a dns peer would:

- remember all answers returned from a broadcast query, including their ttls
- if no answers are returned (based on some timeout), remember that too, again with a ttl
- expire entries based on their ttl
- answer queries from the cache if possible

The shortcomings of this basic approach are principally around the danger of returning stale entries vs achieving good performance and scalability. Still, I reckon it's good enough for starters, and even with a short ttl an improvement over the present situation.

rade:
One possible model is to use a mixture of push & pull. Essentially we take the pull+cache approach from #429 and add a mechanism by which peers a) monitor each other, and b) inform each other when names that have been queried have disappeared. The former would remove entries from the cache when the peer on which the associated containers reside disappears. The latter would remove entries from the cache for containers that have stopped. The latter could use s.t. like a bloom filter for compactly recording which names have been asked for by which peers, so that we don't end up in effect broadcasting container lifecycle events to all peers.

With that in place we can increase the TTL significantly. Except we can't...

We'd also need to find a way to flush cache entries when new name associations are created, e.g. when containers are started.

So perhaps the thing to do is to gossip the information about which peer is interested in which names.

inercia:
So what do you think about this?

* when a WeaveDNS instance _X_ receives a mDNS query from _Y_ for a local name `name1`, it answers the query with a 30s TTL, and registers this information as (eg, _"Y is an observer of name1"_). It could be useful to also register the TTL sent to _Y_  (eg, _"Y is an observer of name1 and it will not ask us again before now+30s"_)
* when `name1` disappears, _X_ will update its local cache and send an update to _Y_ (and any other observers) about that name being removed (maybe with a DNS `UPDATE` message #265 ??). If the TTL was registered and that answer was too long ago, _X_ will skip this message as the answer cached at _Y_ is too old anyway, so it will need to ask _X_...

inercia:
[...]

Issue #226, opened 2014-11-23
[dns] basic load balancing for services


rade:
There is a rather obvious opportunity for weave to facilitate basic load balancing and fault resilience... when, using weavedns, multiple containers are entered into dns under the same name, weave could implicitly treat these as a load balance group, returning all the addresses in a random order.

This requires [caching](225).

awh:
Scope of this issue has been reduced to load balancing only for 0.11.0. The implementation will return a single randomly chosen address from the possible set of addresses for a name.

Issue #364, opened 2015-01-26
[dns] idiomatic support for containers with more than one name


rade:
With weavedns one can create multiple DNS records for containers - with PUTs on the weavedns HTTP API. There are a few problems with that...

1. Adding the records requires peeking under the hood of weavedns, i.e. there is no corresponding weave or docker command.

2. The extra records do not survive a weavedns restart, since, unlike the hostname, the information is not part of the container's metadata and weavedns does not itself persist any data.

3. weavdns only deals in A records; whereas for some use cases CNAMEs would be more appropriate.

Related to #226.

inercia:
Any suggestion on how to solve the second issue? Do you suggest to add some persistence mechanism to WeaveDNS?

rade:
>  Do you suggest to add some persistence mechanism to WeaveDNS?

No! https://github.com/docker/docker/pull/9013 would offer a solution.

inercia:
I presumed you were not suggesting that... :) 

So I guess this issue is blocked until docker/docker#9013 is solved...

rade:
I've captured points 2 and 3 in #635 and #636, respectively.

Issue #338, opened 2015-01-10
[dns] return all matching local records


rade:
ATM, weavedns only returns the first matching record for a local name / reverse lookup. But there can be multiple matching records; all of them should be returned.

rade:
This should apply to both direct queries and queries received over mDNS.

@squaremo @bboreham any gotchas here? I suppose it could be deemed somewhat inconsistent/strange/surprising to respond with all matching records from one peer, but not across all peers.

squaremo:
> any gotchas here?

Nope, don't think so. There is still the problem of having a consistent way of answering a query for which there are multiple answers on the network, which this may or may not help; *just* doing this will muddy the waters a bit, since instead of getting one answer from some host or no answer, a question will get some answers *all from the same host* or no answer.

inercia:
@rade Current WeaveDNS can return all the answers with "--max-answers=0", so is this issue still valid?  If so, what is missing?

rade:
Users shouldn't have to supply special configuration parameters for something that should just work.

I suspect this issue will get resolved as part of #733.

rade:
> I suspect this issue will get resolved as part of #733.

Pretty sure that is the case, in the light of #826.

PR #429, opened 2015-03-02
weave-225: [dns] caching (1/3)

inercia:
#225 : this pull request implements a caching mechanism in the DNS server. The cache is used for saving positive and negative replies obtained by the local discovery system as well as by the recursive resolver.

Replies found in the cache are modified before being returned to clients by adjusting the remaining TTL and  shuffling answers (implementing a basic load balancing mechanism as a side effect, a first step for #226).

### DIFF
--- a/nameserver/cache.go
+++ b/nameserver/cache.go
@@ -1,0 +1,370 @@
+package nameserver
+
+import (
+	"container/heap"
+	"errors"
+	"github.com/miekg/dns"
+	. "github.com/zettio/weave/common"
+	"math"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+var (
+	errInvalidCapacity = errors.New("Invalid cache capacity")
+	errCouldNotResolve = errors.New("Could not resolve")
+	errTimeout         = errors.New("Timeout while waiting for resolution")
+	errNoLocalReplies  = errors.New("No local replies")
+)
+
+const (
+	defPendingTimeout int = 5  // timeout for a resolution
+)
+
+type entryStatus uint8
+
+const (
+	stPending  entryStatus = iota // someone is waiting for the resolution
+	stResolved entryStatus = iota // resolved
+)
+
+const (
+	CacheNoLocalReplies uint8 = 1 << iota // not found in local network (stored in the cache so we skip another local lookup or some time)
+)
+
+// shuffleAnswers reorders answers for very basic load balancing
+func shuffleAnswers(answers []dns.RR) []dns.RR {
+	n := len(answers)
+	if n > 1 {
+		rand.Seed(time.Now().UTC().UnixNano())
+
+		for i := 0; i < n; i++ {
+			r := i + rand.Intn(n-i)
+			answers[r], answers[i] = answers[i], answers[r]
+		}
+	}
+
+	return answers
+}
+
+// a cache entry
+type cacheEntry struct {
+	Status   entryStatus // status of the entry
+	Flags    uint8       // some extra flags
+	ReplyLen int
+
+	question dns.Question
+	protocol dnsProtocol
+	reply    dns.Msg
+
+	validUntil time.Time // obtained from the reply and stored here for convenience/speed
+	putTime    time.Time
+
+	waitChan chan struct{}
+
+	index int // for fast lookups in the heap
+}
+
+func newCacheEntry(question *dns.Question, reply *dns.Msg, status entryStatus, flags uint8, now time.Time) *cacheEntry {
+	e := &cacheEntry{
+		Status:   status,
+		Flags:    flags,
+		question: *question,
+		index:    -1,
+	}
+
+	if e.Status == stPending {
+		e.validUntil = now.Add(time.Duration(defPendingTimeout) * time.Second)
+		e.waitChan = make(chan struct{})
+	} else {
+		e.setReply(reply, flags, now)
+	}
+
+	return e
+}
+
+// Get a copy of the reply stored in the entry, but with some values adjusted like the TTL
+func (e *cacheEntry) getReply(request *dns.Msg, maxLen int, now time.Time) (*dns.Msg, error) {
+	if e.Status != stResolved {
+		return nil, nil
+	}
+
+	// if the reply has expired or is invalid, force the caller to start a new resolution
+	if e.hasExpired(now) {
+		return nil, nil
+	}
+
+	if e.Flags&CacheNoLocalReplies != 0 {
+		return nil, errNoLocalReplies
+	}
+
+	if e.ReplyLen >= maxLen {
+		Debug.Printf("[cache] returning truncated reponse: %d > %d", e.ReplyLen, maxLen)
+		return makeTruncatedReply(request), nil
+	}
+
+	// create a copy of the reply, with values for this particular query
+	reply := e.reply
+	reply.SetReply(request)
+	reply.Rcode = e.reply.Rcode
+	reply.Authoritative = true
+
+	// adjust the TTLs
+	passedSecs := uint32(now.Sub(e.putTime).Seconds())
+	for _, rr := range reply.Answer {
+		ttl := rr.Header().Ttl
+		if passedSecs < ttl {
+			rr.Header().Ttl = ttl - passedSecs
+		} else {
+			rr.Header().Ttl = 0
+		}
+	}
+
+	// shuffle the values, etc...
+	reply.Answer = shuffleAnswers(reply.Answer)
+
+	return &reply, nil
+}
+
+func (e cacheEntry) hasExpired(now time.Time) bool {
+	return e.validUntil.Before(now) || e.validUntil == now
+}
+
+// set the reply for the entry
+// returns True if the entry has changed the validUntil time
+func (e *cacheEntry) setReply(reply *dns.Msg, flags uint8, now time.Time) bool {
+	shouldNotify := (e.Status == stPending)
+
+	var prevValidUntil time.Time
+	if e.Status == stResolved {
+		prevValidUntil = e.validUntil
+	}
+
+	e.Status = stResolved
+	e.Flags = flags
+	e.putTime = now
+
+	if e.Flags&CacheNoLocalReplies != 0 {
+		// use a fixed timeout for negative local resolutions
+		e.validUntil = now.Add(time.Second * time.Duration(negLocalTTL))
+	} else {
+		// calculate the validUntil from the reply TTL
+		var minTtl uint32 = math.MaxUint32
+		for _, rr := range reply.Answer {
+			ttl := rr.Header().Ttl
+			if ttl < minTtl {
+				minTtl = ttl // TODO: improve the minTTL calculation (maybe we should skip some RRs)
+			}
+		}
+		e.validUntil = now.Add(time.Second * time.Duration(minTtl))
+	}
+
+	if reply != nil {
+		e.reply = *reply
+		e.ReplyLen = reply.Len()
+	}
+
+	if shouldNotify {
+		close(e.waitChan) // notify all the waiters by closing the channel
+	}
+
+	return (prevValidUntil != e.validUntil)
+}
+
+// wait until a valid reply is set in the cache
+func (e *cacheEntry) waitReply(request *dns.Msg, timeout time.Duration, maxLen int, now time.Time) (*dns.Msg, error) {
+	if e.Status == stResolved {
+		return e.getReply(request, maxLen, now)
+	}
+
+	if timeout > 0 {
+		select {
+		case <-e.waitChan:
+			return e.getReply(request, maxLen, now)
+		case <-time.After(time.Second * timeout):
+			return nil, errTimeout
+		}
+	}
+
+	return nil, errCouldNotResolve
+}
+
+func (e *cacheEntry) close() {
+	if e.Status == stPending {
+		close(e.waitChan)
+	}
+}
+
+//////////////////////////////////////////////////////////////////////////////////////
+
+// An entriesPtrHeap is a min-heap of cache entries.
+type entriesPtrsHeap []*cacheEntry
+
+func (h entriesPtrsHeap) Len() int           { return len(h) }
+func (h entriesPtrsHeap) Less(i, j int) bool { return h[i].validUntil.Before(h[j].validUntil) }
+func (h entriesPtrsHeap) Swap(i, j int) {
+	h[i], h[j] = h[j], h[i]
+	h[i].index = i
+	h[j].index = j
+}
+
+func (h *entriesPtrsHeap) Push(x interface{}) {
+	// Push and Pop use pointer receivers because they modify the slice's length,
+	// not just its contents.
+	n := len(*h)
+	entry := x.(*cacheEntry)
+	entry.index = n
+	*h = append(*h, entry)
+}
+
+func (h *entriesPtrsHeap) Pop() interface{} {
+	old := *h
+	n := len(old)
+	item := old[n-1]
+	item.index = -1 // for safety
+	*h = old[0 : n-1]
+	return item
+}
+
+//////////////////////////////////////////////////////////////////////////////////////
+
+type cacheKey dns.Question
+type entries map[cacheKey]*cacheEntry
+
+// Cache is a thread-safe fixed capacity LRU cache.
+type Cache struct {
+	Capacity int
+
+	entries  entries
+	entriesH entriesPtrsHeap // len(entriesH) <= len(entries), as pending entries can be in entries but not in entriesH
+	lock     sync.RWMutex
+}
+
+// NewCache creates a cache of the given capacity
+func NewCache(capacity int) (*Cache, error) {
+	if capacity <= 0 {
+		return nil, errInvalidCapacity
+	}
+	c := &Cache{
+		Capacity: capacity,
+		entries:  make(entries, capacity),
+	}
+
+	heap.Init(&c.entriesH)
+	return c, nil
+}
+
+// Clear removes all the entries in the cache
+func (c *Cache) Clear() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	c.entries = make(entries, c.Capacity)
+	heap.Init(&c.entriesH)
+}
+
+// Purge removes the old elements in the cache
+func (c *Cache) Purge(now time.Time) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	for i, entry := range c.entriesH {
+		if entry.hasExpired(now) {
+			heap.Remove(&c.entriesH, i)
+			delete(c.entries, cacheKey(entry.question))
+		} else {
+			return // all remaining entries must be still valid...
+		}
+	}
+}
+
+// Add adds a reply to the cache.
+func (c *Cache) Put(request *dns.Msg, reply *dns.Msg, flags uint8, now time.Time) int {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	question := request.Question[0]
+	key := cacheKey(question)
+	ent, found := c.entries[key]
+	if found {
+		Debug.Printf("[cache msgid %d] replacing response in cache", request.MsgHdr.Id)
+		updated := ent.setReply(reply, flags, now)
+		if updated {
+			heap.Fix(&c.entriesH, ent.index)
+		}
+	} else {
+		// If we will add a new item and the capacity has been exceeded, make some room...
+		if len(c.entriesH) >= c.Capacity {
+			lowestEntry := heap.Pop(&c.entriesH).(*cacheEntry)
+			lowestEntry.close()
+			delete(c.entries, cacheKey(lowestEntry.question))
+		}
+		ent = newCacheEntry(&question, reply, stResolved, flags, now)
+		heap.Push(&c.entriesH, ent)
+		c.entries[key] = ent
+	}
+	return ent.ReplyLen
+}
+
+// Look up for a question's reply from the cache.
+// If no reply is stored in the cache, it returns a `nil` reply and no error. The caller can then `Wait()`
+// for another goroutine `Put`ing a reply in the cache.
+func (c *Cache) Get(request *dns.Msg, maxLen int, now time.Time) (reply *dns.Msg, err error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	question := request.Question[0]
+	key := cacheKey(question)
+	if ent, found := c.entries[key]; found {
+		reply, err = ent.getReply(request, maxLen, now)
+		if ent.hasExpired(now) {
+			Debug.Printf("[cache msgid %d] expired: removing", request.MsgHdr.Id)
+			if ent.index > 0 {
+				heap.Remove(&c.entriesH, ent.index)
+			}
+			delete(c.entries, key)
+			reply = nil
+		}
+	} else {
+		// we are the first asking for this name: create an entry with no reply... the caller must wait
+		Debug.Printf("[cache msgid %d] addind in pending state", request.MsgHdr.Id)
+		c.entries[key] = newCacheEntry(&question, nil, stPending, 0, now)
+	}
+	return
+}
+
+// Wait for a reply for a question in the cache
+// Notice that the caller could Get() and then Wait() for a question, but the corresponding cache
+// entry could have been removed in between. In that case, the caller should retry the query (and
+// the user should increase the cache size!)
+func (c *Cache) Wait(request *dns.Msg, timeout time.Duration, maxLen int, now time.Time) (reply *dns.Msg, err error) {
+	// do not try to lock the cache: otherwise, no one else could `Put()` the reply
+	question := request.Question[0]
+	if entry, found := c.entries[cacheKey(question)]; found {
+		reply, err = entry.waitReply(request, timeout, maxLen, now)
+	}
+	return
+}
+
+// Remove removes the provided question from the cache.
+func (c *Cache) Remove(question *dns.Question) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	key := cacheKey(*question)
+	if entry, found := c.entries[key]; found {
+		if entry.index > 0 {
+			heap.Remove(&c.entriesH, entry.index)
+		}
+		delete(c.entries, key)
+	}
+}
+
+// Len returns the number of entries in the cache.
+func (c *Cache) Len() int {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return len(c.entries)
+}
--- a/nameserver/server.go
+++ b/nameserver/server.go
@@ -1,29 +1,26 @@
 package nameserver
 
 import (
+	"bytes"
 	"fmt"
 	"github.com/miekg/dns"
 	. "github.com/zettio/weave/common"
 	"net"
 	"sync"
-	"bytes"
+	"time"
 )
 
 const (
-	DEFAULT_LOCAL_DOMAIN = "weave.local."     // The default name used for the local domain
-	DEFAULT_SERVER_PORT  = 53                 // The default server port
-	DEFAULT_CLI_CFG_FILE = "/etc/resolv.conf" // default "resolv.conf" file to try to load
-	DEFAULT_UDP_BUFLEN   = 4096               // bigger than the default 512
-	DEFAULT_IFACE_NAME   = "default interface"
+	DEFAULT_LOCAL_DOMAIN   = "weave.local."     // The default name used for the local domain
+	DEFAULT_SERVER_PORT    = 53                 // The default server port
+	DEFAULT_CLI_CFG_FILE   = "/etc/resolv.conf" // default "resolv.conf" file to try to load
+	DEFAULT_UDP_BUFLEN     = 4096               // bigger than the default 512
+	DEFAULT_CACHE_LEN      = 8192               // default cache capacity
+	DEFAULT_RESOLV_WORKERS = 8                  // default number of resolution workers
+	DEFAULT_TIMEOUT        = 5                  // default timeout for DNS resolutions
+	DEFAULT_IFACE_NAME     = "default interface"
+	DEFAULT_RESOLV_TRIES   = 3 // max # of times a worker tries to resolve a query
 )
-
-func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
-	m := new(dns.Msg)
-	m.SetReply(r)
-	m.RecursionAvailable = true
-	m.Rcode = dns.RcodeNameError
-	return m
-}
 
 type DNSServerConfig struct {
 	// (Optional) client config file for resolving upstream servers
@@ -34,16 +31,54 @@ type DNSServerConfig struct {
 	Port int
 	// (Optional) local domain (ie, "weave.local.")
 	LocalDomain string
+	// (Optional) cache size
+	CacheLen int
+	// (Optional) timeout for DNS queries
+	Timeout int
+	// (Optional) UDP buffer length
+	UdpBufLen int
+}
+
+type dnsProtocol uint8
+
+const (
+	protUdp dnsProtocol = iota // UDP protocol
+	protTcp dnsProtocol = iota // TCP protocol
+)
+
+func (proto dnsProtocol) String() string {
+	switch proto {
+	case protUdp:
+		return "UDP"
+	case protTcp:
+		return "TCP"
+	}
+	return "unknown"
+}
+
+// get a new dns.Client for a protocol
+func (proto dnsProtocol) GetNewClient(bufsize int) *dns.Client {
+	switch proto {
+	case protTcp:
+		return &dns.Client{Net: "tcp"}
+	case protUdp:
+		return &dns.Client{Net: "udp", UDPSize: uint16(bufsize)}
+	}
+	return nil
 }
 
 type DNSServer struct {
-	udpSrv   *dns.Server
-	tcpSrv   *dns.Server
-	mdnsCli  *MDNSClient
-	mdnsSrv  *MDNSServer
-	zone     Zone
-	iface    *net.Interface
-	upstream *dns.ClientConfig
+	udpSrv      *dns.Server
+	tcpSrv      *dns.Server
+	mdnsCli     *MDNSClient
+	mdnsSrv     *MDNSServer
+	cache       *Cache
+	zone        Zone
+	iface       *net.Interface
+	upstream    *dns.ClientConfig
+	timeout     int
+	udpBuf      int
+	listenersWg *sync.WaitGroup
 
 	Domain     string // the local domain
 	IfaceName  string // the interface where mDNS is working on
@@ -53,8 +88,9 @@ type DNSServer struct {
 // Creates a new DNS server
 func NewDNSServer(config DNSServerConfig, zone Zone, iface *net.Interface) (s *DNSServer, err error) {
 	s = &DNSServer{
-		zone:  zone,
-		iface: iface,
+		zone:        zone,
+		iface:       iface,
+		listenersWg: new(sync.WaitGroup),
 
 		Domain:     DEFAULT_LOCAL_DOMAIN,
 		IfaceName:  DEFAULT_IFACE_NAME,
@@ -79,6 +115,12 @@ func NewDNSServer(config DNSServerConfig, zone Zone, iface *net.Interface) (s *D
 			return nil, err
 		}
 	}
+	if config.Timeout > 0 {
+		s.timeout = config.Timeout
+	}
+	if config.UdpBufLen > 0 {
+		s.udpBuf = config.UdpBufLen
+	}
 	s.mdnsCli, err = NewMDNSClient()
 	if err != nil {
 		return
@@ -90,19 +132,28 @@ func NewDNSServer(config DNSServerConfig, zone Zone, iface *net.Interface) (s *D
 	if iface != nil {
 		s.IfaceName = iface.Name
 	}
+	cacheLen := DEFAULT_CACHE_LEN
+	if config.CacheLen > 0 {
+		cacheLen = config.CacheLen
+	}
+	Debug.Printf("[dns] Initializing cache: %d entries", config.CacheLen)
+	s.cache, err = NewCache(cacheLen)
+	if err != nil {
+		return
+	}
 
 	// create two DNS request multiplexerers, depending on the protocol used by clients
 	// (we use the same protocol for asking upstream servers)
-	mux := func(client *dns.Client) *dns.ServeMux {
+	mux := func(proto dnsProtocol) *dns.ServeMux {
 		m := dns.NewServeMux()
-		m.HandleFunc(s.Domain, s.queryHandler([]Lookup{s.zone, s.mdnsCli}))
-		m.HandleFunc(RDNS_DOMAIN, s.rdnsHandler([]Lookup{s.zone, s.mdnsCli}, client))
-		m.HandleFunc(".", s.notUsHandler(client))
+		m.HandleFunc(s.Domain, s.queryHandler([]Lookup{s.zone, s.mdnsCli}, proto))
+		m.HandleFunc(RDNS_DOMAIN, s.rdnsHandler([]Lookup{s.zone, s.mdnsCli}, proto))
+		m.HandleFunc(".", s.notUsHandler(proto))
 		return m
 	}
 
-	s.udpSrv = &dns.Server{Addr: s.ListenAddr, Net: "udp", Handler: mux(&dns.Client{Net: "udp", UDPSize: DEFAULT_UDP_BUFLEN})}
-	s.tcpSrv = &dns.Server{Addr: s.ListenAddr, Net: "tcp", Handler: mux(&dns.Client{Net: "tcp"})}
+	s.udpSrv = &dns.Server{Addr: s.ListenAddr, Net: "udp", Handler: mux(protUdp)}
+	s.tcpSrv = &dns.Server{Addr: s.ListenAddr, Net: "tcp", Handler: mux(protTcp)}
 
 	return
 }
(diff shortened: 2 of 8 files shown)


squaremo:
> The DNS server is also improved by using a limited number of concurrent resolvers.

Could this be changed in a separate PR, to make things more incremental?

inercia:
@squaremo It could be done, but that would require some surgery in the `server` code, and it would make many parts of the `cache` useless (as the cache is designed for concurrent access)...

inercia:
@squaremo I have reverted the changes in the server, so it uses the old mechanism + the cache...

squaremo:
Thanks @inercia, I'll take another look

nameserver/cache.go
```diff
@@ -0,0 +1,318 @@
...
+
+	return &reply, nil
+}
+
+func (e cacheEntry) hasExpired(now time.Time) bool {
+	return e.validUntil.Before(now) || e.validUntil == now
+}
+
+// set the reply for the entry
+func (e *cacheEntry) setReply(reply *dns.Msg, flags uint8, now time.Time) {
```
squaremo: Naming this (and to a lesser extent) the same as the dns.Msg method is really confusing.

nameserver/cache.go
```diff
@@ -0,0 +1,318 @@
...
+
+type entryStatus uint8
+
+const (
+	stPending  entryStatus = iota // someone is waiting for the resolution
+	stResolved entryStatus = iota // resolved
+)
+
+const (
+	CacheLocalReply uint8 = 1 << iota // the reply was obtained from a local resolution
```
squaremo: Does this really need to be a bitset rather than just the one boolean?

nameserver/cache.go
```diff
@@ -0,0 +1,318 @@
...
+	// if the reply has expired or is invalid, force the caller to start a new resolution
+	if e.hasExpired(now) {
+		return nil, nil
+	}
+
+	// create a copy of the reply, with values for this particular query
+	reply := e.reply
+	reply.SetReply(request)
+	reply.Rcode = e.reply.Rcode
+	reply.Authoritative = (e.Flags&CacheLocalReply != 0) // we are only authoritative for local questions
```
squaremo: Is this what gofmt did -- ugh

nameserver/cache.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// adjust the TTLs
+	passedSecs := uint32(now.Sub(e.putTime).Seconds())
```
squaremo: I don't think you need both `putTime` and `validUntil`; here you can use `validUntil - now` for the TTL, and `putTime` isn't used anywhere else.

nameserver/cache.go
inercia: I considered that option, but my plan is to add more flags in the future...

nameserver/cache.go
inercia: I run `gofmt` periodically, so I guess it is... :-)

nameserver/cache.go
inercia: I didn't know dns.Msg has a setReply method but... isn't it better? It will be easier to remember, I guess...

nameserver/cache.go
inercia: I think you are right: `putTime` is not needed anymore... I'll fix that

But `passedSecs >= TTL` cannot be used as an exit condition as TTL must be the minimum TTL... it is easier (and that was the original intention) to check with `now>=validUntil`

nameserver/cache.go
```diff
@@ -0,0 +1,318 @@
...
+func (c *Cache) Len() int {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return len(c.entries)
+}
+
+// removeOldest removes the oldest item(s) from the cache.
+// note: this method is not thread safe (it is a responsability of the caller function...)
+func (c *Cache) removeOldest(atLeast int, now time.Time) {
```
squaremo: This uses a pass through the entries, then a copy, then a sort. Why don't you sort them first and drop as many as you have to (expired things will be at the back anyway)? Or use a heap, so they are always in order.

rade: http://golang.org/pkg/container/heap/

inercia: (This would depend on Go's implementation of hash tables, but) I think a heap would make the most common operation (Get's) more expensive.

And, if we keep the hash table, the "sort them first" would imply the copy I do as, AFAIK, you cannot sort a hash table in Go. Then you can use the sorted data for either a) deleting entries in the hash table or b) re-inserting in the table...

Anyway, we are doing an early optimization... I would wait until this becomes a bottleneck. My guess is that new entries in the cache will come at the same rate as they expire their TTL, so we will never reach this expensive part of `removeOldest`...

squaremo: Oh you wouldn't *just* use a heap, you'd keep a map of key->heap-entry and the heap would have keys. See http://openmymind.net/Writing-An-LRU-Cache/ for a similar structure.

inercia: Well, it could be done... but I still think this is a bit of overengineering and early optimization... :)

inercia: And, following Karl Seguin instructions, I should use [this](https://github.com/karlseguin/ccache). Absolutely overengineering... ;)

nameserver/server.go
```diff
@@ -6,24 +6,21 @@ import (
...
-	DEFAULT_LOCAL_DOMAIN = "weave.local."     // The default name used for the local domain
-	DEFAULT_SERVER_PORT  = 53                 // The default server port
-	DEFAULT_CLI_CFG_FILE = "/etc/resolv.conf" // default "resolv.conf" file to try to load
-	DEFAULT_UDP_BUFLEN   = 4096               // bigger than the default 512
-	DEFAULT_IFACE_NAME   = "default interface"
+	DEFAULT_LOCAL_DOMAIN   = "weave.local."     // The default name used for the local domain
+	DEFAULT_SERVER_PORT    = 53                 // The default server port
+	DEFAULT_CLI_CFG_FILE   = "/etc/resolv.conf" // default "resolv.conf" file to try to load
+	DEFAULT_UDP_BUFLEN     = 4096               // bigger than the default 512
+	DEFAULT_CACHE_LEN      = 1024               // default cache capacity
```
rade: That seems rather low, given that typical records take up just tens of bytes and I see nothing wrong with the cache taking up a few tens of megabytes. 8192?

rade:
Surely we need to significantly lower the default ttl in dns.go...
````
const (
	localTTL uint32 = 300 // somewhat arbitrary; we don't expect anyone
	// downstream to cache results
)
````
See the discussion in #225. IMO anything higher than 30s is going to be problematic in a fair number of situations. I suggest we make that the default, with a facility to override the value via an option to `weave launch-dns`.

Incidentally, if this PR closes #225, then a separate PR needs to be opened to address the scalability issues identified in #225.

File: nameserver/server.go (hunk 6 above)
Comment (inercia) [inline nameserver/server.go]: I'll fix that.

**Comment by inercia** [comment]
I agree, I will reduce the `localTTL`... Maybe we could also adopt a common practice in CDNs: to return answers with tiny TTLs for some queries, in this particular case, for local queries. Even when the nominal TTL would be used for things like determining the time answers live in the cache, users would be forced to re-query to WeaveDNS each time they want to resolve a name, and that would give us the opportunity to shuffle answers, or load-balance in some more elaborated way. I think that wouldn't be a problem unless a WeaveDNS servers thousands of containers... 

I will open a new PR later on for the scalability issues you described...

**Comment by rade** [comment]
Thinking about it, 30s still seems quite high. 10s?

File: nameserver/cache.go (hunk 1 above)
Comment (squaremo) [inline nameserver/cache.go]: I see you are just as confused by it -- you use dns.Msg#setReply at L93.

File: nameserver/server.go [hunk 7]
```diff
@@ -33,16 +30,47 @@ type DNSServerConfig struct {
 	Port int
 	// (Optional) local domain (ie, "weave.local.")
 	LocalDomain string
+	// (Optional) cache size
+	CacheLen int
+	// (Optional) number of resolution workers for local queries
+	NumLocalWorkers int
```
squaremo: What are these for?

nameserver/cache.go
```diff
@@ -0,0 +1,318 @@
...
+			c.removeOldest(1, now)
+		}
+		c.entries[key] = newCacheEntry(&question, reply, stResolved, now)
+	}
+}
+
+// Look up for a question's reply from the cache.
+// If no reply is stored in the cache, it returns a `nil` reply and no error. The caller can then `Wait()`
+// for another goroutine `Put`ing a reply in the cache.
+func (c *Cache) Get(request *dns.Msg, proto dnsProtocol, now time.Time) (reply *dns.Msg, err error) {
```
squaremo: I'm not clear on why the protocol appear in this signature (and in the cache key). Would you explain ..

squaremo:
I don't see how there will ever be multiple RRs for a local (on the weave network) name: both the zone and the mdns client return just one RR, which the cache will take as a replacement for any existing answer. The only time multiple answers will come through is from a fallback query (arguably we ought not be permuting those).

nameserver/cache.go
inercia: A previous version of the cache did not have this distinction, but I realized that truncated responses would be trickier. 

First, the cached response for TCP and for UDP does not need to be the same and, in fact, they will most probably be different.

Second, _UDP+truncated_ responses would always be followed by another TCP request, and both responses would have to be stored in the same entry in the cache. So a _UDP+truncated_ response would never last long in the cache. Adding the protocol to the key removes this inconsistency and only adds one byte...

nameserver/server.go
inercia: Oh, `NumLocalWorkers` does not really belong to this PR but to the next one... ;)

nameserver/cache.go
squaremo: Why would the UDP+truncated response go in the cache at all?

nameserver/cache.go
inercia: Any suggestion for an alternative name?

nameserver/cache.go
inercia: Because, even if it is truncated, it is as valid as any other entry (eg, it can be requested 3 million times per second, so better you have it in the cache...)

inercia:
I assume this comment is about the shuffle operation. I agree that we are not currently using this feature, but we could return multiple answers in the future (eg, #338). It should have no effect or performance cost when returning only one answer...

squaremo:
> I assume this comment is about the shuffle operation.

In part; but see the second paragraph which I think is a more serious problem, unless I've misunderstood the code.

inercia:
Yes, I thought about that, but we could _a)_ store in the cache the first response we receive (as you said) or _b)_ apply some heuristics and merge all the responses, setting some reasonable values for things like TTLs, etc...

squaremo:
> b) apply some heuristics and merge all the responses,

Something like that would have to be done, yes, but also lookups would have to be tried in parallel (since in general you'd want to return all answers on the network). I think the interaction with the cache would be quite different, and possibly using mDNS wouldn't be workable.

In any case, at the minute the behaviour will be:

 1. Answer with whatever's in the cache, until it expires
 2. Get another answer

whereas if we are sticking to single answers I think we want

 1. Answer with one thing that's in the cache that's not expired
 2. If there are no answers, start looking for some

inercia:
Yes, there could be a periodic process that, for local queries, could try to get more answers with mDNS and merge answers and store them in the cache for future queries...

nameserver/cache.go
inercia: @squaremo I cannot get rid of `putTime`: setting the TTL to `validUntil - now` would set all TTLs to the same value (as `validUntil` is calculated with the minimum TTL in the reply), and that is not accurate...

nameserver/cache.go
inercia: Fixed in #447

nameserver/cache.go
squaremo: YAGNI. You can make it a bitset when that happens.

inercia: That is a procrastination argument :)

In a case like this, I'm fond of flags because _1)_ they look better when calling the function (eg, I think `s.cache.Put(..., CacheLocalReply, ...)` is more intuitive than `s.cache.Put(..., true, ...)`)  _2)_ they do not create functions with a one zillion arguments _3)_ there is no storage/performance gain in using a bool _4)_ and because, in conclusion, there is nothing bad in using a flag and, at the same time, you do not break the API when you add more things... :)

nameserver/cache.go
squaremo: > Second, UDP+truncated responses would always be followed by another TCP request, and both responses would have to be stored in the same entry in the cache.

What's wrong with just overwriting the cache entry when you have a full answer?

If you are proxying UDP and you have a full answer (that won't fit), you can truncate it. This could be the case if you had a full UDP answer but it can't fit it in the client's buffer size.

If you only have the truncated answer and you are asked over TCP, you can go fetch the full answer.

Keeping them as distinct means we can't answer a UDP question with a TCP answer. Is that a problem? Not for correctness -- you can always send the truncated reply then wait for the client to ask over TCP. But it's needless, if you have the answer right there.

Am I missing something?

nameserver/cache.go
squaremo: The constant v boolean arguments argument is a good point (especially if you can make them varargs) -- I hadn't considered that.

nameserver/cache.go
inercia: > If you are proxying UDP and you have a full answer (that won't fit), you can truncate it.

I don't want to do any processing at all at the cache, let alone to be truncating responses by ourselves. Having a cache with, for example, the default 8192 entries is something very cheap (in the worst case we are wasting 8192 bytes). The whole point in having a cache is that you can return a response immediately, so I would not add all that complexity and increase the CPU consumption for saving just a few bytes...

nameserver/cache.go
squaremo: OK, I see. You'd have to go back and figure out passedSecs by finding the minimum TTL and subtracting that from validUntil, to get putTime, ...

While we are here, why is an entry is expired when the *minimum* TTL has passed, rather than when the maximum has passed? This means the whole answer gets thrown away when any RR in it expires. To be honest, I struggle to figure out what is the right thing to do, when it's whole sets of RRs in one cache entry, so I'm curious to hear the rationale.

nameserver/cache.go
squaremo: I'm not suggesting that the cache truncate answers, I'd expect whatever was writing the answer to do that. The cache doesn't know how big the client's buffer is.

It's not about saving bytes, it's about avoiding falling back to TCP if you don't have to -- i.e., returning a response immediately if you can.

nameserver/cache.go
inercia: I have taken the conservative path: an entry is valid until all the RR in the response are valid. I was specially concerned about _additional_ records in the response that could increase the `validUntil` if it was calculated that way.

For example, if asking about `some.domain.com NS` you could get the list of nameservers (`some.domain.com. 21223 IN NS ns1.some.domain.com.`) and, additionally, the list of IPs for those nameservers (`ns1.some.domain.com. 559 IN A 144.168.1.100.`), both with different TTLs, but both the nameservers and the names would have to be up-to-date in order to be useful for the client. Otherwise, the client would bother us again with another query, it would take more space from the cache...

nameserver/cache.go
inercia: I think this will be the typical use case with truncated responses:

_1)_ the client sends a non-local query to WeaveDNS with UDP
_2)_ WeaveDNS sends the same request to the fallback server with UDP
_3)_ the fallback server realizes the response will not fit in a UDP packet, so it fills the packet with as much data as it can but it marks the packet as _truncated_ so the client will know some info is missing
_4)_ the fallback server sends this truncated packet to WeaveDNS
_5)_ WeaveDNS caches the _"UDP+truncated"_ response
_6)_ WeaveDNS sends it back to the client
_7)_ the client sees the _truncated_flags and realizes there is some info missing
_8)_ the client retries the query, but this time with TCP
_9)_ WeaveDNS proxies the query to the fallback server, this time with TCP
_10)_ the fallback server sends all the info to WeaveDNS through the TCP connection
_11)_ WeaveDNS caches the _"TCP"_ response
_12)_ WeaveDNS sends the info to the client
_13)_ the client is done

So we will end up having the "UDP+truncated" response in the cache as well as the "TCP" response. Both responses will be immediately returned to clients the next time someone asks about the same query, so we will skip _2_, _3_, _4_, _5_, _9_, _10_ and _11_.

squaremo: > So we will end up having the "UDP+truncated" response in the cache as well as the "TCP" response. Both responses will be immediately returned to clients the next time someone asks about the same query, so we will skip 2, 3, 4, 5, 9, 10 and 11.

And if you don't distinguish between TCP and UDP responses, you can potentially just skip from 1 to 12 if the client's buffer is large enough, without incurring a cost if it isn't.

Also, if the client's buffer is smaller than the one WeaveDNS uses for upstream queries, WeaveDNS may have to forward a truncated response to the client even if it got a full response from upstream -- as things stand, when the client subsequently connects via TCP, WeaveDNS will have to repeat the query upstream despite having the full answer already.

inercia: I think you are misunderstanding how truncation works: it has nothing to do with buffer sizes. The _TC_ bit is set by the DNS server when the number of responses do not fit the 512 bytes limit of the packet. It has nothing to do with the client buffer size. So the TC bit is like a _"please ask me again, but with TCP"_ flag.  

So you cannot jump from 1 to 12 because _a)_ you must send the _"UDP+truncated"_ response to the client, you cannot skip this step (otherwise, the client will not ask again with TCP) _b)_ _1_ involves an incoming UDP connection and _12_ is an outgoing TCP connection.

So I have distinguishes TCP and UDP responses in the cache because they are different. As I said, _"UDP+truncated"_ responses contain the TC bit set and the RRs the server could put in the message. _"TCP"_ responses are full responses. They are just different.

squaremo: > I think you are misunderstanding how truncation works: it has nothing to do with buffer sizes.

Have I misunderstood EDNS0? https://tools.ietf.org/html/rfc2671

inercia: I admit there is a possible optimization here if we make use of EDNS, but we still need to make distinction between UDP and TCP in the cache...

As you said, when `cache[query+UDP]="UDP-truncated response"`, we could check if a) the client is using EDNS b) `cache[query+TCP]` exists and c) it can fit in the payload for this client. In that case, it could be returned.

But I don't think we should assume clients support EDNS, and I don't think we should use these EDNS extensions for other parts of the system either. I think our DNS client to the fallback server in WeaveDNS is using a user-configurable buffer size (`UDPSize`) and I think this is used for announcing the maximum payload size with EDNS, but what happens if we send requests to `ns1.google.com` announcing, for example, a buffer size of 65535 bytes? Are we going to have a cache filled with 65535-bytes responses? Then maybe no client of WeaveDNS is capable of receiving that, and the whole cache could be useless then... so I think WeaveDNS should not announce any buffer size at all...

squaremo: > I think our DNS client to the fallback server in WeaveDNS is using a user-configurable buffer size (UDPSize) and I think this is used for announcing the maximum payload size with EDNS

It is: https://github.com/zettio/weave/pull/361. It's set to 4KB.

A little wiresharking suggests that libresolv won't use (or advertise) a larger buffer by default. So it's certainly a possibility that we get a non-truncated UDP response from upstream which we can't send downstream without truncating; and that we could answer a subsequent TCP request with the answer we got via UDP (if we didn't segregate answers according to how we obtained them).

> But I don't think we should assume clients support EDNS

Aye, we can only send larger responses if the client has itself advertised a larger buffer. Using EDNS0 is an option in resolv.conf, and "security-aware resolvers" per DNSSEC are required to use EDNS0 OPT RRs and larger buffer sizes (RFC 4035). I don't know how common those are.

> I don't think we should use these EDNS extensions for other parts of the system either.

That ought to be logged as an issue.

> what happens if we send requests to ns1.google.com announcing, for example, a buffer size of 65535 bytes?

WeaveDNS doesn't do that; why would it?


In sum: the current algorithm is

 * For a UDP request:
   * Do we have a cached UDP response? If so send that, truncating if necessary
      * If not, ask upstream, cache the result, and send it downstream, truncating if necessary
   * For a TCP request:
    * Do we have a cached TCP response? If so send that
      * If not, ask upstream over TCP, cache the result and send it downstream

Without segregating cache entries, the algorithm would be

 * For a request:
    * Do we have a cached response? If so, send that, truncating if necessary
      * If not, ask upstream*, cache the result, and send it downstream, truncating if necessary

*There's two tactics wrt using TCP upstream: it could be done in response to the client using TCP, or it could be immediately on getting a truncated result.

inercia: >> what happens if we send requests to ns1.google.com announcing, for example, a buffer size of 65535 bytes?

> WeaveDNS doesn't do that; why would it?

As you said, it is 4K, and it could be ok to do so when no cache was used, but caching that response would be useless for some clients. A client could not accept such a long response, and no cache entry would be useful. Yes, you could truncate a long response in the cache, but I think the point in having a cache is not only to avoid requests to upstream servers but to avoid this kind of computations...

By the way, current WeaveDNS implementation does not truncate responses to clients if they support less than 4K, right? I think the recursive resolver always uses 4Kb, and I don't remember to have seen the part of the code that truncates responses...

> In sum: the current algorithm is
> 
>  * For a UDP request:
>    * Do we have a cached UDP response? If so send that, truncating if necessary
>       * If not, ask upstream, cache the result, and send it downstream, truncating if necessary
>    * For a TCP request:
>     * Do we have a cached TCP response? If so send that
>       * If not, ask upstream over TCP, cache the result and send it downstream

No, my proposal never truncates responses. As I said, I think it is not a good thing to do... The only things I modify in responses is the TTL (and, in some cases, some shuffling)

> * For a request:
>    * Do we have a cached response? If so, send that, truncating if necessary
>      * If not, ask upstream*, cache the result, and send it downstream, truncating if necessary

I think these algorithms could be classified more like:

1) to have full responses in the cache, obtained either with TCP or with UDP, and truncate as needed, according to client needs.
2) to have the short responses, obtained with UDP, some (hopefully, just a few) with the TC bit set and some (hopefully, just a few) full responses (obtained with TCP), and never process the packet.

So your algorithm pros are (and correct me if I'm wrong)

+ it would save one RTT to the upstream server for long responses >512bytes.
+ it would save a byte per cache entry as well as the entries for short, truncated responses when long responses are also present.

but the cons would be

- it would be useless if the upstream server does not support EDNS
- hopefully, it will only make a difference for a tiny fraction of queries (as most queries fit in the original 512-bytes payload limit), and when previous condition is met. 
- it would require implementing the truncation heuristics (eg, you have to decide which RR should be removed...)
- it could lead to many truncations (for example, if the response in the cache for `google.com NS` is 4Kb and all WeaveDNS clients are not-EDNS capable or with short buffers) 

In conclusion, I think your algorithm is nice but it adds more complexity and just saves us some bytes in some corner cases... :|

inercia: Anyway, I think WeaveDNS should calculate and log some statistics. It could give us some information on possible future optimizations...

squaremo: > In conclusion, I would consider your algorithm an optimization more than the base case.

I'm not as concerned with optimisation as I am with simplicity, so I would prefer to say that it's removing complication (e.g., protocol arguments to procedures).

If you include changing the upstream client buffer size back to 512B, then I can see that your scheme works. (By the way you are entirely right that we don't truncate responses -- and the dns library doesn't either so far as I can tell).

Let me try to remove some complication, though: wrt to UDP and TCP, all you really care about is whether you might need to send a truncated result -- with TCP you never need to, and with UDP you need to if the upstream server needed to (assuming both downstream and upstream clients have the same size buffer, i.e., 512B). That's why two different answers might be cached, and the protocol is threaded through methods, to the cache.

Instead of storing two answers, you could look at whether the answer needs to be truncated *when you go to write it out*. You seem to think this is a major bit of computation, but it's really not: you just need to look at the length. And to truncate it, you can just set the bit and remove all RRs (or just don't copy them over from the cached result!). None of this needs to be done inside the cache code.

(It's not a problem to send an empty answer in a truncated response: that's what djbdns does, and miekg/dns doesn't even bother parsing the RRs of packets with the truncate bit set.)

inercia: I understand, and this proposal looks much better to me. To create and send a dummy, truncated response is not a big deal. And checking a response length is probably not too difficult...

But I still have a concern. I don't know if truncating responses in such a simple way is a wise solution: maybe we are forcing the client to reconnect with TCP when it would be more than happy with some answers. I want to think that a server like BIND implements some smart heuristics for giving the most valuable info in a truncated response, but maybe I'm assuming too much... And I don't know what is the common behavior in DNS clients... probably they will always try to get a full response...

So I'll implement the change you suggest. Last chance if you have a concern... :)

inercia:
@squaremo I have removed the distinction between protocols in the cache in [last commit](https://github.com/inercia/weave/commit/aefc926ef73547e2baad92485681718b4d64aca6) as you suggested...

inercia:
By the way, it seems that setting `UDPSize` to a bigger values does not result in that value being announced to the server with _EDNS0_. Unless you use `(dns *Msg) SetEdns0(udpsize uint16, do bool)` the server will not detect you are capable of using bigger payloads...

squaremo:
> By the way, it seems that setting UDPSize to a bigger values does not result in that value being announced to the server with EDNS0.

You are right -- for some reason (possibly because it *used* to be this way, but I've not verified that) we assumed miekg/dns added EDNS0 when you increased the UDP buffer size.

squaremo:
> for some reason [...] we assumed miekg/dns added EDNS0

@bboreham corrected me -- it's the other way around, it'll increase the buffer if you set EDNS0.

squaremo:
@inercia I have done a couple of cleanups in squaremo/weave branch pr429 (https://github.com/squaremo/weave/commits/pr429), can you take a look? If you think they are worthwhile I'll include them when I merge.

squaremo:
@rade points out that this doesn't fix what's described in #225 -- I've changed the name of that, since it was far too vague. To do that the cache for *local* answers would have to keep (and expire) individual RRs, and assemble replies based on those that are current.

The cache here is OK for external answers (although you could argue that it should also keep and expire individual RRs, but it is fit for purpose as it is)

squaremo:
I should be clear that I don't think that prevents this being merged, since it represents progress -- just that we can't close #225.

rade:
Well, on reflection, I actually think this *does* address #225. In particular it takes care of three points at the top. What comes after that is arguably an implementation detail.

However, this PR also does *more* than #225. In particular it both caches and shuffles results of external queries. #225 makes no mention of that. And for good reason...

- External DNS queries do not result in a broadcast and are therefore not placing anywhere near as much load on the network and machines.
- If performance of external DNS queries is an issue, a user can always insert their own caching DNS server.
- Shuffling local results is ok, because we control the semantics of dns resolution in the local domain. But we shouldn't mess with results from external queries. Do any of the well-known DNS caches shuffle results?

inercia:
@rade You are right when you say that this does a bit more than #225. It didn't make sense to cache local queries and just forget about recursive resolutions, specially when it was so easy to do it once the local queries problem was solved. Users could add their own solutions, but that would add a lot of unnecessary complexity and I think this would be a valuable feature...

But you are wrong when you say that this PR shuffles recursive results: [we check where we obtained the result](https://github.com/inercia/weave/blob/weave-225/nameserver/cache.go#L117) and use that information for deciding if we must shuffle or not... Doing that would be absolutely wrong and we would burn in hell.

@squaremo When you say that the cache should _keep (and expire) individual RRs, and assemble replies based on those that are current_, I think that goes beyond of simple caching. WeaveDNS receives responses with only one RR when it queries with mDNS, so I assumed we could skip that granularity level (_individual RRs_) and just work at the _reply level_ in the cache. We could implement more elaborated solutions for improving the cache contents (discussed at the end of #225, like [this](https://github.com/zettio/weave/issues/225#issuecomment-77922996)), but they will be in the _"cache 2.0"_. Maybe #225 was too generic and we should have several sub-tasks...

rade:
So, as was pointed out earlier, is the shuffling actually dead code at the moment? We don't shuffle results of external queries, and local queries only ever produce one result, so there is nothing to shuffle.

>  WeaveDNS receives responses with only one RR when it queries with mDNS

But

- mDNS queries can result in responses from multiple servers, and at some point - certainly when tackling #226 - we are going to have to do something smarter than just stomp on existing answers with new ones.

- eventually the responses from even just a single WeaveDNS server may contain multiple results. As per #338.

> so I assumed we could skip that granularity level (individual RRs) and just work at the reply level in the cache

Even right now, external queries can contain multiple RRs.

The granularity of facts known to a DNS server are RRs. And TTLs are on RRs. And the same RR can be part of the result to a variety of queries. Therefore the cache should operate at the granularity of RRs. Answers are an ephemeral construct; I'd consider caching them in a secondary cache, at best.

The purpose of #225 is to reduce the network/CPU load and latency resulting from issuing mDNS queries for every single inbound local query. *That* is where the bottleneck is. Not in external queries. And not in assembling answers from facts.

inercia:
> So, as was pointed out earlier, is the shuffling actually dead code at the moment? We don't shuffle results of external queries, and local queries only ever produce one result, so there is nothing to shuffle.

Yes, it is dead code right now. It could be safely removed.

> The granularity of facts known to a DNS server are RRs. And TTLs are on RRs. And the same RR can be part of the result to a variety of queries. Therefore the cache should operate at the granularity of RRs. Answers are an ephemeral construct; I'd consider caching them in a secondary cache, at best.

Maybe that could be true for local queries, but I'm not sure if that is true for external queries. A DNS server is not just a key-value database, so I don't know if you can think of RRs as isolated facts. When you query about `somecontent.akamai.com` you get a particular set of RRs, with a particular order and a very particular group of TTLs. So when caching `somecontent.akamai.com`, can we safely decompose the response in RRs and store them in a key-value cache? If the RRs show a transitive resolution (eg, `somecontent.akamai.com` `->` some CNAME `->` some IP), can we safely store those individual RRs in the cache, or they are only valid for us, for `somecontent.akamai.com` and now?  

Besides that, I think the idea of a cache is to avoid that _answer assembly_ you mentioned: a cache should be something where you look for a key and you just get the answer. Facts could be stored in a regular database, with more appropriate data structures for assembling answers on demand and store them in the cache. I don't think caching RRs is the solution as not only answers are ephemeral: RRs are ephemeral too.

So I think we are talking about the same here but there is some confusion about what is what. The cache in this PR is the _secondary cache_ you mentioned, and the _facts database_ would be the [zone database](https://github.com/inercia/weave/blob/weave-338/nameserver/zone.go). That zone database should be the place where you store higher level facts, the source and destination for gossiping information, accessible through an API, etc... And the cache should just store _ready to send_ answers.

squaremo:
> WeaveDNS receives responses with only one RR when it queries with mDNS, so I assumed we could skip that granularity level (individual RRs) and just work at the reply level in the cache.

For now, yes for sure. I mean once there is a way to have more than one result, either kept in the zone db, sent by the mDNS peer, or accumulated, then it makes sense to treat RRs individually. This would be different in nature to caching the proxied results.

inercia:
Yes, of course, @squaremo, but I'm reluctant to use the term RR for this information because it reminds me of the wire, binary data. I would prefer to keep that information (those _facts_) in the zone database in a more high level form...

So, in my view, we should have

* the zone database, the source of information for queries, with a richer internal structure (#364, #338) with its own information retrieval mechanisms (eg, mDNS, gossiping, etc), a rich HTTP API (#432), etc
* a responses cache, for local and recursive _ready-to-send_ responses, offering an invalidation mechanims to the zone database

I think the next steps would be

1. to implement the invalidation mechanism in the cache (currently working on it, see [this](https://github.com/zettio/weave/pull/447#issuecomment-78267078))
2. to enrich the zone database, so we can store multiple IPs per name and (potentially) other information (eg, SRV information, etc) (started work in [this branch](https://github.com/inercia/weave/blob/weave-338/))
3. to improve the HTTP API
4. to implement the gossiping mechanims described [here](https://github.com/zettio/weave/issues/225#issuecomment-77922996)

What do you guys think?

squaremo:
> So, in my view, we should have
> 
>     the zone database, the source of information for queries, with a richer internal structure (#364, #338) with its own information retrieval mechanisms (eg, mDNS, gossiping, etc), a rich HTTP API (#432), etc

Sounds good.

>     a responses cache, for local and recursive ready-to-send responses, offering an invalidation mechanims to the zone database

I'm not sure this is needed. If TTLs for local results are small, which they should be, you'll be invalidating the cache all the time. In which case there's little benefit to caching whole result sets over just assembling the matching records whenever you're asked. If the zone DB has O(low) selection then there won't be much difference, if any, to what's done to construct results now.

At the least it's worth doing some calculations to demonstrate whether there's any benefit or not.

> I think the next steps would be
> 
>     to implement the invalidation mechanism in the cache (currently working on it, will be PR "[dns] caching (4/4)", see this)
>     to enrich the zone database, so we can store multiple IPs per name and (potentially) other information (eg, SRV information, etc) (started work in this branch). It should also provide a better API (more than AddRecord/DelRecord, eg GetRecord, GetRecordsFor*, etc) that could be used by an improved HTTP API
>     to improve the HTTP API (#432)
>     to implement the gossiping mechanims described here

I'd really like to see multiple local results supported, however it's done.

inercia:
> I'm not sure this is needed. If TTLs for local results are small, which they should be, you'll be invalidating the cache all the time. In which case there's little benefit to caching whole result sets over just assembling the matching records whenever you're asked. If the zone DB has O(low) selection then there won't be much difference, if any, to what's done to construct results now.

Aha, @squaremo, but, [quoting](https://github.com/zettio/weave/issues/225#issuecomment-77922996) @rade, _"the whole point of doing something more complicated is to be able to use a longer TTL"_. The cache invalidation part is needed in order to implement the gossiping mechanism, so we could use very long TTLs (or we could even forget about TTLs).

squaremo:
If you mean that you will always hear "quickly" about arriving or leaving containers, and when you hear about them, you'll change the set of records you give in response to queries .. what is the difference between this and an enriched zone DB? I don't think you need both.

inercia:
Do you mean we do not need a cache and a enriched zone db? Well, having both has some advantages.

First, I think it provides a better evolution path for WeaveDNS. Once we have a functional cache, DNS resolutions will be greatly improved, both for local and for recursive resolutions. So we will not need to do a complete refactoring of WeaveDNS and base everything in gossiping. And then any refinement in the mechanism we use for updating the zone db will not be incompatible with having a cache in front. In fact, most of the queries should be answered from the cache, and that will probably make a big difference...

I think that the complexity will not be in this cache: the complexity will really be in the gossiping mechanism...

squaremo:
> think that the complexity will not be in this cache: the complexity will really be in the gossiping mechanism...

The complexity will be in the gossiping, in the zone DB, in the cache, and in the interactions among them and with other code. Please can we solve one thing at a time, until it's demonstrated that e.g., caching local query responses -- as well as having the records in the zone DB -- is necessary.

rade:
> Please can we solve one thing at a time, until it's demonstrated that [...] is necessary.

IMO that also applies to caching results of recursive queries.

rade:
Do we know how well-known caching DNS servers, such as dnsmasq, perform caching? I am concerned that a cache keyed on the octet-sequence of a question, and storing entire answer sets, might be semantically suspect.

inercia:
@rade I think solutions applied by others do not need to be valid or applicable for us, as we do not have the resources others have (otherwise, we could reimplement BIND9 for Weave). Using the question as the key should be perfectly valid. A better granularity could lead to better cache usage, but that would add more complexity and I don't think we really need to optimize space here...   

Regarding recursive lookups, I do not understand why you are so reluctant to use the cache for them. I wouldn't underestimate these resolutions: they could have a big impact for users in some scenarios (e.g., a user setting `HostnameLookups On` in `apache.conf`). Caching these resolutions comes for free and, in fact, using the same mechanism for both local and recursive lookups simplifies the code a lot, so it is a win-win for me...

squaremo:
> I think solutions applied by others do not need to be valid or applicable for us, as we do not have the resources others have

The suggestion is to look for indications of what is the correct way to handle proxied responses -- whether it's OK to pass every record on. djbdns, for example, appears to do a lot of work to sanitise responses. If we aren't prepared to figure this out we have no business providing a caching name server.

> Caching these resolutions comes for free

Only if you think development and maintenance effort on our part, and troubleshooting and operation on our users' part, are free.

inercia:
@squaremo In conclusion, do you want to remove the cache for recursive queries?

squaremo:
> In conclusion, do you want to remove the cache for recursive queries?

No, I just want us to be clear where we are providing benefit, and what the trade-off is. For example, many deployments will already have a caching nameserver, or be happily using 8.8.8.8, and in those deployments we are effectively adding needless machinery; some won't, but solutions exist already -- use dnsmasq or whatever -- and are we sure ours is really better?

I know you've done some benchmarking, so perhaps you can make a quantitative argument in this direction.

rade:
There are a bunch of RFCs pertaining to caching...
- [RFC1034](http://www.rfc-archive.org/getrfc.php?rfc=1034) (NB: results of inverse queries must not be cached!)
- [RFC1035](http://www.rfc-archive.org/getrfc.php?rfc=1035)
- [RFC2181](http://www.rfc-archive.org/getrfc.php?rfc=2181)
- [RFC2308](http://www.rfc-archive.org/getrfc.php?rfc=2308) (Caching negative results is *hard*)

The upshot is that there is *a lot* to consider when caching answers from recursive queries. Most of these do not apply for local queries because, a) as Michael said, we really should consider the collection of all weaveDNS servers as a single server, and b) our servers only deal with a very limited number of queries / record types.

inercia:
Ok, no problem, I'll remove caching for recursive resolutions.

nameserver/server.go
```diff
@@ -144,24 +194,47 @@ func (s *DNSServer) Start() error {
...
 	return nil
 }
 
-func (s *DNSServer) queryHandler(lookups []Lookup) dns.HandlerFunc {
+func (s *DNSServer) queryHandler(lookups []Lookup, proto dnsProtocol) dns.HandlerFunc {
 	return func(w dns.ResponseWriter, r *dns.Msg) {
 		q := r.Question[0]
+		maxLen := getMaxReplyLen(r, proto)
 		Debug.Printf("Query: %+v", q)
 		if q.Qtype == dns.TypeA {
```
awh: If this type is no matched, should we be returning 'Not Implemented' rather than 'Name Error'?

nameserver/server.go
```diff
@@ -173,15 +246,33 @@ func (s *DNSServer) queryHandler(lookups []Lookup) dns.HandlerFunc {
...
 
-func (s *DNSServer) rdnsHandler(lookups []Lookup, client *dns.Client) dns.HandlerFunc {
-	fallback := s.notUsHandler(client)
+func (s *DNSServer) rdnsHandler(lookups []Lookup, proto dnsProtocol) dns.HandlerFunc {
+	fallback := s.notUsHandler(proto)
 	return func(w dns.ResponseWriter, r *dns.Msg) {
 		q := r.Question[0]
+		maxLen := getMaxReplyLen(r, proto)
 		Debug.Printf("Reverse query: %+v", q)
 		if q.Qtype == dns.TypePTR {
```
awh: If this type is not matched, we return nothing? Refused or Not Implemented might be better?

nameserver/server.go
```diff
@@ -89,19 +131,28 @@ func NewDNSServer(config DNSServerConfig, zone Zone, iface *net.Interface) (s *D
...
 
 	// create two DNS request multiplexerers, depending on the protocol used by clients
 	// (we use the same protocol for asking upstream servers)
-	mux := func(client *dns.Client) *dns.ServeMux {
+	mux := func(proto dnsProtocol) *dns.ServeMux {
 		m := dns.NewServeMux()
-		m.HandleFunc(s.Domain, s.queryHandler([]Lookup{s.zone, s.mdnsCli}))
-		m.HandleFunc(RDNS_DOMAIN, s.rdnsHandler([]Lookup{s.zone, s.mdnsCli}, client))
-		m.HandleFunc(".", s.notUsHandler(client))
+		m.HandleFunc(s.Domain, s.queryHandler([]Lookup{s.zone, s.mdnsCli}, proto))
```
awh: I wonder if the cache and the fallback resolver could be implemented as components that support the `Lookup` interface? Then you might be able to have a single handler that loops over Cache -> Zone -> mDNS -> Fallback ...

inercia: It is a good idea. Maybe it could add some complexity to the `Put()` when you get a valid result with the `Lookup()` from the cache, but I could investigate this...

nameserver/server.go
inercia: I agree. I will return a more specific error...

inercia: @awh done. We return a "Not implemented" error now...

inercia:
@rade and @squaremo , I have updated this PR with the following changes:

* the cache for recursive queries has been disabled 
* I have squashed all the commits in the branch (rebasing master was being really annoying...)
* I have merged the changes in the PR 3/3 (#447) (to use a heap for optimizing garbage collections)

Let me know what you think about this...

squaremo:
What is the rationale behind keeping the heap of entries in order of expiry? I would have thought either of insertion time (nice and easy) or LRU (more trouble but more useful perhaps) would be preferable.

inercia:
@squaremo insertion time wouldn't give you the best order for getting rid of entries (could lead to useful entries being evicted before useless entries), and I'm not sure about the LRU (it would require more logic, and maybe it could make `Get()`s more expensive...). I think the heap for `validUntil` is the _nice and easy_ solution...

squaremo:
> insertion time wouldn't give you the best order for getting rid of entries (could lead to useful entries being evicted before useless entries)

It depends what you think is useful or useless. An entry with a long TTL that is asked for very infrequently could be considered useless, while an fast-moving entry that gets asked for continually could be considered useful. Ordering the evictions by how soon they expire will tend to preserve the former while evicting the latter, sometimes ahead of schedule.

Ordering by time of insertion means that you will tend to refresh (reinsert because of expiry) fast-moving entries well before they can get evicted, while slow-moving entries might make it to the top of the heap before being refreshed, but if so weren't being asked for very often anyway.

PR #435, opened 2015-03-04
434 Update README with CoreOS install instructions

fintanr:


README.md
```diff
@@ -34,6 +34,9 @@ Ensure you are running Linux (kernel 3.5 or later) and have Docker
       https://github.com/zettio/weave/releases/download/latest_release/weave
     sudo chmod a+x /usr/local/bin/weave
 
+If you are using CoreOS you will need to install weave using cloud-config.
+An example cloud-config file is provided [here](https://github.com/fintanr/weave-gs/blob/master/coreos-simple/user-data).
+
```
rade: The issue I have with this is that people who are not familiar with CoreOS or weave will immediately think that a) installing weave on CoreOS is quite complicated (which is true, compared to typing a single command), and b) that this is somehow weave's fault (which it isn't).

squaremo: I don't think it especially suggests b). In any case, how about just

> CoreOS users see _here_ for an example of installing weave using cloud-config.

rade: +1

PR #440, opened 2015-03-06
weave-225: [dns] caching (2/3)

inercia:
#225: this pull request introduces a limited number of concurrent resolvers in the DNS server. Clients requests are distributed among these resolvers with no duplicates (ie, there will be only one resolution for a given name, even when multiple clients are performing the same query). 

Performance tests with DNSblast show a significant increment in the number of requests that can be served per second as well as a reduction in the number of I/O timeouts to upstream servers.

Important: this PR depends on #429


rade:
> Performance tests with DNSblast

I'm intrigued. Would you mind elaborating what/how you tested?

inercia:
I tested performance of upstream queries, mostly for checking how the cache works. Starting WeaveDNS with

```
sudo weavedns -dnsport=1053 -fallback="8.8.8.8:53" -debug
```

and launching [dnsblast](https://github.com/jedisct1/dnsblast) with something like

```
dnsblast 127.0.0.1 50000 100 1053
```

This would generate 100 reqs/sec of DNS A-queries for names with the form `<random char><random char><random char><random char>.com`. Invoking _dnsblast_ again with the same arguments generates the same sequence of A-queries, a nice feature for testing the cache, even when queries following a Pareto distribution would be much more realistic...

inercia:
My guess in the reduction of I/O timeouts to fallback servers is that the new mechanism reduces the number of concurrent queries to Google's servers, while the old mechanism generates too many requests and they probably drop packets from the same IP when they come too fast...

inercia:
Closed until further notice...

PR #447, opened 2015-03-10
weave-225: [dns] caching (3/3)

inercia:
#225 : this pull request implements an auxiliary heap in the cache that keeps entries sorted by their TTL, so the removal of the entry with the lowest TTL is O(log n). 

This PR depends on #429 and #440


inercia:
There is a pending issue in the cache: entries invalidation. In particular, I'm thinking about the relation between the zones database (`zone.go`) and the cache (`cache.go`). It would be nice if we could invalidate all cache entries that refer to a particular record in the zones database when that record changes or is removed...

I'm still thinking on how to solve this, but maybe we could keep a reference in the cache to the record(s) that produced the `response` we insert with `cache.Put(question, response)`, something like:

```
response, dependencies = lookup.LookupInaddr(question)
cache.Put(question, response, dependencies)
```

However, this would not help unless we get all the entities this response depends on, instead of the records. For example `dependencies = []dependency{an_ident, a_name}`. This would help us when removing cache entries when, for example, a container disappears... but it would lead to the cache holding many hash tables (ie, `entriesByIdent`, `entriesByName`, etc) for this kind of lookups...

So, any suggestion?

inercia:
All the heap stuff has been integrated in the first pull request...

Issue #445, opened 2015-03-09
Very slow TCP writes can hang a Weave process


bboreham:
As seen in this stack trace, if the TCP write gets stuck for some reason, LocalPeer will make no progress, so no connections can be added, removed, the user cannot obtain status, etc.

I think it makes sense to set a deadline on writing similar to what we have for reading.

````
goroutine 6 [IO wait, 5 minutes]:
net.(*pollDesc).Wait(0xc20a990d10, 0x77, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:84 +0x47
net.(*pollDesc).WaitWrite(0xc20a990d10, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:93 +0x43
net.(*netFD).Write(0xc20a990cb0, 0xc20b2c6006, 0x3de, 0x7ffa, 0x117, 0x7fe1432f9730, 0xc20ba0bf50)
	/usr/local/go/src/net/fd_unix.go:335 +0x5ee
net.(*conn).Write(0xc209f0f238, 0xc20b2c6006, 0x3de, 0x7ffa, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:129 +0xdc
encoding/gob.(*Encoder).writeMessage(0xc20ad565a0, 0x7fe1418b74d0, 0xc209f0f238, 0xc20ad565d8)
	/usr/local/go/src/encoding/gob/encoder.go:75 +0x238
encoding/gob.(*Encoder).EncodeValue(0xc20ad565a0, 0x862820, 0xc20aba66c0, 0x57, 0x0, 0x0)
	/usr/local/go/src/encoding/gob/encoder.go:243 +0x69b
encoding/gob.(*Encoder).Encode(0xc20ad565a0, 0x862820, 0xc20aba66c0, 0x0, 0x0)
	/usr/local/go/src/encoding/gob/encoder.go:168 +0x7f
github.com/zettio/weave/router.(*SimpleTCPSender).Send(0xc209f0a9f0, 0xc20be0f000, 0x3d6, 0x400, 0x0, 0x0)
	/github.com/zettio/weave/router/crypto.go:470 +0x89
github.com/zettio/weave/router.(*LocalConnection).sendProtocolMsg(0xc2094ec780, 0x5, 0xc20be0c900, 0x3d5, 0x443, 0x0, 0x0)
	/github.com/zettio/weave/router/connection.go:418 +0x172
[...]

PR #460, opened 2015-03-16
Decouple gossip update sending from receiving, to break deadlock

bboreham:
In order to break the deadlock reported at #445, we create a set of pending updates and wait on a separate goroutine until we are able to send.
A gossipUpdateSender and associated goroutine are created for every outgoing connection that has gossip.
This change makes explicit the fact that Gossip/OnGossip were encoding some sort of data structure where updates converged - this data structure is now represented by GossipData and GossipKey, which must be implemented by the parameters passed to NewGossip().


rade:
Supplanted by #476.

PR #463, opened 2015-03-17
Don't force -d on docker run

andyrj:
This makes it impossible for CoreOS/SystemD users to run containers as a service that utilize Weave with Restart=always and other flags and makes it where you have to use Type=oneshot and RemainAfterExit=yes, in your SystemD/Fleet unit.  This seems like a strange assumption to make on behalf of the user.

### DIFF
--- a/weave
+++ b/weave
@@ -610,7 +610,7 @@ case "$COMMAND" in
         validate_cidr $CIDR
         shift 1
         create_bridge
-        CONTAINER=$(docker run $DNS_ARG $DNS_SEARCH_ARG -d "$@")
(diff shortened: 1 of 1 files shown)


vmatekole:
Hi! This code doesn't seemed to have been merged and is still and issue in current version (1.6.1)
I managed to solve #2514 by setting Type=oneshot and RemainAfterExit=yes

bboreham:
@vmatekole this PR was closed and marked 'invalid' because it cannot work.  We need to run the `attach` command after `docker run` while the container is still running, and removing `-d` will make Docker wait until the container exits.

I haven't managed to look at #2514 yet but you'd be better off suggesting your solution there (or in a new PR).  Thanks!

PR #468, opened 2015-03-18
Replace hand parsing and validation of URLs with Gorilla mux

awh:
[Gorilla mux](http://www.gorillatoolkit.org/pkg/mux) is an enhanced HTTP router for golang's http server, taking care of many routine tasks around HTTP method matching, URL parsing and dispatch. Mux is used by Docker's REST API implementation.

### DIFF
--- a/nameserver/http.go
+++ b/nameserver/http.go
@@ -25,13 +25,13 @@ func ListenHttp(domain string, db Zone, port int) {
 		io.WriteString(w, "ok")
 	})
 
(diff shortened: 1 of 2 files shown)


awh:
I have run unit tests and manually tested all URLs but not run smoke tests - running them now!

awh:
All smoke tests pass.

nameserver/http.go
```diff
@@ -9,18 +8,9 @@ import (
 	"log"
 	"net"
 	"net/http"
-	"strings"
+	"github.com/gorilla/mux"
```
bboreham: Please run 'go fmt' on code before committing (which would move this line earlier, in alphabetical order).

PR #474, opened 2015-03-23
Increment the server port for each test

squaremo:
Occasionally the test servers overlap and the test will fail with "address already in use".


dpw:
This is very welcome.

But the underscore in `setup_` seems a bit un-goish to me.  Is there some precedent for it?

squaremo:
> But the underscore in setup_ seems a bit un-goish to me.

Just a hack because there's already a setup() in that package. I'll change it to ... `setupForTest()`?

dpw:
LGTM (Lets Go To Mars).

PR #482, opened 2015-03-25
simplify actors

### LINKED ISSUE #475: simplify actor code
The way we implement actors is a bit cumbersome, with quite a few pieces that need to be carefully aligned:

1. 'client' methods
2. message types
3. server method handlers

It has been proposed that we collapse these into one by communicating closures. Then there will be just one message type, and the server methods will be closures that make up the bulk of the body of the client methods.

rade:
Instead of actor client methods passing data to the server via the actor's channel, and the server in turn invoking server methods with that data, we get the client methods to pass a closure of the server method via the actor's channel. This reduces boilerplate a fair bit.

Closes #475.


rade:
@awh review, please.

Issue #486, opened 2015-03-25
Administrative function to remove peer


awh:
It's not currently possible to instruct weave to forget about a peer that was specified via `weave launch` or `weave connect`. This causes problems for users with dynamic environments where peers can go away forever; their only recourse currently is to restart the weave network leaving out the decommissioned peer.

Add a script command that instructs the router to forget a specified peer:

`weave forget <CIDR>`

'Forget' is suggested for the command name as 'disconnect' implies simply closing a connection; forget removes the peer from weave's internal list of known seeds. Any advance on a better name @rade?

awh:
Invite @fermayo

bboreham:
The original motivation for this feature was to be used on one peer to forget about one address previously specified _to that peer_. But what if it's inconvenient to remember which peer was told about which other peer?  Maybe 'forget' should work from any peer in the network. 

This would require a way to uniquely identify a peer; IP addresses do not do this in the presence of NAT. Perhaps nicknames could be used?

A 'distributed forget' as I have described might have complex semantics under network partition.

awh:
I think in the long term that could be useful - perhaps we could split it off as a separate ticket? I'm keen to get an MVP of this out in the next stable release if possible...

rade:
'connect' adds to the "list of IP addresses *this* peer should always attempt to connect to". It makes sense to have an inverse operation for that. Let's not boil the ocean here.

kulki:
How is this feature supposed to work? Suppose if I have 10 nodes in my weave network and I want to remove one from the network, do I need to run 'weave forget <ip-of-node-to-be-decommissioned>' from each of the other 9 nodes? 
It would be helpful, as mentioned by @bboreham, if i could run 'weave forget <peer>' from only one of the other nodes.

Following is the experiment I did. Note: Auto discovery is ON (default)
I started 'weave launch' from 3 machines m1, m2 & m3. 
On m1: I ran 'weave connect m2' and 'weave connect m3'. Now all were part of the network. 
Then I ran 'weave reset' on m2 followed by 'weave forget m2' on m1. 
Then on m2: I ran 'weave launch'. Immediately m2 joined the network automatically as if 'forget' had no effect. 

In the second run, I 'weave reset' on m2 and 'weave forget m2' on both m1 and m3 and then I ran 'weave launch' on m2. This time m2 did not join the network. I got what I desired. But this is too tedious if have more peers.

In the first case I am wondering if what I think below is happening. 
[...]

PR #488, opened 2015-03-26
Implement weave forget command

awh:
Addresses #486.

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -55,7 +55,7 @@ func (cm *ConnectionMaker) InitiateConnection(address string) {
 func (cm *ConnectionMaker) ForgetConnection(address string) {
 	cm.actionChan <- func() bool {
 		delete(cm.cmdLineAddress, NormalisePeerAddr(address))
-		return true
+		return false
 	}
 }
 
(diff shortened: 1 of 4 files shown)


router/connection_maker.go
```diff
@@ -52,6 +52,13 @@ func (cm *ConnectionMaker) InitiateConnection(address string) {
 	}
 }
 
+func (cm *ConnectionMaker) ForgetConnection(address string) {
+	cm.actionChan <- func() bool {
+		delete(cm.cmdLineAddress, NormalisePeerAddr(address))
+		return true
```
rade: would a 'false' suffice here?

awh: Yes I think you're right - the extant target for this address will get deleted either when the timer fires or ConnectionTerminated gets called. Will push an update.

site/features.md
```diff
@@ -331,7 +331,15 @@ then on one of the existing hosts runs
...
-establish connections to the new host too.
+establish connections to the new host too. Conversely, it is possible
+to instruct a peer to forget a particular host that was specified
+to it via `weave launch` or `weave connect`:
+
+    host# weave forget $DECOMMISSIONED_HOST
+
+This will prevent the peer from trying to reconnect to that host once
+connectivity to it is lost, and thus can be used to administratively
+remove peers from the network.
```
rade: "administratively remove *decommissioned* peers"? Just to avoid any suggestion that the command itself actually removes/disconnects a peer.

PR #491, opened 2015-03-26
Add lint script, which can also act as an optional presubmit hook.

tomwilkie:


bin/lint
```diff
@@ -0,0 +1,86 @@
...
+		;;
+	esac
+}
+
+if [ $# -gt 0 ]; then
+	for filename in $@; do
+		lint ${filename}
+	done
+else
+	for filename in $(git diff --cached --name-only); do
```
awh: Should this be `$(git diff --cached --name-only | grep '.go$')`? Otherwise you're trying to lint arbitrary file types that may have been staged for commit...

tomwilkie:
re: `gofmt`; done.  Don't need to pipe to `head` with `-l`, I only ever pass it one file at a time.

re: `$lintoutput` - this doesn't mangle whitespace if you `echo` with quotes.

PTAL

bin/lint
```diff
@@ -0,0 +1,81 @@
...
+function lint_go {
+	filename="$1"
+	result=0
+
+	if [ -n "$(gofmt -l "${filename}")" ]; then
+		result=1
+		echo "${filename}: run gofmt -w ${filename}!"
+	fi
+
+	go tool vet ${filename} || result=$?
```
rade: I'm guessing that last `${filename}` should be quoted too, like elsewhere.

PR #496, opened 2015-03-27
Router cleanups

dpw:
Matthias seems to be in the mood for cleaning up the core router code.  While I enjoy nothing more than resolving the many rebase conflicts this creates for me, I feel I should push some obvious cleanups before Matthias gets to them.

More seriously: I haven't pushed these before because I'm reluctant to debate whether they are worthwhile when the best argument is simply to present them along with the more compelling work they lead to.  But on the other hand, I don't want to rebase them more often than necessary.  I think they look fairly obvious, but who knows.


rade:
Nice!

Issue #467, opened 2015-03-17
enable starting of app containers with multiple weave CIDRs


rade:
Prompted by a [discussion on IRC](https://botbot.me/freenode/weavenetwork/2015-03-17/?msg=34404225&page=2)...

Right now, if you want a container to have multiple weave CIDRs, you need to run `weave run` followed by one or more `weave attach`. But that means any app in the container may see only a subset of the addresses on startup.

The distinction will become more pertinent once #47 and #230 are resolved.

Technically there is nothing preventing us from accepting multiple CIDRs in `weave run` and get them all configured before marking the interface as 'up'. It only requires some hideous bash scripting :)

While we are at it, we might as well make *all* CIDR-accepting weave commands take multiple CIDRs.

PR #498, opened 2015-03-30
Allow multiple CIDR arguments to run/start/attach/detach

awh:
Addresses #467.


rade:
docs need updating

weave
```diff
@@ -641,16 +653,20 @@ case "$COMMAND" in
...
+            CIDRS="$CIDRS $1"
+            shift 1
+        done
         CONTAINER=$(docker run $DNS_ARGS -d "$@")
         if [ "$LOCAL" = "--local" ]; then
             create_bridge
-            with_container_netns $CONTAINER attach $CIDR >/dev/null
-            tell_dns PUT $CONTAINER $CIDR
+            with_container_netns $CONTAINER attach $CIDRS >/dev/null
+            for CIDR in $CIDRS; do
```
awh: I will refactor `tell_dns` to take multiple CIDRs and hence eliminate the loop here and elsewhere...

weave
```diff
@@ -641,16 +653,20 @@ case "$COMMAND" in
...
             [ "$1" = "--with-dns" ] && shift 1
         fi
-        CIDR=$1
-        shift 1
-        validate_cidr $CIDR
+        CIDRS=""
+        while is_cidr "$1"; do
+            CIDRS="$CIDRS $1"
+            shift 1
+        done
```
awh: This `while` over `is_cidr` appears in four places, but I'm struggling to see how to factor it out as it modifies `$@` with `shift`. Any thoughts @dpw?

dpw: shift $(collect_cidr_args "$@")

`collect_cidr_args` counts the CIDR parameters and sets `CIDRS` as a side effect (Matthias will love that).

awh: Just spent twenty minutes banging my head against this, wondering why it wasn't working - unfortunately the `$()` expression executes in a subshell, so the modified global `$CIDRS` is not visible to the calling process...

awh:
Updated documentation and factored out some common parts. Unfortunately the CIDR collection loop is still resisting (see my comment elsewhere) - remaining avenues are the CPS transform that @dpw mentioned or a function that exports *two* global variables. Thoughts @rade?

site/features.md
```diff
@@ -90,6 +90,14 @@ each other but not the containers of our first application...
 This isolation-through-subnets scheme is an example of carrying over a
 well-known technique from the 'on metal' days to containers.
 
+If desired, a container can be attached to multiple subnets when it is
+started:
+
+    host1# weave run 10.2.2.1/24 10.2.3.1/24 -t -i ubuntu
+
+In this case, all of the addresses are added to the container's
+weave interface before it is brought up.
```
rade: Drop that last sentence; it's an implementation detail. And also wrong ;) (the interface is called 'ethwe').

rade:
Two globals is fine. No need for the 'GLOBAL_' prefix; all our vars are global.

awh:
Docs updated, CIDR parsing refactored.

rade:
Any reason not do do 'expose' and 'hide' too?

awh:
`weave expose` and `weave hide` now accept multiple addresses.

awh:
Docs updated to use consistent subnets in examples.

awh:
I'm going to have a stab at rebasing this to clean up the history - please hold off merging for now.

rade:
I really don't care. But if you must...

PR #505, opened 2015-04-01
Move away from recursive go test, as it cannot output a merged coverage profile.

tomwilkie:
- Update said profile to coveralls.io
- Add badge to readme

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,9 @@ install:
   - go install -tags netgo std
   - make travis
 
-script: make tests
+script:
+  - make tests
+
+after_success:
+  - go get github.com/mattn/goveralls
+  - $HOME/gopath/bin/goveralls -coverprofile=profile.cov -service=travis-ci
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,15 @@ $(DOCKER_DISTRIB):
 	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
 
 tests:
-	go test -cover -tags netgo ./...
+	echo "mode: count" > profile.cov
+	for dir in $$(find . -type f -name '*_test.go' | xargs -n1 dirname | sort -u); do \
+	    go test -tags netgo -covermode=count -coverprofile=$$dir/profile.tmp $$dir;       \
+	    if [ -f $$dir/profile.tmp ]; then                                                 \
+	        cat $$dir/profile.tmp | tail -n +2 >> profile.cov;                            \
+	        rm $$dir/profile.tmp;                                                         \
+	    fi                                                                                \
+	done
+	go tool cover -html=profile.cov -o=coverage.html
 
 $(PUBLISH): publish_%:
 	$(SUDO) docker tag -f $(DOCKERHUB_USER)/$* $(DOCKERHUB_USER)/$*:$(WEAVE_VERSION)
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -13,4 +13,9 @@ install:
   - go install -tags netgo std
   - make travis
 
-script: make tests
+script:
+  - make tests
+
+after_success:
+  - go get github.com/mattn/goveralls
+  - $HOME/gopath/bin/goveralls -coverprofile=profile.cov -service=travis-ci
```
rade: newline missing at end of file

Makefile
```diff
@@ -61,7 +61,14 @@ $(DOCKER_DISTRIB):
...
 tests:
-	go test -cover -tags netgo ./...
+	echo "mode: count" > profile.cov
+	for dir in $$(find . -type f -name '*_test.go' | xargs -n1 dirname | sort | uniq); do \
+	    go test -tags netgo -covermode=count -coverprofile=$$dir/profile.tmp $$dir;       \
+	    if [ -f $$dir/profile.tmp ]; then                                                 \
+	        cat $$dir/profile.tmp | tail -n +2 >> profile.cov;                            \
+	        rm $$dir/profile.tmp;                                                         \
+	    fi                                                                                \
+	done
```
rade: Shouldn't profile.cov be added to .gitignore? Also, what happens to this file? I.e. when I run this manually, how do I look at the results?

tomwilkie:
PTAL

Makefile
```diff
@@ -61,7 +61,15 @@ $(DOCKER_DISTRIB):
...
+	go tool cover -html=profile.cov -o=coverage.html
```
rade: @dpw mind taking a quick look at the above?

Makefile
```diff
@@ -61,7 +61,15 @@ $(DOCKER_DISTRIB):
 	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
 
 tests:
-	go test -cover -tags netgo ./...
+	echo "mode: count" > profile.cov
+	for dir in $$(find . -type f -name '*_test.go' | xargs -n1 dirname | sort | uniq); do \
```
dpw: This an be expressed more concisely with something like:

    find . -type f -name '*_test.go' -printf '%h' | sort -u

tomwilkie: No -printf on mac os x I'm afraid; probably not important for this repo, but thats why its there.  Still want me to change it?

Makefile
```diff
@@ -61,7 +61,15 @@ $(DOCKER_DISTRIB):
...
+	    go test -tags netgo -covermode=count -coverprofile=$$dir/profile.tmp $$dir;       \
+	    if [ -f $$dir/profile.tmp ]; then                                                 \
+	        cat $$dir/profile.tmp | tail -n +2 >> profile.cov;                            \
```
dpw: No need to invoke cat there:

    tail -n +2 <$$dir/profile.tmp >> profile.cov

Also, does the temp file need to be in the directory?  Better to put it under tmp if it doesn't (e.g. `${TMPDIR:-/tmp}/profile.$$`).

Makefile
dpw: Ok, just the `sort -u` then.

tomwilkie: Done; thanks!

PR #510, opened 2015-04-01
Generalise updater.go and move into its own package

bboreham:
This is so it can be re-used by IPAM


bboreham:
Moved to #511 because I needed to change branch

PR #512, opened 2015-04-01
Use docker stop instead of docker kill

bboreham:
The difference is that `docker stop` sends `SIGTERM` first, whereas `docker kill` just sends `SIGKILL`, and the latter cannot be caught by a process.  Although we don't presently know of anything that needs to shut down nicely, it seems better to use the variant that permits it.

### DIFF
--- a/weave
+++ b/weave
@@ -759,7 +759,7 @@ case "$COMMAND" in
         ;;
     stop)
         [ $# -eq 0 ] || usage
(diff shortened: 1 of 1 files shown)


rade:
The reason we use 'kill' is that in the past we frequently saw 'stop' not terminating. Might have been [this problem](http://stackoverflow.com/questions/22413563/docker-container-refuses-to-get-killed-after-run-command-turns-into-a-zombie), in which case it's fixed in recent kernels.

PR #516, opened 2015-04-05
lower topology update gossip complexity

### LINKED ISSUE #514: gossip of topology changes is O(n^2)
Topology gossip does not use broadcast routing to inform other peers of changes, because routing tables are calculated asynchronously (in order to prevent deadlocks), which means the broadcast routing information has a high probability of being out of date exactly when we would want to use it to for transmission of a topology change.

Instead topology gossip transmits changes via the same mechanism as the period gossip. Currently that involves sending the data to all neighbours, which then apply it and, if anything new was found, transmit that delta to all their neighbours etc.

Unfortunately that means propagating a single change in a fully connected network of peers is O(n^2), since each of the originating peer's neighbours will see the change as new information and hence in turn send it it all its neighbours.

We should fix that.

rade:
Transmit topology updates via gossip broadcast instead of the "send to all neighbours" periodic gossip mechanism.

This requires the following changes:

1. make `Gossip.GossipBroadcast` operate on `GossipData`. We need this so we can perform accumulation (see below). It also gives the opportunity to update data as we relay it, which is an optimisation.
2. add "representation of received update" to return of `Peers.ApplyUpdate`. We need this so we can perform the aforementioned data updating for when relaying topology gossip.
3. use `GossipSender`s for `GossipBroadcast`. Depends on 1. We need this so `GossipBroadcast` becomes properly async, which is required if we want to use it for topology gossip - in order to avoid deadlocks w/o spawning a unbounded number of goroutines. It is also a great optimisation since `GossipData` is accumulated and sent in one go. Note that there is one `GossipSender` per broadcast source since that is the largest granularity at which accumulation is possible. As with the existing `GossipSender`s, we need a way to GC stale entries - in this case for broadcast source peers that have disappeared.
4. introduce `Routes.EnsureRecalculated`. We need this so that when topology gossip uses `GossipBroadcast`, we ensure the broadcast routes are up to date wrt the latest topology.
5. Switch to use `Gossip.GossipBroadcast` for topology updates. Depends on 1,2,3,4.

Closes #514.

### DIFF
--- a/router/gossip.go
+++ b/router/gossip.go
@@ -20,13 +20,15 @@ type Gossip interface {
 	// specific message from one peer to another
 	// intermediate peers relay it using unicast topology.
 	GossipUnicast(dstPeerName PeerName, msg []byte) error
-	// send a message to every peer, relayed using broadcast topology.
-	GossipBroadcast(msg []byte) error
+	// send gossip to every peer, relayed using broadcast topology.
+	GossipBroadcast(update GossipData) error
 }
 
 type Gossiper interface {
 	OnGossipUnicast(sender PeerName, msg []byte) error
-	OnGossipBroadcast(msg []byte) error
+	// merge received data into state and return a representation of
+	// the received data, for further propagation
+	OnGossipBroadcast(update []byte) (GossipData, error)
 	// return state of everything we know; gets called periodically
 	Gossip() GossipData
 	// merge received data into state and return "everything new I've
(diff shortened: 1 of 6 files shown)


rade:
@bboreham would you mind taking a look at this?

The individual commits are self-contained, modulo the dependencies as indicated, so I recommend reviewing them individually. We could also merge them into master individually, if that is helpful. 2+4 are behaviour preserving when merged into master. 1 is an API change that doesn't affect master, but probably does affect IPAM. 3 doesn't affect behaviour on master, and should be an optimisation for IPAM. Only 5 affects the behaviour on master.

bboreham:
"It also ensures we update data as we relay it" - there is nothing in the code to ensure any updating of data, nor any mention in the comment on OnBroadcastUpdate() .
"It gives the opportunity to update data as we relay it" seems more accurate.

bboreham:
Actually, I see that the updating will fall out from the natural usage of GossipData.

PR #518, opened 2015-04-06
Canonical sets

rade:


rade:
@bboreham mind taking a quick look at this?

rade:
@bbohreham never mind, I've merged it. But you may want to take a look anyway, just FYI.

PR #521, opened 2015-04-07
lower topology periodic gossip complexity

### LINKED ISSUE #517: periodic topology gossip is O(n^2)
Even after #514, there is still some O(n^2) complexity remaining in topology gossip, namely in the period gossip. This currently sends the gossip data to all neighbours. So we have n peers performing n-1 sends.

Traditional gossip protocols only send gossip to a random subset (of, usually log(n) size) of neighbours. However, this assumes a relatively regular topology. With highly irregular topologies, especially those containing "bottleneck" links (where one set of peers can reach another only via a very small number of links), propagation delays are prohibitive.

That's why we decided to gossip to all neighbours.

But we can do better. Peers have knowledge of the topology and hence know which links are bottlenecks. We can use that information to select random neighbours for gossip communication, with the random distribution skewed towards the bottleneck links.

rade:
Instead of gossiping to all neighbours, we select up to log(n_peers) neighbours at random, based on a probability distribution that takes into account the topology, i.e. favours neighbours at the end of "bottleneck" links. The distribution is determined from the unicast topology.

Closes #517.

### DIFF
--- a/router/gossip.go
+++ b/router/gossip.go
@@ -182,19 +182,25 @@ func (c *GossipChannel) deliver(_ PeerName, _ []byte, dec *gob.Decoder) error {
 
 func (c *GossipChannel) Send(data GossipData) {
 	connections := c.ourself.Connections() // do this outside the lock so they don't nest
-	retainedSenders := make(connectionSenders)
 	c.Lock()
 	defer c.Unlock()
+	// GC - randomly (courtesy of go's map iterator) pick some
+	// existing entries and stop&remove them if the associated
+	// connection is no longer active.  We stop as soon as we
+	// encounter a valid entry; the idea being that when there is
+	// little or no garbage then this executes close to O(1), whereas
+	// when there is lots of garbage we remove it quickly.
+	for conn, sender := range c.senders {
+		if foundConn, found := c.ourself.ConnectionTo(conn.Remote().Name); !found || foundConn != conn {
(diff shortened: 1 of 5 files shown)


rade:
@bboreham review please.

bboreham:
(discussed offline) selecting `min(n_connections, log(n_peers))` _connections_ to send down (as opposed to `log(n_peers)` peers, which may lie down fewer connections) seems a worthwhile improvement - more resilient to awkward topologies, while retaining the benefit that the overall cost is O(nlogn)

rade:
I've changed the neighbour selection to use the min(n_neighbours, log(n_peers)), excluded the gossip sender, and added some documentation.

router/gossip.go
```diff
@@ -167,7 +167,7 @@ func (c *GossipChannel) deliverBroadcast(srcName PeerName, _ []byte, dec *gob.De
 	return c.relayBroadcast(srcName, data)
 }
 
-func (c *GossipChannel) deliver(_ PeerName, _ []byte, dec *gob.Decoder) error {
+func (c *GossipChannel) deliver(srcName PeerName, _ []byte, dec *gob.Decoder) error {
```
bboreham: (not really in scope of this change) the need for the second parameter on GossipChannel.deliver() seems to have gone away.

Issue #507, opened 2015-04-01
`weave` and `weave --local` behave differently


awh:
`weave` behaves differently depending on whether `--local` is specified:

### `STDERR` is merged into `STDOUT` when run remotely
````
vagrant@ubuntu-14:~/weave$ sudo ./weave status 2>/dev/null
weave container is not present; have you launched it?

weavedns container is not present; have you launched it?
vagrant@ubuntu-14:~/weave$ sudo ./weave --local status 2>/dev/null

vagrant@ubuntu-14:~/weave$
````
### Lines are terminated with CRLF when run remotely
````
vagrant@ubuntu-14:~/weave$ sudo ./weave ps|hexdump -c
0000000   b   b   7   e   6   d   a   0   b   e   1   4       3   6   :
0000010   1   1   :   0   4   :   9   8   :   e   e   :   a   6       1
0000020   0   .   0   .   0   .   1   /   2   4  \r  \n   a   d   1   9
0000030   1   8   0   7   7   8   0   d       7   a   :   b   c   :   a
0000040   c   :   d   3   :   0   4   :   f   f  \r  \n
000004c
vagrant@ubuntu-14:~/weave$ sudo ./weave --local ps|hexdump -c
0000000   b   b   7   e   6   d   a   0   b   e   1   4       3   6   :
0000010   1   1   :   0   4   :   9   8   :   e   e   :   a   6       1
0000020   0   .   0   .   0   .   1   /   2   4  \n   a   d   1   9   1
0000030   8   0   7   7   8   0   d       7   a   :   b   c   :   a   c
0000040   :   d   3   :   0   4   :   f   f  \n
000004a
vagrant@ubuntu-14:~/weave$
````

**Comment by awh** [comment]
The LF->CRLF termination is being caused by the intervening Docker pseudoterminal. Can you see any problems with changing the `weaveexec/Dockerfile` entrypoint to the following @squaremo?

    ENTRYPOINT ["stty -onlcr; /home/weave/weave"]

**Comment by squaremo** [comment]
I think it would have to be

```
ENTRYPOINT "stty -onlcr; /home/weave/weave"
```

but otherwise, worth a crack yep.

dpw:
I suspect neither of those will work as presented due to the need to handle arguments passed to `docker run`.  The docker docs are not very clear on the matter though.

awh:
You're right, it requires the shell form of entrypoint, and consequently can't be used for two reasons:

[...]

PR #523, opened 2015-04-08
Implement weaveexec signal proxy

awh:
Address #507 (`weave` and `weave --local` behave differently).

Introduce an intermediate signal proxying process that:

* Runs as PID 1 inside the weaveexec container
* Installs a SIGINT handler
* Forks and execs the weave script
* Redelivers inbound SIGINTs (received from `docker run --sig-proxy=true`) to its own process group, which includes the shell running the weave script and any subprocesses (such as `curl`) it is executing
* Waits for child to exit and then exits with child's exit status

As a consequence we can disable the problematic docker pseudoterminal allocation without losing the ability to interrupt the script running in the weaveexec container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,8 @@ $(WEAVER_EXE) $(WEAVEDNS_EXE): common/*.go
 
 $(WEAVER_EXE): router/*.go weaver/main.go
 $(WEAVEDNS_EXE): nameserver/*.go weavedns/main.go
+
+# Sigproxy needs separate rule as it fails the netgo check in the main build stanza due to not importing net package 
 $(SIGPROXY_EXE): sigproxy/main.go
 	go build -o $@ ./$(shell dirname $@)
 
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ DOCKERHUB_USER=zettio
 WEAVE_VERSION=git-$(shell git rev-parse --short=12 HEAD)
 WEAVER_EXE=weaver/weaver
 WEAVEDNS_EXE=weavedns/weavedns
+SIGPROXY_EXE=sigproxy/sigproxy
 WEAVER_IMAGE=$(DOCKERHUB_USER)/weave
 WEAVEDNS_IMAGE=$(DOCKERHUB_USER)/weavedns
 WEAVEEXEC_IMAGE=$(DOCKERHUB_USER)/weaveexec
@@ -42,6 +43,8 @@ $(WEAVER_EXE) $(WEAVEDNS_EXE): common/*.go
 
 $(WEAVER_EXE): router/*.go weaver/main.go
 $(WEAVEDNS_EXE): nameserver/*.go weavedns/main.go
+$(SIGPROXY_EXE): sigproxy/main.go
+	go build -o $@ ./$(shell dirname $@)
 
 $(WEAVER_EXPORT): weaver/Dockerfile $(WEAVER_EXE)
 	$(SUDO) docker build -t $(WEAVER_IMAGE) weaver
@@ -51,8 +54,9 @@ $(WEAVEDNS_EXPORT): weavedns/Dockerfile $(WEAVEDNS_EXE)
 	$(SUDO) docker build -t $(WEAVEDNS_IMAGE) weavedns
 	$(SUDO) docker save $(WEAVEDNS_IMAGE):latest > $@
 
-$(WEAVEEXEC_EXPORT): weaveexec/Dockerfile $(DOCKER_DISTRIB) weave
+$(WEAVEEXEC_EXPORT): weaveexec/Dockerfile $(DOCKER_DISTRIB) weave $(SIGPROXY_EXE)
 	cp weave weaveexec/weave
+	cp sigproxy/sigproxy weaveexec/sigproxy
 	cp $(DOCKER_DISTRIB) weaveexec/docker.tgz
 	$(SUDO) docker build -t $(WEAVEEXEC_IMAGE) weaveexec
 	$(SUDO) docker save $(WEAVEEXEC_IMAGE):latest > $@
@@ -81,7 +85,7 @@ publish: $(PUBLISH)
 
 clean:
 	-$(SUDO) docker rmi $(WEAVER_IMAGE) $(WEAVEDNS_IMAGE) $(WEAVEEXEC_IMAGE)
-	rm -f $(WEAVER_EXE) $(WEAVEDNS_EXE) $(WEAVER_EXPORT) $(WEAVEDNS_EXPORT) $(WEAVEEXEC_EXPORT)
+	rm -f $(WEAVER_EXE) $(WEAVEDNS_EXE) $(SIGPROXY_EXE) $(WEAVER_EXPORT) $(WEAVEDNS_EXPORT) $(WEAVEEXEC_EXPORT)
 
 build:
 	$(SUDO) go clean -i net
--- a/sigproxy/main.go
+++ b/sigproxy/main.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"log"
+	"os"
+	"os/exec"
+	"os/signal"
+	"syscall"
+)
+
+func checkArguments() {
+	if len(os.Args) == 1 {
+		log.Fatal("USAGE: sigproxy <command> [arguments ...]")
+	}
+}
+
+func installSignalHandler() chan os.Signal {
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc, os.Interrupt)
+
+	return sc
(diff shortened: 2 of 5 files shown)


awh:
@bboreham @dpw @rade review at your convenience, thankyou!

Makefile
```diff
@@ -10,6 +10,7 @@ DOCKERHUB_USER=zettio
 WEAVE_VERSION=git-$(shell git rev-parse --short=12 HEAD)
 WEAVER_EXE=weaver/weaver
 WEAVEDNS_EXE=weavedns/weavedns
+SIGPROXY_EXE=sigproxy/sigproxy
```
awh: Is this a good name?

Makefile
```diff
@@ -42,6 +43,8 @@ $(WEAVER_EXE) $(WEAVEDNS_EXE): common/*.go
 
 $(WEAVER_EXE): router/*.go weaver/main.go
 $(WEAVEDNS_EXE): nameserver/*.go weavedns/main.go
+$(SIGPROXY_EXE): sigproxy/main.go
+	go build -o $@ ./$(shell dirname $@)
```
awh: I added another `go build` instruction here rather than add `$(SIGPROXY_EXE)` to the target above because the netgo check fails when compiling the signal proxy - possibly because it doesn't import net at all? Should we do this a different way?

sigproxy/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+func checkArguments() {
+	if len(os.Args) == 1 {
+		log.Fatal("USAGE: sigproxy <command> [arguments ...]")
+	}
+}
+
+func installSignalHandler() chan os.Signal {
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc, os.Interrupt)
```
awh: Do we want to support signals other than SIGINT?

sigproxy/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return cmd
+}
+
+func forwardSignals(sc chan os.Signal, cmd *exec.Cmd) {
+	// Unfortunately there's no way to convert the os.Signal type from
+	// the channel into a numeric signal number for syscall.Kill. Wait for
+	// (and discard) os.Interrupt, then deliver SIGINT to process group
+	<-sc
+	syscall.Kill(0, syscall.SIGINT)
```
awh: Should these statements live inside an infinite `for {}` loop?

sigproxy/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// These conveniently default to /dev/null otherwise...
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Start(); err != nil {
+		log.Fatal(err)
+	}
+
```
awh: Should we be closing stdin/stdout/stderr once we've successfully fork & execed?

sigproxy/main.go
```diff
@@ -0,0 +1,71 @@
...
+func checkArguments() {
+	if len(os.Args) == 1 {
+		log.Fatal("USAGE: sigproxy <command> [arguments ...]")
+	}
+}
+
+func installSignalHandler() chan os.Signal {
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc, syscall.SIGINT)
```
awh: Should we be handling any other signals too?

bboreham: SIGTERM seems like a good idea; this is what `docker stop` will send.

sigproxy/main.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func waitAndExit(cmd *exec.Cmd) {
+	if err := cmd.Wait(); err != nil {
+		// Exit status is platform specific so not readily accessible - casts
+		// required to access system-dependent exit information
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			if waitStatus, ok := exitErr.Sys().(syscall.WaitStatus); ok {
+				os.Exit(waitStatus.ExitStatus())
```
bboreham: why do the checked cast here while doing an unchecked one on syscall.Signal?

weave
```diff
@@ -52,7 +52,7 @@ DNS_IMAGE=$BASE_DNS_IMAGE:$IMAGE_VERSION
 EXEC_IMAGE=$BASE_EXEC_IMAGE:$IMAGE_VERSION
 
 exec_remote() {
-    docker run -t --rm \
+    docker run --sig-proxy=true --rm \
```
bboreham: --sig-proxy=true is the default, hence unnecessary

Makefile
bboreham: I concur on the reason for the failure.  Seems OK to do it this way for now.

sigproxy/main.go
```diff
@@ -0,0 +1,71 @@
...
+			} else {
+				os.Exit(1)
+			}
+		} else {
+			os.Exit(1)
+		}
+	} else {
+		os.Exit(0)
+	}
+}
```
rade: Please get rid of all those `else` branches. And that innermost `os.Exit(1)`.

sigproxy/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"os/exec"
+	"os/signal"
+	"syscall"
+)
+
+func checkArguments() {
+	if len(os.Args) == 1 {
+		log.Fatal("USAGE: sigproxy <command> [arguments ...]")
+	}
+}
```
rade: This function is overkill, imo.

sigproxy/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+func main() {
+	checkArguments()
+	sc := installSignalHandler()
+	cmd := execCommand()
+	go forwardSignals(sc)
+	waitAndExit(cmd)
+}
```
rade: I'd inline all those functions.

sigproxy/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+func installSignalHandler() chan os.Signal {
+	sc := make(chan os.Signal, 1)
+	signal.Notify(sc, syscall.SIGINT)
+
+	return sc
+}
+
+func execCommand() *exec.Cmd {
+	// First argument is command to run, remainder are its arguments
```
rade: comment is superfluous.

Makefile
rade: Have you checked that this isn't dragging in any dynamically linked libs? I can't remember the precise reasons for the ldflags magic in the main build stanza.

Also, I suggest you add a comment explaining why this isn't using the main build stanza.

sigproxy/main.go
rade: Any reason why you split the signal handler installation and forwarding, and the command execution and waiting? For the latter, why not just use `cmd.Run()`?

Makefile
awh: Looks OK:
````
$ file sigproxy/sigproxy
sigproxy/sigproxy: ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), statically linked, not stripped
````

sigproxy/main.go
awh: Good question. I think on Linux it's safe to do an unchecked cast for both `syscall.Signal` and `exitErr.Sys()`; the `err` from `cmd.Wait()` can be something other than an `ExitError` though. Do we have a general policy on this kind of thing?

weave
awh: It is unnecessary, I included it for documentation. Would you still prefer it be removed?

rade: yes

sigproxy/main.go
awh: > I'd inline all those functions.

I split it out into functions to make it self documenting - you'd rather I not?

> Any reason why you split the signal handler installation and forwarding

Yes - I wanted to make sure I was listening for signals before the child process started, but only attempt delivery after it had (also I was originally forwarding signals onto the child process rather than the process group so `cmd`was being passed into `forwardSignals` as well)

rade:
LGTM, but would like @dpw to take a look.

dpw:
The long comment line in the makefile needs a M-q.  Fine otherwise.

dpw:
> The long comment line in the makefile needs a M-q.

Even worse - it has trailing whitespace!

sigproxy/main.go
rade: @dpw thoughts on this particular point?

dpw: I don't think it would hurt to close them, and it is perhaps more correct (keep stderr open though, just in case). But I think it would be rare that it makes a difference in practice, and, depending on the details of how docker handles stdout/stderr, it might make no difference at all.

PR #526, opened 2015-04-08
Allow container names to be supplied to `weave ps`

bboreham:
So, `weave ps` will still show you all containers, but `weave ps mycontainer` will just show you that one.

If you give a nonexistent name you get no output, which is consistent with Unix `ps`.

### DIFF
--- a/site/troubleshooting.md
+++ b/site/troubleshooting.md
@@ -146,6 +146,12 @@ IP address/routing prefix length ([CIDR
 notation](http://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing))
 assigned on the weave network.
 
+You can also supply a list of container IDs/names to `weave ps`, like this:
+
+    $ sudo weave ps able baker
+    able ce:15:34:a9:b5:6d 10.2.5.1/24
+    baker 7a:61:a2:49:4b:91 10.2.8.3/24
+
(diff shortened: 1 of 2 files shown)


inercia:
+1

rade:
I am not completely sold in this. `docker ps` doesn't take a container argument. And filtering by container id is just one 'grep' away. So what's the rationale here? Performance?

bboreham:
I needed something like this to run smoke-tests for IPAM.

`docker ps` does have a  `--filter name=foo` option (on master).

But yes, performance. In order to grep the `weave ps` output you need the short form of the container hex ID, which needs a `docker inspect`, possibly remotely.  Also, it's much more inefficient to obtain details on every container then grep out the ones you don't need.

rade:
ok. rationale accepted.

AFAICT, when this is fed a name, the output will contain that name, not the container id. Shouldn't the output *always* contain the id?

Docs could perhaps do with a sentence mentioning the extra functionality.

bboreham:
While putting the hex ID in the output would be consistent, it seems rather a complex implementation to add for little benefit.  If the caller of `weave ps` knows the name of a container and doesn't know the ID, then has to do extra work to be able to match them up.

bboreham:
Doc change added.

PR #527, opened 2015-04-08
IP Allocation Management

bboreham:
Implements the scheme described at https://github.com/weaveworks/weave/wiki/IP-allocation-design, with address ranges controlled by a CRDT gossiped between peers.

`weave launch` takes a new `-alloc` argument specifying the subnet to work within.

Commands updated to take advantage of IPAM: `weave run`, `weave start`, `weave attach`, `weave expose`


router/gossip.go
```diff
@@ -224,6 +228,16 @@ func (c *GossipChannel) GossipBroadcast(update GossipData) error {
...
+func (c *GossipChannel) LeaderElect() PeerName {
+	highest := PeerName(0)
+	c.ourself.Router.Peers.ForEach(func(peer *Peer) {
+		if highest < peer.Name {
+			highest = peer.Name
+		}
+	})
+	return highest
+}
+
```
rade: This looks like it's in the wrong place. There is one call site (outside test code) and that needs a cast.

README.md
```diff
@@ -1,6 +1,6 @@
 # Weave - the Docker network
 
-[![Build Status](https://travis-ci.org/weaveworks/weave.svg?branch=master)](https://travis-ci.org/weaveworks/weave) [![Coverage Status](https://coveralls.io/repos/weaveworks/weave/badge.svg)](https://coveralls.io/r/weaveworks/weave)
+[![Build Status](https://travis-ci.org/zettio/weave.svg?branch=master)](https://travis-ci.org/zettio/weave) [![Coverage Status](https://coveralls.io/repos/zettio/weave/badge.svg)](https://coveralls.io/r/zettio/weave)
```
rade: This has nothing to do with this PR.

ipam/allocator.go
```diff
@@ -0,0 +1,251 @@
...
+	buf.WriteString(fmt.Sprintf("Allocator universe %s+%d\n", alloc.universeStart, alloc.universeSize))
+	buf.WriteString(alloc.ring.String())
+	buf.WriteString(alloc.spaceSet.String())
+	buf.WriteString("\nPending requests for ")
+	for _, pending := range alloc.pending {
+		buf.WriteString(pending.Ident)
+		buf.WriteString(", ")
+	}
+	return buf.String()
+}
```
rade: https://github.com/weaveworks/weave/commit/45740b1419508d442ec681893eb8f4c2aac93736

ipam/space/space_set.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (s *Set) String() string {
+	var buf bytes.Buffer
+	buf.WriteString("Set")
+	for _, space := range s.spaces {
+		buf.WriteString(fmt.Sprintf("\n  %s", space))
+	}
+	return buf.String()
+}
```
rade: https://github.com/weaveworks/weave/commit/45740b1419508d442ec681893eb8f4c2aac93736

ipam/allocator.go
rade: Also, I'd inline this function at the call site.

ipam/allocator.go
```diff
@@ -0,0 +1,251 @@
...
+		alloc.infof("I was elected leader of the universe\n%s", alloc.string())
+		alloc.gossip.GossipBroadcast(alloc.Gossip())
+		alloc.considerOurPosition()
+	} else {
+		alloc.sendRequest(leader, msgLeaderElected)
+	}
+}
+
+// return true if the request is completed, false if pending
+func (alloc *Allocator) tryAllocateFor(ident string, resultChan chan<- net.IP) bool {
```
rade: It's odd for this to take a chan<-. Make it return `(net.IP, bool)`.

ipam/allocator.go
```diff
@@ -0,0 +1,251 @@
...
+	// some other peer decided we were the leader:
+	// if we already have tokens then they didn't get the memo; repeat
+	if !alloc.ring.Empty() {
+		alloc.gossip.GossipBroadcast(alloc.Gossip())
+	} else {
+		// re-run the election here to avoid races
+		alloc.electLeaderIfNecessary()
+	}
+	return nil
+}
```
rade: inline this function.

ipam/actor.go
```diff
@@ -0,0 +1,165 @@
+package ipam
```
rade: Merge this whole file into allocator.go

ipam/allocator.go
```diff
@@ -0,0 +1,251 @@
...
+		universeStart: universeNet.IP,
+		universeSize:  universeSize,
+		universeLen:   ones,
+		ring:          ring.New(utils.Add(universeNet.IP, 1), utils.Add(universeNet.IP, universeSize-1), ourName),
+		owned:         make(map[string][]net.IP),
+	}
+	return alloc, nil
+}
+
+func (alloc *Allocator) string() string {
```
rade: seems odd for this to be private.

ipam/claims.go
```diff
@@ -0,0 +1,90 @@
...
+	alloc.infof("Address %s claimed by %s", addr, ident)
+	if owner, err := alloc.checkClaim(ident, addr); err != nil {
+		resultChan <- err
+	} else if owner != alloc.ourName {
+		alloc.infof("Address %s owned by %s", addr, owner)
+		alloc.claims = append(alloc.claims, pendingClaim{resultChan, ident, addr})
+	} else {
+		resultChan <- nil
+	}
+}
```
rade: inline this function at the call site. Yes, it's quite big, but the special treatment of resultChan is too obscure otherwise.

ipam/claims.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (alloc *Allocator) handleCancelClaim(ident string, addr net.IP) {
+	for i, claim := range alloc.claims {
+		if claim.Ident == ident && claim.IP.Equal(addr) {
+			alloc.claims = append(alloc.claims[:i], alloc.claims[i+1:]...)
+			break
+		}
+	}
+}
+
```
rade: inline this at the call site.

ipam/claims.go
```diff
@@ -0,0 +1,90 @@
+package ipam
```
rade: merge this entire file into allocator.go

ipam/gossip.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// OnGossip (Sync)
+func (alloc *Allocator) OnGossip(msg []byte) (router.GossipData, error) {
+	alloc.debugln("Allocator.OnGossip:", len(msg), "bytes")
+	resultChan := make(chan error)
+	alloc.actionChan <- func() {
+		resultChan <- alloc.updateRing(msg)
+	}
+	err := <-resultChan
```
rade: inline this receive below.

ipam/http.go
```diff
@@ -0,0 +1,123 @@
...
+					err.Error())
+				return
+			}
+
+			w.WriteHeader(204)
+		default:
+			http.Error(w, "Verb not handled", http.StatusBadRequest)
+		}
+	})
+}
```
rade: This really could do with a `badRequest(w, "description", err)` helper.

site/features.md
```diff
@@ -53,6 +54,29 @@ now re-use the same tools and techniques when deploying applications
...
+Weave uses the Docker Events API to learn when a container has exited
+and hence can release its IP address.
+
+You may wish to `weave stop` and re-launch to change some config or to
+upgrade to a new version; provided the underlying protocol hasn't
+changed it will pick up where it left off and learn from peers in the
+network which address ranges it was previously using. If, however, you
+run `weave reset` this will remove the peer from the network so
+if Weave is run again on that node it will start from scratch.
+
```
rade: What happens if you call `weave stop` and then *don't* relaunch? And what happens if the host just dies?

Also, surely this should mention `start` and `expose` too.

And we may want to mention the new (#526) `weave ps <container> ...` to find out what IP address was allocated.

site/features.md
```diff
@@ -53,6 +54,29 @@ now re-use the same tools and techniques when deploying applications
 as containers as we would have done when deploying them 'on metal' in
 our data centre.
 
+### <a name="ipam"></a>IP Address Management
+
+Weave can assign unique IP addresses to each container across the
+network.  To make this work, weave must be told what range of
+addresses to allocate from, on each host.  For example:
```
rade: The wording here is ambiguous. "on each host" could be be interpreted as meaning that each host must be given a *separate* range to allocate from. Whereas usually the specified range should be the same across hosts (what happens if it isn't?).

ipam/actor.go
```diff
@@ -0,0 +1,165 @@
...
+func (alloc *Allocator) Start() {
+	actionChan := make(chan func(), router.ChannelSize)
+	alloc.actionChan = actionChan
+	go alloc.actorLoop(actionChan)
+}
+
+// Actor client API
+
+// GetFor (Sync) - get IP address for container with given name
+func (alloc *Allocator) GetFor(ident string, cancelChan <-chan bool) net.IP {
```
rade: cancelChan looks like a single-shot signalling channel. in which case make it of type `struct{}` and close it instead of sending to it.

ipam/actor.go
```diff
@@ -0,0 +1,165 @@
...
+	resultChan := make(chan net.IP, 1) // len 1 so actor can send while cancel is in progress
```
rade: you could keep this at 0 by making cancellation read from resultChan, and the cancellation handler send nil to the resultChan if it finds the id.

ipam/actor.go
```diff
@@ -0,0 +1,165 @@
...
+	select {
+	case result := <-resultChan:
+		return result
+	case <-cancelChan:
+		alloc.actionChan <- func() {
+			alloc.handleCancelClaim(ident, addr)
+		}
+		return nil
+	}
+}
```
rade: see comments on `GetFor`, which equally apply here.

ipam/actor.go
tomwilkie: This chan comes from go's http package's closeNotifier (http://golang.org/src/net/http/server.go?s=3493:3661#L96) - is it worth the effort translating this to a chan struct{}?

ipam/actor.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		delete(alloc.owned, ident)
+	}
+	return nil // this is to satisfy the ContainerObserver interface
+}
+
+// OnShutdown (Sync)
+func (alloc *Allocator) Shutdown() {
+	alloc.infof("Shutdown")
+	doneChan := make(chan bool)
```
rade: this looks like a one-shot signal channel, in which case make it of type `struct{}` and close it instead of sending `true`.

ipam/actor.go
rade: ah ok. No, in that case it's not worth the hassle. Isn't it great that the go standard libraries don't follow their own best practices?

ipam/actor.go
tomwilkie: As this was mostly boilerplate we chose to keep it out of allocator.go; this seems pretty idiomatic for a go PoV.  It keeps allocator.go more manageable and understandable.  Not as relevant now we don't have all those structs & type switches, but I still think it counts.  Do you have any rules as to when it is appropriate to spit large files?

rade: I wouldn't mind seeing a split along functionality lines. See my comments on claims.go.

As an aside, if we do want to split that "client" and "server" part of the actor, then the *client* part, including the New* constructor, should be what lives in the <actor-name>.go since the functions there define the API; the fact that the API is implemented in terms of an actor is an implementation detail.

tomwilkie:
I would appreciate if this was a mutli-author commit - the history of the bits I wrote was lost when this was squashed into one change.

rade:
"We are all individuals!"

router/gossip.go
bboreham: Yes, the cast is there to remind me I'm not entirely sure which way to go: 

I think a case can be made that leadership is related to gossip, so it can go on the Gossip interface.  

Or the thing that uses it (Allocator, in the first instance) could be passed a reference to a Leadership interface.  If we go that way, Peers could be the implementer.

ipam/actor.go
bboreham: I looked at this - we'd have to make the cancel side synchronous or put it in a separate goroutine, and the resulting overhead seems unwarranted compared to just making the chan size 1.

ipam/allocator.go
tomwilkie: string() is private as it is the implementation for String() that executes on the actor thread.

Didn't want to inline is to preserve the actor / implementation split.

Still want us to inline it?

rade: ah yes. Once again I got confused by which file these methods are in. Yes, inline it. It's short enough.

bboreham: I have used string() while debugging so it would be inconvenient for it to be only accessible from outside the actor

ipam/gossip.go
bboreham: sorry, can you re-explain "inline this receive below"; it may have lost context.

site/features.md
bboreham: There is an http action DELETE /peer/peername which the administrator can use to take a peer out of the ring.  Currently there is no script command to call this action, nor any easy way to get hold of the peername of a dead peer.

site/features.md
bboreham: "what happens if it isn't" is a lot of angry log messages.  We should make this easier for the user to notice / react to.

bboreham: PR #552 will bring a mismatched allocator range out to `weave status` - it will break the connection and show up in the 'reconnects' section

site/features.md
bboreham: Tom added a command `weave rmpeer` which is used in the case that a peer is never coming back.

router/gossip.go
bboreham: Moved to Peers, and Allocator is now given a separate Leadership reference.

bboreham:
@dpw would you kindly consider reviewing this, please?  
We will squash the commits down into ~2 when all done.

dpw:
> @dpw would you kindly consider reviewing this, please?

Ok.

I would like to make sure I understand the context properly first though.  I know this work has been going on a while, and has already had several team members involved as contributors and reviewers.  So it seems reasonable to ask about the purpose of another review.

My guess is that there is some consensus that it is ready now, but there's also a wish for a fresh set of eyes to do what is hopefully a final pass.

Is that correct?

rade:
Correct.

@bboreham @tomwilkie please make sure this is *really* ready. We are rapidly running out of people who can look at this with fresh eyes.

bboreham:
This redesign (with the Ring) has not been reviewed in earnest by anyone outside its authors.

@tomwilkie and I think it's ready, and yes we are looking for a fresh set of eyes.

dpw:
Ok, then I'd suggest doing any commit munging *before* I look at it, so that I am looking at it in a genuinely final state.  I'm expecting to be ready to hit the merge button on this!

I'd also like to check another thing before I look at it.  This may be something you've already taken care of, but just in case:  I understand that there is a design document to explain how the algorithm works.  Is that document needed to properly understand the code, or not?  If it is, then it ought to be contained in its entirety in the git repo, so that in future it is clear what version of the doc corresponds to what version of the code.  Also, please make sure that the code and the doc do correspond. And if the doc describes design alternatives not implemented by the code, these should be clearly marked as such.  I bring it up because this is the kind of thing that is easy to overlook when one is immersed.

Conversely, if the document is not considered essential to understanding the code, please check that it really is possible to understand the algorithm from the code alone, without it seeming like an exercise in reverse engineering.

tomwilkie:
Its worth saying we know of a few limitations (in the design).  These should probably be included in release notes / docs.  We do not intend to fix this in this PR, as we think what we have is 'good enough' for v1.

- our simple master election algorithm will give different answers on different sides of partition, leading to potentially issuing the same IP twice and an unmergeable ring.  This will happen if you bring up a network in a partitioned state, for instance.  The fix is to use a quorum based master election algorithm, like raft or paxos do.
- similarly, if a healthy ring is partitioned & then all state is lost of one side of the ring (for instance, due to a coordinate power event), you could issue IPs twice (related to above) and the rings will not be mergeable.  The fix is to persist a list of peers, and not blow away the weave container each time - so its out of scope.
- if you use rmpeer & tombstones in a partition, this can sometimes lead to unmergable rings and issuing IP twice.  This is hard to fix, and probably never will be.

bboreham:
Design doc is at https://github.com/weaveworks/weave/wiki/IP-allocation-design
There is another, lower-level, doc in the repo at https://github.com/bboreham/weave/blob/22_ipam/ipam/ipam.md

I suspect I am too far immersed in the code to know what the document really ought to say.

I can move the design doc into the repo; the diagram will have to be converted.

dpw:
Well, I'm hoping not look at any of this until I look at all of it, so I'm not going to follow those links.  But it sounds a bit odd that there are two design docs, even if they have a different emphasis.  Would it be unreasonable to suggest that they be folded together, moving some of the low-level exposition into comments in the code if appropriate?

And while I can understand that it might be convenient to author a document as a wiki page, I think a copy of it should go into the git repo, so that in the future, as things evolve, it's clear which version of the doc goes with which version of the code.

rade:
I suggest adding an IPAM section to our [How it Works](http://weaveworks.github.io/weave/how-it-works.html) docs.

dpw:
> I suggest adding an IPAM section to our How it Works docs.

Maybe, but I want to be clear about my expectations, as I sense that fatigue has set in on this sub-project:  the IPAM design doc should match the code and say whatever needs to be said to understand the code, so that when I review it everything seems clear.  Its location in the repo does not seem like a blocker for merging, as it can easily be changed later on, and how we handle developer docs vs outwards-facing technical docs vs user docs does not seem like something we have quite settled on yet.

PR #528, opened 2015-04-08
Update repository name from zettio to weaveworks

bboreham:
Replaced all occurrences of 'zettio/' with 'weaveworks/'.

Note that the DOCKERHUB_USER remains 'zettio', which is necessary until we release images under the new name.

The expected process for next release is:
  * build, test and push to zettio/weave:some_specific_tags
  * update the release script to change DOCKERHUB_USER to 'weaveworks'
  * build, test and push to weaveworks/weave:specific_tags
  * update the documentation to change the Docker image name from zettio to weaveworks
  * push the documentation change


README.md
```diff
@@ -135,12 +135,12 @@ to each other.
  * [Troubleshooting](http://zettio.github.io/weave/troubleshooting.html)
  * [Building](http://zettio.github.io/weave/building.html)
  * [How it works](http://zettio.github.io/weave/how-it-works.html)
- * [WeaveDNS README](https://github.com/zettio/weave/tree/master/weavedns#readme)
+ * [WeaveDNS README](https://github.com/weaveworks/weave/tree/master/weavedns#readme)
```
rade: surely the `zettio.github.io` links should be changed too.

site/building.md
```diff
@@ -49,8 +49,8 @@ $ make
 ```
 
 This will build the weave components and package them into three
-Docker images (`zettio/weave`, `zettio/weavedns`, and
-`zettio/weaveexec`).  These are then exported (as `weave.tar`,
+Docker images (`weaveworks/weave`, `weaveworks/weavedns`, and
+`weaveworks/weaveexec`).  These are then exported (as `weave.tar`,
```
rade: we haven't changed the image names yet, so the above should remain unchanged too.

site/troubleshooting.md
```diff
@@ -8,7 +8,7 @@ layout: default
 Make sure you are running the latest version - you can download it
 with
 
-    docker pull zettio/weave:latest
+    docker pull weaveworks/weave:latest
```
rade: we haven't changed the image names yet, so the above should remain unchanged too.

dpw:
`bin/release` has `GITHUB_USER=${GITHUB_USER:-zettio}`.  Sure about that?

That's the only remaining zettio reference that looks dodgy to me.

bboreham:
OK; I changed both the GITHUB_USER and the DOCKERHUB_USER since I couldn't think why they should be different.

bboreham:
Re-pushed with GITHUB_USER updated and DOCKERHUB_USER left at 'zettio', in case we want to do the next release to there.

dpw:
I still wonder why  `GITHUB_USER` in `bin/release` is not changed to weaveworks.  According to my understanding of the purpose of that script, changing  `GITHUB_USER` (but not `DOCKERHUB_USER`) will give us the option of doing another release while the docker user name remains `zettio`.  It seems pretty unlikely that we'll want to do that, but it is a logical possibility.  To put it another way, I'm not sure what is special about the `GITHUB_USER` in `bin/release` that explains why it should *not* be changed.

It's obviously a minor thing.  All I'm asking for is a rationale.

bboreham:
I did change GITHUB_USER in bin/release, in 632cc3f

dpw:
> I did change GITHUB_USER in bin/release, in 632cc3f

Ah yes. ETOOEARLY.

rade:
LGTM. @dpw mind merging this? It's rather more urgent than I thought given that currently our beloved travis-ci builds are failing elsewhere (see, e.g. #534).

dpw:
github pages updated

PR #501, opened 2015-03-30
Fix for negative TTLs in WeaveDNS

inercia:
Fixes zettio/weave#494 by using a different lookups set depending on the value stored in the cache.

PS: these changes have been tested with this [new WeaveDNS test tool](https://github.com/inercia/weave-dns-tests).

PR #532, opened 2015-04-09
Fix in WeaveDNS: do not modify cache when getting a reply

inercia:
Changes in this PR:

* do not modify the TTL in the reply stored when `Get`ting from the cache
* set all responses obtained with mDNS as authoritative
* minor improvements in the logs

Note: this PR depends on #501


awh:
Thankyou for splitting this up it's much much easier to digest! LGTM.

inercia:
Great. Thanks for the reviews @awh

PR #534, opened 2015-04-09
Add options to set ports on weaver command line

abligh:
This commit adds two command line options, intended to allow multiple
instances of weave to run without container separation between them.

* -port=PORT makes weave bind to port PORT (an integer). Without the
  option, the default value (6783) is used.

* -bindjsonto=BINDSTRING makes weave bind its control interface to
  a different port. For instance "127.0.0.1:6784" can be used to
  bind to localhost only. ":6784" is the default. An empty string
  can be used to disable the control interface.

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -15,11 +15,12 @@ const (
 )
 
 type ConnectionMaker struct {
-	ourself        *LocalPeer
-	peers          *Peers
-	targets        map[string]*Target
-	cmdLineAddress map[string]struct{}
-	actionChan     chan<- ConnectionMakerAction
+	ourself           *LocalPeer
+	peers             *Peers
+	normalisePeerAddr func(string) string
+	targets           map[string]*Target
+	cmdLineAddress    map[string]struct{}
+	actionChan        chan<- ConnectionMakerAction
 }
 
 // Information about an address where we may find a peer
@@ -31,12 +32,13 @@ type Target struct {
 
 type ConnectionMakerAction func() bool
 
-func NewConnectionMaker(ourself *LocalPeer, peers *Peers) *ConnectionMaker {
+func NewConnectionMaker(ourself *LocalPeer, peers *Peers, normalisePeerAddr func(string) string) *ConnectionMaker {
 	return &ConnectionMaker{
-		ourself:        ourself,
-		peers:          peers,
-		cmdLineAddress: make(map[string]struct{}),
-		targets:        make(map[string]*Target)}
+		ourself:           ourself,
+		peers:             peers,
+		normalisePeerAddr: normalisePeerAddr,
+		cmdLineAddress:    make(map[string]struct{}),
+		targets:           make(map[string]*Target)}
 }
 
 func (cm *ConnectionMaker) Start() {
@@ -47,7 +49,7 @@ func (cm *ConnectionMaker) Start() {
 
 func (cm *ConnectionMaker) InitiateConnection(address string) {
 	cm.actionChan <- func() bool {
-		cm.cmdLineAddress[NormalisePeerAddr(address)] = void
+		cm.cmdLineAddress[cm.normalisePeerAddr(address)] = void
 		if target, found := cm.targets[address]; found {
 			target.tryAfter, target.tryInterval = tryImmediately()
 		}
@@ -57,7 +59,7 @@ func (cm *ConnectionMaker) InitiateConnection(address string) {
 
 func (cm *ConnectionMaker) ForgetConnection(address string) {
 	cm.actionChan <- func() bool {
-		delete(cm.cmdLineAddress, NormalisePeerAddr(address))
+		delete(cm.cmdLineAddress, cm.normalisePeerAddr(address))
 		return false
 	}
 }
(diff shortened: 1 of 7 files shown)


abligh:
Hmm... travis failed here, just like a build does locally, because `weaver/main.go` has:

    import (
           ....
           weave "github.com/zettio/weave/router"

and of course a build will fail against the github version of `weave/router`, as this hasn't been merged yet.

Changing this to:

    weave "../router"

makes it work fine, but presumably you don't want to merge that.

rade:
Hmm. Looks like something isn't quite right in our (relatively new) travis setup here. It appears to be pulling down dependencies from git when it should just use the repo it has already checked out. @tomwilkie any idea?

bboreham:
It's because the name is different in the repo. Fix is in #528

rade:
I reckon we should move `NormalisePeerAddr` to `Router.NormaliseAddr`.

router/router.go
```diff
@@ -31,6 +31,7 @@ type Router struct {
 	ConnLimit       int
 	BufSz           int
 	LogFrame        func(string, []byte, *layers.Ethernet)
+	ListenPort	int
```
rade: I'd just call it 'Port', since it not *just* used for listening.

router/router.go
```diff
@@ -86,8 +88,8 @@ func (router *Router) Start() {
 	router.Macs.Start()
 	router.Routes.Start()
 	router.ConnectionMaker.Start()
-	router.UDPListener = router.listenUDP(Port, po)
-	router.listenTCP(Port)
+	router.UDPListener = router.listenUDP(router.ListenPort, po)
+	router.listenTCP(router.ListenPort)
```
rade: Just use `port` here, i.e. the method param.

router/router.go
```diff
@@ -198,7 +200,7 @@ func (router *Router) listenTCP(localPort int) {
 
 func (router *Router) acceptTCP(tcpConn *net.TCPConn) {
 	// someone else is dialing us, so our udp sender is the conn
-	// on Port and we wait for them to send us something on UDP to
+	// on ListenPort and we wait for them to send us something on UDP to
```
rade: should be `router.Port`

weaver/main.go
```diff
@@ -58,6 +60,8 @@ func main() {
 	flag.StringVar(&prof, "profile", "", "enable profiling and write profiles to given path")
 	flag.IntVar(&connLimit, "connlimit", 30, "connection limit (defaults to 30, set to 0 for unlimited)")
 	flag.IntVar(&bufSz, "bufsz", 8, "capture buffer size in MB (defaults to 8MB)")
+	flag.IntVar(&port, "port", weave.Port, fmt.Sprintf("router port (defaults to %d)",weave.Port))
+	flag.StringVar(&bindJsonTo, "bindjsonto", fmt.Sprintf(":%d", weave.HTTPPort), fmt.Sprintf("address to bind JSON interface to (defaults to :%d, set to \"\" to disable)", weave.HTTPPort))
```
rade: `httpaddr`?

abligh:
Rade: I fixed all your comments bar this one:

In router/router.go:

> > @@ -86,8 +88,8 @@ func (router *Router) Start() {
> >  	router.Macs.Start()
> >  	router.Routes.Start()
> >  	router.ConnectionMaker.Start()
> > -	router.UDPListener = router.listenUDP(Port, po)
> > -	router.listenTCP(Port)
> > +	router.UDPListener = router.listenUDP(router.ListenPort, po)
> > +	router.listenTCP(router.ListenPort)
>
> Just use port here, i.e. the method param.

I didn't understand that one. This is in `Start()`, which takes no parameters. The parameters are passed to `NewRouter()` (immediately above it).

Travis appears to like the commit now.

rade:
> I didn't understand that one.

Ah sorry, github showed me a diff with a function boundary collapsed, and I didn't realise that.

rade:
One more thing. The introduction of a dependency of `ConnectionMaker` on `Router` (via `cm.ourself.Router`) is undesirable. So I suggest we pass a function to `NewConnectionMaker`, stash that in a new `ConnectionMaker.normalisePeerAddr field`, and invoke that wherever you currently have `cm.ourself.Router.NormalisePeerAddr`.

abligh:
I think I've done that, and it works here, but it's now failing the coverage test. I am afraid the reasons it's failing (but wasn't before) are a bit opaque to me.

rade:
don't worry about the coverage "failure".

router/connection_maker.go
```diff
@@ -14,12 +14,15 @@ const (
 	MaxInterval     = 10 * time.Minute
 )
 
+type PeerNormaliser func(string) string
```
rade: no need to define a type for this.

router/connection_maker.go
```diff
@@ -14,12 +14,15 @@ const (
...
+
 type ConnectionMaker struct {
 	ourself        *LocalPeer
 	peers          *Peers
 	targets        map[string]*Target
 	cmdLineAddress map[string]struct{}
 	actionChan     chan<- ConnectionMakerAction
+	peerNormaliser PeerNormaliser
```
rade: don't like the name. `normalisePeerAddr` is the best I could come up with. Yes, it's a bit long.

rade: also, that field should come right after `peers`, following the order of arguments to `NewConnectionMaker`.

rade:
LGTM!

abligh:
Fixed

PR #541, opened 2015-04-10
Support using a unix domain socket for the HTTP control interface.

abligh:
When running multiple weaver daemons it is convenient for the HTTP
control interface to bind to a UNIX domain socket, to avoid the
need to allocate an additional unique port. This commit adds the
ability to specify

    httpaddr="/path/to/socket"

on the command line.

Example of use:

    $ weaver/weaver -name="00:11:22:33:44:55" -httpaddr="/tmp/weaversock"

then (`nc` which supports UNIX domain sockets required):

    $ nc -U /tmp/weaversock
    GET /status HTTP/1.1

    HTTP/1.1 200 OK
    Date: Fri, 10 Apr 2015 15:02:03 GMT
    Content-Length: 301
    Content-Type: text/plain; charset=utf-8

    weave router git-c13980af4bd3
    Encryption off
    Our name is 00:11:22:33:44:55(nimrod-ubuntu)
    Sniffing traffic on <nil>
    MACs:
    Peers:
    Peer 00:11:22:33:44:55(nimrod-ubuntu) (v0) (UID 3416334196290595526)
    Routes:
    unicast:
    00:11:22:33:44:55 -> 00:00:00:00:00:00
    broadcast:
    00:11:22:33:44:55 -> []
    Reconnects:

### DIFF
--- a/weaver/main.go
+++ b/weaver/main.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"strings"
 	"syscall"
 )
 
@@ -60,7 +61,7 @@ func main() {
 	flag.IntVar(&connLimit, "connlimit", 30, "connection limit (0 for unlimited)")
 	flag.IntVar(&bufSz, "bufsz", 8, "capture buffer size in MB")
 	flag.IntVar(&port, "port", weave.Port, "router port")
-	flag.StringVar(&httpAddr, "httpaddr", fmt.Sprintf(":%d", weave.HTTPPort), "address to bind HTTP interface to (disabled if blank)")
+	flag.StringVar(&httpAddr, "httpaddr", fmt.Sprintf(":%d", weave.HTTPPort), "address to bind HTTP interface to (disabled if blank, , absolute path indicates unix domain socket)")
 	flag.Parse()
 	peers = flag.Args()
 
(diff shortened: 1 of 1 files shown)


rade:
I'm not sold on the rationale here. Given that one will have to allocate unique weave ports, what's the big deal with also allocating unique http ports?

abligh:
In a large cluster, though I'm only likely to have a handful of weave processes per cluster (for which read *'domain enclosing all weave networks'*), life is significantly easier (as you yourself have pointed out) if the weave port is the same across the entire weave network, and (thus) that the weave port is unique to that network. That limits the total number of weave networks per cluster to the number of available ports. However, if I need a port for the control interface too, that's half as many ports. Weave isn't the only user of ports. If I assign a range of (e.g.) 5,000 ports to weave, not having this means I am limited to 2,500 weave networks per cluster (across possibly thousands of nodes), rather than 5,000. Note I am talking about the limit of the total number of weave networks; each weave network itself will probably only have a handful of participating devices, and the number of weave networks running on each node (on different ports) is also likely to be small.

My immediate fix for that issue is simply to not bind the control port, as I don't need to communicate with weaver for run-time purposes - the previous commit fixed this. However, for debugging purposes, it would be nice to bind it to a socket as well, so I can get the status out. A trivial script around `nc` will do this. So really for me, this just adds back the ability to do debugging when using lots and lots of ports. However, I'm sure other users might actually want to use the control port for other purposes (add/delete peer etc).

rade:
ok.

Onto the code...

1. I suggest we have a common code path of `net.Listen` followed by `http.Serve` for both the domain and network socket case. i.e. the distinction should be taken care of in the code prior to that.

2. the ability to specify a domain socket should be mentioned in the usage.

abligh:
Done, but note the help text now has a fantastically long line.

abligh:
thanks

On 10 Apr 2015, at 20:39, rade <notifications@github.com> wrote:

> Merged #541.
> 
> �
> Reply to this email directly or view it on GitHub.
> 

-- 
Alex Bligh

PR #542, opened 2015-04-10
Only check kernel version and ip command in local mode

errordeveloper:


weave
```diff
@@ -545,15 +529,33 @@ ask_version() {
     [ -n "$DOCKERIMAGE" ] && docker run --rm $DOCKERIMAGE --version
 }
 
-if ! command_exists ip ; then
-    echo "ERROR: ip utility is missing. Please install it." >&2
-    exit 1
-fi
+if [ "$LOCAL" = "--local" ]; then
```
rade: this should just be `-n "$LOCAL"`

errordeveloper: I was just adhering to prevalent style...

rade: fair enough. I've [just](https://github.com/weaveworks/weave/commit/85971193a8aec9f62ca3723e2acb7362f1e473df) changed the prevalent style :)

PR #544, opened 2015-04-10
Make sure FieldValidator methods act on the object we pass in.

### LINKED ISSUE #543: Handshake checks don't work
The methods on `FieldValidator` in handshake.go take a value rather than a pointer, so the update of `err` acts on a copy which is then discarded.

bboreham:
Fixes #543

### DIFF
--- a/router/handshake_test.go
+++ b/router/handshake_test.go
@@ -1,0 +1,29 @@
+package router
+
+import (
+	wt "github.com/weaveworks/weave/testing"
+	"testing"
+)
+
+func TestFieldValidator(t *testing.T) {
+	testMap := map[string]string{
+		"Protocol":        Protocol,
+		"ProtocolVersion": "123",
+	}
+	fv := NewFieldValidator(testMap)
+	err := fv.CheckEqual("ProtocolVersion", "123")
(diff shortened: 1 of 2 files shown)


router/handshake_test.go
```diff
@@ -0,0 +1,29 @@
...
+	err = fv.CheckEqual("ProtocolVersion", "123")
+	wt.AssertFalse(t, err == nil || fv.Err() == nil, "Previous error should be retained")
+
+	fv = NewFieldValidator(testMap)
+	val, err := fv.Value("Protocol")
+	wt.AssertNoErr(t, err)
+	wt.AssertEqualString(t, val, Protocol, "Protocol")
+	val, err = fv.Value("Protocolx")
+	wt.AssertFalse(t, err == nil || fv.Err() == nil, "Expected error from incorrect string check")
+}
```
rade: All the same tests that were performed for `CheckEqual` should also be performed here.

Furthermore, we should also test that CheckValue isn't stomping on CheckEqual and vice versa.

rade:
somehow my manual merge + change + rebase seems to have bypassed this PR. -> closing.

PR #548, opened 2015-04-12
lazy address resolution

### LINKED ISSUE #112: resolve peer hostnames at time of connect
If weave is launched with one or more peers given as names rather than IP addresses, then the simple behaviour added by #111 is to resolve those names once at startup and then use that IP address from then on.

Better to re-resolve the name each time we want to connect, so we get an up-to-date IP address.

Needs some modifications to the structure in ConnectionMaker to remember the resolved IP address while that connection attempt is in flight.

This behaviour would also open the possibility that we have an address which does not resolve - this is perhaps a transitory condition so we should re-try that address later.

rade:
resolve command line peer addresses at the time of connect rather than once at the beginning.

Closes #112.


rade:
Closing because of https://github.com/weaveworks/weave/issues/112#issuecomment-92097447

PR #551, opened 2015-04-12
make weave port configurable via WEAVE_PORT

rade:
This is a follow on of #534.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,10 +92,11 @@ We repeat similar steps on `$HOST2`...
 
 The only difference, apart from the choice of IP address for the
 application container, is that we tell our weave that it should peer
-with the weave on `$HOST1` (specified as the IP address or hostname, and
-optional `:port`, by which `$HOST2` can reach it). NB: if there is a
-firewall between `$HOST1` and `$HOST2`, you must open port 6783 for TCP
-and UDP.
+with the weave on `$HOST1` (specified as the IP address or hostname,
+and optional `:port`, by which `$HOST2` can reach it). NB: if there is
+a firewall between `$HOST1` and `$HOST2`, you must open the weave port
+(6783 by default; this can be overriden by setting `WEAVE_PORT`) for
+TCP and UDP.
 
 Note that we could instead have told the weave on `$HOST1` to connect to
 `$HOST2`, or told both about each other. Order does not matter here;
--- a/site/features.md
+++ b/site/features.md
@@ -316,7 +316,10 @@ data centres. So, for example, one could run an application consisting
 of containers on GCE, EC2 and in local data centres.
 
 To enable this, the network must be configured to permit TCP and UDP
-connections to port 6783 of the docker hosts.
+connections to the weave port of the docker hosts. The weave port
+defaults to 6783. This can be overriden by setting `WEAVE_PORT`, but
+it is highly recommended that all peers in a weave network are given
+the same port setting.
 
 ### <a name="multi-hop-routing"></a>Multi-hop routing
 
--- a/weave
+++ b/weave
@@ -59,6 +59,7 @@ exec_remote() {
         -e WEAVE_DOCKER_ARGS \
         -e WEAVEDNS_DOCKER_ARGS \
         -e WEAVE_PASSWORD \
+        -e WEAVE_PORT \
         -e WEAVE_CONTAINER_NAME \
         -e DOCKER_BRIDGE \
         -e PROCFS=/hostproc --privileged --net=host \
@@ -86,7 +87,7 @@ DNS_CONTAINER_NAME=weavedns
 BRIDGE=weave
 CONTAINER_IFNAME=ethwe
 MTU=65535
-PORT=6783
+PORT=${WEAVE_PORT:-6783}
 HTTP_PORT=6784
 DNS_HTTP_PORT=6785
 
@@ -599,6 +600,7 @@ case "$COMMAND" in
             shift 1
         fi
         NICKNAME=$(hostname)
+        CONTAINER_PORT=$PORT
         while true; do
             case "$1" in
                 -password)
@@ -612,6 +614,11 @@ case "$COMMAND" in
                     NICKNAME="$2"
                     shift 2
                     ;;
+                -port)
+                    [ $# -gt 1 ] || usage
+                    CONTAINER_PORT="$2"
+                    shift 2
+                    ;;
                 *)
                     break
                     ;;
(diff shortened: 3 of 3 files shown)


rade:
@dpw review please

dpw:
Other than the comment about `$PORT`, works for me.

rade:
The general rule I've tried to follow is that `WEAVE_` vars are for thing we expected to be supplied by a user, and that they seed the non-prefixed variants. We could change that to use the same var for the latter, but it should be applied to all such vars in the weave script, not just `PORT`. Hence should go into a separate PR.

dpw:
> The general rule I've tried to follow is that WEAVE_ vars are for thing we expected to be supplied by a user, and that they seed the non-prefixed variants. We could change that to use the same var for the latter, but it should be applied to all such vars in the weave script, not just PORT. Hence should go into a separate PR.

Ok.  The `WEAVE_PORT` idea just seemed convenient to me.  I'd be happy with renaming `PORT` to something else non-generic. But the `WEAVE_PORT` vs `PORT` thing just seems too confusing to ok as is.

rade:
any more confusing than `CONTAINER_NAME`?

dpw:
The confusingness of `CONTAINER_NAME` does not seem pertinent to or exacerbated by this PR. You can fix that if it makes you happy though.

rade:
It is pertinent because if we go with `WEAVE_PORT` throughout then that inconsistent with how other vars are treated. We could go with another non-weave name (suggestions?), but tbh I do not understand what you find confusing about `PORT`. And it's always been called `PORT`; this is not something introduced in this PR.

dpw:
I'm not saying it has to be `WEAVE_PORT` throughout.  That does seem like a
good solution to me, but it's not the only solution to the confusion I see,
so you are welcome to solve it a different way.

The confusion is because we now have two variables, `WEAVE_PORT` and
`PORT`, which both carry the, um, weave port, but with a subtly different
emphasis to their use. This worsens a the preexisting issue: While `PORT`
might have been a fine variable name when there was only one port number in
the script, it is not so good when there is also `HTTP_PORT` etc.  But even
if `HTTP_PORT` did not exsit, the lack of a hint at the distinction between
`PORT` and `WEAVE_PORT` would still be a problem.

I don't have a good idea for a name (hence the attraction of re-using
`WEAVE_PORT`).  But sometimes when I have a difficulty naming something, I
find a useful approach is to start by writing a comment introducing its
purpose, and then to boil that comment down until it can be captured in the
name, so that the comment can be dispensed with.

rade:
PROTOCOL_PORT? PEER_PORT?

awh:
You appear to be having trouble getting this bikeshed painted ;) GOSSIP_PORT?

rade:
> GOSSIP_PORT

The port is used for more than gossip.

dpw:
> PROTOCOL_PORT? PEER_PORT? 

These do not seem to hint at the purpose of the variable as it contrasts with the purpose of `WEAVE_PORT`.

rade:
I will just leave it as PORT then.

dpw:
I'm not raising this point frivolously, I genuinely think the distinction between WEAVE_PORT and PORT is not obvious. And that given state of the weave script, we need an abundance of caution when introducing changes that tend to make it worse.

rade:
I don't think it's a frivolous point. But I cannot come up with a better name. It's not worth holding this up for that.

dpw:
I was clear what I thought the best way to address this was up front (I.e. conflate WEAVE_PORT and PORT). You didn't like that suggestion, because some other thing should be changed for consistency. That might be a valid point, but it's not one I raised, and not one I would have held things up for. And you could have made the wider change for consistency if you believed that proper - it would not have been time consuming.

As for changing the name of PORT, I don't think an earnest attempt to find a better

dpw:
(Cont) name would take more than five or ten minutes.

In short, I don't think I am the one holding this up, except by providing feedback, and guidance on what would or would not satisfy that feedback. If you want someone to rubber stamp your PRs without giving them careful consideration, find someone else, you know that's not my style.

rade:
> I don't think I am the one holding this up

I didn't suggest you were. To be clear, I am not asking you to merge this PR.

>  If you want someone to rubber stamp your PRs without giving them careful consideration, find someone else, you know that's not my style.

I don't mind the careful consideration. Quite the opposite. But sometimes there will be disagreement about a particular piece of feedback. And sometimes it's not worth holding up a PR until that is resolved. 

I'll let this one sit until tomorrow. If somebody comes up with a mutually acceptable name for the var then I'll make that change. Otherwise I'll merge this as is.

dpw:
> I don't mind the careful consideration. Quite the opposite. But sometimes there will be disagreement about a particular piece of feedback. And sometimes it's not worth holding up a PR until that is resolved.

Ah, but who gets to make the call - there's the rub.  If a developer can unilaterally override, then that leads to a low standard of review (because all developers are reluctant to change work they have invested in, or to do further work they regard as unnecessary; and because reviewers have a disincentive to be diligent if their diligence may be disregarded at a whim).

> I'll let this one sit until tomorrow. If somebody comes up with a mutually acceptable name for the var then I'll make that change. Otherwise I'll merge this as is.

We both know how that will go, so this seems like a pretense.

rade:
I disagree, but let's not have that meta discussion in this issue, please.

awh:
The flag help for `-port` in `weaver/main.go` is "router port". `ROUTER_PORT`?

dpw:
My opinion on this one may be by the by, but if PORT became ROUTER_PORT, then WEAVE_PORT could become WEAVE_ROUTER_PORT on exactly the same basis, and we'd be more or less back where we started  I am not suggesting renaming for the sake of renaming, with a name that is merely different from WEAVE_PORT and longer than PORT, but rather I'd like a name that accounts for the specific purpose of that variable.  Yes, it's a port number, yes it's the router port number, but why does it exist in addition to WEAVE_PORT?

awh:
I think the problem here stems from the fact that PORT is used for several things - e.g. both the control plane traffic over TCP and the forwarding path over UDP, both of which are distinct protocols proprietary to the weave router. As a consequence it's quite hard to give it a pithy name! Actually I agree with your suggestion to do `WEAVE_PORT=${WEAVE_PORT:-6783}`, and would actually extend it to the other variables too, with some further changes to the other port names to make them more descriptive:

````
WEAVE_CONTAINER_NAME=${WEAVE_CONTAINER_NAME:-weave}
WEAVEDNS_CONTAINER_NAME=${WEAVEDNS_CONTAINER_NAME:-weavedns}
WEAVE_BRIDGE=weave
WEAVE_CONTAINER_IFNAME=ethwe
WEAVE_MTU=65535
WEAVE_ROUTER_PORT=${WEAVE_ROUTER_PORT:-6783}
WEAVE_REST_PORT=${WEAVE_REST_PORT:-6784}
WEAVEDNS_REST_PORT=${WEAVEDNS_REST_PORT:-6785}
````

rade:
@awh please open a separate issue if you want to go on a var renaming spree.

As for your latest suggestion. See https://github.com/weaveworks/weave/pull/551#issuecomment-92308365... The `WEAVE_` prefix indicates that the var can be supplied by the user. Some of the vars in your list aren't configurable that way. And neither should they be. That also goes for the 'rest' ports - these are only accessed via the container IP and thus cannot clash with anything).

PR #553, opened 2015-04-13
Makefile cleanups

dpw:


rade:
I have cherry-picked the 2nd change (https://github.com/dpw/weave/commit/433d406d83ab9a3489366b7e68c71b4005e0878d) onto master.

The first one LGTM, but it may be a good idea to ask a Mac user to test it.

dpw:
> The first one LGTM, but it may be a good idea to ask a Mac user to test it.

It's in OSX: https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/mktemp.1.html

Could you nominate a mac user? You know their priorities better than me.

rade:
I had looked up the man pages too. I suppose there is no reason not to trust them. And in any case, if this commit really does break something we'll find out pretty quickly.

PR #554, opened 2015-04-13
Pull forward the declaration of the RouterConfig in main()

dpw:
This eliminates a few variables, and the naming awkardness where we
have corresponding but distinct flag and router config variables.

### DIFF
--- a/weaver/main.go
+++ b/weaver/main.go
@@ -35,32 +35,29 @@ func main() {
 	runtime.GOMAXPROCS(procs)
 
 	var (
+		config      weave.RouterConfig
 		justVersion bool
-		port        int
 		ifaceName   string
 		routerName  string
-		nickName    string
 		password    string
 		wait        int
 		debug       bool
 		prof        string
 		peers       []string
-		connLimit   int
-		bufSz       int
 		httpAddr    string
 	)
 
 	flag.BoolVar(&justVersion, "version", false, "print version and exit")
-	flag.IntVar(&port, "port", weave.Port, "router port")
+	flag.IntVar(&config.Port, "port", weave.Port, "router port")
 	flag.StringVar(&ifaceName, "iface", "", "name of interface to capture/inject from (disabled if blank)")
 	flag.StringVar(&routerName, "name", "", "name of router (defaults to MAC of interface)")
-	flag.StringVar(&nickName, "nickname", "", "nickname of peer (defaults to hostname)")
+	flag.StringVar(&config.NickName, "nickname", "", "nickname of peer (defaults to hostname)")
 	flag.StringVar(&password, "password", "", "network password")
 	flag.IntVar(&wait, "wait", 0, "number of seconds to wait for interface to be created and come up (0 = don't wait)")
 	flag.BoolVar(&debug, "debug", false, "enable debug logging")
 	flag.StringVar(&prof, "profile", "", "enable profiling and write profiles to given path")
-	flag.IntVar(&connLimit, "connlimit", 30, "connection limit (0 for unlimited)")
-	flag.IntVar(&bufSz, "bufsz", 8, "capture buffer size in MB")
+	flag.IntVar(&config.ConnLimit, "connlimit", 30, "connection limit (0 for unlimited)")
+	flag.IntVar(&config.BufSz, "bufsz", 8, "capture buffer size in MB")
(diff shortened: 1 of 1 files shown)


weaver/main.go
```diff
@@ -35,33 +35,30 @@ func main() {
...
 	flag.StringVar(&password, "password", "", "network password")
 	flag.IntVar(&wait, "wait", 0, "number of seconds to wait for interface to be created and come up (0 = don't wait)")
 	flag.BoolVar(&debug, "debug", false, "enable debug logging")
 	flag.StringVar(&prof, "profile", "", "enable profiling and write profiles to given path")
-	flag.IntVar(&connLimit, "connlimit", 30, "connection limit (0 for unlimited)")
-	flag.IntVar(&bufSz, "bufsz", 8, "capture buffer size in MB")
-	flag.StringVar(&httpAddr, "httpaddr", fmt.Sprintf(":%d", weave.HTTPPort), "address to bind HTTP interface to (disabled if blank, absolute path indicates unix domain socket)")
+	flag.IntVar(&config.ConnLimit, "connlimit", 30, "connection limit (0 for unlimited)")
+	flag.IntVar(&config.BufSz, "bufsz", 8, "capture buffer size in MB")
+	flag.StringVar(&httpAddr, "httpaddr", fmt.Sprintf(":%d", weave.HTTPPort), "address to bind HTTP interface to (disabled if blank, , absolute path indicates unix domain socket)")
```
rade: This re-introduces a stray comma, which I recently removed from 'master'.

dpw: Ok, will respin

dpw:
Comma gone again.

(I notice that this page currently shows "rade commented on an outdated diff", confirming my suspicion that github can *sometimes* track PR comments across rebases.)

rade:
The treatment of `BufSz` is rather suspect. It would fail a type check in a type system that supported units, i.e. we the type of `RouterConfig.BufSz` is "bytes", yet we read a "MBytes" value into it (and fix it up later). So I reckon we should leave the intermediate var in place for that one.

Overall I find his change makes the code less regular and makes it harder to figure out what the router config contains, since it gets created bit by bit. Is that outweighed by the 11 lines saved?

dpw:
> The treatment of BufSz is rather suspect. It would fail a type check in a type system that supported units, i.e. the type of RouterConfig.BufSz is "bytes", yet we read a "MBytes" value into it (and fix it up later). So I reckon we should leave the intermediate var in place for that one.

Go is rather far from having such a type system.

> Overall I find this change makes the code less regular and makes it harder to figure out what the router config contains, since it gets created bit by bit. Is that outweighed by the 11 lines saved?

The elimination of mess such as name vs ourName and password vs pwSlice is worth it on its own. And 11 lines is 11 lines.

rade:
> Go is rather far from having such a type system.

Right, but my point stands. There is a bad smell to setting a field value to something that is invalid and then mutating it to be valid.

> The elimination of mess such as name vs ourName and password vs pwSlice is worth it on its own.

I don't actually consider that to be messy.

dpw:
> > Go is rather far from having such a type system.
> 
> Right, but my point stands. There is a bad smell to setting a field value to something that is invalid and then mutating it to be valid.

Reintroduced with a clearer name.

rade:
> Reintroduced with a clearer name.

Cheers. As for the overall change. I am still not convinced. Perhaps ask for a 3rd opinion?

dpw:
> Perhaps ask for a 3rd opinion?

@bboreham, will you arbitrate?

bboreham:
I like this change.  Considering the way that `RouterConfig` gets used inside router.go, I might pull `Name` and `NickName` out and make the remaining `RouterConfig` a member of `Router`, saving another 12 lines.

dpw:
> Considering the way that RouterConfig gets used inside router.go, I might pull Name and NickName out and make the remaining RouterConfig a member of Router, saving another 12 lines.

I did think about this, and as you mention the wrinkle is Name andNickName. I anticipated resistance to moving them out of RouterConfig to separate NewRouter parameters on the basis that it would revealing implementation details.

It would be really nice if we could introduce a RouterCommon that had the common parts of Router and RouterConfig, and make it an anonymous field of both.  But this uglifies the corresponding struct literals, because anonymous fields are not magical there.

PR #555, opened 2015-04-14
more discriminate connectivity

### LINKED ISSUE #478: Indefinite reconnects occur if two peers are launched referencing each other
See @nordbergm's comment in issue #341 for original report. If two peers are launched simultaneously referencing each other by IP, they will establish connectivity (confirmed with `weave status` on each peer) however one of them will *also* be trying to reconnect to the other periodically, an attempt which fails as there is already a connection between the peers.

rade:
If a peer was specified on the command line w/o a port, then do not connect to it if we have inbound connections from that IP.

This eliminates unnecessary (and ultimately failing0 connection attempts in the common scenario of several peers having been told about each other on their command lines.

Closes #478.

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -15,12 +15,12 @@ const (
 )
 
 type ConnectionMaker struct {
-	ourself           *LocalPeer
-	peers             *Peers
-	normalisePeerAddr func(string) string
-	targets           map[string]*Target
-	cmdLinePeers      map[string]string
-	actionChan        chan<- ConnectionMakerAction
+	ourself      *LocalPeer
+	peers        *Peers
+	port         int
+	targets      map[string]*Target
+	cmdLinePeers map[string]*net.TCPAddr
+	actionChan   chan<- ConnectionMakerAction
 }
 
 // Information about an address where we may find a peer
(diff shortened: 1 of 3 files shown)


awh:
Looks good to me, although there is a typo in the commit message...

rade:
Fixed the typo. And committed a change with a different internal representation for command line supplied addresses. I think it's marginally cleaner and clearer. If you agree I'll happily squash those two commits into one.

The last commit is related tidying up in other parts of the code base as a consequence of the 2nd commit. Technically it could be split into two - removal of the NormalisePeerAddr call site and removal of NormalisePeerAddr. The former has no dependency on the other changes. The latter would depend on the lot.

router/connection_maker.go
```diff
@@ -15,12 +15,12 @@ const (
...
-	ourself           *LocalPeer
-	peers             *Peers
-	normalisePeerAddr func(string) string
-	targets           map[string]*Target
-	cmdLinePeers      map[string]string
-	actionChan        chan<- ConnectionMakerAction
+	ourself      *LocalPeer
+	peers        *Peers
+	port         int
+	targets      map[string]*Target
```
awh: Not something for this PR, but I feel that the readability of the entire ConnectionMaker would be greatly improved if all these different `map[string]...` used appropriate type aliases for the key instead of a plain string e.g.

````
type TCPAddrString string
...
target map[TCPAddrString]*Target
````

thoughts?

File: router/connection_maker.go [hunk 2]
```diff
@@ -49,16 +49,19 @@ func (cm *ConnectionMaker) Start() {
 }
 
 func (cm *ConnectionMaker) InitiateConnection(peer string) error {
-	addr, err := net.ResolveTCPAddr("tcp4", cm.normalisePeerAddr(peer))
+	host, port, err := net.SplitHostPort(peer)
+	if err != nil {
+		host = peer
+		port = "0" // we use that as an indication that "no port was supplied"
```
awh: Oh for `Maybe Port` :)

router/connection_maker.go
rade: maybe. worth seeing what that looks like. but not just in ConnectionMaker.

awh:
Yep agree the use of `net.TCPAddr` is an improvement - squash as you see fit and I'll merge.

PR #556, opened 2015-04-14
make the Router includer the RouterConfig

rade:
rather than copying the bulk of it


rade:
@bboreham @dpw

dpw:
(I'm on a ports-80-and-443-only wifi network.  So if you want this dealt with today, Bryan will have to do it.)

PR #557, opened 2015-04-14
Ensure `make` stops on error from `go test`

bboreham:
Changed command separator from `;` to `&&`.

In the absence of this (or something like `set -e`), `make` will carry on and report the exit status from the next command, which is generally 0 i.e. success.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,14 +69,18 @@ $(DOCKER_DISTRIB):
 
 tests:
 	echo "mode: count" > profile.cov
+	fail=0 ;                                                                              \
 	for dir in $$(find . -type f -name '*_test.go' | xargs -n1 dirname | sort -u); do     \
-	    output=$$(mktemp cover.XXXXXXXXXX);                                               \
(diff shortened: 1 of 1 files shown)


rade:
This does have the disadvantage of stopping on the first failing test set, doesn't it? @dpw

bboreham:
I just checked, and actually it does the opposite - if an earlier test fails but later one succeeds it will return 0 for the whole lot.  Back to the drawing board.

dpw:
> This does have the disadvantage of stopping on the first failing test set, doesn't it? @dpw

Looks like it,

It should probably be restructured so that there is an `tests` empty target that depends on some per-dir test runner targets.  It can probably be done with a static pattern.

bboreham:
We had two much simpler variants before and after https://github.com/weaveworks/weave/commit/592be90904288c7e6be08e2c27fea3a5c3ff1a96

The multi-line script came in with #505, to generate a consolidated coverage report.

bboreham:
Changed to accumulate the error in a variable `fail`

rade:
Fixed the cosmetic issue. rebased. merged.

PR #559, opened 2015-04-14
Support for multiple responses in mDNS

inercia:
This pull request is the stepping stone for full support of multiple responses in WeaveDNS. It focuses in the mDNS client and server, where it provides

* support for "insistent" `Lookups` (maybe not the best name): mDNS queries that try to obtain as many responses as possible for up to 500ms (and, in consequence, they block for that period of time). This will be used in the next PR by a "updater" process that will _refresh_ the ZoneDb information by sending periodic queries to peers...
* support for multiple responses from all the `Lookup*()` methods
* a common `ZoneRecord` interface for many structs that contain names and IPs (used in `Lookup` methods, but also in the future database records)
* support for multiple records when building packets with `make*()` methods
* improved unit tests for mDNS

### DIFF
--- a/nameserver/addrs.go
+++ b/nameserver/addrs.go
@@ -87,7 +87,7 @@ func addrToIPv4(addr string) (IPv4, error) {
 // Parse a reverse address (eg, "11.12.13.10.in-addr.arpa.") and return the corresponding IPv4 (eg, "10.13.12.11")
 func raddrToIPv4(addr string) (IPv4, error) {
 	l := len(addr)
-	if l < rdnsDomainLen + 1 {
+	if l < rdnsDomainLen+1 {
 		return IPv4{}, &net.ParseError{"too short reverse IP address", addr}
 	}
 
--- a/nameserver/addrs.go
+++ b/nameserver/addrs.go
@@ -1,0 +1,107 @@
+package nameserver
+
+import (
+	"bytes"
+	"fmt"
+	"net"
+)
+
+// A zone record
+// Note: we will try to keep all the zone records information is concentrated here. Maybe not all queries will
+//       use things like priorities or weights, but we do not want to create a hierarchy for dealing with all
+//       possible queries...
+type ZoneRecord interface {
+	Name() string  // The name for this IP
+	IP() net.IP    // The IP (v4) address
+	Priority() int // The priority
+	Weight() int   // The weight
+}
+
+type ZoneLookup interface {
+	// Lookup for a name
+	LookupName(name string) ([]ZoneRecord, error)
+	// Lookup for an address
+	LookupInaddr(inaddr string) ([]ZoneRecord, error)
+}
+
+type ZoneObserver interface {
+	// Observe anything that affects a particular name in the zone
+	ObserveName(name string, observer ZoneRecordObserver) error
+	// Observe anything that affects a particular IP in the zone
+	ObserveInaddr(inaddr string, observer ZoneRecordObserver) error
+}
+
+type ZoneRecordObserver func(ZoneRecord)
+
+/////////////////////////////////////////////////////////////
+
+// A basic record struct (satisfying the ZoneRecord interface)
+type Record struct {
+	name     string
+	ip       net.IP
+	priority int
+	weight   int
+}
+
+func (r Record) Name() string  { return r.name }
+func (r Record) IP() net.IP    { return r.ip }
+func (r Record) Priority() int { return r.priority }
+func (r Record) Weight() int   { return r.weight }
+
+func (i Record) String() string {
+	var buf bytes.Buffer
+	if len(i.Name()) > 0 {
+		buf.WriteString(fmt.Sprintf("%s", i.Name()))
+	}
+	if !i.IP().IsUnspecified() {
+		buf.WriteString(fmt.Sprintf("[%s]", i.IP()))
+	}
+	if i.Priority() > 0 {
+		buf.WriteString(fmt.Sprintf("/P:%d", i.Priority()))
+	}
+	if i.Weight() > 0 {
+		buf.WriteString(fmt.Sprintf("/W:%d", i.Weight()))
+	}
+
+	return buf.String()
+}
+
+/////////////////////////////////////////////////////////////
+
+// simple IPv4 type that can be used as a key in a map (in contrast with net.IP), used for sets of IPs, etc...
+type IPv4 [4]byte
+
+func (ip IPv4) toNetIP() net.IP { return net.IPv4(ip[0], ip[1], ip[2], ip[3]) }
+func (ip IPv4) String() string  { return ip.toNetIP().String() }
+func ipToIPv4(nip net.IP) IPv4  { ip := nip.To4(); return IPv4([4]byte{ip[0], ip[1], ip[2], ip[3]}) }
+
+// Parse an address (eg, "10.13.12.11") and return the corresponding IPv4 (eg, "10.13.12.11")
+func addrToIPv4(addr string) (IPv4, error) {
+	ip := net.ParseIP(addr)
+	if ip == nil {
+		return IPv4{}, &net.ParseError{"IP address", addr}
+	}
+	return ipToIPv4(ip), nil
+}
+
+// Parse a reverse address (eg, "11.12.13.10.in-addr.arpa.") and return the corresponding IPv4 (eg, "10.13.12.11")
+func raddrToIPv4(addr string) (IPv4, error) {
+	l := len(addr)
+	if l < rdnsDomainLen + 1 {
+		return IPv4{}, &net.ParseError{"too short reverse IP address", addr}
+	}
+
+	suffixLen := l - rdnsDomainLen
+	suffix := addr[suffixLen+1:]
+	if suffix != RDNSDomain {
+		return IPv4{}, &net.ParseError{"suffix of reverse IP address", addr}
+	}
+	ipStr := addr[:suffixLen]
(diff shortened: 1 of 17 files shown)


nameserver/addrs.go
```diff
@@ -0,0 +1,99 @@
...
+	if i.Weight() > 0 {
+		buf.WriteString(fmt.Sprintf("/W:%d", i.Weight()))
+	}
+
+	return buf.String()
+}
+
+/////////////////////////////////////////////////////////////
+
+// simple IPv4 type that can be used as a key in a map (in contrast with net.IP), used for sets of IPs, etc...
```
inercia: This type, `IPv4`, will not be really used until I introduce the new `ZoneDb` internal database format...

nameserver/mdns_client.go
```diff
@@ -64,17 +84,21 @@ type mDNSQueryInfo struct {
 }
 
 func NewMDNSClient() (*MDNSClient, error) {
-	if conn, err := net.ListenUDP("udp4", &net.UDPAddr{IP: net.IPv4zero, Port: 0}); err != nil {
```
inercia: It did not make sense to listen for anything here. Binding to ports should be delayed until we `Start()` the server...

nameserver/mdns_client.go
```diff
@@ -93,7 +117,13 @@ func (c *MDNSClient) Start(ifi *net.Interface) error {
 	actionChan := make(chan MDNSAction, MailboxSize)
 	c.actionChan = actionChan
 	go c.actorLoop(actionChan)
+	return nil
+}
 
+func (c *MDNSClient) Stop() error {
```
inercia: We didn't have a proper shutdown mechanism. And it could be improved by waiting for the actor goroutine...

nameserver/mdns_client.go
```diff
@@ -64,17 +84,21 @@ type mDNSQueryInfo struct {
...
+		addr:     ipv4Addr,
+		running:  false,
+		inflight: make(map[string]*inflightQuery)}, nil
 }
 
-func (c *MDNSClient) Start(ifi *net.Interface) error {
+func (c *MDNSClient) Start(ifi *net.Interface) (err error) {
+	if c.running {
+		return nil
+	}
```
inercia: We want to skip multiple initializations on the same mDNS client. This is specially useful when testing, as we can use the same client several times...

nameserver/addrs.go
```diff
@@ -0,0 +1,99 @@
...
+type IPv4 [4]byte
+
+func (ip IPv4) toNetIP() net.IP { return net.IPv4(ip[0], ip[1], ip[2], ip[3]) }
+func (ip IPv4) String() string  { return ip.toNetIP().String() }
+func ipToIPv4(nip net.IP) IPv4  { ip := nip.To4(); return IPv4([4]byte{ip[0], ip[1], ip[2], ip[3]}) }
+
+// Parse an address (eg, "10.13.12.11") and return the corresponding IPv4 (eg, "10.13.12.11")
+func addrToIPv4(addr string) IPv4 {
```
awh: Should this (and `raddrToIPv4`)not return `(IPv4, error)`? I can see that `net.ParseIP` returns `nil`, but that is broken IMO, and clearly when it came to implement `net.ParseCIDR` they realised their mistake with `net.ParseIP` and introduced a `net.ParseError` type...

nameserver/addrs.go
```diff
@@ -0,0 +1,99 @@
...
+	ip := net.ParseIP(addr)
+	if ip == nil {
+		return IPv4{}
+	}
+	return ipToIPv4(ip)
+}
+
+// Parse a reverse address (eg, "11.12.13.10.in-addr.arpa.") and return the corresponding IPv4 (eg, "10.13.12.11")
+func raddrToIPv4(addr string) IPv4 {
+	ipStr := addr[:len(addr)-rdnsDomainLen]
```
awh: It would be better if we actually checked that the expected suffix is present rather than just blindly chopping off `rdnsDomainLen` characters...

nameserver/addrs.go
inercia: You are right... I will change it.

nameserver/addrs.go
```diff
@@ -0,0 +1,99 @@
...
+	revIP := net.ParseIP(ipStr)
+	if revIP == nil {
+		Warning.Printf("Could not parse '%s' (%s)", addr, ipStr)
```
awh: As above - this message should go in a `net.ParseError`

nameserver/addrs.go
inercia: This was how we were doing it so far... I don't know if it is worth the trouble to check that, but it could be done...

nameserver/mdns_lookup.go
```diff
@@ -3,37 +3,69 @@ package nameserver
...
-		return resp, nil
+		Debug.Printf("[mdns] Got response name for %s-query about name '%s': name '%s', addr '%s'",
+			dns.TypeToString[qtype], name, resp.name, resp.addr)
+
+		if !insistent {
+			return []ZoneRecord{resp}, nil // non-background queries return the first response...
+		}
+
+		// add the response, avoiding duplicates
+		// TODO: replace this linear search by something better... (not a big deal as we don't expect long responses)
```
awh: Store responses as map keys?

inercia: I thought the same, but you cannot do that: they are not really `Comparable`...

nameserver/addrs.go
awh: I think if you're going to have a standalone library function that returns a `ParseError` then it should be rigorous - just needs
````
    if !HasSuffix(addr, RDNSDomain) {
        return nil, &ParseError{"Invalid domain suffix", addr}
    }
````
at the top...

**Comment by inercia** [comment]
@awh I have fixed the parsing mechanism for IP addresses as you suggested, and I have also added a unit test for that...

**Comment by awh** [comment]
It looks like a response containing multiple addresses is only permuted if the answer comes from the cache. Shouldn't `shuffleAnswers` be moved outside the cache so that it can treat lookups from the local zone and the mDNS client in the same way?

**Comment by awh** [comment]
In the case where `queryHandler` is asked to resolve a name which is a) not in the cache and b) has records in the `Zone` local to the a weaveDNS instance, it looks like we never go out to the rest of the weaveDNS instances? The presence of an entry locally short circuits the loop over the `ZoneLookup` array...

**Comment by inercia** [comment]
@awh That's right: the presence of an entry locally ends the lookup. We currently do not support multiple responses to clients, so once you have a valid answer to the query, you are done.

And regarding `shuffleAnswers`, I would prefer to delay that change, as we currently we do not really shuffle answers (we do not support multiple answers yet, so there is nothing to shuffle!)

File: nameserver/addrs.go [hunk 9]
```diff
@@ -0,0 +1,111 @@
...
+
+func (r Record) Name() string  { return r.name }
+func (r Record) IP() net.IP    { return r.ip }
+func (r Record) Priority() int { return r.priority }
+func (r Record) Weight() int   { return r.weight }
+
+func (i Record) String() string {
+	var buf bytes.Buffer
+	if len(i.Name()) > 0 {
+		buf.WriteString(fmt.Sprintf("%s", i.Name()))
```
awh: @rade pointed out that there was a concerted effort a while back eliminate `buf.WriteString(fmt.S*(...))` calls in favour of `fmt.F*(&buf,...)` (see commit 45740b1419508d442ec681893eb8f4c2aac93736) - could you update these in a similar vein?

awh:
> @awh That's right: the presence of an entry locally ends the lookup. We currently do not support multiple responses to clients, so once you have a valid answer to the query, you are done.

At first I thought you meant 'before this PR' by 'currently', but on closer reading I can see the PR widens all the interfaces to take a slice but still works with single values. Can you confirm my understanding that this PR doesn't change the user observed behaviour at all - it's just internal restructuring in support of future changes?

inercia:
@awh Exactly. This PR does not provide an user visible change. It only provides "insistent" queries at the mDNS layer, but they are not used at all in this PR...

awh:
OK in that case LGTM. One final comment - and I really can't stress this enough - separating out different logical changes within your PRs into individual commits makes things much easier on the reviewer, with the consequence that things will get reviewed quicker. Perhaps each of your bullet points could have been teased out into different commits, and there are other things hiding in this PR too like the reformatting of the logging and moving the port binding into the start methods. I don't know how familiar you are with `git rebase -i`, but I didn't appreciate until recently how powerful it is for restructuring the commit history of a branch after the fact - if you've not used it in anger I recommend taking a look. Also recently learned about `git add -p` for staging individual hunks in a file, which gives you a way of doing it as you go...

inercia:
Thanks @awh 

Regarding how to organize the PR, I understand that it can get a bit messy because of the mix of different logical changes. But, in fact I usually do a `git rebase -i` but for squashing all the changes in my branch. And I also use `git add -p`. In fact this PR is the result of squashing all the changes in one commit, then resetting and then `git add -p`'ing these changes... :)

I do that because my branches are in _pending_ state for too long, and that means that I have to rebase master almost everyday (for keeping them mergeable in case the PR reviewer decides it can be merged). This daily rebase is really painful when the PR is more than 1 or 2 commits, but it is trivial when all changes have been squashed into only one commit...

Please let me know of any way this workflow could be improved, but I've found this the only suitable solution for this project due to the long time it takes to have the PRs reviewed...

rade:
please move this meta discussion to email.

PR #563, opened 2015-04-15
IP Allocation Management

bboreham:
Implements the scheme described at https://github.com/weaveworks/weave/wiki/IP-allocation-design, with address ranges controlled by a CRDT gossiped between peers.

The design doc on the Wiki will be deleted when this is merged.

weave launch takes a new -iprange argument specifying the subnet to work within.

Commands updated to take advantage of IPAM: weave run, weave start, weave attach, weave expose, weave hide

### Future work

Currently there is no fall-back range if you don't specify one on the command-line; it would be better to pick from a set of possibilities (checking the range is currently unused before we start using it).

How to use IPAM for WeaveDNS?  It needs its own special subnet.

How should we add support for subnets in general?

We get a bit of noise in the weave logs from containers going down, now that the weave script is calling ethtool and curl via containers. It looks like "Container 1c5a03204b5eea3aeb6f43ece41c3f260a87329bfd89ded34ea3056826b59814 died"

~~If we hear that a container has died we should knock it out of pending?~~

Interrogate Docker to check container exists and is alive at the point we assign an IP

~~To lessen the chance of simultaneous election of two leaders at start-up, peers could wait for a bit to see if more peers arrive before running an election.  Also, we could look at the command-line parameters supplied to Weave, and wait longer in the case that we have been asked to connect to someone specific.~~

[It would be good to move PeerName out of package router into a more
central package so both ipam and router can depend on it.]

There is no specific logic to time-out requests such as "give me some space": the logic will be re-run and the request re-sent next time something else happens (e.g. it receives another request, or some
periodic gossip). This means we may send out requests more frequently than required, but this is innocuous and it keeps things simple.  It is possible for nothing to happen, e.g. if everyone else has
disconnected.  We could have a timer to re-consider things.

~~The operation to Split() one space into two, to give space to another Peer, is conceptually simple but the implementation is fiddly to maintain the various lists and limits within MutableSpace. Perhaps a different free-list implementation would make this easier.~~


bboreham:
@dpw this is the new PR, squashed and ready for review.

dpw:
Thanks.  But we seem to have some build failures?

rade:
There also seems to be a whole bunch of stuff missing - I don't see any changes to existing weave - the script, router main, etc.

rade:
18 files changed here vs 29 in #527.

bboreham:
Sorry about missing files.  Should be all there now.

ipam/testutils_test.go
```diff
@@ -0,0 +1,330 @@
...
+type timeProvider interface {
+	Now() time.Time
+}
+
+type defaultTime struct{}
+
+func (defaultTime) Now() time.Time { return time.Now() }
+
+func (alloc *Allocator) setTimeProvider(tp timeProvider) {
+	// fixme
```
tomwilkie: Erm, maybe we should remove this (and calls to it)?

bboreham:
Noted; I will refraim from rewriting previous commits for now because that often loses the context of comments here on GitHub

dpw:
I'll take another look at this once the first commit is hooked into the makefile 'test' target so that I know that it is self-contained, passes its own tests, etc.

I haven't actually looked at the code in the first and second commit, so my eyes remain mostly fresh.

bboreham:
Commits re-organised and test-passing.  Also extracted -pktdebug change out to a separate PR #567

dpw:
I can't see what I said about `-pktdebug` now, but what I meant was split it out into its own commit, not into its own PR (where I see you have incurred the wrath of Matthias due to leaving `-debug` doing nothing).

dpw:
I've seen the following bug several times.  I guess I can reproduce it about one time in three, though I have better and worse runs of luck.  Just reproduced it with a build from this branch checked out sometime this morning.

I'm running weave on my laptop host OS (dwragg-laptop) and in a VM (ubuntu1404):

Host:

    # ~dwragg/work/weaveworks/go/src/github.com/weaveworks/weave/weave launch -alloc 10.0.1.0/24

VM:

    # ~dwragg/work/weaveworks/go/src/github.com/weaveworks/weave/weave launch -alloc 10.0.1.0/24 192.168.122.1

Host:

    # ~dwragg/work/weaveworks/go/src/github.com/weaveworks/weave/weave expose

VM:

```
# ~dwragg/work/weaveworks/go/src/github.com/weaveworks/weave/weave expose
# ~dwragg/work/weaveworks/go/src/github.com/weaveworks/weave/weave status
weave router git-1f40f87d34f0
Encryption off
Our name is f6:43:c5:fe:24:e9(ubuntu1404)
Sniffing traffic on &{34 65535 ethwe 2a:37:24:d7:30:a7 up|broadcast|multicast}
MACs:
02:e0:38:b0:f1:4d -> 02:e0:38:b0:f1:4d(dwragg-laptop) (2015-04-23 12:37:12.861200086 +0000 UTC)
2a:37:24:d7:30:a7 -> f6:43:c5:fe:24:e9(ubuntu1404) (2015-04-23 12:37:10.136586417 +0000 UTC)
92:18:02:b2:25:77 -> f6:43:c5:fe:24:e9(ubuntu1404) (2015-04-23 12:37:10.660167879 +0000 UTC)
f6:43:c5:fe:24:e9 -> f6:43:c5:fe:24:e9(ubuntu1404) (2015-04-23 12:37:11.448019404 +0000 UTC)
0e:59:47:2c:9c:46 -> 02:e0:38:b0:f1:4d(dwragg-laptop) (2015-04-23 12:37:11.843317058 +0000 UTC)
Peers:
f6:43:c5:fe:24:e9(ubuntu1404) (v2) (UID 3027955244246475338)
   -> 02:e0:38:b0:f1:4d(dwragg-laptop) [192.168.122.1:6783]
02:e0:38:b0:f1:4d(dwragg-laptop) (v2) (UID 3288747158550528852)
   -> f6:43:c5:fe:24:e9(ubuntu1404) [172.17.42.1:46939]
Routes:
unicast:
f6:43:c5:fe:24:e9 -> 00:00:00:00:00:00
02:e0:38:b0:f1:4d -> 02:e0:38:b0:f1:4d
broadcast:
f6:43:c5:fe:24:e9 -> [02:e0:38:b0:f1:4d]
02:e0:38:b0:f1:4d -> []
Reconnects:

Allocator subnet 10.0.1.0+256
10.0.1.1 -> f6:43:c5:fe:24:e9(ubuntu1404) (0, 2, 126)
10.0.1.128 -> 02:e0:38:b0:f1:4d(dwragg-laptop) (0, 0, 127)
Address ranges we own:
  10.0.1.1+127 (1/0)
# /home/dwragg/work/weaveworks/go/src/github.com/weaveworks/weave/weave reset
```

Host:

```
# docker logs weave
[...]
panic: Trying to report free on space I don't own
[...]
```

Yes, by resetting weave on one machine, I killed the weave router on another.  See https://gist.github.com/dpw/3766c53f9ae700f8b26e for full panic.

bboreham:
The panic after tombstone should be addressed by https://github.com/bboreham/weave/commit/5e10c237c01b0b40c2850e11b0b3f1fb4c137920

dpw:
Kicking off round 2 on the implementation: Although GetFor is gone, there is still a file called getfor.go.

dpw:
The code under ipam/ring and ipam/space pass addresses around as `net.IP`, but usually need to store and operate on them as integers.  So there is lots of clutter converting them back and forth, and odd things like utils.Add and utils.Subtract.  And the code appears to assume 32-bit IPv4 addresses throughout, so it is not like the abstraction provided by net.IP is buying you anything.

So simply do `type address uint32` in one place, and then use that type throughout the rest of the ipam code.  Only use net.IP where it is really relevant (e.g. when parsing and formatting IP addresses in the HTTP API to the allocator).

dpw:
It seems that space.Space and its associated functions are not actually public - allocator.go works purely with space.Set. It would be nice to make that explicit by lower casing, so that the true API to the Space code is clearer.

Actually, I'd go a bit further.  `space.Set` should probably be called `Space` (as it is the abstraction exposed by the space module). What is currently called `Space` should be called something like `range`, except that is a reserved word in go, so `spaceRange` or something.

dpw:
I think  a rebase is in order so I can do another pass on some of the parts already commented on without it all becoming terribly confusing.

dpw:
The separation of `SpaceSet` and `Space` seems to make things rather verbose, as we often end up having similar code in two forms (`Allocate`, `NumFreeAddresses`, etc.).  It looks like fundamentally we are managing two sets of addresses: those this peer owns, and those this peer has allocated.  I'm not sure why we need the separation to implement that, and by avoiding it, I think we can reduce the line count considerably.

(`GiveUpSpace` does use a heuristic which is based on knowledge of Spaces, but as noted in my line comment, I'm not sure what the rationale for that is.)

I can't see a strong need for the use of bitsets to represent allocated addresses.  A peer will manage addresses for thousands rather than millions of containers, so we can afford to use a few bytes to track each one.

On the other hand, we do need to represent the ranges of addresses owned by a peer compactly, because a peer could own a whole /8 or bigger.

Given that there are no stringent performance requirements, we can use a common representation for both sets.  The "owned" set suggests a representation based on ranges.  I've sketched this idea out in https://gist.github.com/dpw/dd984ed746e80350e485 . I've tried to implement the main operations from SpaceSet; the few omissions should be easy to add.  It's 212 lines, including some tests (although no doubt it could use a few more).

dpw:
I have nothing against asserts.  I like asserts.   Some of my best friends are asserts.  But...

Asserts are a bit like executable comments.  And similarly to comments, the best assert is one that is not there because it would be obvious from the code.  So an assertion should only be added when it cannot reasonably be made superfluous by improvements to the code around it.

That leaves plenty of cases where asserts are appropriate.  And obviousness is subjective.  But I feel that some parts of the IPAM code lean a bit too heavily on asserts.

And because I assume we'll see `Assert` spread to other parts of the codebase, I'd prefer it if the main variant of `Assert` did not take a message.  The condition in an assert should generally be self-explanatory (if it's not, that is a hint that the code around it needs further consideration).  Although it is unfortunate that golang does not have a built-in assert that would include the condition expression as the panic message, I doubt the message is vital: the function name and line number should be sufficient to track down the failing assert.  My worry is that the message (combined with golang's cumbersome error handling) will create a temptation to use `Assert` for conditions that would be better regarded as errors.  If `Assert` just takes a condition, it will be clear that it is only for sanity checks.

tomwilkie:
The asserts in ring were added to prevent future, potentially innocuous
looking changes from breaching invariants.  There are subleties to the
algorithm that need to be expressed. For sure I've learnt on the side of
including more, but I don't think that causes any harm. None of them should
be triggerable by any combination of user actions.

I am concerned that a bug in ipam will take down the router; I'm thinking
perhaps catching these panics and closing the ipam actor thread (with a big
error in the logs) may be preferable?  If we can make future requests error
gracefully.

I don't mind re:including a message in the assert. I don't see the harm in
being verbose, but I will go ahead and remove them.

tomwilkie:
Discussed this with David and he'd prefer to not special case assertions inside the IPAM code.  I've written it anyway, but I'm not suggesting we include this code in this PR: https://github.com/tomwilkie/weave/commit/46570f17e0c8dc9eafd2e083e6761315243d7957

docs/ipam.md
```diff
@@ -0,0 +1,214 @@
...
+
+When a peer owns some range(s), it can allocate freely from within
+those ranges to containers on the same machine. If it runs out of
+space (all owned ranges are full), it will ask another peer for space:
+  - it picks a peer to ask at random, weighted by the amount of space
+    owned by each peer
+    - if the target peer decides to give up space, it unicasts a message
+      back to the asker with the newly-updated ring.
+    - if the target peer has no space, it unicasts a message back to the
+      asker with its current copy of the ring,on the basis tha the
```
rade: typo: "tha"

ipam/paxos/paxos.go
```diff
@@ -0,0 +1,252 @@
...
+		if our_claims.Promise.precedes(claims.Promise) {
+			our_claims.Promise = claims.Promise
+		}
+	}
+
+	// The "Accept Request" step of Paxos: Acting as a proposer,
+	// do we have a proposal that has been promised by a quorum?
+	//
+	// In Paxos, the "proposer" and "acceptor" roles are distinct,
+	// so in principle a node acting as a proposer could could
```
rade: typo: "could could"

site/features.md
```diff
@@ -53,6 +54,38 @@ now re-use the same tools and techniques when deploying applications
...
+
+You must specify the same range with `-iprange` on each host, and you
+cannot mix weaves started with and without -iprange.
+
+You may wish to `weave stop` and re-launch to change some config or to
+upgrade to a new version; provided the underlying protocol hasn't
+changed it will pick up where it left off and learn from peers in the
+network which address ranges it was previously using. If, however, you
+run `weave reset` this will remove the peer from the network so
+if Weave is run again on that node it will start from scratch.
```
rade: it's lower-case 'weave' everywhere expect headings and start of sentence.

router/peers.go
```diff
@@ -32,6 +33,24 @@ func NewPeers(ourself *Peer, onGC func(*Peer)) *Peers {
 		onGC:    onGC}
 }
 
+func (peers *Peers) AddNewPeerFunc(f func(*Peer)) {
+	peers.Lock()
+	defer peers.Unlock()
+
+	peers.onNew = append(peers.onNew, f)
+
+	f(peers.ourself)
```
rade: ourself is in the peers.table, so no point doing this.

ipam/allocator.go
```diff
@@ -0,0 +1,660 @@
...
+		// per RFC 1122, don't allocate the first and last address in the subnet
+		ring:               ring.New(utils.Add(subnetStart, 1), utils.Add(subnetStart, subnetSize-1), ourName),
+		owned:              make(map[string]utils.Address),
+		otherPeerNicknames: make(map[router.PeerName]string),
+	}
+	alloc.paxos.Init(ourName, ourUID, quorum)
+	return alloc, nil
+}
+
+// OnNewPeer is part of the NewPeerWatcher interface, and is called by the
```
rade: `NewPeerWatcher` no longer exists.

weaver/main.go
```diff
@@ -163,7 +183,45 @@ func initiateConnections(router *weave.Router, peers []string) {
 	}
 }
 
-func handleHTTP(router *weave.Router, httpAddr string) {
+func createAllocator(router *weave.Router, apiPath string, iprangeCIDR string, quorum uint) *ipam.Allocator {
+	allocator, err := ipam.NewAllocator(router.Ourself.Peer.Name, router.Ourself.Peer.UID, iprangeCIDR, quorum)
+	if err != nil {
+		log.Fatal(err)
+	}
+	allocator.SetInterfaces(router.NewGossip("IPallocation", allocator))
```
rade: this is racy. The router has already been started, so you could receive IPAM gossip from that point onwards, which will result in errors in the logs. I think.

Really the allocator and router should be wired up before calling `Start()` on either of them.

router/peers.go
```diff
@@ -32,6 +33,24 @@ func NewPeers(ourself *Peer, onGC func(*Peer)) *Peers {
 		onGC:    onGC}
 }
 
+func (peers *Peers) AddNewPeerFunc(f func(*Peer)) {
```
rade: Do we really need all this machinery? Can't we create `Peers` with a (single) `onNew` function?

site/features.md
```diff
@@ -53,6 +54,38 @@ now re-use the same tools and techniques when deploying applications
...
+
+    host1# weave launch -iprange 10.2.3.0/24
+
+The `run`, `start`, `attach` and `expose` commands will then allocate
+an address automatically if none is specified, i.e.:
+
+    host1# D=$(weave run -t -i ubuntu)
+
+Weave uses the Docker Events API to learn when a container has exited
+and hence can release its IP address.
```
rade: Drop the mentioning of the 'Docker Events API' - it's an implementation detail.

weave
```diff
@@ -571,6 +581,44 @@ populate_dns() {
...
+                CONTAINER="${status#true }"
+                ;;
+            *)
+                echo "$1 container is not running." >&2
+                return 1
+                ;;
+        esac
+        CIDR_ARGS="$(http_call $CONTAINER_NAME $HTTP_PORT POST /ip/$CONTAINER)"
+        if [ "$CIDR_ARGS" = "404 page not found" ] ; then
+            echo "No IP address supplied (use the -iprange option on 'weave launch'\nto enable IP allocation)" >&2
```
rade: "enable automatic IP address allocation"

weave
```diff
@@ -801,6 +857,7 @@ case "$COMMAND" in
     expose)
         collect_cidr_args "$@"
         shift $CIDR_COUNT
+        [ $CIDR_COUNT -gt 0 ] || { get_ipam_cidr weave && expose $CIDR_ARGS ; }
```
rade: Surely we need a corresponding change to 'hide' in order to release the IP address.

weave
```diff
@@ -853,6 +910,7 @@ case "$COMMAND" in
         ;;
     reset)
         [ $# -eq 0 ] || usage
+        http_call $CONTAINER_NAME $HTTP_PORT POST /tombstone-self >/dev/null 2>&1 || true
```
rade: I thought tombstones were gone. In any case they are an implementation detail, so the HTTP API should use some more generic terminology.

ipam/allocator.go
```diff
@@ -0,0 +1,660 @@
...
+		dst := 0
+		for src := 1; src < len(peers); src++ {
+			if peers[dst] != peers[src] {
+				dst++
+				peers[dst] = peers[src]
+			}
+		}
+
+		return peers[:dst+1]
+	}
```
rade: cosmetic: get rid of the `else`.

ipam/allocator.go
```diff
@@ -0,0 +1,660 @@
...
+				alloc.gossip.GossipBroadcast(alloc.Gossip())
+			}
+
+			if cons := alloc.consensus(); cons != nil {
+				alloc.createRing(cons)
+			}
+		}
+	}
+
+	return err
```
rade: move that `return err` into the first branch, remove the `else` and then `return nil` here.

ipam/http.go
```diff
@@ -0,0 +1,116 @@
...
+			ident, err := parseURL(r.URL.Path)
+			if err != nil {
+				badRequest(w, err)
+			} else if err = alloc.Free(ident); err != nil {
+				badRequest(w, err)
+			}
+		default:
+			http.Error(w, "Verb not handled", http.StatusBadRequest)
+		}
+	})
```
rade: Why arent' we using mux's URL parsing? See, for example, https://github.com/weaveworks/weave/blob/master/nameserver/http.go#L28

ipam/paxos/paxos.go
```diff
@@ -0,0 +1,252 @@
...
+	if a.Round != b.Round {
+		return a.Round < b.Round
+	} else if a.Proposer.Name != b.Proposer.Name {
+		return a.Proposer.Name < b.Proposer.Name
+	} else if a.Proposer.UID != b.Proposer.UID {
+		return a.Proposer.UID < b.Proposer.UID
+	} else {
+		return false
+	}
+}
```
rade: The 'go way' is to use a `switch`.

ipam/paxos/paxos.go
dpw: Fixed

ipam/allocator.go
dpw: Done

ipam/allocator.go
dpw: Done

ipam/paxos/paxos.go
dpw: Done

site/features.md
```diff
@@ -53,6 +54,38 @@ now re-use the same tools and techniques when deploying applications
...
+network which address ranges it was previously using. If, however, you
+run `weave reset` this will remove the peer from the network so
+if Weave is run again on that node it will start from scratch.
+
+For failed peers, the `weave rmpeer` command can be used to permanently
+remove the ranges allocated to said peer.  This will allow other peers
+to allocate IPs in the ranges previously owner by the rm'd peer, and as such
+should be used with extreme caution - if the rm'd peer later rejoins
+the Weave network, the same IP address may be allocated twice.
+
```
rade: As discussed, the peer rejoining isn't the only problematic situation. We need to document the others.

ipam/paxos/paxos.go
```diff
@@ -0,0 +1,254 @@
+package paxos
```
rade: There are a small number of non-`camel_case` identifiers in this package, which should be turned into `camelCase`

ipam/paxos/paxos.go
```diff
@@ -0,0 +1,254 @@
...
+
+// Has a consensus been reached, based on the known claims of other nodes?
+func (node *Node) Consensus() (bool, AcceptedValue) {
+	counts := map[ProposalID]uint{}
+
+	for _, claims := range node.knows {
+		if claims.Accepted.valid() {
+			origin := claims.AcceptedVal.Origin
+			count := counts[origin] + 1
+			counts[origin] = count
```
rade: `counts[origin]++`?

ipam/utils/utils.go
```diff
@@ -0,0 +1,63 @@
...
+		r[i] = byte(key)
+		key >>= 8
+	}
+	return
+}
+
+func (addr Address) String() string {
+	return AddressIP4(addr).String()
+}
+
```
rade: There is code identical (in purpose) to all the above in the [nameserver code](https://github.com/weaveworks/weave/blob/master/nameserver/addrs.go#L81). Please raise an issue to unify the two post-merge of this PR.

ipam/allocator_test.go
```diff
@@ -0,0 +1,437 @@
...
+	wt.AssertNoErr(t, err)
+	_, addr3 := alloc.Allocate(container3, nil)
+	wt.AssertEqualString(t, addr3.String(), testAddr1, "address")
+}
+
+func (alloc *Allocator) sleepForTesting(d time.Duration) {
+	alloc.actionChan <- func() {
+		time.Sleep(d)
+	}
+}
```
rade: You can avoid a couple of sleeps and resulting races by instead having a `pauseForTesting` function that uses a pair of extra channels... one channel to signal that pausing has started, so this function only returns when that has happened, and one channel, which the function returns, for the caller to resume the actor (by closing that channel).

ipam/http_test.go
```diff
@@ -0,0 +1,156 @@
...
+	alloc := makeAllocatorWithMockGossip(t, "08:00:27:01:c3:9a", testCIDR1, 1)
+	port := rand.Intn(10000) + 32768
+	fmt.Println("Http test on port", port)
+	go listenHTTP(port, alloc)
+
+	time.Sleep(100 * time.Millisecond) // Allow for http server to get going
+
+	alloc.claimRingForTesting()
+	// Ask the http server for a new address
+	cidr1 := HTTPPost(t, fmt.Sprintf("http://localhost:%d/ip/%s", port, containerID))
```
rade: That `fmt` repeats a lot. Would benefit from a `allocURL(port, containerID)` helper.

ipam/paxos/paxos_test.go
```diff
@@ -0,0 +1,351 @@
...
+			m.t.Fatal("Node disagrees about consensus")
+		}
+	}
+}
+
+func TestPaxos(t *testing.T) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	params := TestParams{
+		nodeCount:     10,
```
rade: Why 10? To allow for three partitions, with non-trivial (i.e. not 0,1,2) node counts in each?

ipam/ring/ring_test.go
```diff
@@ -0,0 +1,713 @@
...
+	ring2 := New(start, end, peer2name)
+
+	// Claim everything for peer1
+	ring1.Entries = []*entry{{Token: start, Peer: peer1name, Free: 255}}
+	ring2.merge(*ring1)
+	wt.AssertEquals(t, ring1.Entries, ring2.Entries)
+
+	// Now grant a split range to peer2
+	ring1.GrantRangeToHost(dot10, dot245, peer2name)
+	wt.AssertEquals(t, ring1.Entries, entries{{Token: start, Peer: peer1name, Version: 1, Free: 10},
```
rade: Nearly everywhere the field order is {Token, Peer, Free, Version}, but not here and in a few other places. Is there a reason for that? If not then please make it consistent.

ipam/paxos/paxos.go
dpw: Oh, how I wanted to be able to turn those three lines into

    if ++counts[origin] >= quorum {

Unfortunately, golang.

As I think you are suggesting, I could do

    counts[origin]++
    if counts[origin] >= quorum {

But, depending on how smart the compiler is (and 6g is generally not very smart) that might lead to three accesses to the map, when only two are needed.  That offends my sense of efficiency.  I don't see how to get it down to two lines with only two map accesses, hence the three lines.

ipam/paxos/paxos.go
dpw: Done.  Horrible, but done.

ipam/paxos/paxos.go
rade: Yes, I figured it would be two lines. Not worth arguing about. So leave it as is if you like.

weave
bboreham: I think at the time this was written I was unclear how to get hold of the specific IP address to hide; however this isn't necessary since we can ask IPAM to release _all_ addresses (effectively only one) for the container name ("weave").

the UX will get more complicated when IPAM supports multiple subnets.

rade: the container name should be `weave:expose`. That's what we use for DNS.

rade: incidentally, this code is just wrong. As it stands it ends up associating the IP with the weave container. But the address is in fact attached to the bridge, which survives weave restarts.

bboreham: Turns out it doesn't matter what you told a weave which has been restarted.  For correct operation, populate_ipam() should include the bridge address.

rade:
I've put the "allow Gossiper.Gossip to return nil" and "introduce PeerUID type" changes on master in https://github.com/weaveworks/weave/commit/ace56da7b3639d7a466b24ccd96e54791a11600f and https://github.com/weaveworks/weave/commit/c7dd48d7699ed5e62955f58b14e2405f8a108cae, respectively.

weaver/main.go
```diff
@@ -121,12 +129,23 @@ func main() {
...
+		if iprangeCIDR != "" {
+			allocator = createAllocator(router, apiPath, iprangeCIDR, determineQuorum(peerCount, peers))
+		} else {
+			if peerCount > 0 {
+				log.Fatal("-initpeercount flag specified without -iprange")
+			}
+		}
+	}
+	if httpAddr == "" || iprangeCIDR == "" {
+		router.NewGossip("IPallocation", &ipam.DummyAllocator{})
```
rade: This is frightfully convoluted, and wrong...

- we shouldn't call `handleHTTP` (in the code following this) unless `httpAddr != nil`
- we should create the allocator regardless of whether there is an `httpAddr` present. Otherwise quorum may not be reached.

So I think the whole lot should just be
````
	var allocator *ipam.Allocator
	if ipRangeCIDR != "" {
		allocator = createAllocator(router, apiPath, iprangeCIDR, determineQuorum(peerCount, peers))
	} else if peerCount > 0 {
		log.Fatal("-initpeercount flag specified without -iprange")
	} else {
		router.NewGossip("IPallocation", &ipam.DummyAllocator{})
	}
	router.Start()
	initiateConnections(router, peers)
 	if httpAddr != "" {
		go handleHTTP(router, httpAddr)
	}

dpw: An if .. else if .. else? I'm shocked you didn't turn that into a switch!

rade: Was thinking about `switch` but I think it would be less clear in this case.

dpw: Done

PR #566, opened 2015-04-16
Add optional hostname argument to weave expose

### LINKED ISSUE #375: [dns] naming 'expose'd hosts
moonfish: Back to my question, then: host2# weave expose 10.0.1.102/24.  Now from a weave + dns container, I ping 10.1.0.102 - works. ping host2 - nada
yaronr
because doing expose doesn't register 'host2' as 10.1.0.102 in weavedns
yaronr
(or does it?0
moonfish
no. and neither should it. You could have a weave network of a dozen hosts which are all called 'host2'.
yaronr
tell you why I'm asking:
moonfish
we could perhaps add an option to 'expose' to specify a name that goes into weavedns.
yaronr
back to mesos - I'm running mesos slaves with -host. I had to manually re-write the resolv.conf, putting the bridge ip as nameserver. but the hostname is set by mesos. I also had to change dns resolution to be based on DNS (and not /etc/hosts). so mesos starts, and tries to gethostbyname. but since host2 is not a weave dns name, it returns nothing, and that's a problem.
yaronr
moonfish: so back to 'perhaps add an option to 'expose' to a specific name' - that would be very useful
moonfish
right. At first glance at least, does seem like a reasonable thing one may want to do.
moonfish
So please file an issue :)
yaronr
Also I believe that many admins will like it, especially those that set their names using puppet / chef / whatever  - they usually won't have a dozen 'host2'
yaronr
will do :)

awh:
Fix #375:
````
vagrant@ubuntu-14:~/weave$ ./weave expose 10.1.1.1/24 -h myservice.weave.local
vagrant@ubuntu-14:~/weave$ ./weave status
weave router git-59ae50eb4ef9
Encryption off
Our name is 7a:c5:c4:15:be:33(ubuntu-14)
Sniffing traffic on &{10 65535 ethwe 16:3d:0b:8a:66:e4 up|broadcast|multicast}
MACs:
7a:c5:c4:15:be:33 -> 7a:c5:c4:15:be:33(ubuntu-14) (2015-04-16 09:23:30.45427506 +0000 UTC)
6a:0a:da:89:30:de -> 7a:c5:c4:15:be:33(ubuntu-14) (2015-04-16 09:22:13.804277587 +0000 UTC)
16:3d:0b:8a:66:e4 -> 7a:c5:c4:15:be:33(ubuntu-14) (2015-04-16 09:22:03.786283726 +0000 UTC)
32:dd:63:79:54:df -> 7a:c5:c4:15:be:33(ubuntu-14) (2015-04-16 09:22:04.374157935 +0000 UTC)
Peers:
Peer 7a:c5:c4:15:be:33(ubuntu-14) (v0) (UID 5547860031843093274)
Routes:
unicast:
7a:c5:c4:15:be:33 -> 00:00:00:00:00:00
broadcast:
7a:c5:c4:15:be:33 -> []
Reconnects:


weave DNS git-59ae50eb4ef9
Local domain weave.local.
Listen address :53
mDNS interface &{14 65535 ethwe 6a:0a:da:89:30:de up|broadcast|multicast}
Fallback DNS config &{[66.28.0.45 8.8.8.8] [] 53 1 5 2}
Zone database:
weave:expose 10.1.1.1 myservice.weave.local.
````

There are however a number of caveats as compared to the behaviour of `-h` as used in conjunction with run:

* `weave hide` cannot remove the DNS entries in the same way that `weave detach` does as there is no container in which to persist the metadata 
* We have to use the synthetic container ID of `weave:expose` when we interact with the DNS REST API (@inercia can you foresee any problems with this?)

### DIFF
--- a/site/features.md
+++ b/site/features.md
@@ -190,6 +190,11 @@ invocation:
     host2# weave expose 10.2.1.102/24 10.2.2.102/24
     host2# weave hide 10.2.1.102/24 10.2.2.102/24
 
+Finally, exposed addresses can be added to weaveDNS by supplying a
+hostname:
+
+    host2# weave expose 10.2.1.102/24 -h exposed.weave.local
+
 ### <a name="service-export"></a>Service export
 
 Services running in containers on a weave network can be made
--- a/weave
+++ b/weave
@@ -23,7 +23,7 @@ usage() {
     echo "weave start      <cidr> [<cidr> ...] <container_id>"
     echo "weave attach     <cidr> [<cidr> ...] <container_id>"
     echo "weave detach     <cidr> [<cidr> ...] <container_id>"
-    echo "weave expose     <cidr>"
+    echo "weave expose     <cidr> [-h <hostname>]"
     echo "weave hide       <cidr>"
     echo "weave ps         [<container_id> ...]"
     echo "weave status"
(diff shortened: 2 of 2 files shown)


inercia:
@awh I think it should be ok to have synthetic names like that... in fact, I will introduce a `remote` container ID in my next PR, so `weave:expose` should not be problematic either...

rade:
We should make `hide` work properly for this.

awh:
D'oh - the REST API doesn't require the hostname so fix is trivial...

PR #567, opened 2015-04-16
Rename per-packet debugging flag to `-pktdebug`

bboreham:
For consistency with WeaveDNS, and common usage, `-debug` is hereby re-appropriated to mean "a bit more logging" and "give me a log-line per packet" is relegated to `-pktdebug`.

This change driven by IPAM, where there are some log lines you want to see when debugging.  If this change were merged on its own, `-debug` would have no effect.


rade:
No dead code please. Either get rid of `-debug` or find some use for it on master.

bboreham:
So leave it until IPAM is merged

PR #569, opened 2015-04-17
Simplify address parsing for populate_dns in weave script

bboreham:
Took out one `echo`, one `grep` and one subshell.


rade:
I *think* this will break if the container has died between the `docker ps` and `with_container_netns`.

bboreham:
Amended to put the `if` back in

PR #572, opened 2015-04-17
Separate Forwarder and ForwarderDF

dpw:
Much of the state and code associated with Forwarder only applied in the
DF case.

### DIFF
--- a/router/forwarder.go
+++ b/router/forwarder.go
@@ -38,12 +38,10 @@ func (conn *LocalConnection) ensureForwarders() error {
 	}
 
 	forwarder := NewForwarder(conn, encryptor, udpSender, DefaultPMTU)
-	forwarderDF := NewForwarder(conn, encryptorDF, udpSenderDF, DefaultPMTU)
+	forwarderDF := NewForwarderDF(conn, encryptorDF, udpSenderDF, DefaultPMTU)
 	effectivePMTU := forwarderDF.unverifiedPMTU
-	//NB: only forwarderDF can ever encounter EMSGSIZE errors, and
-	//thus perform PMTU verification
-	forwarder.Start(nil)
-	forwarderDF.Start(make(chan int, ChannelSize))
+	forwarder.Start()
+	forwarderDF.Start()
 
 	// Various fields in the conn struct are read by other processes,
 	// so we have to use locks.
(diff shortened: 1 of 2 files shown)


router/forwarder.go
```diff
@@ -38,11 +38,11 @@ func (conn *LocalConnection) ensureForwarders() error {
...
 
 	forwarder := NewForwarder(conn, encryptor, udpSender, DefaultPMTU)
-	forwarderDF := NewForwarder(conn, encryptorDF, udpSenderDF, DefaultPMTU)
+	forwarderDF := NewForwarderDF(conn, encryptorDF, udpSenderDF, DefaultPMTU)
 	effectivePMTU := forwarderDF.unverifiedPMTU
 	//NB: only forwarderDF can ever encounter EMSGSIZE errors, and
 	//thus perform PMTU verification
-	forwarder.Start(nil)
+	forwarder.Start()
 	forwarderDF.Start(make(chan int, ChannelSize))
```
rade: should just be `forwarderDF.Start()` now, i.e. the chan creation can now happen inside that.

router/forwarder.go
```diff
@@ -38,11 +38,11 @@ func (conn *LocalConnection) ensureForwarders() error {
 	}
 
 	forwarder := NewForwarder(conn, encryptor, udpSender, DefaultPMTU)
-	forwarderDF := NewForwarder(conn, encryptorDF, udpSenderDF, DefaultPMTU)
+	forwarderDF := NewForwarderDF(conn, encryptorDF, udpSenderDF, DefaultPMTU)
 	effectivePMTU := forwarderDF.unverifiedPMTU
 	//NB: only forwarderDF can ever encounter EMSGSIZE errors, and
 	//thus perform PMTU verification
```
rade: I think that comment is now somewhat out of place. It previously served as an explanation of why we called `Start` with two different arguments.

router/forwarder.go
```diff
@@ -351,16 +292,7 @@ func (fwd *Forwarder) flush() {
 	err := fwd.udpSender.Send(fwd.enc.Bytes())
 	if err != nil {
 		if mtbe, ok := err.(MsgTooBigError); ok {
```
rade: It's a shame to have this in the `Forwarder` code given that we can only get `MsgTooBigError` in `ForwarderDF`.

How about we increase the scope of of the custom handler slightly...

````
if err := processSendError(fwd.udpSender.Send(fwd.enc.Bytes())); err != nil && PosixError(err) != syscall.ENOBUFS {
    fwd.conn.Shutdown(err)
}
````
`Forwarder` would supply an identify function for `processSendError`, whereas `ForwarderDF` would supply a function that handles `MsgTooBigError`.

router/forwarder.go
dpw: Yeah.  Indeed, it seems redundant now, as it ought to be clearer from the code that forwarder and forwarderDF are different and how they are different.

router/forwarder.go
dpw: Done

router/forwarder.go
dpw: Done

dpw:
Fixups squashed into the single commit. github kept the comments this time, yay.

rade:
`func(err error) error { return err }}`. Hilarious :)

PR #580, opened 2015-04-20
Replace `ip netns exec` with `nsenter`

### LINKED ISSUE #458: use `nsenter` instead of `ip netns exec`
By replacing our use of `ip netns exec` with `nsenter` we could remove the current [network namespace symlinking hack](https://github.com/zettio/weave/blob/9eeec2fa4835c4e50ea455412068ecfb86ceddc6/weave#L191).

The only downside is that `nsenter` is a fairly recent command. So, for example, [it only appeared in Ubuntu 14.10](http://askubuntu.com/questions/439056/why-there-is-no-nsenter-in-util-linux). This is less of an issue post #388 though.

awh:
Fixes #458. All smoke tests pass (`400_weave_local_test.sh` necessitated updating the Vagrantfiles to Ubuntu 14.10) but nevertheless quite invasive - suggest merging after 0.10.0?

### DIFF
--- a/weave
+++ b/weave
@@ -261,10 +261,10 @@ docker_bridge_ip() {
 
 # the following borrows from https://github.com/jpetazzo/pipework
 
-# Set $NETNS to the network namespace of container $1, $LOCAL_IFNAME
-# and $GUEST_IFNAME to suitable names for two ends of a veth pair,
-# specific to the container, and execute args $2 $3 ... as a command.
-# If an error is caused by container dying, swallow output from error.
+# Set $CONTAINER_PID to the PID of container $1, $LOCAL_IFNAME and
+# $GUEST_IFNAME to suitable names for two ends of a veth pair, specific
+# to the container, and execute args $2 $3 ... as a command. If an error
+# is caused by container dying, swallow output from error.
 with_container_netns () {
     CONTAINER="$1"
     CONTAINER_PID=$(docker inspect --format='{{.State.Pid}}' $CONTAINER)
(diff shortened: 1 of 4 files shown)


rade:
- The repetition of `nsenter --net=$PROCFS/$NETNS/ns/net` is a bit tiresome. Yes, the same was true for `ip netns exec $NETNS`, but at least that was shorter.

- `$NETNS` worked as a var name for `ip netns exec $NETNS`, but is somewhat misleading for use in `$PROCFS/$NETNS`. Could we just use `$CONTAINER_PID` instead?

- We need to check what, if any, new version constraints this imposes on the docker host's kernel. While `nsenter` itself is only a user space tool, it does depend on kernel features, so the questions is whether it depends on more recent features than `ip netns exec`.

awh:
> The repetition of nsenter --net=$PROCFS/$NETNS/ns/net is a bit tiresome. Yes, the same was true for ip netns exec $NETNS, but at least that was shorter.

Abstracted out to a function.

> $NETNS worked as a var name for ip netns exec $NETNS, but is somewhat misleading for use in $PROCFS/$NETNS. Could we just use $CONTAINER_PID instead?

Done.

> We need to check what, if any, new version constraints this imposes on the docker host's kernel. While nsenter itself is only a user space tool, it does depend on kernel features, so the questions is whether it depends on more recent features than ip netns exec.

I've smoke tested this on Ubuntu Precise 12.04/3.5.0-54-generic and it works (modulo #585 and `400_weave_local_test.sh`).

weave
```diff
@@ -261,10 +261,10 @@ docker_bridge_ip() {
...
 # the following borrows from https://github.com/jpetazzo/pipework
 
-# Set $NETNS to the network namespace of container $1, $LOCAL_IFNAME
-# and $GUEST_IFNAME to suitable names for two ends of a veth pair,
-# specific to the container, and execute args $2 $3 ... as a command.
-# If an error is caused by container dying, swallow output from error.
+# Set $LOCAL_IFNAME and $GUEST_IFNAME to suitable names for two ends
+# of a veth pair, specific to the container, and execute args $2 $3 ...
+# as a command. If an error is caused by container dying, swallow output
+# from error.
```
rade: Should note here that this  sets `CONTAINER_PID` (to the pid of the container passed in $1).

awh:
Fixed `with_container_netns` comment.

Issue #576, opened 2015-04-20
dealing with old dockers in the weavexec era


rade:
Our README says we support Docker 0.9.1 or later. I am pretty sure that is not true anymore, since the weavexec image introduced in #388 contains Docker 1.3.1, so that ought to be the oldest daemon version it can work with.

`weave --local` can probably still cope with older versions, but that is not how the majority of user will run weave.

So I think we need to update the docs, to mandate >=1.3.1.  @squaremo correct?

But surely we should also amend the version checks in the script, since otherwise users will get obscure errors when running against older daemons. Or are these errors quite self-explanatory?

If we are going to amend the checks, we really want to only check that the version of the docker *daemon* we are running against is >=1.3.1. And only if we are running in the context of the weavexec image, since the existing checks are sufficient outside that. `docker version` does tell us the server version, but a) does that work when the server is an older version, and b) is the output format the same across server versions?

squaremo:
Docker version requirements:

 * We cannot use docker client version 1.3.0, specifically, for any purpose, since it outputs things to stdout which it shouldn't
 * Using `--with-dns` requires docker client >=1.2.0, and since the server version needs to be >= the client version, a server >= 1.2.0. Unless `run --local` is used, this will be executed in weaveexec, which has docker client v1.3.1.

Without `--local`, the server needs to be >=1.3.1, and the client needs to be !=1.3.0. With `--local` and `--with-dns`, the client needs to be >=1.2.0.

If the docker server is <1.3.1, our embedded docker won't be able to talk to it, which might be confusing if the user's docker client is also <1.3.1, since they will get a message indicating that the versions are mismatched (which isn't true from their point of view). So we may want to head that off.

rade:
Let's not worry about `--local run`.

We could decrease the docker version included in weavexec. No downside to that except for #582 and any bugs that might be in the client-side portion. I have no feel for the latter, unfortunately. Ideally we'd drop all the way back to 0.9.1 since that is in ubuntu trusty. trusty-updates has 1.0.1. utopic has 1.2.0. and vivid has 1.5.0.

Shall we go for 1.2.0? It's what the server needs to be for `--with-dns`. We are integrating weavedns more and more and want more folks to use it, and there is some merit to having just one version requirement. (i.e. server >= 1.2.0).

I think that's the only change I'd make for 0.10. If it turns out that several users are running into version compatible issues and are confused by the error messages, we can address that in a 0.10.1 or 0.11 release.

[...]

PR #584, opened 2015-04-21
lower weavexec docker version

rade:


rade:
@squaremo review please.

NB: this doesn't fully resolve #576, but is intended to be good enough for 0.10.

rade:
Closing due to #587.

PR #589, opened 2015-04-22
return NXDOMAIN instead of NOTIMP

### LINKED ISSUE #588: [dns] return NXDOMAIN instead of NOTIMP for most query types
````
# host seetwo.weave.local
seetwo.weave.local has address 10.2.0.34
Host seetwo.weave.local not found: 4(NOTIMP)
Host seetwo.weave.local not found: 4(NOTIMP)
````
The latter results are from AAAA and MX queries. Surely we should return NXDOMAIN for most of these on the grounds that we do understand them but have no matching records.

### PR DESCRIPTION
**Pull request by inercia**
Fixes #588 by returning NOTIMP for some unimplemented query types, NXDOMAIN otherwise

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -88,6 +88,18 @@ exec_on() {
     docker -H tcp://$host:2375 exec $container "$@"
 }
 
+dig_on() {
+    host=$1
+    cont=$2
+    query=$3
+    docker -H tcp://$host:2375 exec $cont sh -c "dig $query"
+}
+
+# assert_dns_status <host> <container> <name> <query> <status>
+assert_dns_status() {
+    assert_raises "dig_on \"$1\" \"$2\" \"$3\" | grep -q \"status: $4\""
+}
+
 # assert_dns_record <host> <container> <name> <ip>
 assert_dns_record() {
     assert "exec_on $1 $2 getent hosts $3 | tr -s ' '" "$4 $3"
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -74,14 +74,6 @@ func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
 	return m
 }
 
-func makeDNSNotImplResponse(r *dns.Msg) *dns.Msg {
-	m := new(dns.Msg)
-	m.SetReply(r)
-	m.RecursionAvailable = true
-	m.Rcode = dns.RcodeNotImplemented
-	return m
-}
-
 // get the maximum UDP-reply length
 func getMaxReplyLen(r *dns.Msg, proto dnsProtocol) int {
 	maxLen := minUDPSize
--- a/nameserver/server.go
+++ b/nameserver/server.go
@@ -277,7 +277,7 @@ func (s *DNSServer) commonQueryHandler(proto dnsProtocol, kind string, qtype uin
 		// catch unsupported queries
 		if q.Qtype != qtype {
 			Debug.Printf("[dns msgid %d] Unsupported query type %s", r.MsgHdr.Id, dns.TypeToString[q.Qtype])
-			m := makeDNSNotImplResponse(r)
+			m := makeDNSFailResponse(r)
 			s.cache.Put(r, m, negLocalTTL, 0)
 			w.WriteMsg(m)
 			return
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by rade** [comment]
wouldn't it have been easier to list the query types that we *do* implement?

**Comment by rade** [comment]
also, you should use the canonical set representation (#518).

**Comment by rade** [comment]
Hmm. querying the weave.works domain, I have yet to find a query type that *doesn't* result in an NXDOMAIN. Perhaps that is what DNS servers typically do? How did you come up with the list of unimplemented types?

**Comment by inercia** [comment]
I haven't found much information on when to return NOTIMP, just [this](https://tools.ietf.org/html/draft-ogud-dnsop-any-notimp-00) RFC that recommends using it for rejecting ANY queries... I will reduce the set to just this query type...

**Comment by rade** [comment]
That RFC is a very recent draft. So it's rather flimsy evidence.

Is NXDOMAIN a valid response for all query types?

The duplication between queryHandler and rdnsHandler is getting out of hand.

**Comment by inercia** [comment]
We have two options: a) we can just reply with NXDOMAIN for all query types or b) reply with NOTIMPL to ANY queries. I think the safest option is a), but it is up to you @rade

**Comment by rade** [comment]
I am fine with a *provided NXDOMAIN is a valid response for all query types*. Is it?

**Comment by inercia** [comment]
From my understanding of the DNS RFC, I think NXDOMAIN is the general error code (unless otherwise specified). The only documented evidence of the need of NOTIMP is that draft RFC...

**Comment by rade** [comment]
ok. In which case let's just go with NXDOMAIN for everything.

Would be good to add a smoke tests that checks for the (absence of) the behaviour I reported in #588.

Re the the duplication between queryHandler and rdnsHandler, I can take a stab at that after this has been merged.

**Comment by awh** [comment]
Just want to confirm my understanding of what we've decided needs to be done for this PR - I think it becomes a one line change to return NXDOMAIN instead of NOTIMP, and a smoke test to that effect?

**Comment by inercia** [comment]
@awh Yes, returning NXDOMAIN instead of NOTIMP is trivial... no idea of what the smoke test would involve though...

**Comment by awh** [comment]
> no idea of what the smoke test would involve though...

Something that simply asserts that `NXDOMAIN` is in fact returned for query types we don't handle should suffice...

**Comment by rade** [comment]
you could just run what's in #588, asserting that you get only get one line as an answer / get no NOTIMPs.

**Comment by rade** [comment]
@inercia ping

**Comment by inercia** [comment]
@rade Just updated, with some integration tests now...

**Comment by rade** [comment]
Hmm. Do we have to use `dig`? All of the other tests use `getent`.

**Comment by inercia** [comment]
I'm afraid we can't: `getent` is just a frontend to the resolver database, so it does not return low-level return codes obtained from the DNS queries (eg, NXDOMAIN). (in fact, I'm not sure it is a good thing to do in these tests...)

**Comment by rade** [comment]
> `getent` does not return low-level return codes obtained from the DNS queries (eg, NXDOMAIN)

Well, as you can see from #588, it *does* return some error codes, but ignores NXDOMAIN. No idea whether that's a feature of the resolver library or `getent`.

> I'm not sure [using the resolver library] is a good thing to do in these tests

The smoke tests are meant to operate at a very high level, close to what we expect users/applications to do. Few users/apps query DNS directly; they go through the resolver library. Hence `getent` is preferable to `dig`. Except of course when it can't test what we want. So if we really want to check that NXDOMAIN gets returned, I guess we are stuck with using `dig`.

**Comment by rade** [comment]
The code still returns NOTIMP for ANY queries. I thought we had [agreed](#issuecomment-97016054) we'd always return NXDOMAIN.

File: test/config.sh [hunk 1]
```diff
@@ -88,6 +88,24 @@ exec_on() {
...
+    host=$1
+    cont=$2
+    query=$3
+    docker -H tcp://$host:2375 exec $cont sh -c "dig $query"
+}
+
+assert_dig_on() {
+    ok=$(dig_on "$1" "$2" "+short $3")
+    assert "echo $ok" "$4"
+}
```
rade: this function isn't used. remove.

test/config.sh
```diff
@@ -88,6 +88,24 @@ exec_on() {
...
+
+# assert_dns_status <host> <container> <name> <query> <status>
+assert_dns_status() {
+    res=$(dig_on "$1" "$2" "$3" | grep -q "status: $4" && echo "FOUND")
+    assert "echo $res" "FOUND"
```
rade: Combine these two lines into an `assert_raises`, then you won't need the echo.

inercia:
I thought we had agreed to return NOTIMP on ANY queries... anyway, I will remove anything that has to do with NOTIMP

PR #590, opened 2015-04-22
add exec_on test helper

rade:
so we don't repeat the remote docker invocation command (complete with hard-coded port) over and over again.

### DIFF
--- a/test/100_cross_hosts_test.sh
+++ b/test/100_cross_hosts_test.sh
@@ -18,7 +18,7 @@ weave_on $HOST2 launch $HOST1
 
 weave_on $HOST2 run $C2/24 -t --name=c2 ubuntu
 weave_on $HOST1 run $C1/24 -t --name=c1 ubuntu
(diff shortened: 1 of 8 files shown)


rade:
Note that there is a bug here... there are a bunch of tests that used to run as `sh -c`. I stripped that off. But now the error handling doesn't work, i.e. the scripts just terminate at that point. Alas it turns out that was the case before too! I experimented with a few variations but no luck. @awh I welcome suggestions on how to fix that. I wonder whether we should just test the exit status, and dispense with the "&& echo ok".

awh:
I encountered this yesterday - I think it's caused by the `set -e` at the top of `config.sh`, which is sourced into all the test scripts. As a consequence a statement such as
````
ok=$(docker -H tcp://$HOST1:2375 exec c1 sh -c "ping -q -c 4 $C2 >&2 && echo ok")
````
would cause the script to exit abruptly if the ping fails due to it returning a non-zero status. I'll propose a solution as part of the work I'm doing to make `run_all.sh` behave nicely in the case of test failures...

PR #595, opened 2015-04-22
shrink dns tests and prevent exit on failure

rade:


awh:
I've confirmed that the non-zero exit status returned by the `getent` calls in the first commit do trigger the
````
---= !!!ABNORMAL TEST TERMINATION!!! =---
````
message.

PR #606, opened 2015-04-24
Factor out test fixture teardown

### LINKED ISSUE #592: `100_cross_hosts_test.sh` succeeds only once, requiring reboot of test VMs to work again

awh:
Fixes #592.

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -95,6 +95,12 @@ assert_dns_record() {
 }
 
 start_suite() {
+    for HOST in $HOST1 $HOST2; do
+        echo "Cleaning up on $HOST: removing all containers and resetting weave"
+        CONTAINERS=$(docker -H tcp://$HOST:2375 ps -aq)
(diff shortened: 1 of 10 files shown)


test/config.sh
```diff
@@ -95,6 +95,12 @@ assert_dns_record() {
 }
 
 start_suite() {
+    for HOST in $HOST1 $HOST2; do
```
awh: I ended up putting this in `start_suite` rather than `end_suite` as the latter isn't run in cases of abnormal test case termination, leaving the system in a broken state in which none of the tests case run.

test/config.sh
```diff
@@ -95,6 +95,12 @@ assert_dns_record() {
...
+        echo "Cleaning up containers on $HOST and resetting weave"
```
rade: "Cleaning up on $HOST: removing all containers and resetting weave"

awh:
Updated log message.

PR #609, opened 2015-04-26
refactor: eliminate duplication in nameserver query handlers

rade:


rade:
It will likely be easier to review the individual commits instead of the entire diff.

nameserver/server.go
```diff
@@ -215,70 +215,49 @@ func (s *DNSServer) Stop() error {
...
+		notUsHandler(w, r)
 	}
+
+	return s.commonQueryHandler(proto, "Reverse query", dns.TypePTR, zoneLookup, fallback)
 }
 
-func (s *DNSServer) rdnsHandler(proto dnsProtocol) dns.HandlerFunc {
-	fallback := s.notUsHandler(proto)
+func (s *DNSServer) commonQueryHandler(proto dnsProtocol, kind string, qtype uint16,
+	zoneLookup func(ZoneLookup, *dns.Question, *dns.Msg) (*dns.Msg, []ZoneRecord, error), fallback dns.HandlerFunc) dns.HandlerFunc {
```
inercia: A rather long method signature. Maybe we could reduce it by defining a type:

``` Go
type zoneLookupFunc func(ZoneLookup, *dns.Question, *dns.Msg) (*dns.Msg, []ZoneRecord, error)

func (s *DNSServer) commonQueryHandler(proto dnsProtocol, kind string, qtype uint16, zoneLookup zoneLookupFunc, fallback dns.HandlerFunc) dns.HandlerFunc {...
```

nameserver/server.go
```diff
@@ -215,70 +215,49 @@ func (s *DNSServer) Stop() error {
 }
 
 func (s *DNSServer) queryHandler(proto dnsProtocol) dns.HandlerFunc {
```
inercia: Maybe we could go further and rename these functions for better names (eg, `queryHandler`? all these methods are going to handle queries...). I would suggest something like: `queryHandler` -> `localAHandler`, `rdnsHandler` -> `localPTRHandler`, `commonQueryHandler` -> `localHandler`

nameserver/server.go
```diff
@@ -215,70 +215,49 @@ func (s *DNSServer) Stop() error {
...
+	}
 
-		Info.Printf("[dns msgid %d] No results for type %s query %s [caching no-local]",
-			r.MsgHdr.Id, dns.TypeToString[q.Qtype], q.Name)
-		s.cache.Put(r, nil, negLocalTTL, CacheNoLocalReplies, now)
-		w.WriteMsg(makeDNSFailResponse(r))
+	notUsHandler := s.notUsHandler(proto)
+	fallback := func(w dns.ResponseWriter, r *dns.Msg) {
+		Info.Printf("[dns msgid %d] -> sending to fallback server", r.MsgHdr.Id)
+		notUsHandler(w, r)
```
inercia: A picky comment: we could remove the `notUsHandler` definition and use directly `s.notUsHandler(proto)` here, right?

rade: no, because `s.notUsHandler` returns a function. And we only want to construct that function once, not on every fallback.

nameserver/server.go
rade: I suggest you do that separately, after this is merged.

nameserver/server.go
rade: The type would only be used in one place, so no point.

nameserver/server.go
inercia: Ok, I'll do...

nameserver/server.go
inercia: I agree.

nameserver/server.go
inercia: I see...

PR #612, opened 2015-04-28
Post docs move

errordeveloper:


rade:
There are a few things here and in the broader docs move project that do not quite make sense to me, but perhaps that's just due to lack of explanation...

1. Why does the README link in https://github.com/weaveworks/weave/wiki/WorkingOnWeave#working-on-docs point at the head rather than https://github.com/weaveworks/weave/tree/latest_release?
2. I pointed out before that the "Installation Notes" heading in http://docs.weave.works/weave/latest_release/index.html is misleading since it gives the impression that the only two ways to install weave is via boot2docker and systemd. I'd bury that information; it has no place on the main docs page. Instead make the description of the link to the README read "Introduction, Installation and Example".
3. Where can the docs for the 'master' version be found? Is the idea that we, and any of our users working with snapshots, would simply be looking at the .md files in the repo? That's ok in principle, except that none of the cross-file links work!
4. https://github.com/weaveworks/weave/wiki/WorkingOnWeave#working-on-docs needs updating.
5. What is the process for updating the published docs? e.g. when we correct typos and make other improvements. We need to be able to do that.

errordeveloper:
1. I didn't realised that this was documented in the wiki, will update it
  2. I and @fintanr will be reworking the docs soon, but I don't see a reason for making any half-way changes right now
  3. As per #604, the docs for master branch can be found at [docs.weave.works/weave/master](http://docs.weave.works/weave/master/)
  4. see point 1
  5. This question had been raised, which might had been not explicit enough; so in order to update docs for a given release we will need to have those changes checked-in and thereby we will require release branches to accomplish that

rade:
re 5... we really need that process documented, step-by-step on the wiki. including instructions for how to make a change to the published docs *and* have the same change go into master. It's probably all just some git command sequences, but they are outside the ordinary day-to-day workflow and hence easy to get wrong.

rade:
re 2... it's not a half-way change. The bit I am complaining about does not appear at https://github.com/weaveworks/weave#find-out-more; it's something that got introduced elsewhere. You could just remove it.

rade:
Incidentally, 5 is already an issue now. A bunch of changes have been made on the 'master' docs which equally apply to the released version and hence really, really should be published on http://docs.weave.works/weave/latest_release/index.html too. So we also need a documented way of backporting docs from master to latest_release.

errordeveloper:
@rade c9bff8b adds basic logic for updating docs at `docs.weave.works/weave/latest_release/`. I am going to update the wiki page to describe the workflow. I also would like to make this update as soon as this pull-request is merged, are all docs on master correct for `latest_release` at the moment?

rade:
> re all docs on master correct for `latest_release` at the moment?

No, if you document how to cherry pick from master to latest_release, I'll do the rest.

errordeveloper:
The wiki should have [sufficient information](https://github.com/weaveworks/weave/wiki/WorkingOnWeave#working-on-docs) now, and we can always refine that later anyway.

rade:
This LGTM. I have fixed a few typos in the docs. All of my points except 2 have been addressed. I've raised #640 for that.

I'd like @paulbellamy to take a look at the mechanics.

README.md
```diff
@@ -131,11 +131,11 @@ to each other.
...
- * [Features](http://weaveworks.github.io/weave/features.html)
- * [Troubleshooting](http://weaveworks.github.io/weave/troubleshooting.html)
- * [Building](http://weaveworks.github.io/weave/building.html)
- * [How it works](http://weaveworks.github.io/weave/how-it-works.html)
+ * [Documentation homepage](http://docs.weave.works/weave/latest_release/)
+ * [Features](http://docs.weave.works/weave/latest_release/features.html)
+ * [Troubleshooting](http://docs.weave.works/weave/latest_release/troubleshooting.html)
+ * [Building](http://docs.weave.works/weave/latest_release/building.html)
+ * [How it works](http://docs.weave.works/weave/latest_release/how-it-works.html)
  * [Automatic Discovery with WeaveDNS](https://github.com/weaveworks/weave/tree/master/weavedns#readme)
```
paulbellamy: Shouldn't this point to latest_release also?

site/build-and-publish.sh
```diff
@@ -13,6 +13,10 @@ if [ -n "${TRAVIS_TAG}" ]; then
   OUTPUT="weave/${TRAVIS_TAG}"
 fi
 
+if [ "${BRANCH}" = "latest_release_doc_updates" ]; then
```
paulbellamy: I get this shouldn't be master, because it would pull in code changes before release, but I worry this branch will be neglected/forgotten.

paulbellamy:
What is the process for updating docs on a previous release?

rade:
> What is the process for updating docs on a previous release? Is that something we need to worry about?

We don't.

README.md
errordeveloper: well spotted

site/build-and-publish.sh
rade: hmm. good point. What happens when we make a new release? won't this branch go stale? Perhaps the release script should include a merge from the release tag to this branch. or s.t. like that.

errordeveloper:
> Or just make another release (e.g. 0.9.0.1)

Sort of could, but the idea of `latest_release_doc_updates` is to keep the process to the minimum. Also, a patch release would give an impression to the user that they have to pull latest binaries, while there is nothing new there for them.

> ...but I worry this branch will be neglected/forgotten.

Not unlikely, but the worst that can happen is that `latest_release` docs have some typos of sorts. The only better way I can think of is to move docs to a separate repository which can have a slightly different branching structure to it.

rade:
I am completely lost now. I have no idea how latest_release_doc_updates works. Is what gets published under the latest_release URL a *merge* of the latest_release_doc_updates branch into latest_release?

errordeveloper:
> Is what gets published under the latest_release URL a merge of the latest_release_doc_updates branch into latest_release?

A commit with tag `latest_release` is primary cause of release doc update. Secondary cause is a commit on  `latest_release_doc_updates`.

A tag push would occur as result of someone running `bin/release`.
To update docs after the release you can either push directly to `latest_release_doc_updates` branch or make a pull-request to it.

rade:
So this whole process relies on ci to be operational *at the time a change is pushed*?

errordeveloper:
Yes, as otherwise the process would have much more dependencies. We can probably package up a container that developer can use in case CI is screwed, if that worries you.

paulbellamy:
@ilya:
Ah, ok. I hadn't realized that bin/release was updating the latest_release tag.
I do like having the docs tied into the main repo, as it keeps them in sync with the release process. In practice updating docs separately from a release may not be an issue. Let's wait and deal with that if it becomes a problem.

I'm assuming that the gsutils cp command will also be able to remove files?

@rade:
I'm less worried about depending on the CI being available, because you could just re-run the build when it comes back.

paulbellamy:
Also, we should back-fill the docs for older releases.

rade:
> Also, we should back-fill the docs for older releases.

Not bothered. Except, maybe, for 0.10.

paulbellamy:
Ok, pending the ```gsutils cp``` removing files, this LGTM.

errordeveloper:
> Not bothered. Except, maybe, for 0.10.

It's [there already](http://docs.weave.works/weave/v0.10.0/index.html).

rade:
> Yes, as otherwise the process would have much more dependencies.

I am not suggesting that we should be able to publish docs when CI isn't operational. But it would be nice if CI *picked up* any doc changes that may have been pushed while it was down.

errordeveloper:
@paulbellamy yeah, should probably switch to `gsutil rsync`, but I haven't tested that...

paulbellamy:
Cool, @errordeveloper pointed out that if nothing links to the files anymore it doesn't matter if they're left around. so this LGTM.

PR #615, opened 2015-04-28
Background updates for names in the Zone database

inercia:
This PR implements
- an improved Zone database, where names can have multiple IPs.
  - The Zone database can also store (and differentiate) locally introduced information as well as data obtained from remote peers (with mDNS)
  - Names and IPs accept _observers_ that will we invoked when they change, currently used for cache invalidation but with more potential use cases (eg, remote peers notifications)
  - The mDNS client and server has been moved inside the Zone database. The Zone database will use them for obtaining missing information, for keeping information up to date, etc (this integration between mDNS and the Zone database could be even tighter in the future...)
  - The _mDNS server_ responds with all the answers available in the Zone database, and the _mDNS client_ can make use of multiple answers from other peers for adding information to the local database.
  - The internal structure has also been improved by using maps instead of slices for storing records.
- an _update mechanism_ for keeping names in the database up to date. So far, queries about information not in the database resulted in mDNS queries. Clients would then obtain the first response WeaveDNS could get from other peers, as waiting for more responses would also mean more latency. In this new mechanism, we also trigger an immediate update request for that name in order to get all the other IPs we didn't wait for, and we will repeat that update periodically until either a) there is no local interest in that name or b) no other peer answers our request (more details [here](https://github.com/inercia/weave/blob/weave-583-2/nameserver/zone_lookup.go#L198))
- vastly improved unit tests, with
  - use of an external clock provider, specially useful for testing expirations in the cache and names updates (eg, names that appear and disappear in the network).
  - more mocks, in particular the new mocked mDNS client and server (an external, mocked clock provider would lead to unexpected results with real network senders&receivers)

### DIFF
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -2,6 +2,8 @@ package nameserver
 
 import (
 	"github.com/miekg/dns"
+	"math/rand"
+	"time"
 )
 
 const (
@@ -85,4 +87,21 @@ func getMaxReplyLen(r *dns.Msg, proto dnsProtocol) int {
 		maxLen = int(opt.UDPSize())
 	}
 	return maxLen
+}
+
+
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
+}
+
+// shuffleAnswers reorders answers for very basic load balancing
+func shuffleAnswers(answers []dns.RR) []dns.RR {
+	if len(answers) > 1 {
+		for i := range answers {
+			j := rand.Intn(i + 1)
+			answers[i], answers[j] = answers[j], answers[i]
+		}
+	}
+
+	return answers
 }
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -89,7 +89,6 @@ func getMaxReplyLen(r *dns.Msg, proto dnsProtocol) int {
 	return maxLen
 }
 
-
 func init() {
 	rand.Seed(time.Now().UTC().UnixNano())
 }
@@ -103,5 +102,14 @@ func shuffleAnswers(answers []dns.RR) []dns.RR {
 		}
 	}
 
+	return answers
+}
+
+// only take the first `num` answers
+func pruneAnswers(answers []dns.RR, num int) []dns.RR {
+	if num > 0 && len(answers) > num {
+		// TODO: we should have some prefer locally-introduced answers, etc...
+		return answers[:num]
+	}
 	return answers
 }
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -33,6 +33,8 @@ func makeTruncatedReply(r *dns.Msg) *dns.Msg {
 	return reply
 }
 
+type DNSResponseBuilder func(r *dns.Msg, q *dns.Question, addrs []ZoneRecord) *dns.Msg
+
 func makeAddressReply(r *dns.Msg, q *dns.Question, addrs []ZoneRecord) *dns.Msg {
 	answers := make([]dns.RR, len(addrs))
 	header := makeHeader(r, q)
--- a/nameserver/mocks_test.go
+++ b/nameserver/mocks_test.go
@@ -2,6 +2,7 @@ package nameserver
 
 import (
 	"fmt"
+	"github.com/benbjohnson/clock"
 	"github.com/miekg/dns"
 	. "github.com/weaveworks/weave/common"
 	wt "github.com/weaveworks/weave/testing"
@@ -419,6 +420,22 @@ func runLocalTCPServer(laddr string, handler dns.HandlerFunc) (*dns.Server, stri
 
 	Debug.Printf("[mocked fallback] Fallback TCP server listening at %s", l.Addr().String())
 	return server, l.Addr().String(), nil
+}
+
+//////////////////////////////////////////////////////////////////
+
+type mockedClock struct {
+	*clock.Mock
+}
+
+func newMockedClock() *mockedClock {
+	return &mockedClock{clock.NewMock()}
+}
+
+func (clk *mockedClock) Forward(secs int) {
+	Debug.Printf(">>>>>>> Moving clock forward %d seconds - Time traveling >>>>>>>", secs)
+	clk.Add(time.Duration(secs) * time.Second)
+	Debug.Printf("<<<<<<< Time travel finished! We are at %s <<<<<<<", clk.Now())
 }
 
 //////////////////////////////////////////////////////////////////
--- a/nameserver/mocks_test.go
+++ b/nameserver/mocks_test.go
@@ -1,9 +1,15 @@
 package nameserver
 
 import (
+	"fmt"
+	"github.com/miekg/dns"
 	. "github.com/weaveworks/weave/common"
+	wt "github.com/weaveworks/weave/testing"
 	"net"
+	"strconv"
 	"sync"
+	"testing"
+	"time"
 )
 
 const (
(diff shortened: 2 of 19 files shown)


nameserver/server.go
```diff
@@ -54,51 +65,64 @@ func (proto dnsProtocol) String() string {
...
 	switch proto {
 	case protTCP:
-		return &dns.Client{Net: "tcp"}
+		return &dns.Client{Net: "tcp", ReadTimeout: timeout}
 	case protUDP:
-		return &dns.Client{Net: "udp", UDPSize: uint16(bufsize)}
+		return &dns.Client{Net: "udp", ReadTimeout: timeout, UDPSize: uint16(bufsize)}
 	}
 	return nil
 }
```
inercia: `server.go` has changed so much that I doubt it would be worth to review it by using the diff... The same applies to `zone.go`

awh:
@alvaro

This is far, far too much change for a single PR which iterates on existing functionality - it's comparable in size to the IPAM PR, which is adding an entire new subsystem. Here's what I think we need to do:

### Push some of the commits straight to master

You don't need to get absolutely everything reviewed - small obvious stuff can go straight on master:

* _c726a95 - Minor: check the "assert" argument is not "nil"._
* _b903870 - Restored "fallback" argument for forcing a fallback server._

### Pull some commits out as separate PRs

* _32084ce - Use a general signal handler._ This PR should also refactor the weave router to make use of the extracted common signal handler.
* _bb76c7c - Do not invoke the mDNS lookups from the server._ This commit has a nice descriptive log message, but the changes are hard to review all mixed together. Pull this commit out as a single PR, and restructure it with ~one commit per bullet point of the commit message.
* 82cbd6d - Improved Zone database. This is a monster! Each of the five sub-bullet points describing this in the initial PR comment (and the 'update mechanism' bullet) need to be pulled out as individual pull requests, with granular commits as appropriate. 

There are a few other sundry small commits in this PR that introduce interfaces and make some test improvements, I'm not sure what the dependencies are between those and the above so I'll leave you to distribute them across the restructured PRs as you see fit.

I appreciate that this is a pain, and impedes your forward progress, but we need to break out of this cycle where the review backlog is so far behind your current progress. I'm afraid that if we don't, by the time I struggle through reviewing this you'll have built up another ~3k line changeset and we'll be back where we started. I'm hopeful that we can break this cycle, get the backlog down, and start iterating rapidly on small PRs...

awh:
Another thought regarding types of PR: as a reviewer it's very useful to know whether a PR is intended to change user visible behaviour or not. Refactoring PRs that don't change external functionality are absolutely fine; PRs which change behaviour should _almost always_ be associated with an issue, and be clearly marked - the easiest way of doing this is to put a 'Fixes #NNN' message at the top of the PR comment, that way the issue is closed automatically on merge...

inercia:
@awh Apologies if this PR is too big again. I understand there is a lot of code and it is difficult to review. I understand your request for smaller PRs, but I'm not so fond of having so many PRs as you suggest. I think it makes sense in big, stable projects (eg, the Linux kernel) where you do not want to rush for new features and the main goal is to move between very stable points (eg, points where you can get a new release) through incremental changes.

I think that, in a project like this, too many PRs frequently lead to a slower reviewing process (eg, more comments (for things that could change in next PR), more iterations, etc). In my opinion, it makes more sense to work on PRs that focus on specific features, start and end in stable points in the history (eg, we will never get a release from a commit in the middle of this PR!), while being easy to understand and review.

All this said, I'm sure I can make it easier to review this PR. I will try again and redo the history in the PR, so let me know what you think about this plan:

* _c726a95 - Minor: check the "assert" argument is not "nil"._ I will move this commit to master.
* _b903870 - Restored "fallback" argument for forcing a fallback server._ I will move this commit to master.
* _32084ce - Use a general signal handler._  My plan was to get this into master with this PR, and then create another PR where the router would use it too. But I can do it as you suggested and put all together in a separate PR.
* _82cbd6d - Improved Zone database._ There is a problem here, as I think it is difficult to break in different commits (or PRs) as you suggest: the whole ZoneDb internal structure has changed from a slice of records to a tree, so the main chunk must be an atomic change (and this can be about 80% of the code in this commit), and I think that the new [zone.go](https://github.com/weaveworks/weave/blob/82cbd6d9d864158bfb4a251c7402336fc55d70f4/nameserver/zone.go) is relatively easy to understand. I can move to different commits things like the names updater, maybe some other things too... I will try it.
* _bb76c7c - Do not invoke the mDNS lookups from the server._ Once the ZoneDb is responsible for the mDNS subsystem, the server is simplified a lot. And, while this change could seem hard to review by looking at the diff, the code is, in fact, much easier to understand now if you look to the new [server.go](https://github.com/weaveworks/weave/blob/bb76c7c92f4f969feffe1bcbc5359bdc3b731e31/nameserver/server.go). The rest of the commit are unit tests. I will try to split this commit somehow...

rade:
Does this PR resolve any of the open issues relating to DNS?

inercia:
@rade This would address #338 (return all matching local records) and partially #226 (dns load balancing)

rade:
which bits of #226 will be left?

inercia:
@rade ... the "fault resilience" part of the #226

rade:
@alvaro In my mind there is no difference between the load balancing and fault resilience aspects of #226, wrt weaveDNS functionality. What weaveDNS functionality were you thinking of that is specifically targeted at fault tolerance?

inercia:
@rade In its basic form, we can use load-balancing as some form of fault resilience. But, in the long term, we have a privileged position for knowing the state of the network and sending always up-to-date responses to DNS clients (with tiny TTLs, probably equal to the propagation time of this information in the network). And I also think it would be a good thing to support SRV records at some moment...

rade:
@inercia As per its title,  #226 is about *basic* load balancing and fault resilience. What you are describe is more advanced; please open separate issues.

awh:
General question - what do you think about breaking out name/nameset/zonedb (and possibly others) and their related methods into eponymously named files? One thing I've noticed (across the whole project, not just weaveDNS) is that we don't tend to make much use of packages (particularly nested ones), and have relatively few files compared to what you might see in other languages. I'm new to golang so it may just be my one-class-per-file Java/Scala prejudices showing :D Having directories or files named for the structures they contain does certainly aid navigability for those that aren't intimately familiar with the codebase though... thoughts?

rade:
The [docs](https://github.com/weaveworks/weave/tree/latest_release/weavedns#registering-multiple-containers-with-the-same-name) need updating.

inercia:
@rade Docs updated in 42be1a2

rade:
Given that this PR *does* add new user-visible functionality, I'd expect there to be some addition to the smoke tests.

inercia:
@rade It makes sense. I will include some smoke tests...

nameserver/zone_lookup.go
```diff
@@ -0,0 +1,301 @@
...
+
+	if len(res) > 0 {
+		Debug.Printf("[zonedb] '%s' resolved in local database", name)
+	} else {
+		res, err = zone.domainLookup(name, zone.mdnsCli.LookupName)
+	}
+
+	if len(res) > 0 {
+		zone.startUpdatingName(name)
+		return res, nil
```
inercia: We should make responses unique at this point. Otherwise, `['ident1', 'name1', IP1]` and `['ident2', 'name1', IP1]` would produce `[IP1, IP1]`

nameserver/zone_lookup.go
```diff
@@ -0,0 +1,301 @@
...
+
+	if len(res) > 0 {
+		// note: even for reverse addresses, we perform the background updates in the name, not in the IP
+		//       this simplifies the process and produces basically the same results...
+		// note: we do not spend time trying to update names that did not return an initial response...
+		for _, r := range res {
+			zone.startUpdatingName(r.Name())
+		}
+
+		return res, nil
```
inercia: Again, we should make responses unique at this point. Otherwise, `['ident1', 'name1', IP1]` and `['ident2', 'name1', IP1]` would produce `['name1', 'name1']`

inercia: Fixed in b23a6bb

nameserver/zone_lookup.go
inercia: Fixed in b23a6bb

nameserver/dns.go
```diff
@@ -92,3 +96,18 @@ func getMaxReplyLen(r *dns.Msg, proto dnsProtocol) int {
 	}
 	return maxLen
 }
+
+// shuffleAnswers reorders answers for very basic load balancing
+func shuffleAnswers(answers []dns.RR) []dns.RR {
+	n := len(answers)
+	if n > 1 {
+		rand.Seed(time.Now().UTC().UnixNano())
```
awh: This is typically done only once per program instantiation. You might want to think about abstracting it too (as you have done for the clock) so that you can have deterministic behaviour in your unit tests...

rade: http://marcelom.github.io/2013/06/07/goshuffle.html

nameserver/mocks_test.go
```diff
@@ -89,3 +96,403 @@ func (mz *mockedZoneWithRecords) ObserveInaddr(inaddr string, observer ZoneRecor
...
+}
+func newMockedMDNSServerWithRecords(zrs []ZoneRecord) *mockedMDNSServer {
+	return newMockedMDNSServer(newMockedZoneWithRecords(zrs))
+}
+
+//////////////////////////////////////////////////////////////////
+
+// A mocked mDNS client.
+// This mock asks a group of (potentially mocked) mDNS servers
+type mockedMDNSCLient struct {
```
awh: Should be `mockedMDNSClient` not `mockedMDNSCLient` (The 'L' in CLient should be lower case)

nameserver/zone_lookup.go
```diff
@@ -0,0 +1,342 @@
...
+		uniq.add(r)
+	}
+	zone.mx.Unlock()
+
+	return uniq.toSlice(), nil
+}
+
+// Perform a lookup for a name in the zone
+// The name can be resolved locally with the local database or with some other resolution method (eg, a mDNS query)
+func (zone *zoneDb) DomainLookupName(name string) (res []ZoneRecord, err error) {
```
awh: If I am reading this correctly then the very first time I perform a query on a given weaveDNS peer:
* If the name is in the local database, then I will receive only those local addresses
* If the name is not in the local database, then I will receive only addresses from the first weaveDNS peer to respond by mDNS (due to use of `zone.mdnsCli.LookupName` on line 145)

Is that right? The second case is of particular concern, because I'm likely to encounter it the very first time I try to play with load balancing - if I set up two server containers on different peers and give them both the same FQDN, then try to resolve that name from a client container on a third peer I will receive only one of the addresses the first time round. I understand there is a latency tradeoff here, but I think it would be better to use `InsistentLookupName` and put up with the slight pause on the first query...

inercia: You are right: for the first question for a name, we reply with the first answer we can get (and then we start the background periodic update for that name). It is not only a question of latency, is also a question of doing a incremental change on the way we resolve names: the resolution will be as we have being doing so far, but with an additional background update...

As you said, I could replace it by `InsistentLookupName`, or maybe switch between both behaviors with a command line argument... both changes would be trivial (there would be more changes in the unit tests though). What do you guys (@rade @awh) think?

nameserver/zone.go
```diff
@@ -3,151 +3,761 @@ package nameserver
...
 const (
 	RDNSDomain = "in-addr.arpa."
 )
 
+const (
+	DefaultLocalDomain     = "weave.local." // The default name used for the local domain
+	DefaultRefreshInterval = int(localTTL)  // Period for background updates with mDNS
+	DefaultRelevantTime    = 60             // When to forget info about remote info if nobody asks...
+
+	defaultNumUpdaters    = 1              // Number of background updaters
```
awh: Perhaps this should default to something higher? `InsistentLookupName` is blocking...

inercia: Yes, `InsistentLookupName` is a blocking operation, but I'm not sure what the right number would be... `InsistentLookupName` takes 500ms, so we can refresh 2 names per second... maybe I should increase this number, set some low, safe value (4 workers?) and add a command line argument...

nameserver/mocks_test.go
inercia: Fixed

nameserver/dns.go
inercia: Fixed. Added a init() function as @awh suggested, and changed the algorithm to the one provided by @rade.

rade:
needs re-basing.

afaict the docs didn't get updated to reflect the 'pick one' behaviour.

nameserver/server.go
```diff
@@ -111,207 +140,227 @@ func NewDNSServer(config DNSServerConfig, zone Zone, iface *net.Interface) (s *D
...
-			}
+		if answers, err := lookup(q.Name); err != nil {
+			Info.Printf("[dns msgid %d] No results for type %s query for '%s' [caching no-local]",
+				r.MsgHdr.Id, dns.TypeToString[q.Qtype], q.Name)
+			maybeCache(nil, s.negLocalTTL, CacheNoLocalReplies)
+			fallback(w, r)
+		} else {
+			m := msgBuilder(r, &q, answers)
+			m.Authoritative = true
+			m.Answer = pruneAnswers(shuffleAnswers(m.Answer), s.maxAnswers)
```
awh: Should this not happen *after* we've `maybeCached`d the answer list?

inercia: You are absolutely right. Good catch! I'll fix it...

PR #617, opened 2015-04-28
replace random nonces with message sequence numbers

### LINKED ISSUE #564: crypto errors can stall udp receiver loop
Receiving a stray udp packet can cause the decryptor to wait for a new nonce which will never arrive. This stalls the entire udp receiver loop for 30 seconds, until the offending connection gets shut down by heartbeat timeouts. That of course will cause all other connections to miss heartbeats too and shut down too. Not great.

rade:
Fixes #564.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -446,7 +446,7 @@ func (conn *LocalConnection) receiveTCP(decoder *gob.Decoder) {
 	usingPassword := conn.SessionKey != nil
 	var receiver TCPReceiver
 	if usingPassword {
-		receiver = NewEncryptedTCPReceiver(conn.SessionKey)
+		receiver = NewEncryptedTCPReceiver(conn.SessionKey, conn.outbound)
 	} else {
 		receiver = NewSimpleTCPReceiver()
 	}
--- a/router/connection.go
+++ b/router/connection.go
@@ -489,11 +489,6 @@ func (conn *LocalConnection) handleProtocolMsg(tag ProtocolTag, payload []byte) 
 			nil)
 	case ProtocolFragmentationReceived:
 		conn.setStackFrag(true)
-	case ProtocolNonce:
-		if conn.SessionKey == nil {
-			return fmt.Errorf("unexpected nonce on unencrypted connection")
-		}
-		conn.Decryptor.ReceiveNonce(payload)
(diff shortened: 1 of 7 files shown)


rade:
I did some benchmarking of this today. Running `iperf` I observed a ~1% drop compared to master, based on the highest throughput figure from 10 runs. The variance is much greater than 1% though, so I'm not going to read much into that result, and in any case a 1% drop would be perfectly acceptable. It is certainly possible that there might be a slight performance decrease with this change, since we have increased the encapsulation overhead by 6 bytes (2 byte offset in the old code vs 8 byte message sequence number here).

I also ran `multiweave`, attempting to determine how many weaves I can get to fully establish connections. On master I get as far as 32. On this branch to 50. The reason is likely twofold. Firstly, on master we run into #564, which this fixes. Secondly, crypto on the tcp connection should perform much better on this branch, and since connection establishment is mostly about the tcp connections, that likely makes a difference in the test.

dpw:
Ok, I'm going to add a few specific comments about the current `advanceState` implementation. And then explain how the current windowing approach offends my sense of micro-optimization, and should all be different.

rade:
> We never seem to actually use `di.highestSeqNoSeen` except to calculate its window number here.

Not quite. We have the following in the `default` case:
````
		if seqNo > di.highestSeqNoSeen {
			di.highestSeqNoSeen = seqNo
		}
````

rade:
Almost all fixed in the above commit. I will squash that, but figured it would be better to have it around temporarily so it's obvious what i've changed.

Good call on the naming and dispatching on the delta. And good catch on the overflow.

Re `di.currentWindow` - see comment above

Re ordering... I thought about this a fair bit when I wrote the code and experimented with a few variations. In the end I settled on what I have because the `default` case is special. It sticks out visually, due to the keyword and coming last. I therefore chose to make it the logically special case of the five. And that is the 0 case, which a) is the common case, b) the only case with a delta comparison not involving +/-1, and c) the only case where we have a nested conditional.

dpw:
> Not quite. We have the following in the default case:

Yes.  But why can't that be turned into `di.currentWindow = window` in the appropriate places?

rade:
ah true. Just realised that myself :)

Issue #519, opened 2015-04-06
four layers of gob is bad


rade:
The sending of topology gossip currently goes through four layers of gob encoding, and involves n_peers+3 encoders...

1. one GobEncoder for *each* peer, into which we encode the connections
2. one GobEncoder for the set of peers, into which all peers are
   encoded, with their connection information represented by the
   result of the aforementioned encoding.
3. one GobEncoder for encoding the gossip (hash, source, and the
   result of (2)) as a ProtocolMsg payload.
4. one GobEncoder in the TCPSender, which used to encode a byte
   slice - comprising the ProtocolMsg tag and payload - as a single
   message

This has a bad design smell to it and probably is also rather inefficient.

rade:
Two layers of gob is the best we can do when crypto is enabled. Our crypto operates on message streams, i.e. streams of byte slices, rather than byte streams. The 4th gob layer converts between the two. And the 3rd layer converts between between our protocol data objects and byte slices.

We could do something different in the non-crypto case but it's really not worth the hassle.

rade:
I've done some more profiling, running
````
NUM_WEAVES=1 bin/multiweave launch -connlimit 100 -profile /tmp
NUM_WEAVES=49 FIRST_WEAVE=2 bin/multiweave launch -connlimit 100
```
(wait until everything goes quiet - takes about 3 minutes on my machine)
```
sudo kill -INT $(docker inspect --format '{{.State.Pid}}' weave1)
docker cp weave1:/tmp/cpu.pprof .
go tool pprof --nodefraction=0.02 --edgefraction=0.01 weaver/weaver cpu.pprof 
````

The gob codec takes up ~12%, in what is an extremely gob-heavy scenario. So while there is still room for improvement, it's not worth spending a lot of effort on it. So the main motivation to eliminate another layer would be improved design/APIs.

PR #619, opened 2015-04-29
Use same gob encoder/decoder for peers and peer connections

awh:
Partially addresses #519 by merging the first two layers of gob.

### DIFF
--- a/router/peers.go
+++ b/router/peers.go
@@ -32,6 +32,13 @@ type ConnectionSummary struct {
 	Established   bool
 }
 
+type PeerSummary struct {
+	NameByte []byte
+	NickName string
+	UID      uint64
+	Version  uint64
+}
+
 func NewPeers(ourself *Peer, onGC func(*Peer)) *Peers {
 	return &Peers{
 		ourself: ourself,
@@ -197,15 +204,15 @@ func (peers *Peers) decodeUpdate(update []byte) (newPeers map[PeerName]*Peer, de
 	decoder := gob.NewDecoder(updateBuf)
 
 	for {
-		nameByte, nickName, uid, version, connSummaries, decErr := decodePeer(decoder)
+		peerSummary, connSummaries, decErr := decodePeer(decoder)
 		if decErr == io.EOF {
 			break
 		} else if decErr != nil {
 			err = decErr
 			return
 		}
-		name := PeerNameFromBin(nameByte)
-		newPeer := NewPeer(name, nickName, uid, version)
+		name := PeerNameFromBin(peerSummary.NameByte)
+		newPeer := NewPeer(name, peerSummary.NickName, peerSummary.UID, peerSummary.Version)
 		decodedUpdate = append(decodedUpdate, newPeer)
 		decodedConns = append(decodedConns, connSummaries)
 		existingPeer, found := peers.table[name]
@@ -216,7 +223,6 @@ func (peers *Peers) decodeUpdate(update []byte) (newPeers map[PeerName]*Peer, de
 			return
 		}
 	}
-
 
 	for _, connSummaries := range decodedConns {
 		for _, connSummary := range connSummaries {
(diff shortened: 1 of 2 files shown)


router/peers.go
```diff
@@ -25,6 +25,13 @@ type NameCollisionError struct {
 
 type PeerNameSet map[PeerName]struct{}
 
+type ConnectionView struct {
```
awh: Would welcome suggestions for a better name... `GobConnection`?

rade: `ConnectionData`?

router/peers.go
```diff
@@ -233,7 +237,7 @@ func (peers *Peers) decodeUpdate(update []byte) (newPeers map[PeerName]*Peer, de
 	return
 }
 
-func (peers *Peers) applyUpdate(decodedUpdate []*Peer, decodedConns [][]byte) map[PeerName]*Peer {
+func (peers *Peers) applyUpdate(decodedUpdate []*Peer, decodedConns [][]ConnectionView) map[PeerName]*Peer {
 	newUpdate := make(map[PeerName]*Peer)
 	for idx, newPeer := range decodedUpdate {
 		connsBuf := decodedConns[idx]
```
rade: `connsBuf` needs renaming.

router/peers.go
```diff
@@ -298,48 +304,19 @@ func decodePeerNoConns(dec *gob.Decoder) (nameByte []byte, nickName string, uid
...
-		var established bool
-		if err := dec.Decode(&established); err != nil {
-			return err
-		}
-		fun(nameByte, foundAt, outbound, established)
-	}
-}
-
-func readConnsMap(peer *Peer, buf []byte, table map[PeerName]*Peer) map[PeerName]Connection {
+func readConnsMap(peer *Peer, connViews []ConnectionView, table map[PeerName]*Peer) map[PeerName]Connection {
```
rade: function name needs updating; we are not really 'reading' anything anymore here.

router/peers.go
```diff
@@ -210,21 +217,18 @@ func (peers *Peers) decodeUpdate(update []byte) (newPeers map[PeerName]*Peer, de
...
-				return
+				continue
 			}
 			if _, found := peers.table[remoteName]; found {
-				return
+				continue
 			}
 			// Update refers to a peer which we have no knowledge
 			// of. Thus we can't apply the update. Abort.
 			err = UnknownPeerError{remoteName}
```
rade: We should `return` here, and get rid of the `return` below.

rade:
I am curious whether this change makes any difference to the time it takes `NUM_WEAVES=50 bin/multiweave launch --connlimit 100` to settle. (NB: make sure you bump the kernel setting as described in bin/multiweave).

router/peers.go
bboreham: Fowler calls this kind of thing a "Data Transfer Object"; if we adopted that pattern then `ConnectionDTO`.  `ConnectionData` not too bad but "Data" is not really adding much to anyone's understanding.

rade: Ah yes. I knew these things had a name. `ConnectionDTO` is good.

dpw: "DTO".  Yuck.

ConnectionUpdate?  ConnectionReport?

awh: `ConnectionSummary`?

awh:
I've made the suggested changes as well as introducing `PeerSummary` to tidy up peer encoding/decoding...

rade:
The linter is moaning.

And the inclusion of `ConnectionSummary` in `PeerSummary` needs further-reaching changes to make sense, so I suggest leaving that out.

rade:
This has made a massive improvement to cluster formation. Running `NUM_WEAVES=50 bin/multiweave launch --connlimit=100` and measuring the time it takes from the first connection to quiescence, I get ~7 minutes on master, and ~2 minutes in this branch.

PR #625, opened 2015-05-01
Improve the "Generic error" message when compiled against old libpcap

### LINKED ISSUE #623: weave shouldn't fail with "Generic error" when compiled against an old libpcap
Weave needs pcap >= 1.5.0. The errors we get from gopacket when running with an older version are less than helpful. So we should check that we have the right version. Ideally at build time, but that may be more trouble than it's worth.

dpw:
If gopacket is compiled against an older pcap.h that doesn't have pcap_set_immediate_mode, it supplies a dummy definition that  always returns PCAP_ERROR.  That becomes "Generic error", which is not very helpful.  The real pcap_set_set_immediate never returns PCAP_ERROR, so this turns it into a more informative message.

Fixes #623

### DIFF
--- a/router/pcap.go
+++ b/router/pcap.go
@@ -2,6 +2,7 @@ package router
 
 import (
 	"code.google.com/p/gopacket/pcap"
+	"fmt"
 )
 
 type PcapIO struct {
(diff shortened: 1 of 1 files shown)


bboreham:
Could you put a call to this code in a test in pcap_test.go, so it would then be caught at unit-test time?

dpw:
> Could you put a call to this code in a test in pcap_test.go, so it would then be caught at unit-test time?

If some of the tests start requiring pcap, it will get checked, no?  Is there value in checking even if pcap is not actively used?

Also, it is unlikely that you won't already have a working version of libpcap (Tom found ubuntu 12.04 doesn't).  So it's really about making the error message more obvious in the rare event you don't, rather than checking as such.

dpw:
Having second thoughts on this one.  Probably going to cancel this PR unless I need to repurpose it for a very different patch.

dpw:
I've reimagined this and #623.

We could report the "generic error" as an issue against gopacket, but having looked at their code again, I think improvements on their side are unlikely.  Because their approach is to supply a substitute `pcap_set_immediate_mode` in the C section of pcap.go, they are limited to the existing libpcap error status codes.  They could do things differently in order to provide a better error message, but I could totally understand if they weren't convinced by the cost/benefit of doing so.

PR #627, opened 2015-05-03
add `weave dns-add/remove` commands

grkvlt:
Added two new commands `dns-add` and `dns-remove` that call the `tell_dns` function method to store addresses and optionally a specific FQDN for a container ID.

This is useful for circumstances where containers may be configured with IP addresses outside weave, perhaps by some contaienr orchestration service or using a different SDN, but those addresses should still be resolvable and usable by the weaveDNS.

### DIFF
--- a/test/240_dns_add_name_test.sh
+++ b/test/240_dns_add_name_test.sh
@@ -1,0 +1,29 @@
+#! /bin/bash
+
+. ./config.sh
+
+C1=10.2.0.78
+C2=10.2.0.34
+NAME1=seeone.weave.local
+NAME2=seetwo.weave.local
+
+start_suite "Add and remove names on a single host"
+
+weave_on $HOST1 launch-dns 10.2.254.1/24
+
+weave_on $HOST1 run $C2/24 -t --name=c2 ubuntu
+weave_on $HOST1 run --with-dns $C1/24 -t --name=c1 aanand/docker-dnsutils /bin/sh
+
+weave_on $HOST1 dns-add $C2 c2 -h $NAME2
+
+assert_dns_record $HOST1 c1 $NAME2 $C2
+
+weave_on $HOST1 dns-add $C1 c1 -h $NAME1
+
+assert "exec_on $HOST1 c1 getent hosts $C1 | tr -s ' '" "$C1 $NAME1"
+
+weave_on $HOST1 dns-remove $C1 c1
+
+assert_raises "exec_on $HOST1 c1 getent hosts $NAME1" 2
+
+end_suite
--- a/weave
+++ b/weave
@@ -17,6 +17,11 @@ DOCKERHUB_USER=weaveworks
 BASE_EXEC_IMAGE=$DOCKERHUB_USER/weaveexec
 EXEC_IMAGE=$BASE_EXEC_IMAGE:$IMAGE_VERSION
 
+# Define some regular expressions for matching addresses.
+# The regexp here is far from precise, but good enough.
+IP_REGEXP="[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}"
+CIDR_REGEXP="$IP_REGEXP/[0-9]{1,2}"
+
 usage() {
     echo "Usage:"
     echo "weave setup"
@@ -28,6 +33,8 @@ usage() {
     echo "weave start      <cidr> [<cidr> ...] <container_id>"
     echo "weave attach     <cidr> [<cidr> ...] <container_id>"
     echo "weave detach     <cidr> [<cidr> ...] <container_id>"
+    echo "weave dns-add    <ip_address> [<ip_address> ...] <container_id> [-h <fqdn>]"
+    echo "weave dns-remove <ip_address> [<ip_address> ...] <container_id>"
     echo "weave expose     <cidr> [<cidr> ...] [-h <fqdn>]"
     echo "weave hide       <cidr> [<cidr> ...]"
     echo "weave ps         [<container_id> ...]"
@@ -72,8 +79,7 @@ check_docker_version() {
 }
 
 is_cidr() {
-    # The regexp here is far from precise, but good enough.
-    echo "$1" | grep -E '^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}/[0-9]{1,2}$' >/dev/null
+    echo "$1" | grep -E "^$CIDR_REGEXP$" >/dev/null
 }
 
 validate_cidr() {
@@ -94,6 +100,31 @@ collect_cidr_args() {
     done
     if [ $CIDR_COUNT -eq 0 ]; then
         [ $# -gt 0 ] && validate_cidr "$1" || usage
+    fi
+}
+
+is_ip() {
+    echo "$1" | grep -E "^$IP_REGEXP$" >/dev/null
+}
+
+validate_ip() {
+    if ! is_ip "$1" ; then
+        echo "Invalid IP: $1" >&2
+        echo "IP must be of the form <aaa.bbb.ccc.ddd>, e.g. 10.2.1.1" >&2
+        exit 1
+    fi
+}
+
+collect_ip_args() {
+    IP_ARGS=""
+    IP_COUNT=0
+    while is_ip "$1"; do
+        IP_ARGS="$IP_ARGS $1"
+        IP_COUNT=$((IP_COUNT + 1))
+        shift 1
+    done
+    if [ $IP_COUNT -eq 0 ]; then
+        [ $# -gt 0 ] && validate_ip "$1" || usage
     fi
 }
 
@@ -535,21 +566,32 @@ dns_args() {
     DNS_ARGS="$DNS_ARG $DNS_SEARCH_ARG"
 }
 
-# Perform operation $1 on local DNS database, for container ID $2 at
-# addresses $3, $4, ...  This function is only called when we know $2
-# is a valid container name
+# Perform operation $1 on local DNS database, for container ID $2
+# with an FQDN obtained from the container's configured hostname and
+# domain, at addresses $3, $4 ...
 tell_dns() {
     METHOD="$1"
     CONTAINER_ID="$2"
     shift 2
+    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER_ID 2>/dev/null) && true
+    tell_dns_fqdn $METHOD $CONTAINER_ID $CONTAINER_FQDN $@
+}
 
+# Perform operation $1 on local DNS database, for container ID $2
+# with FQDN $3, at addresses $4, $5 ... This function is only called
+# where we know $2 is a valid container name
+tell_dns_fqdn() {
+    METHOD="$1"
+    CONTAINER_ID="$2"
+    CONTAINER_FQDN="$3"
+    shift 3
     if ! status=$(docker inspect --format='{{.State.Running}}' $DNS_CONTAINER_NAME 2>/dev/null) || [ "$status" != "true" ] ; then
         # weavedns not running - silently return
         return
     fi
     # get the long form of the container ID
     CONTAINER=$(docker inspect --format='{{.Id}}' $CONTAINER_ID 2>/dev/null)
-    MORE_ARGS=$(docker inspect --format='--data-urlencode fqdn={{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER 2>/dev/null) && true
+    MORE_ARGS="--data-urlencode fqdn=$CONTAINER_FQDN"
     for ADDR; do
         http_call $DNS_CONTAINER_NAME $DNS_HTTP_PORT $METHOD /name/$CONTAINER/${ADDR%/*} $MORE_ARGS || true
     done
@@ -801,6 +843,26 @@ case "$COMMAND" in
         CONTAINER_ID="$1"
         with_container_netns $CONTAINER_ID detach $CIDR_ARGS >/dev/null
         tell_dns DELETE $CONTAINER_ID $CIDR_ARGS
+        ;;
+    dns-add)
+        collect_ip_args "$@"
+        shift $IP_COUNT
+        [ $# -ge 1 ] || usage
+        CONTAINER_ID="$1"
+        if [ $# -eq 1 ]; then
+            tell_dns PUT $CONTAINER_ID $IP_ARGS
+        else
+            [ $# -eq 3 -a "$2" = "-h" ] || usage
+            FQDN="$3"
+            tell_dns_fqdn PUT $CONTAINER_ID $FQDN $IP_ARGS
+        fi
+        ;;
+    dns-remove)
+        collect_ip_args "$@"
+        shift $IP_COUNT
+        [ $# -eq 1 ] || usage
+        CONTAINER_ID="$1"
+        tell_dns DELETE $CONTAINER_ID $IP_ARGS
         ;;
     expose)
         collect_cidr_args "$@"
--- a/weavedns/README.md
+++ b/weavedns/README.md
@@ -153,11 +153,19 @@ link-local as per [RFC6762](https://tools.ietf.org/html/rfc6762),
 
 If DNS is started after you've attached a container to the weave
 network, or you want to give the container a name in DNS *other* than
-its hostname, you can register it using the HTTP API:
+its hostname, you can register it using the `dns-add` command. If no
+FQDN is specified, it is derived from the container's configured
+hostname and domain. For example:
 
 ```bash
 $ docker start $shell2
-$ curl -X PUT "http://$dns_ip:6785/name/$shell2/10.2.1.27" -d fqdn=shell2.weave.local
+$ weave dns-add 10.2.1.27 $shell2 -h shell2.example.org
+```
+
+The inverse operation can be carried out using the `dns-remove` command:
+
+```bash
+$ weave dns-remove 10.2.1.27 $shell2
 ```
 
 ### Registering multiple containers with the same name


grkvlt:
@errordeveloper Can you have a look and review this, please? it is for the use case we discussed where weaveDNS is being used without the weave SDN, but I think it might be useful in other circumstances also.

rade:
Why do you want to make IP addresses for containers outside the weave network resolvable in the weave domain?

weave
```diff
@@ -532,21 +533,28 @@ dns_args() {
...
+# Perform operation $1 on container ID $2 to local DNS database with addresses at $3
+# followed by an optional fully-qualified domain name
 # This function is only called where we know $2 is a valid container name
 tell_dns() {
     METHOD="$1"
     CONTAINER_ID="$2"
     shift 2
-
+    collect_cidr_args "$@"
+    shift $CIDR_COUNT
```
rade: We have already done all the CIDR collecting at the call sites, so doing it here again, just so we can skip over them and look for a fqdn, is horrible. I suggest you introduce a `tell_dns_fqdn OPERATION CONTAINER FQDN CIDR1 CIDR2 ...` function, and invoke that from `tell_dns` and the `put-dns` code.

weave
```diff
@@ -32,6 +32,7 @@ usage() {
     echo "weave status"
     echo "weave version"
     echo "weave stop"
+    echo "weave put-dns    <cidr> [<cidr> ...] <container_id> [-fqdn fqdn]"
```
rade: This is in rather the wrong place. I suggest putting between after `detach`.

rade:
Is `put-dns` the best command name we can come up with?

Also, surely there should be an inverse operation.

rade:
NB: this partially addresses #364. Note the major caveat listed there: all these "manually" inserted records are lost on weavedns restart. Nevertheless, I suggest that this PR closes #364, and that we open separate issues for the two remaining points identified there (lossy restart, and lack of support for CNAMEs).

weave
grkvlt: @rade heh, I actually _had_ done it that way, then I refactored to what I submitted. I think you'er right, though, it would be better as two separate functions.

weave
grkvlt: OK

grkvlt:
@rade @errordeveloper For the inverse, I assume I just call `tell_dns` but with `DELETE` as the method. Do I need to worry about FQDNs in the delete call?

grkvlt:
@rade How about `add-dns` and `rm-dns` as the two commands?

rade:
> For the inverse, I assume I just call `tell_dns` but with `DELETE` as the method. Do I need to worry about FQDNs in the delete call?

You *should* have to worry about the FQDN. After all, what if you have one IP under multiple names and want to remove just one of records? However, it looks like this functionality is [missing](#631) from the DNS API.

grkvlt:
> You should have to worry about the FQDN

I pass in the domain name inside `$MORE_DATA` for the `http_call` in `tell_dns_fqdn` no matter what the `$METHOD` was set to. I think this means that once the DNS API supports it things will work as expected?

weave
```diff
@@ -28,6 +28,8 @@ usage() {
     echo "weave start      <cidr> [<cidr> ...] <container_id>"
     echo "weave attach     <cidr> [<cidr> ...] <container_id>"
     echo "weave detach     <cidr> [<cidr> ...] <container_id>"
+    echo "weave add-dns    <cidr> [<cidr> ...] <container_id> [-fqdn fqdn]"
```
rade: There is an inconsistency in our option naming here. `weave expose` takes a `-h <hostname>` parameter, which serves the same purpose. I'm leaning towards changing that to `-fqdn <fqdn>` - seems like a more accurate description of what it is. But I'd like to hear from @awh on this.

Also note that user-supplied input should appear in angle brackets in the usage. So it should be `-fqdn <fqdn>`.

weave
```diff
@@ -536,20 +538,30 @@ dns_args() {
 }
 
 # Perform operation $1 on local DNS database, for container ID $2 at
-# addresses $3, $4, ...  This function is only called when we know $2
-# is a valid container name
+# addresses $3, $4 ... Uses FQDN set on container and calls tell_dns_fqdn()
```
rade: "...for container ID $2, with a FQDN obtained from the container's configured hostname&domain, at addresses $3, $4 ...". No need to mention the tell_dns_fqdn bit.

weave
```diff
@@ -536,20 +538,30 @@ dns_args() {
...
 tell_dns() {
     METHOD="$1"
     CONTAINER_ID="$2"
     shift 2
+    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER_ID 2>/dev/null) && true
+    tell_dns_fqdn $METHOD $CONTAINER_ID $CONTAINER_FQDN $@
+}
 
+# Perform operation $1 on local DNS database, for container ID $2
+# with fully-qualified domain name $3 at addresses $4, $5 ... This
```
rade: s/fully-qualified domain name/fqdn

weavedns/README.md
```diff
@@ -153,11 +153,18 @@ link-local as per [RFC6762](https://tools.ietf.org/html/rfc6762),
...
 network, or you want to give the container a name in DNS *other* than
-its hostname, you can register it using the HTTP API:
+its hostname, you can register it using the `add-dns` command:
 
 ```bash
 $ docker start $shell2
-$ curl -X PUT "http://$dns_ip:6785/name/$shell2/10.2.1.27" -d fqdn=shell2.weave.local
+$ weave add-dns 10.2.1.27/24 $shell2 -fqdn shell2.weave.local
+```
+
```
rade: plus "If no FQDN is specified, it is derived from the container's configured hostname and domain."

rade:
still not sure about the command names. `add/remove-name`? It's not perfect, but `add-dns-A-record(s)-for-container-under-name-and-given-ip(s)` is a bit long.

grkvlt:
@rade happy to change to `add-name` and `remove-name` or what about `rm-name` instead?

rade:
Let's see what other suggestions come up from interested bike shed painters.

grkvlt:
OK, I have left it as `add-name` and `rm-name` for the moment. I also didn't include the ability to set an FQDN for the `rm-name` command, since as you pointed out it's not supported in the DNS API yet.

Thanks for taking the time to review this @rade

grkvlt:
I added a test script (called `240_dns_add_name_test.sh`) but was not able to run it properly, although manual testing of the `add-name` command works for me on a local Vargrant build of Weave. Are these tests run automatically anywhere?

awh:
Are `add-name` etc a bit generic for the top level? How about contextualising them like this:
````
weave dns add
weave dns remove
````

**Comment by awh** [comment]
Just discussing with @squaremo - further to this line of thought we could contextualise the existing router commands:
````
weave router launch
weave router attach
weave dns launch
````
obviously we'd still support the existing script commands, but they would become deprecated - this would allow us to address #602 cleanly...

**Comment by inercia** [comment]
I like @awh proposal. I think we should move to a format like _weave \<command\> \<subcommand\>_

**Comment by grkvlt** [comment]
@awh @inercia OK, but we should do that in a different PR. For this feature I think the current command names will be fine, although they could use something likr the `@Beta` annotation Java has?

**Comment by rade** [comment]
let's go with `dns-add` and `dns-remove`. This introduces an inconsistency w.r.t. other composite commands, which are "verb-noun", but all the alternatives we can think of are either  (e.g. `add-dns-record` longer or misleading (e.g. `add-dns`).

**Comment by grkvlt** [comment]
Done ... :frog:

**Comment by rade** [comment]
@grkvlt
> I added a test script but was not able to run it properly

What happened? Our test set up is *supposed* to be easy to install.

File: weave (hunk 2 above)
Comment (awh) [inline weave]: The `-h` option to `weave expose` mirrors the `-h` that can be passed to `weave run` to set the weaveDNS name at container launch, unfortunately that is interpreted by Docker itself and cannot be changed to follow suit...

**Comment by rade** [comment]
> Should this have the `-h shell2.weave.local` appended?

Nah, let's leave that to #631.

Re option naming, @awh and I have come up with `-h <fqdn>`. I've [just updated](https://github.com/weaveworks/weave/commit/e23c15dab27a110a5b622c48f2cdaec6bf4a5a7f) the `expose` usage/docs to reflect that.

**Comment by grkvlt** [comment]
@rade ok, updated and pushed, let me try the tests again...

**Comment by grkvlt** [comment]
Great, the new DNS test 240 now passes, there was a typo the first time:
```
$ ./240_dns_add_name_test.sh
Cleaning up on 192.168.48.11: removing all containers and resetting weave
Cleaning up on 192.168.48.12: removing all containers and resetting weave
Add names on a single host
Weave on 192.168.48.11: launch-dns 10.2.254.1/24
109a79da0b49b603bcccda2adc11b0ad2725d836e7e43595295c9cdadf1e3ac1
Weave on 192.168.48.11: run 10.2.0.34/24 -t --name=c2 ubuntu
d74159c9030e92aca8e29af536803d2fab55d7d9de552137e02ec07e407a48b1
Weave on 192.168.48.11: run --with-dns 10.2.0.78/24 -t --name=c1 aanand/docker-dnsutils /bin/sh
2bf38a74d73fada303d5aed5624079c3200bd7da68765bd50e42d7b67a613fd6
Weave on 192.168.48.11: dns-add 10.2.0.34/24 c2 -h seetwo.weave.local
all 2 tests passed in 8.328s.
```

**Comment by rade** [comment]
you should add a test for `dns-remove` too. And, if you don't mind, rebase and squash everything into a single commit.

File: weavedns/README.md [hunk 7]
```diff
@@ -153,11 +153,20 @@ link-local as per [RFC6762](https://tools.ietf.org/html/rfc6762),
...
 $ docker start $shell2
-$ curl -X PUT "http://$dns_ip:6785/name/$shell2/10.2.1.27" -d fqdn=shell2.weave.local
+$ weave dns-add 10.2.1.27/24 $shell2 -h shell2.weave.local
+```
+
+The inverse operation can be carried out using the `dns-remove` command:
+
+```bash
+$ docker stop $shell2
+$ weave dns-remove 10.2.1.27/24 $shell2
```
rade: Hmm. The rationale for doing that isn't clear. weavedns will see the container stop and remove the dns entries anyway. Well, that's what it is supposed to do.

So I think you need a better example here.

grkvlt:
```
$ ./240_dns_add_name_test.sh 
Cleaning up on 192.168.48.11: removing all containers and resetting weave
Cleaning up on 192.168.48.12: removing all containers and resetting weave
Add and remove names on a single host
Weave on 192.168.48.11: launch-dns 10.2.254.1/24
9a8f3007c9166ce78815f2c2e56336fa0ebdcb9215f0c27c731f2b65fb8a1849
Weave on 192.168.48.11: run 10.2.0.34/24 -t --name=c2 ubuntu
9a8a29ee877e11acdaa4ef309c3d9344553ce941e9fee3189049b5cc3fc68296
Weave on 192.168.48.11: run --with-dns 10.2.0.78/24 -t --name=c1 aanand/docker-dnsutils /bin/sh
cf10f945925795426b5a14463fedcc4acb5a8fba9662fff39240ca42dbd0bdff
Weave on 192.168.48.11: dns-add 10.2.0.34/24 c2 -h seetwo.weave.local
Weave on 192.168.48.11: dns-remove 10.2.0.34/24 c2
all 3 tests passed in 8.012s.
```

weavedns/README.md
grkvlt: I'll just remove the `docker stop` command...

grkvlt:
@rade should be ready to merge, rebased and sqashed, with tests passing

test/240_dns_add_name_test.sh
```diff
@@ -0,0 +1,24 @@
...
+weave_on $HOST1 run $C2/24 -t --name=c2 ubuntu
+weave_on $HOST1 run --with-dns $C1/24 -t --name=c1 aanand/docker-dnsutils /bin/sh
+
+weave_on $HOST1 dns-add $C2/24 c2 -h $NAME
+
+assert_dns_record $HOST1 c1 $NAME $C2
+
+weave_on $HOST1 dns-remove $C2/24 c2
+
+assert_raises "exec_on $HOST1 c1 getent hosts $C2 | grep $NAME"
```
rade: `assert_raises` checks whether the exist status matches $2, defaulting to 0. So the above actually checks whether the reverse lookup of the IP succeeds. Which, surprisingly, it does! Why? Because it got cached, since `assert_dns_record` performs both forward and reverse lookups, and weavedns caches answers.

So we need a better test. I suggest adding a 2nd record, removing it again, and then check that a lookup fails. NB: the check should be done with a *forward* lookup; reverse lookups are expensive when they fail since they go to upstream dns server.

rade:
`add/remove-dns` should take IP addresses (`<ip_address>` in weave usage terminology), not CIDRs.

I suspect this is going to make things a fair bit more involved since all the existing functions you are calling expect CIDRs. :(

test/240_dns_add_name_test.sh
grkvlt: OK, thank, that makes sense. I didn't know much about your test system, sorry... But, why do forward lookups of .local domains get resolved recursively? I would have thought you could control that, although I would need to do a lookup on `host.weave.local.` (note trailing `.`) I guess. Or you could use a `.weave` TLD for local-only addresses?

rade: Forward lookups don't get resolved recursively. Reverse lookups, which is what you had (`$C2` is an IP address), do require asking the upstream if the IP is unknown to us.

test/240_dns_add_name_test.sh
```diff
@@ -0,0 +1,26 @@
...
+weave_on $HOST1 run --with-dns $C1/24 -t --name=c1 aanand/docker-dnsutils /bin/sh
+
+weave_on $HOST1 dns-add $C2/24 c2 -h $NAME1
+
+assert_dns_record $HOST1 c1 $NAME1 $C2
+
+weave_on $HOST1 dns-add $C2/24 c2 -h $NAME2
+weave_on $HOST1 dns-remove $C2/24 c2
+
+assert_raises "exec_on $HOST1 c1 getent hosts $NAME2 | grep $C2" 1
```
rade: You don't need the `grep` here; the `getent` should fail with a status of 2.

grkvlt:
@rade try again :koala: this has IP addresses instead of CIDRs and imoroved tests as siggested

weave
```diff
@@ -97,6 +99,32 @@ collect_cidr_args() {
     fi
 }
 
+is_ip() {
+    # The regexp here is far from precise, but good enough.
+    echo "$1" | grep -E '^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$' >/dev/null
```
rade: perhaps define an `ip_regexp='[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}'` and use that both here and in `is_cidr`.

weave
```diff
@@ -562,13 +601,8 @@ populate_dns() {
...
     fi
     for CONTAINER in $(docker ps -q --no-trunc); do
-        MORE_ARGS=$(docker inspect --format='--data-urlencode fqdn={{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER 2>/dev/null) && true
         if CONTAINER_IPS=$(with_container_netns $CONTAINER container_weave_addrs 2>&1 | sed -n -e 's/inet \([^/]*\)\/\(.*\)/\1/p') ; then
-            for IP in $CONTAINER_IPS; do
-                # NB: CONTAINER_IP is the IP of the weavedns
-                # container; it is set by wait_for_status.
-                http_call_ip $CONTAINER_IP $DNS_HTTP_PORT PUT /name/$CONTAINER/$IP $MORE_ARGS
-            done
+            tell_dns PUT $CONTAINER $CONTAINER_IPS
```
rade: The previous code here was an optimisation. `tell_dns` checks whether weaveDNS is running and looks up its IP address. That is quite expensive, and unnecessary to do for every container; can be done just once instead.

weave
```diff
@@ -79,7 +81,7 @@ is_cidr() {
 validate_cidr() {
     if ! is_cidr "$1" ; then
         echo "Invalid CIDR: $1" >&2
-        echo "CIDR must of be of form <ip_address>/<routing_prefix_length>, e.g. 10.2.1.1/24" >&2
+        echo "CIDR must be of the form <ip_address>/<routing_prefix_length>, e.g. 10.2.1.1/24" >&2
```
rade: well spotted. I've just corrected this on master.

weave
grkvlt: I thought about that as well... If you think its a good idea, I'll make the change

weave
grkvlt: Not convinced. I think the convenience of calling `tell_dns` wins out, as that is what is done everywhere else. So one extra call to `docker` to check if WeaveDNS is running, plus one more to get the ID, makes three per container instead of just the call to get the domain name which is present in both. So, we triple the number of `docker` calls

How many containers do you think need to be running per host before this will be an issue? I think that with small numbers (how many containers per host are people using, on average?) of containers it should not be noticeable, but it would be useful to know some exact numbers, probably.

rade: Run some benchmarks. You'll be surprised how expensive it is to interact with docker.

Anyway, this has nothing to do with this PR. If you want to change this code, create a separate PR.

weave
rade: yes, please. Or I can do it post merge.

weave
grkvlt: OK, will revert.

grkvlt:
@rade we'll get there eventually... thanks again for helping me get this straightened out

rade:
@grkvlt thanks for your patience. Merged.

PR #628, opened 2015-05-04
prevent interference of expose in other subnets

### LINKED ISSUE #605: `weave expose` breaks container connectivity in other subnets
Whilst investigating #592 it has become apparent that you cannot use multiple subnets in conjunction with `weave expose`. To reproduce:

1. Attach address 10.2.1.4/24 to container c1 on host1
2. Attach address 10.2.1.7/24 to container c2 on host2
3. Weave expose 10.2.2.2/24 on host1

If you now ping c2 from c1, you will observe on c2 that the echo request source address has been rewritten to that of the exposed address on host1's weave bridge (10.2.2.2 in this example)  by the masquerading rules installed by `weave expose` on host1:

````
Chain WEAVE (1 references)
 pkts bytes target     prot opt in     out     source               destination
    4   336 MASQUERADE  all  --  *      weave  !10.2.2.0/24          0.0.0.0/0
    0     0 MASQUERADE  all  --  *      !weave  10.2.2.0/24          0.0.0.0/0
````

Furthermore, it's not clear that the masquerading rules work as intended when multiple addresses are exposed - masquerading is defined thusly:

````
'You don't need to put in the source address explicitly with masquerading: it will use the source address of the interface the packet is going out from'
````

What happens when the interface has multiple primary IP addresses?

rade:
Fixes #605.

### DIFF
--- a/test/130_expose_test.sh
+++ b/test/130_expose_test.sh
@@ -1,0 +1,61 @@
+#! /bin/bash
+
+. ./config.sh
+
+C1=c1:10.2.1.34
+C2=c2:10.2.1.37
+C3=c3:10.2.2.34
+C4=c4:10.2.2.37
+EXP=10.2.1.101
+
+PING="ping -nq -W 1 -c 1"
+
+weave_on1() {
+    assert_raises "weave_on $HOST1 $@"
+}
+
+run_on1() {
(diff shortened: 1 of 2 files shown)


awh:
Looks good to me!

PR #637, opened 2015-05-07
Improvements to circle ci integration

tomwilkie:
- Don't need gsutil on circle
- Use a prebuild weave-build image and cache it locally.
- Name VMs after change I'm building, so I can do multiple concurrently
- Don't build gh-pages branch


paulbellamy:
LGTM

PR #642, opened 2015-05-08
Fixes post #612

errordeveloper:
Use rsync instead of cp to avoid uploading site to a subdirectory (when the directory already exists).

### DIFF
--- a/site/build-and-publish.sh
+++ b/site/build-and-publish.sh
@@ -26,7 +26,7 @@ gsutil -m cp -z html,css -a public-read -R _site "gs://docs.weave.works/${OUTPUT
 
 echo "Published at http://docs.weave.works/${OUTPUT}"
 
-if [ -z "${TRAVIS_TAG}" ]; then
+if [ -z "${TRAVIS_TAG}" -a ! "${BRANCH}" = "latest_release_doc_updates" ]; then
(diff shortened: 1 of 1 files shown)


errordeveloper:
The switch to `gsutil rsync` is mostly due to GoogleCloudPlatform/gsutil#274, but also it's somewhat more optimal anyway.

site/build-and-publish.sh
```diff
@@ -22,11 +22,11 @@ export BRANCH COMMIT OUTPUT
 bundle install --path=.bundle
 bundle exec jekyll build --verbose
 
-gsutil -m cp -z html,css -a public-read -R _site "gs://docs.weave.works/${OUTPUT}"
+gsutil -m rsync -d _site "gs://docs.weave.works/${OUTPUT}"
```
errordeveloper: The `-z` and `-a` flags are not supported, but we shouldn't have too big files to compress them anyway and now a default permission set on the bucket for all object to be publicly readable, so `-a` is not needed also.

paulbellamy:
Will this not still make stuff look like: ```${OUTPUT}/_site``` ?

errordeveloper:
> Will this not still make stuff look like: `${OUTPUT}/_site` ?

From what I am seeing, no.

paulbellamy:
LGTM

PR #650, opened 2015-05-13
gsutil rsync does not imply recursion

errordeveloper:


rade:
Did this really require a PR?

errordeveloper:
@rade true, it doesn't really.

PR #652, opened 2015-05-13
Add weave launch-proxy

paulbellamy:
This adds a docker remote API proxy to the weaveexec image. The proxy is
auto-launched when running a container with weave cli.

The proxy currently has hooks for:

 Hook | Action
 ---- | ------
Pre Container Create   | Adjusts the entrypoint of a container to wait for the ethwe interface. 
Post Container Start         | Attaches a weave interface to a container given a WEAVE_CIDR environment entry. 

Other changes include:
* weave start no longer takes CIDRs. CIDRs from run will be re-attached, but not ones attached manually. weave attach should be used to reattach those.


rade:
docs need updating

squaremo:
> docs need updating

Off the top of my head:

 * Usage of weave start, if it appears, should be updated (in readme, possibly; in site/features.md)
 * Any examples that assume containers will run in the background will need changing

I think it'd be worth having a brief description of how weave run goes through the proxy, and how one can use docker directly to do the same thing.

paulbellamy:
As per f2f discussion with @rade, let's not change bash script commands to use proxy, or autolaunch. Just provide launch-proxy as a "beta feature" and document how to use it with the docker cli tool.

paulbellamy:
Closed in favour of #655

Issue #47, opened 2014-09-11
set up app container networking before containers start


rade:
Currently 'weave run' sets up the app container's interface into the weave network *after* the container has been launched with 'docker run -d'. That means the network may not be available to the container process straight away. Depending on what the container is doing, that can be benign, annoying, or disastrous.

Containers can themselves ensure that the interface is available, by running s.t. like https://github.com/jpetazzo/pipework/blob/master/pipework#L30, i.e.

    while ! grep -q ^1$ /sys/class/net/ethwe/carrier 2>/dev/null
    do sleep 1
    done

before starting the actual container process, but this of course requires containers to have been constructed with weave in mind, which is limiting.

There is no way around this issue w/o some changes to docker.

rade:
One challenge with the "rewrite entrypoint to a volume-mounted exe" approach is making it work when we don't have direct access to the docker host and instead run weave itself in a container (as per #312)? Where would the exe live s.t. we can volume-mount it into containers?

The answer is `--volumes-from`... As part of `weave launch` we start a named container that copies the required exe into a volume created with `-v`. For example,  we could add the exe, and the program to copy it, to the weavetools image, in which case we would start that container with something like `docker run --name weavetools -v /home/weave zettio/weavetools /bin/copy-entrypoint-exe /home/weave`. We can then start all other containers with `--volumes-from=weavetools`.

Note that weavetools doesn't need to remain running for this to work, though the container needs to continue to exist. We'd need to take care of removing it with `docker rm -v` on `weave stop`.

rade:
Trying to extract the orginal entrypoint from the command line (in case it was specified there), and determining the image from the command line (so we can lookup the entry point specified in that), is fiendishly hard. Go's option parser library `flag`, which is what docker uses, allows options to be specified in a myriad of ways. Moreover, there is no generic parsing function; parsing requires a specification of all the options. This not not just a matter of missing APIs; the option grammar is ambiguous in the absence of an option specification - consider the difference between `docker run -i a b` (`a` is the image and `b` is the command) and `docker run -h a b` (`a` is the host and `b` is the image).

The best option I can think of here is to invoke docker's `runconfig.Parse` function. This performs nearly all of the `docker run` option parsing and appears to require little/no context. However, it is missing some options, because the function is also used to for parsing `docker create`, so any options not shared with `docker create` would have to be added by us before invoking the function.

Furthermore, the whole thing is highly version specific; there's probably not been a single docker release when the `docker run` options haven't changed.

[...]

Issue #629, opened 2015-05-04
Example to use docker-compose with weave


stephanbuys:
Hi, created this demonstration of weave and docker-compose playing together. At the moment I use two instantiations of the compose tool because weave needs to run after the containers are started. 

Something like https://github.com/docker/compose/pull/686 might solve it later.

https://gist.github.com/stephanbuys/f16f793ef9373b9b28a6

rade:
@squaremo

errordeveloper:
@stephanbuys that's very interesting, let me look into it and I'll get back to you if I have any questions.

stephanbuys:
@errordeveloper, your latest blog post ( https://github.com/errordeveloper/weave-demos/tree/master/quartet/app ), does the WEAVE_CIDR environment variable support multiple networks? I've got several containers that attach to frontends and backends with different networks.

FYI: https://github.com/docker/compose/pull/686

One of the things that weave could potentially support is a way for the "attach" container to wait/listen for the container in the argument for weave attach to appear, and then only attempt to run/attach (perhaps with a timeout), that will make the order in which containers are started (and weave commands are issued) independent.

errordeveloper:
The blog post that will use compose with the `WEAVE_CIDR` is not ready yet, it's waiting for 0.11 release of Weave. I will have to check if this method will support multiple subnets, that's a very good question.

Regarding the container process startup after initialisation of Weave interface, it essentially depends on network plugin support in Docker itself, however the proxy method that my compose blog post will use should provide a basic solution to this.

errordeveloper:
You can look at the proxy implementation in #655.

stephanbuys:
@errordeveloper is there a way to test the proxy pre-release? I'm guessing it will be "swarm" compatible as long as you schedule a proxy instance on each of the nodes in the Swarm cluster. It's a great feature.

errordeveloper:
> ... is there a way to test the proxy pre-release?

The 0.11 release is due this week, so you shouldn't have to wait for too long.

>  I'm guessing it will be "swarm" compatible as long as you schedule a proxy instance on each of the nodes in the Swarm cluster.

Yes, it will be. I'm in the process of writing a blog post on how it will work with Swarm and Compose.

> ...does the `WEAVE_CIDR` environment variable support multiple networks?

Yes, please see [the docs](https://github.com/paulbellamy/weave/blob/b7acf9e1535213b9b61f9fe1a5460ce83203510a/site/proxy.md#usage).

[...]

Issue #401, opened 2015-02-17
make docker daemon initiated container (re)start work properly


rade:
The docker daemon can decide to (re)start containers, e.g. after a reboot or when a `--restart` policy was specified for the container.

We want this to work properly for a) the weave container and any similar weave infrastructure containers (such as weavedns), b) application containers. In the latter case this would entail re-attaching to the weave network, and re-registering with weavedns.

This does require bits of #230/#312/#117.

bboreham:
It would be nice if the container came back with the same IP address after a restart, as described in #1047.

rade:
Presumably any containers started via the proxy will hang in weavewait when restarted by docker.

The proxy, or a separate container, could take care of re-attaching containers. This won't work on docker restart though, since the proxy may not be running yet. Though I guess we could scan all containers on proxy startup and attach any that are weavewait-ing.

bboreham:
> This won't work on docker restart though, since the proxy may not be running yet

The Docker Event mechanism allows clients to fetch events from before the time they were started, so this is not an issue.

Doing that brings a different issue: what if the proxy itself is restarted, and queries for past events, and thus gets told of the start of containers that it already attached to weave (before the proxy was restarted)?

Maybe we can rely on attachment being idempotent, or maybe we can check before attaching.

bboreham:
The fix in #1210 does not do (a) in the original description - dealing with restart of weave itself.

PR #655, opened 2015-05-13
Minimal implementation of the proxy

paulbellamy:
Supersedes #652 

This adds a docker remote API proxy to the weaveexec image. The proxy is NOT
auto-launched when running a container with weave cli. Nor is it used at all by the weave CLI script.
It is merely provided/documented as a beta feature, so that we can gauge feedback.

The proxy currently has hooks for:

 Hook | Action
 ---- | ------
Pre Container Create   | Adjusts the entrypoint of a container to wait for the ethwe interface. 
Post Container Start         | Attaches a weave interface to a container given a WEAVE_CIDR environment entry.


paulbellamy:
"Beta features" or "Experimental features" ?

rade:
"Experimental Features". And I'd put that right after 'Features'.

squaremo:
You checked in the wait-for-weave binary -- is that safe with respect to licensing?

paulbellamy:
Documetation notes from discussion in person:
- [ ] check first/third-person for consistency
- [ ] export DOCKER_HOST instead of -H, and explain how you can forward all commands
- [ ] explain how it connects to docker
- [ ] explain more about advantages (mention foreground run)
- [ ] mention args -H and -L to launch-proxy
- [ ] show how to do DNS this way
- [ ] mention no TLS support, yet

rade:
> You checked in the wait-for-weave binary

We shouldn't have that checked it. Only the source.

paulbellamy:
@squaremo good call, will check.

rade:
or, rather, we should fetch & build it.

rade:
This kinda fixes #47 and #230. Though we should probably not declare victory on these until the proxy is more fully integrated. @paulbellamy please raise an issue for the latter. TLS is covered by #506.

test/600_proxy_docker_py_test.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
+
+. ./config.sh
+
+start_suite "Run docker-py test suite against the proxy"
+
+assert_raises "docker_on $HOST1 ps | grep weaveproxy" 1
```
awh: `start_suite` removes all containers from the test hosts, so this check is redundant

test/600_proxy_docker_py_test.sh
```diff
@@ -0,0 +1,17 @@
...
+weave_on $HOST1 launch-proxy
+assert_raises "docker_on $HOST1 ps | grep weaveproxy"
+docker_on $HOST1 run \
```
awh: Does this run command exit with a non-zero status if `integration_test.py` has test failures? It's not wrapped in an `assert_raises`, so it will cause `run_all.sh` to bail immediately with an `ABNORMAL TEST TERMINATION` message without running the remainder of the test suites...

rade:
I think this also addresses #117, with the same caveats.

rade:
Does this also solve #400?

rade:
and #401?

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -0,0 +1,33 @@
...
+docker_proxy_on() {
+    host=$1
+    shift 1
+    greyly echo "Docker (with proxy) on $host: $@" >&2
+    docker -H tcp://$host:12375 $@
+}
+
+start_suite "Proxy waits for weave to be ready before running container commands"
+weave_on $HOST1 launch-proxy
+assert_raises "docker_proxy_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' --name=c1 gliderlabs/alpine ifconfig ethwe" 0
```
awh: I was going to say that we favour `ip` over the deprecated `ifconfig`, but I've just checked and we are using `ifconfig` in one other test and also mention it in `dockerless.txt`. Should we standardise on `ip` @rade?

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -0,0 +1,33 @@
...
+# Setup and sanity-check
+weave_on $HOST1 launch-proxy
+docker_on $HOST1 build -t ifconfig-ethwe - <<- EOF
+  FROM gliderlabs/alpine
+  ENTRYPOINT ["ifconfig", "ethwe"]
+EOF
+assert "docker_on $HOST1 inspect --format='{{.Config.Entrypoint}}' ifconfig-ethwe" "[ifconfig ethwe]"
+assert "docker_on $HOST1 inspect --format='{{.Config.Cmd}}' ifconfig-ethwe" "<no value>"
+
+# Boot a new container with no entrypoint of it's own, just a command
```
awh: _its_

test/610_proxy_wait_for_weave_test.sh
rade: >  Should we standardise on `ip`

yes.

paulbellamy:
re: licensing, it's Apache2, so since we are "distributing" it, we probably need attribution, and a license included? but I am not a lawyer...

test/config.sh
```diff
@@ -63,21 +63,21 @@ greenly() {
...
-    greyly echo "Running on $host: $@"
+    greyly echo "Running on $host: $@" >&2
     remote $host $SSH $host $@
 }
 
 docker_on() {
     host=$1
     shift 1
-    greyly echo "Docker on $host: $@"
+    greyly echo "Docker on $host: $@" >&2
```
awh: Nice - presumably this means I can do
````
CONTAINER_ID=$(docker_on ...)
````
in my test suites?

Comment (paulbellamy) [inline test/config.sh]: yes

**Comment by paulbellamy** [comment]
re: licensing, will remove binocarlos/wait-for-weave, and reinvent the wheel.

File: weaveexec/main.go [hunk 6]
```diff
@@ -0,0 +1,36 @@
...
+	flag.StringVar(&target, "H", "unix:///var/run/docker.sock", "docker daemon URL to proxy")
+	flag.StringVar(&listen, "L", ":12375", "address on which to listen")
+	flag.BoolVar(&debug, "debug", false, "log debugging information")
+	flag.Parse()
+
+	if debug {
+		InitDefaultLogging(true)
+	}
+
+	p, err := proxy.NewProxy(target)
```
awh: You're not checking the value of `err`, and the compiler isn't noticing the unused value because of the assignment on line 32. I wonder if it's a best practice to always use `:=` for error values due to this?

paulbellamy: Good catch, thanks.

rade:
A few more things to figure out and document...

- Does setting `DOCKER_HOST` to the proxy interfere with `weave ...` operations? Ideally it wouldn't.

- atm the docs only mention `docker run`. Presumably the proxy also does its thing on `create`, i.e. a container `create`d with a `WEAVE_CIDR` will a) have its entry point rewritten, and b) be attached to the weave network when `start`ed. Correct? That needs documenting.

- We need to make this work with IPAM from #563. Strawman: all containers `run` or `create`d+`start`ed via the proxy will receive a weave IP from IPAM if no `WEAVE_CIDR` was specified. Gotcha: it's going to be tricky to make this work in conjunction with the first point, i.e. `DOCKER_HOST` set to the proxy not interfering with `weave ...` operations, since the `weavexec` containers would presumably end up getting attached to the weave network. One way around that might be to do IPAM when an *empty* `WEAVE_CIDR` was specified, and leave containers w/o any `WEAVE_CIDR` env vars well alone.

rade:
Perhaps there could be a flag on the proxy that controls whether it should act on all containers or only containers with a `WEAVE_CIDR`.

rade:
Regarding DNS... the proxy could have a `--with-dns` flag that controls whether to configure container's DNS to point at weaveDNS, much like `weave run --with-dns` does.

stephanbuys:
+1 This is a great feature, of note as ,mentioned in https://github.com/weaveworks/weave/issues/629, WEAVE_CIDR should support multiple CIDR's (for containers connected to other northbound and southbound containers).

rade:
It already does. Per the docs in this PR.

stephanbuys:
@rade, apologies, fired off too fast. Excitement got to me :)

stephanbuys:
Have LABELs been considered instead (or in conjunction with) ENV for the WEAVE_CIDR (http://docs.docker.com/userguide/labels-custom-metadata/#add-labels-to-images-the-label-instruction)? Not a fully-baked thought yet, but reading a LABEL in certain use-cases would curb information-disclosure where knowledge that the container is networked using Weave might matter? Although that could be deduced from the presence of the ethwe bridge too?

rade:
> Have LABELs been considered 

Labels are too recent a feature to depend on.

> reading a LABEL in certain use-cases would curb information-disclosure. ... Although that could be deduced from the presence of the ethwe bridge too?

Exactly.

weavewait/main.go
```diff
@@ -0,0 +1,42 @@
...
+			fmt.Fprintln(os.Stderr, err)
+		}
+	}
+
+	if err := runSubcommand(os.Args); err != nil {
+		os.Exit(1)
+	}
+}
+
+func shouldSkip() bool {
```
rade: why do we need this?

paulbellamy: It was useful to me when debugging the tests, so no harm in leaving it in.

rade: nah. take it out. And inline `runSubCommand`.

paulbellamy:
+1 for `weave launch-proxy --with-dns`. Let's add that.

re: IPAM, let's pull that into a separate issue to be completed after IPAM.

site/experimental-features.md
```diff
@@ -0,0 +1,46 @@
...
+
+    host1$ weave launch-proxy
+
+By default, the proxy will connect to docker at `unix:///var/run/docker.sock`, and listen
+on port 12375. However, we can adjust these via the `-H` and `-L` arguments,
+respectively. All docker commands can be run via the proxy, so it is safe to
+globally adjust your `DOCKER_HOST`.
+
+    host1$ export DOCKER_HOST="tcp://host1:12375"
+    host1$ export DOCKER_TLS_VERIFY=0
```
rade: I'd leave that out. If the user ordinarily connects to docker via tls, turning that off is not a good idea; it's better for that to fail.

site/experimental-features.md
```diff
@@ -0,0 +1,46 @@
...
+    host1$ docker ps
+
+#### Usage
+
+When containers are created via the weave proxy, their entrypoint will be
+modified to wait for the `ethwe` interface to become available. When they are
```
rade: We don't mention `ethwe` anywhere else in the docs. Just stick to "weave network interface".

site/experimental-features.md
```diff
@@ -0,0 +1,46 @@
...
+started via the weave proxy, any network addresses specified in the
+`WEAVE_CIDR` environment variable will be attached. We can create and start a
+container via the weave proxy with
+
+    host1$ docker run -e "WEAVE_CIDR=10.2.1.1/24" -dti ubuntu /bin/sh
```
rade: I'd drop the `-d` here. After all, foreground mode is one of the highlights here.

rade: Are the quotes necessary? Wouldn't have thought so.

site/experimental-features.md
```diff
@@ -0,0 +1,46 @@
...
+
+    host1$ export DOCKER_HOST="tcp://host1:12375"
+    host1$ export DOCKER_TLS_VERIFY=0
+    host1$ docker ps
+
+#### Usage
+
+When containers are created via the weave proxy, their entrypoint will be
+modified to wait for the `ethwe` interface to become available. When they are
+started via the weave proxy, any network addresses specified in the
```
rade: "network addresses" -> "IP&network addresses"?

site/experimental-features.md
```diff
@@ -0,0 +1,46 @@
...
+`WEAVE_CIDR` environment variable will be attached. We can create and start a
+container via the weave proxy with
+
+    host1$ docker run -e "WEAVE_CIDR=10.2.1.1/24" -dti ubuntu /bin/sh
+
+`docker create`, and `docker start`, will also work independently.
```
rade: I'd do this as an example, "or, equivalently with".

site/experimental-features.md
paulbellamy: ?

rade: well, I don't think we use the term 'network address' anywhere else, and it's misleading. "IP address & network", perhaps?

errordeveloper:
I am getting `Unable to create proxied request`, when I try to run an image that I haven't pulled yet... Just guessing, this behaviour might differ between daemon API versions.

```
Client version: 1.6.0
Client API version: 1.18
Go version (client): go1.4.2
Git commit (client): 4749651
OS/Arch (client): darwin/amd64
Server version: 1.6.0
Server API version: 1.18
Go version (server): go1.4.2
Git commit (server): 4749651
OS/Arch (server): linux/amd64
```

paulbellamy:
@errordeveloper Thanks, try it now.

proxy/common.go
```diff
@@ -0,0 +1,36 @@
...
+func callWeave(args ...string) ([]byte, error) {
+	args = append([]string{"--local"}, args...)
+	Debug.Print("Calling weave", args)
+	cmd := exec.Command("./weave", args...)
+	cmd.Env = []string{"PROCFS=/hostproc", "PATH=/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}
+	out, err := cmd.CombinedOutput()
+	return out, err
+}
+
+func weaveCidrsFromConfig(config *docker.Config) ([]string, bool) {
```
rade: go convention is to all-caps acronyms, so this should be weaveCIDRsFromConfig

proxy/proxy.go
```diff
@@ -0,0 +1,112 @@
...
+	if isWeaveStatus(r) {
+		proxy.WeaveStatus(w, r)
+	} else if isCreateContainer(r) {
+		proxy.CreateContainer(w, r)
+	} else if isStartContainer(r) {
+		proxy.StartContainer(w, r)
+	} else {
+		proxy.ProxyRequest(w, r)
+	}
+}
```
rade: use a `switch`.

site/experimental-features.md
```diff
@@ -0,0 +1,64 @@
...
+
+ * [Proxy](#proxy)
+
+### <a name="proxy"></a>Proxy
+
+Instead of the weave command-line utility, you may prefer to use the standard
+docker command-line utility to manage containers. This will allow containers to
+be started in foreground mode. It will also ensure that the weave network
+interface is available before running your container. However, it does not
+currently support TLS.
```
rade: I'd put the TLS bit under a `Limitations` section at the end.

site/experimental-features.md
```diff
@@ -0,0 +1,64 @@
...
+
+#### Usage with WeaveDNS
+
+If we wish to use WeaveDNS with the proxy, we need to launch the proxy with dns support, via
+
+    host1$ weave launch
+    host1$ weave launch-dns 10.2.254.1/24
+    host1$ weave launch-proxy --with-dns
+
+With this done, any containers launched through the proxy which have `WEAVE_CIDR` and a hostname in the `.weave.local` domain will be registered in weaveDNS. We can launch a container to be added into weaveDNS via
```
rade: That is misrepresenting what `--with-dns` does. It tells the container to use WeaveDNS for name resolution. That is quite orthogonal from possibly creating a DNS record in WeaveDNS. The latter should happen for all containers with a `WEAVE_CIDR`, regardless of whether `--with-dns` was specified.

site/experimental-features.md
```diff
@@ -0,0 +1,64 @@
...
+
+    host1$ docker create -e WEAVE_CIDR=10.2.1.1/24 -ti ubuntu /bin/sh
+    5ef831df61d50a1a49272357155a976595e7268e590f0a2c75693337b14e1382
+    host1$ docker start 5ef831df61d50a1a49272357155a976595e7268e590f0a2c75693337b14e1382
+
+Multiple IP addresses and networks can be supplied in the WEAVE_CIDR variable by space-separating them, as in `WEAVE_CIDR=10.2.1.1/24 10.2.2.1/24`.
+
+#### Usage with WeaveDNS
+
+If we wish to use WeaveDNS with the proxy, we need to launch the proxy with dns support, via
```
rade: "Containers started via the proxy can be automatically configured to use WeaveDNS for name resolution. To accomplish this we need to launch the proxy with the `--with-dns` option..."

weave
```diff
@@ -128,6 +129,55 @@ collect_ip_args() {
...
+            ;;
+    esac
+}
+
+wait_for_proxy_to_boot() {
+    for n in `seq 1 20`; do
+      if curl --max-time 1 --connect-timeout 1 -s -X GET http://$PROXY_HOST:$PROXY_PORT/weave ; then
+        return 0
+      else
+        sleep 0.5
```
rade: fractional sleeps are not in posix.

weave
```diff
@@ -128,6 +129,55 @@ collect_ip_args() {
...
+            PROXY_HOST="localhost"
+            ;;
+    esac
+}
+
+wait_for_proxy_to_boot() {
+    for n in `seq 1 20`; do
+      if curl --max-time 1 --connect-timeout 1 -s -X GET http://$PROXY_HOST:$PROXY_PORT/weave ; then
+        return 0
+      else
```
rade: drop the `else`.

weave
```diff
@@ -128,6 +129,55 @@ collect_ip_args() {
...
+    esac
+    PROXY_CONTAINER=$(docker run --privileged --net=host -d \
+        --name=$PROXY_CONTAINER_NAME \
+        -p $PROXY_PORT:12375/tcp -v /var/run/docker.sock:/var/run/docker.sock \
+        -v /proc:/hostproc -e PROCFS=/hostproc \
+        --entrypoint=/home/weave/proxy $EXEC_IMAGE "$@")
+
+    wait_for_proxy_to_boot
+}
+
```
rade: all the code added above this point is in the wrong place. Only code needed for `run` and `exec_remote` should go there.

weave
```diff
@@ -128,6 +129,55 @@ collect_ip_args() {
...
+            ;;
+        false*)
+            docker rm -f $PROXY_CONTAINER_NAME >/dev/null
+            ;;
+    esac
+    PROXY_CONTAINER=$(docker run --privileged --net=host -d \
+        --name=$PROXY_CONTAINER_NAME \
+        -p $PROXY_PORT:12375/tcp -v /var/run/docker.sock:/var/run/docker.sock \
+        -v /proc:/hostproc -e PROCFS=/hostproc \
+        --entrypoint=/home/weave/proxy $EXEC_IMAGE "$@")
```
rade: we should have a `WEAVEPROXY_DOCKER_ARGS` env var here, threaded through like `WEAVE_DOCKER_ARGS` and friends.

weave
```diff
@@ -128,6 +129,55 @@ collect_ip_args() {
...
+      else
+        sleep 0.5
+      fi
+    done
+    echo "Proxy did not start within 10 seconds" >&2
+    return 1
+}
+
+ensure_proxy_running() {
+    proxy_host
```
rade: that call should be moved down, since we don't need the result for the "is it already there?" check.

rade:
btw, in the interest of getting this merged, we could leave out the DNS support and add file an issue for adding it later.

proxy/client.go
```diff
@@ -0,0 +1,169 @@
...
+		Dial:        dial,
+	}
+}
+
+func (c *client) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	req, err := c.InterceptRequest(r)
+	if err == docker.ErrNoSuchImage {
+		http.NotFound(w, r)
+		return
+	} else if err != nil {
```
rade: drop the `else`.

proxy/client.go
```diff
@@ -0,0 +1,169 @@
...
+			Warning.Print(err)
+		}
+	}()
+	go func() {
+		if _, err := io.Copy(up, downBuf); err != nil {
+			Warning.Print(err)
+		}
+		err = up.(interface {
+			CloseWrite() error
+		}).CloseWrite()
```
rade: wow.

proxy/client.go
```diff
@@ -0,0 +1,169 @@
...
+		return
+	}
+
+	go func() {
+		defer close(end)
+
+		if _, err := io.Copy(down, io.MultiReader(rem, up)); err != nil {
+			Warning.Print(err)
+		}
+	}()
```
rade: why are we running this in a goroutine and do this whole `end` dance? Could just run it directly, after creating the 'down' goroutine.

proxy/client.go
```diff
@@ -0,0 +1,169 @@
...
+		if _, err := io.Copy(up, downBuf); err != nil {
+			Warning.Print(err)
+		}
+		err = up.(interface {
+			CloseWrite() error
+		}).CloseWrite()
+		if err != nil {
+			Debug.Printf("Error Closing upstream: %s", err)
+		}
+	}()
```
rade: So this goroutine can stick around beyond the termination of the outer function. Whereas the 'up' goroutine doesn't. is that intentional?

proxy/client.go
```diff
@@ -0,0 +1,169 @@
...
+	// Because we can't go back to request/response after we
+	// hijack the connection, we need to close it and make the
+	// client open another.
+	w.Header().Add("Connection", "close")
+	w.WriteHeader(resp.StatusCode)
+
+	down, _, up, rem, err := hijack(w, client)
+	defer up.Close()
+	defer down.Close()
+	if err != nil {
```
rade: won't this end up calling `nil.Close()` in the event of error?

proxy/client.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+}
+
+func doRawStream(w http.ResponseWriter, resp *http.Response, client *httputil.ClientConn) {
+	down, downBuf, up, rem, err := hijack(w, client)
+	defer down.Close()
+	defer up.Close()
+
+	if err != nil {
```
rade: won't this end up calling `nil.Close()` in the event of error?

proxy/common.go
```diff
@@ -0,0 +1,36 @@
...
+	. "github.com/weaveworks/weave/common"
+)
+
+func cidrsToAddresses(cidrs []string) []string {
+	addresses := []string{}
+	for _, cidr := range cidrs {
+		addresses = append(addresses, strings.Split(cidr, "/")[0])
+	}
+	return addresses
+}
```
rade: This function appears to be unused.

proxy/create_container_interceptor.go
```diff
@@ -0,0 +1,141 @@
...
+		if segments[i] == "--dns" {
+			dockerBridgeIP = segments[i+1]
+		}
+	}
+
+	if dockerBridgeIP == "" {
+		return "", fmt.Errorf("Docker bridge IP not found in weave output: %s", string(out))
+	}
+	return dockerBridgeIP, nil
+}
```
rade: Gross. I suggest adding a `weave docker-bridge-ip` command. Hidden just like `dns-args`.

proxy/proxy.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (proxy *proxy) transport() *http.Transport {
+	return &http.Transport{
+		Dial: func(string, string) (conn net.Conn, err error) {
+			conn, err = proxy.Dial()
+			return
+		},
+	}
+}
```
rade: All the above functions should be inlined.

proxy/start_container_interceptor.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return res, nil
+}
+
+func containerInfo(transport http.RoundTripper, containerId string) (body map[string]interface{}, err error) {
+	body = nil
+	client := &http.Client{
+		Transport: transport,
+	}
+	if res, err := client.Get("http://localhost/v1.16/containers/" + containerId + "/json"); err == nil {
```
rade: that version number should not be hard-coded here. should be a constant at least, but ideally it would be set at compile time.

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -0,0 +1,17 @@
...
+# Ensure the base image does not exist, so that it will be pulled
+if (docker_on $HOST1 images $BASE_IMAGE | grep $BASE_IMAGE); then
+  docker_on $HOST1 rmi $BASE_IMAGE
+fi
+
+assert_raises "docker_proxy_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' --name=c1 $BASE_IMAGE ip link show ethwe" 0
+assert "docker_on $HOST1 inspect --format='{{.Config.Entrypoint}}' c1" "[/home/weavewait/weavewait]"
+assert "docker_on $HOST1 inspect --format='{{.Config.Cmd}}' c1" "[ip link show ethwe]"
+
+end_suite
```
rade: This whole test is too low-level. Smoke tests are meant to simulate what a user/app would do. So I think the first line is enough, though of course it could succeed by luck of timing. Nothing we can do about w/o jumping through hoops.

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -0,0 +1,17 @@
...
+
+start_suite "Proxy waits for weave to be ready before running container commands"
+weave_on $HOST1 launch-proxy
+BASE_IMAGE=ubuntu
+# Ensure the base image does not exist, so that it will be pulled
+if (docker_on $HOST1 images $BASE_IMAGE | grep $BASE_IMAGE); then
+  docker_on $HOST1 rmi $BASE_IMAGE
+fi
+
+assert_raises "docker_proxy_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' --name=c1 $BASE_IMAGE ip link show ethwe" 0
```
rade: 0 is the default.

test/620_proxy_entrypoint_inheritance_test.sh
```diff
@@ -0,0 +1,19 @@
...
+  ENTRYPOINT ["ip", "link", "show", "ethwe"]
+EOF
+assert "docker_on $HOST1 inspect --format='{{.Config.Entrypoint}}' inspect-ethwe" "[ip link show ethwe]"
+assert "docker_on $HOST1 inspect --format='{{.Config.Cmd}}' inspect-ethwe" "<no value>"
+
+# Boot a new container with no entrypoint of its own, just a command
+assert_raises "docker_proxy_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' --name=c2 inspect-ethwe" 0
+assert "docker_on $HOST1 inspect --format='{{.Config.Entrypoint}}' c2" "[/home/weavewait/weavewait]"
+assert "docker_on $HOST1 inspect --format='{{.Config.Cmd}}' c2" "[ip link show ethwe]"
+end_suite
```
rade: This whole test is too low-level. We should test that the right entry point executes by checking for the result of running the container.

test/630_proxy_dns_test.sh
```diff
@@ -0,0 +1,21 @@
...
+
+weave_on $HOST1 launch-dns 10.2.254.1/24
+weave_on $HOST1 launch-proxy --with-dns
+docker_proxy_on $HOST1 run -e WEAVE_CIDR=$C2/24 -dt --name=c2 -h $NAME gliderlabs/alpine /bin/sh
+docker_proxy_on $HOST1 run -e WEAVE_CIDR=$C1/24 -dt --name=c1 aanand/docker-dnsutils /bin/sh
+
+assert_dns_record $HOST1 c1 $NAME $C2
+
+assert_dns_status $HOST1 c1 "MX   seetwo.weave.local" NXDOMAIN
+assert_dns_status $HOST1 c1 "AAAA seetwo.weave.local" NXDOMAIN
```
rade: These two asserts seem to have nothing to do with with we are trying to test here.

proxy/create_container_interceptor.go
rade: Also, this really should be invoked just once, when the proxy starts.

weavewait/main.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/weaveworks/weave/net"
+)
+
+func main() {
+	if _, err := net.EnsureInterface("ethwe", 30); err != nil {
```
rade: Should be 20s, same as stick in the Dockerfile for weaver and weavedns.

proxy/client.go
```diff
@@ -0,0 +1,169 @@
...
+	// Copy the chunked response body to downstream,
+	// stopping at the end of the chunked section.
+	rawResponseBody := io.MultiReader(rem, up)
+	if _, err := io.Copy(ioutil.Discard, httputil.NewChunkedReader(io.TeeReader(rawResponseBody, down))); err != nil {
+		Error.Fatal("Error copying chunked response body", http.StatusInternalServerError)
+		return
+	}
+	resp.Trailer.Write(down)
+	// a chunked response ends with a CRLF
+	down.Write([]byte{13, 10})
```
rade: `[]byte("\r\n")` (for consistency with the `[]byte("\n")` used elsewhere)

proxy/proxy.go
paulbellamy: Why?

rade: because they are short and add a level of indirection that offers no benefit.

paulbellamy: Offers the benefit of telling what the code is expected to do.

proxy/proxy.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (proxy *proxy) ProxyRequest(w http.ResponseWriter, r *http.Request) {
+	NewClient(proxy.Dial, NullInterceptor).ServeHTTP(w, r)
+}
+
+// Supplied so that we can use with http.Client as a Transport
+func (proxy *proxy) RoundTrip(req *http.Request) (*http.Response, error) {
+	res, err := proxy.transport().RoundTrip(req)
+	return res, err
+}
```
dpw: This can just do `return proxy.transport().RoundTrip(req)`

proxy/proxy.go
rade: You could have a single helper function that does `NewClient(proxy.Dial, interceptor).ServeHTTP(w,r)`.

I'm guessing `RoundTrip` cannot be inlined because it is part of some interface. But `transport` can.

paulbellamy:
The proxy/interceptors thing is probably a bit overwrought, but I can revisit that while doing TLS.

errordeveloper:
Works with Swarm and Compose for me :+1:

paulbellamy:
For other issues related to this:

* Provides a workaround for #47
* Mostly solves #230, though passing network info is still a bit unwieldy
* Same for #117, and inspect still won't show IPs attached after creation.
* Partially solves #400 (except for IPs attached after creation, and docker restart policies)
* Does not solve #401 (docker restart policies)
* SSL was pulled out into separate issue (#506)

proxy/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM scratch
+MAINTAINER Weaveworks Inc <help@weave.works>
+WORKDIR /home/weave
+ADD ./proxy /home/weave/
+ENTRYPOINT ["/home/weave/proxy", "-L", ":12375"]
+CMD ["-H", "unix:///var/run/docker.sock"]
```
rade: I gather the proxy is actually embedded in the weaveexec image, so this file should go.

site/experimental-features.md
```diff
@@ -0,0 +1,63 @@
+---
+title: Experimental features
```
rade: Let's instead have a proxy.md, and a short section in features.md - analogous to what we have for weaveDNS.

paulbellamy:
rebased on upstream/master

rade:
we really want a separate `proxy` package, rather than having everything in `main`.

proxy/client.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	err = down.(interface {
+		CloseWrite() error
+	}).CloseWrite()
+	if err != nil {
+		Debug.Printf("Error Closing downstream: %s", err)
+	}
+
+	<-end
+}
```
rade: I have only just realised how similar the up and down paths are. So I reckon we should shove the common code into a helper function that we can call with `go copyStream(up, downBuf, upDone)` and `go copyStream(down, io.MultiReader(rem, up), downDone)`, and then we can run

````
select {
    case <-upDone:
    case <-downDone:
}
````
to wait for either to finish. (assuming that is the semantics we want).

**Comment by rade** [comment]
> Partially solves #400 (except for IPs attached after creation, and docker restart policies)

#400 only covers the `docker restart` command; restart policies are the subject of #401. As for IPs attached after creation, I think we should file a separate issue for that. Ditto for the similar leftover from #117.

So, in summary, this PR should close #47,#117,#230,#400. It may not be the perfect solution for any of them, but let our users be the judge of that.

We should add a test for `docker restart`.

**Comment by rade** [comment]
On reflection, I think it's a bit of a stretch to claim that this solves ##117. So let's leave that one open. I have split off the "dynamically attached" part of #400 into #688.

So, in summary, this PR should close #47,#230,#251,#400.

Things left to do here:

- [ ] make sure stream proxying correctly handles half-closed connections
- [ ] add a smoke test for re-attachement on `docker restart`, including DNS re-population

**Comment by rade** [comment]
Please merge/cherry-pick https://github.com/rade/weave/commit/42cf68959b3194ea2c0cc90425aa868d1a657945 into this.

File: proxy/interceptor.go [hunk 43]
```diff
@@ -0,0 +1,19 @@
...
+type nullInterceptor struct {
+}
+
+func (i nullInterceptor) InterceptRequest(r *http.Request) (*http.Request, error) {
+	return r, nil
+}
+
+func (i nullInterceptor) InterceptResponse(r *http.Response) (*http.Response, error) {
+	return r, nil
+}
```
rade: I don't understand why these functions return a request/response, respectively. In all implementations we actually *modify* the request/response given as a parameter. Do you envisage situations where we'd want to construct completely fresh request/response objects? Even if we do, it's not something needed atm, so I'd favour in keeping the API minimal. Also, the fact that these functions *may* mutate the request/response should be noted.

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
+
+. ./config.sh
+
+start_suite "Proxy waits for weave to be ready before running container commands"
+weave_on $HOST1 launch-proxy
+BASE_IMAGE=ubuntu
```
rade: please use `gliderlabs/alpine` here. It's smaller than ubuntu, has all you need, and is an image we use in other tests.

test/640_proxy_restart_reattaches_test.sh
```diff
@@ -0,0 +1,20 @@
...
+
+start_suite "Proxy restart reattaches networking to containers"
+
+weave_on $HOST1 launch-dns 10.2.254.1/24
+weave_on $HOST1 launch-proxy --with-dns
+docker_proxy_on $HOST1 run -e WEAVE_CIDR=$C2/24 -dt --name=c2 -h $NAME gliderlabs/alpine /bin/sh
+docker_proxy_on $HOST1 run -e WEAVE_CIDR=$C1/24 -dt --name=c1 aanand/docker-dnsutils /bin/sh
+
+docker_proxy_on $HOST1 restart c2
+assert_raises "exec_on $HOST1 c1 ip link show ethwe | grep 'state UP'"
```
rade: Surely that should be `exec_on $HOST1 c2 ...`

test/620_proxy_entrypoint_inheritance_test.sh
```diff
@@ -0,0 +1,16 @@
...
+start_suite "Proxy uses entrypoint from the container with weavewait"
+# Setup and sanity-check
+weave_on $HOST1 launch-proxy
+docker_on $HOST1 build -t inspect-ethwe - <<- EOF
+  FROM gliderlabs/alpine
+  ENTRYPOINT ["ip", "link", "show", "ethwe"]
+EOF
+
+# Boot a new container with no entrypoint of its own, just a command
+assert_raises "docker_proxy_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' inspect-ethwe | grep 'state UP'"
```
rade: "just a command" - I can't see any command here.

test/620_proxy_entrypoint_inheritance_test.sh
```diff
@@ -0,0 +1,16 @@
...
+
+end_suite
```
rade: Surely we should also test that overriding the entrypoint with `--entrypoint` works.

weaveproxy/main.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func main() {
+	var target, listen string
+	var withDNS, debug bool
+
+	getopt.BoolVarLong(&debug, "debug", 'd', "log debugging information")
+	getopt.StringVar(&target, 'H', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultTarget))
+	getopt.StringVar(&listen, 'L', fmt.Sprintf("address on which to listen (default %s)", defaultListen))
+	getopt.BoolVarLong(&withDNS, "with-dns", 'w', "register connected containers with weaveDNS")
```
rade: The description of `with-dns` is wrong.

proxy/client.go
```diff
@@ -0,0 +1,167 @@
...
+		if _, err := io.Copy(w, resp.Body); err != nil {
+			Warning.Print(err)
+		}
+	}
+}
+
+func doRawStream(w http.ResponseWriter, resp *http.Response, client *httputil.ClientConn) {
+	down, downBuf, up, rem, err := hijack(w, client)
+	if err != nil {
+		Error.Fatal(w, "Unable to hijack connection for raw stream mode", http.StatusInternalServerError)
```
rade: Do we really want this to be fatal?

proxy/client.go
```diff
@@ -0,0 +1,167 @@
...
+func doChunkedResponse(w http.ResponseWriter, resp *http.Response, client *httputil.ClientConn) {
+	// Because we can't go back to request/response after we
+	// hijack the connection, we need to close it and make the
+	// client open another.
+	w.Header().Add("Connection", "close")
+	w.WriteHeader(resp.StatusCode)
+
+	down, _, up, rem, err := hijack(w, client)
+	if err != nil {
+		Error.Fatal("Unable to hijack response stream for chunked response", http.StatusInternalServerError)
```
rade: Do we really want this to be fatal?

proxy/client.go
```diff
@@ -0,0 +1,167 @@
...
+		return
+	}
+	defer up.Close()
+	defer down.Close()
+	// Copy the chunked response body to downstream,
+	// stopping at the end of the chunked section.
+	rawResponseBody := io.MultiReader(rem, up)
+	if _, err := io.Copy(ioutil.Discard, httputil.NewChunkedReader(io.TeeReader(rawResponseBody, down))); err != nil {
+		Error.Fatal("Error copying chunked response body", http.StatusInternalServerError)
```
rade: Do we really want this to be fatal?

rade:
The proxy tests take longer to run than the entire existing test suite! 4 mins vs 3min30s. Over half of that is in running the docker integration test suite. Not sure what, if anything, to do about this.

test/610_proxy_wait_for_weave_test.sh
paulbellamy: Can't remove `gliderlabs/alpine`, because it's used by weaveexec. But, I'll swap to busybox, which is small enough.

paulbellamy:
Yeah, I agree the docker-py tests are quite slow. They have several sleeps in them. I had tried, the go-dockerclient tests, but they just use a mock server.

Overall, I'm not sure what to do about it at this point.

paulbellamy:
Notes after f2f with @rade, the docker-py tests are quite slow, and *possibly* a bit flakey, but have been useful in catching proxy bugs. If they prove troublesome, could check if there is a cli suite as part of the main docker repo.

Or put some work into fixing up/speeding up the docker-py tests...

squaremo:
100!

PR #665, opened 2015-05-15
Merged two run commands into one in Dockerfile

pradeepchhetri:


rade:
Why? To save a layer?

@squaremo thoughts?

pradeepchhetri:
@rade Yes, i guess this is what Dockerfile's best practices also says.

PR #667, opened 2015-05-15
revise weaveDNS docs

rade:
- add ToC and anchors for all headings
- move docker bridge settings into "How it works" - it's a rarely used
  feature
- highlight a bunch of features, extracting them from obscurely titled
  sub-sections of "manual usage"
  - basic rendundancy
  - adding/removing extra DNS entries
  - service hot-swapping
  - ability to retain entries for stopped containers
  - using a different local domain
- rename remainder of "manual usage" section to "Using weaveDNS
  without `weave run`"
- remove "Unregistering" section - it was out of date and is not
  covered by the "add/remove" section

### DIFF
--- a/site/weavedns.md
+++ b/site/weavedns.md
@@ -160,10 +160,10 @@ link-local as per [RFC6762](https://tools.ietf.org/html/rfc6762),
 
 ## <a name="without-run"></a>Using weaveDNS without `weave run`
 
-If you use the `--with-dns` option, `weave run` automatically supplies
-the DNS server address to the new container. And both `weave run` and
-`weave attach` register the hostname of the given container against
-the given weave network IP address.
+When weaveDNS is running, both `weave run` and `weave attach` register
+the hostname of the given container against the given weave network IP
+address. And if you use the `--with-dns` option, `weave run`
+automatically supplies the DNS server address to the new container.
 
 In some circumstances, you may not want to use the `weave run` command
 to start containers. You can still take advantage of a running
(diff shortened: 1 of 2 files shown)


site/weavedns.md
```diff
@@ -32,6 +43,12 @@ $ docker attach $shell1
 ...
 ```
 
+Note that it is permissible to register multiple containers with the
+same name. weaveDNS picks one address to return when asked for the
+name. Since weaveDNS removes any container that dies, this is a simple
```
awh: Inconsistent use of 'weaveDNS' vs 'WeaveDNS' here an in other places

site/weavedns.md
```diff
@@ -78,30 +143,31 @@ weave run --with-dns 10.2.1.4/24 -ti \
...
-So that containers can connect to a stable and always routable IP
-address, weaveDNS publishes its port 53 to the Docker bridge device,
-which is assumed to be `docker0`.
+## <a name="without-run"></a>Using weaveDNS without `weave run`
 
-Some configurations may use a different Docker bridge device. To
-supply a different bridge device, use the environment variable
-`DOCKER_BRIDGE`, e.g.,
+If you use the `--with-dns` option, `weave run` automatically supplies
+the DNS server address to the new container. And both `weave run` and
```
awh: This 'And' is confusing, because it implies that `run` and `attach` only register if `--with-dns` has been supplied, when in fact they are othogonal.

site/weavedns.md
rade: Nobody knows how it's meant to be spelled. I did not introduce this inconsistency and am not going to attempt to resolve it in this PR.

rade: I have removed most of the inconsistency.

site/weavedns.md
rade: Fixed

PR #695, opened 2015-05-18
merge IPAM tests with rest

rade:
keeping them separate leads to too much duplication


bboreham:
I note that, after these changes, there is really only one test that launches weave cross-host without the `-iprange` option - `110_encryption_test.sh`.  We should take care that this number doesn't go to zero.

Issue #656, opened 2015-05-13
HTTP/JSON interface issues


abligh:
I've come across some practical problems using the HTTP/JSON interface.

What I want to do in essence is replace the current set of peers with a known good set of peers.

Issue 1: I can only add (using `/connect`) or remove (using `/forget`) peers, not specify a list of peers replacing the current set. It would be nice if we could do an atomic replace of the peer list with the new peer list, forgetting any that aren't on the new list, and connecting to any new ones. However, you can only add or remove peers one by one.

Issue 2: in order to work around that deficiency, I need to get the list of current peers from weave (using `/status-json`), parse the JSON (which is a pain), then add or remove the peers as appropriate. There are two sub-issues here: 2A. `/status-json` does not list the peer IP address and port, so it's impossible to find a given peer. 2B. if the peer has never connected at all (i.e. it's merely configured), it's not listed in `/status-json` at all; this means that if for example a peer was added which isn't running weave (e.g. is down), it would never get removed. Either that, or `/status-json`'s `Peers` entry is meant to return only a list of peers that it has negotiated with, in which case it would be very useful to return a list of peers that are configured (perhaps separately).

My current workaround is 'restart `weaver`' which is not nice.

I think Issue #1 is best addressed through another entry in `HandleHTTP` which simply takes a comma delimited list of peers and does the appropriate thing. This avoids the need to incorporate a JSON library my end. I suggest a URL of `/replace`. I'm quite happy to add this myself. However, I'm not entirely sure how to forget a connection that has been initiated but not yet connected - there's a race condition here I think.

I think Issue #2A and #2B are straight bugs. WDYT?

rade:
> It would be nice if we could do an atomic replace of the peer list

#658 

> get the list of current peers from weave

#659, #660.

Issue #658, opened 2015-05-14
`weave connect --replace` to replace command line peer addresses


rade:
`weave connect` and `weave forget` allow dynamic addition/removal of peer addresses, but we are lacking an operation to 'replace' all addresses, which is useful in some automation scenarios.

I suggest adding a `--replace` flag to `weave connect`. Happy to entertain suggestions for adding a separate command.

NB: `weave connect --replace` w/o any addresses would remove all command line addresses.

PR #697, opened 2015-05-19
weave connect/forget take multiple peers

paulbellamy:
A first step towards solving #656.

Are there some tests for this that should be updated?

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -48,30 +48,38 @@ func (cm *ConnectionMaker) Start() {
 	go cm.queryLoop(actionChan)
 }
 
-func (cm *ConnectionMaker) InitiateConnection(peer string) error {
-	host, port, err := net.SplitHostPort(peer)
-	if err != nil {
-		host = peer
-		port = "0" // we use that as an indication that "no port was supplied"
-	}
-	addr, err := net.ResolveTCPAddr("tcp4", fmt.Sprintf("%s:%s", host, port))
-	if err != nil {
-		return err
+func (cm *ConnectionMaker) InitiateConnections(peers []string) error {
+	addrs := make(map[string]*net.TCPAddr)
+	for _, peer := range peers {
+		host, port, err := net.SplitHostPort(peer)
+		if err != nil {
+			host = peer
+			port = "0" // we use that as an indication that "no port was supplied"
+		}
+		addr, err := net.ResolveTCPAddr("tcp4", fmt.Sprintf("%s:%s", host, port))
+		if err != nil {
+			return err
+		}
+		addrs[peer] = addr
 	}
 	cm.actionChan <- func() bool {
-		cm.cmdLinePeers[peer] = addr
-		// curtail any existing reconnect interval
-		if target, found := cm.targets[addr.String()]; found {
-			target.tryAfter, target.tryInterval = tryImmediately()
+		for peer, addr := range addrs {
+			cm.cmdLinePeers[peer] = addr
+			// curtail any existing reconnect interval
+			if target, found := cm.targets[addr.String()]; found {
+				target.tryAfter, target.tryInterval = tryImmediately()
+			}
 		}
 		return true
 	}
 	return nil
 }
 
-func (cm *ConnectionMaker) ForgetConnection(peer string) {
+func (cm *ConnectionMaker) ForgetConnections(peers []string) {
 	cm.actionChan <- func() bool {
-		delete(cm.cmdLinePeers, peer)
+		for _, peer := range peers {
+			delete(cm.cmdLinePeers, peer)
+		}
 		return false
 	}
 }
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -48,7 +48,8 @@ func (cm *ConnectionMaker) Start() {
 	go cm.queryLoop(actionChan)
 }
 
-func (cm *ConnectionMaker) InitiateConnections(peers []string) error {
+func (cm *ConnectionMaker) InitiateConnections(peers []string) []error {
+	errors := []error{}
 	addrs := make(map[string]*net.TCPAddr)
 	for _, peer := range peers {
 		host, port, err := net.SplitHostPort(peer)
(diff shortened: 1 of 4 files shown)


router/connection_maker.go
```diff
@@ -48,30 +48,38 @@ func (cm *ConnectionMaker) Start() {
...
+	addrs := make(map[string]*net.TCPAddr)
+	for _, peer := range peers {
+		host, port, err := net.SplitHostPort(peer)
+		if err != nil {
+			host = peer
+			port = "0" // we use that as an indication that "no port was supplied"
+		}
+		addr, err := net.ResolveTCPAddr("tcp4", fmt.Sprintf("%s:%s", host, port))
+		if err != nil {
+			return err
```
rade: This fails if one fails, which is a) bad, and b) a change from existing behaviour.

weave
```diff
@@ -870,12 +875,12 @@ case "$COMMAND" in
         echo $PROXY_CONTAINER
         ;;
     connect)
-        [ $# -eq 1 ] || usage
-        http_call $CONTAINER_NAME $HTTP_PORT POST /connect -d "peer=$1"
+        [ $# -gt 0 ] || usage
+        http_call $CONTAINER_NAME $HTTP_PORT POST /connect -d `peer_args $@`
```
rade: We don't use backquotes. They are evil. use `$(...)` instead. And shouldn't that `$@` be quoted?

weave
```diff
@@ -554,6 +554,11 @@ echo_addresses() {
     echo $1 $2 $3
 }
 
+peer_args() {
+  local joined=$(printf "&peer=%s" "${@}")
+  echo ${joined:1}
+}
```
rade: @dpw is there a better way to do this, ideally completely inside sh?

Also, this is the first use of a local var in the script. How do we feel about that?

router/connection_maker.go
paulbellamy: If we are trying to later do an atomic replace of a all peers, I'd personally expect it to replace all-or-none of the peers I provided.

Since it only took one peer before, I'm not sure which existing behaviour you're talking about.

paulbellamy: Ah, weave launch...

paulbellamy: Let's have it be best-effort, and output any it failed on (exiting 1 if any failed).

weave
dpw:
```
peer_args() {
  res=''
  sep=''
  for p in "$@" ; do
      res="$res${sep}peer=$p"
      sep="&"
  done
  echo "$res"
}
```

(Note that `local` is not in POSIX, even though dash supports it. I don't know whether we care about that.)

rade:
> Are there some tests for this that should be updated?

I believe you were going to write some ;)

paulbellamy:
Upon further inspection the old (and new) behaviour of `weave launch` is to `log.Fatal` if any addresses fail to resolve. `weave connect` will now connect to as many as possible, and output errors for any that fail.

router/connection_maker.go
```diff
@@ -48,30 +48,42 @@ func (cm *ConnectionMaker) Start() {
...
+			continue
+		}
+		addrs[peer] = addr
 	}
-	cm.actionChan <- func() bool {
-		cm.cmdLinePeers[peer] = addr
-		// curtail any existing reconnect interval
-		if target, found := cm.targets[addr.String()]; found {
-			target.tryAfter, target.tryInterval = tryImmediately()
+	if len(addrs) > 0 {
```
rade: premature optimisation. Drop it.

test/150_connect_test.sh
```diff
@@ -0,0 +1,20 @@
...
+weave_on $HOST2 launch
+
+start_container $HOST1 $C1/24 --name=c1
+start_container $HOST2 $C2/24 --name=c2
+
+assert_raises "exec_on $HOST1 c1 $PING $C2" "1"
+assert_raises "weave_on $HOST2 connect $HOST1"
+assert_raises "exec_on $HOST1 c1 $PING $C2"
+
+end_suite
```
rade: so where are the tests for n-ary `connect` and `forget`?

weaver/main.go
```diff
@@ -143,7 +144,9 @@ func main() {
 	}
 
 	router.Start()
-	initiateConnections(router, peers)
+	if errors := router.ConnectionMaker.InitiateConnections(peers); len(errors) > 0 {
+		log.Fatal(strings.Join(errorMessages(errors), ", "))
```
rade: That looks gross in the logs. Make it a "\n", which also means the `join` can be handled by `errorMessages`, since "\n" is what gets used at the other call site

weaver/main.go
```diff
@@ -250,13 +253,19 @@ func handleHTTP(router *weave.Router, httpAddr string, allocator *ipam.Allocator
 	})
 
 	muxRouter.Methods("POST").Path("/connect").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if err := router.ConnectionMaker.InitiateConnection(r.FormValue("peer")); err != nil {
-			http.Error(w, fmt.Sprint("invalid peer address: ", err), http.StatusBadRequest)
+		if err := r.ParseForm(); err != nil {
+			http.Error(w, fmt.Sprint("unable to parse form: ", err), http.StatusBadRequest)
+		}
+		if errors := router.ConnectionMaker.InitiateConnections(r.Form["peer"]); len(errors) > 0 {
+			http.Error(w, fmt.Sprint("invalid peer addresses:\n", strings.Join(errorMessages(errors), "\n")), http.StatusBadRequest)
```
rade: That "invalid peer addresses:\n" looks quite gross in the CLI. Drop it.

router/connection_maker.go
```diff
@@ -48,30 +48,42 @@ func (cm *ConnectionMaker) Start() {
...
+		if err != nil {
+			host = peer
+			port = "0" // we use that as an indication that "no port was supplied"
+		}
+		addr, err := net.ResolveTCPAddr("tcp4", fmt.Sprintf("%s:%s", host, port))
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		addrs[peer] = addr
```
rade: make this an `if-then-else`, including the `addr, err := ...`

weaver/main.go
```diff
@@ -155,6 +158,14 @@ func main() {
 	SignalHandlerLoop(router)
 }
 
+func errorMessages(errors []error) []string {
+	result := []string{}
```
rade: iirc the linter usually whines about this, and wants `var result []string`.

test/150_connect_test.sh
paulbellamy: Not sure what to do about those, as we only have 2 vms when testing.

rade: Well, you could still test that `weave connect/forget` succeed for multiple args.

Also should test failure cases.

Also just noticed that there is no `forget` test at all. That could be be tested by invoking `forget` followed by `weave stop` and `weave launch` of the other weave, and then checking that there is no connectivity.

abligh:
Thanks.

So I don't think there is currently a way to *replace* all the direct connections with a list provided by JSON - i.e. add the ones that aren't currently on the direct peer list and forget the others. This would be really handy as it would save parsing the JSON at all. Would you take a patch for that?

rade:
> So I don't think there is currently a way to replace all the direct connections

#658

Issue #684, opened 2015-05-18
[ipam] log message from http API is not very clear


bboreham:
Example command:

    curl -v -X DELETE http://172.17.0.14:6784/ip/foo

Log message:

    WARNING: 2015/05/18 09:48:53.910295 No addresses for foo

The message should at least mention ipam, and possibly refer to the DELETE operation, or 'free' which is what is implied by DELETE.

PR #701, opened 2015-05-19
Improve ipam error messages

bboreham:
Make clear that they come from the IP allocator, and reduce repetition in the text.
Also include the nickname for other peer, if known.
Addresses #684


rade:
the linter is complaining.

bboreham:
Good catch, linter!

PR #702, opened 2015-05-19
proxy needs to add weavewait to execs as well

paulbellamy:
Bug came up in a test where we are restarting and doing an exec checking
a network interface right away. weavewait is only added if the container
has a WEAVE_CIDR var set, the same as at container creation.

### DIFF
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -8,6 +8,10 @@ import (
 	. "github.com/weaveworks/weave/common"
 )
 
+var (
+	weaveWaitEntrypoint = []string{"/home/weavewait/weavewait"}
+)
+
 func callWeave(args ...string) ([]byte, error) {
 	args = append([]string{"--local"}, args...)
 	Debug.Print("Calling weave", args)
--- a/proxy/create_exec_interceptor.go
+++ b/proxy/create_exec_interceptor.go
@@ -1,0 +1,65 @@
+package proxy
+
+import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/fsouza/go-dockerclient"
+	. "github.com/weaveworks/weave/common"
+)
+
+type createExecInterceptor struct {
+	client *docker.Client
+}
+
+type createExecRequestBody struct {
+	*docker.Config
+	HostConfig *docker.HostConfig `json:"HostConfig,omitempty" yaml:"HostConfig,omitempty"`
+	MacAddress string             `json:"MacAddress,omitempty" yaml:"MacAddress,omitempty"`
+}
+
+func (i *createExecInterceptor) InterceptRequest(r *http.Request) error {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+	r.Body.Close()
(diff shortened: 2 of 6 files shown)


proxy/common.go
```diff
@@ -8,6 +8,10 @@ import (
 	. "github.com/weaveworks/weave/common"
 )
 
+var (
+	weaveWaitEntrypoint = []string{"/home/weavewait/weavewait"}
+)
+
```
rade: shouldn't this be a const?

proxy/create_exec_interceptor.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	r.Body.Close()
+	Info.Println("CreateExecInterceptor Intercepting: ", string(body))
+
+	options := docker.CreateExecOptions{}
+	if err := json.Unmarshal(body, &options); err != nil {
+		return err
+	}
+
+	subs := containerIDRegexp.FindStringSubmatch(r.URL.Path)
```
rade: this is accessing `containerIDRegexp` from `start_container_interceptor.go`. If it's a common var then it should live in a more common file.

proxy/create_exec_interceptor.go
```diff
@@ -0,0 +1,65 @@
...
+	MacAddress string             `json:"MacAddress,omitempty" yaml:"MacAddress,omitempty"`
+}
+
+func (i *createExecInterceptor) InterceptRequest(r *http.Request) error {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+	r.Body.Close()
+	Info.Println("CreateExecInterceptor Intercepting: ", string(body))
```
rade: we don't have a similar `Info` line in the other interceptors.

proxy/create_exec_interceptor.go
rade: though actually perhaps we should have helper function that extracts the container id from a path. so we don't end up duplicating the following lines. Perhaps even include the `InspectContainer`.

proxy/common.go
paulbellamy: Can't have a const array with some weird hacks.

proxy/create_exec_interceptor.go
paulbellamy: ah yeah. forgot to remove that, thanks.

proxy/create_exec_interceptor.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	newBody, err := json.Marshal(options)
+	if err != nil {
+		return err
+	}
+	r.Body = ioutil.NopCloser(bytes.NewReader(newBody))
+	r.ContentLength = int64(len(newBody))
+
+	return nil
```
rade: This is 8 lines of c&p from `createContainerInterceptor.InterceptRequest`. Introduce a helper function.

proxy/common.go
rade: I was afraid of that. *sigh*. never mind.

paulbellamy:
github won't let me re-open this, so see #703. Eugh.

PR #703, opened 2015-05-19
proxy needs to add weavewait to execs as well

paulbellamy:
Bug came up in a test where we are restarting and doing an exec checking
a network interface right away. weavewait is only added if the container
has a WEAVE_CIDR var set, the same as at container creation.

### DIFF
--- a/proxy/client.go
+++ b/proxy/client.go
@@ -27,14 +27,17 @@ func newClient(dial func() (net.Conn, error), i interceptor) *client {
 }
 
 func (c *client) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	err := c.InterceptRequest(r)
-	if err == docker.ErrNoSuchImage {
-		http.NotFound(w, r)
-		return
-	}
-	if err != nil {
-		http.Error(w, "Unable to intercept request", http.StatusInternalServerError)
-		Warning.Print("Error intercepting request: ", err)
+	if err := c.InterceptRequest(r); err != nil {
+		_, isNoSuchContainer := err.(*docker.NoSuchContainer)
+		switch {
+		case isNoSuchContainer:
+			http.Error(w, err.Error(), http.StatusNotFound)
+		case err == docker.ErrNoSuchImage:
+			http.Error(w, err.Error(), http.StatusNotFound)
+		default:
+			http.Error(w, "Unable to intercept request", http.StatusInternalServerError)
+			Warning.Print("Error intercepting request: ", err)
+		}
 		return
 	}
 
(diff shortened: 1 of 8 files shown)


proxy/client.go
```diff
@@ -29,12 +29,17 @@ func newClient(dial func() (net.Conn, error), i interceptor) *client {
...
 	if err != nil {
-		http.Error(w, "Unable to intercept request", http.StatusInternalServerError)
-		Warning.Print("Error intercepting request: ", err)
+		switch err.(type) {
+		case *docker.NoSuchContainer:
+			http.Error(w, err.Error(), http.StatusNotFound)
+		default:
+			http.Error(w, "Unable to intercept request", http.StatusInternalServerError)
+			Warning.Print("Error intercepting request: ", err)
+		}
```
rade: Sure the preceding `docker.ErrNoSuchImage` check should be folded into this `switch`.

proxy/create_exec_interceptor.go
```diff
@@ -0,0 +1,60 @@
...
+		Warning.Printf("No container id found in request with path %s", r.URL.Path)
+		return nil
+	}
+	containerID := subs[1]
+
+	container, err := i.client.InspectContainer(containerID)
+	if err != nil {
+		Warning.Printf("Error inspecting container %s: %v", containerID, err)
+		return err
+	}
```
rade: we should have a `container, err := inspectContainerInPath(r.URL.Path)` here (and in `start_container_interceptor.InterceptResponse`), that does all the above, including printing of warnings.

proxy/client.go
paulbellamy: ErrNoSuchImage is not a type, just an `errors.New()`, so we need 2 separate checks.

rade: yuck. Ok. in that case fold it all into the true branch of `if err := c.InterceptRequest(r); err {`

rade: also, you could perhaps use the other form of `switch`, with an `==` in one case and a type check in the other.

paulbellamy: a switch with different types of things in the cases? lol go.

paulbellamy:
re: state UNKNOWN.
weave/net.EnsureInterface, does check the that the interface has the Up flag set, so it should definitely be waiting. I believe it's `ip link show ethwe` being a bit overly-specific. `ifconfig ethwe` shows the interface as UP.

rade:
That seems a little odd. Could it be timing related, e.g. perhaps `ifconfig` is slower than `ip`? Or perhaps `ip` works off some information that is updated asynchronously. I wonder whether our resident kernel sleuth knows the answer. @dpw?

PR #704, opened 2015-05-19
Drop traffic from docker0 bridge to weave bridge

### LINKED ISSUE #620: `weave expose` breaks subnet isolation
````
$ ./weave launch
5987350638a5ff67836575ff6964542a49b23ae7068d5aacb92be133b96f458e
$ ./weave expose 10.2.1.101/24
$ ./weave run 10.2.1.1/24 -ti ubuntu
6d32f68169305659b47b8a735765f49ca09c8c5027e858d1befc89485e4c2328
$ docker attach $(./weave run 10.2.2.1/24 -ti ubuntu)
root@4877c8267567:/# ping -c 1 10.2.1.1
PING 10.2.1.1 (10.2.1.1) 56(84) bytes of data.
64 bytes from 10.2.1.1: icmp_seq=1 ttl=63 time=0.059 ms

--- 10.2.1.1 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.059/0.059/0.059/0.000 ms
````
This happens even without the masquerading rules added by `weave expose`, i.e. simply giving the bridge an IP is enough.

Based on the interface packet stats, the request goes out through the default route via the eth0 interface to the docker bridge, the kernel routes it to the weave bridge, since the destination IP is in the subnet of the bridge's IP, and the bridge routes it to the destination container's ethwe.

The reply apparently travels the reverse route, i.e. out of the ethwe interface, over the weave bridge, docker bridge and into the eth0 of the other container. I cannot work out why that happens - the reply destination ought to be in the wrong subnet to go through ethwe.

### PR DESCRIPTION
**Pull request by bboreham**
Fixes #620, as discussed in comments there.  All smoke tests still pass.

### DIFF
--- a/test/130_expose_test.sh
+++ b/test/130_expose_test.sh
@@ -25,10 +25,9 @@ check_container_connectivity() {
     exec_on1 "c1 $PING $C2"
     exec_on1 "c3 $PING $C4"
     exec_on1 "c5 $PING $C6"
-    # fails due to #620
-    # exec_on1 "c3 ! $PING $C1"
-    # exec_on1 "c5 ! $PING $C1"
-    # exec_on1 "c5 ! $PING $C3"
+    exec_on1 "c3 sh -c \"! $PING $C1\""
+    exec_on1 "c5 sh -c \"! $PING $C1\""
+    exec_on1 "c5 sh -c \"! $PING $C3\""
 }
 
 start_suite "exposing weave network to host"
--- a/test/130_expose_test.sh
+++ b/test/130_expose_test.sh
@@ -21,13 +21,14 @@ exec_on1() {
     assert_raises "exec_on  $HOST1 $@"
 }
 
+# Containers in the same subnet should be able to talk; different subnet not.
 check_container_connectivity() {
     exec_on1 "c1 $PING $C2"
     exec_on1 "c3 $PING $C4"
     exec_on1 "c5 $PING $C6"
-    exec_on1 "c3 sh -c \"! $PING $C1\""
+    exec_on1 "c1 sh -c \"! $PING $C3\""
+    exec_on1 "c3 sh -c \"! $PING $C5\""
     exec_on1 "c5 sh -c \"! $PING $C1\""
-    exec_on1 "c5 sh -c \"! $PING $C3\""
 }
 
 start_suite "exposing weave network to host"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by bboreham** [comment]
Hang on, need to amend the smoke test to check that this actually works

**Comment by bboreham** [comment]
OK, have extended the smoke test - had to kludge the exec_on thing a bit to get the negation in the right place.

**Comment by rade** [comment]
Why is the kludge necessary? Would wrapping in a `sh -c` execution do the trick?

**Comment by rade** [comment]
Alternatively, `assert_raises "exec_on  $HOST1 $@" 1` should work. No need to create a helper function for it, since there is just one call site.

**Comment by rade** [comment]
Except once you've rebased from master there will be three call sites, so an `exec_on1_fail` would probably still be a good idea.

**Comment by rade** [comment]
Hmm, there's a problem with both your existing `exec_on1_n` and my proposed `exec_on1_fail` - they can fail (and hence report success) for entirely the wrong reasons. That's why I went for the 'negate the result of the specific test we expect to fail' approach elsewhere. So perhaps a `sh -c` is the way to go after all.

**Comment by bboreham** [comment]
Changed to `sh -c` and rebased

File: test/130_expose_test.sh [hunk 1]
```diff
@@ -25,10 +25,9 @@ check_container_connectivity() {
     exec_on1 "c1 $PING $C2"
     exec_on1 "c3 $PING $C4"
     exec_on1 "c5 $PING $C6"
-    # fails due to #620
-    # exec_on1 "c3 ! $PING $C1"
-    # exec_on1 "c5 ! $PING $C1"
-    # exec_on1 "c5 ! $PING $C3"
+    exec_on1 "c3 sh -c \"! $PING $C1\""
+    exec_on1 "c5 sh -c \"! $PING $C1\""
+    exec_on1 "c5 sh -c \"! $PING $C3\""
```
rade: nitpick (and this is entirely my fault...)... the last three rules should be c1->$C3,c3->$C5,c5->$C1.

bboreham:
Annularized

rade:
I've just a) committed a [change](https://github.com/weaveworks/weave/commit/c5d5dd5fdc07239001cd853a9137091e6406d4c1) for simplicity and consistency.

PR #706, opened 2015-05-19
[dns] delete individual records

### LINKED ISSUE #631: [dns] permit removal of individual records
ATM, the DELETE operation in the HTTP API, and the underlying `Zone.DeleteRecord` operation, take a container and IP address. They do *not* accept a name. So even though it is possible to add multiple records for the same container-IP combination, with different names, it is not possible to remove such entries individually.

This functionality is needed for #364 and #627.

### LINKED ISSUE #654: [dns] slight inefficiency in weave script tell_dns
The function `tell_dns` is called in four places, two of which know the long-form container ID and two don't.  It would be more efficient to break out the part of tell_dns which re-fetches the long-form ID so we can bypass it in `weave run` and `weave start`.

awh:
Fixes #631
Fixes #654

### DIFF
--- a/nameserver/http.go
+++ b/nameserver/http.go
@@ -17,6 +17,13 @@ func httpErrorAndLog(level *log.Logger, w http.ResponseWriter, msg string,
 	level.Printf("[http] "+logmsg, logargs...)
 }
 
+func extractFQDN(r *http.Request) (string, string) {
+	if fqdnStr := r.FormValue("fqdn"); fqdnStr != "" {
+		return fqdnStr, fqdnStr
+	}
+	return "*", ""
+}
+
 func ServeHTTP(listener net.Listener, version string, server *DNSServer, domain string, db Zone) {
 
 	muxRouter := mux.NewRouter()
@@ -64,28 +71,39 @@ func ServeHTTP(listener net.Listener, version string, server *DNSServer, domain 
 	})
 
 	muxRouter.Methods("DELETE").Path("/name/{id:.+}/{ip:.+}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		reqError := func(msg string, logmsg string, logargs ...interface{}) {
-			httpErrorAndLog(Warning, w, msg, http.StatusBadRequest, logmsg, logargs...)
-		}
-
 		vars := mux.Vars(r)
 		idStr := vars["id"]
 		ipStr := vars["ip"]
 
 		ip := net.ParseIP(ipStr)
 		if ip == nil {
-			reqError("Invalid IP in request", "Invalid IP in request: %s", ipStr)
+			httpErrorAndLog(
+				Warning, w, "Invalid IP in request", http.StatusBadRequest,
+				"Invalid IP in request: %s", ipStr)
 			return
 		}
-		Info.Printf("[http] Deleting %s (%s)", idStr, ipStr)
-		if err := db.DeleteRecord(idStr, ip); err != nil {
-			if _, ok := err.(LookupError); !ok {
-				httpErrorAndLog(
-					Error, w, "Internal error", http.StatusInternalServerError,
-					"Unexpected error from DB: %s", err)
-				return
-			}
-		}
+
+		fqdnStr, fqdn := extractFQDN(r)
+
+		Info.Printf("[http] Deleting ID %s, IP %s, FQDN %s", idStr, ipStr, fqdnStr)
+		db.DeleteRecords(idStr, fqdn, ip)
+	})
+
+	muxRouter.Methods("DELETE").Path("/name/{id:.+}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		idStr := vars["id"]
+
+		fqdnStr, fqdn := extractFQDN(r)
+
+		Info.Printf("[http] Deleting ID %s, IP *, FQDN %s", idStr, fqdnStr)
+		db.DeleteRecords(idStr, fqdn, net.IP{})
+	})
+
+	muxRouter.Methods("DELETE").Path("/name").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fqdnStr, fqdn := extractFQDN(r)
+
+		Info.Printf("[http] Deleting ID *, IP *, FQDN %s", fqdnStr)
+		db.DeleteRecords("", fqdn, net.IP{})
 	})
 
 	http.Handle("/", muxRouter)
--- a/nameserver/mocks_test.go
+++ b/nameserver/mocks_test.go
@@ -83,6 +83,10 @@ func (mz *mockedZoneWithRecords) AddRecord(ident string, name string, ip net.IP)
 	notImplWarn()
 	return nil
 }
+func (mz *mockedZoneWithRecords) DeleteRecords(ident string, name string, ip net.IP) int {
+	notImplWarn()
+	return 0
+}
 func (mz *mockedZoneWithRecords) DeleteRecord(ident string, ip net.IP) error {
 	notImplWarn()
 	return nil
--- a/nameserver/mocks_test.go
+++ b/nameserver/mocks_test.go
@@ -87,12 +87,7 @@ func (mz *mockedZoneWithRecords) DeleteRecords(ident string, name string, ip net
 	notImplWarn()
 	return 0
 }
-func (mz *mockedZoneWithRecords) DeleteRecord(ident string, ip net.IP) error {
-	notImplWarn()
-	return nil
-}
-func (mz *mockedZoneWithRecords) DeleteRecordsFor(ident string) error { notImplWarn(); return nil }
-func (mz *mockedZoneWithRecords) Status() string                      { notImplWarn(); return "nothing" }
+func (mz *mockedZoneWithRecords) Status() string { notImplWarn(); return "nothing" }
 func (mz *mockedZoneWithRecords) ObserveName(name string, observer ZoneRecordObserver) error {
 	notImplWarn()
 	return nil
--- a/nameserver/zone.go
+++ b/nameserver/zone.go
@@ -49,6 +49,8 @@ type Zone interface {
 	Domain() string
 	// Add a record in the local database
 	AddRecord(ident string, name string, ip net.IP) error
+	// Delete matching records (uninitialised values act as wildcards)
+	DeleteRecords(ident string, name string, ip net.IP) (count int)
 	// Delete a record in the local database
 	DeleteRecord(ident string, ip net.IP) error
 	// Delete all records for an ident in the local database
@@ -590,6 +592,47 @@ func (zone *ZoneDb) AddRecord(ident string, name string, ip net.IP) (err error) 
 	record := Record{dns.Fqdn(name), ip, 0, 0, 0}
 	_, err = zone.getNamesSet(ident).addIPToName(record, zone.clock.Now())
 	return
+}
+
+// Delete matching records (uninitialised values act as wildcards)
+func (zone *ZoneDb) DeleteRecords(ident string, name string, ip net.IP) int {
+	zone.mx.Lock()
+	defer zone.mx.Unlock()
+
+	count := 0
+
+	if name != "" {
+		name = dns.Fqdn(name)
+	}
+
+	for identK, identV := range zone.idents {
+		if ident == "" || ident == identK {
+			for nameK, nameV := range identV.names {
+				nameModified := false
+				if name == "" || name == nameK {
+					for ipv4K, ipv4V := range nameV.ipv4 {
+						if ip.Equal(net.IP{}) || ip.Equal(ipv4K.toNetIP()) {
+							ipv4V.notifyIPObservers()
+							delete(nameV.ipv4, ipv4K)
+							nameModified = true
+							count++
+						}
+					}
+				}
+				if nameModified {
+					nameV.notifyNameObservers()
+					if nameV.empty() {
+						delete(identV.names, nameK)
+					}
+				}
+			}
+		}
+		if identV.empty() {
+			delete(zone.idents, identK)
+		}
+	}
+
+	return count
 }
 
 // Delete all records for an ident
--- a/nameserver/zone.go
+++ b/nameserver/zone.go
@@ -192,15 +192,27 @@ func (n *name) addIP(zr ZoneRecord, now time.Time) (*recordEntry, error) {
 	return ne, nil
 }
 
-// Delete an IP for this name
-func (n *name) deleteIP(ip IPv4) bool {
-	if ipRecord, found := n.ipv4[ip]; found {
-		ipRecord.notifyIPObservers()
-		n.notifyNameObservers()
-		delete(n.ipv4, ip)
-		return true
+// Delete by IP. Pass uninitialized value for wildcard
+func (n *name) deleteIP(ip net.IP) int {
+	count := 0
+	if !ip.Equal(net.IP{}) {
+		ipv4 := ipToIPv4(ip)
+		if ipv4V, found := n.ipv4[ipv4]; found {
+			ipv4V.notifyIPObservers()
+			delete(n.ipv4, ipv4)
+			count++
+		}
+	} else {
+		for ipv4K, ipv4V := range n.ipv4 {
+			ipv4V.notifyIPObservers()
+			delete(n.ipv4, ipv4K)
+			count++
+		}
 	}
-	return false
+	if count > 0 {
+		n.notifyNameObservers()
+	}
+	return count
 }
 
 // Update the list of IPs for this name, adding new ones and removing old IPs...
@@ -330,36 +342,25 @@ func (ns *namesSet) addIPToName(zr ZoneRecord, now time.Time) (*recordEntry, err
 	return n.addIP(zr, now)
 }
 
-// Delete a name in the names set
-func (ns *namesSet) deleteName(n string) error {
-	if name, found := ns.names[n]; found {
-		for _, ipRecord := range name.getAllEntries() {
-			ipRecord.notifyIPObservers()
+// Delete by name and IP. Pass uninitialized values for wildcard
+func (ns namesSet) deleteNameIP(name string, ip net.IP) int {
+	count := 0
+	if name != "" {
+		if nameV, found := ns.names[name]; found {
+			count += nameV.deleteIP(ip)
+			if nameV.empty() {
+				delete(ns.names, name)
+			}
 		}
-		name.notifyNameObservers()
-		delete(ns.names, n)
-		return nil
-	}
-	return LookupError(n)
-}
-
-// Delete an IPv4 from all names
-func (ns *namesSet) deleteIP(ip IPv4) error {
-	cnt := 0
-	for nkey, n := range ns.names {
-		if n.deleteIP(ip) {
-			cnt++
-		}
-		if n.empty() {
-			delete(ns.names, nkey)
+	} else {
+		for nameK, nameV := range ns.names {
+			count += nameV.deleteIP(ip)
+			if nameV.empty() {
+				delete(ns.names, nameK)
+			}
 		}
 	}
-
-	// we must return an error if no records have been found and deleted
-	if cnt == 0 {
-		return LookupError(fmt.Sprintf("%+v", ip))
-	}
-	return nil
+	return count
 }
 
 // Get the name query time
@@ -592,43 +593,14 @@ func (zone *ZoneDb) AddRecord(ident string, name string, ip net.IP) (err error) 
 
 // Delete matching records (uninitialised values act as wildcards)
 func (zone *ZoneDb) DeleteRecords(ident string, name string, ip net.IP) int {
-	zone.mx.Lock()
-	defer zone.mx.Unlock()
-
-	count := 0
-
 	if name != "" {
 		name = dns.Fqdn(name)
 	}
 
-	for identK, identV := range zone.idents {
-		if ident == "" || ident == identK {
-			for nameK, nameV := range identV.names {
-				nameModified := false
-				if name == "" || name == nameK {
-					for ipv4K, ipv4V := range nameV.ipv4 {
-						if ip.Equal(net.IP{}) || ip.Equal(ipv4K.toNetIP()) {
-							ipv4V.notifyIPObservers()
-							delete(nameV.ipv4, ipv4K)
-							nameModified = true
-							count++
-						}
-					}
-				}
-				if nameModified {
-					nameV.notifyNameObservers()
-					if nameV.empty() {
-						delete(identV.names, nameK)
-					}
-				}
-			}
-		}
-		if identV.empty() {
-			delete(zone.idents, identK)
-		}
-	}
+	zone.mx.Lock()
+	defer zone.mx.Unlock()
 
-	return count
+	return zone.deleteIdentNameIP(ident, name, ip)
 }
 
 // Observe a name.
@@ -776,16 +748,23 @@ func (zone *ZoneDb) getNamesSet(ident string) *namesSet {
 	return ns
 }
 
-// Delete a ident
-func (zone *ZoneDb) deleteIdent(ident string) {
-	Info.Printf("[zonedb] Removing everything for ident '%s'", ident)
-	if ns, found := zone.idents[ident]; found {
-		for _, n := range ns.names {
-			for _, entry := range n.getAllEntries() {
-				entry.notifyIPObservers()
+// Delete by ident, name and IP. Pass uninitialized value for wildcard
+func (zone *ZoneDb) deleteIdentNameIP(ident string, name string, ip net.IP) int {
+	count := 0
+	if ident != "" {
+		if identV, found := zone.idents[ident]; found {
+			count += identV.deleteNameIP(name, ip)
+			if identV.empty() {
+				delete(zone.idents, ident)
 			}
-			n.notifyNameObservers()
 		}
-		delete(zone.idents, ident)
+	} else {
+		for identK, identV := range zone.idents {
+			count += identV.deleteNameIP(name, ip)
+			if identV.empty() {
+				delete(zone.idents, identK)
+			}
+		}
 	}
+	return count
 }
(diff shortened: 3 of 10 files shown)


awh:
@inercia @rade ready for initial review - thankyou

test/240_dns_add_name_test.sh
```diff
@@ -19,11 +20,17 @@ weave_on $HOST1 dns-add $C2 c2 -h $NAME2
...
 
-assert "exec_on $HOST1 c1 getent hosts $C1 | tr -s ' '" "$C1 $NAME1"
+assert "exec_on $HOST1 c1 getent hosts $NAME1 | tr -s ' '" "$C1 $NAME1"
+assert "exec_on $HOST1 c1 getent hosts $NAME3 | tr -s ' '" "$C1 $NAME3"
 
-weave_on $HOST1 dns-remove $C1 c1
+weave_on $HOST1 dns-remove $C1 c1 -h $NAME1
+
+whitely echo "Waiting 30s for cache expiry..."
+sleep 30
```
rade: #711

weave
```diff
@@ -584,34 +584,68 @@ dns_args() {
...
-        return
+# Iff the container in $1 has an FQDN, invoke $2 as a command passing
+# the FQDN as the first argument and $3.. as additional arguments
+if_has_fqdn() {
+    CONTAINER="$1"
+    COMMAND="$2"
+    shift 2
+
+    CONTAINER_HOSTNAME=$(docker inspect --format='{{.Config.Hostname}}' $CONTAINER 2>/dev/null) || true
+    CONTAINER_DOMAIN=$(docker inspect --format='{{.Config.Domainname}}' $CONTAINER 2>/dev/null) || true
```
rade: docker interactions are expensive, so I'd do this in one call and then split the result.

weave
```diff
@@ -961,19 +987,25 @@ case "$COMMAND" in
         [ $# -ge 1 ] || usage
         CONTAINER_ID="$1"
         if [ $# -eq 1 ]; then
-            tell_dns PUT $CONTAINER_ID $IP_ARGS
+            if_dns_running with_full_container_id $CONTAINER_ID if_has_fqdn put_dns_fqdn $IP_ARGS
         else
             [ $# -eq 3 -a "$2" = "-h" ] || usage
             FQDN="$3"
-            tell_dns_fqdn PUT $CONTAINER_ID $FQDN $IP_ARGS
+            with_full_container_id $CONTAINER_ID put_dns_fqdn $FQDN $IP_ARGS
```
rade: are you missing an `if_dns_running` here?

weave
```diff
@@ -584,34 +584,68 @@ dns_args() {
...
-# domain, at addresses $3, $4 ...
-tell_dns() {
-    METHOD="$1"
-    CONTAINER_ID="$2"
-    shift 2
-    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER_ID 2>/dev/null) || true
-    tell_dns_fqdn $METHOD $CONTAINER_ID $CONTAINER_FQDN $@
+# Execute $@ only if the weaveDNS container is running
+if_dns_running() {
+    if status=$(docker inspect --format='{{.State.Running}}' $DNS_CONTAINER_NAME 2>/dev/null) || [ "$status" != "true" ] ; then
```
rade: This conditional is wrong. The original had a `!` at the beginning and was inverted, i.e. the whole expression evaluated to true when the container *wasn't* running. You didn't apply DeMorgan's laws correctly when negating the expression.

nameserver/zone.go
```diff
@@ -121,6 +122,37 @@ func (zone *ZoneDb) AddRecord(ident string, name string, ip net.IP) error {
...
+			w++
+		}
+	}
+
+	count = len(zone.recs) - w
+	zone.recs = zone.recs[:w]
+
+	return
+}
+
```
rade: Surely `DeleteRecord` and `DeleteRecordsFor` should be implemented in terms of the above. Also, they are rather misnamed. I'd expect `DeleteRecord` to be inverse of `AddRecord`, but it isn't.

nameserver/zone.go
```diff
@@ -18,6 +18,7 @@ var rdnsDomainLen = len(RDNSDomain) + 1
 
 type Zone interface {
 	AddRecord(ident string, name string, ip net.IP) error
+	DeleteRecords(ident *string, ip *net.IP, name *string) (count int)
```
rade: That signature is odd. The order of args should be consistent with `AddRecord`.

rade: Also, a `nil for wildcard` comment would help.

nameserver/mocks_test.go
```diff
@@ -75,6 +75,10 @@ func (mz *mockedZoneWithRecords) AddRecord(ident string, name string, ip net.IP)
 	notImplWarn()
 	return nil
 }
+func (mz *mockedZoneWithRecords) DeleteRecords(ident *string, ip *net.IP, name *string) int {
```
inercia: There is something I don't like: `DeleteRecords` takes **pointers** to strings/IPs, while `DeleteRecord` does not. I understand it is for passing `nil` as a _wildcard argument_, but maybe we could use the empty values for that (eg, `ip=net.IP{}`)...

nameserver/zone.go
inercia: ... and what about returning an `error` (in addition to the `count`)? All the other `Delete*` methods have it...

nameserver/mocks_test.go
rade: I agree; using the initial value (i.e. what you'd get from `var x string` and `var x net.IP`) as a wildcard would be better, I think.

nameserver/zone.go
rade: `error` returns don't make sense here, imo.

nameserver/zone.go
awh: I did make a comment on the commit to this effect - I wasn't sure if we should do that, or remove the other two methods altogether

nameserver/zone.go
inercia: It is for consistency with the other methods `Delete*` methods: they return a `LookupError` if no records are deleted...

awh: 1) It's not clear to me that we should retain those other methods at all - we could make all clients use `DeleteRecords`
2) I'm not sure what the error could be? Failure to delete something at all?

nameserver/zone.go
rade: I'd prefer to remove the other two, but only if that doesn't introduce clutter at the call sites.

nameserver/mocks_test.go
awh: Or perhaps those other methods should go away entirely, since their functionality is subsumed by `DeleteRecords`? And if you don't like points, how about `DeleteRecords` takes a predicate function?

nameserver/zone.go
inercia: I think we should leave only one `DeleteRecords` as we are close to reach the maximum number of permutations in arguments. I think we should have a `DeleteRecords` where `ident`, `name` or `ip` could be wildcarded only in the following forms:

ident, *, *
ident, name, *
ident, name, ip

nameserver/zone.go
inercia: Yes, I'm sure there is a historic reason for returning the error... maybe it is time to get rid of it...

rade: I'd argue that the error return on the other methods is wrong too. The only method were it makes sense would be one that takes all three params, non-wildcarded. i.e. the proper inverse of `AddRecord`.

nameserver/mocks_test.go
awh: Also I feel the complete opposite about the nil values - it makes it clear that I am not specifying any value at all, rather than a value which is treated specially

rade: ...whereas I am looking at the signature and wonder whether that method modifies the args I am passing in.

nameserver/zone.go
awh: What do you think of `DeleteRecords` taking a predicate function?

inercia: I think it would be a good idea to use that predicate for the internal implementation, but not for the API. I also I think that we will not really need that level of flexibility. We will probably need:

* to remove all records for an ident (ident, *, *) (eg, a container has been removed)
* to remove all records for a name in a ident (ident, name, *)
* to remove a particular record (ident, name, ip)

Can you guys think of any other cases?

awh: Good point. @rade also points out that it forces a traversal of the underlying data structure, which would prevent optimisation.

In terms of other use cases - remove all records (*, *, *) perhaps? That being said, actively *preventing* combinations which we don't consider to be useful involves a bunch of extra checks - do we want to do that and return an error?

inercia: Maybe these use cases are not easy to implement now, but they can be optimally implemented when #615 is merged, as the underlying data structure would be a tree...

Maybe the `(*, *, *)` case makes sense... Or maybe it should be explicitly done in a `Flush()` operation or something similar...

And, regarding returning error codes, maybe we could `panic()` if you find a *no-wildcard* after processing a *wildcard*...

awh:
@inercia @bboreham ready for another review pass - thankyou

nameserver/zone.go
```diff
@@ -121,33 +121,37 @@ func (zone *ZoneDb) AddRecord(ident string, name string, ip net.IP) error {
...
-		return nil
+
+	if name != "" {
+		name = dns.Fqdn(name)
+	}
+
+	match := func(r dbRecord) bool {
+		return (ident == "" || r.Ident == ident) &&
+			(name == "" || r.Name == name) &&
+			(ip.Equal(net.IP{}) || r.IP.Equal(ip))
```
inercia: Just a suggestion, but we could check for the `(ident, *, *)`, `(ident, name, *)`, `(ident, name, ip)` cases with something like:

```code:Go
match := func(r dbRecord) bool {
return (ident == "" || (r.Ident == ident && (name == "" || (r.Name == name && (ip.IsUnspecified() || r.IP.Equal(ip))))))
}
```

nameserver/http.go
```diff
@@ -16,6 +16,15 @@ func httpErrorAndLog(level *log.Logger, w http.ResponseWriter, msg string,
 	level.Printf("[http] "+logmsg, logargs...)
 }
 
+func extractFQDN(r *http.Request) (fqdnStr string, fqdn string) {
+	fqdnStr = r.FormValue("fqdn")
+	if fqdnStr == "" {
+		return "*", ""
+	} else {
+		return fqdnStr, fqdnStr
+	}
```
rade:
````
if fqdnStr := r.FormValue("fqdn"); fqdnStr != "" {
	return fqdnStr, fqdnStr
}
return "*", ""
````

File: nameserver/http.go [hunk 10]
```diff
@@ -16,6 +16,15 @@ func httpErrorAndLog(level *log.Logger, w http.ResponseWriter, msg string,
 	level.Printf("[http] "+logmsg, logargs...)
 }
 
+func extractFQDN(r *http.Request) (fqdnStr string, fqdn string) {
```
rade: don't name the return values

test/240_dns_add_name_test.sh
```diff
@@ -19,11 +20,17 @@ weave_on $HOST1 dns-add $C2 c2 -h $NAME2
...
 
 assert_raises "exec_on $HOST1 c1 getent hosts $NAME1" 2
+assert "exec_on $HOST1 c1 getent hosts $NAME3 | tr -s ' '" "$C1 $NAME3"
```
rade: This repetition is getting tiresome. create a helper.

Also, we should be testing reverse lookups.

rade:
I am not sold on the URL scheme. Can we come up with something more consistent and versatile? How about:

`PUT /name/<fqdn>/ip/<ip>/id/<containerId>`

?

We could, at some point, permit leaving out the `id`, which is why I put it last.

And for delete we can leave out various bits,  e.g.

`DELETE /name/<fqdn>/id/<containerId>`
`DELETE /id/<containerId>`

test/240_dns_add_name_test.sh
rade: Also, elswhere, i.e. in #615, we `weave launch-dns -no-cache`, to disable the cache. Could we do this here, for the entire test suite, or would it adversely affect what we are testing?

inercia:
Shouldn't it be better to move to optional arguments? For example, `DELETE /name?fqdn=something.com&ip=10.10.0.1`

test/240_dns_add_name_test.sh
awh: I mentioned this possibility in my initial comments - I think we can change it to use -no-cache once #615 is merged...

rade:
I thought URL params were frowned upon for RESTful APIs.

re deferring... I disagree. The existing URL scheme gets called into question because of this issue. and we are "making stuff up" for DELETE here.

inercia:
But changing the URL scheme would also affect other parts of the HTTP API
(eg, `AddRecord`, so we would have to change the `tell_dns` in the weave
script for adding records, unit tests for adding records, etc...), so this
PR would not be limited to the _"delete a record"_ functionality anymore...

rade:
So the existing URL scheme (after this PR) is
````
PUT    /name/<containerId>/<ip>?fqdn=<fqdn>
DELETE /name/<containerId>/<ip>[?fqdn=<fqdn>]
DELETE /name/<containerId>[?fqdn=<fqdn>]
DELETE /name?[fqdn=<fqdn>]
````
(modulo the fact that for `PUT` the fqdn is transmitted in the body)

Correct?

That's consistent actually. Just not the most sensible scheme. Let's clean it up later.

File: test/240_dns_add_name_test.sh (hunk 11 above)
Comment (awh) [inline test/240_dns_add_name_test.sh]: In my experience reverse lookups of an address return a single canonical hostname, even when multiple hostnames point to that address. Are we returning them all? Might we break some tools (e.g. getent) if we violate this norm? Thoughts @inercia?

File: weave (hunk 3 above)
Comment (awh) [inline weave]: I was, thanks

File: weave (hunk 4 above)
Comment (awh) [inline weave]: D'oh! Good catch, thankyou

**Comment by awh** [comment]
@inercia @rade hopefully this is in a mergeable state now, but we'll hold off until #615 is in master...

**Comment by rade** [comment]
> we'll hold off until #615 is in master

why?

**Comment by rade** [comment]
needs rebasing and heed https://github.com/weaveworks/weave/commit/901f0bf04ead1b0e305e5673ac9044c51983d5b3 and https://github.com/weaveworks/weave/commit/dd33c9418875c0de4bfaf0668d48aaa06530ce9b

**Comment by rade** [comment]
This is exploding rather badly when the container isn't there...
````
$ weave dns-add 10.2.0.1 foo
/home/weave/weave: line 1100: 10.2.0.1: not found
````

File: weave [hunk 12]
```diff
@@ -600,34 +600,67 @@ dns_args() {
...
-    METHOD="$1"
-    CONTAINER_ID="$2"
-    CONTAINER_FQDN="$3"
-    shift 3
-    if ! status=$(docker inspect --format='{{.State.Running}}' $DNS_CONTAINER_NAME 2>/dev/null) || [ "$status" != "true" ] ; then
-        # weavedns not running - silently return
-        return
+# Iff the container in $1 has an FQDN, invoke $2 as a command passing
+# the FQDN as the first argument and $3.. as additional arguments
+if_has_fqdn() {
```
rade: The docs don't match the functionality - $2 is invoked with $1, the fqdn, and $3.....

Also, the `if_` prefix breaks emacs indentation. I suggest renaming to `with_fqdn`.

weave
```diff
@@ -600,34 +600,67 @@ dns_args() {
...
+# Convert the container name or short ID in $2 to its full form, then
+# invoke $1 as a command passing the full form as the first argument and
+# $3.. as additional arguments
+with_full_container_id() {
+    CONTAINER="$1"
+    COMMAND="$2"
+    shift 2
+
+    $COMMAND $(docker inspect --format='{{.Id}}' $CONTAINER 2>/dev/null) "$@"
+}
```
rade: I think it's overkill making this a closure. Just make it spit out the full id and invoke it with `$(container_id $CONTAINER_ID)`.

Throwing away stderr is a bad idea. When the container isn't there that contains a nice descriptive error message.

rade:
`get_ipam_cidr` should call the new `container_id` function, instead of replicating the functionality.

weave
```diff
@@ -600,34 +600,67 @@ dns_args() {
...
-# with an FQDN obtained from the container's configured hostname and
-# domain, at addresses $3, $4 ...
-tell_dns() {
-    METHOD="$1"
-    CONTAINER_ID="$2"
-    shift 2
-    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER_ID 2>/dev/null) || true
-    tell_dns_fqdn $METHOD $CONTAINER_ID $CONTAINER_FQDN $@
+# Execute $@ only if the weaveDNS container is running
+if_dns_running() {
```
rade: The `if_` prefix trips up emacs indentation. `when_dns_is_running`?

rade:
It seems rather dubious for `add/remove-dns` to succeed when dns isn't running. Yes, it was doing that previously. But it's still wrong.

weave
rade: Also, the arg order is the reverse of what is documented.

nameserver/zone.go
```diff
@@ -121,33 +121,37 @@ func (zone *ZoneDb) AddRecord(ident string, name string, ip net.IP) error {
 	return nil
 }
 
-func (zone *ZoneDb) DeleteRecord(ident string, ip net.IP) error {
+func (zone *ZoneDb) DeleteRecords(ident string, name string, ip net.IP) (count int) {
```
rade: don't use a named return value here. It saves you typing five characters later, which isn't worth the obscurity.

test/config.sh
```diff
@@ -137,3 +147,4 @@ end_suite() {
 
 WEAVE=../weave
 DOCKER_NS=./bin/docker-ns
+
```
rade: stray

test/config.sh
```diff
@@ -115,12 +115,22 @@ container_ip() {
...
+# assert_dns_ptr_record <host> <container> <name> <ip>
+assert_dns_ptr_record() {
     assert "exec_on $1 $2 getent hosts $4 | tr -s ' '" "$4 $3"
 }
 
+# assert_dns_record <host> <container> <name> <ip>
+assert_dns_record() {
+    assert_dns_a_record $1 $2 $3 $4
+    assert_dns_ptr_record $1 $2 $3 $4
```
rade: this offends my vertical alignment sensibilities.

weave
awh: It eliminates a call to `docker inspect` in the event that weaveDNS isn't running; still want it removed?

awh:
> It seems rather dubious for add/remove-dns to succeed when dns isn't running.

What should it do instead?

> Yes, it was doing that previously. But it's still wrong.

Do you want it addressed in this PR?

nameserver/zone.go
awh: OK as discussed will leave it in the interface declaration but remove it from the function definition.

weave
awh: As discussed, tried alternative using shortcircuit evaluation in combination with `$(container_id ...)` but this is incompatible with failing gracefully when an unknown container is passed in. It also means all the statements gain an unsightly `|| true` at the end of the line to prevent failure of the overall command if weaveDNS isn't running. Consequently, am leaving as is an augmenting `with_full_container_id` to exit if the container doesn't exist.

nameserver/zone.go
```diff
@@ -585,30 +583,45 @@ func (zone *ZoneDb) AddRecord(ident string, name string, ip net.IP) (err error)
...
+					}
+				}
+				if nameModified {
+					nameV.notifyNameObservers()
+					if nameV.empty() {
+						delete(identV.names, nameK)
+					}
+				}
+			}
+		}
```
rade: Given that we now do have map data structures here, we should do better than a full scan when ident and/or name and/or IP are supplied. Can't be that much more code.

rade:
> `get_ipam_cidr` should call the new `container_id` function, instead of replicating the functionality.

Still to do, in some form or another; we can't have the same logic in two places.

>  It seems rather dubious for `add/remove-dns` to succeed when dns isn't running. Yes, it was doing that previously. But it's still wrong.

Still to do, or file an issue for it.

awh:
>> get_ipam_cidr should call the new container_id function, instead of replicating the functionality.
> Still to do, in some form or another; we can't have the same logic in two places.

Agreed; I left it until you decide whether you're happy with the higher order function approach (as an aside, the logic is already duplicated in master...)

>> It seems rather dubious for add/remove-dns to succeed when dns isn't running. Yes, it was doing that previously. But it's still wrong.
> Still to do, or file an issue for it.

Indeed - issue is still assigned to me for this reason - push was for backup purposes

nameserver/zone.go
awh: I considered it, but given you have to iterate over all the IPs anyway when you remove a name to notify observers, I decided against. I've a feeling that it'll be a bit messy because you'll end up pulling the innards of the name and IP loops out into separate functions to avoid duplication between the direct lookup/full scan paths - I'll have a play and see what it looks like.

nameserver/zone.go
```diff
@@ -585,30 +583,45 @@ func (zone *ZoneDb) AddRecord(ident string, name string, ip net.IP) (err error)
...
+			for nameK, nameV := range identV.names {
+				nameModified := false
+				if name == "" || name == nameK {
+					for ipv4K, ipv4V := range nameV.ipv4 {
+						if ip.Equal(net.IP{}) || ip.Equal(ipv4K.toNetIP()) {
+							ipv4V.notifyIPObservers()
+							delete(nameV.ipv4, ipv4K)
+							nameModified = true
+							count++
+						}
```
inercia: I was wondering if we could use methods like `zone.go/nameSet.deleteName()/deleteIP()`, `zone.go/name.deleteIP()`, etc. For example, maybe you could move the `ip.Equal(net.IP{})` comparison out of the loop and just `if ip.Equal(net.IP{}) { identV.name.deleteName(name) } else { nameV.deleteIP(ipv4K) }`...

nameserver/zone.go
awh: > Can't be that much more code.

I've had a play, and it turns out it can - I'm going to leave it as it is, unless you object strongly...

nameserver/zone.go
awh: Actually I was going to suggest that all those methods could go away. If not, to be used as you suggest, they all need to be modified to return a count of what has been removed...

awh: Hmm I see those methods are used from elsewhere...

awh:
@rade @inercia I've pushed an update that addresses the outstanding `weave` script issues - let me know what you think. I've left the PR assigned to myself whilst I have another go on optimising `DeleteRecords` along the lines suggested...

weave
```diff
@@ -600,34 +620,84 @@ dns_args() {
...
+# the container as the first argument, the FQDN as the second argument
+# and $3.. as additional arguments
+when_has_fqdn() {
+    CONTAINER="$1"
+    COMMAND="$2"
     shift 2
-    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER_ID 2>/dev/null) || true
-    tell_dns_fqdn $METHOD $CONTAINER_ID $CONTAINER_FQDN $@
+
+    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER 2>/dev/null) || true
```
rade: `|| return 0` ?

weave
```diff
@@ -950,7 +1001,7 @@ case "$COMMAND" in
         CONTAINER=$(docker start $CONTAINER_ID)
         get_ipam_cidr $CONTAINER_ID $CONTAINER
         with_container_netns $CONTAINER_ID attach $CIDR_ARGS >/dev/null
-        tell_dns PUT $CONTAINER_ID $CIDR_ARGS
+        when_dns_running when_has_fqdn $CONTAINER put_dns_fqdn $CIDR_ARGS
```
rade: The `with_container_netns` call ends up setting `$CONTAINER` to `$CONTAINER_ID`, so this is not doing what you want.

Simple fix is to call `with_container_netns` with `$CONTAINER` instead of `$CONTAINER_ID`.

rade: [Fixed on master](https://github.com/weaveworks/weave/commit/6f57bcb47701b3fd0d8c62fa12f9ecc199457f0c)

weave
```diff
@@ -512,6 +512,26 @@ container_ip() {
...
+        true*)
+            echo "${STATUS#true }"
+            ;;
+        *)
+            echo "$1 container is not running." >&2
+            return 2
+            ;;
+    esac
+}
+
```
rade: I have [just](https://github.com/weaveworks/weave/commit/79eb4c298c004db7c637fd6a50216bd7224cfcf2) put this on master, but called it `container_id`, and changed `get_ipam_cidr` to only take one argument - the full container id. There was only one call site where we didn't have that, and now use the helper to find it.

awh: I think there's a nice pattern in this function that we could apply elsewhere (in `container_ip` at least) - if the function succeeds the result is printed to stdout (rather than being assigned to a global variable), otherwise it prints an error to stderr and `return`s an indicative exit status (rather than `exit`ing). This seems to give maximum flexibility to the caller:

* You can suppress the stderr messages if so desired by using `2>/dev/null`
* If such a function is invoked directly with `set -e` in operation, it will cause the script to exit
* You can call it and use its result conditionally with the `if RESULT=$(...); then` pattern
* You've got access to explicit failure reason through `$?`
* Doesn't rely on global variables

awh:
@rade @inercia added another commit that integrates @rade's counting wildcard deletion helpers.

inercia:
LGTM

rade:
I've tweaked the weave script for this somewhat in [my branch](https://github.com/rade/weave/tree/dns_delete_individual_records)

awh:
Rebased on the cache test PR.

awh:
> I've tweaked the weave script for this somewhat in my branch

Commits cherry-picked.

Issue #723, opened 2015-05-21
#704 may have broken "weave as docker bridge" usage


rade:
We know some folks configure docker to use the weave bridge. With the change from #704 I foresee the following problems:

1. The user runs `weave --local create-bridge` with `DOCKER_BRIDGE=weave`, causing us to add a rule that `DROP`s traffic across the `weave` bridge.... followed by the normal rule to `ACCEPT` it. The former makes the latter non-functional. We could fix that by reversing the order, or suppressing the `DROP` rule when the docker bridge is the same as the weave bridge. The latter is cleaner.
2. In reality the user will probably run `weave --local create-bridge` w/o setting `DOCKER_BRIDGE`, since hitherto that wasn't a requirement. Nothing especially bad happens as a result. My testing indicates that it is ok to add filter rules for non-existing interfaces. So the only downside is that we end up with some useless rules. I can't think of a way of avoiding that, and it's probably not worth the hassle in any case.

bboreham:
I did think about this, and tested (2).  I didn't think anyone would run `weave create-bridge` with `DOCKER_BRIDGE=weave`, but I guess it's plausible.

Note that #704 introduces the requirement that if you run `weave` with `DOCKER_BRIDGE` set at the point where the weave bridge is created then you should also run `weave reset` with the same setting. Since `DOCKER_BRIDGE` is only mentioned in the weaveDNS docs I have commented there.

PR #724, opened 2015-05-21
Check if Docker using weave bridge before inserting DROP rule

bboreham:
Addresses #723.
Also comment the use of `-I`, as noted on c5d5dd5fdc07239001cd853a9137091e6406d4c1


rade:
I think we ought to have the same logic in `destroy_bridge`, for symmetry, even though it's not strictly necessary.

rade:
> I think we ought to have the same logic in `destroy_bridge`

Did that myself in https://github.com/weaveworks/weave/commit/5148ed793660750cca630baba6030e983587e4fc.

PR #728, opened 2015-05-22
allow test/run_all.sh to run specific test files

paulbellamy:
`./run_all.sh 110_encryption_test.sh`

Handles the assert_* calls better than running the test bare. If people like/use this, it would be nice to rename `run_all.sh` to something more appropriate.


paulbellamy:
Is this related to what you were talking about, @awh ?

awh:
> Is this related to what you were talking about, @awh ?

Yep. Needs rebase...

PR #737, opened 2015-05-26
WeaveDNS cache tests

inercia:
Improved unit tests for the cache:
* cache invalidations as a result of `AddRecord()`/`DeleteRecord()` calls
* cache expirations

### DIFF
--- a/nameserver/server_cache_test.go
+++ b/nameserver/server_cache_test.go
@@ -11,6 +11,160 @@ import (
 	wt "github.com/weaveworks/weave/testing"
 )
 
+// Check that AddRecord/DeleteRecord/... in the Zone database lead to cache invalidations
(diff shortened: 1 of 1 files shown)


nameserver/server_cache_test.go
```diff
@@ -11,6 +11,247 @@ import (
...
+		Zone:              zone,
+		Cache:             cache,
+		Clock:             clk,
+		ListenReadTimeout: testSocketTimeout,
+		UpstreamCfg:       fallback.CliConfig,
+		MaxAnswers:        4,
+	})
+	wt.AssertNoErr(t, err)
+	defer srv.Stop()
+	go srv.Start()
```
awh: It doesn't make a huge amount of difference, but `go ...Start()` appears before `defer ...Stop()` everywhere else...

Issue #745, opened 2015-05-27
don't hash gossip channel names


rade:
Our gossip implementation currently uses a 32-bit hash of the channel name as the key to distinguish channels. We do this so we a) get a short, fixed-length field for transmission, and b) don't require a "channel numbering authority". The downside is that it makes debugging harder. This is going to become more of an issue if we permit routers with different channels to establish connections - something we are currently avoiding with protocol versioning.

Options:

1. give up on the "short, fixed-length" requirement and transmit the name instead
2. give up on the "no channel numbering authority" requirement and assign unique numbers to channels
3. expand the protocol to negotiate channel name <-> number mappings between peers

Option 1+2 are easy. Option 3 is quite fiddly.

I suggest we go with option 1; the extra few bytes and associated encoding/decoding cost, and the cost of a string-based instead of number-based hash lookup, is not going to have a noticeable performance impact.

dpw:
I have an implementation of option 1 sitting around.

rade:
great. can you extract it into a PR, please?

PR #748, opened 2015-05-27
Make gossip use channel names on the wire, not channel name hashes

dpw:
Addresses #745.

### DIFF
--- a/router/gossip.go
+++ b/router/gossip.go
@@ -85,23 +85,22 @@ type GossipChannel struct {
 	ourself      *LocalPeer
 	routes       *Routes
 	name         string
-	hash         uint32
 	gossiper     Gossiper
 	senders      connectionSenders
 	broadcasters peerSenders
 }
 
+type GossipChannels map[string]*GossipChannel
+
 func (router *Router) NewGossip(channelName string, g Gossiper) Gossip {
-	channelHash := hash(channelName)
 	channel := &GossipChannel{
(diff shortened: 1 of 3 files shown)


rade:
protocol version needs bumping.

router/gossip.go
```diff
@@ -85,23 +85,23 @@ type GossipChannel struct {
 	ourself      *LocalPeer
 	routes       *Routes
 	name         string
-	hash         uint32
 	gossiper     Gossiper
 	senders      connectionSenders
 	broadcasters peerSenders
 }
 
+// Type of GossipChannels in the router
```
rade: that comment seems superfluous

router/gossip.go
```diff
@@ -123,13 +123,13 @@ func (router *Router) SendAllGossipDown(conn Connection) {
...
-	if err := decoder.Decode(&channelHash); err != nil {
+	var channelName string
+	if err := decoder.Decode(&channelName); err != nil {
 		return err
 	}
-	channel, found := router.GossipChannels[channelHash]
+	channel, found := router.GossipChannels[channelName]
 	if !found {
-		return fmt.Errorf("[gossip] received unknown channel with hash %v", channelHash)
+		return fmt.Errorf("[gossip] received unknown channel with name %v", channelName)
```
rade: `'%s'` would be better, I think.

dpw:
Re-spun with comments addressed.

Issue #660, opened 2015-05-14
add ConnectionMaker status to status-json


rade:
Unlike the ordinary status output, the JSON flavour does not contain any information about the connection maker, e.g. currently pending connection requests.

See also #659.

PR #764, opened 2015-05-27
Add "Direct Peers" to weave status

### LINKED ISSUE #659: add command line peers to status output
Since command line peers can be modified with `weave connect/forget`, the `weave status` output should include the current list.

paulbellamy:
Also added connection maker status ("DirectPeers", and "Reconnects") to json status output.

Closes #659 and #660

Text Output is:
```
Direct Peers:
192.168.48.11
Reconnects:
->[192.168.48.11:6783] trying since 2015-05-27 15:03:51.389284634 +0000 UTC
```

JSON Output is:
```json
...
  "DirectPeers" : [
      "192.168.48.11"
   ],
   "Reconnects" : {
      "192.168.48.11:6783" : {
         "TryInterval" : "4.111466676s",
         "Attempting" : true,
         "TryAfter" : "2015-05-27T15:03:51.389284634Z",
         "LastError" : "dial tcp4 192.168.48.11:6783: connection timed out"
      }
   },
...
```

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -25,10 +25,10 @@ type ConnectionMaker struct {
 
 // Information about an address where we may find a peer
 type Target struct {
-	Attempting  bool          `json:"Attempting,omitempty"`  // are we currently attempting to connect there?
-	LastError   error         `json:"LastError,omitempty"`   // reason for disconnection last time
-	TryAfter    time.Time     `json:"TryAfter,omitempty"`    // next time to try this address
-	TryInterval time.Duration `json:"TryInterval,omitempty"` // backoff time on next failure
+	attempting  bool          // are we currently attempting to connect there?
+	lastError   error         // reason for disconnection last time
+	tryAfter    time.Time     // next time to try this address
+	tryInterval time.Duration // backoff time on next failure
 }
 
 type ConnectionMakerAction func() bool
@@ -68,7 +68,7 @@ func (cm *ConnectionMaker) InitiateConnections(peers []string) []error {
 			cm.directPeers[peer] = addr
 			// curtail any existing reconnect interval
 			if target, found := cm.targets[addr.String()]; found {
-				target.TryAfter, target.TryInterval = tryImmediately()
+				target.tryAfter, target.tryInterval = tryImmediately()
 			}
 		}
 		return true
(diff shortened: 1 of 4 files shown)


router/connection_maker.go
```diff
@@ -15,31 +15,31 @@ const (
...
 // Information about an address where we may find a peer
 type Target struct {
-	attempting  bool          // are we currently attempting to connect there?
-	lastError   error         // reason for disconnection last time
-	tryAfter    time.Time     // next time to try this address
-	tryInterval time.Duration // backoff time on next failure
+	Attempting  bool          `json:"Attempting,omitempty"`  // are we currently attempting to connect there?
+	LastError   error         `json:"LastError,omitempty"`   // reason for disconnection last time
+	TryAfter    time.Time     `json:"TryAfter,omitempty"`    // next time to try this address
+	TryInterval time.Duration `json:"TryInterval,omitempty"` // backoff time on next failure
```
paulbellamy: I'm not crazy about having the json tags buried in here, but seemed cleaner than defining an identical struct and copying all the data across...

bboreham:
In the text version of `weave status` there can be an error message against each reconnect (if an error caused the disconnect) - should we have that in the JSON too?

paulbellamy:
It is tagged to be in the JSON, as well, but was nil in my example...

paulbellamy:
Although, the formatting of the "LastError" in json is not great. Let me revisit that...

router/connection_maker.go
bboreham: This is inconsistent with the way all the other JSON is done; surely if we want this in one place we should have it in all.

Personally, I prefer to keep JSON completely separate from the main logic of weave.

paulbellamy: Sure. Moved it with the rest.

test/150_connect_forget_test.sh
```diff
@@ -13,11 +17,14 @@ weave_on $HOST2 launch
 start_container $HOST1 $C1/24 --name=c1
 start_container $HOST2 $C2/24 --name=c2
 
+assert_raises "has_direct_peer $HOST2 $HOST1" 1
 assert_raises "exec_on $HOST1 c1 $PING $C2" 1
```
rade: I should have spotted this before... it's dangerous to use `assert_raises` with a non-0 status check, since the test can fail for a multitude of reasons, and not necessarily the specific condition you are interested in. Use a `!`, or, if not in a shell context, `sh -c \"! ..."`.

router/router.go
```diff
@@ -10,6 +9,8 @@ import (
 	"net"
 	"syscall"
 	"time"
+
+	"code.google.com/p/gopacket/layers"
```
rade: why?

paulbellamy: goimports does:
```
import (
  "stdlib1"
  "stdlib2"

  "non-stdlib-stuff"
)
```

PR #769, opened 2015-05-28
shorten proxy entrypoint

### LINKED ISSUE #768: [proxy] long entrypoint obscures command in `docker ps`
````
$ docker ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS             PORTS               NAMES
963b43f0d7f2        ubuntu:latest       "/home/weavewait/wea   5 minutes ago       Up 5 minutes                            admiring_nobel      
````

It's ok with `docker ps --no-trunc`.

I ran some quick experiment in shortening the entrypoint...

````
$ docker run -e WEAVE_CIDR=10.2.0.2/24 --entrypoint=/bin/sh -tdi ubuntu -c "sleep 10"
0d95a274792c694d9bbaf74e823eab27b282a48e714e19f529c02e3f44bcb283
$ docker ps
CONTAINER ID        IMAGE                         COMMAND                CREATED             STATUS              PORTS                                            NAMES
0d95a274792c        ubuntu:latest                 "/w/w /bin/sh -c 'sl   2 seconds ago       Up 1 seconds                                                         focused_bardeen      
....
````

Worth it?

I suspect most CLIs / UIs (including Scope?) will truncate output by default, since otherwise the layout looks hideous.

rade:
...so that the default output of the likes of `docker ps` shows some of the real entrypoint/command.

Closes #768.

### DIFF
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -16,7 +16,7 @@ import (
 
 var (
 	containerIDRegexp   = regexp.MustCompile("^/v[0-9\\.]*/containers/([^/]*)/.*")
-	weaveWaitEntrypoint = []string{"/home/weavewait/weavewait"}
+	weaveWaitEntrypoint = []string{"/w/w"}
 )
 
 func callWeave(args ...string) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


paulbellamy:
`/w/w` seems a bit cryptic, but I don't have a better suggestion that is as short...

PR #774, opened 2015-05-28
allow proxy to intercept unversioned (i.e. 1.0) urls

### LINKED ISSUE #770: [proxy] does not intercept unversioned API calls
The docker remote API functions can be invoked without specifying a version, but the proxy's interception regexps always require a version, and hence unversioned API calls are not intercepted, so none of the proxy magic - rewriting entrypoints, etc - is applied.

This isn't an issue for the docker CLI, afaik, since it always supplies a version. I believe programmatic docker clients tend not to though. e.g. I am pretty sure that the client created by the [go docker client example](https://github.com/fsouza/go-dockerclient#example) does not send requests with a version, since it uses `NewClient` whereas a versioned client is created with `NewVersionedClient`.

NB: I haven't tested this; it's just my reading of the code.

paulbellamy:
Closes #770

### DIFF
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	containerIDRegexp   = regexp.MustCompile("^/v[0-9\\.]*/containers/([^/]*)/.*")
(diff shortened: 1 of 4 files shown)


paulbellamy:
Need to try this a bit with curl

rade:
superseded by #777

PR #775, opened 2015-05-28
add load balancing example

### LINKED ISSUE #736: [dns][docs] load balancing example
Currently there is only a brief mentioning of load balancing. I suggest we add a simple example - just two containers on different hosts, and a third in which we run 'ping' a few times and would expect to see different IPs.

rade:
Closes #736.


inercia:
LGTM

PR #779, opened 2015-05-28
[ipam] Fix race conditions in unit tests

bboreham:
`go test -race` found a couple of race conditions in the ipam unit tests

Note the race conditions arise purely from short-cuts in the unit tests, not from the deployed code.


bboreham:
replaced by #793

PR #783, opened 2015-05-28
[ipam] fix panic when granting space from one peer to another

### LINKED ISSUE #782: [ipam] panic when granting space from one peer to another
Example stack-trace: https://gist.github.com/inercia/08e28278dde093c0f2b8, but this does not give enough detail to know exactly what went wrong.

We can see a bug in `Ring.GrantRangeToHost()` where, if a peer selects a free block that (a) spans another token in the ring that it owns and (b) ends at an address that has no existing token, then the free space is wrong, which will cause the observed panic.

bboreham:
fixes #782

### DIFF
--- a/ipam/address/address.go
+++ b/ipam/address/address.go
@@ -52,3 +52,10 @@ func Subtract(a, b Address) Offset {
 	common.Assert(a >= b)
 	return Offset(a - b)
 }
+
+func Min(a, b Offset) Offset {
+	if a > b {
+		return b
+	}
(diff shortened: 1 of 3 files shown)


tomwilkie:
LGTM; are you going to stick this on master too?

rade:
@bboreham needs to merge 0.11 into master after this has been merged. That's what @paulbellamy has been doing with the other 0.11 PRs that have landed to date.

PR #789, opened 2015-05-28
make `weave status` fail when weave isn't running

### LINKED ISSUE #784: `weave status` should only succeed if weave is running
I recently encountered a strange issue when executing the command `weave --local status` in a script where I set `set -e` at the beginning.
The weave container was stopped at that moment and this command printed the following message  ...

```
weave container is not present; have you launched it?

```

... and the script kept going without exiting ! 

I think it might be due to a wrong exit code.

As a workaround, I needed to do that
`/weave --local status | grep -qv "weave container is not present; have you launched it?"`

rade:
but still always show weavedns status, if present

Closes #784.

### DIFF
--- a/weave
+++ b/weave
@@ -894,9 +894,11 @@ case "$COMMAND" in
         http_call $CONTAINER_NAME $HTTP_PORT POST /forget -d $(peer_args "$@")
         ;;
     status)
-        http_call $CONTAINER_NAME $HTTP_PORT GET /status || true
+        res=0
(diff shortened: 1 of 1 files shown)


rade:
@dpw mind taking a quick look at this? Can this be done even more neatly?

dpw:
> @dpw mind taking a quick look at this? Can this be done even more neatly?

Not that I can think of.  There are other variations, but none are substantially better.

PR #792, opened 2015-05-29
use an automatically assigned port for http tests

### LINKED ISSUE #720: [dns] Unit tests can pick an in-use port for test http server
And once they have, it happens on every run as the golang random number generator isn't being explicitly seeded. Perhaps it would be better to let the kernel pick an actually unused number (not sure if golang exposes this capability) or to loop until we bind successfully.

rade:
instead of random port, which can clash

Fixes #720.

### DIFF
--- a/ipam/http_test.go
+++ b/ipam/http_test.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
+	"net"
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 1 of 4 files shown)


weavedns/main.go
```diff
@@ -127,8 +127,13 @@ func main() {
...
 	}
 
+	httpListener, err := net.Listen("tcp", fmt.Sprintf(":%d", httpPort))
+	if err != nil {
+		Error.Fatal("[main] Unable to create http listener: ", err)
+	}
+
 	go SignalHandlerLoop(srv)
-	go weavedns.ListenHTTP(version, srv, domain, zone, httpPort)
+	go weavedns.ServeHTTP(httpListener, version, srv, domain, zone)
```
inercia: I think the `main()` function is not the place for initializing low level things like a TCP listener. Following the same interface provided by [the standard library](http://golang.org/pkg/net/http/), I think there should be a [`ListenAndServe()`](http://golang.org/pkg/net/http/#ListenAndServe) that creates the listener automatically, and a [`Serve()`](http://golang.org/pkg/net/http/#Serve) where you can provide your own listener...

rade: I wanted to avoid creating two five-arg functions.

This will all come out in the wash when merging weavedns into weaver.

bboreham:
Note ipam has the same pattern - https://github.com/weaveworks/weave/blob/f95b3971ab0a5523051db699aeb312fe8f0c1ced/ipam/http_test.go#L64

rade:
@bboreham please take a quick look at the ipam test changes.
@inercia any other comments?

ipam/http_test.go
```diff
@@ -98,9 +102,8 @@ func TestBadHttp(t *testing.T) {
 
 	alloc := makeAllocatorWithMockGossip(t, "08:00:27:01:c3:9a", testCIDR1, 1)
 	defer alloc.Stop()
-	port := rand.Intn(10000) + 32768
-	fmt.Println("BadHttp test on port", port)
-	go listenHTTP(port, alloc)
+	port := listenHTTP(alloc)
+	fmt.Println("Http test on port", port)
```
bboreham: test is called 'BadHttp' and Println used to reflect this

bboreham:
LGTM

PR #793, opened 2015-05-29
[ipam] Fix race conditions in unit tests

bboreham:
Removes  a couple of race conditions in the ipam unit tests found by `go test -race`.

Note that no adverse effects have been observed from this; just fixing them to make the unit tests closer to the way the code is used in production, and reduce noise when looking for a real race condition.


tomwilkie:
LGTM

PR #797, opened 2015-05-29
Various fixes to circleci integration tests

tomwilkie:
- Build the weave build container if we don't have it in the circle cache.  That way, when it changes, we just flush the circle cache.  It can be removed from the docker hub now.
- Add some timeouts to the test runs and GCE destroy vm function; this should help prevent stranded VMs.
- Checkin assert.sh instead of fetching it each time.  This allows us to modify it to expose test run time.  Its LGPL, so it should be fine.
- Add a simple scheduler to decide what tests run on which shard in circle.  This should balance tests more accurately, and bring down total test run time down by about a minute or two.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -41,6 +41,7 @@ test:
         parallel: true
     - cd $SRCDIR/test; . ./gce.sh hosts; ./run_all.sh:
         parallel: true
+        timeout: 300
   post:
     - cd $SRCDIR/test; ./gce.sh destroy:
         parallel: true
--- a/circle.yml
+++ b/circle.yml
@@ -26,6 +26,8 @@ dependencies:
         docker save weaveworks/weave-build >$WEAVE_BUILD;
       fi
   post:
+    - sudo apt-get install bc
+    - pip install requests
     - curl https://sdk.cloud.google.com | bash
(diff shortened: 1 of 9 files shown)


circle.yml
```diff
@@ -21,11 +21,13 @@ dependencies:
...
         docker load -i $WEAVE_BUILD;
       else
-        docker pull weaveworks/weave-build;
+        docker build -t weaveworks/weave-build build;
         mkdir -p $(dirname "$WEAVE_BUILD");
         docker save weaveworks/weave-build >$WEAVE_BUILD;
       fi
   post:
+    - sudo apt-get install bc
+    - pip install requests
```
paulbellamy: Are `requests` and `bc` needed?

paulbellamy:
Maybe a moving average, instead of all of history?

tomwilkie:
Update to use an exponential moving average.

paulbellamy:
Do we need `total_runs` anymore?

tomwilkie:
I do not, but I thought it would be nice to keep.

paulbellamy:
LGTM, then.

tomwilkie:
Tests on the different shards took 03:13 and 03:02 - yay for scheduler!

PR #801, opened 2015-05-29
Improve ipam status message before and after pending consensus

### LINKED ISSUE #787: [ipam] misleading status when in virgin state
When no IP address allocation requests have been made yet, the `weave status` output shows something like
````
Allocator subnet 10.0.0.0/16
Awaiting consensus: Nodes known: 0, Quorum size: 1
````
which does rather convey a sense of something being not quite right.

This alarmed and mislead at least one user - see #786.

bboreham:
Use the presence of a ticker object to determine whether we are actually waiting for consensus.
I also pulled the 'pending' bit out of the if-I-have-a-ring clause, because it is of interest either way.

Status before anything has happened now looks like this:

    Allocator subnet 10.3.0.0/16

and after one `weave run`:

    Allocator subnet 10.3.0.0/16 awaiting consensus: Nodes known: 1, Quorum size: 5
    Pending requests for Allocate for 8450fe9a4cf18299cc745d32b8d052ac9e6cb3f069710a58eaff19abed78d607, 

Fixes #787

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -456,10 +456,12 @@ func (alloc *Allocator) actorLoop(actionChan <-chan func()) {
 
 func (alloc *Allocator) string() string {
 	var buf bytes.Buffer
-	fmt.Fprintf(&buf, "Allocator subnet %s/%d\n", alloc.subnetStart.String(), alloc.prefixLen)
+	fmt.Fprintf(&buf, "Allocator subnet %s/%d", alloc.subnetStart.String(), alloc.prefixLen)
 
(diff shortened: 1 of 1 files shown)


rade:
I think the output when there are pending requests should look like this:
````
Allocator subnet 10.3.0.0/16 awaiting consensus: Nodes known: 1, Quorum size: 5
Pending requests:
  Allocate for 8450fe9a4cf18299cc745d32b8d052ac9e6cb3f069710a58eaff19abed78d607
  Allocate for ...
````

PR #806, opened 2015-06-01
weave launch-proxy should handle short-form DOCKER_HOST (w/o tcp://)

### LINKED ISSUE #803: [proxy] doesn't recognise short-form DOCKER_HOST setting
The weave script sets the `PROXY_HOST` to the host found in a `DOCKER_HOST` setting of `DOCKER_HOST=tcp://<host>:<port>`. However, the `tcp://` portion of that [is optional](https://github.com/docker/docker/blob/master/pkg/parsers/parsers.go#L22). If such a shorter form `DOCKER_HOST` is set, then the script does not set `PROXY_HOST` correctly.

paulbellamy:
Closes #803

### DIFF
--- a/weave
+++ b/weave
@@ -170,10 +170,6 @@ HTTP_PORT=6784
 DNS_HTTP_PORT=6785
 PROXY_PORT=12375
 PROXY_CONTAINER_NAME=weaveproxy
-PROXY_HOST="localhost"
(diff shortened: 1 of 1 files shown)


paulbellamy:
Realised that the `PROXY_HOST` var (because it's evaluated inside weaveexec), will never have `DOCKER_HOST` set, so this code can be removed. How was the bug found?

PR #813, opened 2015-06-02
First pass at Proxy SSL Support

### LINKED ISSUE #506: [proxy] accept client connections over TLS, using same certs as in the non-proxy case

paulbellamy:
I have looked at uploading the certs provided into a docker image, and mounting them into weaveexec that way, but I'm not really keen on messing around with people's private keys.

Will eventually close #506 

#### Massive Caveats:
* TLS files must be present on the target server (and readable)
* TLS file paths cannot contain spaces
* TLS file paths must be absolute (i.e. cannot be relative)
* Bash argument parsing is terrible

#### Possible nice to haves:
* Detect which certs docker is using and auto-configure


paulbellamy:
@errordeveloper

errordeveloper:
On a Boot2Docker VM provisioned by Machine, the args in use look like this:
```
 --tlsverify \
 --tlscacert=/var/lib/boot2docker/ca.pem \
 --tlskey=/var/lib/boot2docker/server-key.pem \
 --tlscert=/var/lib/boot2docker/server.pem
```

Supposedly, if proxy accepted exact same flags, it could simply grab these from them form the process table, filtering `--tls` prefix.

errordeveloper:
Additionaly, I'd like to note that there are [less environment variables](https://github.com/docker/docker/blob/v1.6.2/docker/flags.go#L15-L16) then there are [flags](https://github.com/docker/docker/blob/v1.6.2/docker/flags.go#L65-L67), may be we don't have to support both...

paulbellamy:
From reading the [https docs](https://docs.docker.com/articles/https/), it seems that the environment variable is primarily (only?) used for the client, while the cli flags are for the daemon.

errordeveloper:
> From reading the https docs, it seems that the environment variable is primarily (only?) used for the client, while the cli flags are for the daemon.

Essentially, yes, although with self-signed certs setup in a certain way (like ones @grkvlt is using in Clocker), the env vars actually don't cut it and client also has to use flags.

weaveproxy/main.go
```diff
@@ -16,13 +16,29 @@ var (
...
 	getopt.BoolVarLong(&debug, "debug", 'd', "log debugging information")
 	getopt.BoolVarLong(&justVersion, "version", 0, "print version and exit")
-	getopt.StringVar(&target, 'H', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultTarget))
 	getopt.StringVar(&listen, 'L', fmt.Sprintf("address on which to listen (default %s)", defaultListen))
+	getopt.StringVar(&target, 'H', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultTarget))
+	getopt.StringVarLong(&tlsCACert, "tlscacert", 0, "Trust certs signed only by this CA")
+	getopt.StringVarLong(&tlsCert, "tlscert", 0, "Path to TLS certificate file")
+	getopt.BoolVarLong(&tlsEnabled, "tls", 0, "Use TLS; implied by --tlsverify")
+	getopt.StringVarLong(&tlsKey, "tlskey", 0, "Path to TLS key file")
+	getopt.BoolVarLong(&tlsVerify, "tlsverify", 0, "Use TLS and verify the remote")
```
rade: default values of 0 for string vars?!?

paulbellamy: no short-flag, not default values.

errordeveloper:
So, passing all but `--tlsverify` flag works on Machine/Boot2Docker. Here is how I set it up, which may be of interest for integration tests.
```
> docker-machine create -d virtualbox test-1
...
> docker-machine config 
--tlsverify --tlscacert="/Users/ilya/.docker/machine/machines/test-1/ca.pem" --tlscert="/Users/ilya/.docker/machine/machines/test-1/cert.pem" --tlskey="/Users/ilya/.docker/machine/machines/test-1/key.pem" -H=tcp://192.168.99.126:2376
> for c in weave weavedns weaveexec; do docker `docker-machine config` load -i ${c}.tar; done
...
> export DOCKER_CLIENT_ARGS="`docker-machine config`"
> ./weave launch
> ./weave launch-dns 10.11.1.1/24
```

Docker run on this box with
```
/usr/local/bin/docker -d -D -g /var/lib/docker -H unix:// -H tcp://0.0.0.0:2375 --tlsverify --tlscacert=/var/lib/boot2docker/ca.pem --tlskey=/var/lib/boot2docker/server-key.pem --tlscert=/var/lib/boot2docker/server.pem --label=provider=virtualbox -H tcp://0.0.0.0:2376
```

I first copied all of the `--tls*` flags and passing those to `launch-proxy` as they are; then tried 
using `docker $(docker-machine config | sed 's/:2376/:12375/') ps` and

client output
```
FATA[0000] An error occurred trying to connect: Get https://192.168.99.126:12375/v1.18/containers/json: remote error: bad certificate 
```
proxy log
```
2015/06/02 11:25:45 http: TLS handshake error from 192.168.99.1:53208: tls: failed to verify client's certificate: x509: certificate signed by unknown authority
```

Without `--tlsverify` on both ends it works okay, so it looks like the way proxy veriefies the cert is just a bit stricter then how Docker seems to do it. This is a self-signed CA, as far as I understand...

paulbellamy:
Accidentally merged this. D'oh. Reverted in #814, and created #815 to replace this.

paulbellamy:
@errordeveloper What is the output of `docker-machine config` ?

errordeveloper:
@paulbellamy see https://github.com/weaveworks/weave/pull/813#issuecomment-107926497

Issue #805, opened 2015-06-01
[proxy] Add support for custom port


paulbellamy:

rade:
It's probably worth tackling this together with #810. `-listen <ip> | :<port> | <ip>:<port>`? and a space separated list for specifying more than one.

paulbellamy:
Docker uses multiple `-H` arguments for multiple listening ports, so let's match that: `weave launch-proxy -L <ip> -L :<port> -L <ip>:<port>`.

Would be nice to rename the `-H` and `-L` arguments to the proxy to closer match docker daemon's arguments, but, that will break backwards-compat.

rade:
We can't match that w/o doing our own flag parsing since golang's flag parser processes multiple occurrences of the same flag by stomping on the old values.

 *That*'s why I suggested what I did.

> Would be nice to rename the `-H` and `-L` arguments to the proxy to closer match docker daemon's arguments, but, that will break backwards-compat.

backward compatibility is of no concern.

paulbellamy:
We already do our own arg parsing using [this](https://godoc.org/code.google.com/p/getopt), at the moment. It would just be a case of swapping libraries to something that supports multi-args, (and maybe has nicer syntax for args, e.g. [http://godoc.org/github.com/codegangsta/cli#StringSlice](http://godoc.org/github.com/codegangsta/cli#StringSliceFlag))

rade:
argh. I missed the esoteric option parsing library in the proxy review. Are you saying that isn't supporting multi-args either?

paulbellamy:
No, it's just for the double-hyphen, at the moment.

paulbellamy:
Actually, getopt does multi-args. Was hidden further down in the docs.

rade:
I have just noticed that the proxy already has an (undocumented!) `-L` option to specify what address and port to listen on. So what actually needs doing here? Other than documentation.

paulbellamy:
See https://github.com/weaveworks/weave/issues/810#issuecomment-109920081

PR #815, opened 2015-06-02
Proxy TLS Support

### LINKED ISSUE #506: [proxy] accept client connections over TLS, using same certs as in the non-proxy case

paulbellamy:
Accidentally merged and reverted #813. Continued work here...

Closes #506

### DIFF
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -16,13 +16,25 @@ var (
 )
 
 func main() {
-	var target, listen string
-	var withDNS, withIPAM, debug, justVersion bool
+	var (
+		debug       bool
+		justVersion bool
+		listen      string
+		target      string
+		tlsConfig   proxy.TLSConfig
+		withDNS     bool
+		withIPAM    bool
+	)
 
 	getopt.BoolVarLong(&debug, "debug", 'd', "log debugging information")
 	getopt.BoolVarLong(&justVersion, "version", 0, "print version and exit")
-	getopt.StringVar(&target, 'H', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultTarget))
 	getopt.StringVar(&listen, 'L', fmt.Sprintf("address on which to listen (default %s)", defaultListen))
+	getopt.StringVar(&target, 'H', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultTarget))
+	getopt.StringVarLong(&tlsConfig.CACert, "tlscacert", 0, "Trust certs signed only by this CA")
+	getopt.StringVarLong(&tlsConfig.Cert, "tlscert", 0, "Path to TLS certificate file")
+	getopt.BoolVarLong(&tlsConfig.Enabled, "tls", 0, "Use TLS; implied by --tlsverify")
+	getopt.StringVarLong(&tlsConfig.Key, "tlskey", 0, "Path to TLS key file")
+	getopt.BoolVarLong(&tlsConfig.Verify, "tlsverify", 0, "Use TLS and verify the remote")
 	getopt.BoolVarLong(&withDNS, "with-dns", 'w', "instruct created containers to use weaveDNS as their nameserver")
 	getopt.BoolVarLong(&withIPAM, "with-ipam", 'i', "automatically allocate addresses for containers without a WEAVE_CIDR")
 	getopt.Parse()
@@ -30,6 +42,10 @@ func main() {
 	if justVersion {
 		fmt.Printf("weave proxy %s\n", version)
 		os.Exit(0)
+	}
+
+	if debug {
+		InitDefaultLogging(true)
 	}
 
 	if target == "" {
@@ -40,17 +56,17 @@ func main() {
 		listen = defaultListen
 	}
 
-	if debug {
-		InitDefaultLogging(true)
-	}
-
-	p, err := proxy.NewProxy(version, target, listen, withDNS, withIPAM)
+	p, err := proxy.NewProxy(
+		version,
+		target,
+		listen,
+		withDNS,
+		withIPAM,
+		&tlsConfig,
+	)
 	if err != nil {
 		Error.Fatalf("Could not start proxy: %s", err)
 	}
-
-	Info.Printf("Listening on %s", listen)
-	Info.Printf("Proxying %s", target)
 
 	if err := p.ListenAndServe(); err != nil {
 		Error.Fatalf("Could not listen on %s: %s", listen, err)
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -19,7 +19,10 @@ func main() {
 	var (
 		debug       bool
 		justVersion bool
-		c           proxy.Config
+		c           = proxy.Config{
+			DockerAddr: defaultDockerAddr,
+			ListenAddr: defaultListenAddr,
+		}
 	)
 
 	c.Version = version
@@ -43,14 +46,6 @@ func main() {
 
 	if debug {
 		InitDefaultLogging(true)
-	}
-
-	if c.DockerAddr == "" {
-		c.DockerAddr = defaultDockerAddr
-	}
-
-	if c.ListenAddr == "" {
-		c.ListenAddr = defaultListenAddr
 	}
 
 	p, err := proxy.NewProxy(c)
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -10,33 +10,30 @@ import (
 )
 
 var (
-	version       = "(unreleased version)"
-	defaultTarget = "unix:///var/run/docker.sock"
-	defaultListen = ":12375"
+	version           = "(unreleased version)"
+	defaultDockerAddr = "unix:///var/run/docker.sock"
+	defaultListenAddr = ":12375"
 )
 
 func main() {
 	var (
 		debug       bool
 		justVersion bool
-		listen      string
-		target      string
-		tlsConfig   proxy.TLSConfig
-		withDNS     bool
-		withIPAM    bool
+		c           proxy.Config
 	)
 
+	c.Version = version
 	getopt.BoolVarLong(&debug, "debug", 'd', "log debugging information")
 	getopt.BoolVarLong(&justVersion, "version", 0, "print version and exit")
-	getopt.StringVar(&listen, 'L', fmt.Sprintf("address on which to listen (default %s)", defaultListen))
-	getopt.StringVar(&target, 'H', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultTarget))
-	getopt.StringVarLong(&tlsConfig.CACert, "tlscacert", 0, "Trust certs signed only by this CA")
-	getopt.StringVarLong(&tlsConfig.Cert, "tlscert", 0, "Path to TLS certificate file")
-	getopt.BoolVarLong(&tlsConfig.Enabled, "tls", 0, "Use TLS; implied by --tlsverify")
-	getopt.StringVarLong(&tlsConfig.Key, "tlskey", 0, "Path to TLS key file")
-	getopt.BoolVarLong(&tlsConfig.Verify, "tlsverify", 0, "Use TLS and verify the remote")
-	getopt.BoolVarLong(&withDNS, "with-dns", 'w', "instruct created containers to use weaveDNS as their nameserver")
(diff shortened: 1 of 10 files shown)


paulbellamy:
@errordeveloper, please give this a try with docker-machine, and `--tlsverify`.

rade:
This is supposed to address #506, right?

paulbellamy:
Yes. This does not use TLS to talk to the upstream docker daemon.

prog/weaveproxy/main.go
```diff
@@ -16,13 +16,29 @@ var (
...
+	var (
+		debug       bool
+		justVersion bool
+		listen      string
+		target      string
+		tlsCACert   string
+		tlsCert     string
+		tlsEnabled  bool
+		tlsKey      string
+		tlsVerify   bool
```
rade: I suggest reading the `tls*` directly into a `TLSConfig` instance.

It's also worth considering the introduction of a `Config` struct for *all* proxy configuration.

test/tls/generate_certs.go
```diff
@@ -0,0 +1,80 @@
...
+					caKeyPath,
+					org,
+					bits,
+				)
+				if err != nil {
+					log.Fatalf("error generating server cert: %s", err)
+				}
+			}
+		}
+	}
```
rade:
````
if len(os.Args) <= 1 {
    return
}
for ...
````

**Comment by paulbellamy** [comment]
Need to update docs, yet

**Comment by rade** [comment]
docs

**Comment by paulbellamy** [comment]
Check with the docker daemon using tls. Does it still listen on unix socket when following: https://docs.docker.com/articles/https/ ?

File: site/proxy.md [hunk 3]
```diff
@@ -102,7 +112,34 @@ resolution. WeaveDNS is used in addition to any dns servers specified
...
+If you are using docker with TLS enabled, you probably want to talk to
+the proxy over TLS. Launching the proxy with TLS support is done
+with the same command-line flags as the docker daemon. For example, if
+you have generated your certificates and keys into the docker host's
+`/tls` directory, we can launch the proxy with:
+
+    host1$ weave launch-proxy --tlsverify --tlscacert=/tls/ca.pem \
+             --tlscert=/tls/server-cert.pem --tlskey=/tls/server-key.pem
+
+The paths to your certificates and key, Make sure that your
```
rade: typo

paulbellamy:
It never uses TLS on the unix socker, but needs to be specifically told to listen there. Added a note to the docs.

errordeveloper:
I have tested this now and it works with Machine/Swarm. [Here](https://github.com/errordeveloper/weave-demos/blob/7254f7b/quartet/scripts/setup-cluster-dev.sh) is how I set it up, if anyone is curious...

rade:
what's left to do here?

paulbellamy:
Afaik it is ready to go.

We still don't support certificate/key paths with spaces in them, but have decided to ignore that for now.

Better arg parsing can potentially be addressed in #805.

prog/weaveproxy/main.go
```diff
@@ -32,27 +41,24 @@ func main() {
...
-	if listen == "" {
-		listen = defaultListen
+	if c.DockerAddr == "" {
+		c.DockerAddr = defaultDockerAddr
 	}
 
-	if debug {
-		InitDefaultLogging(true)
+	if c.ListenAddr == "" {
+		c.ListenAddr = defaultListenAddr
```
rade: surely getopt takes care of this.

proxy/proxy.go
```diff
@@ -90,20 +103,20 @@ func (proxy *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 }
 
 // Return status string
 func (proxy *Proxy) Status() string {
 	var buf bytes.Buffer
-	fmt.Fprintln(&buf, "Listen address is", proxy.listenAddr)
-	fmt.Fprintln(&buf, "Docker address is", proxy.dockerAddr)
-	if proxy.withDNS {
+	fmt.Fprintln(&buf, "Listen address is", proxy.ListenAddr)
+	fmt.Fprintln(&buf, "Docker address is", proxy.DockerAddr)
```
rade: We also need to say something like "TLS is on/off/verify".

prog/weaveproxy/main.go
```diff
@@ -10,21 +10,33 @@ import (
...
-	var target, listen string
-	var withDNS, withIPAM, debug, justVersion bool
-
+	var (
+		debug       bool
+		justVersion bool
+		c           = proxy.Config{
+			DockerAddr: defaultDockerAddr,
+			ListenAddr: defaultListenAddr,
+		}
```
rade: I don't get it. Are you saying that

    getopt.StringVar(&c.ListenAddr, 'L', fmt.Sprintf("address on which to listen (default %s)", defaultListenAddr))

does *not* set `c.ListenAddr = defaultListenAddr` when the option isn't specified?

site/proxy.md
```diff
@@ -102,7 +112,38 @@ resolution. WeaveDNS is used in addition to any dns servers specified
...
+If you are using docker with TLS enabled, you probably want to talk to
+the proxy over TLS. Launching the proxy with TLS support is done
+with the same command-line flags as the docker daemon. For example, if
+you have generated your certificates and keys into the docker host's
+`/tls` directory, we can launch the proxy with:
+
+    host1$ weave launch-proxy --tlsverify --tlscacert=/tls/ca.pem \
+             --tlscert=/tls/server-cert.pem --tlskey=/tls/server-key.pem
+
+The paths to your certificates and key, must be provided as absolute
```
rade: stray comma

site/proxy.md
```diff
@@ -102,7 +112,38 @@ resolution. WeaveDNS is used in addition to any dns servers specified
...
+    host1$ docker --tlsverify --tlscacert=ca.pem --tlscert=cert.pem \
+             --tlskey=key.pem -H=tcp://host1:12375 version
+
+The certificates and keys used by the client will need to be present
+on your client machine.
+
+Because the proxy connects to the docker daemon at
+`unix:///var/run/docker.sock`, you must ensure that the daemon is
+listening there. To do this, you will need to pass the `-H
+unix:///var/run/docker.sock` when booting the docker daemon.
```
rade: missing 'option'. also, should say something along the line of "... in addition to the -H options for configuring the TCP listener". Check the docker docs for what the correct terminology is here. And if you can dredge up any docker documentation anywhere that shows multiple `-H`, please refer to that.

test/setup.sh
```diff
@@ -15,3 +15,9 @@ for HOST in $HOSTS; do
     run_on $HOST chmod a+x $DOCKER_NS ./weave
     run_on $HOST sudo service docker restart
 done
+
+echo Setting up Docker TLS on hosts
+(cd ./tls && go get ./... && go run generate_certs.go $HOSTS)
+for HOST in $HOSTS; do
+    rsync -az -e "$SSH" ./tls/ $HOST:~/tls
+done
```
rade: make this part of the earlier `for` loop.

Also, I am guessing there is a reason we don't use rsync elsewhere. @squaremo might know.

squaremo: No particular reason; so if it's more convenient than `cp` then seems OK to me.

prog/weaveproxy/main.go
paulbellamy: Exactly. That is just a usage string. If the option is not set, it doesn't modify the value of ListenAddr.

weave
```diff
@@ -746,6 +746,53 @@ get_ipam_cidr() {
...
+            PROXY_ARGS="$PROXY_ARGS --tlskey /home/weave/tls/key.pem"
+            ;;
+          *)
+            PROXY_ARGS="$PROXY_ARGS $1"
+            ;;
+        esac
+        shift
+    done
+}
+
```
rade: A `tls_arg` function would do wonders here. You've invoke it s.t. like this:
````
        --tlscacert)
            shift
            tls_arg "cacert" "ca" $1
            ;;
          --tlscacert=*)
            tls_arg "cacert" "ca" ${1#*=}
            ;;
````

And you wouldn't even need the 2nd param if the CA cert file was named `cacert.pem` instead of `ca.pem`.

Comment (rade) [inline weave]: (you may need double quotes around the last arg; I have no idea).

File: prog/weaveproxy/main.go (hunk 6 above)
Comment (rade) [inline prog/weaveproxy/main.go]: That is retarded.

**Comment by rade** [comment]
can we change the title of this PR, please?

File: site/proxy.md [hunk 11]
```diff
@@ -102,7 +112,42 @@ resolution. WeaveDNS is used in addition to any dns servers specified
...
+on your client machine.
+
+Because the proxy connects to the docker daemon at
+`unix:///var/run/docker.sock`, you must ensure that the daemon is
+listening there. To do this, you will need to pass the `-H
+unix:///var/run/docker.sock` option when booting the docker daemon, in
+addition to the `-H` options for configuring the TCP listener. Passing
+multiple `-H` options to the docker daemon is documented on the
+[docker basics
+documentation](https://docs.docker.com/articles/basics/).
```
rade: https://docs.docker.com/articles/basics/#bind-docker-to-another-hostport-or-a-unix-socket is a better link. And just say "See [the Docker documentation](https://docs.docker.com/articles/basics/#bind-docker-to-another-hostport-or-a-unix-socket) for an example."

site/proxy.md
```diff
@@ -102,7 +112,42 @@ resolution. WeaveDNS is used in addition to any dns servers specified
...
+
+More details on generating certificates, and configuring docker with
+TLS can be found in the [docker
+documentation](https://docs.docker.com/articles/https/).
```
rade: I think this paragraph should be tagged onto the end of "The paths to your existing certificates ..." paragraph.

weave
```diff
@@ -880,12 +927,15 @@ case "$COMMAND" in
...
+        proxy_args "$@"
         PROXY_CONTAINER=$(docker run --privileged -d --name=$PROXY_CONTAINER_NAME --net=host \
             -p $PROXY_PORT:$PROXY_PORT/tcp \
             -v /var/run/docker.sock:/var/run/docker.sock -v /proc:/hostproc \
             -e PROCFS=/hostproc \
+            $PROXY_VOLUMES \
             --entrypoint=/home/weave/weaveproxy \
-            $WEAVEPROXY_DOCKER_ARGS $EXEC_IMAGE "$@")
+            $WEAVEPROXY_DOCKER_ARGS $EXEC_IMAGE $PROXY_ARGS)
+
```
rade: stray newline

weave
```diff
@@ -880,12 +927,15 @@ case "$COMMAND" in
         # Set WEAVEPROXY_DOCKER_ARGS in the environment in order to supply
         # additional parameters, such as resource limits, to docker
         # when launching the weaveproxy container.
+        proxy_args "$@"
         PROXY_CONTAINER=$(docker run --privileged -d --name=$PROXY_CONTAINER_NAME --net=host \
             -p $PROXY_PORT:$PROXY_PORT/tcp \
             -v /var/run/docker.sock:/var/run/docker.sock -v /proc:/hostproc \
             -e PROCFS=/hostproc \
+            $PROXY_VOLUMES \
```
rade: I'd tag that onto the beginning of the `-v` line.

site/proxy.md
```diff
@@ -102,7 +112,41 @@ resolution. WeaveDNS is used in addition to any dns servers specified
 via the `--dns` option. More details on weaveDNS can be found in the
 [weaveDNS documentation](weavedns.html).
 
-## Limitations
+## <a name="usage-with-tls"></a>Usage with TLS
+
+If you are using docker with TLS enabled, you probably want to talk to
```
rade: s/enabled/

site/proxy.md
```diff
@@ -102,7 +112,41 @@ resolution. WeaveDNS is used in addition to any dns servers specified
...
+    host1$ weave launch-proxy --tlsverify --tlscacert=/tls/ca.pem \
+             --tlscert=/tls/server-cert.pem --tlskey=/tls/server-key.pem
+
+The paths to your certificates and key must be provided as absolute
+paths which exist on the docker host.
+
+Because the proxy connects to the docker daemon at
+`unix:///var/run/docker.sock`, you must ensure that the daemon is
+listening there. To do this, you will need to pass the `-H
+unix:///var/run/docker.sock` option when booting the docker daemon, in
```
rade: s/booting/starting

rade:
If you are done here, please fix the title and assign to me.

PR #817, opened 2015-06-02
move exe and docker image producing dirs to subdir

### LINKED ISSUE #420: shrink top-level dir by moving exe/image sub-dirs
I have an idea to simplify the structure of the weave repo, and make things more uniform.

Currently, we have the 'router' directory, and a corresponding 'weaver' directory for the weave container Dockerfile (and main.go). Then we have the 'nameserver' directory, and a corresponding 'weavedns' directory containing the weavedns container Dockerfile (and main.go).  It's not hugely confusing, but it does lead to an impression of clutter in the top-level directory of the repo.

It would be nice to reduce that clutter.  It would be nice if the names of the directories containing Dockerfiles uniformly matched the container image names.  And one day it might be nice to have a go package name that is the same as a container name.

So I suggest moving:

* weaver -> containers/weave
* weavedns -> containers/weavedns
* tools -> containers/tools
* (And build -> containers/build if #387 gets merged)

And updating the makefile and any other references, of course.

rade:
Closes #420.


paulbellamy:
LGTM, but at the risk of prolonging bikeshedding, I'm curious why `prog/`?

PR #822, opened 2015-06-02
Fix race condition in allocator fuzz-test

### LINKED ISSUE #802: [ipam] race condition in unit test
In `TestAllocatorFuzz()`, it is possible for the `allocateAgain` action to select a target, then for that target's address to be freed on another goroutine, then to attempt the allocation and hence get a different result.

One possible fix at https://github.com/weaveworks/weave/commit/7aec12063187790bc6541f17d28d060fa6b4926c

bboreham:
Introduce a flag 'block' to prevent an address from being freed during re-test.

Fixes #802

### DIFF
--- a/ipam/allocator_test.go
+++ b/ipam/allocator_test.go
@@ -298,6 +298,7 @@ func TestAllocatorFuzz(t *testing.T) {
 	type result struct {
 		name  string
 		alloc int32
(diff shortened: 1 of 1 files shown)


tomwilkie:
LGTM

PR #824, opened 2015-06-02
[proxy] We can only rely on receiving a signal when we are a container

paulbellamy:
Due to https://github.com/docker/docker/issues/9098, there is no way to
signal an exec. Instead we'll have to just poll the interface and hope.

Adds `weavewait -s` which is the way to tell weavewait to skip waiting for a signal, which we use on execs

### DIFF
--- a/prog/weavewait/main.go
+++ b/prog/weavewait/main.go
@@ -1,32 +1,42 @@
 package main
 
 import (
-	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


paulbellamy:
We use `-s` as a negative bool on execs (vs a positive bool on containers) to avoid further polluting the cmd-line on `docker ps`

Issue #675, opened 2015-05-16
[ipam] manage allocation on multiple subnets


rade:
ATM, IPAM can only allocate IPs in a single subnet. We should extend that to multiple subnets.

Strawman UI: When the IP address(es) given to `weave run` et al are the base address of the network then we allocate an IP address in that network. Note that this is compatible with existing usage.

The alternative is to use some sort of prefix or other indicator to distinguish the 'use this IP' and 'allocate me an IP in this network' cases. And check that the address given to the former is not a network address, and the latter is not a non-network address. IMO that is a worse UI.

Some implementation challenges...

1. Single ring or one ring per sub-net?
2. Should we permit overlapping ranges? non-ipam weave allows that. is there a compelling use case?
3. ATM the range from which IPAM allocates needs to be specified in `weave launch`. We don't want this here, since the subnets may not be known at launch time. The ip range given to weave launch on startup should essentially be treated as the default allocation range, i.e. what gets used when no range has been specified at the time of allocation.

bboreham:
For part 3, istm that we would need a consensus approach to avoid multiple peers independently initializing a subnet.  Maybe split this feature into pre-specified subnets and dynamically-specified?

rade:
yes, it would require consensus, but we already do that for the currently single subnet. why can't we repeat the process for others? It's still a 'one off' consensus, but *per subnet*.

bboreham:
I was just thinking it could be a separate bit of engineering and documentation.  E.g. there will be new error conditions - allocating an address can hang in a partition, if it happens to be the first one in that subnet.

rade:
That is no different to what happens now, for the single IPAM-managed subnet. Since, iirc, ring creation consensus is always deferred until an allocation is requested (and we haven't received a ring via gossip by them).

bboreham:
Good point, well made.

bboreham:
Proposed change to URL scheme to accommodate the strawman UI:

#### Current URL scheme

Alloc: POST /ip/name
Claim: PUT /ip/name/addr/prefixlen

#### Suggested change

Alloc: POST /ip/name/addr/prefixlen
 - if address part (after subnetmask) is 0 then this acts like Alloc; if non-zero then acts like Claim.
 - in both cases, returns (allocated or claimed) address in CIDR format
 - previous version (with no addr/prefixlen) is still allowed and acts as before

tomwilkie:
[...]

Issue #763, opened 2015-05-27
[dns][ipam] have a default subnet for DNS


rade:
Similar to #743, and as an improvement on  #675, we should be able to launch DNS without specifying any network.

rade:
I wonder whether we can get away with *not* having a default subnet specifically for DNS, i.e. just use the main default subnet.

Besides isolation, does anything actually break when apps share a subnet with DNS?

rade:
> Besides isolation, does anything actually break when apps share a subnet with DNS?

#899 lists a few things. Anything else? In particular, are there possible *accidental* breakages, e.g. apps interfering with weaveDNS or vice versa? Obviously apps using multicast on port 53 would be problematic, but that's an edge case.

bboreham:
mDNS is on port 5353. Not that that changes the argument.

Issue #743, opened 2015-05-26
[ipam] have a default for -iprange


bboreham:
In general, we don't know what IP address ranges are in use on the target, so we ask the user to pick a free range.

However, we could have a default, check whether that default is OK to use, and refuse to start if not.  The default might be 10.20.0.0/16; this is fairly large and steers clear of ranges like 10.0.2 which are used by virtualbox.

We can use the same approach that Docker does to check if a range is in use: see if there is any routing rule on the host that intersects the range.

Having just a single default is better than having several and picking the first one that is free, because the latter approach can lead to inconsistent choices across hosts which are much harder to diagnose.

rade:
afaict virtualbox uses `10.0.2`.

Would be good to have something larger than /16, but might be tricky to pick one that has a low probability of clashing.

bboreham:
After #675 is fixed, the requirement will be to pick both an overall range and a default subnet within that range.

Suggest 10.0.0.0/8 and 10.20.0.0/16, respectively.

bboreham:
After discussion, we should probably insist that `-iprange` is completely available for ipam to allocate within, so 10/8 not so good.  How about 10.16.0.0/12 and 10.16.0.0/16 ?

bboreham:
Do we need an option to turn ipam off (i.e. what used to happen if you didn't specify -iprange) ?

Issue #676, opened 2015-05-16
[ipam] allocate subnets


rade:
IPAM at present manages allocation of IP addresses in a given subnet (or, with #675, multiple given subnets). But determining these subnets is in itself an allocation problem, i.e. a user often won't care what specific subnet is used for an application, just that is distinct from others and sufficiently large to accommodate all their containers.

Dealing with this pushes us in the direction of symbolic subnets / application names. And probably consensus to manage their mapping to real subnets.

bboreham:
I'm thinking the `weave run` UX should allow just a subnet name, and if that name is not currently associated with a subnet then weave will allocate a new one.

This implies that weave will choose a moderate size of subnet in order to allow many to be used at once.  For example, the 256 subnets between 10.0/16 and 10.255/16.

If the user needs to put specific requirements on a subnet, such as the size, they should pre-allocate it with a separate command, e.g. `weave subnet`

bboreham:
Proposed amendment to 'alllocate' URL scheme to allow symbolic names for subnets

Current: POST /ip/name
Proposed: POST /ip/subnetname/name

('claim' does not need to change, as the full address being claimed is known)

tomwilkie:
What if a user typos a subnet name on weave run?  It would be safe to refuse (subnet not found), and have an explicit weave subnet create command.

squaremo:
For the purpose of the docker plugin, it is necessary to be able to

 1. Allocate a subnet, supplying a name for it
 2. Allocate an IP on a subnet, supplying the subnet name

NB the libnetwork driver SPI assumes that trying to create a network (subnet, for weave) that already exists will result in an error; so, subnet creation should be explicit rather than implicit.

rade:
> so, subnet creation should be explicit rather than implicit.

We can have both.

squaremo:
> so, subnet creation should be explicit rather than implicit.
> 
> We can have both.

OK: it must be possible to explicitly allocate a subnet, and that must fail if it has already been allocated.

squaremo:
What is blocking this? It is the one thing that stops plugin (k8s, docker) work from progressing.

rade:
Somewhat related to #179.

Also, high-level description/UI from #242:

> ````bash
root@a:~# weave run --name=pong.foo -d -ti ubuntu
root@a:~# weave run --name=pong.bar -d -ti ubuntu
root@a:~# weave run --name=.foo -ti ubuntu
````
TODO: flesh this out
>
[...]

PR #827, opened 2015-06-02
multi-subnet ipam

bboreham:
Addresses #675, #861 and #862.


ipam/allocator.go
```diff
@@ -46,16 +44,15 @@ type operation interface {
...
-	prefixLen        int                        // network prefix length, e.g. 24 for a /24 network
-	ring             *ring.Ring                 // information on ranges owned by all peers
-	space            space.Space                // more detail on ranges owned by us
-	owned            map[string]address.Address // who owns what address, indexed by container-ID
-	nicknames        map[router.PeerName]string // so we can map nicknames for rmpeer
-	pendingAllocates []operation                // held until we get some free space
-	pendingClaims    []operation                // held until we know who owns the space
-	gossip           router.Gossip              // our link to the outside world for sending messages
+	universe         address.CIDR // superset of all ranges
+	defaultSubnet    address.CIDR
```
rade: cosmetic: add a comment here, and make them all align vertically.

ipam/http.go
```diff
@@ -33,16 +33,41 @@ func (alloc *Allocator) HandleHTTP(router *mux.Router) {
...
+				badRequest(w, fmt.Errorf("Unable to allocate: %s", err))
+				return
+			}
+
+			fmt.Fprintf(w, "%s/%d", newAddr.String(), cidr.PrefixLen)
+		} else {
+			fmt.Fprintf(w, "%s/%d", addr.String(), cidr.PrefixLen)
+		}
+	})
+
```
rade: I'd turn this into
````
addr, cidr, err := ...
if err != nil {
    ....
}
if addr == cidr.Start {
...
}
fmt.Fprintf(w, "%s/%d", addr.String(), cidr.PrefixLen)
````

Also, perhaps turn the result into a `CIDR`, which ought to produce stringify to the intended result.

File: ipam/http.go [hunk 3]
```diff
@@ -33,16 +33,41 @@ func (alloc *Allocator) HandleHTTP(router *mux.Router) {
...
 		ident := mux.Vars(r)["id"]
-		newAddr, err := alloc.Allocate(ident, closedChan)
+		newAddr, err := alloc.AllocateInSubnet(ident, alloc.defaultSubnet, closedChan)
 		if err != nil {
 			badRequest(w, err)
 			return
 		}
 
-		fmt.Fprintf(w, "%s/%d", newAddr.String(), alloc.prefixLen)
+		fmt.Fprintf(w, "%s/%d", newAddr.String(), alloc.defaultSubnet.PrefixLen)
```
rade: As above, perhaps turn the result into a CIDR, which ought to produce stringify to the intended result.

rade:
I'm not keen on the `In[Subnet|Range]` method name suffixes. There doesn't seem to be any ambiguity, so they just add noise.

rade:
docs need updating.

squaremo:
The CLI component to this invites confusion: an allocate-on-subnet argument (e.g., 10.20.45.128/25) looks very similar to an exactly-this-IP argument (e.g., 10.20.46.129/25), but these have quite different (and not immediately distinguishable) outcomes. I would much prefer this to be explicit; e.g.,

    weave run ip:10.20.46.129/25 net:10.20.45.128/25 ...

I don't always favour spelling things out in a user interface; but in this case, I think checking the user's stated intention against the supplied data will head off many support queries.

rade:
I know we don't care all that much about backward compatibility, but being explicit here will break every single weave usage out there.

I suppose we could make the prefix optional... "If you fear you might accidentally supply a network when you meant a specific ip address, or vice versa, use the `ip:` and `net:` prefixes"

Also, what would the env var (for the proxy) look like? `WEAVE_CIDR='ip:10.20.46.129/25 net:10.20.45.128/25'`?

squaremo:
> I suppose we could make the prefix optional

Better would be to use just one prefix, `net:`, and let unadorned addresses be interpreted in the usual way.

Or, my preference (which I didn't mention above, because I thought it would introduce parsing difficulties): `--ip <address>` and `--net <address>`, with an unaccompanied address being treated in the usual way.

For the proxy, perhaps `WEAVE_CIDR="..."` and `WEAVE_NET="..."` would work. There's far less of a backwards compatibility concern of course, but this would anyway avoid problems.

squaremo:
Actually, we should probably try not to clash with docker arguments, while we're still doing "best guess parsing" of command line arguments. So, instead of `--net`, perhaps `--subnet`. (`--network` is not currently a docker argument, but would be courting danger.)

By the way, we need to account for using the default subnet alongside other addresses. `net:default` or `--net default` would be adequate.

awh:
> Or, my preference (which I didn't mention above, because I thought it would introduce parsing difficulties): `--ip <address>` and `--net <address>`, with an unaccompanied address being treated in the usual way.

+1

squaremo:
> For the proxy, perhaps WEAVE_CIDR="..." and WEAVE_NET="..." would work

.. or just expect the verbatim arguments as suggested by @rade , since they get spooged to the weave script anyway (I think).

rade:
This PR is at the root of quite a large dependency tree. So let's go for the most expedient to implement UI to start with. We can deal with any improvements in separate issues/PRs.

bboreham:
For the CLI syntax, I went with `--subnet 10.0.0.0/8` means allocate in that subnet, and `10.0.0.1/8` means use that exact address without involving IPAM.

ipam/address/address.go
```diff
@@ -20,6 +27,34 @@ func ParseIP(s string) (Address, error) {
...
+func (cidr CIDR) End() Address { return Add(cidr.Start, cidr.Size()) }
+func (cidr CIDR) Blank() bool  { return cidr.Start == 0 && cidr.PrefixLen == 0 }
+
+func (cidr CIDR) String() string {
+	return fmt.Sprintf("%s/%d", cidr.Start.String(), cidr.PrefixLen)
+}
+
+func (cidr1 CIDR) Overlaps(cidr2 CIDR) bool {
+	return cidr1.Start >= cidr2.Start && cidr1.Start < cidr2.End() ||
+		cidr2.Start >= cidr1.Start && cidr2.Start < cidr1.End()
```
tomwilkie: Interview question!

!(cidr1.End() <= cidr2.Start || cidr1.Start > cidr2.End())

tomwilkie:
I think docs/ipam.md needs updating too.

ipam/allocator.go
```diff
@@ -197,29 +176,46 @@ func hasBeenCancelled(cancelChan <-chan bool) func() bool {
 
 // Actor client API
 
+func (alloc *Allocator) SetDefaultSubnet(subnet address.CIDR) error {
+	resultChan := make(chan error)
+	if subnet.Size() < 4 {
```
tomwilkie: Why 4?  And maybe make this a module-level constant?

bboreham: First and last addresses in a subnet are excluded, so 4 is the smallest size where any allocation will succeed.

ipam/allocator.go
```diff
@@ -197,29 +177,46 @@ func hasBeenCancelled(cancelChan <-chan bool) func() bool {
 
 // Actor client API
 
+func (alloc *Allocator) SetDefaultSubnet(subnet address.CIDR) error {
+	resultChan := make(chan error)
+	if subnet.Size() < MinSubnetSize {
+		return fmt.Errorf("Allocation range smaller than minimum size %d: %s", MinSubnetSize, subnet)
+	}
+	alloc.actionChan <- func() {
+		if !alloc.universe.Overlaps(subnet) {
```
rade: don't you want `Contains` here?

ipam/allocate.go
```diff
@@ -23,23 +24,34 @@ func (g *allocate) Try(alloc *Allocator) bool {
...
+	if addrs, found := alloc.owned[g.ident]; found {
+		for _, addr := range addrs {
+			if g.subnet.Contains(addr) {
+				g.resultChan <- allocateResult{addr, nil}
+				return true
+			}
+		}
+	}
+
+	if !alloc.universe.Overlaps(g.subnet) {
```
rade: don't you want `Contains` here?

ipam/allocator.go
```diff
@@ -264,6 +251,33 @@ func (alloc *Allocator) free(ident string) error {
 	return <-errChan
 }
 
+// Free (Sync) - release single IP address for container
+func (alloc *Allocator) Free(ident string, addr address.Address) error {
+	errChan := make(chan error)
+	alloc.actionChan <- func() {
+		addrs := alloc.owned[ident]
+		for i := range addrs {
+			if addrs[i] == addr {
```
rade: extract the value as part of the `range`.

ipam/allocator.go
```diff
@@ -573,8 +590,18 @@ func (alloc *Allocator) propose() {
...
 
-func (alloc *Allocator) sendRequest(dest router.PeerName, kind byte) {
-	msg := router.Concat([]byte{kind}, alloc.encode())
+func (alloc *Allocator) sendSpaceRequest(dest router.PeerName, subnet address.CIDR) {
+	buf := new(bytes.Buffer)
+	enc := gob.NewEncoder(buf)
+	if err := enc.Encode(subnet); err != nil {
+		panic(err)
+	}
+	msg := router.Concat([]byte{msgSpaceRequest}, buf.Bytes())
```
rade: because of this, the protocol version needs bumping.

ipam/testutils_test.go
```diff
@@ -163,7 +165,7 @@ func (alloc *Allocator) claimRingForTesting(allocs ...*Allocator) {
 func (alloc *Allocator) NumFreeAddresses() address.Offset {
 	resultChan := make(chan address.Offset)
 	alloc.actionChan <- func() {
-		resultChan <- alloc.space.NumFreeAddresses()
+		resultChan <- alloc.space.NumFreeAddressesInRange(alloc.defaultSubnet.Start+1, alloc.defaultSubnet.End()-1)
```
rade: `NumFreeAddresses` is misnamed. Should be called `NumFreeAddressesInDefaultSubnet`. Which is a bit of a mouthful, admittedly.

ipam/space/space_test.go
```diff
@@ -52,12 +52,12 @@ func TestLowlevel(t *testing.T) {
 
 	s := New()
 	wt.AssertEquals(t, s.NumFreeAddresses(), address.Offset(0))
-	ok, got := s.Allocate()
+	ok, got := s.Allocate(0, 1000)
```
rade: I think `Space.NumFreeAddresses` is now only used in tests and hence should be moved into this file.

prog/weaver/main.go
```diff
@@ -135,9 +138,9 @@ func main() {
 
 	var allocator *ipam.Allocator
 	if iprangeCIDR != "" {
-		allocator = createAllocator(router, apiPath, iprangeCIDR, determineQuorum(peerCount, peers))
+		allocator = createAllocator(router, apiPath, iprangeCIDR, defltSubnet, determineQuorum(peerCount, peers))
 	} else if peerCount > 0 {
-		log.Fatal("-initpeercount flag specified without -iprange")
+		log.Fatal("-initpeercount flag specified without -iprange or -ipuniverse")
```
rade: `-ipuniverse` does not exist.

prog/weaver/main.go
```diff
@@ -69,6 +71,7 @@ func main() {
 	flag.IntVar(&bufSzMB, "bufsz", 8, "capture buffer size in MB")
 	flag.StringVar(&httpAddr, "httpaddr", fmt.Sprintf(":%d", weave.HTTPPort), "address to bind HTTP interface to (disabled if blank, absolute path indicates unix domain socket)")
 	flag.StringVar(&iprangeCIDR, "iprange", "", "IP address range to allocate within, in CIDR notation")
+	flag.StringVar(&defltSubnet, "defaultsubnet", "", "default subnet for IP address allocation, in CIDR notation")
```
rade: That's a horrid name. It's not at all clear that this option pertains to ipam. arguably the same is true of `iprange`. `ipamuniverse` and `ipamnet`?

Descriptions: "IP address range reserved for automatic allocation." and "subnet of ipamuniverse to allocate within by default".

rade: also, the value of `ipamnet` should default to that of `ipamuniverse`. Makes sense, imo, and simplifies the logic in a few places.

test/130_expose_test.sh
```diff
@@ -7,7 +7,8 @@ C2=10.2.1.37
 C3=10.2.2.34
 C4=10.2.2.37
 EXP=10.2.2.101
-UNIVERSE=10.2.3.0/24
+UNIVERSE=10.2.0.0/16
+SUBNET=10.2.3.0/24
```
rade: I don't see the need to change anything in this test.

weave
```diff
@@ -98,7 +98,7 @@ validate_cidr() {
 collect_cidr_args() {
     CIDR_ARGS=""
     CIDR_COUNT=0
-    while is_cidr "$1" ; do
+    while is_cidr "$1" || [ "$1" = "--subnet" ] ; do
         CIDR_ARGS="$CIDR_ARGS $1"
         CIDR_COUNT=$((CIDR_COUNT + 1))
```
rade: `CIDR_COUNT` is now really rather misnamed. `CIDR_ARG_COUNT`?

prog/weaver/main.go
dpw: I consider 'universe' to be a poor name, and said so during my IPAM review.  It's even worse in the context of multi-subnet IPAM.  But while it is confined to identifiers in the code, at least there is the prospect that it will be replaced with a better name at some point.  Please let's not use it in a user-visible context.

weave
```diff
@@ -979,6 +1007,7 @@ case "$COMMAND" in
         shift $CIDR_COUNT
         [ $# -eq 1 ] || usage
         CONTAINER=$(container_id $1)
+        ipam_cidrs $CONTAINER $CIDR_ARGS
```
rade: huh? We allocate IP addresses when *detaching* a container? Surely the semantics for subnets here should be "remove (and free in IPAM) and IP address in the given subnet that belongs to the container".

It seems to me that there is actually a bug / missing feature on default here; surely a) detaching an allocated IP should free it, and b) zero-arg detach should detach any allocation from the default ipam subnet.

weave
```diff
@@ -1041,8 +1070,10 @@ case "$COMMAND" in
                 delete_iptables_rule nat WEAVE -s $CIDR ! -d $CIDR -j MASQUERADE
                 when_dns_running delete_dns 'weave:expose' $CIDR
             fi
+            # Tell IPAM we don't need this address, but ignore errors because
+            # the address might not have come from IPAM
+            http_call $CONTAINER_NAME $HTTP_PORT DELETE /ip/weave:expose/${CIDR%/*} >/dev/null 2>&1 || true
```
rade: hmm. this seems rather prone to generate alarming messages in the weave log. Can we do better?

rade:
I find the idea of asking a user to type `docker run -e WEAVE_CIDR="--subnet 10.2.1.0/24" -ti ubuntu` rather abhorrent.

Also, as @squaremo said, we need to account for using the default subnet alongside other addresses.

site/ipam.md
```diff
@@ -22,7 +26,11 @@ You can see which address was allocated with
     host1# weave ps $C
     a7aff7249393 7a:51:d1:09:21:78 10.2.3.1/24
 
-The `-iprange` parameter is given in [CIDR
+To use a subnet other than the default, use `--subnet`:
```
rade: "To request the allocation of an address from a subnet other than the default..."

prog/weaver/main.go
bboreham: Making the default subnet be the "whole universe" makes it impossible to have two subnets, which I thought was still a requirement for weaveDNS?

rade: The user would have to pick a subnet for DNS outside the ipam universe. Which is no different to what they have to do for any other manual allocation.

This changes with #743 and #763.

ipam/allocate.go
bboreham: Overlapped seemed more broadly useful than contained, as you might want both IPAM-allocated and manually-allocated addresses in the same subnet.  I.e. you could give IPAM 10.3.0.0/16 but run all containers in 10.0.0.0/8.
I agree that this way round is harder to grasp, initially.

ipam/allocator.go
bboreham: The code refers to some variant of `addrs[i]` three times, so to give one of them a different name seems more confusing to me.  I presume the compiler can figure out it's the same thing either way.

prog/weaver/main.go
bboreham: How about `-iprange` and `-ipsubnet`, with the latter defaulting to the former?  This would mean that anyone using the feature as shipped in v0.11 could continue unchanged.

Also, as noted elsewhere, anyone wanting to give some containers manually-assigned addresses in the same subnet as ipam-assigned addresses would use a default subnet _larger_ than iprange, so "subnet of xxx" is not strictly the right phrasing.

rade: Happy with `-iprange` and `-ipsubnet`.

So using an `ipsubnet` that is larger than `iprange` actually works? IPAM will only allocate in the intersection of the two? Needs documenting.

weave
bboreham: I'm fixing the bug that DELETE was never called, but it still seems easiest to go through the allocation process first, since it will return the previously allocated address when you do things in the right order, and will only temporarily allocate an unnecessary address when you do 'detach' with no preceding 'attach'.

The alternative is to have both POST and GET variants of `ipam_cidrs`, and handle the different error cases.

rade: fair enough. stick a comment in the code.

rade: also, please file a bug.

bboreham:
I believe I have addressed all review comments to date.

ipam/allocator.go
rade: I doubt the compiler figures that out. To me it looks confusing to use a single-return `range` when the very first expression in the block extracts the value.

ipam/space/space_test.go
```diff
@@ -19,6 +19,15 @@ func ip(s string) address.Address {
 	return addr
 }
 
+// Helper function to avoid editing dozens of lines of test
```
rade: I f I thought that was the case I wouldn't have suggested keeping the function.

It's useful because the alternative is having dozens of lines invoking a function with the same two args.

weave
```diff
@@ -25,19 +25,19 @@ CIDR_REGEXP="$IP_REGEXP/[0-9]{1,2}"
 usage() {
     echo "Usage:"
     echo "weave setup"
-    echo "weave launch       [-password <password>] [-nickname <nickname>] [-iprange <cidr>] <peer> ..."
+    echo "weave launch       [-password <password>] [-nickname <nickname>] [-iprange <cidr>] [-defaultsubnet <cidr>] <peer> ..."
```
rade: s/defaultsubnet/ipsubnet

weave
```diff
@@ -25,19 +25,19 @@ CIDR_REGEXP="$IP_REGEXP/[0-9]{1,2}"
...
+    echo "weave run          [--with-dns] [[--subnet] <cidr> ...] <docker run args> ..."
+    echo "weave start        [[--subnet] <cidr> ...] <container_id>"
+    echo "weave attach       [[--subnet] <cidr> ...] <container_id>"
+    echo "weave detach       [[--subnet] <cidr> ...] <container_id>"
     echo "weave dns-add      <ip_address> [<ip_address> ...] <container_id> [-h <fqdn>]"
     echo "weave dns-remove   <ip_address> [<ip_address> ...] <container_id> [-h <fqdn>]"
-    echo "weave expose       [<cidr> ...] [-h <fqdn>]"
-    echo "weave hide         [<cidr> ...]"
+    echo "weave expose       [[--subnet] <cidr> ...] [-h <fqdn>]"
+    echo "weave hide         [[--subnet] <cidr> ...]"
```
rade: this is not accurate; it implies that `--subnet` is specified *once*, in front of all `<cidr>`s.

Also, you don't mention the `default`.

I suggest introducing a new arg "type".

ipam/allocator.go
```diff
@@ -46,16 +45,15 @@ type operation interface {
...
-	prefixLen        int                        // network prefix length, e.g. 24 for a /24 network
-	ring             *ring.Ring                 // information on ranges owned by all peers
-	space            space.Space                // more detail on ranges owned by us
-	owned            map[string]address.Address // who owns what address, indexed by container-ID
-	nicknames        map[router.PeerName]string // so we can map nicknames for rmpeer
-	pendingAllocates []operation                // held until we get some free space
-	pendingClaims    []operation                // held until we know who owns the space
-	gossip           router.Gossip              // our link to the outside world for sending messages
+	defaultSubnet    address.CIDR
+	universe         address.CIDR                 // superset of all ranges
```
rade: swap these two. and add an explanation for `defaultSubnet`.

ipam/allocator_test.go
```diff
@@ -24,36 +24,49 @@ func TestAllocFree(t *testing.T) {
...
+	_, cidr1, _ := address.ParseCIDR(subnet1)
+	_, cidr2, _ := address.ParseCIDR(subnet2)
 
 	alloc.claimRingForTesting()
-	addr1, _ := alloc.Allocate(container1, nil)
+	addr1, _ := alloc.Allocate(container1, cidr1, nil)
 	wt.AssertEqualString(t, addr1.String(), testAddr1, "address")
 
+	addr2, err := alloc.Allocate(container1, cidr2, nil)
+	wt.AssertNoErr(t, err)
```
rade: Why do we need this here, and not above?

ipam/allocator_test.go
```diff
@@ -24,36 +24,49 @@ func TestAllocFree(t *testing.T) {
...
+	wt.AssertEqualString(t, addr2.String(), testAddr2, "address")
+	CheckAllExpectedMessagesSent(alloc)
```
rade: as somebody not familiar with the internals of ipam this suprised me. why is it needed? and why not above?

ipam/allocator_test.go
```diff
@@ -80,7 +93,7 @@ func TestBootstrap(t *testing.T) {
 	ExpectBroadcastMessage(alloc1, nil) // alloc1 will try to form consensus
 	done := make(chan bool)
 	go func() {
-		alloc1.Allocate("somecontainer", nil)
+		alloc1.Allocate("somecontainer", alloc1.defaultSubnet, nil)
```
rade: I suggest we introduce a helper so this becomes `alloc1.alloc("somecontainer")`

ipam/http_test.go
```diff
@@ -59,45 +59,59 @@ func listenHTTP(alloc *Allocator) int {
...
+	cidr1a := HTTPPost(t, allocURL(port, testCIDR1, containerID))
+	wt.AssertEqualString(t, cidr1a, testAddr1, "address")
+
+	cidr2a := HTTPPost(t, allocURL(port, testCIDR2, containerID))
+	wt.AssertEqualString(t, cidr2a, testAddr2, "address")
+
+	check := HTTPGet(t, allocURL(port, testCIDR1, containerID))
+	wt.AssertEqualString(t, check, cidr1a, "address")
+	check = HTTPGet(t, allocURL(port, testCIDR2, containerID))
+	wt.AssertEqualString(t, check, cidr2a, "address")
```
rade: would be nice to have a comment above that whole block, to explain what it's doing.

ipam/ring/ring.go
```diff
@@ -409,15 +409,21 @@ func (r *Ring) ReportFree(freespace map[address.Address]address.Offset) {
...
 		totalSpacePerPeer = make(map[router.PeerName]address.Offset) // Compute total free space per peer
 	)
 
 	// iterate through tokens
-	for _, entry := range r.Entries {
+	for i, entry := range r.Entries {
+		// Ignore entries that don't span the range we want
+		if i+1 < len(r.Entries) && r.Entries.entry(i+1).Token-1 < start {
+			continue
+		} else if entry.Token >= end {
```
rade: drop the `else`

ipam/space/space.go
```diff
@@ -96,12 +100,19 @@ func (s *Space) Free(addr address.Address) error {
...
 
 	for i := 0; i < len(s.free); i += 2 {
-		size := address.Subtract(s.free[i+1], s.free[i])
+		start, end := s.free[i], s.free[i+1]
+		if start < rangeStart {
+			start = rangeStart
+		}
+		if end > rangeEnd {
+			end = rangeEnd
+		}
```
rade: those 7 lines also appear in `NumFreeAddressesInRange`. extract.

rade: and it looks like you are missing the `if start >= end { continue }` from `NumFreeAddressesInRange`.

rade: And why does this not have a `break` when the start of the free space is beyond the end of the range? You *do* have that in `Allocate`.

`NumFreeAddressesInRange` should have that too.

I am hoping that in the end all three functions can use some common helper.

bboreham: Adding either `break` or `continue` before computing the size of this portion would be wrong.  Adding the tests again after the computation would be correct, but long-winded.

ipam/space/space_test.go
```diff
@@ -52,12 +62,12 @@ func TestLowlevel(t *testing.T) {
 
 	s := New()
 	wt.AssertEquals(t, s.NumFreeAddresses(), address.Offset(0))
-	ok, got := s.Allocate()
+	ok, got := s.Allocate(0, 1000)
```
rade: add a helper so this can become `s.alloc()`.

ipam/space/space_test.go
```diff
@@ -67,40 +77,46 @@ func TestLowlevel(t *testing.T) {
 	wt.AssertErrorInterface(t, s.Free(0), (*error)(nil), "free not allocated")
 	wt.AssertErrorInterface(t, s.Free(100), (*error)(nil), "double free")
 
-	start, size, ok := s.Donate()
+	start, size, ok := s.Donate(0, 1000)
```
rade: add a helper to this can become `s.donate()`.

ipam/space/space_test.go
```diff
@@ -128,28 +144,28 @@ func TestSpaceFree(t *testing.T) {
 	space := makeSpace(ip(testAddr1), 20)
 
 	// Check we are prepared to give up the entire space
-	pos, size := space.biggestFreeRange()
+	pos, size := space.biggestFreeRange(ip(testAddr1), ip(testAddry))
```
rade: add a helper so this becomes `space.biggestFree()`.

ipam/space/space_test.go
```diff
@@ -128,28 +144,28 @@ func TestSpaceFree(t *testing.T) {
...
 	wt.AssertTrue(t, space.free[pos] == ip(testAddr1) && size == 20, "Wrong space")
 
 	for i := 0; i < 20; i++ {
-		ok, _ := space.Allocate()
+		ok, _ := space.Allocate(ip(testAddr1), ip(testAddry))
```
rade: add a helper.

ipam/space/space_test.go
```diff
@@ -128,28 +144,28 @@ func TestSpaceFree(t *testing.T) {
...
 		wt.AssertTrue(t, ok, "Failed to get address")
 	}
 
 	// Check we are full
-	ok, _ := space.Allocate()
+	ok, _ := space.Allocate(ip(testAddr1), ip(testAddry))
 	wt.AssertTrue(t, !ok, "Should have failed to get address")
-	_, size, ok = space.Donate()
+	_, size, ok = space.Donate(ip(testAddr1), ip(testAddry))
```
rade: add a helper.

ipam/space/space_test.go
```diff
@@ -128,28 +144,28 @@ func TestSpaceFree(t *testing.T) {
...
 	wt.AssertTrue(t, size == 0, "Wrong space")
 
 	// Free in the middle
 	wt.AssertSuccess(t, space.Free(ip("10.0.3.13")))
-	pos, size = space.biggestFreeRange()
+	pos, size = space.biggestFreeRange(ip(testAddr1), ip(testAddry))
```
rade: add a helper.

ipam/allocator.go
rade: actually, I am starting to think that `Allocator` should have no knowledge of the defaultSubnet. It never uses it. And neither should it.

ipam/space/space.go
rade: I don't understand. Why would adding a `break` when the start of the free space is beyond the end of the range be wrong?

rade: and, `start >= end` means the range is entirely beyond the free space chunk. So why would a `continue` be wrong?

ipam/allocator.go
```diff
@@ -46,16 +45,15 @@ type operation interface {
...
+	ring             *ring.Ring                   // information on ranges owned by all peers
+	space            space.Space                  // more detail on ranges owned by us
+	owned            map[string][]address.Address // who owns what address, indexed by container-ID
+	nicknames        map[router.PeerName]string   // so we can map nicknames for rmpeer
+	pendingAllocates []operation                  // held until we get some free space
+	pendingClaims    []operation                  // held until we know who owns the space
+	gossip           router.Gossip                // our link to the outside world for sending messages
 	paxos            *paxos.Node
```
rade: adding that blank line before the `paxos` line has confused the diff. I suggest you add it on master.

site/ipam.md
```diff
@@ -22,7 +22,35 @@ You can see which address was allocated with
     host1# weave ps $C
     a7aff7249393 7a:51:d1:09:21:78 10.2.3.1/24
 
-The `-iprange` parameter is given in [CIDR
+You may wish to run containers on different subnets, as described
+[here](features.html#application-isolation). To set a default subnet,
```
rade: make "run containers on different subnets" the link and ditch the "as described here".

weave
```diff
@@ -988,12 +1023,12 @@ case "$COMMAND" in
...
         collect_cidr_args "$@"
-        shift $CIDR_COUNT
+        shift $CIDR_ARG_COUNT
         CONTAINER=$(docker run $DNS_ARGS -d "$@")
         create_bridge
-        get_ipam_cidr $CONTAINER
-        with_container_netns_or_die $CONTAINER attach $CIDR_ARGS
-        when_dns_running with_container_fqdn $CONTAINER put_dns_fqdn $CIDR_ARGS
+        ipam_cidrs $CONTAINER $CIDR_ARGS
+        with_container_netns_or_die $CONTAINER attach $CIDR_ARGS $IPAM_CIDRS
```
rade: A lot of noise is added by having to write `$CIDR_ARGS $IPAM_CIDRS` in so many places. That would go away of `$CIDR_ARGS` included `$IPAM_CIDRS`. Then `$IPAM_CIDRS` would only appear in a couple of places, on its own.

Also, it's rather misleading that `ipam_cidrs` rewrites `CIDR_ARGS`, which was already set by `collect_cidr_args`. Especially since the type is different, i.e. it's not really args anymore, but just cidrs. So I suggest `ipam_cidrs` sets `ALL_CIDRS` and `IPAM_CIDRS`.

ipam/allocator.go
bboreham: There is no blank line before `paxos`.  The diff is larger than you might expect because adding `[]` to the type of `owned` pushed all the comments to the right.

See with whitespace ignored: https://github.com/weaveworks/weave/compare/master...bboreham:675-ipam-subnets?w=1#diff-4e74ab9efea46a74c7a5519eca56df7cR45

ipam/allocator_test.go
bboreham: It isn't needed.  It doesn't hurt to check we sent all expected messages, but since we didn't expect any messages it's a no-op.

rade: remove it then. Or, alternatively, come up with a rule for when to perform that check, and apply that consistently.

ipam/space/space.go
```diff
@@ -96,26 +103,27 @@ func (s *Space) Free(addr address.Address) error {
...
 }
 
-func (s *Space) Donate() (address.Address, address.Offset, bool) {
+func (s *Space) Donate(rangeStart, rangeEnd address.Address) (address.Address, address.Offset, bool) {
 	if len(s.free) == 0 {
 		return 0, 0, false
 	}
 
-	pos, biggest := s.biggestFreeRange()
+	pos, biggest := s.biggestFreeRange(rangeStart, rangeEnd)
```
rade: The code below this looks wrong. We donate a chunk from the end of the biggest free range, but because we are using `pos` this isn't clipped to the requested range.

This would be avoided if `biggestFreeRange` returned the start and end, instead of pos and size.

That would also simplify the new `Space.walkFree`, since we'd no longer need to pass the index to the invoked function.

site/ipam.md
```diff
@@ -117,8 +145,7 @@ reports on the current status of the weave router and IP allocator:
 ````
 weave router git-8f675f15c0b5
 ...
-Allocator subnet 10.2.1.0/24
-  Free IPs: ~98.0%, 62 local, ~189 remote
+Allocator universe 10.2.0.0/16 (default subnet 10.2.1.0/24)
```
rade: needs updating to reflect new position of default subnet in output.

ipam/space/space.go
bboreham: (Although GitHub is helpfully retaining many comments here, all of them relate to older versions of the code)

What I meant was, it takes four if-statements to get the correct answer while applying short-circuits, so I had omitted the short-circuits in one case.  I have extracted the logic into a helper function so being long-winded isn't such a problem.

prog/weaver/main.go
```diff
@@ -194,18 +198,35 @@ func logFrameFunc(debug bool) weave.LogFrameFunc {
...
+	ipRange := parseAndCheckCIDR(ipRangeStr)
+	allocator := ipam.NewAllocator(router.Ourself.Peer.Name, router.Ourself.Peer.UID, router.Ourself.Peer.NickName, ipRange, quorum)
+	defaultSubnet := ipRange
+	if defaultSubnetStr != "" {
+		defaultSubnet = parseAndCheckCIDR(defaultSubnetStr)
+		if !ipRange.Overlaps(defaultSubnet) {
+			log.Fatalf("Default subnet %s out of bounds: %s", defaultSubnet, ipRange)
+		}
+	}
+
```
rade: I suggest switching the check and allocator creation.

ipam/space/space.go
bboreham: I changed biggestFreeRange() as suggested

ipam/space/space_test.go
bboreham: can you expand a bit on how you see that working?  it should figure out the minimum and maximum addresses?

rade:
don't worry about the tests. They were hideous before and are only marginally more hideous now. I will fix them myself.

docs/ipam.md
```diff
@@ -46,18 +46,23 @@ freed.
 
 ### The Allocation Process
 
+Each allocation request either specifies a subnet to allocate within,
+or if no subnet is specified this is taken as a request to allocate in
+a pre-defined default subnet.
+
```
rade: out-of-date

ipam/allocator.go
```diff
@@ -573,8 +570,18 @@ func (alloc *Allocator) propose() {
...
+	enc := gob.NewEncoder(buf)
+	if err := enc.Encode(subnet); err != nil {
+		panic(err)
+	}
+	msg := router.Concat([]byte{msgSpaceRequest}, buf.Bytes())
+	alloc.gossip.GossipUnicast(dest, msg)
+}
+
+func (alloc *Allocator) sendRingUpdate(dest router.PeerName) {
+	msg := router.Concat([]byte{msgRingUpdate}, alloc.encode())
```
rade: inline these two functions.

ipam/space/space.go
```diff
@@ -37,14 +37,40 @@ func (s *Space) Clear() {
...
+			break
+		}
+		// Restrict this block of free space to be in range
+		if start < rangeStart {
+			start = rangeStart
+		}
+		if end > rangeEnd {
+			end = rangeEnd
+		}
+		if start >= end { // no space left in range
```
rade: The comment is misleading. Everywhere else in this function `range` refers to the passed in range. Also, it's not that there is no space left, but that the current free space chunk(?) is completely before the range.

I reckon it would be more obvious to test for `end <= rangeStart` earlier.

bboreham: If `rangeStart==rangeEnd` then your assumption that `start>=end` implies 'current free space chunk is completely before the range' falls down.  It seems more bullet-proof to test this last, then it's obvious that, for instance, `Allocate` can never receive an empty range.

weave
```diff
@@ -734,17 +740,47 @@ ipam_claim() {
...
-# one from IPAM. $1 is the full container id.
-get_ipam_cidr() {
-    if [ -n "$CIDR_ARGS" ] ; then
+# Call IPAM as necessary to allocate addresses
+# $1 is the full container id and following args are previously parsed CIDR_ARGS.
+# Returns ALL_CIDRS and IPAM_CIDRS
+ipam_cidrs() {
+    CONTAINER_ID="$1"
+    shift 1
+    if [ $# -gt 0 ] ; then
```
rade: I suggest inverting this, putting the (shorter) 0-arg code into the `then` block.

ipam/space/space.go
rade: Is `rangeStart == rangeEnd` permissible? I thought `rangeEnd` was the first address outside the range.

bboreham: It would mean an empty range: not a useful thing to ask for, but I don't think we can assume it never happens without some check somewhere.  I'll write that version and you can see which you prefer.

ipam/address/address.go
```diff
@@ -20,6 +27,32 @@ func ParseIP(s string) (Address, error) {
...
+	} else if ipnet.IP.To4() == nil {
+		return 0, CIDR{}, &net.ParseError{Type: "Non-IPv4 address not supported", Text: s}
+	} else {
+		prefixLen, _ := ipnet.Mask.Size()
+		return FromIP4(ip), CIDR{Start: FromIP4(ipnet.IP), PrefixLen: prefixLen}, nil
+	}
+}
+
+func (cidr CIDR) Size() Offset { return 1 << uint(32-cidr.PrefixLen) }
+func (cidr CIDR) End() Address { return Add(cidr.Start, cidr.Size()) }
```
rade: So this an exclusive end. not obvious.

rade:
We have an `address.Range` type. Let's *always* use that.

And I reckon `Allocator` should know nothing about CIDRs; it should operate on Ranges. CIDRs are a UI convenience, so they should be confined to UI code (i.e. main.go and the http handlers).

ipam/address/address.go
```diff
@@ -13,6 +15,17 @@ type Range struct {
 	Start, End Address // [Start, End); Start <= End
 }
 
+func MakeRange(start Address, size Offset) Range { return Range{Start: start, End: Add(start, size)} }
```
rade: isn't `NewRange` the idiomatic name for this?

ipam/address/address.go
```diff
@@ -20,6 +33,31 @@ func ParseIP(s string) (Address, error) {
...
+		prefixLen, _ := ipnet.Mask.Size()
+		return FromIP4(ip), CIDR{Start: FromIP4(ipnet.IP), PrefixLen: prefixLen}, nil
+	}
+}
+
+func (cidr CIDR) Size() Offset { return 1 << uint(32-cidr.PrefixLen) }
+
+// End is exclusive, same as in Range
+func (cidr CIDR) End() Address { return Add(cidr.Start, cidr.Size()) }
+func (cidr CIDR) Range() Range { return Range{Start: cidr.Start, End: Add(cidr.Start, cidr.Size())} }
```
rade: This should call `MakeRange(cidr.Start, cidr.Size())` instead.

ipam/address/address.go
```diff
@@ -20,6 +33,31 @@ func ParseIP(s string) (Address, error) {
...
+func (cidr CIDR) HostRange() Range {
+	// Respect RFC1122 exclusions of first and last addresses
+	return Range{Start: cidr.Start + 1, End: cidr.End() - 1}
```
rade: `MakeRange(cidr.Start+1, cidr.Size()-2)`? Somewhat less obvious, but respects `Range` encapsulation and also means we can get rid of `CIDR.End()`.

ipam/address/address.go
rade: also, i suggest making this multi-line, for better visual similarity with `HostRange` just below.

ipam/allocator.go
```diff
@@ -199,22 +178,23 @@ func hasBeenCancelled(cancelChan <-chan bool) func() bool {
 
 // Allocate (Sync) - get IP address for container with given name
 // if there isn't any space we block indefinitely
-func (alloc *Allocator) Allocate(ident string, cancelChan <-chan bool) (address.Address, error) {
+func (alloc *Allocator) Allocate(ident string, r address.Range, cancelChan <-chan bool) (address.Address, error) {
 	resultChan := make(chan allocateResult)
-	op := &allocate{resultChan: resultChan, ident: ident,
+	op := &allocate{r: r, resultChan: resultChan, ident: ident,
```
rade: Is there any rationale to the arg order here? It does match neither the function parameter order nor the struct field order.

ipam/space/space.go
```diff
@@ -96,35 +112,31 @@ func (s *Space) Free(addr address.Address) error {
...
-func (s *Space) biggestFreeRange() (int, address.Offset) {
-	pos := -1
-	biggest := address.Offset(0)
-
-	for i := 0; i < len(s.free); i += 2 {
-		size := address.Subtract(s.free[i+1], s.free[i])
-		if size >= biggest {
-			pos = i
-			biggest = size
+func (s *Space) biggestFreeRange(r address.Range) (biggest address.Range, biggestSize address.Offset) {
```
rade: There's really no need to return `biggestSize` here, since it's easily obtained at the call site with `biggest.Size()`

weave
```diff
@@ -48,9 +48,11 @@ usage() {
...
     echo "weave rmpeer       <peer_id>"
     echo
-    echo "where <peer>    is of the form <ip_address_or_fqdn>[:<port>], and"
-    echo "      <cidr>    is of the form <ip_address>/<routing_prefix_length>"
-    echo "      <peer_id> is a <nickname> or weave internal peer ID"
+    echo "where <peer>         is of the form <ip_address_or_fqdn>[:<port>], and"
+    echo "      <cidr>         is of the form <ip_address>/<routing_prefix_length>"
+    echo "      <peer_id>      is a <nickname> or weave internal peer ID"
+    echo "      <address_spec> is either a single address in <cidr> format or"
+    echo "         --subnet <cidr> or --subnet default to fetch from the allocator"
```
rade: Following my [commit just now](https://github.com/weaveworks/weave/commit/2d701972211129ed6246d9d7b523a20dfb805e2d) this should be `= <cidr> | --subnet <cidr> | --subnet default`

No need to get more descriptive.

rade:
Can we come up with any tests for #861 and #862 that aren't gross? And ideally for `hide` too.

rade:
Please make the linter happy.

ipam/allocator.go
bboreham: That is a disimprovement, in my opinion.  It separates the encoding and decoding of spaceRequest into two files, and sendRingUpdate has to be a function anyway because it is deferred.

rade: ok

rade:
Those linter changes look wrong; we have configured the linter not to complain about comments.

test/500_weave_multi_cidr_test.sh
```diff
@@ -42,36 +42,41 @@ assert_bridge_cidrs() {
...
 # Run container with three cidrs
-CID=$(start_container $HOST1 10.2.1.1/24 10.2.2.1/24 10.2.3.1/24 --name=multicidr -h $NAME | cut -b 1-12)
+CID=$(start_container $HOST1 10.2.1.1/24 10.2.2.1/24 --subnet 10.2.3.0/24 --name=multicidr -h $NAME | cut -b 1-12)
 assert_container_cidrs $HOST1 $CID 10.2.1.1/24 10.2.2.1/24 10.2.3.1/24
 assert_zone_records $HOST1 $CID $NAME. 10.2.1.1 10.2.2.1 10.2.3.1
 
 # Remove two of them
-weave_on $HOST1 detach 10.2.1.1/24 10.2.3.1/24 $CID
+weave_on $HOST1 detach 10.2.1.1/24  --subnet 10.2.3.0/24 $CID
 assert_container_cidrs $HOST1 $CID 10.2.2.1/24
```
rade: hmm. you really want `assert_container_cidrs` to check that *exactly* those addresses are present, don't you? Which it doesn't.

test/500_weave_multi_cidr_test.sh
```diff
@@ -42,36 +42,41 @@ assert_bridge_cidrs() {
...
 # Hide some cidrs
-weave_on $HOST1 hide 10.2.1.2/24 10.2.3.2/24
+weave_on $HOST1 hide 10.2.1.2/24  --subnet 10.2.3.0/24
 assert_bridge_cidrs $HOST1 weave 10.2.2.2/24
 
+# Now detach and run another container to check we have released IPs in IPAM
+weave_on $HOST1 detach $CID
+CID2=$(start_container $HOST1 --subnet 10.2.3.0/24)
+assert_container_cidrs $HOST1 $CID2 10.2.3.1/24
+
```
rade: So the big underlying assumption in all these tests here is that IPAM will give us the lowest available address. That deserves a comment.

ipam/allocator.go
```diff
@@ -470,20 +477,14 @@ func (alloc *Allocator) actorLoop(actionChan <-chan func()) {
...
 		}
 	} else {
-		localFreeSpace := alloc.space.NumFreeAddresses()
-		remoteFreeSpace := alloc.ring.TotalRemoteFree()
-		percentFree := 100 * float64(localFreeSpace+remoteFreeSpace) / float64(alloc.subnetSize)
-		fmt.Fprintf(&buf, "  Free IPs: ~%.1f%%, %d local, ~%d remote\n",
-			percentFree, localFreeSpace, remoteFreeSpace)
-
-		fmt.Fprint(&buf, "Owned Ranges:")
+		fmt.Fprint(&buf, "\nOwned Ranges:")
```
rade: Docs need updating since we no longer show the "Free" info.

Also, I believe `Ring.TotalRemoteFree()` is now unused.

ipam/allocator.go
```diff
@@ -46,16 +45,14 @@ type operation interface {
...
-	space            space.Space                // more detail on ranges owned by us
-	owned            map[string]address.Address // who owns what address, indexed by container-ID
-	nicknames        map[router.PeerName]string // so we can map nicknames for rmpeer
-	pendingAllocates []operation                // held until we get some free space
-	pendingClaims    []operation                // held until we know who owns the space
-	gossip           router.Gossip              // our link to the outside world for sending messages
+	universe         address.Range                // superset of all ranges
+	ring             *ring.Ring                   // information on ranges owned by all peers
+	space            space.Space                  // more detail on ranges owned by us
+	owned            map[string][]address.Address // who owns what address, indexed by container-ID
```
rade: "address*es*"

ipam/allocator.go
```diff
@@ -264,6 +244,27 @@ func (alloc *Allocator) free(ident string) error {
...
+// Free (Sync) - release single IP address for container
+func (alloc *Allocator) Free(ident string, addrToFree address.Address) error {
+	errChan := make(chan error)
+	alloc.actionChan <- func() {
+		addrs := alloc.owned[ident]
+		for i, ownedAddr := range addrs {
+			if ownedAddr == addrToFree {
+				alloc.debugln("Freed", addrToFree, "for", ident)
+				addrs = append(addrs[:i], addrs[i+1:]...)
+				alloc.owned[ident] = addrs
```
rade: Should this delete the entry when `len(addrs) == 0`? (or `addr == nil`; I can never remember whether these two are equivalent in go).

ipam/allocator.go
```diff
@@ -680,13 +690,15 @@ func (alloc *Allocator) reportFreeSpace() {
 // Owned addresses
 
 func (alloc *Allocator) addOwned(ident string, addr address.Address) {
-	alloc.owned[ident] = addr
+	alloc.owned[ident] = append(alloc.owned[ident], addr)
```
rade: We are assuming that `addr` isn't already in the list here, otherwise we end up with duplicate entries which `Free` won't remove.

This is guaranteed by the calling context, but it took me a while to figure out. Not sure what to do about that; add a note perhaps, "NB: addr must not be owned by ident already"?

prog/weaver/main.go
```diff
@@ -240,6 +261,7 @@ func handleHTTP(router *weave.Router, httpAddr string, allocator *ipam.Allocator
 		fmt.Fprintln(w, router.Status())
 		if allocator != nil {
 			fmt.Fprintln(w, allocator.String())
+			fmt.Fprintln(w, "Allocator default subnet:", defaultSubnet)
```
rade: that also needs adding to the example `weave status` output in the ipam troubleshooting docs.

rade: Ah, you've done that already. ignore me.

ipam/allocator.go
rade: ah, you've done the first bit already. apologies for missing that.

site/ipam.md
```diff
@@ -22,7 +22,35 @@ You can see which address was allocated with
...
+addresses, and have the containers communicate with each other. For
+example, if you say:
+
+    host1# weave launch -iprange 10.9.0.0/17 -ipsubnet 10.9.0.0/16
+
+then you can run all containers in the 10.9.0.0/16 subnet, with IPAM
+using the lower half and leaving the upper half free for manual
+allocation.
+
+The subnet parameters are written in [CIDR
```
rade: @squaremo happy with the above?

rade:
Nearly there. Is there any squashing that can be done w/o much effort?

site/ipam.md
squaremo: Yes, thank you. Perhaps it should mention that you can mix `--subnet` with bare (static) addresses.

rade: right. Though I have to say that to me it's not at all obvious which is which when looking at something like

    weave run 10.2.0.1/24 --subnet 10.2.1.0/24 10.2.2.1/24

We could use a prefix instead, i.e. `subnet:10.2.1.0/24`

rade: `weave run ip:10.2.0.1/24 net:10.2.1.0/24 ip:10.2.2.1/24 ...`

with the `ip:` being optional.

ipam/allocator.go
bboreham: The free-space summaries could be brought back when IPAM tracks subnets in use, e.g. for #676.

However that would require something like `RemoteFree(Range)` so I'll remove `TotalRemoteFree()`

bboreham:
I believe everything up to 'ip:' prefix has been addressed, and some squashing done.

weave
```diff
@@ -1021,32 +1057,37 @@ case "$COMMAND" in
         ;;
     start)
         collect_cidr_args "$@"
-        shift $CIDR_COUNT
+        shift $CIDR_ARG_COUNT
         [ $# -eq 1 ] || usage
-        CONTAINER=$(docker start $1)
+        docker start $1 > /dev/null
```
rade: This conflicts with #884, which has a better fix.

bboreham:
You made a single-word comment "addresses", which I took as a request to rename the variable.

rade:
Terribly sorry about that; I should have been clearer. I placed an emphasis on the 'es', i.e. I wanted the comment to use the plural.

weave
bboreham: I'm unclear which part you think is better.  Returning the ID passed in seems worse, to me.

rade: we are not returning the ID passed in. We are returning whatever `docker start` returns.

bboreham:
Have reverted 'owned'->'addresses' change and squashed a couple more commits.

PR #828, opened 2015-06-02
no timeout on wait for status

### LINKED ISSUE #666: wait_for_status can fail on busy machine
I suppose there should be some option around this, either to disable timeout or make it much longer then the default. I have just seen it happening on a Kubernetes node where there was at least about a dozen containers running and I needed to start weave after it exited due to a Docker crash. The containers were all running already, just needed to start weaver back up, the crash is not exactly related to this.

rade:
Fixes #666.

### DIFF
--- a/weave
+++ b/weave
@@ -554,32 +554,22 @@ http_call() {
     http_call_ip $CONTAINER_IP "$@"
 }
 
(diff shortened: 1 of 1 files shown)


paulbellamy:
LGTM, but needs rebase.

PR #830, opened 2015-06-02
kill -0 $CONTAINER_PID does not work inside weaveexec container

### LINKED ISSUE #829: `weave ...` network configuration errors swallowed and misreported as container deaths
`weave launch/run/etc` do not report network configuration errors as intended, instead incorrectly reporting that the container has died.

To reproduce apply
````patch
--- a/weave
+++ b/weave
@@ -392,7 +392,7 @@ ask_version() {
 ######################################################################
 
 launch() {
-    if ! netnsenter ip link show eth0 >/dev/null ; then
+    if ! netnsenter ip link show eth1 >/dev/null ; then
         echo "Perhaps you are running the docker daemon with container networki
         return 1
     fi
````
and then
````
$ weave launch
Container 473bff6d84a86d1ab65b46f92e12d8ccb7e4ebdecfac4776a0b1b0e45e2fc471 died
````
whereas the expected result in this case was
````
Failure during network configuration for container 473bff6d84a86d1ab65b46f92e12d8ccb7e4ebdecfac4776a0b1b0e45e2fc471:
Device "eth1" does not exist.
Perhaps you are running the docker daemon with container networking disabled (-b=none).
````

### PR DESCRIPTION
**Pull request by rade**
since $CONTAINER_PID is in a different process namespace.

Fixes #829.

### DIFF
--- a/weave
+++ b/weave
@@ -335,7 +335,7 @@ with_container_netns() {
     shift 1
     if ! "$@" >$IP_TMPOUT 2>$IP_TMPERR ; then
         STATUS=1
-        if ! kill -0 "$CONTAINER_PID" 2>/dev/null ; then
+        if [ ! -d $PROCFS/$CONTAINER_PID ] ; then
             echo "Container $CONTAINER died" >&2
         else
             echo "Failure during network configuration for container $CONTAINER:" >&2

### CODE REVIEW

File: weave [hunk 1]
```diff
@@ -335,7 +335,7 @@ with_container_netns () {
     shift 1
     if ! "$@" >$IP_TMPOUT 2>$IP_TMPERR ; then
         STATUS=1
-        if ! kill -0 "$CONTAINER_PID" 2>/dev/null ; then
+        if [ ! -d /$PROCFS/$CONTAINER_PID ] ; then
```
paulbellamy: No leading slash needed.

PR #831, opened 2015-06-02
kill container on network config error

### LINKED ISSUE #192: failure during network configuration should not leave container running
When `weave launch` fails to configure the networking on the weave container, it should remove the container since it's odd to see the command fail and subsequently (e.g. a subsequent `weave launch`) be told that weave is in fact running.

Same goes for `weave launch-dns` and `weave run`.

Possible downsides...
1. removing the container removes its logs. IMO that is ok; whatever is in the container logs is unlikely to shed light on why network configuration failed
2. we would be interrupting container processes that perform a bunch of work, and indeed even function correctly, even in the absence of having a configured network. IMO that too is ok; if you launch a container with `weave run` you are intending for it to have access to the weave network, so the fact that it might kinda work w/o weave is irrelevant.

rade:
Closes #192.

### DIFF
--- a/weave
+++ b/weave
@@ -349,6 +349,13 @@ with_container_netns() {
     return $STATUS
 }
 
+with_container_netns_or_die() {
+    if ! with_container_netns "$@" >/dev/null ; then
+        docker kill $1 2>/dev/null || true
+        exit 1
+    fi
+}
+
 # Execute arguments as a command within the network namespace of $CONTAINER_PID
 netnsenter() {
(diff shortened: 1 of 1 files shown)


paulbellamy:
With this change, should the proxy still kill the container if `weave attach` fails? Seems a bit belt-and-braces, but could catch other errors.

rade:
> should the proxy still kill the container if `weave attach` fails?

You may want to read the code and/or commit message ;)

paulbellamy:
Ah `_or_die` is not used in attach.

PR #832, opened 2015-06-03
fix connection retry delay calculation logic

### LINKED ISSUE #412: Weave can re-try connections too quickly
Example, from weave logs, where the next attempt occurs 80ms after the previous attempt:

````
20:59:34.196292 ->[104.236.19.13:59829] attempting connection
20:59:34.241031 ->[104.236.19.13:59829] error during connection attempt: dial tcp4 104.236.19.13:59829: connection refused
20:59:34.277109 ->[104.236.19.13:59829] attempting connection
````

This behaviour can generate a lot of uninteresting log lines, making troubleshooting difficult.

From inspection of connection_maker.go, the time intervals are randomized, so could be very tiny.  Suggest we add a minimum delay of a few seconds to the next interval in `tryImmediately()`.

### PR DESCRIPTION
**Pull request by rade**
Fixes #412.

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -73,7 +73,7 @@ func (cm *ConnectionMaker) InitiateConnections(peers []string, replace bool) []e
 			cm.directPeers[peer] = addr
 			// curtail any existing reconnect interval
 			if target, found := cm.targets[addr.String()]; found {
-				target.tryAfter, target.tryInterval = tryImmediately()
+				target.tryNow()
 			}
 		}
 		return true
@@ -95,7 +95,7 @@ func (cm *ConnectionMaker) ConnectionTerminated(address string, err error) {
 		if target, found := cm.targets[address]; found {
 			target.attempting = false
 			target.lastError = err
-			target.tryAfter, target.tryInterval = tryAfter(target.tryInterval)
+			target.retry()
 		}
 		return true
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rade** [comment]
I have no idea why that cosmetic change is showing up here, since it's already on master. I think I hadn't pushed it at the time the PR was created.

File: weave [hunk 1]
```diff
@@ -850,13 +850,11 @@ case "$COMMAND" in
             -p $PORT:$CONTAINER_PORT/tcp -p $PORT:$CONTAINER_PORT/udp -e WEAVE_PASSWORD -v /var/run/docker.sock:/var/run/docker.sock \
             $WEAVE_DOCKER_ARGS $IMAGE -iface $CONTAINER_IFNAME -port $CONTAINER_PORT -name "$PEERNAME" -nickname "$(hostname)" $IPRANGE "$@")
         with_container_netns $CONTAINER launch >/dev/null
-
         wait_for_status $CONTAINER_NAME $HTTP_PORT
```
bboreham: Change in weave script seems unrelated to rest of PR.

router/connection_maker.go
```diff
@@ -304,14 +304,16 @@ func (cm *ConnectionMaker) attemptConnection(address string, acceptNewPeer bool)
 }
 
 func tryImmediately() (time.Time, time.Duration) {
-	interval := time.Duration(rand.Int63n(int64(InitialInterval)))
-	return time.Now(), interval
+	return time.Now(), InitialInterval
 }
 
+// The delay at the nth retry is a random value in the range
+// [i-i/2,i+i/2], where i = InitialInterval * 1.5^(n-1).
```
bboreham: This comment is only true indirectly, because of the way the function is called.  Maybe `tryAfter` and `tryImmediately` should be methods on `Target*`, and manipulate `tryInterval` directly?

PR #848, opened 2015-06-04
add proxy-env and proxy-config which query weaveproxy for port

### LINKED ISSUE #753: [proxy] make `weave launch-proxy` return DOCKER_HOST settings
This would be convenient:

````
DOCKER_HOST=$(weave launch-proxy)
docker run ...
````

### PR DESCRIPTION
**Pull request by paulbellamy**
If the overall approach looks fine, I want to modify the tests to use this.

Closes #753

### DIFF
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -40,12 +40,17 @@ interfaces. This can be adjusted with the `-H` argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:9999
 
-All docker commands can be run via the proxy, so it is safe to
-globally adjust your `DOCKER_HOST` to point at the proxy, e.g.
+All docker commands can be run via the proxy, so it is safe to adjust
+your `DOCKER_HOST` to point at the proxy. Weave provides a convenient
+command for this:
 
-    host1$ export DOCKER_HOST=tcp://localhost:12375
+    host1$ eval "$(weave proxy-env)"
     host1$ docker ps
     ...
+
+Alternatively, the proxy host can be set on a per-command basis with
+
+    host1$ docker $(weave proxy-config) ps
 
 If you are working with a remote docker daemon, then `localhost` in
 the above needs to be replaced with the docker daemon host, and any
@@ -137,12 +142,12 @@ to point at the latter:
     host1$ weave launch -iprange 10.2.3.0/24
     host1$ weave launch-dns 10.2.4.1/24
     host1$ weave launch-proxy --with-ipam --with-dns
-    host1$ export DOCKER_HOST=tcp://localhost:12375
+    host1$ eval "$(weave proxy-env)"
 
     host2$ weave launch -iprange 10.2.3.0/24 host1
     host2$ weave launch-dns 10.2.4.2/24
     host2$ weave launch-proxy --with-ipam --with-dns
-    host2$ export DOCKER_HOST=tcp://localhost:12375
+    host1$ eval "$(weave proxy-env)"
 
 NB: Note that the two weaveDNS instances must be given unique IPs, on
 a subnet different from that used for IP allocation.
@@ -195,7 +200,8 @@ with
 
     $ mkdir -pv ~/.docker
     $ cp -v {ca,cert,key}.pem ~/.docker
-    $ export DOCKER_HOST=tcp://host1:12375 DOCKER_TLS_VERIFY=1
+    $ eval "$(weave proxy-env)"
+    $ export DOCKER_TLS_VERIFY=1
     $ docker version
     ...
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: weave [hunk 1]
```diff
@@ -889,6 +918,14 @@ case "$COMMAND" in
...
         echo $PROXY_CONTAINER
         ;;
+    proxy-env)
+        proxy_addr "export DOCKER_HOST="
+        exit $?
+        ;;
+    proxy-config)
+        proxy_addr "-H="
+        exit $?
+        ;;
```
rade: why the exit? we don't do that elsewhere. And why not make `proxy_addr` echo the address to stdout? See `container_id` for an example of a function that works that way.

rade:
[I said](https://github.com/weaveworks/weave/issues/753#issuecomment-106864096) that ideally these commands would run w/o weavexec.

And why do we need to inspect the proxy container? We know what the port is; it's hard-coded in the script.

One reason I can see for both is the looming #805. So perhaps we should tackle that one first.

paulbellamy:
You [later said](https://github.com/weaveworks/weave/issues/753#issuecomment-107855154) that running it locally won't work due to #805.

I agree, that doing #805 first is a good plan.

paulbellamy:
The situation this doesn't handle, is if you have:
```sh
$ weave_on <remote_host> launch-proxy -L 127.0.0.1:12375
083896110ec5162b18dec9b833433812314a82a8be719642ff9effdb5d481425
$ weave proxy-env
export DOCKER_HOST=tcp://127.0.0.1:12375
```
Which won't work if you are not on the remote host, but aside from throwing an error, I'm not sure what we can do in that situation.

test/650_proxy_env_test.sh
```diff
@@ -0,0 +1,13 @@
...
+
+. ./config.sh
+
+start_suite "Configure the docker daemon for the proxy"
+
+weave_on $HOST1 launch-proxy
+
+CMD="run -e WEAVE_CIDR=10.2.1.4/24 $BASE_IMAGE $CHECK_ETHWE_UP"
+assert_raises "eval '$(weave_on $HOST1 proxy-env)' ; docker $CMD"
+assert_raises "docker $(weave_on $HOST1 proxy-config) $CMD"
```
rade: Nice!

rade:
> `weave_on <remote_host> launch-proxy -L 127.0.0.1:12375`

I don't understand this comment. What does `-L` have to do with anything? Did you mean `-H`?

Also, it seems to me that `-H` isn't handled at all, and that we are simply stashing and retrieving 12375 - in a very, very, complicated way!

paulbellamy:
Yeah. Needs a bit of a rework now that #805 is done.

rade:
We could hold our noses and extract the listen address from the proxy's docker logs. Seems preferable to me to doing a lot of arg parsing and introducing bogus `-p`.

weave
```diff
@@ -805,6 +811,18 @@ proxy_args() {
     done
 }
 
+proxy_addr() {
+    if addr=$(docker logs weaveproxy | grep -oE "proxy listening on .*"); then
```
rade: I wonder how this will perform when the proxy log is long. Is there a way to short-circuit this? Would adding a reasonably-sized `head` work? Or would the sub-expression still only terminate after the entire log has been retrieved?

rade: Or `grep -m 1 -oE`?

weave
```diff
@@ -805,6 +811,18 @@ proxy_args() {
     done
 }
 
+proxy_addr() {
+    if addr=$(docker logs weaveproxy | head -n3 | grep -oE "proxy listening on .*"); then
```
rade: I had suggested `grep -m 1`, which would make this resilient to extra lines being. Does that not have the desired effect?

rade: argh, `-m` is not in posix, so scratch that idea.

rade:
docs need updating

weave
paulbellamy: `-m 1`, also didn't seem to terminate input from stdin early.

site/proxy.md
```diff
@@ -40,13 +40,18 @@ interfaces. This can be adjusted with the `-H` argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:9999
 
-All docker commands can be run via the proxy, so it is safe to
-globally adjust your `DOCKER_HOST` to point at the proxy, e.g.
+All docker commands can be run via the proxy, so it is safe to use
+`weave proxy-env` to globally adjust your `DOCKER_HOST` to point at
+the proxy, e.g.
```
rade: "All docker commands can be run via the proxy, so it is safe to adjust your `DOCKER_HOST` to point at the proxy. Weave provides a convenient command for this:"

rade:
done?

paulbellamy:
Yes.

PR #852, opened 2015-06-05
make `weave launch` args insensitive to order

### LINKED ISSUE #816: weave launch arguments are order sensitive
If an argument other than `-password`, `-port` or `-iprange` is given before one of those arguments, the `weave` script behaves as if you had not specified that later argument.

Example:

````
$ sudo weave launch -nickname foo -port 9999
9b3dc46e2
$ sudo docker inspect weave
...
    "Ports": {
        "6783/tcp"
    ...
````

This is caused by the `case` statement in `weave launch` bailing out the first time it sees an argument not in the set.

(we don't actually document `-port` but it's easier to demonstrate than `-iprange`, which we specifically place after `-nickname` in the usage text.)

### PR DESCRIPTION
**Pull request by rade**
Fixes #816.

### DIFF
--- a/weave
+++ b/weave
@@ -864,7 +864,8 @@ case "$COMMAND" in
             shift 1
         fi
         CONTAINER_PORT=$PORT
-        while true ; do
+        ARGS=""
+        while [ $# -gt 0 ] ; do
             case "$1" in
                 -password)
                     [ $# -gt 1 ] || usage
@@ -884,10 +885,12 @@ case "$COMMAND" in
                     shift 2
                     ;;
                 *)
-                    break
+                    ARGS="$ARGS '$(echo "$1" | sed "s|'|'\"'\"'|g")'"
+                    shift
                     ;;
             esac
         done
+        eval "set -- $ARGS"
         # Set WEAVE_DOCKER_ARGS in the environment in order to supply
         # additional parameters, such as resource limits, to docker
         # when launching the weave container.

### CODE REVIEW

File: weave [hunk 1]
```diff
@@ -847,7 +849,7 @@ case "$COMMAND" in
             -p $PORT:$CONTAINER_PORT/tcp -p $PORT:$CONTAINER_PORT/udp \
             -v /var/run/docker.sock:/var/run/docker.sock \
             -e WEAVE_PASSWORD \
-            $WEAVE_DOCKER_ARGS $IMAGE -iface $CONTAINER_IFNAME -port $CONTAINER_PORT -name "$PEERNAME" -nickname "$(hostname)" $IPRANGE "$@")
+            $WEAVE_DOCKER_ARGS $IMAGE -iface $CONTAINER_IFNAME -port $CONTAINER_PORT -name "$PEERNAME" -nickname "$(hostname)" $IPRANGE $ARGS)
```
bboreham: There is a subtle change in behaviour here, in that arguments are no longer quoted here and so will be interpreted by the shell.  For example. `weave launch -nickname "foo bar"` now results in an error: "bar: no such host".

rade: any idea how to fix this?

bboreham: I tried a few things but fundamentally `"$@"` is very special and I can't see a way to synthesise something similar using regular shell variables.

If the case-statement did a 'break' as before, but covered every expected parameter, then the original problem wouldn't arise.  But this is rather fragile.

bboreham: BTW, it's not quite true that arguments are interpreted by the shell, in the sense that `-nickname "foo > bar"` will not do a redirection.  It results in weave seeing `-nickname foo` followed by a peername `> bar`

rade: @dpw any suggestions?

dpw: You can quote the arguments you want to keep (using a bit of sed to protect any quotes that occur), and then `eval set` to put that back into the positional parameters, like so:

```
# set with no args does something different, so we need to eliminate that case
if [ $# -gt 0 ] ; then
   ARGS=
   while [ $# -gt 0 ] ; do
       case "$1" in
           -password)
               shift
               ;;
           *)
               ARGS="$ARGS '$(echo "$1" | sed "s|'|'\"'\"'|g")'"
               shift
               ;;
       esac
   done

   eval "set -- $ARGS"
fi

echo "$@"
echo $#
```

Alternatively, I think it can be also done using computed variable names (e.g. save the argument values to $arg0, $arg1, $arg2 etc., then repeatedly do `set -- "$@" "${arg$n}"` to put them back into the positional parameters).

bboreham: LGTM

bboreham: I mean, I tried the `sed` idea and it passed my tests

rade: Thanks! Based on the [docs](http://pubs.opengroup.org/onlinepubs/009695399/utilities/set.html) and experiments, the 0-arg guard is unnecessary:

> The command set -- without argument shall unset all positional parameters and set the special parameter '#' to zero.

dpw: > The command set -- without argument shall unset all positional parameters and set the special parameter '#' to zero.

Great, the special case has a special case.

PR #853, opened 2015-06-05
introduce '-nodiscovery' option to disable automatic peer discovery

### LINKED ISSUE #847: option to disable peer discovery
In some situations users may want complete control over the weave peer topology, i.e. exactly which peers connect to which other peers, and by what addresses.

`-nodisco`.

rade:
Closes #847.


paulbellamy:
`-nodiscovery`, please.

rade:
I intentionally went for the 70s pun :)

paulbellamy:
Yeah, I appreciate the pun.

rade:
anything else?

paulbellamy:
In the status output, we say on/off for other items, so it would be good to use that instead of enabled/disabled.

But other than that, LGTM.

rade:
blocked on #863.

PR #856, opened 2015-06-05
Respect `DOCKER_CLIENT_ARGS` in `weave run` (close #855)

### LINKED ISSUE #855: `weave run` does not respect DOCKER_CLIENT_ARGS
As any other commands do.

errordeveloper:


rade:
needs to be branched off 0.11

errordeveloper:
Will pull-request into 0.11 then.

Issue #819, opened 2015-06-02
IP leak for short-lived containers


rade:
When a container dies just before we obtain an IP address for it, we leak that address. To reproduce, apply

````patch
--- a/weave
+++ b/weave
@@ -975,6 +975,7 @@ case "$COMMAND" in
         [ $# -eq 1 ] || usage
         CONTAINER=$(container_id $1)
         create_bridge
+        docker kill $CONTAINER
         get_ipam_cidr $CONTAINER
         with_container_netns $CONTAINER attach $CIDR_ARGS >/dev/null
         when_dns_running with_container_fqdn $CONTAINER put_dns_fqdn $CIDR_ARGS
````
and then run

````
$ weave launch -iprange 10.2.0.0/24
$ weave run -ti ubuntu
$ weave status | grep "Allocator"
Allocator subnet 10.2.0.0/24  Free IPs: ~98.8%, 253 local, ~0 remote
````
The 253 is the giveaway here that something is wrong. Should be 254.

rade:
Suggested fix is for IPAM to check (using the docker api) whether the container is alive after adding it to the datastructure which is cleaned up by the docker event stream handler.

PR #865, opened 2015-06-05
Remove stale records for short-lived containers

### LINKED ISSUE #821: stale records for short-lived containers
When a container dies just before we tell DNS about it, we end up creating a stale record that is never removed. To reproduce, apply

````patch
--- a/weave
+++ b/weave
@@ -684,6 +684,8 @@ put_dns_fqdn() {
     FQDN="$2"
     shift 2
 
+    docker kill $CONTAINER_ID
+
     for ADDR in "$@" ; do
         http_call $DNS_CONTAINER_NAME $DNS_HTTP_PORT PUT /name/$CONTAINER_ID/${
     done
````
and then run
````
$ weave launch
$ weave launch-dns 10.2.1.1/24
$ weave run 10.2.0.1/24 -h foo.weave.local -ti ubuntu
$ weave status | grep "foo.weave.local"
51b239d2adc1: foo.weave.local.[10.2.0.1]
````
That last output is proof of the stale record.

### PR DESCRIPTION
**Pull request by inercia**
This PR removes stale records for short-lived containers by asking Docker about the container that corresponds to the new name:IP that has been introduced.

This can be tested by following the procedure described by @rade in #821: with (and without) `docker kill $CONTAINER_ID` in `weave:put_dns_fqdn()` and then running something like:

```language:bash
#!/bin/sh

make clean
make all

./weave launch
./weave launch-dns 10.2.1.1/24 --debug
./weave run 10.2.0.1/24 -h foo.weave.local -ti ubuntu
echo "Checking if 'foo.weave.local' is there"
./weave status | grep "foo.weave.local"
echo "Stopping everything"
./weave stop-dns
./weave stop
```

Fixes #821


nameserver/http.go
```diff
@@ -65,6 +67,26 @@ func ServeHTTP(listener net.Listener, version string, server *DNSServer, domain
...
+				running := true
+
+				// check with Docker whether the container is really running, removing otherwise...
+				Debug.Printf("[http] Checking whether '%s' is running...", idStr)
+				if container, err := dockerCli.InspectContainer(idStr); err != nil {
+					if _, notThere := err.(*docker.NoSuchContainer); notThere {
+						running = false
+					}
+				} else {
+					running = container.State.Running
```
inercia: Maybe we should allow other states (like _Paused_) as well... (?)

rade:
I think the abstraction isn't quite right here. There's too much machinery in weavedns and too little in updater. Bear in mind that solving #819 will likely require very similar machinery.

nameserver/http.go
rade: Interesting question. Whatever we do here should be consistent with our processing of the event stream. I don't think we pay attention to 'pause' transitions there, do we? Perhaps we should. raise an issue.

rade: > raise an issue.

#867.

inercia:
@rade If with _"too much machinery in weavedns"_ you mean [lines 72-86](https://github.com/inercia/weave/blob/weave-821/nameserver/http.go#L72), IMO fourteen lines are not so much machinery, specially when they are pretty specific of WeaveDNS (eg, [deleting records](https://github.com/inercia/weave/blob/weave-821/nameserver/http.go#L85) in the Zone).

Adding this mechanism in the _"updater"_ would not be the right place either, as the _"updater"_ is really a _Docker listener_, and this does not listen for anything: it just checks whether the container is running. I raised the question in HipChat whether we should have a general _"docker client"_ in `common`...

So I would wait until #819 before trying to do an early generalization...

nameserver/http.go
inercia: So I guess the safest option is to leave it and only consider _"Running"_ containers until we have a _"quarantine"_...

rade:
fourteen lines is a lot of machinery when we could have something like `if isContainerRunning(container) {...}`.

Also, there's all the startup machinery in weavedns/main.go

inercia:
Yes, it would be nice to have `isContainerRunning()`, but the thing is where to put that  method. And, as you mentioned, _"the startup machinery in weavedns/main.go"_ is the Docker client initialization, used both by the _"updater"_ and by this code. So I would prefer not to transform the issue _"stale records for short-lived containers"_ into _"refactor all our Docker code into a common client"_...

rade:
weavedns/main.go has no business initialising a docker client. This should be well abstracted away. The new import of "github.com/fsouza/go-dockerclient" in main.go is a dead giveaway.

inercia:
Ok, then we will have a `common/docker/client.go` file... :)

common/docker/client.go
```diff
@@ -0,0 +1,81 @@
...
+// An observer for container events
+type ContainerObserver interface {
+	ContainerDied(ident string) error
+}
+
+type Client struct {
+	client *docker.Client
+}
+
+// NewClint creates a new Docker client
```
rade: typo

common/docker/client.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}()
+	return nil
+}
+
+// IsContainerRunning returns true if the container is up and running
+func (c *Client) IsContainerRunning(idStr string) bool {
+	running := true
+
+	// check with Docker whether the container is really running, removing otherwise...
```
rade: "removing otherwise" is out of place

common/docker/client.go
```diff
@@ -0,0 +1,81 @@
...
+	if container, err := c.client.InspectContainer(idStr); err != nil {
+		if _, notThere := err.(*docker.NoSuchContainer); notThere {
+			running = false
+		}
+	} else {
+		running = container.State.Running
+	}
+
+	return running
+}
```
rade: I don't understand the logic here; why do we return `true` when inspecting the container fails with something other than `NoSuchContainer`?

rade:
The wrapping can be simpler:

````patch
diff --git a/common/docker/client.go b/common/docker/client.go
index ecd213d..73c5cfd 100644
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -11,7 +11,7 @@ type ContainerObserver interface {
 }
 
 type Client struct {
-       client *docker.Client
+       *docker.Client
 }
 
 // NewClient creates a new Docker client
@@ -21,10 +21,7 @@ func NewClient(apiPath string) (*Client, error) {
                return nil, err
        }
 
-       c := Client{
-               client: dc,
-       }
-       return &c, nil
+       return &Client{dc}, nil
 }
 
 // Start starts the client
@@ -47,7 +44,7 @@ func (c *Client) Start(apiPath string) error {
 // AddObserver adds an observer for docker events
 func (c *Client) AddObserver(ob ContainerObserver) error {
        events := make(chan *docker.APIEvents)
-       if err := c.client.AddEventListener(events); err != nil {
+       if err := c.AddEventListener(events); err != nil {
                Error.Printf("[docker] Unable to add listener to Docker API: %s"
                return err
        }
@@ -69,7 +66,7 @@ func (c *Client) IsContainerRunning(idStr string) bool {
        running := true
 
        // check with Docker whether the container is really running
-       if container, err := c.client.InspectContainer(idStr); err != nil {
+       if container, err := c.InspectContainer(idStr); err != nil {
                if _, notThere := err.(*docker.NoSuchContainer); notThere {
                        running = false
                }
````

File: common/docker/client.go (hunk 4 above)
Comment (inercia) [inline common/docker/client.go]: Good catch. Fixed.

**Comment by inercia** [comment]
IMO that would not be appropriate here. Defining a struct like `type Client struct { *docker.Client }` does not make much sense, not only because it provides nothing on top of the original struct but also because we do not want to expose the whole `docker.Client` API, and that would do it... I think `common/docker/client.go` should remain as a _facade_ for the docker API, not as an alias.

**Comment by rade** [comment]
>  it provides nothing on top of the original struct

But it does. Namely the methods defined here.

> we do not want to expose the whole `docker.Client` API

I actually see no problem with that.

**Comment by rade** [comment]
what's happening with this?

**Comment by inercia** [comment]
@rade finishing #836, then I will resume this PR...

(anyway, I think you last comment is personal preference more than a real problem you found when reviewing the code, and I think this kind of comments slow down the review...)

**Comment by rade** [comment]
The code I posted is shorter and just as readable, if not more so, than the original.

**Comment by inercia** [comment]
I also considered that solution when writing the code, and I discarded it because I think type aliases are not a good idea, even when they lead to shorter code (_eg_, `type i int` is not a good idea, even when produces shorter code and you can also add your own methods). But I will change it because I don't like to engage in endless discussions like this...

**Comment by rade** [comment]
Can it actually be done with a type alias? I tried and failed.

**Comment by inercia** [comment]
(facepalm)

**Comment by inercia** [comment]
You can do this:
```Golang
type s string
func (something s) IsEmpty() bool { return len(something) == 0 }
```
so should we replace all our strings with `s`? :confused:

rade:
I know about type aliases, but I cannot get them to work in this situation. Can you? The problem I am encountering is that we get a *pointer* to the docker client, which I cannot figure out how to turn into the right type, if that is even possible.

rade:
Found the [answer](https://groups.google.com/d/msg/golang-nuts/Hbxekd9g09c/ZccBG_qiePAJ); so the wrapper struct actually works better than an alias.

inercia:
Changed the client...

rade:
Does this handle the non-container ids case correctly, e.g. when we add records for `weave:expose`?

rade:
and the linter is unhappy.

inercia:
Good point about the `weave:expose` thing: I think I will need a `if`...

And I will make the _linter_ happy again...

rade:
To address the `weave:expose` problem, I suggest you exclude idents from the aliveness check that are not container IDs, e.g. `^[a-f0-9]+$`.

inercia:
Fixed the container id check.

common/docker/client.go
```diff
@@ -0,0 +1,81 @@
...
+	ContainerDied(ident string) error
+}
+
+type Client struct {
+	*docker.Client
+}
+
+// IsContainerId returns True if the string provided is a valid container id
+func IsContainerId(idStr string) bool {
+	matched, err := regexp.MatchString("^[a-f0-9]+$", idStr)
```
rade: The golang way is to define a global `var` with `regexp.MustCompile`, and then call `MatchString` on that, which returns just a bool.

See, e.g. https://github.com/weaveworks/weave/blob/master/proxy/proxy.go#L21 and https://github.com/weaveworks/weave/blob/master/proxy/proxy.go#L89

nameserver/http.go
```diff
@@ -65,6 +67,17 @@ func ServeHTTP(listener net.Listener, version string, server *DNSServer, domain
...
+				} else if running, err := dockerCli.IsContainerRunning(idStr); err != nil {
+					Error.Printf("[http] Could not check container status: %s", err)
+				} else if !running {
+					Info.Printf("[http] '%s' is not running: removing", idStr)
+					server.Zone.DeleteRecords(idStr, name, ip)
+				}
+			}
 		} else {
 			Info.Printf("[http] Ignoring name %s, not in %s", name, domain)
 		}
```
rade: The nesting is getting a bit deep here. I suggest changing the outer `if` s.t. we end up with

````
if !dns.IsSubDomain(domain, name) {
    Info.Printf("[http] Ignoring..."...)
    return
}
Info.Printf("[http] Adding..."...)
...
```

rade: Ditto for the `if dockerCli`, i.e.

````
if dockerCli == nil {
    return
}
...
````

**Comment by rade** [comment]
also, it looks like this needs rebasing.

File: nameserver/http.go (hunk 6 above)
Comment (inercia) [inline nameserver/http.go]: I agree about the first comment, but I think that `if dockerCli == nil { return }` is a statement that has little meaning (besides breaking the flow) (and makes the code longer)...

Comment (rade) [inline nameserver/http.go]: The *intention* is to break the flow. And it only costs one line, and reduces indentation of 7.

It's not a big deal though. Leave it as is if you must.

File: common/docker/client.go (hunk 5 above)
Comment (rade) [inline common/docker/client.go]: NB: this is still relevant.

**Comment by rade** [comment]
The outdent still needs doing. Rest looks fine.

File: prog/weavedns/main.go [hunk 7]
```diff
@@ -95,14 +96,18 @@ func main() {
...
-	if err != nil {
+	if err := zone.Start(); err != nil {
 		Error.Fatal("[main] Unable to start the Zone database", err)
 	}
 	defer zone.Stop()
 
+	dockerCli, err = docker.NewClient(apiPath)
+	if err != nil {
+		Error.Fatal("[main] Unable to start docker client", err)
+	}
```
bboreham: This is a change in semantics - the point of fetching the version from Docker was to check that we could actually communicate, otherwise the AddObserver() silently fails to do anything.

rade: well spotted. looks like `Start()` isn't called anywhere. I suggest we fold the `Start()` behaviour into `NewClient()`.

bboreham:
I'm going to take this over in @inercia's absence.

PR #873, opened 2015-06-08
Do not overwrite negLocal entries in the cache

inercia:
#849 ported to master:

* Do not overwrite negLocal entries in the cache
* Unit tests


rade:
Why do we need a PR for this? It should just be a merge from 0.11 to master. That's how all other 0.11 changes have been handled.

inercia:
The reason is because master and 0.11 are very different now, while cherry picking a commit with 26 lines of code and merging it in master is trivial...

rade:
did you try the merge?

inercia:
No, but I've seen the output from `git diff master 0.11`...

rade:
That is irrelevant. There's only a problem if the code on master is substantially different in the area that has changed.

Issue #810, opened 2015-06-02
[proxy] listen on specific IPs


Zaephor:
The proxy is configured for :12375 but it appears to end up binding on all host nics, which feels dangerous due to --privileged --net=host switches being used and TLS not being available/ready yet.

I know there's already a feature request for custom ports #805, would be nice to also be able to restrict responses to a specific nic.

For example only respond to eth0:12375, or lo:12375, or any combination of ethX:12375 and lo:12375 or other nic devices.

rade:
Usually this is done by permitting the user to specify specific IP addresses, rather than interfaces, to listen on. So let's do that.

Zaephor:
Good point, I've been trying to play in iptables for too long... :rage1:

rade:
@paulbellamy I have just noticed that the proxy already has an (undocumented!) `-L` option to specify what address and port to listen on. So what actually needs doing here? Other than documentation.

paulbellamy:
We need to configure the exported port binding from the container. And, ideally rename `-L` to `-H`, to match docker config options.

Bregor:
Is it available for weave-1.5?

bboreham:
@Bregor yes, you pass the `-H` flag - take a look at https://www.weave.works/docs/net/latest/weave-docker-api/

Bregor:
@bboreham sorry, I confused it with router.
Is there any possibility to bind router to specific address instead of 0.0.0.0?

Bregor:
I mean `weacer` process is always listening 0.0.0.0 like following:

    tcp        0      0 0.0.0.0:6783            0.0.0.0:*               LISTEN      8761/weaver

Is there any way to change this?

bboreham:
Looking at the code, it accepts a `--host` argument which should take an IP address (and optional port number), like `172.31.20.13`.

Do you mind explaining why you want this?  The option is currently undocumented, and understanding what you want it for might help us to write the documentation.

bboreham:
(note that changing the port almost certainly won't work without other changes.  this is probably connected with the fact that it's undocumented)

Bregor:
@bboreham thank you very much!
`--host` is what I missed.

Explanation is simple: I want to run weave only on internal interfaces (I'm on bare metal) and be calm about external connections.

bboreham:
OK, thanks!  

[...]

PR #875, opened 2015-06-08
Add support for custom proxy bind address

### LINKED ISSUE #805: [proxy] Add support for custom port

paulbellamy:
Closes #805 and #810

`weave status` will pick a random port to connect on (due to map ordering of `docker inspect`).
Renames `-L` flag to `-H` to match docker daemon, which means renaming the existing `-H` flag to `-D`. The `-H` flag doesn't *completely* match the docker daemon's, which takes a `tcp://` prefix, but is at least closer.

Debating adding a smoke test for this. Would be nice to have, but adds to the test run speed.

### DIFF
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -29,8 +29,8 @@ func main() {
 	c.Version = version
 	getopt.BoolVarLong(&debug, "debug", 'd', "log debugging information")
 	getopt.BoolVarLong(&justVersion, "version", 0, "print version and exit")
-	getopt.StringVar(&c.ListenAddr, 'L', fmt.Sprintf("address on which to listen (default %s)", defaultListenAddr))
-	getopt.StringVar(&c.DockerAddr, 'H', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultDockerAddr))
+	getopt.StringVar(&c.ListenAddr, 'H', fmt.Sprintf("address on which to listen (default %s)", defaultListenAddr))
+	getopt.StringVar(&c.DockerAddr, 'D', fmt.Sprintf("docker daemon URL to proxy (default %s)", defaultDockerAddr))
 	getopt.StringVarLong(&c.TLSConfig.CACert, "tlscacert", 0, "Trust certs signed only by this CA")
 	getopt.StringVarLong(&c.TLSConfig.Cert, "tlscert", 0, "Path to TLS certificate file")
 	getopt.BoolVarLong(&c.TLSConfig.Enabled, "tls", 0, "Use TLS; implied by --tlsverify")
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -13,7 +13,7 @@ import (
 var (
 	version           = "(unreleased version)"
 	defaultDockerAddr = "unix:///var/run/docker.sock"
-	defaultListenAddr = ":12375"
+	defaultListenAddr = "tcp://0.0.0.0:12375"
 )
 
 func main() {
@@ -51,6 +51,16 @@ func main() {
 
 	Info.Println("weave proxy", version)
 	Info.Println("Command line arguments:", strings.Join(os.Args[1:], " "))
+
+	protoAddrParts := strings.SplitN(c.ListenAddr, "://", 2)
+	if len(protoAddrParts) == 2 {
+		if protoAddrParts[0] != "tcp" {
+			Error.Fatalf("Invalid protocol format: %q", protoAddrParts[0])
+		}
+		c.ListenAddr = protoAddrParts[1]
+	} else {
+		c.ListenAddr = protoAddrParts[0]
+	}
 
 	p, err := proxy.NewProxy(c)
 	if err != nil {
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -35,9 +35,13 @@ argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:4243
 
-The proxy listens on port 12375, on all network interfaces. All docker
-commands can be run via the proxy, so it is safe to globally adjust
-your `DOCKER_HOST` to point at the proxy, e.g.
+By default, the proxy listens on port 12375, on all network
+interfaces. This can be adjusted with the `-L` argument, e.g.
+
+    host1$ weave launch-proxy -L 127.0.0.1:9999
+
+All docker commands can be run via the proxy, so it is safe to
+globally adjust your `DOCKER_HOST` to point at the proxy, e.g.
 
     host1$ export DOCKER_HOST=tcp://localhost:12375
     host1$ docker ps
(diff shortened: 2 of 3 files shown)


rade:
I am not convinced that the benefit of handling multiple `-H` justifies the complexity.

weave
```diff
@@ -765,6 +780,15 @@ proxy_args() {
     PROXY_ARGS=""
     while [ $# -gt 0 ]; do
         case "$1" in
+          -H)
+            shift
+            PROXY_PORTS="$PROXY_PORTS $1"
+            PROXY_ARGS="$PROXY_ARGS -H $1"
+            ;;
+          -H=*)
```
rade: We don't need to support `-H=`.

rade:
and for further simplification, we may only want to support "<host>:<port>", not any other flavour.

paulbellamy:
Some of the complexity in the sh script (like `container_exposed_ports`) is needed for `weave status` even with just one `-H`. I've done a second pass at the go listener instantiation, and cleaned it up a bit.

weave
paulbellamy: We support the `=` for all the TLS args, so it's consistent to match that here.

paulbellamy:
Which other flavours? Do you mean `eth0:12375`?

rade:
`:<port>`.

paulbellamy:
Ah. I don't think supporting blank host is that much more work. We need a "bind everywhere" default, anyway.

weave
rade: I didn't think `-<flag>=` was supported. Apparently it is. Nuts.

weave
```diff
@@ -28,7 +28,7 @@ usage() {
     echo "weave launch       [-password <password>] [-nickname <nickname>] [-nodiscovery]"
     echo "                       [-iprange <cidr>] <peer> ..."
     echo "weave launch-dns   <cidr>"
-    echo "weave launch-proxy [-H <docker_endpoint>] [--with-dns] [--with-ipam]"
+    echo "weave launch-proxy [-D <docker_endpoint>] [-L <listen_address> ...] [--with-dns] [--with-ipam]"
```
rade: Neither `<docker_endpoint>` nor `<listen_address>` are defined.

Also, it's `-H`, not `-L`.

And the format for `-L` is wrong, since presumably it needs to precede every address, and not just be supplied once, for all addresses.

rade:
Please take out the multi `-H` support.

paulbellamy:
https://github.com/paulbellamy/weave/commit/13098c78643cbf7e153915f4a7c7b674dbff7c5d?diff=unified#commitcomment-11566795 is correct, we don't need to export the port. But it allows us to find the port again for `weave status`. We could use a label for that instead, if you'd prefer?

rade:
>  it allows us to find the port again

Indeed, I realised that after posting my comment.

> We could use a label for that instead

Nah, they are horrible and require recent dockers.

Just add a "only needed so we can rediscover the port later" comment

weave
```diff
@@ -958,7 +983,8 @@ case "$COMMAND" in
         echo
         http_call $DNS_CONTAINER_NAME $DNS_HTTP_PORT GET /status 2>/dev/null || true
         echo
-        http_call_ip 127.0.0.1 $PROXY_PORT GET /status 2>/dev/null || true
+        container_exposed_port $PROXY_CONTAINER_NAME 2>/dev/null || true
```
rade:
````
if proxy_addr=$(docker inspect --format='{{range .HostConfig.PortBindings}}{{range .}}{{.HostIp}}:{{.HostPort}}{{end}}{{end}}' $PROXY_CONTAINER_NAME) ; then
...
````

File: weave [hunk 4]
```diff
@@ -48,9 +48,10 @@ usage() {
     echo "weave reset"
     echo "weave rmpeer       <peer_id>"
     echo
-    echo "where <peer>    is of the form <ip_address_or_fqdn>[:<port>], and"
-    echo "      <cidr>    is of the form <ip_address>/<routing_prefix_length>"
-    echo "      <peer_id> is a <nickname> or weave internal peer ID"
+    echo "where <peer>           is of the form <ip_address_or_fqdn>[:<port>], and"
+    echo "      <cidr>           is of the form <ip_address>/<routing_prefix_length>"
+    echo "      <peer_id>        is a <nickname> or weave internal peer ID"
+    echo "      <listen_address> is of the form [<ip_address>]:<port>"
```
rade: I thought we also allowed just `<ip_address>`.
And perhaps call the whole thing just `<address>`.

paulbellamy:
Decided not to rename the flags, as the `-L` taken by the proxy is a different format from the docker `-H`, so didn't want to confuse them.

weave
```diff
@@ -51,6 +51,7 @@ usage() {
     echo "where <peer>    = <ip_address_or_fqdn>[:<port>]"
     echo "      <cidr>    = <ip_address>/<routing_prefix_length>"
     echo "      <peer_id> = <nickname> or weave internal peer ID"
+    echo "      <address> = <ip_address>]:<port>"
```
rade: typo

rade:
Hmm. It is rather unfortunate that we have different formats for the inbound and outbound side.

paulbellamy:
yeah, it would be nice for `-L` to take the full `tcp://foo:12345`, might be doable now that we aren't parsing it to expose ports. Let me have a go at that.

site/proxy.md
```diff
@@ -35,9 +35,13 @@ argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:4243
 
-The proxy listens on port 12375, on all network interfaces. All docker
-commands can be run via the proxy, so it is safe to globally adjust
-your `DOCKER_HOST` to point at the proxy, e.g.
+By default, the proxy listens on port 12375, on all network
+interfaces. This can be adjusted with the `-L` argument, e.g.
+
+    host1$ weave launch-proxy -L 127.0.0.1:9999
```
rade: Add a `tcp://` prefix here.

rade:
> Decided not to rename the flags, as the proxy `-L` is a different format from the docker `-H`, so didn't want to cause confusion.

It's much closer now.

IMO, `-H` makes more sense. If you `weave launch-proxy -H <blah>` then odds are you'll subsequently want invoke docker with `-H <blah>`, or set `DOCKER_HOST=<blah>`. Which is exactly the same as what you'd have to do if you ran the daemon with `-H <blah>`.

prog/weaveproxy/main.go
```diff
@@ -52,6 +52,16 @@ func main() {
 	Info.Println("weave proxy", version)
 	Info.Println("Command line arguments:", strings.Join(os.Args[1:], " "))
 
+	protoAddrParts := strings.SplitN(c.ListenAddr, "://", 2)
+	if len(protoAddrParts) == 2 {
+		c.ListenAddr = protoAddrParts[1]
+		if protoAddrParts[0] != "tcp" {
+			Error.Fatalf("Invalid protocol format: %q", protoAddrParts[0])
+		}
```
rade: nitpick: these two statements should be swapped, i.e. the check should come first.

Issue #854, opened 2015-06-05
proxy status missing from docs


rade:
When we added the proxy status to `weave status in #800, we failed to update the docs. There should be a link at the bottom of http://docs.weave.works/weave/latest_release/troubleshooting.html#status-report to a section in the proxy docs.

PR #800, opened 2015-05-29
Add proxy info to 'weave status' output

paulbellamy:
Closes #756

Issue #756, opened 2015-05-27
[proxy] `weave status` should report on proxy status


rade:

rade:
so what should it report? just that it's running, and whether `--with-dns` or `--with-ipam` are in effect.

PR #878, opened 2015-06-08
remove most of proxy status

### LINKED ISSUE #876: Enabling TLS removes proxy from `weave status` output
Because curl cannot connect to the proxy over TLS.

### LINKED ISSUE #877: remove proxy from weave status output
it's way more hassle to than its worth. See, e.g. #876.

rade:
There are two problems with obtaining the proxy status via the normal
proxy endpoint...

1. when TLS is enabled, this requires using TLS at the client end

2. the status URL is mixed in with the docker remote API, which is
rather unclean

There is limited value to having some status output from the proxy,
since all the output is constant.

So instead we log the version and command line args.

We still want a way to wait for the proxy to start up, and log a nice
error if it dies. We do this by polling the docker logs for a startup
message.

Fixes #877. Fixes #876.

Undoes #756, #800, #854.

### DIFF
--- a/weave
+++ b/weave
@@ -562,11 +562,12 @@ http_call() {
     http_call_ip $CONTAINER_IP "$@"
 }
 
-# Wait until container $1 on IP $2, port $3 is alive enough to respond
-# to its status call
-wait_for_status_ip() {
+# Wait until container $1 is alive enough to respond to "GET /status"
+# http request on its docker-assigned IP, port $2
+wait_for_status() {
+    container_ip $1 "$1 container has died." "$1 container has died." || return 1
     while true ; do
-        http_call_ip $2 $3 GET /status >/dev/null && return 0
+        http_call_ip $CONTAINER_IP $2 GET /status >/dev/null && return 0
         if ! container_id $1 >/dev/null 2>&1 ; then
             echo "$1 container has died." >&2
             return 1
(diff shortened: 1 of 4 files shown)


rade:
I am not sure about this one. It would be nice to have a proxy status. But not when we have to jump through hoops as in #875.

One possible improvement here is that we could report in `weave status` that the proxy is running.

paulbellamy:
It is nice for debugging to just say "Please send us `weave status`", so having the proxy config output as part of it is handy. Checking the container is running, then grabbing the status output from the logs is probably a reasonable compromise (either CLI flags, or formatted as now).

rade:
I've added a line to `weave status` when the proxy is running.

Processing the docker logs and extracing readable status output is quite horrible, so I didn't do that.

weave
```diff
@@ -575,9 +576,18 @@ wait_for_status_ip() {
...
 }
 
-wait_for_status() {
-    container_ip $1 "$1 container has died." "$1 container has died." || return 1
-    wait_for_status_ip $1 $CONTAINER_IP $2
+# Wait until container $1 outputs a line containing $2 in its log
+wait_for_log() {
+    fifo=/tmp/tmpfifo.$$
+    mkfifo $fifo || exit 1
+    docker logs -f $1 2>/dev/null >$fifo &
```
paulbellamy: Won't this be left hanging?

rade: try it

paulbellamy: Ah, it gets killed when the container exits, so there's probably no point removing the fifo either.

rade: The more interesting case is what happens when the container does *not* exit.

paulbellamy: ? How would the weave script never exit?

rade: I thought you were talking about the proxy container.

But this reminds me, we need to test this with `weave --local`.

paulbellamy: with `--local` it leaves the `docker logs -f weaveproxy` hanging.

rade: fixed

PR #884, opened 2015-06-09
feed full container id on `weave start` to DNS/IPAM

### LINKED ISSUE #881: `weave start` with container name or short id leaks allocated IPs and DNS records
To reproduce
````
$ weave launch -iprange 10.3.0.0/24
c86dea6734a9bee5788e07a9cafa064178d5f930d4582e60d8d59ef7f7f3604b
$ weave launch-dns 10.3.1.1/24
6615e65710c480b30f492b03529b70313e7b6a89e31719f3274ab08d2b1e5328
$ docker create --name=foo -h foo.weave.local -ti ubuntu
ba6622c1301f614d5e38422324a718199629f45b6e22331b163e2878edd71b71
$ weave status | grep -E '(Free IPs|foo.weave.local)'
Allocator subnet 10.3.0.0/24  Free IPs: ~99.2%, 254 local, ~0 remote
$ weave start foo
foo
$ weave status | grep -E '(Free IPs|foo.weave.local)'
Allocator subnet 10.3.0.0/24  Free IPs: ~98.8%, 253 local, ~0 remote
foo: foo.weave.local.[10.3.0.1]
$ docker kill foo
foo
$ weave status | grep -E '(Free IPs|foo.weave.local)'
Allocator subnet 10.3.0.0/24  Free IPs: ~98.8%, 253 local, ~0 remote
foo: foo.weave.local.[10.3.0.1]
````

The problem stems from the fact that `docker start` just echoes back what it is given, rather than the full container id. Which is most useless.

This was [spotted by @bboreham](https://github.com/bboreham/weave/commit/064dfc6db0ce38db3dd733ad601250ee38cce782)

### PR DESCRIPTION
**Pull request by rade**
We erroneously assumed that `docker start` returned the full container id. It doesn't and as a result we populated DNS with records with ids other than full container ids. And that's also what we told IPAM. As a result, DNS and IPAM didn't clean up after the container's death, resulting in stale DNS entries and leaked IPs.

Fixes #881.

### DIFF
--- a/weave
+++ b/weave
@@ -1023,12 +1023,13 @@ case "$COMMAND" in
         collect_cidr_args "$@"
         shift $CIDR_COUNT
         [ $# -eq 1 ] || usage
-        CONTAINER=$(docker start $1)
+        RES=$(docker start $1)
+        CONTAINER=$(container_id $1)
         create_bridge
         get_ipam_cidr $CONTAINER
         with_container_netns_or_die $CONTAINER attach $CIDR_ARGS
         when_dns_running with_container_fqdn $CONTAINER put_dns_fqdn $CIDR_ARGS
-        echo $CONTAINER
+        echo $RES
         ;;
     attach)
         collect_cidr_args "$@"
--- a/test/260_dns_removal_test.sh
+++ b/test/260_dns_removal_test.sh
@@ -1,0 +1,33 @@
+#! /bin/bash
+
+. ./config.sh
+
+C1=10.2.0.67
+C2=10.2.0.43
+NAME=seetwo.weave.local
+
+DOPTS="--name=c2 -h $NAME"
+
+check() {
+    assert_dns_record $HOST1 c1 $NAME $C2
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: weave [hunk 1]
```diff
@@ -1023,12 +1023,13 @@ case "$COMMAND" in
...
         [ $# -eq 1 ] || usage
-        CONTAINER=$(docker start $1)
+        RES=$(docker start $1)
+        CONTAINER=$(container_id $1)
         create_bridge
         get_ipam_cidr $CONTAINER
         with_container_netns_or_die $CONTAINER attach $CIDR_ARGS
         when_dns_running with_container_fqdn $CONTAINER put_dns_fqdn $CIDR_ARGS
-        echo $CONTAINER
+        echo $RES
```
bboreham: If we expected to get the full hex ID back, why should we not do the same for our users?

rade: The idea always was that you'd get back the result of `docker start`.

rade:
@bboreham perhaps you want to grab this from @inercia?

PR #886, opened 2015-06-09
Avoid gossip test race conditions

### LINKED ISSUE #885: gossip unit test failing intermittently
For example https://travis-ci.org/weaveworks/weave/builds/66049086:

````
--- FAIL: TestGossipTopology-2 (0.00s)
	util.go:134: Expected connections not found: from 02:00:00:02:00:00 to map[03:00:00:03:00:00:{}]
...
	github.com/weaveworks/weave/router.implTestGossipTopology(0xc2091082d0)
		/home/travis/gopath/src/github.com/weaveworks/weave/router/gossip_test.go:116 +0xc0b
````

Appears to be caused by topology updates being delivered on a different goroutine, which sometimes happens later than expected.  There is a 'FIXME' comment to the effect that `sendPendingGossip()` may not work.

`go test -race` reports many race conditions in this test, which may or may not be relevant.

bboreham:
Make flushing of GossipSender wait till done, and take extra locks to avoid race conditions.

Fixes #885.

### DIFF
--- a/router/gossip.go
+++ b/router/gossip.go
@@ -332,15 +332,17 @@ func (router *Router) sendPendingGossip() {
 	for _, channel := range router.GossipChannels {
 		channel.Lock()
 		for _, sender := range channel.senders {
-			ch := make(chan struct{})
-			sender.flushch <- ch
-			<-ch
+			sender.flush()
 		}
 		for _, sender := range channel.broadcasters {
-			ch := make(chan struct{})
-			sender.flushch <- ch
-			<-ch
+			sender.flush()
 		}
 		channel.Unlock()
 	}
+}
+
+func (sender *GossipSender) flush() {
+	ch := make(chan struct{})
+	sender.flushch <- ch
+	<-ch
 }
--- a/router/gossip.go
+++ b/router/gossip.go
@@ -63,7 +63,14 @@ func (sender *GossipSender) run() {
 			}
 			sender.send(pending)
 		case ch := <-sender.flushch:
-			sender.flush()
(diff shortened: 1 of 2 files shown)


bboreham:
Yes, adding a sleep is a kludge.  I tried various ideas to push the data through the `GossipSender`, but couldn't come up with a way that had no impact on regular running and didn't deadlock.

rade:
is `GoSched()` of any use, perhaps?

rade:
The linter is unhappy.

One way to make `flush()` work properly, is to introduce a `flush` channel of channels on `GossipSender`, which is read by the `run()` loop, does the flushing and signals completion to the caller by closing the passed in channel.

bboreham:
Re-implemented with the flush channel of channels.  Had to change the test logic slightly to take out a deadlock, which may have been getting in the way yesterday.

rade:
`sender.flush()` should be name the API called by the tests, i.e. it should do the send/receive on `flushch` (yuck, but I can't come up with anything better). Just move the current `flush()` code into the `run()` loop; it's quite short.

Also, the flush handling really should still pay attention to receiving a `nil` value.

Which brings up another issue... things will break if the `GossipSender` terminated.

bboreham:
Tests don't have access to `GossipSender`.  Maybe `GossipChannel.flush()`?

rade:
When I said "'tests" I meant `sendPendingGossip()`

rade:
anything left to do here?

bboreham:
I'm happy with it.

PR #890, opened 2015-06-11
Use container name as hostname

### LINKED ISSUE #750: [dns] set hostname based on container name
Instead of

    weave run --name=foo -h foo.weave.local -ti ubuntu

it would be nice to be able to just write

    weave run --name=foo -ti ubuntu

for the same effect.

at least when `--with-dns` is specified.

i.e. if no hostname is specified it should be set to `<container-name>.<weavedns-domain>`.

Gotcha: the weaveDNS domain can be overridden, so this should respect that. Which involves asking weaveDNS. Which may not be running yet, in which case we should just stick to weave.local.

awh:
Fixes #750.


proxy/create_container_interceptor.go
```diff
@@ -88,9 +90,40 @@ func (i *createContainerInterceptor) setWeaveDNS(container *createContainerReque
...
+func (i *createContainerInterceptor) getDNSDomain() (domain string) {
+	domain = nameserver.DefaultLocalDomain
+	dnsContainer, err := i.proxy.client.InspectContainer("weavedns")
+	if err != nil ||
+		dnsContainer.NetworkSettings == nil ||
+		dnsContainer.NetworkSettings.IPAddress == "" {
+		return
+	}
+
+	url := fmt.Sprintf("http://%s:6785/domain", dnsContainer.NetworkSettings.IPAddress)
```
rade: Define that port as a constant.

site/weavedns.md
```diff
@@ -30,7 +30,8 @@ on every weave host, by invoking the weave script command
 `launch-dns`. Application containers are then instructed to use
 weaveDNS as their nameserver by supplying the `--with-dns` option when
 starting them. Giving any container a hostname in the `.weave.local`
-domain registers it in weaveDNS.  For example:
+domain registers it in weaveDNS; containers started with `--with-dns`
+have a default hostname derived from their container name. For example:
```
rade: We need to update (nearly) all examples to use this feature. Explicitly specifying a hostname should be done just *once* (with an explanation).

test/260_use_name_as_hostname_test.sh
```diff
@@ -0,0 +1,30 @@
...
+
+assert "docker_inspect_fqdn $HOST1 c1-name" "c1-name.weave.local."
+assert "docker_inspect_fqdn $HOST1 c2-name" "c2-hostname.weave.local."
+assert "docker_inspect_fqdn $HOST1 c3-name" "c3-hostname.weave.local."
+assert "docker_inspect_fqdn $HOST1 c4-name" "c4-hostname.weave.local."
+assert "docker_inspect_fqdn $HOST1 c5-name" "c5-hostname.weave.local."
+assert "docker_inspect_fqdn $HOST1 c6-name" "c6-hostname.weave.local."
+assert "docker_inspect_fqdn $HOST1 c7-name" "c7-hostname.weave.local."
+
+end_suite
```
rade: Start a container, do the check, stop the container. Rinse, repeat. Then you can use the same name and hostname for all and introduce a convenient helper:

````
check --name c
check --name c -h $hname
check --name --hostname=$hname
...
````

File: test/config.sh [hunk 4]
```diff
@@ -124,12 +124,12 @@ assert_dns_record() {
...
 
     [ -z "$DEBUG" ] || greyly echo "Checking whether $name exists at $host:$container"
-    got_ip=$(exec_on $host $container getent hosts $name)
+    got_ip=$(exec_on $host $container getent hosts $name | awk '{print $1}')
     assert_raises "echo '$got_ip' | grep -q '$exp_ips_regex'"
 
     [ -z "$DEBUG" ] || greyly echo "Checking whether the IPs '$@' exists at $host:$container"
     for ip in "$@" ; do
-        assert "exec_on $host $container getent hosts $ip | tr -s ' '" "$ip $name"
+        assert "exec_on $host $container getent hosts $ip | awk '{print \$2}'" "$name"
```
rade: I don't understand this change or the one above. What's going on here?

weave
```diff
@@ -641,16 +641,45 @@ dns_args() {
...
             --dns-search=*)
                 DNS_SEARCH_ARG=""
+                shift
+                ;;
+            --name)
+                HOSTNAME_ARG="--hostname=$2.$(dns_domain)"
+                shift 2
+                ;;
+            --name=*)
+                HOSTNAME_ARG="--hostname=${1#*=}.$(dns_domain)"
```
rade: I suggest deferring the `$(dns_domain)`, so we don't execute it unnecessarily when `$HOSTNAME_SPECIFIED` turns out to be true.

paulbellamy:
Rebased onto master

weave
```diff
@@ -680,9 +709,9 @@ with_container_fqdn() {
     COMMAND="$2"
     shift 2
 
-    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}.' $CONTAINER 2>/dev/null) || return 0
+    CONTAINER_FQDN=$(docker inspect --format='{{.Config.Hostname}}.{{.Config.Domainname}}' $CONTAINER 2>/dev/null) || return 0
 
-    if echo "$CONTAINER_FQDN" | grep -qv '\.\.$' ; then
+    if echo "$CONTAINER_FQDN" | grep -Eq '[^.]+\..+' ; then
```
rade: I don't understand this change.

rade: Ah, the [commit](https://github.com/awh/weave/commit/4f6ace8b109ae9d9e91474dd95054e7fababa7fc) explains it. Nevermind.

test/config.sh
paulbellamy: getent hosts was returning the hostname twice, so the output wouldn't match. Could have changed it to be a grep instead, like above.
```
$ getent hosts $ip
10.2.0.78      c1.weave.local      c1.weave.local
```

rade: *why* is it returning the name twice? Obviously it's not doing that for all the existing tests, otherwise they'd fail.

paulbellamy: I *believe* it might be due to differences in alpine linux's use of [musl vs libc](http://wiki.musl-libc.org/wiki/Functional_differences_from_glibc#Name_Resolver_.2F_DNS), as it only returns once on ubuntu.

rade: Please investigate further. If weaveDNS actually returned the same name twice in response to a reverse lookup that would be bad.

Also, if your explanation is correct then I still don't understand how the tests can have passed previously, since the test containers always been using alpine.

paulbellamy: Yeah, swapping the image used to the dnsutils image (ubuntu based), also fixes it, so I've done that instead.

rade:
anything left to do here?

awh:
I've just pushed another doc update that refocuses around `--name` and treats `-h` as the exception - if you're happy I think that's everything.

awh:
Let me re-run all the smoketests one final time!

awh:
Unit + smoke tests pass locally.

PR #891, opened 2015-06-11
Match README instructions to main website

fintanr:


README.md
```diff
@@ -32,8 +32,7 @@ capabilities, so these can continue to be used by containers.
 Ensure you are running Linux (kernel 3.8 or later) and have Docker
 (version 1.3.1 or later) installed. Then install weave with
 
-    sudo wget -O /usr/local/bin/weave \
-      https://github.com/weaveworks/weave/releases/download/latest_release/weave
+    sudo curl -L git.io/weave -o /usr/local/bin/weave
```
rade: If we are using `curl` here then we should also do that [here](http://docs.weave.works/weave/master/head/troubleshooting.html#snapshots)

fintanr: @errordeveloper is there a particular reason we want to use curl rather than wget before I go further with this

fintanr:
updated.

site/troubleshooting.md
```diff
@@ -191,8 +191,7 @@ We sometimes publish snapshot releases, to provide previews of new
 features, assist in validation of bug fixes, etc. One can install the
 latest snapshot release with
 
-    sudo wget -O /usr/local/bin/weave \
-      https://raw.githubusercontent.com/weaveworks/weave/master/weave
+    sudo curl -L git.io/weave -o /usr/local/bin/weave
```
rade: Look where the original URL is pointing. It's *not* the latest release. So the short url is no use here.

errordeveloper: And to make it a bit more consistent and easier to remember, we can use `http://git.io/weave-snapshot`, which I just created.

Issue #870, opened 2015-06-08
enable IPAM by default


rade:
`weave launch` should launch with IPAM.

This requires #743.

The question is whether there should be a way to disable IPAM. The main reason to have that, I reckon, is that IPAM has the need to establish a quorum on initialisation, which may be problematic in some situations. However, that only occurs when IPAM is actually being used, i.e. an IP address allocation is requested. Hence always launching with IPAM enabled is fine.

Issue #758, opened 2015-05-27
[ipam][docs] specify maximum peer count


errordeveloper:
We currently state:

> It isn’t a problem to over-estimate by a bit, but if you supply a number that is too small then multiple independent groups may form.

We should be a bit more specific about maximum and minimum counts.

PR #893, opened 2015-06-11
Default value for -iprange

### LINKED ISSUE #743: [ipam] have a default for -iprange

bboreham:
Somewhat arbitrary default of 10.16.0.0/12 chosen to be pretty big and yet avoid regions like 10.0.2.0/24 used by VirtualBox.

The default subnet will also be 10.16.0.0/12. (This will need to change to accommodate various other IPAM requirements).

A new helper program 'netcheck' is introduced to see if any existing routes cover that address range, which is a proxy for 'is that space unused' (technique borrowed from Docker). Netcheck is put on the path inside the `weaveexec` container, to simplify the problem for people doing `weave --local launch` who want the default `-iprange`.

Fixes #743 

Also addresses #870, although it will perhaps annoy some people who don't want to use IPAM, if they do have a clash with the default, since it will force them to pick a non-clashing range.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ prog/weavedns/weavedns
 prog/weaveproxy/weaveproxy
 prog/sigproxy/sigproxy
 prog/weavewait/weavewait
+prog/netcheck/netcheck
 tools/bin
 tools/build
 releases
@@ -45,6 +46,7 @@ prog/weaveexec/weave
 prog/weaveexec/weaveproxy
 prog/weaveexec/weavewait
 prog/weaveexec/sigproxy
+prog/weaveexec/netcheck
 prog/weaveexec/docker*.tgz
 Vagrantfile.local
 test/tls/*.pem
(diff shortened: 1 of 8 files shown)


bboreham:
As from latest commit, `weave --local launch` with no available `netcheck` command will result in no default `-iprange` and IPAM turned off.

rade:
docs?

rade:
Also we said that if `netcheck` doesn't exist then just use the range w/o checking.

rade:
I've given the docs an [overhaul](https://github.com/rade/weave/commit/df918ab3c4040af5f9332a3a5a658c98cbb4543d). Please review. If you want to include that in this PR then great. If not I will apply it post-merge.

bboreham:
Looks ok.  Noting that #758 is still outstanding.

bboreham:
Looks like 10.128/9 clashes with addresses used on CircleCI, e.g. 10.240.0.1/32

PR #898, opened 2015-06-11
Weavedns default resolver

### LINKED ISSUE #383: make weavedns the name resolver for containers by default
Right now, containers started with `weave run` only use weavedns for name resolution when the `--with-dns` option has been specified. We did that because weavedns used to lack some key features required to act as a nameserver in a fair few cases. That has changed now, and hence that choice should be inverted, i.e. we should have a `--without-dns` option, though, arguably, only after #382.

NB: we'd want to avoid an ordering dependency here; i.e. it should be possible to start containers with dns pointing at weavedns even when weave/weavedns isn't running.

awh:
Fixes #383.

### DIFF
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -81,7 +81,12 @@ func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container *docker.Co
 }
 
 func (i *createContainerInterceptor) setWeaveDNS(container *createContainerRequestBody, r *http.Request) error {
-	if !i.proxy.WithDNS {
+	if i.proxy.WithoutDNS {
+		return nil
+	}
+
+	dnsDomain, dnsRunning := i.getDNSDomain()
+	if !(dnsRunning || i.proxy.WithDNS) {
 		return nil
 	}
 
@@ -95,13 +100,13 @@ func (i *createContainerInterceptor) setWeaveDNS(container *createContainerReque
 	if container.Hostname == "" && name != "" {
 		container.Hostname = name
 		// Strip trailing period because it's unusual to see it used on the end of a host name
-		container.Domainname = strings.TrimSuffix(i.getDNSDomain(), ".")
+		container.Domainname = strings.TrimSuffix(dnsDomain, ".")
 	}
 
 	return nil
 }
 
-func (i *createContainerInterceptor) getDNSDomain() (domain string) {
+func (i *createContainerInterceptor) getDNSDomain() (domain string, running bool) {
 	domain = nameserver.DefaultLocalDomain
 	dnsContainer, err := i.proxy.client.InspectContainer("weavedns")
 	if err != nil ||
@@ -122,7 +127,7 @@ func (i *createContainerInterceptor) getDNSDomain() (domain string) {
 		return
 	}
 
-	return string(b)
+	return string(b), true
 }
 
 func (i *createContainerInterceptor) InterceptResponse(r *http.Response) error {
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -115,25 +115,18 @@ are handled as before.
 
 ## <a name="dns"></a>Automatic discovery
 
-Containers started via the proxy are
-automatically registered in [weaveDNS](weavedns.html) if they have a
-hostname in the weaveDNS domain (usually `.weave.local`). In order for
-containers to be able to look up such names, their DNS resolver needs
-to be configured to point at weaveDNS. This will be done automatically by the proxy
- while weaveDNS is running. To disable this launch the proxy with the `--without-dns` option, i.e.
-
-    host1$ weave launch-proxy --without-dns
-
-To enable weaveDNS resolution when weaveDNS is not running, launch the
-proxy with the `--with-dns` option, i.e.
-
-    host1$ weave launch-proxy --with-dns
-
-Now any containers started via the proxy, and with a `WEAVE_CIDR=...`
-environment variable (or even without it, if IPAM is
-[configured](#ipam)), will use weaveDNS for name resolution. If not
-provided, the hostname will default to the container name and will be
-registered in weaveDNS.
+Containers started via the proxy are automatically registered in
+[weaveDNS](weavedns.html) if they have a hostname in the weaveDNS
+domain (usually `.weave.local`). In order for containers to be able to
+look up such names, their DNS resolver needs to be configured to point
+at weaveDNS. This will be done automatically by the proxy while
+weaveDNS is running. To override this behaviour launch the proxy with
+either `--with-dns` or `--without-dns`, which will force the proxy to
+always/never set the resolver to weaveDNS. If there is no hostname
+provided, the container will be registered in weaveDNS using its
+container name. Otherwise, if there is no container name, and no
+hostname (or a hostname outside the weaveDNS domain), the container
+will not be registered in weaveDNS.
 
 ## <a name="multi-host"></a>Multi-host example
 
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -115,13 +115,17 @@ are handled as before.
 
 ## <a name="dns"></a>Automatic discovery
 
-Containers started via the proxy are automatically registered in
-[weaveDNS](weavedns.html) if they have a hostname in the weaveDNS
-domain (usually `.weave.local`). In order for containers to be able to
-look up such names, their DNS resolver needs to be configured to point
-at weaveDNS. This can be done [manually](weavedns.html#without-run),
-or automatically by the proxy by launching it with the `--with-dns`
-option, i.e.
+Containers started via the proxy are
+automatically registered in [weaveDNS](weavedns.html) if they have a
+hostname in the weaveDNS domain (usually `.weave.local`). In order for
+containers to be able to look up such names, their DNS resolver needs
+to be configured to point at weaveDNS. This will be done automatically by the proxy
+ while weaveDNS is running. To disable this launch the proxy with the `--without-dns` option, i.e.
+
+    host1$ weave launch-proxy --without-dns
+
+To enable weaveDNS resolution when weaveDNS is not running, launch the
+proxy with the `--with-dns` option, i.e.
 
     host1$ weave launch-proxy --with-dns
 
--- a/site/weavedns.md
+++ b/site/weavedns.md
@@ -27,16 +27,27 @@ hosts.
 WeaveDNS is deployed as a set of containers that communicate with each
 other over the weave network. One such container needs to be started
 on every weave host, by invoking the weave script command
-`launch-dns`. Application containers are then instructed to use
-weaveDNS as their nameserver by supplying the `--with-dns` option when
-starting them; containers so started also automatically register their
-container name in the weaveDNS domain. For example:
+`launch-dns`.
 
 ```bash
 $ weave launch
 $ weave launch-dns 10.2.254.1/24
-$ weave run --with-dns 10.2.1.25/24 -ti --name=pingme ubuntu
-$ shell1=$(weave run --with-dns 10.2.1.26/24 -ti --name=ubuntu ubuntu)
+```
+
+Containers started while weaveDNS is running are automatically
+registered in [weaveDNS](weavedns.html) if they have a hostname in the
+weaveDNS domain (usually `.weave.local`). In order for containers to
+be able to look up such names, their DNS resolver needs to be
+configured to point at weaveDNS. This will be done automatically while
+weaveDNS is running. If there is no hostname provided, the container
+will be registered in weaveDNS using its container name. Otherwise, if
+there is no container name, and no hostname (or a hostname outside the
+weaveDNS domain), the container will not be registered in weaveDNS.
+For example, to use the container name as the hostname:
+
+```bash
+$ weave run 10.2.1.25/24 -ti --name=pingme ubuntu
+$ shell1=$(weave run 10.2.1.26/24 -ti --name=ubuntu ubuntu)
 $ docker attach $shell1
 
 root@ubuntu:/# ping pingme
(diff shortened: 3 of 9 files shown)


proxy/create_container_interceptor.go
```diff
@@ -94,34 +99,33 @@ func (i *createContainerInterceptor) setWeaveDNS(container *createContainerReque
...
 	defer resp.Body.Close()
 
 	b, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return
+		return nameserver.DefaultLocalDomain, false
 	}
 
-	return string(b)
+	return string(b), true
```
rade: one of the few places where named returns would work well, i.e. all the `false` cases can just be `return`.

site/proxy.md
```diff
@@ -115,13 +115,17 @@ are handled as before.
...
-at weaveDNS. This can be done [manually](weavedns.html#without-run),
-or automatically by the proxy by launching it with the `--with-dns`
-option, i.e.
+Containers started via the proxy are
+automatically registered in [weaveDNS](weavedns.html) if they have a
+hostname in the weaveDNS domain (usually `.weave.local`). In order for
+containers to be able to look up such names, their DNS resolver needs
+to be configured to point at weaveDNS. This will be done automatically by the proxy
+ while weaveDNS is running. To disable this launch the proxy with the `--without-dns` option, i.e.
+
```
rade: line length should wrap at 72.

site/proxy.md
```diff
@@ -115,13 +115,17 @@ are handled as before.
...
+    host1$ weave launch-proxy --without-dns
+
+To enable weaveDNS resolution when weaveDNS is not running, launch the
+proxy with the `--with-dns` option, i.e.
```
rade: I don't think we need examples of `--with/without-dns`.

So just "This will be done automatically by the proxy while weaveDNS is running. To override this behaviour launch the proxy with either `--with-dns` or `--without-dns`, which will force the proxy to always/never set the resolver to weaveDNS."

site/weavedns.md
```diff
@@ -24,19 +24,18 @@ hosts.
...
-`launch-dns`. Application containers are then instructed to use
-weaveDNS as their nameserver by supplying the `--with-dns` option when
-starting them; containers so started also automatically register their
-container name in the weaveDNS domain. For example:
+WeaveDNS is deployed as a set of containers that communicate with each other
+over the weave network. One such container needs to be started on every weave
+host, by invoking the weave script command `launch-dns`. Application
+containers are automatically configured to use weaveDNS as their nameserver if
+they are started whilst weaveDNS is running; such containers also register
+their container name in the weaveDNS domain. For example:
```
rade: Seems to be formatted to a different line length than before. Should be 72.

site/weavedns.md
```diff
@@ -51,6 +50,10 @@ it in weaveDNS simply by giving the container a hostname in the
 $ weave run 10.2.1.25/24 -ti -h pingme.weave.local ubuntu
 ```
 
+It's also possible to force or forbid an application container's use of
```
rade: "It is"

weave
```diff
@@ -145,9 +177,11 @@ collect_ip_args() {
 if [ "$1" = "run" ] ; then
     shift 1
     check_docker_version
-    if [ "$1" = "--with-dns" ] ; then
-        shift 1
-        DNS_ARGS=$(exec_remote dns-args "$@")
+    collect_dns_args "$@"
```
rade: don't collect them; the *must* come first.

weave
```diff
@@ -29,10 +29,10 @@ usage() {
     echo "weave launch       [-password <password>] [-nickname <nickname>] [-nodiscovery]"
     echo "                       [-iprange <cidr> [-ipsubnet <cidr>]] <peer> ..."
     echo "weave launch-dns   <cidr>"
-    echo "weave launch-proxy [-D <endpoint>] [-H <endpoint>] [--with-dns] [--with-ipam]"
+    echo "weave launch-proxy [-D <endpoint>] [-H <endpoint>] [--with-dns|--without-dns] [--with-ipam]"
```
rade: line is too long on display; needs to be <80 chars. I suggest putting all the `with` options on the 2nd line.

rade:
The tests are too direct. We want the tests to reflect what users/applications might do. So here we should check that name resolution works as expected.

rade:
[This error message](https://github.com/weaveworks/weave/pull/898/files#diff-ed6579eaedd0e67b3ed51c7ee7071044R679) needs updating.

proxy/create_container_interceptor.go
awh: I came to the same conclusion when this function was introduced in the original PR...

weave
awh: This has the exact same semantics as `collect_cidr_args` - it consumes zero or more dns related options from the _start_ of `$@`, stopping as soon as it hits the first thing that isn't...

proxy/create_container_interceptor.go
rade: Hah! I see what you mean. The previous code does indeed use named returns. Let's keep it that way.

awh:
> This error message needs updating.

Fixed.

weave
rade: ah, quite right; I didn't see the `break`.

rade: but I don't see why we need a loop.

awh:
> The tests are too direct. We want the tests to reflect what users/applications might do. So here we should check that name resolution works as expected.

I considered this, but concluded that since we have other tests that check whether the full behaviour of '-with-dns' is as required, that it would be OK here to merely check whether the option had been triggered or not. How do you feel about leaving this as is for 1.0, and creating an issue to come back and refactor the whole suite of DNS tests into something more streamlined?

awh:
Also the tests `270_use_name_as_hostname_test.sh` in the previous PR take the same stance; I took the merge of those as tacit approval to some extent.

rade:
Two wrongs don't make a right.

rade:
What irks me about these tests is that they really peek very deep under the covers and introduce a whole bunch of extra machinery to do so. By contrast, 270 was very clean and light.

awh:
> By contrast, 270 is very clean and light.

How about I change these tests to look at the container FQDN instead, as the 270 tests do? Then we could even share the helper functions between the two suites.

awh:
> Two wrongs don't make a right.

Interestingly, programmatic testing is one of the few areas where they do ;)

weave
awh: We did it for parity with the proxy implementation - that also errors when you specify both options.

rade: Meh.

rade:
> How about I change these tests to look at the container FQDN instead, as the 270 tests do?
> Then we could even share the helper functions between the two suites.

I think that would be ok. No promises though; will need to see the finished article.

site/proxy.md
```diff
@@ -116,20 +116,14 @@ are handled as before.
...
-environment variable (or even without it, if IPAM is
-[configured](#ipam)), will use weaveDNS for name resolution. If not
-provided, the hostname will default to the container name and will be
-registered in weaveDNS.
+[weaveDNS](weavedns.html) if they have a hostname in the weaveDNS domain
+(usually `.weave.local`). In order for containers to be able to look up
+such names, their DNS resolver needs to be configured to point at
+weaveDNS. This will be done automatically by the proxy while weaveDNS is
+running. To override this behaviour launch the proxy with either
+--with-dns or --without-dns, which will force the proxy to always/never
```
rade: literal quoting missing.

paulbellamy:
Rebased conflicts in weave script usage

weave
```diff
@@ -138,6 +139,38 @@ collect_ip_args() {
...
+                ;;
+            *)
+                break
+                ;;
+        esac
+    done
+    if [ "$WITH_DNS" = "true" -a "$WITHOUT_DNS" = "true" ] ; then
+        usage
+    fi
+}
```
rade: oops. accidentally deleted my comments here...

We don't need the looping, or checking whether both `--with` options were given. keep it simple.

weave
```diff
@@ -138,6 +139,30 @@ collect_ip_args() {
...
+            WITH_DNS=true
+            DNS_ARG_COUNT=1
+            shift
+            ;;
+        --without-dns)
+            WITHOUT_DNS=true
+            DNS_ARG_COUNT=1
+            shift
+            ;;
+    esac
```
rade: you don't need those `shift`s.

weave
```diff
@@ -145,9 +170,11 @@ collect_ip_args() {
 if [ "$1" = "run" ] ; then
     shift 1
     check_docker_version
-    if [ "$1" = "--with-dns" ] ; then
-        shift 1
-        DNS_ARGS=$(exec_remote dns-args "$@")
+    collect_dns_args "$@"
+    shift $DNS_ARG_COUNT
+    [ $# -gt 0 ] || usage
```
rade: why did this check suddenly appear?

rade: ...and because we don't need that check here or in the `run` handler below, we can combine `collect_dns_args` and `dns_available` into one function, since they are now invoked in sequence. Which is great since that side-channel information flow from the former to the latter via global vars is gross.

test/280_with_or_without_dns_test.sh
```diff
@@ -0,0 +1,41 @@
+#! /bin/bash
+
+. ./config.sh
+
+IP=10.2.0.34
+TARGET=seetwo.weave.local
+TARGET_IP=10.2.0.78
+
+# Start container with args $2.. and assert HostConfig Dns of $1
```
rade: that comment is wrong. remove.

weave
```diff
@@ -145,9 +170,11 @@ collect_ip_args() {
...
+    if dns_available exec_remote dns-running ; then
+      DNS_ARGS=$(exec_remote dns-args "$@")
```
rade: We don't want to perform two `exec_remote`s here, especially not in the common case. Conversely, I wouldn't be too bothered if we *always* performed *one* exec_remote here, if that simplified matters.

weave
paulbellamy: Are you sure you meant #901?

rade: #910

paulbellamy:
Fixed minor issues. Having a look at combining `collect_dns_args`, and `dns_available`

test/280_with_or_without_dns_test.sh
```diff
@@ -0,0 +1,40 @@
...
+    container=$(weave_on $HOST1 run "$@" $IP/24 -t $DNS_IMAGE /bin/sh)
+    assert_no_dns_record $HOST1 "$container" $TARGET
+    rm_containers $HOST1 "$container"
+}
+
+assert_resolution() {
+    container=$(weave_on $HOST1 run "$@" $IP/24 -t $DNS_IMAGE /bin/sh)
+    assert_dns_record $HOST1 "$container" $TARGET $TARGET_IP
+    rm_containers $HOST1 "$container"
+}
```
rade: quoting of `$container` is unnecessary in the above.

test/config.sh
```diff
@@ -111,6 +111,14 @@ start_container_with_dns() {
     weave_on $host run --with-dns "$@" -t $DNS_IMAGE /bin/sh
 }
 
+rm_containers() {
+    host=$1
+    shift
+    if [ $# -gt 0 ] ; then
+        docker_on $host rm -f "$@" >/dev/null 2>&1
+    fi
+}
```
rade: `[ $# -eq 0 ] || docker_on $host rm -f "$@" >/dev/null 2>&1`

Why the `2>&1`? Given that this whole thing fails when the rm fails, it seems odd that we'd swallow the error message. I know we had that indirection in `start_suite`, but I don't know why. Does `rm` produce output on stderr in some success cases?

paulbellamy:
I've eliminated one of the `exec_remote` calls (so there's only one now), but I don't see a way to combine `collect_dns_args` and `dns_available` without breaking things.

paulbellamy:
dns multicast smoketest fails, trying to diagnose why...

paulbellamy:
Removed the last few commits, as they broke smoke tests.

We need 2 exec-remote calls to A) determine if the dns is running, and B) do the dns args collection. We could combine them into one if we have it return a specific format, and do some pattern-matching on the output, but.

rade:
needs re-basing, even though gh doesn't think so.

rade:
I've rebased and aggressively squashed this. I'd love to squash it further but that would entail splitting some of the commits, which is more hassle than it's worth.

site/weavedns.md
```diff
@@ -27,16 +27,16 @@ hosts.
...
 other over the weave network. One such container needs to be started
 on every weave host, by invoking the weave script command
-`launch-dns`. Application containers are then instructed to use
-weaveDNS as their nameserver by supplying the `--with-dns` option when
-starting them; containers so started also automatically register their
-container name in the weaveDNS domain. For example:
+`launch-dns`. Application containers are automatically configured to
+use weaveDNS as their nameserver if they are started whilst weaveDNS
+is running; such containers also register their container name in the
+weaveDNS domain. For example:
```
rade: This fails to mention the setting of the hostname.

site/proxy.md
```diff
@@ -119,17 +119,12 @@ Containers started via the proxy are automatically registered in
...
-Now any containers started via the proxy, and with a `WEAVE_CIDR=...`
-environment variable (or even without it, if IPAM is
-[configured](#ipam)), will use weaveDNS for name resolution. If not
-provided, the hostname will default to the container name and will be
-registered in weaveDNS.
+at weaveDNS. This will be done automatically by the proxy while
+weaveDNS is running. To override this behaviour launch the proxy with
+either `--with-dns` or `--without-dns`, which will force the proxy to
+always/never set the resolver to weaveDNS.  If not provided, the
+hostname will default to the container name and will be registered in
```
rade: that's not quite right; the hostname is set to container name in the weaveDNS domain.

paulbellamy:
Rebased to include #929

Issue #869, opened 2015-06-08
make use of default IPAM throughout documentation


rade:
IPAM should be documented as an integral rather than separate feature. i.e. throughout the docs we should generally make use of IPAM. Usually with the default sub-net, i.e. not specifying any IP/network. And specifying a subnet (but no IP) should become the primary means to deal with multiple subnets. I reckon specifying individual IP addresses may well be something we only cover once, in a separate section.

We will still want to keep docs specifically about IPAM, to explain stuff about quorum and cover various exotic uses.

rade:
Ideally we'd tackle this after #675, #762, #763, #862, #870.

bboreham:
Following #900 we can update the `launch-dns` examples to remove any IP addresses.

PR #900, opened 2015-06-12
IPAM for weavedns

### LINKED ISSUE #762: [dns][ipam] launch dns w/o the need to specify an IP address
Once we have #675, we should make it possible to launch DNS with just specifying a network, not an IP.

rade:
Same as for ordinary containers.

Closes #762.


rade:
The discussion in #763 about the dangers (or otherwise) of running weavedns in the same subnet as apps is relevant here.

Docs still need updating.

Meanwhile, @bboreham, I'd appreciate some quick feedback on the code/approach.

rade:
The docs are a bit sparse, but I really didn't want to introduce a reference to IPAM into the weaveDNS docs at this stage. Will all become more obvious as part of #869.

PR #902, opened 2015-06-12
Keep flushing gossip until it's all gone

### LINKED ISSUE #885: gossip unit test failing intermittently

bboreham:
Fix #885 more thoroughly, by looping across multiple routers until none of them have anything to say.

### DIFF
--- a/router/gossip.go
+++ b/router/gossip.go
@@ -39,9 +39,11 @@ type Gossiper interface {
 // Accumulates GossipData that needs to be sent to one destination,
 // and sends it when possible.
 type GossipSender struct {
-	send    func(GossipData)
-	cell    chan GossipData
-	flushch chan chan struct{}
+	send func(GossipData)
+	cell chan GossipData
+	// for testing
+	sent    bool
+	flushch chan chan bool
 }
 
 func NewGossipSender(send func(GossipData)) *GossipSender {
@@ -50,7 +52,7 @@ func NewGossipSender(send func(GossipData)) *GossipSender {
 
 func (sender *GossipSender) Start() {
 	sender.cell = make(chan GossipData, 1)
-	sender.flushch = make(chan chan struct{})
+	sender.flushch = make(chan chan bool)
 	go sender.run()
 }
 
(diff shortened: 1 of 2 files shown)


router/gossip_test.go
```diff
@@ -32,6 +32,17 @@ func (conn *mockChannelConnection) SendProtocolMsg(protocolMsg ProtocolMsg) {
...
+func sendPendingGossip(routers ...*Router) {
+	done := false
+	// Loop until all routers report they didn't send anything
+	for !done {
+		done = true
+		for _, router := range routers {
+			done = !router.sendPendingGossip() && done
+		}
+	}
+}
```
rade:
````
func sendPendingGossip(routers ...*Router) {
	more := true
	// Loop until all routers report they didn't send anything
	for more {
		for _, router := range routers {
			more = router.sendPendingGossip() || more
		}
	}
}
````

Comment (rade) [inline router/gossip_test.go]: or, even better, `for more := true; more; {...}`

Comment (rade) [inline router/gossip_test.go]: ah, that's wrong. never mind.

Comment (rade) [inline router/gossip_test.go]:
````
func sendPendingGossip(routers ...*Router) {
	for sentSomething := true; sentSomething; {
		sentSomething = false
		for _, router := range routers {
			sentSomething  = router.sendPendingGossip() || sentSomething
		}
	}
}
````

File: router/gossip.go [hunk 2]
```diff
@@ -328,21 +329,23 @@ func (c *GossipChannel) log(args ...interface{}) {
...
-			sender.flush()
+			didSomething = sender.flush() || didSomething
 		}
 		for _, sender := range channel.broadcasters {
-			sender.flush()
+			didSomething = sender.flush() || didSomething
 		}
 		channel.Unlock()
 	}
+	return didSomething
```
rade: s/didSomething/sentSomething

router/gossip.go
```diff
@@ -68,6 +68,7 @@ func (sender *GossipSender) run() {
 			case pending := <-sender.cell:
 				if pending != nil {
 					sender.send(pending)
+					ch <- true
```
rade: we don't actually need the nil check here.

Also, I reckon we should send true or false, and *not* close the channel.

PR #903, opened 2015-06-12
In test/setup.sh, remove existing containers

bboreham:
Leaving containers lying around when doing a `service docker restart` seems to contribute to the symptom that docker takes ages to do the next removal. The old containers are in state "dead" if you do `docker ps -a`.

This pr simply repeats the code from `config.sh` to remove any containers known to docker before doing the restart.


rade:
Let's just ditch the docker restart instead.

rade:
> Let's just ditch the docker restart instead.

[done](https://github.com/weaveworks/weave/commit/97d692df809dba79473636c90abddf042f3b00c7)

PR #905, opened 2015-06-12
Use /10 instead of /9 for default iprange

bboreham:
... so it doesn't clash with CircleCI address

We were getting a lot of failures like this:

````
---= Running 150_connect_forget_test.sh =---
Connecting and forgetting routers after launch
network 10.128.0.0/9 would overlap with route 10.240.0.1/32
Default -iprange 10.128.0.0/9 overlaps with existing route.
````


rade:
Proof that our check works!!!

PR #906, opened 2015-06-12
DNS HTTP API listens on eth0 only

### LINKED ISSUE #899: http handler listens on all interfaces, which is bad
The DNS server's http handler is listening on all interfaces. In particular that means it is listening on the ethwe interface. That is bad since it makes sharing a subnet between the DNS servers and app containers more dangerous than it should be - apps could modify DNS entries.

The http handler should listen on eth0 / the docker-assigned IP only, at least by default.

Then, at least when full app isolation has been configured (see the bottom of the [isolation](http://docs.weave.works/weave/latest_release/features.html#application-isolation) section in our docs), the weaveDNS http interface is inaccessible to apps.

Even then apps on the same subnet as weaveDNS can modify it - by setting up an mdns responder that participates in the weaveDNS chatter. But that is more involved. And not easily preventable.

awh:
Fixes #899.

### DIFF
--- a/prog/weavedns/main.go
+++ b/prog/weavedns/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/weaveworks/weave/common/updater"
 	weavedns "github.com/weaveworks/weave/nameserver"
 	weavenet "github.com/weaveworks/weave/net"
+	"strconv"
 )
 
 var version = "(unreleased version)"
@@ -22,6 +23,7 @@ func main() {
 		apiPath         string
 		domain          string
 		dnsPort         int
+		httpIfaceName   string
 		httpPort        int
 		wait            int
 		ttl             int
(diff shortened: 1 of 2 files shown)


prog/weavedns/main.go
```diff
@@ -46,6 +48,7 @@ func main() {
 	flag.StringVar(&domain, "domain", weavedns.DefaultLocalDomain, "local domain (ie, 'weave.local.')")
 	flag.IntVar(&wait, "wait", 0, "number of seconds to wait for interface to be created and come up")
 	flag.IntVar(&dnsPort, "dnsport", weavedns.DefaultServerPort, "port to listen to DNS requests")
+	flag.StringVar(&httpIfaceName, "http-iface", "", "interface on which to listen for HTTP requests")
```
rade: No dashes, please. Yes, I know some have crept in, but for more exotic options only.

Should mention what the default is.

weave
```diff
@@ -1021,7 +1022,7 @@ case "$COMMAND" in
         DNS_CONTAINER=$(docker run --privileged -d --name=$DNS_CONTAINER_NAME \
             -p $DOCKER_BRIDGE_IP:53:53/udp \
             -v /var/run/docker.sock:/var/run/docker.sock \
-            $WEAVEDNS_DOCKER_ARGS $DNS_IMAGE -iface $CONTAINER_IFNAME "$@")
+            $WEAVEDNS_DOCKER_ARGS $DNS_IMAGE -iface $CONTAINER_IFNAME --http-iface "$DNS_HTTP_IFNAME" "$@")
```
rade: use a single dash, for consistency with the others.

prog/weavedns/main.go
```diff
@@ -74,15 +77,44 @@ func main() {
...
+			Error.Fatal("[main] No addresses on HTTP interface")
+		}
+
+		ip, _, err := net.ParseCIDR(addrs[0].String())
+		if err != nil {
+			Error.Fatal(err)
+		}
+
+		httpAddr = net.JoinHostPort(ip.String(), strconv.Itoa(httpPort))
+	}
```
rade: The `if` should just compute an `httpIP` string; and the joining with the port should happen afterwards.

prog/weavedns/main.go
```diff
@@ -74,15 +77,44 @@ func main() {
...
+	Info.Println("[main] HTTP API listening on", httpAddr)
+
```
rade: Hmm. Perhaps that should come right before the `net.Listen`? Or right after, so it gets only output on success? I can't remember what we do elsewhere. Worth checking.

PR #914, opened 2015-06-14
cope with peers having different gossip channels

### LINKED ISSUE #911: permit connectivity between peers that have different gossip channels
When weave receives gossip from an unknown channel, it [closes the connection with an error](https://github.com/weaveworks/weave/blob/eae9928e98d14557ec1f0e59b63d7b7f1d11b294/router/gossip.go#L149). This means we cannot introduce new channels without breaking protocol compatibility, which is rather inconvenient.

rade:
Closes #911.


rade:
I've tested this as follows:

````
host1> weave launch -nodiscovery -initpeercount 2
host2> weave launch -nodiscovery -initpeercount 2
host3> weave launch -iprange "" host1 host2
host1> weave run -ti ubuntu
host2> weave run -ti ubuntu
````

`weave status` on the third peer confirms that no IPAM is running, and that the topology is as intended i.e. "1 <-> 3 <-> 2".

Would be nice to have a smoke test for the above, but it requires three nodes.

I suppose I could add some unit tests, though they are not going to be as neat as the above :(

PR #918, opened 2015-06-15
Output error message if http call to weave or weavedns fails

### LINKED ISSUE #916: `weave run` terminates silently when weave is stopped during IP allocation
````
$ weave launch -initpeercount 2
$ weave run -ti ubuntu
[this blocks; now run `weave stop` in another shell, at which point it exits]
$ echo $?
52
````

running `WEAVE_DEBUG=1 weave run -ti ubuntu` shows that the weave script exits in `ipam_cidrs()` because it doesn't check the return status of the `http_call`s.

We have other `http_call`s in the code that are similarly unguarded, but they are less susceptible to stalling.

### PR DESCRIPTION
**Pull request by bboreham**
Fixes #916

Will print an error message if any of the calls to `weave` or `weavedns` fails after we've decided the container was running.

I've added `-S` to the curl options for all calls, so we get some explanation too.  In the example of #916, we now get:

````
curl: (52) Empty reply from server
Call to weave failed.
````

This implied an extra redirect to `/dev/null` in `wait_for_status` as we expect some of those calls to fail and don't want the user to see error messages.

### DIFF
--- a/weave
+++ b/weave
@@ -538,17 +538,21 @@ http_call_ip() {
     http_verb="$3"
     url="$4"
     shift 4
-    curl --connect-timeout 3 -s -X $http_verb "$@" http://$ip:$port$url
+    curl --connect-timeout 3 -s -S -X $http_verb "$@" http://$ip:$port$url
 }
 
 # Call url $4 with http verb $3 on container $1 at port $2
 http_call() {
-    container_ip $1 \
-        "$1 container is not present. Have you launched it?" \
-        "$1 container is not running." \
+    HTTP_CALL_CONTAINER=$1
+    container_ip $HTTP_CALL_CONTAINER \
+        "$HTTP_CALL_CONTAINER container is not present. Have you launched it?" \
+        "$HTTP_CALL_CONTAINER container is not running." \
         || return 1
     shift 1
-    http_call_ip $CONTAINER_IP "$@"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: weave [hunk 1]
```diff
@@ -547,16 +547,20 @@ http_call() {
         "$1 container is not present. Have you launched it?" \
         "$1 container is not running." \
         || return 1
+    HTTP_CALL_CONTAINER=$1
```
rade: I'd move that to the top and use it throughout.

weave
```diff
@@ -547,16 +547,20 @@ http_call() {
...
-    http_call_ip $CONTAINER_IP "$@"
+    if ! http_call_ip $CONTAINER_IP "$@" ; then
+        echo "Call to $HTTP_CALL_CONTAINER failed." >&2
+        return 1
+    fi
 }
 
 call_weave() {
-    http_call $CONTAINER_NAME $HTTP_PORT "$@"
+    http_call $CONTAINER_NAME $HTTP_PORT "$@" --show-error
```
rade: Let's just change the `curl` to use `-S` insead of `-s`. Any place that doesn't want the error output can `2>/dev/null` it.

bboreham: You actually need both `-s -S` to get (no progress messages) and (error messages)

bboreham:
Update in line with review comments.

PR #920, opened 2015-06-15
forward/backward compatible protocol versioning

### LINKED ISSUE #871: build forward and backward compatibility into the protocol
ATM weave peers must all be running the same version of the protocol; both sides of a new connection will shut it down when detecting that the protocol version of the remote peer differs from their own.

This makes upgrading weave when the protocol version has changed a major event.

We should build some mechanisms into the protocol for handling forward and backward compatibility.

Extracted from https://github.com/weaveworks/weave/issues/403#issuecomment-104048264

rade:
Closes #871.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -41,6 +41,7 @@ type LocalConnection struct {
 	sync.RWMutex
 	RemoteConnection
 	TCPConn           *net.TCPConn
+	version           byte
 	tcpSender         TCPSender
 	remoteUDPAddr     *net.UDPAddr
 	receivedHeartbeat bool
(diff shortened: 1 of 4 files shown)


rade:
A is old weave, B is new weave
- A says "connection shutting down due to error during handshake: extra data in buffer"
- B says "connection shutting down due to error during handshake: unrecognised protocol header: [14 255 129 4 1 2 255]"

A is [1,1], B is [2,2]
- A says "connection shutting down due to error during handshake: remote version [2,2] is incompatible with our version [1,1]"
- B says "connection shutting down due to error during handshake: remote version [1,1] is incompatible with our version [2,2]"

A is [0,1], B is [2,3]
- A says "connection shutting down due to error during handshake: remote version [2,3] is incompatible with our version [0,1]"
- B says "connection shutting down due to error during handshake: remote version [0,1] is incompatible with our version [2,3]"

A is [1,1], B is [1,1]
- A says "completed handshake; using protocol version 1"
- B says "completed handshake; using protocol version 1"

A is [0,2], B is [1,3]
- A says "completed handshake; using protocol version 2"
- B says "completed handshake; using protocol version 2"

dpw:
The general scheme looks ok.   I added a couple of line notes.  Do you intend to fold the two commits together?  The second one overrides so much of the first that the first one seems like a detour.

rade:
> Do you intend to fold the two commits together?

Will squash them, together with the other changes.

PR #932, opened 2015-06-16
Kill the container we just started if we fail to allocate an IP address

### LINKED ISSUE #921: `weave launch-dns|start` w/o running weave reports error but still launches
````
$ weave launch-dns
weave container is not present. Have you launched it?
$ weave launch-dns
weavedns is already running.
````

Ditto for `weave start` and `weave --local run`.

The failure is in attempting to obtain an IP address.

In #831 we fixed a similar issue wrt network configuration by introducing a `with_container_netns*_or_die*`. Perhaps we need a `ipam_cidrs_or_die`.

### PR DESCRIPTION
**Pull request by bboreham**
Fixes #921 and #889

### DIFF
--- a/weave
+++ b/weave
@@ -788,7 +788,9 @@ ipam_cidrs() {
     IPAM_CIDRs=""
     if [ $# -eq 0 ] ; then
         # If no addresses passed in, get one in the default subnet
-        IPAM_CIDRS=$(call_weave POST /ip/$CONTAINER_ID)
+        if ! IPAM_CIDRS=$(call_weave POST /ip/$CONTAINER_ID) ; then
+            return 1
+        fi
         if [ "$IPAM_CIDRS" = "404 page not found" ] ; then
             echo "No IP address supplied (use the -iprange option on 'weave launch' to enable IP address allocation)" >&2
             return 1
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: weave [hunk 1]
```diff
@@ -788,7 +788,9 @@ ipam_cidrs() {
     IPAM_CIDRs=""
     if [ $# -eq 0 ] ; then
         # If no addresses passed in, get one in the default subnet
-        IPAM_CIDRS=$(call_weave POST /ip/$CONTAINER_ID)
+        if ! IPAM_CIDRS=$(call_weave POST /ip/$CONTAINER_ID) ; then
```
paulbellamy: Isn't this the same as just: `IPAM_CIDRS=$(call_weave POST /ip/$CONTAINER_ID)`

rade:
Should the proxy now call `weave attach --or-die` instead of doing its own cleaning up?

weave
bboreham: No; in the absence of anyone looking at the return value from `call_weave` its return value is overwritten by the next command, i.e. `[`.
This doesn't matter when `set -e` is in effect, but my next change calls `call_cidr` in an if-statement so removing the instant-exit behaviour.

paulbellamy:
+1 for proxy using `weave attach --or-die`

weave
paulbellamy: Ah, right. Thanks.

PR #935, opened 2015-06-16
Fix all reports from `go test -race` in weavedns unit tests

bboreham:
Most are artefacts of the test code, but having so many reports made it difficult to check for any real problems.

The only one which looks like a real problem is a34d2ec45c5d4065010f9ce035a270313a8cd228, though aa85e19256c6bceb3fa2b0f82a45ea033787e7ac could conceivably result in a crash.

### DIFF
--- a/nameserver/mdns_client.go
+++ b/nameserver/mdns_client.go
@@ -128,12 +128,13 @@ func (c *MDNSClient) Start(ifi *net.Interface) (err error) {
 		}
 	}
 
-	c.listener = &dns.Server{Unsafe: true, PacketConn: multicast, Handler: dns.HandlerFunc(handleMDNS)}
-	go c.listener.ActivateAndServe()
-
 	actionChan := make(chan MDNSAction, MailboxSize)
 	c.actionChan = actionChan
 	go c.actorLoop(actionChan)
+
+	c.listener = &dns.Server{Unsafe: true, PacketConn: multicast, Handler: dns.HandlerFunc(handleMDNS)}
+	go c.listener.ActivateAndServe()
+
 	return nil
 }
 
--- a/nameserver/mdns_client.go
+++ b/nameserver/mdns_client.go
@@ -97,7 +97,6 @@ type MDNSClient struct {
 	addr       *net.UDPAddr
 	inflight   map[string]*inflightQuery
 	actionChan chan<- MDNSAction
-	running    bool
 }
 
 type mDNSQueryInfo struct {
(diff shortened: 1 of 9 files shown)


rade:
linter is unhappy.

rade:
@inerica aside from the linter issue, are you happy with the rest?

nameserver/mdns_client.go
```diff
@@ -272,13 +265,12 @@ func (c *MDNSClient) checkInFlightQueries() time.Duration {
...
 	run := func() { timer.Reset(c.checkInFlightQueries()) }
-	c.running = true
-	for c.running {
+	for {
 		select {
 		case action := <-actionChan:
 			if action == nil {
 				c.listener.Shutdown()
-				c.running = false
+				return
```
inercia: This changes the logic. By returning instead of breaking the loop, the code below is never executed, and response channels are never closed (probably not important because the client is going to be destroyed and the channels will be closed anyway, but in any case...)

nameserver/server.go
```diff
@@ -194,17 +206,6 @@ func (s *DNSServer) Start() error {
...
-	if err != nil {
-		return err
-	}
-	s.pc = pc
-
-	_, port, err := net.SplitHostPort(pc.LocalAddr().String())
-	if err != nil {
-		return err
-	}
-	s.ListenAddr = fmt.Sprintf(":%s", port)
```
inercia: Is this change really necessary? Do we need to move the `Listen()` to the `New()` method? I've never been fond of mixing strict initialization tasks (usually performed in `New()` methods, like memory allocations, checking parameters, etc...) with things that have side effects (an are usually delayed until we really `Start()`). I think we do not call `Start()` from multiple threads, do we?

inercia:
@rade I've added a couple of comments....

nameserver/server.go
bboreham: Yes, we call `Start()` on a different goroutine, because it blocks.  Hence a race condition is reported, although it is not likely to do harm.

Not clear that creating a listener can be called a side-effect. It's an essential part of the server.

Maybe separating out a third `ListenAndServe()` routine which blocked would make it clearer?

nameserver/mdns_client.go
bboreham: Good point, thanks.

nameserver/server.go
inercia: Well, I think that `New*()` methods should not try to be too smart. For example, imagine you create a `NewMyDatabaseClient()`: I don't think it should create a connection pool until you `Start()` or `Connect()` or something else... It is a pattern I always try to follow, but this is my personal opinion... 

And, yes, I've thought several times about refactoring some of our components from `Start()`/`Stop()` to  the `ListenAndServe()`/`Shutdown()` _standard_ pattern. That would make things more clear, a much better API, but I've never done it because it would also break the convenient [SignalReceiver interface](https://github.com/weaveworks/weave/blob/master/common/signals.go#L11) found in `common/signal.go`...

bboreham:
I re-did the last commit with an `ActivateAndServe()` method, and rebased the commits to remove detours.

inercia:
Thanks, @bboreham, it looks great now! Everything looks good to me now. Please merge it if all tests pass...

PR #948, opened 2015-06-17
Update documentation to leverage IPAM and DNS

### LINKED ISSUE #471: make use of DNS throughout the documentation
ATM the weaveDNS documentation is completely separate from the main documentation, with just the odd reference to it from the latter. This was desirable when weaveDNS was new and experimental, but now it actually hinders adoption.

The weave 1.0 walkthrough in #242 presents a much more integrated view of weaveDNS, and hence can serve as guidance for this issue, though we won't be able to go quite as far until a few more features have landed.

### LINKED ISSUE #869: make use of default IPAM throughout documentation

awh:
@rade @bboreham I'm not expecting this to be actually merged yet, I'm just using it to make changes available to you ASAP to get early feedback; I'll continue pushing self contained updates to other areas of the documentation + incorporate your feedback as you make it.

This PR updates just the main README initially - examples look much much better! Any advance on the rather boring h1c1/h2c1 container names? They're a bit mechanistic... something fun that we can carry through as a theme would be better IMO.

Fixes #869. Fixes #471. (eventually...)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ On `$HOST1` run
 
     host1$ weave launch
     host1$ weave launch-dns
-    host1$ weave run --name h1c1 -t -i ubuntu
+    host1$ weave run --name h1c1 -ti ubuntu
 
 The first two lines start the weave router and DNS, each inside their
 own container - this needs to be done once per host. Whilst the
@@ -81,7 +81,7 @@ Next we repeat similar steps on `$HOST2`...
 
     host2$ weave launch $HOST1
     host2$ weave launch-dns
-    host2$ weave run --name h2c1 -t -i ubuntu
+    host2$ weave run --name h2c1 -ti ubuntu
 
 The only difference, apart from the name of the application container,
 is that we tell our weave that it should peer with the weave on
--- a/README.md
+++ b/README.md
@@ -55,47 +55,41 @@ two containers, one on each host.
 On `$HOST1` run
 
     host1$ weave launch
-    host1$ C=$(weave run 10.2.1.1/24 -t -i ubuntu)
+    host1$ weave launch-dns
+    host1$ weave run --name h1c1 -t -i ubuntu
 
-The first line starts the weave router, in a container. This needs to
-be done once on each host. The required docker image for the weave
-router container is downloaded automatically. There is also a `weave
-setup` command for downloading this and other images required for
+The first two lines start the weave router and DNS, each inside their
+own container - this needs to be done once per host. Whilst the
+required docker images are downloaded automatically, there is also a
+`weave setup` command for downloading all the images required for
 weave operation; this is a strictly optional step which is especially
 useful for automated installation of weave and ensures that any
 subsequent weave commands do not encounter delays due to image
 downloading.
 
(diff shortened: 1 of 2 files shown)


bboreham:
For the README I think c1 and c2 would work fine.  Is there a point where there is an h1c2 or h2c2?

README.md
```diff
@@ -56,47 +56,41 @@ two containers, one on each host.
...
+    host1$ weave run --name h1c1 -t -i ubuntu
 
-The first line starts the weave router, in a container. This needs to
-be done once on each host. The required docker image for the weave
-router container is downloaded automatically. There is also a `weave
-setup` command for downloading this and other images required for
+The first two lines start the weave router and DNS, each inside their
+own container - this needs to be done once per host. Whilst the
+required docker images are downloaded automatically, there is also a
+`weave setup` command for downloading all the images required for
```
bboreham: `weave setup` could go elsewhere; it is not a key feature to learn about on first contact.

awh:
I've updated the manual addressing feature so it's a replay of the first example - this means that the following application subnet isolation makes sense without changes (unless there is a way that can be IPAMified - thoughts?)

site/features.md
```diff
@@ -55,6 +56,49 @@ now re-use the same tools and techniques when deploying applications
 as containers as we would have done when deploying them 'on metal' in
 our data centre.
 
+### <a name="manual-addressing"></a>Manual addressing
+
+The examples so far have leveraged weave's [address allocation](#ipam)
+and [discovery](#dns) features, meaning that we've not had to concern
+ourselves with IP addresses at all. It is however possible to specify
```
bboreham: "It is, however, possible"
but "possible" plus "if you want" seems overkill.

bboreham:
You can request IPAM allocate an address in a subnet, e.g. `net:10.180.0.0/16`

site/features.md
```diff
@@ -174,31 +232,38 @@ anywhere.
...
 
-    host2$ weave expose 10.2.1.102/24
+    host2$ weave expose
 
-choosing an unused IP address in the application subnet. (There is a
-corresponding 'hide' command to revert this step.)
+(There is a corresponding 'hide' command to revert this step.)
 
-Now
+Now,  after finding allocated IPs via `weave ps`
```
awh: This is a bit janky, but weaveDNS names won't resolve from the host...

bboreham:
LGTM, apart from the names h1c2 etc., which I still find ugly.

awh:
@rade suggests a1, a2 and b1, b2 (letter == subnet, number == host) - thoughts @bboreham?

bboreham:
Better.  Or you could start with c1, c2 as before and then move to b1, b2 when you have a second subnet?

PR #955, opened 2015-06-17
Replacing custom testing helpers with testify, where possible

### LINKED ISSUE #358: replace homegrown testing library with something off-the-shelf
Our [testing util](https://github.com/zettio/weave/blob/master/testing/util.go) library is growing. It's still quite small, but at some point we should probably bite the bullet and pull in something like [gocheck](http://labix.org/gocheck) or [testify](https://github.com/stretchr/testify).

paulbellamy:
Closes #358

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
   - pip install --user --upgrade gcloud gsutil
   - bin/setup-ci-secrets $encrypted_5ba036e89377_key $encrypted_5ba036e89377_iv
   - make travis
+  - go get ./...
 
 script:
   - bin/lint .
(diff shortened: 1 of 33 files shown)


rade:
@inercia @awh would appreciate you taking a look at this.

I have suggested to @paulbellamy that we should file PRs against testify for adding the two helpers we are currently left with.

inercia:
Looks good to me. It's great to use someone's else library...

inercia:
Regarding `testify` missing the _stacktraces_, it would be great if we could be more general and call any user-provided callback when `Fatalf` is invoke. This callback could be a function for dumping the stack trace, but it could also be a function for dumping all the _goroutines_ stacks, stop the debugger, etc...

PR #964, opened 2015-06-18
Update documentation to leverage proxy

### LINKED ISSUE #951: make use of proxy throughout documentation
Leverage the proxy in examples so we can relegate `weave run`

### LINKED ISSUE #965: Revise proxy, IPAM and DNS documentation
Now that the main README example and features documentation uses proxy/IPAM/DNS pervasively, go through the dedicated pages for those components and:

* Remove any duplication that has arisen
* Convert them to use proxy/IPAM/DNS as appropriate
* Document `weave run` as an exception

awh:
Fixes #951.
Fixes #965.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,36 +52,30 @@ Say you have docker running on two hosts, accessible to each other as
 `$HOST1` and `$HOST2`, and want to deploy an application consisting of
 two containers, one on each host.
 
-On `$HOST1` run
+First start weave on $HOST1:
 
-    host1$ weave launch
-    host1$ weave launch-dns
-    host1$ weave run --name a1 -ti ubuntu
+    host1$ weave launch && weave launch-dns && weave launch-proxy
 
-The first two lines start the weave router and DNS, each inside their
-own container - this needs to be done once per host. Whilst the
-required docker images are downloaded automatically, there is also a
-`weave setup` command for downloading all the images required for
-weave operation; this is a strictly optional step which is especially
-useful for automated installation of weave and ensures that any
-subsequent weave commands do not encounter delays due to image
-downloading.
+this runs the weave router, DNS and proxy, each in their own
+container. Next we configure our `DOCKER_HOST` environment variable to
+point to the proxy, so that containers launched via the docker command
+line are automatically attached to the weave network:
 
-The third line runs our application container, allocating it an IP
-address on the weave network and registering it in DNS.  `weave run`
-invokes `docker run -d` behind the scenes, so we could be launching
-any container this way; here we just take a stock ubuntu container and
-launch a shell in it.  There's also a `weave start` command, which
-invokes `docker start` for starting existing containers.
+    host1$ eval $(weave proxy-env)
 
-If our application consists of more than one container on this host we
-simply launch them with a variation on third second line.
+Finally we run our application container; this happens via the proxy
+so it is automatically allocated an IP address and registered in DNS:
+
+    host1$ docker run --name a1 -ti ubuntu
+
+That's it! If our application consists of more than one container on
+this host we simply launch them with `docker run` as appropriate.
 
 Next we repeat similar steps on `$HOST2`...
 
-    host2$ weave launch $HOST1
-    host2$ weave launch-dns
-    host2$ weave run --name a2 -ti ubuntu
+    host2$ weave launch $HOST1 && weave launch-dns && weave launch-proxy
+    host2$ eval $(weave proxy-env)
+    host2$ docker run --name a2 -ti ubuntu
 
 The only difference, apart from the name of the application container,
 is that we tell our weave that it should peer with the weave on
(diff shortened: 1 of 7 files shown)


rade:
There are still a fair few references to `weave run` in the dns docs. We need to excise most of them and also figure out where to introduce `weave run` - should go into the proxy docs, perhaps, e.g. in a "running containers without the proxy" section.

We should use the `weave launch && weave launch-dns && weave launch-proxy` short version everywhere all three are launched - a few places in the DNS docs dont' do that.

awh:
> There are still a fair few references to weave run in the dns docs

The only ones that are left now are those documenting things specific to `weave run` e.g. overriding DOCKER_BRIDGE and all the --dns-* options...

rade:
There's a whole section titled "Using weaveDNS without weave run". Perhaps that should be "...without the proxy or `weave run`"

awh:
> Perhaps that should be "...without the proxy or weave run"

Or excised altogether? It made sense when we didn't have the proxy, but it's looking something of a corner case now. We already mention that the DNS listens on the docker bridge IP so anyone trying to do something this obscure can presumably work out they need to do `--dns=$DOCKER_IP` - thoughts?

rade:
The new `weave run` section in the proxy docs should mention the differences w.r.t. `docker run`, namely that containers are always started in the background and that the weave network interface may not be immediately available on startup.

The section should also be the (only) place that mentions `weave start`.

README.md
```diff
@@ -52,36 +52,30 @@ Say you have docker running on two hosts, accessible to each other as
...
-launch a shell in it.  There's also a `weave start` command, which
-invokes `docker start` for starting existing containers.
-
-If our application consists of more than one container on this host we
-simply launch them with a variation on third second line.
+First start weave on $HOST1:
+
+    host1$ weave launch && weave launch-dns && weave launch-proxy
+
+this pulls and runs the router, DNS and proxy containers. Next we
```
rade: - we have lost the "this needs to be done once per host". Do you think that's obvious enough?
- "this runs the weave router, DNS and proxy, each in their own container".
- we do need to mention `weave setup` somewhere in our docs. It's quite an important command, used by many automated weave deployments.

README.md
```diff
@@ -103,7 +97,6 @@ can talk to each other...
 
 On `$HOST1`...
```
rade: "In the container started on $HOST1" - yes, if you actually run the example it is obvious that the following command runs inside the container, but that's not so clear when just casually reading the docs.

README.md
```diff
@@ -112,7 +105,6 @@ On `$HOST1`...
 
 Similarly, on `$HOST2`...
```
rade: ditto

rade:
- "Working with the Docker CLI and API" appears as a heading/link in a few places. How about "Configuring the interaction with Docker"?
- the multi host example in the proxy docs is the same as the main example now, isn't it? In which case we should remove that whole section.

rade:
In features/manual_addressing, we have "including those delegated to IPAM" - this is the first mentioning of IPAM! Replace with "...to weave's [automatic IP address allocator](ipam.html)"

rade:
In features/dynamic_network_attachment we have "In some scenarios containers are started independently, e.g. via some existing tool chain, or require more complex startup sequences than
provided by `weave run`. And sometimes the decision which application network a container should be part of is made post-startup. For these situations, weave", which still mentions `weave run`. I suggest getting rid of the first sentence and just saying "Sometimes the decision ..."

rade:
features/application_isolation uses "net:default" without explanation

rade:
I suggest we git rid of the IPAM, DNS and proxy sections at the end of features and replace them with a single "Further reading" (better title welcome!) section that says something along the lines of "We have made use of weave's automatic IP address allocation, DNS server, and proxy throughout the examples. Each of these components offers additional functionality, which is explained in ... links".

rade:
The proxy/automatic_discovery section is a bit dense. Should at least be broken into two paragraphs, and ideally reworded somewhat since right now it reads like a giant nested 'if' statement.

awh:
> The proxy/automatic_discovery section is a bit dense. Should at least be broken into two paragraphs, and ideally reworded somewhat since right now it reads like a giant nested 'if' statement.

This is actually quite similar to the old 'usage' section of the weaveDNS doc which I rewrote from scratch for the same reason. There's no point having it in two places, I'll reduce this to a pointer to the weaveDNS docs.

site/proxy.md
```diff
@@ -218,6 +174,33 @@ which is exactly the same configuration as when connecting to the
...
+## <a name="without-proxy"></a>Launching containers without the proxy
+
+If you cannot or do not want to use the proxy you can launch
+containers on the weave network with `weave run`:
+
+    $ weave run -ti ubuntu
+
+The arguments after `run` are passed through to `docker run` so you
+can freely specify whichever docker options are appropriate. Once the
+container is started `weave run` attaches it to the weave network, in
```
rade: missing comma: "Once the container is started, `weave run` ..."

site/weavedns.md
```diff
@@ -254,60 +232,13 @@ can force weaveDNS to use a different domain by launching it
...
 ```bash
-$ weave launch-dns 10.2.254.1/24 --domain="mycompany.local."
+$ weave launch-dns --domain="mycompany.local."
 ```
 
 The local domain should end with `local.`, since these names are
 link-local as per [RFC6762](https://tools.ietf.org/html/rfc6762),
 (though this is not strictly necessary).
 
-## <a name="without-run"></a>Using weaveDNS without `weave run`
```
rade: The link from the ToC is still there. Also check there isn't a reference to it from elsewhere.

rade:
the weavedns/domain-search-path section still uses `weave run`. it can be removed from there, and the proxy used instead.

rade:
boot2docker, how-it-works and ipam have one reference to`weave run` each. I think they can all be eliminated.

awh:
Have updated everything except the boot2docker docs (the changes will be a bit fiddly to get right due to the `eval $(weave proxy-env)` step, and I have no way of testing them - boot2docker is OSX/Windows only). Is that page even needed? I would have thought it obvious that if you're using boot2docker, that $HOST1/$HOST2 etc must be boot2docker VMs...

rade:
LGTM

Please
1. file an issue for updating the boot2docker docs - I don't think we should get rid of them
2. do some squashing
3. merge into lastest_release_doc_updates, and from there into the 1.0 branch, and from there master

awh:
Done squashing; will close this PR and merge manually as requested.

PR #969, opened 2015-06-20
include image name in errors

### LINKED ISSUE #967: proxy complains about `no such image` when starting containers that require image downloading
while using weave proxy
for example if the docker host is clean and doesn't have an ubuntu image there
if i run `docker run -it ubuntu`
a message saying `Error response from daemon: no such image` is being returned
might be while facing this error, proxy could try to pull the image instead?

rade:
The go docker client's `ErrNoSuchImage` does not include the image name. However, it is imperative that this name appears in error messages sent to clients, since the docker client, as of version
1.7.0, only automatically pulls images when their name is present in the error.

Fixes #967.

### DIFF
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -21,6 +21,17 @@ type createContainerRequestBody struct {
 	MacAddress string             `json:"MacAddress,omitempty" yaml:"MacAddress,omitempty"`
 }
 
+// Replacement for docker.NoSuchImage, which does not contain the
+// image name, which in turn breaks docker clients post 1.7.0 since
+// they expect the image name to be present in errors.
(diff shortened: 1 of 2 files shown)


squaremo:
Why is this a PPR into 1.0.0 and not master?

rade:
It's a PR on 1.0.0 because it is going to get released in a 1.0.1. So it should have been merged into 1.0.0 and then from there into master.

squaremo:
Ops, I didn't get that memo. Does it matter if it happens the other way around?

rade:
> Does it matter if it happens the other way around?

Yes. Because stuff has landed on master since 1.0.0.

Why not merge this PR as intended, and then merge 1.0.0 into master (which should be a no-op)?

PR #977, opened 2015-06-22
Change one IPAM proxy example to use a network

bboreham:
Previously, no proxy examples used the `net:` syntax, and this particular example duplicated the one 18 lines above.

### DIFF
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -91,9 +91,9 @@ special environment variables or other options*.
 
     host1$ docker run -ti ubuntu
 
-To use a specific IP, we pass a `WEAVE_CIDR` to the container, e.g.
+To use a specific subnet, we pass a `WEAVE_CIDR` to the container, e.g.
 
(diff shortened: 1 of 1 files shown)


rade:
Hmm. perhaps the `net:` should also feature in the "Multiple IP addresses and networks can be supplied..." example earlier on?

bboreham:
If you do that it becomes tenuous why we would have an 'Automatic IP address assignment' heading afterwards.  So bigger re-org needed.

awh:
Only comment is that we've used `10.2.1.0/24` as the default `net:10.2.2.0/24` as an alternative subnet elsewhere (e.g. features docs on subnet isolation and dynamic attach), so you might want to use 10.2.2.0/24 instead of 10.2.9.0/24 here; otherwise LGTM.

We appear to be operating a process for doc updates where author merges the PR themselves post LGTM then manually merges that from latest_release_doc_updates -> 1.0 and then 1.0 -> master so please go ahead!

bboreham:
Actually, without any `-iprange` on this page, we should use a `net:` in the 10.128/10 range.

PR #980, opened 2015-06-22
[proxy] Only prepend weavewait if not already present

### LINKED ISSUE #979: weaveWaitEntrypoint is appended multiple times
file: https://github.com/weaveworks/weave/blob/v1.0.1/proxy/create_container_interceptor.go
function: `setWeaveWaitEntrypoint`
line 92: `container.Entrypoint = append(weaveWaitEntrypoint, container.Entrypoint...)`

I haven't had any experiences with golang,
but I believe this line allows the `/w/w` entrypoint to be inserted at the front for multiple times.

I am using dokku-alt with weave & weave proxy,
I could see the entrypoint of container launching with dokku becomes `/w/w /w/w /w/w /w/w /bin/bash`.

It's probably because in `setWeaveWaitEntrypoint`,
it's trying to insert `/w/w` to the container's entrypoint from docker inspect,
however the container has already included `/w/w` before.

paulbellamy:
This way `docker commit` will not stack up weavewaits.

Fixes #979

### DIFF
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -89,7 +89,10 @@ func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container *docker.Co
 		}
 	}
 
-	container.Entrypoint = append(weaveWaitEntrypoint, container.Entrypoint...)
+	if len(container.Entrypoint) == 0 || container.Entrypoint[0] != weaveWaitEntrypoint[0] {
+		container.Entrypoint = append(weaveWaitEntrypoint, container.Entrypoint...)
+	}
+
 	return nil
 }
 
--- a/test/610_proxy_wait_for_weave_test.sh
+++ b/test/610_proxy_wait_for_weave_test.sh
@@ -2,6 +2,10 @@
 
 . ./config.sh
 
+entrypoint() {
+  docker_on $HOST1 inspect --format="{{.Config.Entrypoint}}" "$@"
+}
+
 start_suite "Proxy waits for weave to be ready before running container commands"
 weave_on $HOST1 launch-proxy
 BASE_IMAGE=busybox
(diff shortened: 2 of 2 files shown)


proxy/create_container_interceptor.go
```diff
@@ -71,25 +70,30 @@ func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
...
+		if config.Entrypoint == nil {
+			config.Entrypoint = image.Config.Entrypoint
 		}
 	}
 
-	container.Entrypoint = append(weaveWaitEntrypoint, container.Entrypoint...)
+	if len(config.Entrypoint) == 0 || config.Entrypoint[0] != weaveWaitEntrypoint[0] {
+		config.Entrypoint = append(weaveWaitEntrypoint, config.Entrypoint...)
+	}
+	container.HostConfig.VolumesFrom = append(container.HostConfig.VolumesFrom, "weaveproxy")
```
rade: I'd rather leave the volume setting where it was, so the function can continue to take just a `*docker.Config`.

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -12,4 +12,10 @@ fi
 
 assert_raises "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' $BASE_IMAGE $CHECK_ETHWE_UP"
 
+# Check committed containers only have one weavewait prepended
+LAST_CONTAINER=$(proxy docker_on $HOST1 ps -a -q -n=1)
```
rade: give the container a name instead, then you won't need this.

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -12,4 +12,10 @@ fi
...
+COMMITTED_IMAGE=$(proxy docker_on $HOST1 commit $LAST_CONTAINER)
+assert_raises "proxy docker_on $HOST1 run --name c2 $COMMITTED_IMAGE"
+assert "proxy docker_on $HOST1 inspect --format='{{.Config.Entrypoint}}' c2" "{[/w/w]}"
```
rade: This is too low level. Nothing in the current tests knows that the entrypoint is called `w/w`. We should keep it that way.

Can we observe some difference in the behaviour, instead?

rade: or perhaps check that the entrypoint of this container is the same as its parent?

rade:
LGTM. Please merge, and the merge 1.0 into master.

PR #991, opened 2015-06-23
Fast datapath preparations

dpw:
These are the changes to the existing router code to prepare the way for fast datapath.  I didn't include the short peer ID changes, as they are much more tailored to ODP specifics, and so they make more sense as part of the fast datapath PR.

I think this is clean enough for a review pass.  There is still one point I'm not quite happy with yet:  The names "IntraHost" and "InterHost" were intended as stand-ins, until I came up with better names.  But I still haven't come up with replacements that seem obviously better (the best so far is PacketCapture and PacketTransport but they are longer, and PacketCapture isn't right because that interface includes sending, plus PacketCapture has too much affinity with pcap).  I'm hoping inspiration will strike by the time this is merge-ready, and welcome any suggestions.


dpw:
(github is having a hard time dealing with the extent of the rebasing that has gone on in this branch.  It's probably easier to look at the branch itself.)

dpw:
I've become concerned that 9e2a79d816 adds more information to the unencrypted handshake.  In fact, very little of the information in the handshake is required to establish encryption.  The downside of moving that information out of the initial handshake is that, in the encrypted case, we will need a second encrypted handshake (in the unencrypted case they can be combined).  But I'm not sure that avoiding an extra round trip is justified in this context.

rade:
> I've become concerned that 9e2a79d adds more information to the unencrypted handshake.

Don't be. Let's fix https://github.com/weaveworks/weave/issues/1029.

rade:
Any objections to me cherry-picking be13eaf and 80e6cbe onto master?

dpw:
> Any objections to me cherry-picking be13eaf and 80e6cbe onto master?

No.  (I think it's unnecessary, but you knew that already.)

rade:
> cherry-picking be13eaf and 80e6cbe

Done.

rade:
re the `PeerSummary` change... (putting this here so it doesn't get lost in a rebase)...

- can we come up with a better name than `PeerSummary`? That name made sense when the summary was used only for encoding, but it looks really odd for `Peer` to contain `PeerSummary`.
- what all the test code *really* wants is a `Peer.NewPeerFrom(p *Peer)`
- `NewPeer` should call `NewPeerFromSummary`

dpw:
> can we come up with a better name than PeerSummary? That name made sense when the summary was used only for encoding, but it looks really odd for Peer to contain PeerSummary.

I don't really see this.   It is to be expected that a PeerSummary can be derived from a Peer; in this case, that happens to be by way of a straightforward projection, so we can express it concisely in go (for once).  That seems quite natural and logical to me, so I might have difficulty coming up with another name.

router/router.go
```diff
@@ -48,24 +49,21 @@ type Router struct {
...
-	ReadPacket() ([]byte, error)
-}
+func NewRouter(config Config, name PeerName, nickName string) *Router {
+	router := &Router{Config: config, gossipChannels: make(GossipChannels)}
 
-type PacketSink interface {
-	WritePacket([]byte) error
-}
+	// If the caller didn't set the intrahost, replace it with a
+	// null implementation
```
rade: That comment is stating the obvious. Remove.

router/sleeve.go
```diff
@@ -0,0 +1,1121 @@
...
+		// destination MAC address?  Because even
+		// non-broadcast frames can be broadcast, if the
+		// destination MAC was not in our MAC cache.
+		if broadcast {
+			log.Printf("dropping too big DF broadcast frame (%v -> %v): PMTU= %v\n", dec.IP.DstIP, dec.IP.SrcIP, effectivePMTU)
+			return
+		}
+
+		// Send an ICMP back to where the frame came from
+		fn, err := dec.makeICMPFragNeeded(effectivePMTU)
```
rade: Several times now I've stumbled across this code and initially thought `fn` stands for `function`. `fragNeededFrame`?

rade: or `fragNeededPacket`, given that you call it a "packet" in the comment below.

rade:
I still cringe every time I read `IntraHost` and `InterHost`. Perhaps @squaremo, our naming czar, can come up with something better?

dpw:
> I still cringe every time I read IntraHost and InterHost. Perhaps @squaremo, our naming czar, can come up with something better?

As I noted, they were and are intended to be stand-ins.  Unless you had said "actually they seem fine to me", my plan is to change them.

rade:
*when*? I am close to being happy with the `IntraHost` changes, *except* for the name.

dpw:
Shrug.  It's just a naming thing, so if it is the last thing I fix, that's fine.  It's not like it's the only thing keeping this from being ready to merge.

rade:
I was planning to cherry-pick the IntraHost change.

dpw:
You'll have to wait until the branch as a whole is ready.

rade:
Are you planning to make changes *other* than in response to my review comments?

dpw:
Of course, if you have further feedback on the other commits, that would be much appreciated.  Whether intrahost is renamed or cherry picked or not will not alter their substance.

dpw:
> Are you planning to make changes other than in response to my review comments?

The "features" commit will change, or possibly go away entirely, due to the handshake changes I'm working on.

Other than that: No, I don't have any further changes planned.

rade:
I may need something like https://github.com/weaveworks/weave/blob/master/docs/architecture.txt to get through this. Ideally with some diagrams.

router/pcap.go
```diff
@@ -1,19 +1,39 @@
...
+	bufSz int
+
+	// The libpcap handle for writing packets. It's possible thata
+	// single handle could be used for reading and writing, but
+	// we'd have to examine the performance implications.
+	writeHandle *pcap.Handle
+
+	// pcap handles are single-threaded, so we need to lock around
+	// uses of writeHandle.
+	writeHandleMutex sync.Mutex
```
rade: I've been staring at the code for half an hour now and still cannot work out where the multiple goroutines come from that call `Send`. Or did you introduce the lock to make the API thread-safe *just in case*?

dpw:
The spirit of the code hasn't changed, and I have tried to make it more readable than before (this is not to criticise the readability before, but the UDP forwarding code was scattered throughout the codebase, and some of the objects / goroutines / layers have also been consolidated).  So if it didn't require a document before, and does now, I have failed, and I will need to figure out how.  If you can point me to something that seems obscure, just as a starting point, that would be great.

Alternatively, if what you are saying is "that new code might be perfectly readable, but it would *always* have been nice if we had a document describing how the UDP forwarding works" I'm not sure if I would agree or not.  But maybe the fact that we have got by without such a thing so far suggests that the need cannot be that pressing?  In which case it can be filed as a separate issue.

rade:
I am finding the control flow much harder to follow that before. Part of that is obviously due to familiarity with the previous code base.

router/sleeve.go
```diff
@@ -0,0 +1,1121 @@
...
+
+// A "special" message over UDP, or a control message.  The sender is
+// nil for control messages.
+type specialFrame struct {
+	sender *net.UDPAddr
+	frame  []byte
+}
+
+func (sleeve *SleeveInterHost) MakeForwarder(
+	params ForwarderParams) (InterHostForwarder, error) {
```
rade: curious line break there. Remove.

router/pcap.go
dpw: In the commit in which that change is introduced (69318506975), they correspond to the two uses of `router.IntraHost.InjectPacket` in `router.go`:  One is in the packet capturing goroutine, the other is in the udp listener goroutine.  This structure is as before, but before the packet capturing goroutine used the "reading" pcap handle to send, whereas now both cases use the "writing" handle, to avoid cluttering the `IntraHost` interface.

There are two corresponding cases after the introduction of `FlowOps`:

* `readUDP` -> `handleFrame` -> `sendToConsumer` -> `handleForwardedPacket` -> (via `injectFop`) `Pcap Send`
* `sniff` -> (via `fop` based on router's decision to forward over UDP) `curriedForward Send` -> `sendToConsumer` -> ditto (but in a different goroutine)

rade: I see. I was looking at the final state of this PR, i.e. your 2nd explanation with `FlowOps`, which is considerably more obscure than the first.

dpw:
> I am finding the control flow much harder to follow that before. Part of that is obviously due to familiarity with the previous code base.

It seems like you are looking at the diff of the whole branch. It's certainly necessary for the code to be comprehensible on that basis, and I would like to work out what we can do to address that.  But if you want to transfer your familiarity with the previous code base, then I suspect you'll find it easier if you look at each commit in turn.  In particular, while the commit that introduces `FlowOp` is a pure refactoring and in a sense quite straightforward, it is undeniable that it convolutes things.  Of course, that is the crucial convolution required to implement fast data path in terms of the existing router code base.

rade:
> But if you want to transfer your familiarity with the previous code base, then I suspect you'll find it easier if you look at each commit in turn.

I actually have mostly been doing that. But, as you have pointed out on a number of occasions now when I had questions on specific commits, some of the design decisions in earlier commits are motivated by what comes later.

It would help me a lot if I was able to merge all but the interhost/sleeve and flow changes.

dpw:
> I actually have mostly been doing that. But, as you have pointed out on a number of occasions now when I had questions on specific commits, some of the design decisions in earlier commits are motivated by what comes later.

The commits are related but separate, like chapters in a book.  They are not intended to stand alone, but neither should their separation be discarded.  Sometimes the answer to a question "why is this done like this" will be in terms of what happens in a later commit.  In such cases, you won't be able to fully evaluate that explanation until you reach the later commit.  But the separation is important, particularly when trying to understand the changes in terms of the prior state of the code base.  That is why I divided the commits up, and included extended commit messages on the more interesting ones to provide rationale and context.

> It would help me a lot if I was able to merge all but the interhost/sleeve and flow changes.

Ok, let's do this synchronously then.  I'll let you know when I think everything is ready to merge.

dpw:
Naming idea: `IntraHost` -> `Bridge`, `InterHost` -> `Overlay`.  Pros: short, accurate in the context of the router codebase.  Cons: pretty ambiguous outside of that context (but I think they fit well enough to override that concern).

squaremo:
> Naming idea: IntraHost -> Bridge, InterHost -> Overlay

SGTM

rade:
> Naming idea: IntraHost -> Bridge, InterHost -> Overlay

+1

dpw:
>     what all the test code really wants is a Peer.NewPeerFrom(p *Peer)

Done

>     NewPeer should call NewPeerFromSummary

Done

>     can we come up with a better name than PeerSummary? That name made sense when the summary was used only for encoding, but it looks really odd for Peer to contain PeerSummary.

Not done.  I still find PeerSummary reasonable here.  Its use is now broader than it was before, but that doesn't make it a bad name.

router/router.go
dpw: Done

router/sleeve.go
dpw: Done

router/sleeve.go
dpw: Done

dpw:
Other items of feedback that github has now swallowed have also been addressed.

router/router.go
```diff
@@ -53,21 +55,17 @@ type Router struct {
...
+		router.Bridge = NullBridge{}
+	}
 
-type PacketSourceSink interface {
-	PacketSource
-	PacketSink
-}
+	if router.Overlay == nil {
+		router.Overlay = NullOverlay{}
+	}
```
rade: IMO `Config` should be immutable. Perhaps introduce a `NewConfig()` that sets `Bridge` and `Overlay`?

dpw: Are you sure about that?  I would agree that we shouldn't modify the caller-provided `Config` here, but that's not what we are doing.  We are modifying fields in `Router` that happen to be inherited from `Config`.  So how would a `NewConfig` help here?

rade: Ah, `config` gets copied, right? Nevermind.

rade:
What's the protocol compatibility situation? If we merged this pre 1.1, would it (or could easily be made to) be compatible with 1.0?

dpw:
A general note about this PR: The commits are not currently protocol-compatible with 1.0/master.  My expectation is that it will be possible to make them compatible with a couple of hours of work.  But I'm deferring that work in case it turns out that FDP proper fundamentally breaks compatibility somewhere else (I'm not aware that it does, but it was not implemented with compatibility in mind, and there are some areas that will require careful thought before declaring that it is compatible).

dpw:
> What's the protocol compatibility situation? If we merged this pre 1.1, would it (or could easily be made to) be compatible with 1.0?

Oh, I was just writing about that.  Yes, I expect so.  Do you want to merge for 1.1? I don't know of a good reason not to, but I don't think we gain anything by doing so, so my inclination is not to merge.

inercia:
@dpw Is there any architecture document that I could use as an introduction to this PR?

dpw:
> @dpw Is there any architecture document that I could use as an introduction to this PR?

While some implementation aspects of the router change, the architecture does not change, and as far as I am aware, and existing architecture documents continue to apply.  Feedback about any places where the implementation is unclear or unnecessarily obscure is welcome though!

dpw:
(This morning it occurred to me that the Peer / PeerSumary change might not be gob-compatible.  The gob documentation is quite unclear on such matters.  So I have checked.  It is compatible.)

router/flow.go
```diff
@@ -0,0 +1,50 @@
...
+	PacketKey
+}
+
+type FlowOp interface {
+	// The caller must supply an EthernetDecoder specific to this
+	// thread, which has already been used to decode the frame.
+	// The broadcast parameter is a hint whether the packet is
+	// being broadcast.
+	Send(frame []byte, dec *EthernetDecoder, broadcast bool)
+}
```
awh: Out of curiosity - is there a reason `FlowOp` has type interface rather than func? Do you anticipate adding additional method signatures here?

awh:
I've spent a few hours reading this so far; initial thoughts:

* The chosen abstractions (bridge/overlay + router core currying) seem reasonable and uncontroversial. No superior alternatives spring immediately to mind, but I will spend some time thinking about it
* New naming + commit structure is good. Extended commit comments are helpful
* The commit that introduces overlay & sleeve requires the most attention. I'm going to go back and refamiliarise myself with the existing implementation before tackling that in detail

router/flow.go
dpw: Kind of.  Fast datapath introduces `FlowOps` that still represent a way to send a packet, but gain another interpretation as elements of ODP flow rules passed to the kernel.  That doesn't actually result in more methods being added to `FlowOp`, as not all `FlowOps` are fast datapath `FlowOps`.  But it requires the ability to test if a `FlowOp` implements another interface.  That wouldn't work with a function type.

dpw:
> The commit that introduces overlay & sleeve requires the most attention. I'm going to go back and refamiliarise myself with the existing implementation before tackling that in detail

Great.  I did find a small bug there earlier this week, but I didn't get around to pushing the fix yet.  Shall I hold on to it to give you something to look for?

awh:
>  I did find a small bug ... Shall I hold on to it to give you something to look for?

Is it the only one then? :wink:

router/router.go
```diff
@@ -135,70 +121,39 @@ func onOff(b bool) string {
...
-	dec.DecodeLayers(frameData)
-	decodedLen := len(dec.decoded)
-	if decodedLen == 0 {
-		return
-	}
-	srcMac := dec.Eth.SrcMAC
-	srcPeer, found := router.Macs.Lookup(srcMac)
 	// We need to filter out frames we injected ourselves. For such
 	// frames, the srcMAC will have been recorded as associated with a
 	// different peer.
```
awh: This comment could use an explanation similar to the one in `handleForwardedPacket` e.g. '(because we only ever *inject* packets that have come from a different peer)'

router/router.go
```diff
@@ -230,138 +185,44 @@ func (router *Router) acceptTCP(tcpConn *net.TCPConn) {
...
-		if router.Macs.Enter(srcMac, srcPeer) {
-			log.Println("Discovered remote MAC", srcMac, "at", srcPeer)
-		}
-		if po != nil {
-			router.LogFrame("Injecting", frame, dec)
-			checkWarn(po.WritePacket(frame))
-		}
+	srcMac := net.HardwareAddr(key.SrcMAC[:])
+	dstMac := net.HardwareAddr(key.DstMAC[:])
+	if router.Macs.Enter(srcMac, key.SrcPeer) {
```
awh: Is there a reason not to record the mac/peer mapping for relayed packets too? The previous code has the same behaviour

router/router.go
```diff
@@ -135,70 +121,39 @@ func onOff(b bool) string {
...
+		return nil
 	}
 
-	err := router.Ourself.Forward(dstPeer, frameCopy, dec)
-	if ftbe, ok := err.(FrameTooBigError); ok {
-		err = dec.sendICMPFragNeeded(ftbe.EPMTU, po.WritePacket)
+	dstMac := net.HardwareAddr(key.DstMAC[:])
+	switch dstPeer := router.Macs.Lookup(dstMac); dstPeer {
+	case router.Ourself.Peer:
+		return nil
```
awh: This clause could benefit from a comment along the lines of 'We've sniffed a packet that was destined for a MAC that's directly attached to the bridge - ignore it'

router/router.go
```diff
@@ -135,70 +121,39 @@ func onOff(b bool) string {
...
+	case nil:
+		// If we don't know which peer corresponds to the dest
+		// MAC, broadcast it.
+		return router.Ourself.Broadcast(key)
+	default:
+		router.PacketLogging.LogPacket("Forwarding", key)
```
awh: Is there a reason we log forwarded but not broadcast packets?

router/router.go
dpw: Is there a reason to do so?  You might say that it is potentially useful information, but it's not clear that it is genuinely useful: just because a peer relays a packet, is it likely to want to send to the source peer?  Even though the MacCache is not currently size-limited, putting redundant information in there still has some cost.

So the existing behaviour seems reasonable to me.

router/local_peer.go
```diff
@@ -27,54 +27,54 @@ func NewLocalPeer(name PeerName, nickName string, router *Router) *LocalPeer {
 	return peer
 }
 
-func (peer *LocalPeer) Forward(dstPeer *Peer, frame []byte, dec *EthernetDecoder) error {
-	return peer.Relay(peer.Peer, dstPeer, frame, dec)
+func (peer *LocalPeer) Forward(dstPeer *Peer, key PacketKey) FlowOp {
```
awh: It feels a bit odd to me to have these four methods (Forward/Broadcast/Relay/RelayBroadcast)  in local_peer.go rather than router.go - thoughts?

awh:
It may just be me, but the name `ConsumePackets` on overlay & bridge made me think that they would be synchronous when in fact they start goroutines that continue invoking the supplied callback after the function has returned.

router/local_peer.go
dpw: I might agree with you, but the code was already like that, and I have strenuously resisted the temptation to improve the router code in ways not crucial to fast datapath.

router/sleeve.go
```diff
@@ -0,0 +1,1117 @@
...
+		case frame := <-aggDFChan:
+			err = fwd.aggregateAndSend(frame, aggDFChan,
+				fwd.crypto.EncDF, fwd.senderDF, fwd.maxPayload)
+
+		case special := <-specialChan:
+			err = fwd.handleSpecialFrame(special)
+
+		case _, ok := <-confirmedChan:
+			if !ok {
+				// specialChan is closed to indicate
```
awh: `s/specialChan/confirmedChan/`

router/local_peer.go
awh: Good point - I'm sure rebasing this constantly is enough of a nightmare already!

awh:
I have similar (mild) concern about `Overlay.MakeForwarder` - it causes me to wonder if it's doing something similar to Go's `make` function, and it doesn't sound asynchronous. `StartForwarder`? `GoForward`?

dpw:
> I have similar (mild) concern about Overlay.MakeForwarder - it causes me to wonder if it's doing something similar to Go's make function, and it doesn't sound asynchronous. StartForwarder? GoForward?

I see your point about `ConsumePackets`, but the comment about `MakeForwarder` seems like a misapprehension of the intent.  `MakeForwarder`, uhmm, makes a forwarder.  And essentially, a forwarder is something you can call `Forward` on to relay a packet to the peer concerned.  A forwarder may, as an implementation detail, have goroutines and other active machinery behind it, but it would be wrong to reflect that in the name.  In contrast to `ConsumePackets`, the asynchrony is not key to the interface.  The only asynchrony from the caller's point of view is indirectly through the forwarder's `SetListener` method.

awh:
> `MakeForwarder`, uhmm, makes a forwarder. And essentially, a forwarder is something you can call `Forward` on to relay a packet to the peer concerned. A forwarder may, as an implementation detail, have goroutines and other active machinery behind it, but it would be wrong to reflect that in the name.

If the forwarder were completely inert unless you used it to `Forward` something (which it may accomplish with a goroutine at its discretion), I would agree. That is not the case here - there is a continuous UDP heartbeat process going on in the background. The presence of a `Close()` method on `OverlayForwarder` is also indicative IMO.

The aforementioned notwithstanding, why `MakeForwarder` and not `NewForwarder`?

dpw:
> If the forwarder were completely inert unless you used it to Forward something (which it may accomplish with a goroutine at its discretion), I would agree. That is not the case here - there is a continuous UDP heartbeat process going on in the background. The presence of a Close() method on OverlayForwarder is also indicative IMO.

The UDP heartbeat is an implementation detail of the `sleeveForwarder`.  The point of the `Overlay` interface et al is to hide such details.  And the fast data path code in the "preview release" demonstrates that `OverlayForwarder` can be implemented without any associated goroutines, etc.

> The aforementioned notwithstanding, why MakeForwarder and not NewForwarder?

There seems to be a golang convention that `NewForwarder` is a function that allocates a concrete type called a `Forwarder`.  You might say that convention only applies to package-level functions.  But it seemed better not to allude to it here.  `NewForwarder` might tend to suggest a passive allocation, and, as you point out, implementations might be doing much more than that.  Hence I chose a name that I regard as more neutral.

awh:
> The UDP heartbeat is an implementation detail of the sleeveForwarder. The point of the Overlay interface et al is to hide such details. And the fast data path code in the "preview release" demonstrates that OverlayForwarder can be implemented without any associated goroutines, etc.

I am persuaded :+1:

awh:
> Hence I chose a name that I regard as more neutral.

Except there is the `make` builtin. How about `OpenForwarder` to mirror `OverlayForwarder.Close()`?

dpw:
> Except there is the make builtin

Given the rather specialized purpose of the `make` keyword, and the capitalization difference, and the fact that here the word `Make` appears in a longer identifier, I'm not sure what kind of confusion can arise.

There's a `new` keyword too, and that clearly doesn't preclude the use of the word `New` in identifiers.

There are quite a few uses of `[mM]ake` in identifiers in the go source tree.

> How about OpenForwarder to mirror OverlayForwarder.Close()?

`OpenForwarder` might suggest the creation of some kind of connection or channel associated with the forwarder (as when one opens a TCP socket).  That's not the case for any implementation I have in mind.  How about I rename `Close` to `Stop` or `Terminate`?

router/flow.go
inercia: `FlowOps` could be used for more than just sending packets, right? Maybe another name (eg, `Exec()` instead of `Send()`) would be more appropriate...

router/sleeve.go
```diff
@@ -0,0 +1,1117 @@
...
+
+	fwd.heartbeatTimeout.Reset(HeartbeatTimeout)
+	return nil
+}
+
+func (fwd *sleeveForwarder) setRemoteAddr(addr *net.UDPAddr) {
+	// Although we don't need to lock when reading remoteAddr,
+	// because this thread is the only one that modifies
+	// remoteAddr, we do need to lock when writing it, because
+	// memory models.
```
awh: I'm mystified by this comment. Firstly, it's not clear what 'this thread' is referring to; secondly, 'because memory models', whilst amusing, is insufficiently explicatory :smile:

router/sleeve.go
```diff
@@ -0,0 +1,1117 @@
...
+	fwd.lock.Lock()
+	fwd.remoteAddr = addr
+	fwd.lock.Unlock()
+}
+
+func (fwd *sleeveForwarder) handleHeartbeatAck() error {
+	// The connection is nowregarded as established
```
awh: Missing space.

router/flow.go
```diff
@@ -0,0 +1,50 @@
...
+func NewMultiFlowOp(broadcast bool) *MultiFlowOp {
+	return &MultiFlowOp{broadcast: broadcast}
+}
+
+func (mfop *MultiFlowOp) Add(ops ...FlowOp) {
+	mfop.ops = append(mfop.ops, ops...)
+}
+
+func (mfop *MultiFlowOp) Send(frame []byte, dec *EthernetDecoder,
+	broadcast bool) {
```
awh: I wonder if it's a good idea to use `_ bool` here to signify that we're ignoring the passed value?

awh:
> How about I rename Close to Stop or Terminate?

:+1:

awh:
`sleeve.go` is by far the biggest individual file now as it contains then entire UDP forwarding mechanism - do you have any plans to address this post FDP merge? Is the entire ODP overlay in a single file too? Am wondering about the use of packages or some other organising principle (e.g. `bridge_pcap.go`/`bridge_odp.go`, `overlay_sleeve.go`,`overlay_sleeve_forwarder.go`/`overlay_odp.go`)

dpw:
> sleeve.go is by far the biggest individual file now as it contains then entire UDP forwarding mechanism - do you have any plans to address this post FDP merge?

I don't have any such plans, although that is certainly not to say that it should not be addressed.

I don't consider the length of sleeve.go to be much of a problem at this point.  There is a high level of cohesion among its contents, and coupling to the rest of the code base is reasonably loose, so it forms a nice unit.  As file boundaries have no semantic significance in golang, it clearly could be split up into multiple files, but that would tend to diminish its status as a component with a well-drawn boundary.

(So, for example, I find it much easier to navigate the UDP encapsulation code now that it is all in one file, albeit a rather large one, compared to the previous state where it was spread out over multiple files that often had other concerns.)

Of course, at what point a file becomes unmanageably long is a subjective matter.  I'm not sure if one can take the go source tree as an authority, but that contains a number of files of similar or greater length (even excluding test files and syscall definition files and other such special cases).

> Is the entire ODP overlay in a single file too? 

Yes.  It's not currently as long as sleeve.go, though how long it will end up remains to be seen.

> Am wondering about the use of packages or some other organising principle

I think that if the router were to be decomposed into multiple packages, with sleeve.go in its own package, I would feel more comfortable with splitting it up into multiple files (as the conceptual boundary would be represented by the package boundary, there would be no need to represent it by a file boundary).

But as no other part of the router is in its own package, doing so would represent a noteworthy step.  I'm not sure the need here is great enough to justify crossing that line.  I certainly don't think there's a strong case to do it before merging.  After merging, the decision can be made at leisure.

> (e.g. bridge_pcap.go/bridge_odp.go, overlay_sleeve.go,overlay_sleeve_forwarder.go/overlay_odp.go)

I'm sceptical of the advantages of such a split.  Maybe I just don't like typing long filenames.

awh:
I too appreciate the entirety of the sleeve implementation being in a single place - I find it much easier to understand. The concern is less about the length of the file, and more about being able to see the pieces that comprise sleeve (and its FDP counterpart) without trawling through the file; it would be nice if we could have the best of both worlds. I think we're in agreement that fine grained packages would achieve this, but that it is a fair amount of work; the idea of the overlay_$IMPL_$SUBCOMPONENT.go approach was a lightweight compromise. Related, parts of crypto.go are entirely sleeve specific...

dpw:
> and more about being able to see the pieces that comprise sleeve

I think that is a fine wish when those pieces are intended to be taken in isolation.  But that is not the case here.  The "pieces" of sleeve.go are somewhat tightly coupled.  Moving them into separate files as-is would suggests more of a boundary than is justified.  If they were moved into separate files, it would be appropriate to restructure the code to make them less tightly coupled.  But that creates work, for the author, the reader, and those who maintain the code.  At around a thousand lines, I don't think there is a problem that needs solving here.

> Related, parts of crypto.go are entirely sleeve specific...

Other parts of the router codebase interact with crypto.go, and the interface between crypto.go and sleeve.go is narrow and simple.  So the file boundary there corresponds to a clear and clean design boundary.

router/sleeve.go
```diff
@@ -0,0 +1,1117 @@
...
+	case fwd.specialChan <- specialFrame{nil, msg}:
+	case <-fwd.finishedChan:
+	}
+}
+
+func (fwd *sleeveForwarder) Close() {
+	sleeve := fwd.sleeve
+	sleeve.lock.Lock()
+	if sleeve.forwarders[fwd.remotePeer.Name] == fwd {
+		delete(sleeve.forwarders, fwd.remotePeer.Name)
```
awh: We have separate `addForwarder` and `lookupForwarder` methods on `Sleeve` - why not `removeForwarder` too?

router/sleeve.go
```diff
@@ -0,0 +1,1117 @@
...
+	fwd.lock.RLock()
+	defer fwd.lock.RUnlock()
+	if fwd.listener != nil {
+		fwd.listener.Established()
+	}
+}
+
+func (fwd *sleeveForwarder) sendFragTest() error {
+	log.Debug(fwd.logPrefix(), "sendFragTest")
+	fwd.stackFrag = false
```
awh: I appreciate that the previous implementation did the same thing, but I find this a bit odd - we're setting `stackFrag` to false periodically (with concomitant effects in `Send`) even when stack fragmentation is continuously available and working. The only benefit from doing this that I can think of is avoiding an additional timer to notice a lack of `FragTestAck` - is that your assessment too?

Also shouldn't the modification of `fwd.stackFrag` here and in `handleFragTestAck` be protected by a write lock as `Send` (which obtains a read lock before accessing `stackFrag`) is called from a different goroutine?

dpw: > I appreciate that the previous implementation did the same thing, but I find this a bit odd - we're setting stackFrag to false periodically (with concomitant effects in Send) even when stack fragmentation is continuously available and working. The only benefit from doing this that I can think of is avoiding an additional timer to notice a lack of FragTestAck - is that your assessment too?

As you say, it is doing what the old implementation did.

`stackFrag == false` should only have implications for performance, not behaviour, and only for packets that are larger than the MTU, which tend to be rare in practice.  And other normal circumstances, it will be set back to true promptly.  It doesn't seem like a major concern.

> Also shouldn't the modification of fwd.stackFrag here and in handleFragTestAck be protected by a write lock as Send (which obtains a read lock before accessing stackFrag) is called from a different goroutine?

Although the golang memory model is entirely undocumented AFAICT, and I have not been able to find an authoritative pronouncement on the topic, it seems safe to assume that reads and writes of bools are atomic. They are compiled to single bytes, not bit-packed (the ability to take pointers to them makes this almost inevitable). So modifying it without holding a lock is safe.

router/sleeve.go
```diff
@@ -0,0 +1,1117 @@
...
+
+func (fwd *sleeveForwarder) sendSpecial(enc Encryptor, sender udpSender,
+	data []byte) error {
+	enc.AppendFrame(fwd.sleeve.localPeerBin, fwd.remotePeerBin, data)
+	return fwd.flushEncryptor(enc, sender)
+}
+
+func (fwd *sleeveForwarder) handleSpecialFrame(special specialFrame) error {
+	if special.sender == nil {
+		return fwd.handleControlMsg(special.frame)
```
awh: I'd be tempted to inline `handleControlMsg` so that all the specialFrame dispatch happens in a single location, rather than being staggered (asymmetrically) across two functions...

router/sleeve.go
awh: I found this: https://golang.org/ref/mem

> it seems safe to assume that reads and writes of bools are atomic

From linked doc "Reads and writes of values larger than a single machine word behave as multiple machine-word-sized operations in an unspecified order." implies you are correct.

dpw: > From linked doc "Reads and writes of values larger than a single machine word behave as multiple machine-word-sized operations in an unspecified order." implies you are correct.

It's a shame it's not explicit about bools though.  They can be a special case, because for example, if you have `struct { bool a, bool b }` those two fields could in principle be packed within a single byte (as with C bit fields).  But then assignments like `s.a = true` can end up compiling to the equivalent of `*s |= 1`, and so assignments are not atomic.  But then pointers to bit-packed fields need special handling, so I think it's unlikely that any golang implementation would do that even aside from the atomicity concerns.

awh:
I've just run the smoke tests again for the first time after the most recent rebase, and there are some things that no longer pass:

````
>>> Ran 34 tests, 5 failed
>>> Fail ./250_dns_lookup_2_test.sh
>>> Fail ./160_claim_3_test.sh
>>> Fail ./210_dns_cross_hosts_3_test.sh
>>> Fail ./660_proxy_ipam_2_test.sh
>>> Fail ./120_discovery_3_test.sh
````

The coverage report is interesting - `sleeve.go` is pretty well covered, except for the manual fragmentation path. Interestingly that *is* covered on current master due to @rade's new `105_frag_2_test.sh` so we should get some coverage there post upcoming rebase...

**Comment by rade** [comment]
Rebasing should fix 160.

**Comment by awh** [comment]
@dpw just to put our conversation yesterday on the record - from a design/maintainability/handover POV I am ready to merge this PR; remaining activity is concerned with convincing myself that the behaviour hasn't changed, but I think that can be time bounded by however long it takes you to rebase + address outstanding comments.

**Comment by awh** [comment]
`docs/architecture.txt` needs updating - it speaks of two separate forwarder goroutines

File: router/sleeve.go (hunk 18 above)
Comment (awh) [inline router/sleeve.go]: Hmm actually now I notice that the control messages are coming over the TCP channel and we're smuggling them through `specialChan` by setting the sender to nil. I suspect you will feel otherwise due to the large number of fields already present on `sleeveForwarder`, but I would prefer to use a separate channel for this for the sake of clarity...

**Comment by awh** [comment]
We have a mix of ways we're testing or setting the value of `heartbeatInterval`:

````
759:	if fwd.heartbeatInterval != 0 {
800:		if fwd.heartbeatInterval != time.Duration(0) {
836:	if fwd.heartbeatInterval != SlowHeartbeat {
````
````
767:	fwd.heartbeatInterval = time.Duration(FastHeartbeat)
837:		fwd.heartbeatInterval = SlowHeartbeat
````

File: router/sleeve.go (hunk 18 above)
Comment (dpw) [inline router/sleeve.go]: It's not the fields I'd be worried about.  I haven't studied it first-hand, but my understanding of golang's select implementation is that on each pass through the select, it re-evaluates all the channel expressions into an array and then calls some run-time function to check if any are ready, and sleep if not.  So adding an extra case, even one that rarely fires, has a performance impact.  For that reason I was concerned about the performance implications of that big select in a performance-critical area.  Fortunately, when I measured the end-result, it seems to match mainline weave.  But the concern hasn't completely gone away.

If one wanted to address that performance concern properly, then it would be done by merging the various channels, with an auxiliary goroutine to fan-in the various timer events onto the merged channel.  One way to look at use of the `specialChan` for both special UDP packets and control messages is as that kind of merging, in a case where it happens to be straightforward to do so.

File: router/connection.go [hunk 19]
```diff
@@ -307,77 +202,63 @@ func (conn *LocalConnection) run(actionChan <-chan ConnectionAction, finished ch
...
+	params := ForwarderParams{
+		RemotePeer:         conn.remote,
+		LocalIP:            conn.TCPConn.LocalAddr().(*net.TCPAddr).IP,
+		RemoteAddr:         conn.remoteUDPAddr,
+		ConnUID:            conn.uid,
+		Crypto:             conn.forwarderCrypto(),
+		SendControlMessage: conn.sendOverlayControlMessage,
 	}
-	if err = conn.Router.Ourself.AddConnection(conn); err != nil {
+	if conn.forwarder, err = conn.Router.Overlay.MakeForwarder(params); err != nil {
```
awh: This will fail if there is already a forwarder for `conn.remote`; doesn't that make the tie breaking code in `AddConnection` moot?

dpw: Good point.  Looks like adding the forwarder to the forwarders map needs to be deferred into `confirmed`.

router/sleeve.go
awh: Sigh. It's a shame that we have to spend so much effort (manual locking, this) avoiding use of the one feature that Go is supposed to be good at...

router/sleeve.go
```diff
@@ -0,0 +1,1117 @@
...
+	}
+
+	fwd.fragTestTicker = time.NewTicker(FragTestInterval)
+	if err := fwd.sendFragTest(); err != nil {
+		return err
+	}
+
+	// Send a large frame down the DF channel in order to prompt
+	// PMTU discovery to start.
+	return fwd.sendSpecial(fwd.crypto.EncDF, fwd.senderDF,
```
awh: For the default PMTU to be reduced down from 65535 to something sensible we're relying on receiving an error back from the stack in `processSendError`, however the architecture docs state that the reason for the PMTU discovery mechanism is that such messages may not be reliably received. Are we always guaranteed to get an error here even if ICMP has been filtered out because the packet is bigger than the MTU of the egress interface? Or do we not care about defending against ICMP filtering stupidity on the hosts, only that which may be happening in the intervening network?

router/sleeve.go
awh: Related - was the move to a single goroutine (and hence select) per forwarder motivated from the same concern about overheads?

awh: Actually thinking about this more - doesn't that exacerbate the problem you described? It means you're re-evaluating all channel expressions any time any of them fires, which would not be the case if you had a goroutine per channel. Perhaps there's some other concern there though. Just thinking out loud :smile:

dpw: Yes, using a single goroutine leads to the large select, which was always a concern for me.  OTOH, using a single goroutine might also yield performance benefits to, due to improved locality etc.  My performance measurements suggest it's a wash.

The reason for using a single goroutine was simplicity.  Compared to the original code, there is less locking and inter-goroutine communication involved, because responsibilities that were dispersed between the Connection and the two forwarder goroutines are now confined to a single goroutine.

router/sleeve.go
dpw: Yes, I agree that code is a bit sloppy, and it could be better integrated into the EPMTU search code so that, even if we never see an EMSGSIZE, we would still arrive at the correct EPMTU.

But I think it is unlikely to be a problem in practice for two reasons:

* As you suggest, unless the outgoing netdev has an MTU of 65535, we'll get an EMSGSIZE anyway, and then we initiate the EPMTU search.
* By this point, the kernel's TCP code will have established the PMTU to the remote peer's IP, and that works even in the absence of frag-needed responses.  Though I suppose it's possible that there could be NAT in the way that maps a peer to different IP addresses for the TCP and UDP sides.

I haven't checked, and it's been a while since I wrote that code, but I'd be surprised if it doesn't just recapitulate what the existing code does.

awh:
> As you suggest, unless the outgoing netdev has an MTU of 65535, we'll get an EMSGSIZE anyway, and then we initiate the EPMTU search.

I've had a look at the kernel source and it uses the device MTU in the absence of a discovered PMTU value so this is guaranteed to happen.

> By this point, the kernel's TCP code will have established the PMTU to the remote peer's IP, and that works even in the absence of frag-needed responses

Is that [Packetization Layer Path MTU Discovery](https://tools.ietf.org/html/rfc4821)?

> I haven't checked, and it's been a while since I wrote that code, but I'd be surprised if it doesn't just recapitulate what the existing code does.

I have, and it does :smile:

router/sleeve.go
awh: > As you suggest, unless the outgoing netdev has an MTU of 65535, we'll get an EMSGSIZE anyway, and then we initiate the EPMTU search.

I've had a look at the kernel source and it uses the device MTU in the absence of a discovered PMTU value so this is guaranteed to happen.

> By this point, the kernel's TCP code will have established the PMTU to the remote peer's IP, and that works even in the absence of frag-needed responses

Is that [Packetization Layer Path MTU Discovery](https://tools.ietf.org/html/rfc4821)?

> I haven't checked, and it's been a while since I wrote that code, but I'd be surprised if it doesn't just recapitulate what the existing code does.

I have, and it does :smile:

awh: I think the unfamiliar would find it useful if the comment here had a short explanation of how the large frame manifests in `processSendError` at a later time - it wasn't immediately obvious to me.

dpw: > > By this point, the kernel's TCP code will have established the PMTU to the remote peer's IP, and that works even in the absence of frag-needed responses
> 
> Is that Packetization Layer Path MTU Discovery?

Yes.  (Although the kernel code doesn't refer to that RFC, and I wouldn't be surprised if the technique came first and the RFC codified it.)

As that RFC notes, TCP doesn't allow you to pad packets.  So TCP-based PMTU discovery can only proceed when the application supplies enough data for the TCP layer to make a segment of the candidate size.  I don't think our connection initiation protocol is guaranteed to do big sends, so it's less of a sure thing than I thought that TCP will force discovery of the PMTU.

router/flow.go
dpw: The problem with bools is that if you eliminate the name, the meaning becomes obscure.  In this case it might be ok, because the corresponding interface is nearby in the same file. But I'm not keen on removing the names of unused parameters as a general rule.

awh:
Shouldn't `MaxUDPPacketSize` in `router/consts.go` be 65535 and not 65536? (it's wrong on master too...)

awh:
Name change suggestion for clarity:

`sleeveForwarder.effectivePMTU` -> `sleeveForwarder.overlayPMTU`
`msgTooBigError.PMTU` -> `msgTooBigError.underlayPMTU`

Not sure what to do with `sleeveForwarder.epmtu*` though :frowning:

awh:
I made this diagram to understand the PMTU discovery - I think it would make a helpful comment in `sleeve.go`:
````
+----------+----+-----+--------+--------+----------+-------------------------+                              
| Ethernet | IP | UDP | Sleeve | Sleeve | Overlay  | Overlay Layer 3 Payload |                              
|          |    |     | Packet | Frame  | Ethernet |                         |                              
|          |    |     | Header | Header |          |                         |                              
+----------+----+-----+--------+--------+----------+-------------------------+                              
                                                                                                            
+----------+ EthernetOverhead                                                                               
                                                                                                            
           +----------+ UDPOverhead                                                                         
                                                                                                            
                      +--------+ Encryptor.PacketOverhead                                                   
                                                                                                            
                               +--------+ Encryptor.FrameOverhead                                           
                                                                                                            
           +---------------------------------------+ sleeveForwarder.overheadDF                             
                                                                                                            
           +-----------------------------------------------------------------+ msgTooBigError.PMTU          
                                                                                                            
                      +------------------------------------------------------+ sleeveForwarder.maxPayload   
                                                                                                            
                                                   +-------------------------+ sleeveForwarder.effectivePMTU

````
(import into [ASCIIFlow](http://asciiflow.com) for serious edits)

dpw:
> I made this diagram to understand the PMTU discovery - I think it would make a helpful comment in sleeve.go:

I wonder if it could be made to fit in 80 columns by transposing it so that the packet structure runs from top to bottom.

More substantially, I think the EthernetOverhead bar is in the wrong place.  It should apply to the "Overlay Ethernet" box.  I don't think the outer ethernet header is ever relevant, and it can be dropped.  The outermost thing we ever care about is the PMTU reported by the kernel, and that's the packet size at the IP layer.

awh:
````
+----+-----+--------+--------+----------+-------------------------+ 
| IP | UDP | Sleeve | Sleeve | Overlay  | Overlay Layer 3 Payload | 
|    |     | Packet | Frame  | Ethernet |                         | 
|    |     | Header | Header |          |                         | 
+----+-----+--------+--------+----------+-------------------------+ 
                                                                    
+----------+ UDPOverhead                                            
                                                                    
           +--------+ Encryptor.PacketOverhead                      
                                                                    
                    +--------+ Encryptor.FrameOverhead              
                                                                    
                             +----------+ EthernetOverhead          
                                                                    
+---------------------------------------+ sleeveForwarder.overheadDF
                                                                    
+------------------------ msgTooBigError.PMTU --------------------+ 
                                                                    
           +------------- sleeveForwarder.maxPayload -------------+ 
                                                                    
          sleeveForwarder.effectivePMTU +-------------------------+ 

````

awh:
It looks like `log.Debug` behaves differently to `log.Println` in that it doesn't interpolate arguments with spaces:

````
 956         log.Println(fwd.logPrefix(),
 957             "Effective PMTU verified at", epmtu)
````

produces

````
<elided> Effective PMTU verified at 1438
````

whereas

````
 903     log.Debug(fwd.logPrefix(),
 904         "sendPMTUTest: epmtu candidate", fwd.epmtuCandidate)
````

produces

````
<elided> sendPMTUTest: epmtu candidate1438
````

bboreham:
This is following Go standard behaviour: 
 *  `fmt.Sprint` : "Spaces are added between operands when neither is a string." 
 *  `fmt.Sprintln` : "Spaces are always added between operands and a newline is appended."

dpw:
> Shouldn't MaxUDPPacketSize in router/consts.go be 65535 and not 65536? (it's wrong on master too...)

Indeed it should.  It's possible to trigger a spurious EMSGSIZE log entry by doing `ping -M dont -s $[65536-28-6-14-14-20-8]` across the weave network, when really it should be dropped with the same warning as for larger packets.

dpw:
> trigger a spurious EMSGSIZE log entry by doing ping -M dont -s $[65536-28-6-14-14-20-8] across the weave network

Nope, I just looked at the code/logs in more detail:  It actually triggers the connection to be dropped and re-established.

router/router.go
dpw: Actually, that comment and the check it concerns are redundant (pcap sets a BPF filter to get only inbound packets, the fdp bridge does not hairpin packets).  I think it should be removed, and add a comment to bridge.go to make it explicit.

rade: The check here is a failsafe.

dpw:
> msgTooBigError.PMTU -> msgTooBigError.underlayPMTU

Ok

> sleeveForwarder.effectivePMTU -> sleeveForwarder.overlayPMTU
> sleeveForwarder.emptu* -> sleeveForwarder.opmtu*

Considering this now, one thing I notice is that the term "PMTU" in "effectivePMTU"   is only tenuously linked to PMTUs (the notion of a PMTU for the L2 overlay network doesn't make sense).  So I think we should just refer to MTU for these, not PMTU.

While I agree that the established naming is not ideal, I'm not sure I agree that we need to qualify it with "overlay".  It's not like we have a lot of mixing of overlay and underlay MTUs.  Apart from the few uses of the underlay PMTU in msgTooBigError (to be addressed as proposed), all uses refer to the overlay MTU.  And I don't think omtuCandidate is greatly less obscure than mtuCandidate.  So I think we can simply drop the "effective", and just refer to the MTU, with a comment for the fields to clarify that they refer to the overlay MTU.

router/sleeve.go
dpw: > I'd be tempted to inline handleControlMsg so that all the specialFrame dispatch happens in a single location, rather than being staggered (asymmetrically) across two functions...

I'll hoist the call to `handleControlMsg` up into the main forwarder loop to make the distinction clearly and for improved symmetry.

awh:
> So I think we can simply drop the "effective", and just refer to the MTU, with a comment for the fields to clarify that they refer to the overlay MTU.

Concur.

dpw:
Looking at @awh's diagram, it suggests that we don't need all three of `maxPayload`, `overheadDF` and `mtu` (as `overheadDF + mtu = maxPayload + UDPOverhead`, with `UDPOVerhead` constant).  But I don't propose to clean that up now.

dpw:
> This is following Go standard library behaviour:
> 
>     fmt.Sprint : "Spaces are added between operands when neither is a string."
>     fmt.Sprintln : "Spaces are always added between operands and a newline is appended."

but we're dealing with logrus, which is based around log entries, so the newline part doesn't apply.  So, there is no difference between `log.Println` and `log.Print`, except for the space insertion behaviour.  Oh, and they are exactly the same as `Infoln` and `Info`.  Obviously.

(And while I'm hating on logrus, why do the methods check the log level in `logger.go` when deciding whether to log an entry, and then immediately check it again in the corresponding method in `entry.go`?)

router/flow.go
dpw: > FlowOps could be used for more than just sending packets, right? Maybe another name (eg, Exec() instead of Send()) would be more appropriate...

Ok.  But `Exec()` seems grammatically wrong to me (saying `op.Exec(packet)` sounds like op executes the packet).  So I've changed it to `Process` instead.

dpw:
Ok, I've rebased and I think I've addressed all the comments that have not been discussed with some other conclusion.  I've put it the results in a separate branch to avoid orphaning comments here - https://github.com/dpw/weave/tree/fdp-prep-fixed .  All the changes in response to the discussion here are broken out to make it easy to review them.  If this meets with approval, I'll squash all the commits marked "Fix" to produce something ready for merging.

Tests currently don't pass on circle CI due to the "docker-py" breakage that affects everything.  I guess I'll try again tomorrow.

awh:
> Ok, I've rebased and I think I've addressed all the comments

Thankyou - will take a look.

dpw:
> Tests currently don't pass on circle CI due to the "docker-py" breakage that affects everything. I guess I'll try again tomorrow.

Tests passing with Bryan's docker-py fix: https://circleci.com/gh/weaveworks/weave/2642

awh:
That seems to cover everything - a few minor comments:
````
+	// we can receive a heartbeat before confirmed() has set up
+	// heartbeatTimeout
+	if fwd.heartbeatTimeout != nil {
+		fwd.heartbeatTimeout.Reset(HeartbeatTimeout)
+	}
+
````
This would be the bug I didn't find then :smile:
> I'll hoist the call to handleControlMsg up into the main forwarder loop to make the distinction clearly and for improved symmetry.

Great compromise - much improved.

````
+	switch newSrcMac, conflictPeer := router.Macs.Add(srcMac, router.Ourself.Peer); {
+	case newSrcMac:
+  		log.Println("Discovered local MAC", srcMac)
+
+	case conflictPeer != nil:
+		// The MAC cache has an entry for the source MAC
+		// associated with another peer.  This probably means
+		// we are seeing a frame we injected ourself.  That
+		// shouldn't happen, but discard it just in case.
+		log.Errorln("Captured frame from MAC (", srcMac, ") associated with another peer", conflictPeer)
+		return nil
 	}
````
This use of switch is pretty obscure - I had to scour the language spec for a few minutes to figure out what it does...

dpw:
> This use of switch is pretty obscure - I had to scour the language spec for a few minutes to figure out what it does...

Yes. Fun, isn't it?

dpw:
Rebased and fixed squashed.

dpw:
I am ashamed that I didn't manage to annoy `bin/lint` at all in this PR.  I will try harder in future.

PR #993, opened 2015-06-23
sleep for 0.1s instead of 1s in the wait_for_status loop

### LINKED ISSUE #992: shorten launch time by reducing sleep
I have [discovered](https://github.com/weaveworks/weave/issues/942#issuecomment-112973008) that sleeping for 1 second in `wait_for_status` retries is slowing down `weave launch` a fair bit. Ideally we'd sleep for a fractional time, but that is non-POSIX so we need to find a way around that.

rade:
This reduces `weave launch|launch-dns` execution time.

Closes #992.

On my machine, `time weave launch` takes ~1.95s before this change, and ~1.15s after.

### DIFF
--- a/weave
+++ b/weave
@@ -183,6 +183,21 @@ command_exists() {
     command -v $1 >/dev/null 2>&1
 }
 
+fractional_sleep() {
+    case $1 in
+        *.*)
+            if [ -z "$NO_FRACTIONAL_SLEEP" ] ; then
+                sleep $1 >/dev/null 2>&1 && return 0
+                NO_FRACTIONAL_SLEEP=1
+            fi
+            sleep $((${1%.*} + 1))
+            ;;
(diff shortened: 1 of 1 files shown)


dpw:
I'm bothered by the redundant `sleep` to check for the support for fractional sleeps.  How about:
```
fractional_sleep() {
    case $1 in
        *.*)
            if [ -z "$NO_FRACTIONAL_SLEEP" ] ; then
                sleep $1 >/dev/null 2>&1 && return 0
                NO_FRACTIONAL_SLEEP=1
            fi
            sleep $((${1%.*} + 1))
            ;;
        *)
            sleep $1
            ;;
    esac
}
```

rade:
> I'm bothered by the redundant `sleep`

Thanks for that code. Incorporated.

PR #994, opened 2015-06-23
separate the image build and export phases

rade:
That way the export can be just a single tar, which is more convenient.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,9 +39,10 @@ prog/netcheck/netcheck
 tools/bin
 tools/build
 releases
+.weaver.uptodate
+.weavedns.uptodate
+.weaveexec.uptodate
 weave.tar
-weavedns.tar
-weaveexec.tar
 prog/weaveexec/weave
 prog/weaveexec/weaveproxy
 prog/weaveexec/weavewait
--- a/Makefile
+++ b/Makefile
@@ -18,23 +18,25 @@ NETCHECK_EXE=prog/netcheck/netcheck
 
 EXES=$(WEAVER_EXE) $(WEAVEDNS_EXE) $(SIGPROXY_EXE) $(WEAVEPROXY_EXE) $(WEAVEWAIT_EXE) $(NETCHECK_EXE)
 
+WEAVER_UPTODATE=.weaver.uptodate
+WEAVEDNS_UPTODATE=.weavedns.uptodate
+WEAVEEXEC_UPTODATE=.weaveexec.uptodate
+
+IMAGES_UPTODATE=$(WEAVER_UPTODATE) $(WEAVEDNS_UPTODATE) $(WEAVEEXEC_UPTODATE)
+
 WEAVER_IMAGE=$(DOCKERHUB_USER)/weave
 WEAVEDNS_IMAGE=$(DOCKERHUB_USER)/weavedns
 WEAVEEXEC_IMAGE=$(DOCKERHUB_USER)/weaveexec
 
 IMAGES=$(WEAVER_IMAGE) $(WEAVEDNS_IMAGE) $(WEAVEEXEC_IMAGE)
 
-WEAVER_EXPORT=weave.tar
-WEAVEDNS_EXPORT=weavedns.tar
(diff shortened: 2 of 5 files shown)


rade:
@tomwilkie so this is [breaking CircleCI](https://circleci.com/gh/weaveworks/weave/1237) because it looks like the docker version on the build VM is too old - multi-arg `docker save` got [introduced in 1.3](https://github.com/docker/docker/commit/e64131d1bf7c79358d5ee7a1a98626730db3593b). Is this easily fixable?

tomwilkie:
We could potentially build in a container with a newer version of docker,
but it might not be fixable.

rade:
where is that ancient version coming from? CircleCI? GCE?

rade:
isn't this using the weaveworks/weave-build image, and doesn't that contain a version of docker?

rade:
> isn't this using the weaveworks/weave-build image, and doesn't that contain a version of docker?

I reckon that's the problem. weave-build's Dockerfile [installs docker.io](https://github.com/weaveworks/weave/blob/master/build/Dockerfile#L3), not lxc-docker.

Am I on the right track?

tomwilkie:
Old version comes from circle.

We do build in the container, and we bind mount the docker socket in. But
we might be able to self host a newer docker.

rade:
Hmm. [Allegedly](https://github.com/circleci/docker/issues/1) CircleCI runs Docker 1.4.1. Which [does support multi-arg save](http://beta-docs.docker.io/v1.4/reference/commandline/cli/#save). Something isn't right here.

rade:
So I reckon the issue is what I said earlier. The multi-arg save gets rejected because the docker *client* inside the build container is too old a version, since it gets installed from the docker.io instead of lxc-docker package.

rade:
Note that we wouldn't be able to just switch to lxc-docker; we'd have to pin the version to 1.4.1 too, since otherwise the client won't be able to talk to the CircleCI-provided docker daemon.

rade:
Success!

I changed the weave-build image to include docker 1.3.1, which is what we use elsewhere, notably in the weaveexec image. It's a bit gross since it requires two `apt-get -y update` - the first to install `apt-transport-https` in order for apt to be able to access the docker.io repo. If anybody has a better idea, let me know.

@tomwilkie I had to `Rebuild without the cache` in order to force a rebuild of weave-build. Will this need to be done on a few other builds until this has run on all the build VMs we have lying around? If so, is there a better way?

rade:
> is there a better way

@tomwilkie says to "just press the button until it works".

dpw:
I just noticed a surprising consequence of this change, although it seems benign:

I have a VM where I had previously done `docker load` for each of the three separate tar files.  When I `docker load` the new unified tar file, I get:

```
The image weaveworks/weave:latest already exists, renaming the old one with ID 97543ea4be18 to empty string
The image weaveworks/weavedns:latest already exists, renaming the old one with ID e0da4ecffe48 to empty string
The image weaveworks/weaveexec:latest already exists, renaming the old one with ID a53e12301065 to empty string
```

These appear to be just warnings, and don't appear for subsequent loads of the single tar file.  But I don't know what is special about this case to trigger them.

I can reproduce this by checking out 1.0 and building and loading separate tar files again, and then loading the single tar file again.

I expect these warnings are confined to `docker load`, so they are not serious.  But it's still a bit of a mystery to me.

bboreham:
Here is the message:
https://github.com/docker/docker/blob/04c6f09fbdf60c7765cc4cb78883faaa9d971fa5/graph/tags.go#L274

called from:
https://github.com/docker/docker/blob/04c6f09fbdf60c7765cc4cb78883faaa9d971fa5/graph/load.go#L77

I can't see what special circumstances would cause it to come just once.

rade:
It's a new feature on 1.7: https://github.com/docker/docker/commit/7d0053d554c453f0ec426bde2899dd55a265f09a

dpw:
Indeed it is:  I get the same warnings when loading the single tar file after a build that changed things.

PR #997, opened 2015-06-24
Eliminate start methods

dpw:
Most start methods in the router are trivially redundant.  Apart from the redundancy, start methods raise the question of the state of an object prior to the start method being called (i.e. which fields are valid, which methods can be called, etc.).  So they should only be used when there is a reason.

(Extracted from #991)

### DIFF
--- a/router/gossip_test.go
+++ b/router/gossip_test.go
@@ -18,15 +18,10 @@ type mockChannelConnection struct {
 	dest *Router
 }
 
-// Construct a "passive" Router, i.e. without any goroutines, except
-// for Routes and GossipSenders.
 func NewTestRouter(name string) *Router {
 	peerName, _ := PeerNameFromString(name)
-	router := NewRouter(Config{}, peerName, "")
-	// need to create a dummy channel otherwise tests hang on nil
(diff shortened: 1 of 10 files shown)


router/router.go
```diff
@@ -84,6 +84,9 @@ func NewRouter(config Config, name PeerName, nickName string) *Router {
 	return router
 }
 
+// Start listening for packets from containers, TCP connections and
```
rade: Shouldn't really mention 'containers' here; the router is largely unaware of them.

router/gossip_test.go
```diff
@@ -22,11 +22,8 @@ type mockChannelConnection struct {
 // for Routes and GossipSenders.
 func NewTestRouter(name string) *Router {
```
rade: The comment above this needs updating since we now certainly do have a bunch of goroutines.

rade:
rest looks fine.

router/gossip_test.go
dpw: I'm just going to delete the comment - `NewTestRouter` seems to cover it.

router/router.go
dpw: Done

PR #999, opened 2015-06-24
Let Scope use these integration test helpers

paulbellamy:
Since scope has to check out the latest weave release anyway, we may as well use the test helpers from there, and only manage them in one place.

This must be merged, and into a release before: https://github.com/weaveworks/scope/pull/273

### DIFF
--- a/test/config.sh
+++ b/test/config.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
 # Protect against being sourced multiple times to prevent
 # overwriting assert.sh global state
 if ! [ -z "$SOURCED_CONFIG_SH" ]; then
@@ -25,9 +27,11 @@ fi
 HOST1=$(echo $HOSTS | cut -f 1 -d ' ')
 HOST2=$(echo $HOSTS | cut -f 2 -d ' ')
 
-. ./assert.sh
+. "$DIR/assert.sh"
 
-SSH=${SSH:-ssh -l vagrant -i ./insecure_private_key -o UserKnownHostsFile=./.ssh_known_hosts -o CheckHostIP=no -o StrictHostKeyChecking=no}
+
+SSH_DIR=${SSH_DIR:-$DIR}
+SSH=${SSH:-ssh -l vagrant -i "$SSH_DIR/insecure_private_key" -o "UserKnownHostsFile=$SSH_DIR/.ssh_known_hosts" -o CheckHostIP=no -o StrictHostKeyChecking=no}
 
 SMALL_IMAGE="gliderlabs/alpine"
 DNS_IMAGE="aanand/docker-dnsutils"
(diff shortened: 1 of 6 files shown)


test/config.sh
```diff
@@ -182,5 +198,5 @@ end_suite() {
     whitely assert_end
 }
 
-WEAVE=../weave
-DOCKER_NS=./bin/docker-ns
+WEAVE=$DIR/../weave
+DOCKER_NS=$DIR/../bin/docker-ns
```
rade: these can go now, can't they?

test/config.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
rade: This is horrific. As I understand it the complication arises from the fact that we are 'sourcing' the script. @dpw is there a better way.

dpw: No.

The test scripts all live in the same directory as `config.sh`, so why not just use `$0`?

test/config.sh
paulbellamy: Almost. We still use them in a few places.

rade: `weave_on` and `660_proxy_env_test.sh`. Could just use `weave` in both. After all, that's what `140_weave_local_test.sh` does.

circle.yml
```diff
@@ -37,13 +40,24 @@ test:
   override:
     - docker run -v /var/run/docker.sock:/run/docker.sock -v /home/ubuntu:/home/go weaveworks/weave-build:
         parallel: true
-    - cd $SRCDIR/test; ./gce.sh setup:
-        parallel: true
-    - cd $SRCDIR/test; . ./gce.sh hosts; ./setup.sh:
-        parallel: true
-    - cd $SRCDIR/test; . ./gce.sh hosts; ./run_all.sh:
```
tomwilkie: One nice-ish side effect of having this as three commands is that you could see in the tests how long each one took.

tomwilkie:
Have you tested this on the weaveworks/weave?

circle.yml
```diff
@@ -37,13 +40,24 @@ test:
...
+    - ? >
+        if [[ -e "/tmp/gce_private_key.json" ]]; then
+          cd $SRCDIR/integration;
```
tomwilkie: Is this the right path?

circle.yml
```diff
@@ -29,7 +29,10 @@ dependencies:
     - sudo apt-get install bc
     - pip install requests
     - curl https://sdk.cloud.google.com | bash
-    - bin/setup-circleci-secrets "$SECRET_PASSWORD"
+    - if [[ -n "$SECRET_PASSWORD" ]]; then
```
tomwilkie: Not a big fan of these tests everywhere - prs without the credentials will be filtered out by the ignore stanza at the top of the file.

test/config.sh
paulbellamy: 140 is using the `run_on` for the weave installed onto `$HOST1`, whereas `weave_on` and 660 are using the local weave script (hence `../weave`).

paulbellamy:
Reverted the commit with the modifications to `circle.yml`, as they're tangential to the main point of this.
Tested on weaveworks/weave at: https://circleci.com/gh/weaveworks/weave/1329

tomwilkie:
LGTM

PR #1000, opened 2015-06-24
Include stylesheet in repo, update footer

davkal:
need because website's custom.css is fingerprinted

### DIFF
--- a/site/_includes/footer.html
+++ b/site/_includes/footer.html
@@ -2,38 +2,49 @@
 <div class="section footer">
   <div class="container">
     <div class="row clearfix">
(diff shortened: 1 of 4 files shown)


bboreham:
1000!

rade:
shouldn't this be against `latest_release_doc_updates` rather than `master`?

Also, I have no idea what this does.

davkal:
@rade the docs page used to reference stylesheets from the main website, which breaks formatting sometimes when they are not in sync.

I'll redo this against `latest_release_doc_updates`.

davkal:
Continued in #1002

PR #1002, opened 2015-06-24
Include stylesheet in repo, update footer

davkal:
`style.css` needs to be included in this repo to untangle it from style changes made in the website repo.

Also needed, because website's custom.css is fingerprinted.

Using the opportunity to update header and footer include files.


fintanr:
LGTM, confirmed at http://docs.weave.works/weave/include-css/00744fb9b496c1aa17fe50dde2df54f601c1af03/index.html

PR #1005, opened 2015-06-24
Try more peers for space if we get errors

### LINKED ISSUE #896: Long delay if ipam asks an absent host for space; can break launch-dns
When one allocator needs more space, it picks another peer at random to ask for space.  If it picks one that is no longer in the network then the message will be dropped; it will be retried when something else happens, but that may be 30 seconds later, on a timer-driven gossip update.

If the program that needs an IP address has a time-out shorter than 30 seconds, as weaveDNS does, then this will inevitably lead to failure.

A basic mitigation would be to pick a different peer if the chosen target is not even in the peer topology, but fixing this may require some crossing of layer abstractions.

bboreham:
Fixes #896

### DIFF
--- a/ipam/allocate.go
+++ b/ipam/allocate.go
@@ -42,9 +42,12 @@ func (g *allocate) Try(alloc *Allocator) bool {
 	}
 
 	// out of space
-	if donor, err := alloc.ring.ChoosePeerToAskForSpace(g.r.Start, g.r.End); err == nil {
+	donors := alloc.ring.ChoosePeersToAskForSpace(g.r.Start, g.r.End)
+	for _, donor := range donors {
 		alloc.debugln("Decided to ask peer", donor, "for space in range", g.r)
-		alloc.sendSpaceRequest(donor, g.r)
+		if err := alloc.sendSpaceRequest(donor, g.r); err == nil {
+			break
+		}
 	}
 
 	return false
--- a/router/gossip_channel.go
+++ b/router/gossip_channel.go
@@ -3,6 +3,7 @@ package router
 import (
 	"bytes"
 	"encoding/gob"
+	"fmt"
 	"log"
 	"sync"
 )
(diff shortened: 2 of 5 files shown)


router/gossip_channel.go
```diff
@@ -160,15 +164,18 @@ func (c *GossipChannel) GossipBroadcast(update GossipData) error {
...
 	} else if conn, found := c.ourself.ConnectionTo(relayPeerName); !found {
-		c.log("unable to find connection to relay peer", relayPeerName)
+		err = fmt.Errorf("unable to find connection to relay peer %s", relayPeerName)
 	} else {
 		conn.(ProtocolSender).SendProtocolMsg(ProtocolMsg{ProtocolGossipUnicast, buf})
 	}
-	return nil
+	if err != nil {
+		c.log(err)
+	}
```
rade: IMO it is generally not a good idea to *both* log and return errors. So I suggest we don't log here and only log in `deliverUnicast` (and not `GossipUnicast`).

ipam/allocate.go
```diff
@@ -42,9 +43,18 @@ func (g *allocate) Try(alloc *Allocator) bool {
...
+		donor, _ := alloc.ring.ChoosePeerToAskForSpace(g.r.Start, g.r.End)
+		if _, triedAlready := tried[donor]; triedAlready {
+			continue
+		}
 		alloc.debugln("Decided to ask peer", donor, "for space in range", g.r)
-		alloc.sendSpaceRequest(donor, g.r)
+		if err := alloc.sendSpaceRequest(donor, g.r); err == nil {
+			break
+		}
+		tried[donor] = struct{}{}
```
rade: This has horrific complexity when there are lots of peers, most of which are dead.

I suggest we change the API s.t. `ChoosePeer*s*ToAskForSpace` returns a list in priority order.

rade:
also, please branch this off 1.0.

bboreham:
The obvious way to return a prioritised list - re-run the randomised pick N(peers) times - has O(N^2) complexity every time, even when nothing is dead.
I can think of improvements like sorting first, which probably reduce that to NlogN, but perhaps the added code isn't worth it.
Either would be a big improvement on current worst-case which is somewhere around N(size of IP space)*N(peers)

rade:
Do we need the randomness? Could we just have a list sorted by free space?

bboreham:
If your gossip messages are getting dropped en-route to the target but you don't get any error back (e.g. because the fault is in an intermediate peer), then you want some mechanism to try a different peer.

rade:
I see. Well googling for "weighted random shuffle" throws up some interesting results. Including [this little haskell beauty](https://hackage.haskell.org/package/random-extras-0.19/docs/Data-Random-Shuffle-Weighted.html)

PR #1009, opened 2015-06-25
Weave launch all the things

### LINKED ISSUE #962: make `weave launch` launch everything
Running the weave router, dns and proxy is now the common case, rather than the exception. So we should make that more convient...

1. `weave launch` should launch all three
2. `weave stop` should stop all three
3. `weave launch-router` should launch just the router (i.e. it's the current `weave launch`)
4. `weave stop-router` should stop just the router (i.e. it's the current `weave stop`)

Rather than getting all clever about arg parsing, `weave launch` should simply pass all args to (just) the router. In the unusual cases where a user wants to pass arguments to the dns or proxy, they can launch all three separately instead.

`weave launch` should fail with a suitable "already running" error (as now) if *any* of the three are already running, and *not* start any of the others. And if starting any of them fails, it should exit with a failure.

`weave stop` should stop *any* containers running, report a "not running" error (as now), and *succeed*.

awh:
Fixes #962.

### DIFF
--- a/site/weavedns.md
+++ b/site/weavedns.md
@@ -28,13 +28,23 @@ hosts.
 
 WeaveDNS is deployed as a set of containers that communicate with each
 other over the weave network. One such container needs to be started
-on every weave host, by invoking the weave script command
-`launch-dns`:
+on every weave host, either simultaneously with the router and proxy
+via `launch`:
 
 ```bash
-host1$ weave launch && weave launch-dns && weave launch-proxy
+host1$ weave launch
 host1$ eval $(weave proxy-env)
 ```
+or independently via `launch-dns`:
+
+```bash
+host1$ weave launch-router && weave launch-dns && weave launch-proxy
+host1$ eval $(weave proxy-env)
+```
+
+The first form is more convenient, however you can only pass weaveDNS
+related configuration arguments to `launch-dns` so if you need to
+modify the default behaviour you will have to use the latter.
 
 Application containers will use weaveDNS automatically if it is
 running at the point when they are started. They will use it for name
@@ -67,7 +77,11 @@ however specify an address in CIDR format manually. In this case you
 are responsible for ensuring that the IP addresses specified are
 uniquely allocated and not in use by any other container.
 
-Finally, WeaveDNS containers can be stopped with `stop-dns`.
+Finally, weaveDNS can be stopped independently with
+
+    host1$ weave stop-dns
+
+or in conjunction with the router and proxy via `stop`.
 
 ## <a name="how-it-works"></a>How it works
 
@@ -112,7 +126,7 @@ Returning to our earlier example, let us start an additional `pingme`
 container, this time on the 2nd host, and then run some ping tests...
 
 ```bash
-host2$ weave launch && weave launch-dns && weave launch-proxy
+host2$ weave launch
 host2$ eval $(weave proxy-env)
 host2$ docker run -dti --name=pingme ubuntu
 
--- a/test/150_connect_forget_test.sh
+++ b/test/150_connect_forget_test.sh
@@ -15,8 +15,8 @@ assert_peers() {
 
 start_suite "Connecting and forgetting routers after launch"
 
-weave_on $HOST1 launch
-weave_on $HOST2 launch
+weave_on $HOST1 launch-router
+weave_on $HOST2 launch-router
 
 start_container $HOST1 $C1/24 --name=c1
 start_container $HOST2 $C2/24 --name=c2
@@ -33,8 +33,8 @@ assert_raises "exec_on $HOST1 c1 $PING $C2"
 
 # Forget everyone and disonnect
 assert_raises "weave_on $HOST2 forget $HOST1 $HOST2"
-assert_raises "weave_on $HOST1 stop"
-assert_raises "weave_on $HOST1 launch"
+assert_raises "weave_on $HOST1 stop-router"
+assert_raises "weave_on $HOST1 launch-router"
 assert_peers $HOST2 ""
 assert_raises "exec_on $HOST1 c1 sh -c '! $PING $C2'"
 
--- a/weave
+++ b/weave
@@ -26,22 +26,22 @@ CIDR_REGEXP="$IP_REGEXP/[0-9]{1,2}"
 usage() {
     echo "Usage:"
     echo "weave setup"
-    echo "weave launch       [-password <password>] [-nickname <nickname>] [-nodiscovery]"
-    echo "                       [-iprange <cidr> [-ipsubnet <cidr>]] <peer> ..."
-    echo "weave launch-dns   [<addr>]"
-    echo "weave launch-proxy [-H <endpoint>] [--with-dns | --without-dns]"
-    echo "                       [--no-default-ipam]"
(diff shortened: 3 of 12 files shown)


awh:
> So do we actually have any test that uses weave launch?

````
grep -r 'weave.*launch[^-_]'
100_cross_hosts_test.sh:weave_on $HOST1 launch -iprange $UNIVERSE -ipsubnet $SUBNET_1
100_cross_hosts_test.sh:weave_on $HOST2 launch -iprange $UNIVERSE -ipsubnet $SUBNET_1 $HOST1
130_expose_test.sh:weave_on $HOST1 launch -iprange $UNIVERSE
110_encryption_test.sh:weave_on $HOST1 launch -password notverysecure
110_encryption_test.sh:weave_on $HOST2 launch -password notverysecure $HOST1
270_use_name_as_hostname_test.sh:weave_on $HOST1 launch -iprange 10.2.0.0/24
140_weave_local_test.sh:run_on $HOST1 sudo ./weave --local launch -iprange 10.2.5.0/24
150_connect_forget_test.sh:assert_raises "weave_on $HOST1 launch"

````
I only modified the ones that needed to launch things individually for example because they were passing parameters to proxy/DNS.

File: weave [hunk 1]
```diff
@@ -1261,6 +1288,13 @@ case "$COMMAND" in
         ;;
     stop)
         [ $# -eq 0 ] || usage
+        stop $PROXY_CONTAINER_NAME "Proxy"
+        stop $DNS_CONTAINER_NAME "WeaveDNS"
+        stop $CONTAINER_NAME "Weave"
+        conntrack -D -p udp --dport $PORT >/dev/null 2>&1 || true
```
awh: I did consider pulling the body of `stop-router` out to a function to avoid duplicating this `conntrack` command, but it seemed unbalanced unless I did the same for `stop-proxy` and `stop-dns` too and that felt like overkill...

rade: not worth it

awh:
Addressed vertical alignment in rebase.

README.md
```diff
@@ -104,7 +104,8 @@ The router, DNS and proxy need to be started once per host. The
 relevant container images are pulled down on demand, but if you wish
 you can preload them by running `weave setup` - this is particularly
 useful for automated deployments, and ensures that there are no delays
-during later operations.
+during later operations. It's also possible to start the containers
+independently with `launch-router`, `launch-dns` and `launch-proxy`.
```
rade: Drop that.

site/weavedns.md
```diff
@@ -67,7 +77,7 @@ however specify an address in CIDR format manually. In this case you
 are responsible for ensuring that the IP addresses specified are
 uniquely allocated and not in use by any other container.
 
-Finally, WeaveDNS containers can be stopped with `stop-dns`.
+Finally, WeaveDNS containers can be stopped with `stop` or `stop-dns`.
```
rade: use similar language here as elsewhere, i.e. "in conjunction with... or independently..."

rade:
We need to put the README changes into a separate PR, to be merged when we make the next feature release, since the README on master needs to work with the currently released version.

awh:
> We need to put the README changes into a separate PR

#1011. Do you want me to move the 'fixes' from this PR to that one so that the issue is only closed once the docs are updated?

rade:
> Do you want me to move the 'fixes'

No.

test/150_connect_forget_test.sh
```diff
@@ -15,8 +15,8 @@ assert_peers() {
 
 start_suite "Connecting and forgetting routers after launch"
 
-weave_on $HOST1 launch
-weave_on $HOST2 launch
+weave_on $HOST1 launch-router
+weave_on $HOST2 launch-router
```
rade: why the change?

Also note that it's using `stop` and `launch` below.

rade: without the change it's failing for me with
````
test #6 "exec_on 192.168.48.11 c1 ping -nq -W 1 -c 1 10.2.1.7" failed:
	program terminated with code 1 instead of 0
````
But why? Please explain.

Comment (awh) [inline test/150_connect_forget_test.sh]: > why the change?

Hah yes this one was a bit non-obvious - it's to do with IPAM. As far as I can tell using `weave launch` meant that two rings would form, as each peer allocated an IP for weaveDNS - this meant that the peers couldn't connect later due to an invalid state error. I'm guessing that happens in a lazy fashion, so if nothing asks for an IP then the peering is successful.

Let me try `weave launch -initpeercount` instead.

> Also note that it's using stop and launch below.

Good spot - will adjust.

Comment (awh) [inline test/150_connect_forget_test.sh]: > Let me try weave launch -initpeercount instead.

Interesting results from this - the first `weave launch -initpeercount 2` fails:
````
$ time ./150_connect_forget_test.sh 
Connecting and forgetting routers after launch
Container 016b4593cc2184e2b5794994ffe3c892fc14bb20d14f262cc5b1f15b227c1867 not running.

real	0m23.154s
user	0m0.029s
sys	0m0.044s
````
I discussed this with @bboreham and he pointed me towards #1001 which features weaveDNS dying after twenty seconds due to failure to allocate an IP address to the mDNS interface. The root cause is different - in this case we're lacking quorum so the allocation request blocks - but the effect is the same.

Since this problem already affects `weave launch-dns`, I propose not to fix it here - I'll create a separate issue for it instead.

**Comment by awh** [comment]
>> Also note that it's using stop and launch below.

> Good spot - will adjust.

Done.

File: weave [hunk 5]
```diff
@@ -918,6 +921,114 @@ proxy_addr() {
...
+            *)
+                ARGS="$ARGS '$(echo "$1" | sed "s|'|'\"'\"'|g")'"
+                shift
+                ;;
+        esac
+    done
+    eval "set -- $ARGS"
+    if [ -z "$IPRANGE_SPECIFIED" ] ; then
+        IPRANGE="10.128.0.0/10"
+        if command_exists netcheck && ! netcheck $IPRANGE ; then
```
inercia: Shouldn't this be like:
```
if [ -z "$IPRANGE_SPECIFIED" ] ; then
       IPRANGE="10.128.0.0/10"
fi
if command_exists netcheck && ! netcheck $IPRANGE ; then
   ...
```

bboreham: It's intentional; if the user specifies a range then we leave checking it up to them.  Maybe they have a good reason for using a range that spans some existing route.

bboreham:
I found it a bit surprising that extra arguments like `-debug` only get passed to the router, not to dns or proxy.

rade:
`-debug` isn't documented. Are there any other common flags?

bboreham:
https://github.com/weaveworks/weave/blob/master/site/troubleshooting.md

rade:
I see. It's not in the 'usage' (and neither should it be, imo).

So `-debug` is the only one. We could make an exception for that, I suppose. File an issue.

bboreham:
Should update the README.

rade:
> Should update the README.

That is in #1012.

dpw:
This broke `bin/multiweave`.

rade:
> This broke `bin/multiweave`.

Well spotted. [Fixed](https://github.com/weaveworks/weave/commit/03ac7a86c6244e596bc8c199afcd75753818debb).

PR #1010, opened 2015-06-25
Try other peers for space if we get errors

### LINKED ISSUE #896: Long delay if ipam asks an absent host for space; can break launch-dns

bboreham:
Replacing #1005 which was against master not 1.0

Note I have done it a slow and stupid way in one commit, then improved the algorithmic complexity using a CDF and binary search in the next commit.

Fixes #896

### DIFF
--- a/ipam/allocate.go
+++ b/ipam/allocate.go
@@ -42,9 +42,12 @@ func (g *allocate) Try(alloc *Allocator) bool {
 	}
 
 	// out of space
-	if donor, err := alloc.ring.ChoosePeerToAskForSpace(g.r.Start, g.r.End); err == nil {
+	donors := alloc.ring.ChoosePeersToAskForSpace(g.r.Start, g.r.End)
+	for _, donor := range donors {
 		alloc.debugln("Decided to ask peer", donor, "for space in range", g.r)
-		alloc.sendSpaceRequest(donor, g.r)
+		if err := alloc.sendSpaceRequest(donor, g.r); err == nil {
+			break
+		}
 	}
 
 	return false
--- a/ipam/allocate.go
+++ b/ipam/allocate.go
@@ -44,8 +44,10 @@ func (g *allocate) Try(alloc *Allocator) bool {
 	// out of space
 	donors := alloc.ring.ChoosePeersToAskForSpace(g.r.Start, g.r.End)
 	for _, donor := range donors {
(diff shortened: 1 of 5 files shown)


rade:
https://github.com/weaveworks/weave/pull/1005#discussion_r33182574 still needs addressing

bboreham:
`sort.Search` is a binary search.
The algorithm implemented seems to me to be the same as the one you linked to.

rade:
`sort.Search` requires the slice to be sorted. That is the case initially but not after the first iteration, since you subtract from the value of all entries in the tail.

>  The algorithm implemented seems to me to be the same as the one you linked to.

No, it's not. The 2nd answer says "One way to shuffle a list is to assign random numbers to each element in the list and sort by those numbers. We can extend that idea, we just have to pick weighted random numbers."

rade:
My bad, it's the *first* answer I was referring to.

bboreham:
You seem to be quoting the first answer; http://programmers.stackexchange.com/a/233552/124690.  Which, as noted in the comments and at the top of the answer, gives the wrong result.

rade:
> Which, as noted in the comments and at the top of the answer, gives the wrong result.

Hence I said "It's not perfect, but perhaps good enough?"

bboreham:
I have implemented the scheme suggested by that StackOverflow answer.  It's a bit longer in Go than in Python...

rade:
I put some comments in https://github.com/bboreham/weave/commit/b9a069a62c04170d0f4a31a7e6fd49b79fc701cb

rade:
I've merged this, but now 1.0 won't merge cleanly into master. @bboreham please do that merge and resolve the conflict.

rade:
> now 1.0 won't merge cleanly into master

[fixed](https://github.com/weaveworks/weave/commit/ee2a66323e18b9e89c61dbed10973fd31dcdd796)

PR #1011, opened 2015-06-25
Update README to use new single `weave launch` command

awh:
Defer merge until #962 is released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ two containers, one on each host.
 
 First start weave on $HOST1:
 
-    host1$ weave launch && weave launch-dns && weave launch-proxy
+    host1$ weave launch
 
 this runs the weave router, DNS and proxy, each in their own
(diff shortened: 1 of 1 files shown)


rade:
I'd like to go a bit further here, if possible, and excise any mention of weaveDNS or the proxy from the README. On that note, I have been wondering whether we should rename `weave proxy-[env|config]` to just `weave env|config`.

rade:
This PR can become a collector for all the README changes that need to be merged at release time. I suggest we put the branch into the main repo, so its easier for everybody to work on it.

awh:
> excise any mention of weaveDNS or the proxy

I'll give it a go.

> On that note, I have been wondering whether we should rename weave proxy-[env|config]

Seems reasonable.

> I suggest we put the branch into the main repo, so its easier for everybody to work on it.

Good idea. I'll have to close this PR and make a new one.

rade:
I have raised #1014 for `weave env|config`.

PR #1015, opened 2015-06-25
Extend 'go get' to tests

bboreham:
The `-t` flag asks Go to fetch anything required for tests, which is more necessary now we have added testify.

I considered adding a `go get` line to the `make tests` target, but it's already unwieldy and not necessary if people do 'make' first.


rade:
This doesn't actually work. *sigh*

````
make clean; make; make tests
...
common/sched_queue_test.go:8:2: cannot find package "github.com/stretchr/testify/require" in any of:
...
````

tomwilkie:
bin/test in scope handles this nicely...

rade:
Fixed in 82da45a

PR #1016, opened 2015-06-25
Don't inspect the DF flag earlier than we need to

dpw:
The router extracts the DF flag from a received frame early on and passes it around.  But it doesn't use it in any significant way until it reaches the UDP forwarder code.  These commits defer extracting it until it is really needed.

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -11,7 +11,6 @@ import (
 	"runtime"
 	"strings"
 
-	"code.google.com/p/gopacket/layers"
 	"github.com/davecheney/profile"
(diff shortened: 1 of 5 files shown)


rade:
I've managed to [eliminate the DF passing API strangeness](https://github.com/rade/weave/commit/7de278d7ab3d9466ec4d88ed18b9012f1a2fa449). This should perhaps be squashed with https://github.com/dpw/weave/commit/64ac7745fec17dafe1c75a38f2999d43bba0e76e.

Am happy with the rest.

dpw:
Fine.

PR #1017, opened 2015-06-25
Flush messages from mock gossip in IPAM test

### LINKED ISSUE #982: ''No entries for peer found' failure in TestTransfer-4
From the [Travis CI report](https://travis-ci.org/weaveworks/weave/builds/67902970)...

````
--- FAIL: TestTransfer-4 (1.01s)
	util.go:134: Expected success, got 'No entries for peer found'
		goroutine 10 [running]:
		github.com/weaveworks/weave/testing.stackTrace(0x2adaf7e2b900, 0x0, 0x0)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:143 +0x88
		github.com/weaveworks/weave/testing.StackTrace(0x0, 0x0)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:138 +0x35
		github.com/weaveworks/weave/testing.Fatalf(0xc20807e5a0, 0x9a03b0, 0x1a, 0xc20913ed90, 0x1, 0x1)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:134 +0x27
		github.com/weaveworks/weave/testing.AssertSuccess(0xc20807e5a0, 0x2adaf7e384c0, 0xc2091052c0)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:114 +0x11a
		github.com/weaveworks/weave/ipam.TestTransfer(0xc20807e5a0)
			/home/travis/gopath/src/github.com/weaveworks/weave/ipam/allocator_test.go:264 +0x6cf
		testing.tRunner(0xc20807e5a0, 0xddcf98)
			/usr/local/go/src/testing/testing.go:447 +0xbf
		created by testing.RunTests
			/usr/local/go/src/testing/testing.go:555 +0xa8b
````

NB: the test was run against a branch off 1.0, though I doubt that makes a difference.

### PR DESCRIPTION
**Pull request by bboreham**
Refactored mock router, and added a `flush()` function which we then call every time gossip messages are generated in `TestTransfer()`, thus avoiding the problem that they may be dropped if the previous message has not been removed from the channel, or may not have arrived by the time we check what has happened.

`flush()` is not called in the other two tests that use the mock router; one of them hardly does anything, and the other benefits from the message-dropping behaviour, in that it makes it also a test of recovery from network issues.

Fixes #982.

Also replaced a `Sleep` with the `flush`, so the unit tests run faster.

### DIFF
--- a/ipam/testutils_test.go
+++ b/ipam/testutils_test.go
@@ -195,23 +195,26 @@ func AssertNothingSentErr(t *testing.T, ch <-chan error) {
 }
 
 // Router to convey gossip from one gossiper to another, for testing
-type gossipMessage struct {
-	isUnicast bool
-	sender    *router.PeerName
-	buf       []byte            // for unicast
-	data      router.GossipData // for broadcast
-	exitChan  chan struct{}
+type unicastMessage struct {
+	sender *router.PeerName
+	buf    []byte
+}
+type broadcastMessage struct {
+	data router.GossipData
+}
+type exitMessage struct {
+	exitChan chan struct{}
 }
 
 type TestGossipRouter struct {
-	gossipChans map[router.PeerName]chan gossipMessage
+	gossipChans map[router.PeerName]chan interface{}
 	loss        float32 // 0.0 means no loss
 }
 
 func (grouter *TestGossipRouter) GossipBroadcast(update router.GossipData) error {
 	for _, gossipChan := range grouter.gossipChans {
 		select {
-		case gossipChan <- gossipMessage{data: update}:
+		case gossipChan <- broadcastMessage{data: update}:
 		default: // drop the message if we cannot send it
 		}
 	}
@@ -221,7 +224,7 @@ func (grouter *TestGossipRouter) GossipBroadcast(update router.GossipData) error
 func (grouter *TestGossipRouter) removePeer(peer router.PeerName) {
 	gossipChan := grouter.gossipChans[peer]
 	resultChan := make(chan struct{})
-	gossipChan <- gossipMessage{exitChan: resultChan}
+	gossipChan <- exitMessage{exitChan: resultChan}
 	<-resultChan
 	delete(grouter.gossipChans, peer)
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ipam/testutils_test.go [hunk 1]
```diff
@@ -232,27 +246,29 @@ type TestGossipRouterClient struct {
...
-					message.exitChan <- true
+			case gossip := <-gossipChan:
+				if message, isExit := gossip.(exitMessage); isExit {
+					close(message.exitChan)
 					return
 				}
-
+				if message, isFlush := gossip.(flushMessage); isFlush {
+					close(message.flushChan)
+				}
```
rade: I reckon this would be easier to follow if the above were part of the type dispatch below. This does mean duplicating the loss conditional, but I think that is justifiable in this case.

rade:
> `flush()` is not called in the other two tests that use the mock router; one of them hardly does anything, and the other benefits from the message-dropping behaviour, in that it makes it also a test of recovery from network issues.

But message loss is part of the mock router, so surely it gets tested anyway?

I'd rather see `flush()` everywhere.

bboreham:
The `loss` parameter is not actually set in the current codebase.  But I could do that.

bboreham:
I don't think it makes sense to call `flush()` in `TestAllocatorFuzz()`; all interactions with the mock router are indirect, via `Allocator`, and we don't know which ones will generate gossip, nor do we care when it is delivered. All we care is that the process of allocating and freeing makes progress at some reasonable rate. (I should note that isn't tested either, but is perhaps a separate issue)

rade:
> I don't think it makes sense to call `flush()` in `TestAllocatorFuzz()`

ok.

rade:
anything left to do here?

bboreham:
I don't think there is anything left for the original issue.  I ran the ipam unit tests repeatedly in a loop for an hour with no failures.

rade:
ok. pls assign back to me then.

rade:
*argh*. Branched incorrectly. This was meant to be off 1.0.

bboreham:
Replaced by #1024

PR #1026, opened 2015-06-26
Make proxy listen on unix:///var/run/weave.sock

### LINKED ISSUE #1003: allow proxy to listenen on unix socket

tomwilkie:
- proxy still listens on 0.0.0.0; can now listen on multiple endpoints
- bind mounts /var/run instead of /var/run/docker.sock so the host can see the proxy sock.
- copies owner & permissions from /var/run/docker.sock to /var/run/weave.sock

Fixes #1003

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -2,10 +2,13 @@ package proxy
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"regexp"
 	"strings"
+	"syscall"
 
 	"github.com/fsouza/go-dockerclient"
 	. "github.com/weaveworks/weave/common"
@@ -15,6 +18,8 @@ const (
 	defaultCaFile   = "ca.pem"
 	defaultKeyFile  = "key.pem"
 	defaultCertFile = "cert.pem"
+	dockerSock      = "/var/run/docker.sock"
+	dockerSockUnix  = "unix://" + dockerSock
 )
 
 var (
@@ -24,7 +29,7 @@ var (
 )
 
 type Config struct {
-	ListenAddr    string
+	ListenAddrs   []string
 	NoDefaultIPAM bool
 	TLSConfig     TLSConfig
 	Version       string
@@ -45,7 +50,7 @@ func NewProxy(c Config) (*Proxy, error) {
 		Error.Fatalf("Could not configure tls for proxy: %s", err)
 	}
 
-	client, err := docker.NewClient("unix:///var/run/docker.sock")
+	client, err := docker.NewClient(dockerSockUnix)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


weave
```diff
@@ -921,8 +921,8 @@ proxy_args() {
 }
 
 proxy_addr() {
-    if addr=$(docker logs $PROXY_CONTAINER_NAME 2>/dev/null | head -n3 | grep -oE "proxy listening on .*"); then
-      addr=${addr##* }
+    if addr=$(docker logs $PROXY_CONTAINER_NAME 2>/dev/null | grep -oE "proxy listening on .*" | head -n1); then
```
rade: The `head -n3` was there so we don't end up fetching the entire log.

rade:
Need to update the `weave` usage. And the proxy docs.

weave
```diff
@@ -921,8 +921,8 @@ proxy_args() {
...
+      addr=${addr##* tcp://}
```
rade: There is no guarantee that the first address appearing in the output is the one we want.
Plus, because there is no separation between the listen and serve phases, we can potentially have lots of output (from the proxy doing its stuff) interspersed with the "proxy listening on ..."

proxy/proxy.go
```diff
@@ -83,17 +87,84 @@ func (proxy *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	proxy.Intercept(i, w, r)
 }
 
-func (proxy *Proxy) ListenAndServe() error {
-	listener, err := net.Listen("tcp", proxy.ListenAddr)
+func (proxy *Proxy) ListenAndServe() {
+	errs := make(chan error)
+	for _, addr := range proxy.ListenAddrs {
+		Info.Println("proxy listening on", addr)
+	}
```
rade: This breaks the intended semantics of "wait_for_log` in the weave script.

You really ought to get all the listening (and associated error checking) out of the way *before* printing out any "proxy listening on".

rade:
Neither the docs nor the usage mention that *multiple* -H can be specified.

weave
```diff
@@ -926,10 +926,8 @@ proxy_args() {
 proxy_addr() {
     if addr=$(docker logs $PROXY_CONTAINER_NAME 2>/dev/null | head -n3 | grep -oE "proxy listening on .*"); then
       addr=${addr##* }
-      host=${addr%:*}
-      [ "$host" = "0.0.0.0" ] && host=$PROXY_HOST
-      port=${addr#*:}
-      echo "${1}tcp://${host:-$PROXY_HOST}:${port:-$PROXY_PORT}"
+      addr=$(echo ${addr} | sed "s/0.0.0.0/$PROXY_HOST/g")
+      echo "${1}${addr}"
```
rade: or instead of three lines just have

`echo "${1}${addr##* }" | sed "s/0.0.0.0/$PROXY_HOST/"`

(no need for the `g`)

PR #1038, opened 2015-06-29
Disable hostname derivation when container name is too long

### LINKED ISSUE #1006: long container name cause breakage when deriving hostname
````
$ weave launch && weave launch-dns && weave launch-proxy
233175e68b3f1acdb4b575365b65873bb4e6f2fdde4540ee58fb5a940fa1cc0e
eab7f5de2f66576835a4e119570304ab16b6bbae26a8f752315d46b6c5d29e53
9f22c0618999ba92de2ca1fe038ca6700470e6ee5a5c3e6b8adae7772c23deaf
$ eval $(weave proxy-env)
$ docker run --name=a123456789012345678901234567890123456789012345678901234567890 -ti ubuntu
invalid argument
Error response from daemon: 18944953c27c4ed8f71f2fbd1989e7bd8f79cbd9c1a77b23c31d6fd25f94d093 container is not running.
````

### PR DESCRIPTION
**Pull request by awh**
Fixes #1006.

### DIFF
--- a/weave
+++ b/weave
@@ -714,8 +714,13 @@ dns_args() {
     DNS_ARGS="--dns $DOCKER_BRIDGE_IP"
     if [ -n "$NAME_ARG" -a -z "$HOSTNAME_SPECIFIED" ] ; then
         DOMAIN=$(dns_domain)
-        DNS_ARGS="$DNS_ARGS --hostname=$NAME_ARG.${DOMAIN%.}"
-        HOSTNAME_SPECIFIED=1
+        HOSTNAME="$NAME_ARG.${DOMAIN%.}"
+        if [ ${#HOSTNAME} -gt 64 ] ; then
+            echo "Container name too long to be used as hostname" >&2
+        else
+            DNS_ARGS="$DNS_ARGS --hostname=$HOSTNAME"
+            HOSTNAME_SPECIFIED=1
+        fi
     fi
     if [ -z "$DNS_SEARCH_SPECIFIED" ] ; then
       if [ -z "$HOSTNAME_SPECIFIED" ] ; then
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -13,6 +13,8 @@ import (
 	"github.com/weaveworks/weave/nameserver"
 )
 
+const MaxDockerHostname = 64
+
 type createContainerInterceptor struct{ proxy *Proxy }
 
 type createContainerRequestBody struct {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by rade** [comment]
do you know where the limitation of 64 for the *entire* name comes from? Is it a docker bug?

**Comment by awh** [comment]
I suspect it's due to it passing the fully qualified name into `sethostname`, which according to `getconf HOST_NAME_MAX` may be only 64 characters long. You're not  supposed to do that (the UTS namespace has two values, one for `sethostname` and one for `setdomainname`) but in my experience it's a pretty common misconfiguration, usually inflicted on systems by putting the FQDN into `/etc/hostname`. I'll confirm my findings and may raise an issue against Docker, but there's a good chance that stuff has been built expecting this broken behaviour so they may be unwilling to change it...

**Comment by awh** [comment]
Quick detour via the kernel source reveals that both the hostname and domainname values are limited to 64 characters a piece. The limit is compiled in.

File: weave [hunk 1]
```diff
@@ -711,8 +711,13 @@ dns_args() {
     DNS_ARGS="--dns $DOCKER_BRIDGE_IP"
     if [ -n "$NAME_ARG" -a -z "$HOSTNAME_SPECIFIED" ] ; then
         DOMAIN=$(dns_domain)
-        DNS_ARGS="$DNS_ARGS --hostname=$NAME_ARG.${DOMAIN%.}"
-        HOSTNAME_SPECIFIED=1
+        HOSTNAME="$NAME_ARG.${DOMAIN%.}"
+        if [ ${#HOSTNAME} -gt 64 ] ; then
```
inercia: Maybe `if [ ${#HOSTNAME} -gt ``getconf HOST_NAME_MAX`` ] ; then`?

awh:
OK I've not traced it through completely because it's quite convoluted, but although the `runconfig.Config` at the top level has separate `Hostname` and `Domainname` fields, by the time it wends its way into libcontainer's `config.Config` (via a serialisation-over-fd step in the middle, hence the difficulty in tracing) the `Domainname` field is missing from the struct entirely and the `Hostname` field (which I'm *assuming* by now contains the FQDN) gets fed straight to `syscall.Sethostname`. Do you want me to raise it with Docker @rade?

weave
awh: Not supported by Alpine's `getconf` unfortunately...

rade:
So a `-h <hostname>` [gets split](https://github.com/docker/docker/blob/v1.7.0/runconfig/parse.go#L226) into a hostname and domain name portion, which is what [appears](https://docs.docker.com/reference/api/docker_remote_api_v1.19/#create-a-container) in the Docker Remote API. But there is no `setdomainnname` call, only `sethostname`, so somewhere along the way the hostname and domainname must get munched together.

> Do you want me to raise it with Docker @rade?

Yes.

bboreham:
Note message from linter:

    ./proxy/create_container_interceptor.go:115:3: var trimmedDnsDomain should be trimmedDNSDomain

awh:
>> Do you want me to raise it with Docker @rade?

> Yes.

docker/docker#14282

> Note message from linter:

Fixed.

rade:
We should add a smoke test.

awh:
* Added a smoke test
* Rebased atop logrus merge
* Include container name in proxy log message

rade:
The test no longer asserts that the container start succeeds, which is far more important than checking the warning message.

awh:
I've reverted the test change back.

rade:
needs re-basing off 1.0

awh:
I tried adjusting the merge base of this PR via the REST API last night and it isn't possible - will close and open another PR.

Issue #664, opened 2015-05-15
replace homegrown logging library with something off-the-shelf


paulbellamy:

tomwilkie:
Docker use logrus...

paulbellamy:
Logrus is fine. https://github.com/inconshreveable/log15 is also ok.

paulbellamy:
What advantages do we want from changing the logging library? structured logging?

The logging code we have is only 40 lines, and with either logrus, or log15 we'll need about the same amount of setup (maybe *slightly* less, but not enough of a saving to justify it alone).

awh:
Support for ordered log levels (see #329). Right now we have four separate output streams (debug, info, warn, error) but they aren't really levels - implementing something like 'show me info or above' would be a bit of a faff...

bboreham:
I took a look at the main results from a web search of 'best golang logger':

https://github.com/cihub/seelog seems extremely full-featured, probably much more than we need. Over 10,000 lines including tests.
https://github.com/Sirupsen/logrus is ~3200 lines.
https://github.com/inconshreveable/log15 is ~3000 lines.
https://github.com/op/go-logging is ~2000 lines.

Based on my reading so far, I could not find any slam-dunk reason to prefer one of the smaller ones, so am planning to just use the same as Docker - logrus.  It seems to be actively maintained.

rade:
I think one requirement that we have is that we don't want our log statements to become much more unwieldy than they currently are, and without introducing our own custom wrapper functions.

rade:
> don't want ... much more unwieldy

So on that note, if we do go with logrus then we shouldn't use its 'fields' feature, at least to start with.

Issue #329, opened 2015-01-08
offer choice of all log levels


rade:
We use four log levels in our code - debug, info, warn, error - yet only allow selection of "all except debug" (the default) and "all" (when `-debug` is supplied).

inercia:
I think _warn_ and _error_ should never be silenced, right?. I cannot imagine a situation where you don't want to see errors. If this is the case, that would leave _info_ and _debug_ as optional. Maybe we could `-verbose`/`-v` for enabling _info_, and a second `-v` or `-debug` could enable _debug_...

PR #1044, opened 2015-06-29
Use off-the-shelf logger logrus

### LINKED ISSUE #1039: All log messages should go to stderr
Currently, in the router and nameserver containers, Debug, Info and Warning messages go to stdout while Error and Fatal messages go to stderr.

This is unexpected for some people, and not really compatible with the off-the-shelf libraries considered in #664.

So we should just send all log messages to stderr.

bboreham:
As discussed in #664 

We have our own formatting routine, because the standard logrus one has surprising behaviour like adding extra spaces on the end of lines and inserting colour-changing escape sequences.

One side-effect of this change is that all logging now comes on stderr, rather than some on stdout and some on stderr as before.  The weave script needs two changes to accommodate this. Closes #1039.

The previous style of having `Debug.Println()` or `Error.Println()` isn't supported by logrus, so such code changes to `Log.Debugln()` and `Log.Errorln()` respectively.  We provide an `Info` reference so `Info.Println()` still works.

### DIFF
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -44,7 +44,7 @@ func NewClient(apiPath string) (*Client, error) {
 func (c *Client) AddObserver(ob ContainerObserver) error {
 	events := make(chan *docker.APIEvents)
 	if err := c.AddEventListener(events); err != nil {
-		Error.Printf("[docker] Unable to add listener to Docker API: %s", err)
+		Log.Errorf("[docker] Unable to add listener to Docker API: %s", err)
 		return err
 	}
 
@@ -63,7 +63,7 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
 // IsContainerNotRunning returns true if we have checked with Docker that the ID is not running
 func (c *Client) IsContainerNotRunning(idStr string) bool {
 	if !IsContainerID(idStr) {
-		Debug.Printf("[docker] '%s' does not seem to be a container id", idStr)
+		Log.Debugf("[docker] '%s' does not seem to be a container id", idStr)
 		return false
 	}
 
@@ -75,6 +75,6 @@ func (c *Client) IsContainerNotRunning(idStr string) bool {
 	if _, notThere := err.(*docker.NoSuchContainer); notThere {
 		return true
 	}
-	Error.Printf("[docker] Could not check container status: %s", err)
+	Log.Errorf("[docker] Could not check container status: %s", err)
 	return false
 }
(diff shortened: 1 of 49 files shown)


rade:
I am curious what the resolution to #329 would look like with this change. To date that's the only justification provided for dragging in an external logging library.

bboreham:
#329 is implemented at https://github.com/weaveworks/weave/commit/11ef0fe0312fc84891924433951890a36d94a96e

(actually more than was asked for, since #329 is only about DNS)

inercia:
LGTM.

I was wondering if logrus could make things easier for identifying the module/package where the log line was generated, so we could get rid of things like `[mdns]` in `"[mdns] ..."` lines...

bboreham:
@inercia not sure whether you meant easier in the code that generates the log line or easier for the person reading the log file.  I'll answer the former.

The typical pattern I've seen is using `runtime.Caller`, but it requires that you know how many stack frames to go up before you hit the "real" call. Which is worse than where we are now, IMO.

It should be possible to use logrus' "fields" feature to set the module and bring it out again in the formatter.

inercia:
@bboreham Yes, the `runtime.Caller` would be too hackish for me, so I guess we could use `log.WithFields(log.Fields{"module": "mdns",})`... or maybe we should create a common `fun LogM(m string) { return Log.WithFields(log.Fields{"module": m}) } `

bboreham:
I suggest (a) that should be filed as a separate issue and (b) we should carefully consider what the log-file format of such messages should be.  It is very useful to be able to reliably search them, so a format like `module=mdns;` is more search-friendly and less human-friendly, and `[mdns]` more the other way round.

inercia:
I agree with both points. Parseability is probably not really necessary. I was more concerned with a) a common calling convention and b) a common log line formatting...

bboreham:
Quick note: I should fix the way millisecond numbers are currently stripped of trailing zeros.

bboreham:
I fixed the milliseconds and rebased off master.

paulbellamy:
Why not get rid of `common.Info`?

bboreham:
It was an extra ~50 edits, and `Info.Printf` seemed a reasonable way to do logging, so I aliased Log to make it work.

I see the argument that it's more consistent to lose it. I could be swayed.

paulbellamy:
sway sway sway

inercia:
I would remove it... :+1:

rade:
remove it.

bboreham:
Removed.

PR #1045, opened 2015-06-29
Add -initpeercount to usage

### LINKED ISSUE #1041: initpeercount is missing from usage
...even though we do mention it in the docs.

squaremo:
Fixes #1041.


rade:
> NB it's a suboption of -iprange, since it only applies if you turn IPAM on.

...which we do by default. Hence it should be shown separately.

rade:
also, usage should fit into 80 columns.

squaremo:
> usage should fit into 80 columns.

Like the `weave run` line?

rade:
> Like the `weave run` line?

*sigh* that got longer when we introduced `weave launch-proxy` and then `weave launch-router`. Needs fixing.

rade:
> `weave run` ... needs fixing.

9127eb8

PR #1050, opened 2015-06-30
Resolve unqualified names in the local WeaveDNS domain

### LINKED ISSUE #987: resolve unqualified names in weavedns domain
musl/busybox/alpine's dns resolver [is broken](https://github.com/weaveworks/weave/issues/985#issuecomment-114448705); it doesn't handle the 'search' directive. As a result unqualified names won't resolve in containers derived from busybox/alpine.

We could patch over that by getting weavedns to resolve unqualified names in the weavedns domain.

inercia:
Fixes #987

### DIFF
--- a/nameserver/names.go
+++ b/nameserver/names.go
@@ -1,0 +1,28 @@
+package nameserver
+
+import (
+	"strings"
+
+	"github.com/miekg/dns"
+)
+
+// Get the number of components of a name
+// For example, "redis" is 1, "redis.weave.local"  is 3...
+func nameNumComponents(name string) int {
+	if len(name) == 0 || name == "." {
+		return 0
+	}
+	return strings.Count(dns.Fqdn(name), ".")
+}
+
+// normalize a name by
+// a) adding the local domain when it is a single component name
+// b) converting the result to FQDN
+func normalizeName(n string, domain string) string {
+	// we should never add anything to a FQDN like "redis.", but that's the only thing that
+	// some broken resolvers will do: to convert the name to FQDN...
+	if nameNumComponents(n) == 1 {
+		n = dns.Fqdn(n) + domain
+	}
+	return dns.Fqdn(n)
+}
--- a/nameserver/server.go
+++ b/nameserver/server.go
@@ -279,9 +279,6 @@ func (s *DNSServer) Stop() error {
 // We must create two DNS request multiplexers, depending on the protocol used by
 // clients (as we use the same protocol for asking upstream servers)
 func (s *DNSServer) createMux(proto dnsProtocol) *dns.ServeMux {
-	failFallback := func(w dns.ResponseWriter, r *dns.Msg) {
-		w.WriteMsg(makeDNSFailResponse(r))
-	}
 	notUsHandler := s.notUsHandler(proto)
 	notUsFallback := func(w dns.ResponseWriter, r *dns.Msg) {
 		Log.Infof("[dns msgid %d] -> sending to fallback server", r.MsgHdr.Id)
@@ -291,7 +288,7 @@ func (s *DNSServer) createMux(proto dnsProtocol) *dns.ServeMux {
 	// create the multiplexer
 	m := dns.NewServeMux()
 	m.HandleFunc(s.Zone.Domain(), s.localHandler(proto, "Query", dns.TypeA,
-		s.Zone.DomainLookupName, makeAddressReply, s.Zone.ObserveName, failFallback))
+		s.Zone.DomainLookupName, makeAddressReply, s.Zone.ObserveName, failHandleFunc))
 	m.HandleFunc(RDNSDomain, s.localHandler(proto, "Reverse query", dns.TypePTR,
 		s.Zone.DomainLookupInaddr, makePTRReply, s.Zone.ObserveInaddr, notUsFallback))
 	m.HandleFunc(".", s.notUsHandler(proto))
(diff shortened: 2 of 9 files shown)


nameserver/names.go
```diff
@@ -0,0 +1,28 @@
...
+	if len(name) == 0 || name == "." {
+		return 0
+	}
+	return strings.Count(dns.Fqdn(name), ".")
+}
+
+// normalize a name by
+// a) adding the local domain when it is a single component name
+// b) converting the result to FQDN
+func nameNormalize(n string, domain string) string {
```
rade: should be called `normalizeName`. And `n` should be `name`.

inercia: `normalizeName` would break the scheme where all functions in `names.go` have the form `name*`...

rade:
I think we want some smoke test for this.

Would like to hear from @bboreham and @awh about any gotchas to this 'hack'.

nameserver/names.go
rade: There are only two functions!

inercia: That is no excuse! :smile:

bboreham:
It seems plausible to me that a user has some set-up where they are already using unqualified names in their environment outside of Docker.  So whilst I agree with the premise of #987, it might be safer to have an option to disable this new behaviour.

inercia:
@bboreham suggestion seems reasonable. I will add a command line flag...

nameserver/names.go
rade: We use verb+noun for function names everywhere (modulo some exceptions, which I am sure exist).

rade:
Nooooooo! Not more command line flags! We can think about adding that when we have a user running into problems because of it.

nameserver/names.go
inercia: Ok, I'll rename the function, but IMO noun+verb is a better scheme (it groups related functions when listing alphabetically, and it makes things easier when looking for something...)

inercia:
Ok, then no more command line flags...

nameserver/server.go
```diff
@@ -372,9 +369,19 @@ func (s *DNSServer) localHandler(proto dnsProtocol, kind string, qtype uint16,
 func (s *DNSServer) notUsHandler(proto dnsProtocol) dns.HandlerFunc {
 	dnsClient := proto.GetNewClient(DefaultUDPBuflen, s.timeout)
 
+	localNameResolver := s.localHandler(proto, "Query", dns.TypeA,
+		s.Zone.DomainLookupName, makeAddressReply, s.Zone.ObserveName, failHandleFunc)
+
```
rade: btw, it's great how the `localHandler` abstraction we built makes this so easy and concise.

awh:
> Would like to hear from @bboreham and @awh about any gotchas to this 'hack'.

It will prevent fallback lookup of TLDs. Probably not common use case though.

inercia:
That's right: `dig @weavedns .com` would be equivalent to `dig @weavedns com.weave.local.`, but, as @awh mentioned, it is not a big deal...

rade:
@inercia anything left to do here

inercia:
If there are no more comments on your side, I think I'm done with it...

rade:
> I think I'm done with it...

ok. next time please assign back to me.

test/230_dns_unqualified_test.sh
```diff
@@ -26,4 +26,8 @@ check c2
 check c3
 check "$container"
 
+# check that unqualified names are automatically qualified when looking up
+weave_on $HOST1 dns-add $C2 c2 -h name1.$DOMAIN
+assert_dns_a_record $HOST1 c2 name1 $C2 name1.$DOMAIN
+
```
rade: Is this test actually failing on master? I wouldn't have thought so, since the image we use in `start_container_with_dns` has a functioning resolver library and hence will append the domain to the query.

I also don't understand what the point is of adding another record and extending the `assert_dns_record` helper function, instead of using `check` defined above.

So, imo, this should just start another container, `c5`, with a broken resolver library (`start_container` will do that), and run `check c5`. That *should* fail on master and succeed here.

inercia: You are right: the intention was to test it with a broken resolver. I'll fix it.

inercia:
Seems that alpine will not even add a `.` at the end of a short name like `redis`, so our DNS server mux is not catching those names. When a user does a `getent hosts redis` in alpine, our DNS server will receive a query about just `redis`. Not `redis.`, let alone `redis.weave.local`. And the problem is that we cannot install a handler in our DNS library for that kind of names...

rade:
> the problem is that we cannot install a handler in our DNS library for that kind of names...

why not?

inercia:
Because the minimum pattern [seems to be](https://github.com/miekg/dns/blob/master/server.go#L143) `.`

rade:
@awh @bboreham any idea what to do with this?

rade:
@awh @bboreham ping

awh:
Raise an issue against `miekg/dns` to allow the empty string so it can serve broken resolvers? I had a look for musl's issue tracker to see if this is getting fixed soon enough for us to ignore it, [there doesn't appear to be one](http://www.openwall.com/lists/musl/2014/09/05/1)...

bboreham:
Doesn't [this line](https://github.com/miekg/dns/blob/master/server.go#L135) mean that pattern `"."` will match a string like `"redis"` ?

inercia:
@bboreham I'm think it doesn't... `.` is just the minimum FQDN name.

bboreham:
I added a debug print at the top of `notUsHandler` and tried your suggested test; it seems to get the period appended:

````
$ sudo weave reset
$ sudo weave launch-router
a6d98c4c87518cd78c46e26cac11c6bc04d09e83a98b301ee6498f0c979dc9a5
$ sudo weave launch-dns -debug
027cc73866fae420fb08ac0660bffde24cec3cb306be6f64381ab45ec3c9378c
$ sudo weave run -ti gliderlabs/alpine /bin/sh
5d22fbb050d3c4ceb4b192fb5e1b7eb0d0a5729bd0f842eb1d4a528e25c4d487
$ sudo docker exec 5d22 getent hosts redis
$ sudo docker logs weavedns
INFO: 2015/07/07 10:19:31.148076 [main] WeaveDNS version git-149fe870abe1

INFO: 2015/07/07 10:19:31.148143 [main] Waiting for mDNS interface ethwe to come up
INFO: 2015/07/07 10:19:31.248575 [main] Interface ethwe is up
INFO: 2015/07/07 10:19:31.248624 [main] Waiting for HTTP interface eth0 to come up
INFO: 2015/07/07 10:19:31.248959 [main] Interface eth0 is up
INFO: 2015/07/07 10:19:31.249128 [zonedb] Using mDNS on &{Index:127 MTU:65535 Name:ethwe HardwareAddr:2e:8a:e3:89:84:16 Flags:up|broadcast|multicast}
INFO: 2015/07/07 10:19:31.250556 [docker] Using Docker API on unix:///var/run/docker.sock: &[KernelVersion=3.13.0-53-generic Os=linux Version=1.6.2 ApiVersion=1.18 Arch=amd64 GitCommit=7c8fca2 GoVersion=go1.4.2]
INFO: 2015/07/07 10:19:31.251298 [main] HTTP API listening on 172.17.0.26:6785
INFO: 2015/07/07 10:19:31.251358 [dns] Upstream server(s): &{Servers:[8.8.8.8 8.8.4.4] Search:[localdomain] Port:53 Ndots:1 Timeout:5 Attempts:2}
INFO: 2015/07/07 10:19:31.251423 [dns] Cache: 8192 entries
DEBU: 2015/07/07 10:19:31.251454 [dns] Creating UDP DNS client with 5s timeout
DEBU: 2015/07/07 10:19:31.251464 [dns] Creating UDP DNS client with 5s timeout
DEBU: 2015/07/07 10:19:31.251480 [dns] Creating TCP DNS client with 5s timeout
DEBU: 2015/07/07 10:19:31.251485 [dns] Creating TCP DNS client with 5s timeout
INFO: 2015/07/07 10:19:31.252286 [dns] Listening for DNS on :53 (UDP)
INFO: 2015/07/07 10:19:31.253136 [dns] Listening for DNS on :53 (TCP)
INFO: 2015/07/07 10:19:31.410668 [zonedb] Container e2ab206d967b9c699ce4dd5c32c43a44a108ac038d54f89aa4fff32424d02865 down. Removing records
INFO: 2015/07/07 10:19:35.826293 [zonedb] Container f0210ae3ffa7ece9e3315e03523ebc3bff34720c83e933eb3a311170c69165b6 down. Removing records
INFO: 2015/07/07 10:19:36.362694 [zonedb] Container a8d1b78881706ccda5b39d401155f6167691cf6d4719dc66fd6afbed7805f27f down. Removing records
DEBU: 2015/07/07 10:19:44.881717 NotUsHandler ;; opcode: QUERY, status: NOERROR, id: 48265
;; flags: rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;redis.	IN	 A

DEBU: 2015/07/07 10:19:44.881818 [dns msgid 48265] Short name 'redis.' -> resolving locally
DEBU: 2015/07/07 10:19:44.881896 [dns msgid 48265] Query: {Name:redis. Qtype:1 Qclass:1}
DEBU: 2015/07/07 10:19:44.881926 [cache msgid 48265] addind in pending state
DEBU: 2015/07/07 10:19:44.882002 [zonedb] Looking for name 'redis.weave.local.' in local(&remote) database
DEBU: 2015/07/07 10:19:44.882037 [zonedb] 'redis.weave.local.' not in local database... trying with mDNS
DEBU: 2015/07/07 10:19:44.882272 [mdns] srv: mDNS query from local host ('10.128.0.1:47436'): ignored
DEBU: 2015/07/07 10:19:45.382892 [zonedb] mDNS lookup error for 'redis.weave.local.': Unable to find redis.weave.local.
INFO: 2015/07/07 10:19:45.383049 [dns msgid 48265] No results for type A query for 'redis.' [caching no-local]
````

**Comment by rade** [comment]
What is left to do here?

**Comment by inercia** [comment]
I will close this PR, as it will have to be reimplemented once the new DNS code is in place...

**Comment by rade** [comment]
> I will close this PR, as it will have to be reimplemented once the new DNS code is in place.

If this is ready to merge then we should do so. What is missing?

**Comment by inercia** [comment]
It seems that broken resolvers add a `.` after all, as @bboreham discovered. So this PR could be reviewed again...

File: test/230_dns_unqualified_test.sh [hunk 4]
```diff
@@ -6,24 +6,26 @@ C1=10.2.0.78
...
 start_suite "Resolve unqualified names"
 
-launch_dns_on $HOST1 10.2.254.1/24
+launch_dns_on $HOST1 10.2.254.1/24 --debug
 
 start_container          $HOST1 $C1/24 --name=c1 -h $NAME
 start_container_with_dns $HOST1 $C2/24 --name=c2 -h seetwo.$DOMAIN
 start_container_with_dns $HOST1 $C3/24 --name=c3 --dns-search=$DOMAIN
 container=$(start_container_with_dns $HOST1 $C4/24)
+start_container          $HOST1 $C5/24 --name=c5 -h short.$DOMAIN
```
inercia: Not needed.

bboreham:
LGTM

nameserver/server.go
```diff
@@ -373,32 +370,42 @@ func (s *DNSServer) localHandler(proto dnsProtocol, kind string, qtype uint16,
...
+				if reply != nil && reply.Rcode != dns.RcodeSuccess {
+					Log.Debugf("[dns msgid %d] Failure reported by %s for query %s",
+						r.MsgHdr.Id, server, q.Name)
+					continue
+				}
+				Log.Debugf("[dns msgid %d] Given answer by %s for query %s",
 					r.MsgHdr.Id, server, q.Name)
-				continue
+				w.WriteMsg(reply)
+				return
```
rade: Previously, an empty query would follow what is now the `nc > 1` path. Why the change?

inercia: Because, once we have spent some CPU cycles calculating the number of components in the name, we could use that number for skipping empty names...

rade:
I've had to revert this since it fails the integration tests on master.

PR #1051, opened 2015-06-30
Draw attention to --hostname precedence

### LINKED ISSUE #1046: Clarify weaveDNS behaviour around --hostname
The docs don't sufficiently draw attention to the fact that `--hostname` disables registration unless the specified FQDN is in the `weave.local` domain.

awh:
Fixes #1046.


awh:
Updated to substitute markdown blockquote for alert div.

Issue #1030, opened 2015-06-26
frequent failure in 150_connect_forget_test


rade:
e.g. [here](https://circleci.com/gh/weaveworks/weave/1323)

````
---= Running 150_connect_forget_test.sh =---
Connecting and forgetting routers after launch
7e52b95f812d0bf678250a792b650fac3be0950e5fcd7a2a5a1f8ab9479129e7
fedc53aa1431eb10bcd877f3d19b5aefc7ad1cd3f7da28c6e81755ac99543d0a
fcb9ef937002787a01ba8a53c3e23d26c9e339c398e3b7292e662f738be577de
1c0712d8f0f27a8e659edef20370e0da8eb8db67946aafd08b60effb060ef201
test #6 "exec_on host1-1323-1.us-central1-a.positive-cocoa-90213 c1 ping -nq -W 1 -c 1 10.2.1.7" failed:
	program terminated with code 1 instead of 0
1 of 11 tests failed in 17.464s.
````
I suspect it's just timing; we've just connected to another peer and odds are the topology and routing info hasn't fully updated yet.

awh:
I've seen this a couple of times locally with 110_encryption_test.sh, not quite enough times to make an issue of it yet but I'll keep an eye on it.

rade:
> I've seen this a couple of times locally with 110_encryption_test.sh

So have I, but that's a separate issue. Though possibly the same cause.

awh:
@bboreham do you have any thoughts on whether #1052 fixes this too?

bboreham:
See https://github.com/weaveworks/weave/pull/1052#issuecomment-117635018

rade:
I haven't seen this fail for a long time. Let's assume #1052 fixed it.

bboreham:
Failed on my local PC today.

bboreham:
I see that the two peers can disconnect because they have inconsistent IPAM ring data.
Suggest starting with `--init-peer-count=2`.

tomwilkie:
The test does the following:
1. start 2 routers, not connected
2. start 2 containers (1 on each), with manually assigned IPs
3. connect the 2 routers

The change which stops this working (#1200) is that we now check manually assigned IPs with IPAM, which means a ring is being created at 2 and 3 is failing.

bboreham:
Is it wrong to add the `--init-peer-count=2`?  I guess it's unnecessary given your change, but I'm thinking we should follow our own rules.

Issue #1042, opened 2015-06-29
Frequent failure in 110_encryption_test


awh:
@bboreham and @rade report seeing this in addition to myself. Root cause possibly the same as #1030.

bboreham:
I tried raising the count of packets sent by `ping` to 3, on the hypothesis that it was perhaps losing just the first packet, and subsequent packets would get through.  No change; it still fails around 20% of the time.

As part of this exercise, at one point I had two ping commands - one that just printed out what it saw and one that was the original test.  In every test I ran, either both succeeded or both failed.

I also tried changing `weave launch` to `weave launch-router`, to avoid starting weavedns, but still saw some failures.

bboreham:
With `-pktdebug` turned on,
Logfile from successful run:

````
INFO: 2015/06/30 10:33:20.09492 Command line options: map[wait:20 iface:ethwe iprange:10.128.0.0/10 name:4e:14:17:0a:c6:44 nickname:host1 pktdebug:true port:6783]
INFO: 2015/06/30 10:33:20.095032 Command line peers: []
INFO: 2015/06/30 10:33:20.095184 Communication between peers is encrypted.
INFO: 2015/06/30 10:33:20.095512 Our name is 4e:14:17:0a:c6:44(host1)
INFO: 2015/06/30 10:33:20.095541 Assuming quorum size of 1
INFO: 2015/06/30 10:33:20.096874 [docker] Using Docker API on unix:///var/run/docker.sock: &[KernelVersion=3.16.0-37-generic Version=1.7.0 ApiVersion=1.19 GitCommit=0baf609 GoVersion=go1.4.2 Os=linux Arch=amd64]
INFO: 2015/06/30 10:33:20.099151 Sniffing traffic on &{1284 65535 ethwe ee:e6:0c:b6:42:b9 up|broadcast|multicast}
INFO: 2015/06/30 10:33:20.099419 Discovered our MAC ee:e6:0c:b6:42:b9
INFO: 2015/06/30 10:33:20.242878 Sniffed 90 bytes ( a929601304384903349e74bdba4d727cbea7d612c9f9f08b53d40167d56497c0 )
INFO: 2015/06/30 10:33:20.242916 Discovered local MAC 4e:14:17:0a:c6:44
[...]

PR #1052, opened 2015-06-30
Call `weave reset` earlier in smoke-test set-up, to improve reliability

bboreham:
As discussed at #1042, calling `conntrack -D` before shutting down older containers seems to improve the reliability of our smoke tests.

Also it gives a more gentle shut-down, so a better model to copy in case someone is trying to do something similar.


rade:
We should check whether this also happens to resolve #1030.

bboreham:
I tried, but couldn't get test 150 to fail in the 'before' case, after running it over a hundred times.

rade:
Please branch this off 1.0, since I don't want spurious test failures to contaminate 1.0 CI.

bboreham:
Replaced by #1054

Issue #480, opened 2015-03-24
check that subnets aren't in use


rade:
Several users have run into problems with weave by accidentally picking CIDRs in `weave run` or `weave expose` for networks which are in use by the host.

We could perform some basic checks to prevent common cases of that, i.e. check overlaps with other interfaces and routing table entries on the host.

I guess this may be partially superseded by #22.

rade:
See also #437.

unitymind:
In the same manner `weave run` or `weave expose` can execute on another peer host, using already allocated on another host CIDR.

rade:
This is about preventing weave subnets from clashing with host subnets. It's not about IP addresses. IPAM (#22) takes care of ensuring uniqueness of IPs within the weave network.

unitymind:
Okey, thanks for clarification

PR #1053, opened 2015-07-01
Check addresses supplied by users for conflict with existing routes

### LINKED ISSUE #437: [dns] subnet collision hard to detect
A user [accidentally launched weavedns with a CIDR that is in use](https://botbot.me/freenode/weavenetwork/2015-03-06/?msg=33505887); in fact the host's nameserver lives on that sub-net. There is no indication of anything being amiss in the weavedns logs - just failed lookups for external domains.

We could detect this, i.e. check whether the entry in resolv.conf clashes with our weave CIDR.

Admittedly this is a very special case. There are plenty of other opportunities for subnet collisions. However, unlike the others, this one is easy to detect.

bboreham:
Fixes #437, #480.

Issues a warning but carries on.

Note that we are still not checking ranges passed to IPAM; I'm still considering whether that is a good idea.

### DIFF
--- a/net/route.go
+++ b/net/route.go
@@ -16,7 +16,7 @@ func CheckNetworkFree(subnet *net.IPNet) error {
 	}
 	for _, route := range routes {
 		if route.Dst != nil && overlaps(route.Dst, subnet) {
-			return fmt.Errorf("Network %s would overlap with existing route %s", subnet, route.Dst)
+			return fmt.Errorf("Network %s overlaps with existing route %s", subnet, route.Dst)
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


net/route.go
```diff
@@ -26,3 +26,18 @@ func CheckNetworkFree(subnet *net.IPNet) error {
...
+// For a specific address, we only care if it is actually *inside* an
+// existing route, because weave-local traffic never hits IP routing.
+func CheckAddressOverlap(addr net.IP) error {
+	routes, err := netlink.RouteList(nil, netlink.FAMILY_V4)
+	if err != nil {
+		return err
+	}
+	for _, route := range routes {
+		if route.Dst != nil && route.Dst.Contains(addr) {
+			return fmt.Errorf("Address %s would overlap with existing route %s", addr, route.Dst)
```
rade: s/would overlap/overlaps

...since we don't treat this as an error, and carry on.

rade:
Where does the warning pop up when using the proxy? In the CLI? The proxy logs?

bboreham:
> Where does the warning pop up when using the proxy? In the CLI? The proxy logs?

Nowhere, it seems.

bboreham:
#1056 created for proxy issue

PR #1057, opened 2015-07-01
Give the user more choice of log levels

bboreham:
Addresses #329.

Note that some log levels do not sit well with some combinations of usage, e.g. the parts of the `weave` script that scan logfiles looking for specific messages are broken if your log-level inhibits those messages.  But it offers flexibility to advanced users.

### DIFF
--- a/common/logging.go
+++ b/common/logging.go
@@ -41,22 +41,28 @@ var (
 	Log *logrus.Logger
 )
 
-func InitLogging(level logrus.Level) {
-	if Log == nil {
-		Log = &logrus.Logger{
-			Out:       os.Stderr,
-			Formatter: standardTextFormatter,
-			Hooks:     make(logrus.LevelHooks),
-			Level:     level,
-		}
+func init() {
+	Log = &logrus.Logger{
+		Out:       os.Stderr,
+		Formatter: standardTextFormatter,
+		Hooks:     make(logrus.LevelHooks),
+	}
+}
+
+func SetLogLevel(levelname string) {
+	level, err := logrus.ParseLevel(levelname)
+	if err != nil {
+		Log.Fatal(err)
 	}
 	Log.Level = level
 }
 
-func InitDefaultLogging(debug bool) {
-	level := logrus.InfoLevel
-	if debug {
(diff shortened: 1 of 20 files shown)


rade:
- I think we should ditch the `-debug` flag and instead have a sensible default for `-loglevel`
- We should document the new flag in the troubleshooting docs.

site/troubleshooting.md
```diff
@@ -20,7 +20,7 @@ Check the weave container logs with
 
 A reasonable amount of information, and all errors, get logged there.
 
-The log verbosity can be increased by supplying the `-debug` flag when
+The log verbosity can be increased by supplying the flag `-log-level=debug` when
```
rade: `--log-level=debug`, surely. So that we don't have to change this *again* when moving to unix-getopt style option parsing.

common/logging.go
```diff
@@ -60,3 +60,20 @@ func InitDefaultLogging(debug bool) {
...
+	return nil
+}
+
+// Combination of InitDefaultLogging and SetLogLevel, for convenience
+// of existing programs that support both options
+func InitDefaultLoggingOrDie(loglevel string) {
+	if err := SetLogLevel(loglevel); err != nil {
+		Log.Fatal(err)
+	}
+}
```
rade: I am perplexed by this variety of log initialisers. Why can't we have a *single function*, `InitLogging(logLevel string)`, that does the job of `initDefaultLoggingOrDie`?

site/troubleshooting.md
bboreham: to clarify: you are suggesting to document it as `--log-level`, even though the code and the usage message will show `-log-level` ?  It seems to be impossible to have single-minus and double-minus options in the same program, without much bigger changes.

common/logging.go
bboreham: It was to minimize impact on unrelated code.  But I think four init functions is taking this too far; I will clean it up.

rade:
I've addresses a whole bunch of cosmetic issues, removed some stray debug logging from master, and rebased.

PR #1058, opened 2015-07-02
Register the "dns.weave.local" name in WeaveDNS

### LINKED ISSUE #984: Register weaveDNS resolver address in weaveDNS
nginx can reverse proxy/load balance across servers by domain name as long as a resolver is specified explicitly in the configuration (system resolver cannot be used, presumably as it does not have a non-blocking API). At the moment this would require determining the IP address from `/etc/resolv.conf` and injecting it into the configuration, which is a little awkward. Fortunately the resolver address can itself be specified by hostname, and this is looked up (with the system resolver) when nginx starts. Consequently if each weaveDNS peer were to publish a well known name (e.g. `weavedns.weave.local`) pointing to the correct resolver IP, this name could be used in the nginx configuration and the system would bootstrap itself automatically with no need to preprocess configuration files.

inercia:
Fixes #984 by registering `dns.weave.local.` in a private ident that is not exported via mDNS, so containers can query about `dns.weave.local.` and get the IP of the local WeaveDNS instance.

### DIFF
--- a/test/200_dns_test.sh
+++ b/test/200_dns_test.sh
@@ -5,15 +5,17 @@
 C1=10.2.0.78
 C2=10.2.0.34
 NAME=seetwo.weave.local
+DNS1=10.2.254.1
 
 start_suite "Resolve names on a single host"
 
-launch_dns_on $HOST1 10.2.254.1/24
+launch_dns_on $HOST1 $DNS1/24
 
 start_container          $HOST1 $C2/24 --name=c2 -h $NAME
 start_container_with_dns $HOST1 $C1/24 --name=c1
 
 assert_dns_record $HOST1 c1 $NAME $C2
+assert_dns_record $HOST1 c1 "dns.weave.local" $DNS1
 
 assert_raises "exec_on $HOST1 c1 dig MX $NAME | grep -q 'status: NXDOMAIN'"
 
--- a/weave
+++ b/weave
@@ -1039,6 +1039,8 @@ launch_dns() {
     ipam_cidrs_or_die $DNS_CONTAINER $CIDR_ARGS
     with_container_netns_or_die $DNS_CONTAINER attach $ALL_CIDRS
     populate_dns
+    # register "dns.weave.local." privately (not exported via mDNS)
+    put_dns_fqdn weave:private dns.$(dns_domain) $ALL_CIDRS
 }
 
 launch_proxy() {
--- a/nameserver/zone.go
+++ b/nameserver/zone.go
@@ -21,7 +21,8 @@ const (
 	DefaultRefreshInterval = int(DefaultLocalTTL) // Period for background updates with mDNS
 	DefaultRelevantTime    = 60                   // When to forget info about remote info if nobody asks...
 
-	defaultRemoteIdent = "weave:remote" // Ident used for info obtained from mDNS
+	defaultRemoteIdent  = "weave:remote"  // Ident used for info obtained from mDNS
+	defaultPrivateIdent = "weave:private" // Ident used for local host data (we never export)
 )
 
 // +1 to also exclude a dot
--- a/nameserver/zone_lookup.go
+++ b/nameserver/zone_lookup.go
@@ -39,11 +39,15 @@ func (uzr *uniqZoneRecords) toSlice() []ZoneRecord {
 
 //////////////////////////////////////////////////////////////////////////////
 
+func (zone *ZoneDb) isExportableIdent(ident string) bool {
+	return (ident != defaultRemoteIdent) && (ident != defaultPrivateIdent)
+}
+
 // Lookup in the database for locally-introduced information
 func (zone *ZoneDb) lookup(target string, lfun func(ns *nameSet) []*recordEntry) (res []ZoneRecord, err error) {
 	uniq := newUniqZoneRecords()
 	for identName, nameset := range zone.idents {
-		if identName != defaultRemoteIdent {
+		if zone.isExportableIdent(identName) {
 			for _, ze := range lfun(nameset) {
 				uniq.add(ze)
 			}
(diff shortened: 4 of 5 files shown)


test/200_dns_test.sh
```diff
@@ -14,6 +14,7 @@ start_container          $HOST1 $C2/24 --name=c2 -h $NAME
 start_container_with_dns $HOST1 $C1/24 --name=c1
 
 assert_dns_record $HOST1 c1 $NAME $C2
+assert_dns_record $HOST1 c1 "dns.weave.local" 10.2.254.1
```
awh: Could we make this an variable to match the multicast tests

awh:
Am ready to merge this once minor comment is addressed!

weave
```diff
@@ -1027,6 +1027,8 @@ launch_dns() {
     ipam_cidrs_or_die $DNS_CONTAINER $CIDR_ARGS
     with_container_netns_or_die $DNS_CONTAINER attach $ALL_CIDRS
     populate_dns
+    # register "dns.weave.local." privately (not exported via mDNS)
+    put_dns_fqdn 'weave:private' dns.`dns_domain` $ALL_CIDRS
```
rade: - single quotes aren't necessary here; I've [just](https://github.com/weaveworks/weave/commit/5559f1df2e4d50b4f08d600a633e57ced5490b6b) removed them elsewhere on master.
- never use backticks; always `$(...)`.

test/200_dns_test.sh
```diff
@@ -5,15 +5,17 @@
...
 start_suite "Resolve names on a single host"
 
-launch_dns_on $HOST1 10.2.254.1/24
+launch_dns_on $HOST1 $DNS1/24
 
 start_container          $HOST1 $C2/24 --name=c2 -h $NAME
 start_container_with_dns $HOST1 $C1/24 --name=c1
 
 assert_dns_record $HOST1 c1 $NAME $C2
+assert_dns_record $HOST1 c1 "dns.weave.local" DNS1
```
awh: This test is failing - should be `$DNS1`

squaremo:
Is there a reason to pollute the user-owned namespace `weave.local` with this name? Why not use something else, e.g., `.local` (I know, the user is allowed to change the name of the domain, but that does not excuse the pollution).

Also: unless WeaveDNS is in the same subnet as the container asking, they will not be able to talk. This seems like a bit of a sticking point for the issue which motivated this change.

rade:
> Also: unless WeaveDNS is in the same subnet as the container asking, they will not be able to talk. This seems like a bit of a sticking point for the issue which motivated this change.

Why not populate it with the docker bridge ip? As per https://github.com/weaveworks/weave/issues/984#issuecomment-114425450

squaremo:
> Why not populate it with the docker bridge ip? As per #984 (comment)

And if you're doing that, why not put an entry in `/etc/hosts` while you're at it (e.g., with `--add-host` or `--link`).

rade:
@fintanr see above - could you give that a try?

rade:
i.e. `docker run --add-host weavedns:$(weave docker-bridge-ip) ...` should do the trick and allow the nginx to resolve `weavedns`.

fintanr:
@rade sure, starting a test/update for nginx guides now.

squaremo:
> `docker run --add-host weavedns:$(weave docker-bridge-ip) ...`

or even

    docker run --link weavedns ...

HOWEVER: linking also copies the environment variables from the source container over to the target container (!?). So perhaps not recommended.

inercia:
What is the problem with `polluting` the `weave.local.` domain? We are not introducing a random name. We are setting the dns server of that domain. It is a common convention to have a `dns.somedomain.com` server for the `somedomain.com` domain. Registering `dns` in a different domain provides little benefit and much complexity.

squaremo:
> What is the problem with polluting the weave.local. domain?

It's meant to be under control of the user -- withholding a name in it is a bad smell.

Sure, it may be a convention for DNS admins to put dns.example.com -- but it's not done automatically by the nameserver (is it?). In any case I don't think we need this for fixing #984.

inercia:
@squaremo So you are suggesting to fix this with `--add-host`. Taking into account that `nginx` uses its own resolver, have we verified that it reads `/etc/hosts`?

So I guess I should close this PR.

squaremo:
> @squaremo So you are suggesting to fix this with --add-host. Taking into account that nginx uses its own resolver, have we verified that it reads /etc/hosts?

According to the description in #984, nginx uses the _system resolver_ first, to find the host you told it to use for _its own resolver_.

> So I guess I should close this PR.

Only if there's otherwise no use for this feature -- i.e., is there another scenario in which it would be handy to be able to resolve the DNS server itself?

inercia:
> Only if there's otherwise no use for this feature -- i.e., is there another scenario in which it would be handy to be able to resolve the DNS server itself?

I think there isn't... I'm closing it.

rade:
Let's wait for @fintanr to confirm that the `--add-host` approach actually works.

inercia:
Any news on this, @fintanr?

fintanr:
--add-host worked fine for me

kpdecker:
For anyone who stumbles on this via google: IPv6 must be turned off for the resolver to work, at least as of Nginx 1.9.15.
https://trac.nginx.org/nginx/ticket/974

TL;DR: If the resolver setting is not working, try `ipv6=off` resolver flag.

bboreham:
@kpdecker I believe you are describing the same problem as #2244, which is fixed now and will be in the next release.  Sorry for the trouble.

kpdecker:
@bboreham thanks for the update. I have a work around for now and will certainly give the next version a try!

bboreham:
@kpdecker we just released version 1.5.1 with the fix to #2244 included.

kpdecker:
@bboreham thanks, will make some time to test it out.

PR #1064, opened 2015-07-02
Wait forever for IPAM

### LINKED ISSUE #1023: `weave launch-dns` fails if IPAM allocation blocks
Observe:
````
$ ./weave launch -initpeercount 2
d9fd4b2eb9af2b0b5e5a5a22853b42fabcf4f2bee525df0c934123a073e3c0ec
$ time ./weave launch-dns
Container 9d01006caae41717fc5924df4edc725fd5573ab6abddf66e03b7b6bc172541e8 not running.

real	0m20.726s
user	0m0.014s
sys	0m0.005s
$ echo $?
1
````
This also affects the new launch everything behaviour (#962):
````
$ time ./weave launch -initpeercount 2
Container 6734b7dac75dc7c6800d0f212b81351315afd427958f37c0569bf575b34048ec not running.

real	0m22.109s
user	0m0.008s
sys	0m0.003s
$ echo $?
1
$ docker ps
CONTAINER ID        IMAGE                     COMMAND                CREATED             STATUS              PORTS                                            NAMES
09699147a938        weaveworks/weave:latest   "/home/weave/weaver    3 minutes ago       Up 3 minutes        0.0.0.0:6783->6783/tcp, 0.0.0.0:6783->6783/udp   weave 
````
These semantics aren't particularly helpful - the weaveDNS launch part should either block indefinitely, or succeed immediately and wait for the allocation to complete in the background.

### PR DESCRIPTION
**Pull request by awh**
Fixes #1023.

Is this an acceptable approach?

### DIFF
--- a/net/if.go
+++ b/net/if.go
@@ -2,6 +2,7 @@ package net
 
 import (
 	"fmt"
+	"math"
 	"net"
 	"time"
 )
@@ -10,11 +11,17 @@ const (
 	sleepTime = 100 * time.Millisecond
 )
 
+// Wait `wait` seconds for an interface to come up. Pass zero to check once
+// and return immediately, or a negative value to wait indefinitely.
 func EnsureInterface(ifaceName string, wait int) (iface *net.Interface, err error) {
 	if iface, err = findInterface(ifaceName); err == nil || wait == 0 {
 		return
 	}
-	for i := time.Duration(wait) * time.Second / sleepTime; err != nil && i > 0; i-- {
+	i := int64(math.MaxInt64)
+	if wait > 0 {
+		i = int64(time.Duration(wait) * time.Second / sleepTime)
+	}
+	for ; err != nil && i > 0; i-- {
 		time.Sleep(sleepTime)
 		iface, err = findInterface(ifaceName)
 	}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by rade** [comment]
So -1 means 'forever'? The usage should say so.

**Comment by rade** [comment]
Would you mind summarising what now waits for what and for how long?

Also, please double-check that there aren't any cases where we leave just-started containers running in the event of a failure in network configuration; the timeout provided a fail-safe mechanism, which we are now losing.

**Comment by awh** [comment]
> what now waits

Everything

> for how long

Forever :)

Router and weaveDNS launches wait forever for their interfaces, as do application containers launched via the proxy. In all cases these use `with_container_netns_or_die`, so if anything goes terminally wrong with attachment the container is killed.

File: net/if.go [hunk 1]
```diff
@@ -10,13 +10,22 @@ const (
...
+	if wait < 0 {
+		for err != nil {
+			time.Sleep(sleepTime)
+			iface, err = findInterface(ifaceName)
+		}
+	} else {
+		for i := time.Duration(wait) * time.Second / sleepTime; err != nil && i > 0; i-- {
+			time.Sleep(sleepTime)
+			iface, err = findInterface(ifaceName)
+		}
```
rade: refactor this so there is only one loop.

awh:
Use a single loop in `EnsureInterface`

PR #1065, opened 2015-07-02
Gossip-based DNS embedded in the router.

### LINKED ISSUE #338: [dns] return all matching local records

### LINKED ISSUE #583: [dns] remember and return *all* answers to a broadcast query
Per https://github.com/weaveworks/weave/issues/225#issuecomment-94563050, when WeaveDNS is asked about a local name it will

1) send a mDNS query and wait for up to 500ms for an answer
2) get the first answer
3) store that answer in the cache and
4) return that answer to the client

Following answers to the mDNS query will be simply ignored.

We should keep *all* answers instead, regardless of when they arrive, and return them in response to ordinary (non-mdns) queries.

### LINKED ISSUE #645: [dns] list *all* records
As of #465, `weave status` shows DNS records. But only records held in the zone db of the weaveDNS instance on the local host. Some users have requested a means by which to obtain a 'global' view.

### LINKED ISSUE #733: [dns] basic fault resilience for services
Follow on from #226. Extend load balancing behaviour to return all addresses for a name in a randomised order, allowing sophisticated clients to connect to multiple service instances for fault resilience.

### LINKED ISSUE #826: Replace mDNS peering with gossip
WeaveDNS peers currently use mDNS to communicate registrations amongst themselves in response to queries. This has a number of disadvantages:

*  It is not possible to tell when we have received the set of all available answers, leading to the use of a heuristic timeout
* Reverse lookups are similarly hard to implement, as we must rely on the absence of answers within a defined time period to know when to fallback to external DNS. This introduces unnecessary delay (#473), and packet loss can result in false negatives
* Name updates (least common operation) require no network operations, whereas lookups (most common operation) do

These issues can be mitigated to a degree by adopting additional mDNS features intended to promote performance and scalability, however:

* Doing so would introduce much additional implementation complexity to address problems which have already been effectively solved by the router gossip protocol
* The features in question were designed on the assumption that multicast traffic is free, which is not true in weave networks. Extending them further to work around this results in even more complexity

As a consequence, we intend to replace mDNS with the router gossip mechanism so that updates are proactively propagated to all peers, resulting in the following advantages:

* Each peer operates on the assumption that its local view (maintained by gossip) of the distributed name mapping is authoritative. Forward and reverse lookups are answered immediately from this mapping with no timeouts or network operations; updates are propagated asynchronously so registration is effectively a local operation from the point of view of clients updating the mapping
* The caching and cache invalidation mechanism can be removed, resulting in a smaller codebase
* Reuse of the gossip protocol obviates the need for additional implementation work on mDNS; existing mDNS code can be removed

Depends on #741 for access to the router gossip mechanism.

### LINKED ISSUE #833: Refactor nameserver codebase into separate files and subpackages
The current implementation uses a handful of files in a single package, despite the large number of defined types and operations:

````
addrs.go:type ZoneRecord interface {
addrs.go:type ZoneLookup interface {
addrs.go:type ZoneLookupFunc func(target string) ([]ZoneRecord, error)
addrs.go:type ZoneObservable interface {
addrs.go:type ZoneObserverFunc func(string, ZoneRecordObserver) error
addrs.go:type ZoneRecordObserver func()
addrs.go:type Record struct {
addrs.go:type IPv4 [4]byte
cache.go:type entryStatus uint8
cache.go:type cacheEntry struct {
cache.go:type entriesPtrsHeap []*cacheEntry
cache.go:type ZoneCache interface {
cache.go:type cacheKey dns.Question
cache.go:type entries map[cacheKey]*cacheEntry
cache.go:type Cache struct {
dns.go:type DNSResponseBuilder func(r *dns.Msg, q *dns.Question, addrs []ZoneRecord, ttl int) *dns.Msg
mdns_client.go:type Response struct {
mdns_client.go:type responseInfo struct {
mdns_client.go:type inflightQuery struct {
mdns_client.go:type MDNSClient struct {
mdns_client.go:type mDNSQueryInfo struct {
mdns_client.go:type MDNSAction func()
mdns.go:type ZoneMDNS interface {
mdns.go:type ZoneMDNSServer interface {
mdns.go:type ZoneMDNSClient interface {
mdns_server.go:type MDNSServer struct {
mdns_server.go:type LookupFunc func(ZoneLookup, *dns.Msg, *dns.Question) *dns.Msg
server.go:type DNSServerConfig struct {
server.go:type dnsProtocol uint8
server.go:type DNSServer struct {
zone.go:type LookupError string
zone.go:type DuplicateError struct {
zone.go:type Zone interface {
zone.go:type recordEntry struct {
zone.go:type name struct {
zone.go:type namesSet struct {
zone.go:type identRecordsSet map[string]*namesSet
zone.go:type refreshRequest struct {
zone.go:type ZoneConfig struct {
zone.go:type ZoneDb struct {
zone_lookup.go:type uniqZoneRecordKey struct {
zone_lookup.go:type uniqZoneRecords map[uniqZoneRecordKey]ZoneRecord
````

to aid discoverabilty and navigability of the codebase these types should be split out into separate files, grouped into subpackages where appropriate. Depends on #741,#826 as these are likely to prune a lot of code which shouldn't be needlessly reorganised.

### LINKED ISSUE #944: sched_queue_test.go fails with wrong number of calls
`sched_queue_test.go` fails sometimes with _"expected X and got X+/-1"_ errors.

```
--- FAIL: TestSchedCallsGap-4 (0.01s)
	sched_queue_test.go:62: Now: 1970-01-01 00:16:40 +0000 UTC - calls: 900
	util.go:134: Expected Number of calls 901 but got 900
		goroutine 33 [running]:
		github.com/weaveworks/weave/testing.stackTrace(0xc20809e000, 0x0, 0x0)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:143 +0x88
		github.com/weaveworks/weave/testing.StackTrace(0x0, 0x0)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:138 +0x35
		github.com/weaveworks/weave/testing.Fatalf(0xc208090090, 0x583170, 0x19, 0xc20806de08, 0x3, 0x3)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:134 +0x27
		github.com/weaveworks/weave/testing.AssertEqualInt(0xc208090090, 0x384, 0x385, 0x56c190, 0xf)
			/home/travis/gopath/src/github.com/weaveworks/weave/testing/util.go:51 +0x1b3
		github.com/weaveworks/weave/common.TestSchedCallsGap(0xc208090090)
			/home/travis/gopath/src/github.com/weaveworks/weave/common/sched_queue_test.go:63 +0x55c
		testing.tRunner(0xc208090090, 0x6201b8)
			/usr/local/go/src/testing/testing.go:447 +0xbf
		created by testing.RunTests
			/usr/local/go/src/testing/testing.go:555 +0xa8b
INFO: 2015/06/16 18:58:06.106500 TestSchedCallsStop starting
FAIL
coverage: 73.3% of statements
FAIL	github.com/weaveworks/weave/common	0.030s
ok  	github.com/weaveworks/weave/ipam	7.379s	coverage: 67.5% of statements
ok  	github.com/weaveworks/weave/ipam/paxos	3.669s	coverage: 96.4% of statements
ok  	github.com/weaveworks/weave/ipam/ring	1.718s	coverage: 90.4% of statements
ok  	github.com/weaveworks/weave/ipam/space	0.003s	coverage: 71.6% of statements
ok  	github.com/weaveworks/weave/nameserver	18.372s	coverage: 84.3% of statements
ok  	github.com/weaveworks/weave/net	0.003s	coverage: 5.3% of statements
ok  	github.com/weaveworks/weave/router	0.378s	coverage: 28.9% of statements
make: *** [tests] Error 1
```

### LINKED ISSUE #987: resolve unqualified names in weavedns domain

### PR DESCRIPTION
**Pull request by tomwilkie**
- Nameserver holds a sorted list of (hostname, peer, container id, ip) tuples.
- Entries are gossiped using topology gossip / broadcast - only new entries are
  propogated as expected.
- Entries for a given peer are removed when the connection to the peer is lost.
- Entries are tombstoned when containers die.  Tombstones are kept for 10mins,
  as they do not need to survive partition, only gossip propagation delay.
- Hostname lookups are O(logn) in memory.
- All other operations (reverse DNS, tombstoning, deletion, etc) are O(n) in memory.

Fixes #826, Fixes #987, Fixes #733, Fixes #944, Fixes #833, Fixes #645, Fixes #583, Fixes #338

Left to do:
- [x] Make ```weave launch-dns``` and ```weave stop-dns``` print warning and exit cleanly
- [x] Add useful status output
- [x] Audit logging
- [x] Update proxy to send dns commands to the router.
- [x] Make sure restart still works & add tests for that
- [x] make ttl configurable, audit args, arg naming
- [x] Add tests for large responses
- [x] Make ``when_weave_running`` change
- [x] Docs
- [x] Deal with OnGC-gossip race
- [x] Proactively broadcast tombstones
- [x] Add EDNS and truncation
- [x] Manually test mixed weave 1.0 / weave 1.1

After the bulk of the review, we need to:
- [x] squash & rebase
- [x] remove old weavedns code.
- [x] rename package to nameserver
- [x] update for mflags

Not doing in this PR (left to follow-on issues, to be filed)
- Reduce the complexity of non-hostname-lookups (all of which are O(n))
- Rate limit new entry broadcasts
- Remove dns entries from weave status and add a separate command for them

### CODE REVIEW

**Comment by tomwilkie** [comment]
@rade made a PR as I don't want you comments on changesets to get lost as I push -f.  Not ready to merge yet, but it just passed all the tests - https://circleci.com/gh/weaveworks/weave/1438

**Comment by tomwilkie** [comment]
From @rade: many `weave ...` commands actually work w/o weave running *as long as ipam isn't needed*. So I wouldn't just drop the `when_dns_running`. Replace it with `when_weave_running` instead.

**Comment by rade** [comment]
>  Not ready to merge yet

-> assigning to you in that case

**Comment by rade** [comment]
btw, I'd prefer if you didn't rebase unless necessary, and especially didn't squash commits - until the very end. Otherwise I'll get bored very quickly following this, since I'd have to re-read the entire diff over and over.

**Comment by tomwilkie** [comment]
Fair enough, I'll stop doing it until the end.

**Comment by rade** [comment]
> Hostname lookups are O(logn), reverse DNS lookups are O(n), all in memory.

What is the complexity of:
- removing an individual record (e.g. as a result of `weave detach`)
- removing all records for a container (e.g. as a result of container death)
- removing all records for a peer (e.g. as a result of the peer disappearing)
?

**Comment by tomwilkie** [comment]
Everything other than lookup-by-hostname is O(n), in memory.

There is a potential optimisation for receive gossip msg, to make merge ~O(logn).  This is on the assumption most gossip message are only a few entries.  For large ones O(n) is as good as you can get.

**Comment by rade** [comment]
We should file a follow-on PR to reduce the complexity.

**Comment by tomwilkie** [comment]
This PR also doesn't include rate limiting entry addition, which was part of the original design thoughts.  Happy to leave that to another PR?

**Comment by rade** [comment]
> rate limiting entry addition

File an issue for it.

**Comment by inercia** [comment]
I think this code does not reuse any of our current WeaveDNS code. What is the plan? Are we going to throw all the code we have now?

**Comment by rade** [comment]
I think there is a race here: I believe it is possible to receive some DNS gossip pertaining to a peer *after* that peer has been removed from `Peers` by GC. That would result in permanently stale records.

**Comment by tomwilkie** [comment]
@inercia thats right, its a fresh implementation.  Some of the code has been borrowed (mainly the dns.go bits) but they've been rewritten.  The API should be the same.

There is no plan, and what to do with the existing code is up to you and @rade.  I guess they could coexist for a while if there is a good reason, but in the end I don't see why we would need both.

**Comment by inercia** [comment]
@tomwilkie I don't see the point in throwing all the code we currently have for DNS, specially when most of the code in our current nameserver is _transport_ agnostic. The gossiping mechanism could be plugged in our current zone database, remove the mDNS code (and maybe the cache), and we could reuse the rest of the weaveDNS code... It is well tested code, and many features and solutions for corner cases are already there (for example, the code in this `dns.go` is far from implementing all the features we current have in our nameserver).

**Comment by tomwilkie** [comment]
@inercia what kind of features are missing from gossipdns/dns.go?

**Comment by inercia** [comment]
@tomwilkie Some things...
* fallback connections should use the same transport used by clients (eg, if the client uses TCP, we should use TCP to fallback servers)
* you should also handle EDNS buffer sizes
* you could also prune responses (there is no reason for returning 1000 answers, even when we have them)
* this server is not tested, and maybe it is not easy to test (eg, things like `etcResolvConf` should be configurable)
* you could also optionally cache responses...

**Comment by tomwilkie** [comment]
- fallback connections should use the same transport used by clients (eg, if the client uses TCP, we should use TCP to fallback servers)

The intention is that they do.  Have you spotted a bug?

- you should also handle EDNS buffer sizes

Whats EDNS?

- you could also prune responses (there is no reason for returning 1000 answers, even when we have them)

I intentionally don't want to do that.

 - this server is not tested, and maybe it is not easy to test (eg, things like etcResolvConf should be configurable)

dns.go is just a shim onto nameserver.go, which is well tested.  Its is also tested by the integration tests.  I'm not convinced dns.go and http,go need tests.

- you could also optionally cache responses...

All authoritative answers are served from memory by design, so no point in caching.  All recursive answers are intentionally uncached - they would be doing a network call anyway, so why cache them?

Thanks for taking a look at the code - this is really useful feedback.

**Comment by tomwilkie** [comment]
> I think there is a race here: I believe it is possible to receive some DNS gossip pertaining to a peer after that peer has been removed from Peers by GC. That would result in permanently stale records.

@rade under what situation are you thinking?  I'm not familiar enough with the gossip / broadcast / peers code.  I could add a list of "live" peers to the nameserver, and filter incoming gossips by that.  I would need to add a "peer added" callback to peers.go if one doesn't already exist.  Then we'd need to make sure we get the peer-added callback before any gossips from the peer.  Thoughts?

**Comment by rade** [comment]
> under what situation are you thinking?

There is no connection between the peer topology gossip and nameserver gossip. Especially not when multiple peers are involved. e.g. Peer A could detect B's death and still receive nameserver gossip from B containing records pertaining to C.

So I'd think it can happen quite easily.

> we'd need to make sure we get the peer-added callback before any gossips from the peer.

That's just the same problem, in reverse. But it's less of an issue as long as it only goes wrong rarely - period gossip will fill in the blanks.

Also, instead of a callback you could just ask `peers` whenever adding records.

**Comment by tomwilkie** [comment]
@rade I've added code to filter gossip messages based on the contents of peers.  Had to be careful to avoid deadlocks, which meant I had to take the peers lock in a bit of a smelly way.  Let me know what you think.

Other than EDNS (what is it?) I think this might be close to done.  Who do you want to review this?

File: gossipdns/nameserver.go [hunk 1]
```diff
@@ -0,0 +1,234 @@
...
+		return nil, nil, err
+	}
+
+	// Filter out entries which we think have died.  Have to be
+	// careful to avoid a deadlock - other threads can take the
+	// peers write lock, then call out OnGC callback, taking our
+	// write lock.
+	if n.peers != nil {
+		n.peers.RLock()
+		defer n.peers.RUnlock()
```
rade: I don't understand why you need this lock. `peers.Fetch` already takes a lock, and we are not holding any other locks at this point.

In fact as it stands the code is actually *more prone* to deadlocks. If we take a read lock, then something else attempts to take a write lock, then the call to `Fetch` below will deadlock, since locks are re-entrant and no read locks are granted when there is a pending write lock.

tomwilkie: Yes you're right.   The lock is there to stop inconsistent results (not a deadlock, sorry) - when a peer is removed by another thread halfway through the filter.

I could add a FetchUnlocked method to solve this?  The filter-and-apply-gossip needs to be atomic wrt removal of peers.  I originally did the filter inside the n.Lock below, but that could deadlock.

rade: `FetchUnlocked` - gross. Could the `OnGC` callback take place outside the lock? If it can then it definitely should; invoking callbacks while holding locks is bad.

tomwilkie: It might take some jiggling but I'm sure I could do that.

On Mon, Jul 6, 2015 at 1:56 PM, Matthias Radestock <notifications@github.com

router/peers.go
```diff
@@ -43,7 +43,21 @@ func NewPeers(ourself *LocalPeer, onGC func(*Peer)) *Peers {
...
+		onGC:    []func(*Peer){onGC}}
+}
+
+func (peers *Peers) OnGC(callback func(*Peer)) {
+	peers.Lock()
+	defer peers.Unlock()
+	peers.onGC = append(peers.onGC, callback)
+}
+
+func (peers *Peers) triggerOnGC(removed []*Peer) {
```
rade: `invokeOnGCCallbacks`?

rade:
How sure are your that moving the `onGC` callback invocation outside the lock is safe? I am concerned about race conditions where another thread might re-add a peer before the callback has been invoked. I reckon this is ok for the existing `onGC` callback, the `MacCache` cleanup, since it's a cache so worst case an incorrect removal results in a performance hit. But is it ok for the new one?

rade:
So [apparently](https://github.com/weaveworks/weave/issues/1091#issuecomment-118972128) this is also meant to fix #733. I'd rather see that done separately. Otherwise we need extra tests that will just distract from the main purpose of this PR.

tomwilkie:
I can do it separately, but it would be adding code (to pick the first
entry) which would then just be removed.

On Mon, Jul 6, 2015 at 8:38 PM, Matthias Radestock <notifications@github.com

rade:
If you keep it you must add some smoke tests. And it will make it harder to review this PR.

ipam/address/address.go
```diff
@@ -98,3 +98,10 @@ func Min(a, b Offset) Offset {
 	}
 	return a
 }
+
+func (addr Address) Reverse() Address {
+	return ((addr >> 24) & 0xff) | // move byte 3 to byte 0
+		((addr << 8) & 0xff0000) | // move byte 1 to byte 2
+		((addr >> 8) & 0xff00) | // move byte 2 to byte 1
+		((addr << 24) & 0xff000000) // byte 0 to byte 3
+}
```
rade: It's odd for `nameserver` to import something from IPAM. Let's move `address`, including the above function, into `net`. In a separate PR.

gossipdns/entry_test.go
```diff
@@ -0,0 +1,126 @@
...
+	now = func() int64 { return 1234 }
+
+	entries := Entries{}
+	entries.add("A", "", router.UnknownPeerName, address.Address(0))
+	expected := Entries{
+		Entry{Hostname: "A", Origin: router.UnknownPeerName, Addr: address.Address(0)},
+	}
+	if !reflect.DeepEqual(entries, expected) {
+		t.Fatalf("Unexpected: %v", entries)
+	}
```
rade: use testify's `require.Equal`.

site/weavedns.md
```diff
@@ -283,17 +246,16 @@ guide](troubleshooting.html#status-report) for more detail.
 
 The second section is pertinent to weaveDNS, and includes:
 
+* Port on which the DNS server is listening
 * The local domain suffix which is being served
-* The address on which the DNS server is listening
-* The interface being used for multicast DNS
-* The fallback DNS which will be used to resolve non local names
-* The names known to the local weaveDNS server. Each entry comprises
+* The respinse ttl
```
rade: typo

site/weavedns.md
```diff
@@ -66,47 +56,38 @@ root@ubuntu:/# ping pingme
...
+`.weave.local` domain then weaveDNS records the association of that name with
+the container's weave IP address(es) in its in-memory database, and broadcasts
+the association to other weave peers in the cluster.
 
-When weaveDNS is queried for a name in the `.weave.local` domain, it
-first checks its own records. If the name is not found there, it asks
-the weaveDNS servers on the other hosts in the weave network.
+When weaveDNS is queried for a name in the `.weave.local` domain, it looks
+up the hostname its in memory database and responds with the IPs all containers
+for that hostname across the entire cluster, in a random order.
```
rade: "with the IPs *of* all containers *with* that hostname"

rade:
The lines in the .md files are too long - should wrap at 72; This has resulted in some spurious changes.

gossipdns/http.go
```diff
@@ -0,0 +1,99 @@
...
+			return
+		}
+
+		if err := n.AddEntry(dns.Fqdn(hostname), container, n.ourName, ip); err != nil {
+			badRequest(w, fmt.Errorf("Unable to add entry: %s", err))
+			return
+		}
+
+		if r.FormValue("check-alive") == "true" && n.docker.IsContainerNotRunning(container) {
+			common.Log.Infof("[gpssipdns] '%s' is not running: removing", container)
```
rade: typo

site/weavedns.md
```diff
@@ -66,47 +56,38 @@ root@ubuntu:/# ping pingme
...
-specify an address in CIDR format manually. In this case you are
-responsible for ensuring that the IP addresses specified are uniquely
-allocated and not in use by any other container.
+`weave run` and `weave launch-proxy`.
 
-Finally, weaveDNS can be stopped independently with
+Finally, weaveDNS is stopped automatically when the router is stopped:
 
-    host1$ weave stop-dns
+    host1$ weave stop-router
```
rade: I'd drop this whole paragraph.

weave
```diff
@@ -1345,6 +1300,10 @@ case "$COMMAND" in
         PEER=$1
         call_weave DELETE /peer/$PEER
         ;;
+    launch-dns|stop-dns)
+        echo "$COMMAND is deprecated." >&2
+        exit 0
+        ;;
```
rade: Let's be smarter and more helpful here:

- for `launch-dns` echo "The 'launch-dns' command has been removed; DNS is launched as part of 'launch' and 'launch-router'.". And if weave is not running then exit 1.
- for `stop-dns` echo "The 'stop-dns command has been removed; DNS is stopped as part of 'stop' and 'stop-router'.". Always exit 0.

prog/weaver/main.go
```diff
@@ -70,6 +78,13 @@ func main() {
 	flag.StringVar(&ipsubnetCIDR, "ipsubnet", "", "subnet to allocate within by default, in CIDR notation")
 	flag.IntVar(&peerCount, "initpeercount", 0, "number of peers in network (for IP address allocation)")
 	flag.StringVar(&apiPath, "api", "unix:///var/run/docker.sock", "Path to Docker API socket")
+
+	flag.StringVar(&dnsDomain, "dns-domain", gossipdns.DefaultDomain, "local dns domain")
+	flag.IntVar(&dnsPort, "dns-port", 53, "port to listen to DNS requests")
+	flag.IntVar(&dnsTTL, "dns-ttl", 30, "TTL for DNS request")
+	flag.IntVar(&dnsUDPbufsize, "dns-udp-bufsize", 4096, "size of udp buffer for fallback dns requests")
+	flag.DurationVar(&dnsClientTimeout, "dns-client-timeout", 5*time.Second, "timeout for fallback dns requests")
```
rade: Those option names don't obviously correspond to their description. Can we do better?

rade:
I reckon `gossipdns` should be the new `nameserver` package.

ipam/address/address.go
bboreham: q

bboreham: Just pondering what package name `address` would end up with.  Code that imports both the standard library `net` and `weaveworks/weave/net` will rename one, probably the latter.  Might be less confusing to move `address` as a sub-package of `net`.

rade: > Might be less confusing to move `address` as a sub-package of `net`.

That is what I had in mind.

tomwilkie:
I'll rename the package once we've done the bulk of the review, if thats okay - it will make rebasing / merging easier.

So after the bulk of the review I will:
- squash & rebase
- remove old weavedns code.
- rename package to nameserver

Let me know when you think the right time to do that is.

tomwilkie:
I've also broken the ipam/address move in #1095

gossipdns/nameserver.go
```diff
@@ -0,0 +1,232 @@
...
+		if containerid != "*" && e.ContainerID != containerid {
+			return false
+		}
+
+		if ipStr != "*" && e.Addr != ip {
+			return false
+		}
+		return true
+	})
+}
```
rade: This should broadcast the change. Ditto in `ContainerDied`.

Makefile
```diff
@@ -55,7 +55,7 @@ $(WEAVER_EXE) $(WEAVEDNS_EXE) $(WEAVEPROXY_EXE) $(NETCHECK_EXE): common/*.go com
 		false; \
 	}
 
-$(WEAVER_EXE): router/*.go ipam/*.go ipam/*/*.go prog/weaver/main.go
+$(WEAVER_EXE): router/*.go ipam/*.go ipam/*/*.go gossipdns/*.go prog/weaver/main.go
```
inercia: Why is `gossipdns` and not a subdirectory in `router/`? It is part of the router now...

gossipdns/dns.go
```diff
@@ -0,0 +1,239 @@
...
+	s := &DNSServer{
+		ns:        ns,
+		domain:    domain,
+		ttl:       ttl,
+		port:      port,
+		tcpClient: &dns.Client{Net: "tcp", ReadTimeout: clientTimeout},
+		udpClient: &dns.Client{Net: "udp", ReadTimeout: clientTimeout, UDPSize: udpBuffSize},
+	}
+	var err error
+	if s.upstream, err = dns.ClientConfigFromFile(etcResolvConf); err != nil {
```
inercia: I think this should be configurable, otherwise it cannot be tested...

gossipdns/dns.go
```diff
@@ -0,0 +1,239 @@
...
+func (d *DNSServer) Stop() error {
+	for _, server := range d.servers {
+		if err := server.Shutdown(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func badDNSRequest(r *dns.Msg, w dns.ResponseWriter) {
```
inercia: I would move all the message serialization functions to a different file...

gossipdns/dns.go
```diff
@@ -0,0 +1,239 @@
...
+	Log.Debugf("[gossipdns] request: %+v", *req)
+	if len(req.Question) != 1 || req.Question[0].Qtype != dns.TypeA {
+		badDNSRequest(req, w)
+		return
+	}
+
+	hostname := req.Question[0].Name
+	addrs := d.ns.Lookup(hostname)
+
+	response := dns.Msg{}
```
inercia: I think all the message builders should be refactored and moved to a different file...

Makefile
tomwilkie: This is mirroring ipam, also part of the router but a top level directory.  I think its different enough to warrant this.

gossipdns/dns.go
tomwilkie: The integration tests cover this nicely.  I'm keen not to introduce unnecessary configurables.

gossipdns/dns.go
tomwilkie: Why?  This is nice and self contained as it stands, and the code reads in a very straight-line fashion.

gossipdns/dns.go
tomwilkie: Why?  There isn't that much common code between them, and inlining them like this makes the code more readable.

gossipdns/dns.go
inercia: Integration tests are not a substitute of good unit tests. Otherwise, we could just `rm -rf */*_test.go`... ;)

We currently test things like _"a truncated UDP response falls back to a TCP query"_. I would like to see things like that in the DNS unit tests...

gossipdns/dns.go
inercia: I think it would keep the code tidier by getting rid of these low level things. The DNS server should deal with higher level stuff, not with DNS messages fields...

Makefile
inercia: We have discussed in the past that we should move things to subdirectories, specially when they are not shared between different components... I think this would be a good candidate for such kind of change...

rade: > I think this would be a good candidate for such kind of change...

Perhaps. But not as part of this issue.

gossipdns/dns.go
tomwilkie: This file (dns.go) is about low-level interface specific DNS handling.  The high level code is in nameserver.go.  The pattern is repeated with http.go

tomwilkie:
Some comments got disappeared by the push -f.

Alvaro said: "Integration tests are not a substitute of good unit tests. Otherwise, we could just rm -rf */*_test.go... ;)

We currently test things like "a truncated UDP response falls back to a TCP query". I would like to see things like that in the DNS unit tests..."

tomwilkie:
I'm not suggesting we remove all unit tests.  I'm suggesting we have a mix of unit tests and integration tests.  The integration tests cover the code in dns.go nicely.

> We currently test things like "a truncated UDP response falls back to a TCP query". I would like to see things like that in the DNS unit tests..."

Why are you testing client behaviour?

inercia:
The truncated thing test is for the fallback functionality, not for the client. Please take a look at our current DNS server tests [here](https://github.com/weaveworks/weave/blob/master/nameserver/server_test.go).

So our unit tests check that we behave when relaying traffic to fallback servers, and our integration tests do not (currently) do that. In fact, I think it would be difficult to do that with integration tests... And I think that your `dns.go` tests don't check fallback servers functionality either...

nameserver/dns.go
```diff
@@ -1,117 +1,239 @@
...
+			return err
+		}
 	}
-	return makeReply(r, answers)
+	return nil
 }
 
-func makeDNSFailResponse(r *dns.Msg) *dns.Msg {
-	m := new(dns.Msg)
+func badDNSRequest(r *dns.Msg, w dns.ResponseWriter) {
```
inercia: This name leads to think that the request was malformed...

tomwilkie:
The integration tests check fallback behaviour in 290_dns_fallback_test.sh - what more do we need to test?  Specifically, I don't understand the whole truncated thing.

nameserver/dns.go
```diff
@@ -1,117 +1,239 @@
...
+	}
+
+	for i := range *answers {
+		j := rand.Intn(i + 1)
+		(*answers)[i], (*answers)[j] = (*answers)[j], (*answers)[i]
 	}
-	return answers
+}
+
+func (d *DNSServer) HandleHTTP(router *mux.Router) {
```
inercia: Why is this HTTP code here? Seems out of context...

tomwilkie: On the DNS server knows about the domain its serving; all the other code is domain-agnostic.  Hence, the only place for the handler is in the DNS server.

nameserver/dns.go
tomwilkie: I've renamed it errorResponse and added the error code as a parameter.

nameserver/dns.go
inercia: Yes, but at the cost of having a `HandleHTTP` inside the DNS server and dealing with HTTP there... It would be better to just pass the `domain` when you create the HTTP server.

tomwilkie: What cost?  The exact same code has to exist somewhere.  Right now I'd prefer to limit the scope of the configuration data to where its needed - I don't like passing lots of config around.

I'm not particularly fussed (its a small change).

inercia: You have several `HandleHTTP` (I think that in the `dns` and in the `nameserver`), and they are not there because they access private methods/data of their respective classes, and they do things like extracting parameters from forms, write HTTP responses, etc... I think that does not look very elegant, but of course that is my personal opinion... I think it would be much better to have all the HTTP code in `http.go` and invoke public methods in the right objects (for xample, maybe you could get the `Domain` from the DNS server in the HTTP `/status` handler).

tomwilkie: I've moved the code to http.go (see 4117109)

rade:
Presumably `common/sched_queue.go` can go.

nameserver/dns.go
```diff
@@ -1,117 +1,263 @@
...
 }
 
-// shuffleAnswers reorders answers for very basic load balancing
-func shuffleAnswers(answers []dns.RR) []dns.RR {
-	if len(answers) > 1 {
-		for i := range answers {
-			j := rand.Intn(i + 1)
-			answers[i], answers[j] = answers[j], answers[i]
+func (d *DNSServer) handleRecursive(client *dns.Client, defaultMaxResponseSize int) func(dns.ResponseWriter, *dns.Msg) {
+	return func(w dns.ResponseWriter, req *dns.Msg) {
```
inercia: This should also deal with the unqualified names case that #1050 solves.

nameserver/dns.go
```diff
@@ -1,117 +1,263 @@
...
+	tcpClient *dns.Client
+	udpClient *dns.Client
+}
+
+func NewDNSServer(ns *Nameserver, domain string, port int, ttl uint32,
+	clientTimeout time.Duration) (*DNSServer, error) {
+
+	s := &DNSServer{
+		ns:        ns,
+		domain:    domain,
```
inercia: Do we verify the domain is FQDN at any point?

nameserver/dns_test.go
```diff
@@ -1,39 +1,68 @@
...
-	questionMsg := new(dns.Msg)
-	questionMsg.SetQuestion("name", dns.TypeA)
-	questionMsg.RecursionDesired = true
-	question := &questionMsg.Question[0]
-	records := []ZoneRecord{
-		Record{"name", net.ParseIP("10.0.1.1"), 0, 0, 0},
-		Record{"name", net.ParseIP("10.0.1.2"), 0, 0, 0},
-		Record{"name", net.ParseIP("10.0.1.3"), 0, 0, 0},
-		Record{"name", net.ParseIP("10.0.1.4"), 0, 0, 0},
+func TestLargeUDPResonse(t *testing.T) {
```
inercia: `TestLargeUDPResonse` implies that this only tests UDP, but it also tests TCP.

inercia: And t should be `Response`, not `Resonse`

nameserver/dns_test.go
```diff
@@ -1,39 +1,68 @@
...
-	pruned = pruneAnswers(reply.Answer, 0)
-	require.Equal(t, len(records), len(pruned), "wrong number of answers")
+	// do a tcp query, ensure we don't get a truncated response
+	{
+		tcpClient := dns.Client{Net: "tcp"}
+		response := doRequest(&tcpClient, &dns.Msg{}, tcpPort)
+		require.False(t, response.MsgHdr.Truncated)
+		require.Equal(t, len(response.Answer), 100)
+	}
 }
```
inercia: We should also test malformed queries, not FQDN names, unsopported query types (eg, MX), etc

nameserver/dns.go
tomwilkie: It looks like there are a bunch of open questions related to #1050.  If #1050 gets merged before this change, I'll duplicate the approach here.  Otherwise I'll help adapt #1050 for the new code after this is merged.

nameserver/dns_test.go
tomwilkie: Renamed (in https://github.com/weaveworks/weave/commit/95e7ffc3d680a8852300a4a88038aa7d5dd3dcb3)

nameserver/dns.go
tomwilkie: Added in https://github.com/weaveworks/weave/commit/1bbcd7bbf2922d95b9761b7f2f5756444a068801

nameserver/entry_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/weaveworks/weave/net/address"
+	"github.com/weaveworks/weave/router"
+)
+
+func TestAdd(t *testing.T) {
+	oldNow := now
+	defer func() { now = oldNow }()
+	now = func() int64 { return 1234 }
```
inercia: This _time swap_ thing happens in several tests. Maybe it could it be refactored...

tomwilkie: I've done a similar thing here: https://github.com/weaveworks/scope/blob/master/probe/docker/registry_test.go#L118 but its pretty ugly.  Unfortunately code repetition is common in Go.

nameserver/dns_test.go
tomwilkie: We have tests for unsupported query types in the integration tests.

Re malformed queries, not FQDNs etc - I don't see the point of adding these tests.  (1) the parsing is not down to us, but meikg/dns and (2) we know there will fail.

tomwilkie:
@inercia @rade mind if I squash this all up again - are you guys up to date?

inercia:
It is ok for me squash commits...

inercia:
I see that coverage has been reduced from our current 83% to 56%...

tomwilkie:
Yes coverage has gone down; even if this code have 100% coverage, as it is so much shorter it would still go down, as the figure is dominated by the amount of uncovered code in the rest of the code base.

According to the covertool:
- nameserver.go 73.3%
- entry.go 85.2%
- dns.go 50.8%
- http.go 0%

The core (nameserver.go & entry.go) are well covered.  The interface glue (dns.go & http.go) are not, but don't need to be - they are pretty much just boilerplate, and are exercised by the integration tests.

inercia:
IMO integration tests cannot replace good unit tests: they can overlap, but they are intended to be used for different things. And we currently don't know what is really covered by integration tests because we do not have coverage reports for them...

There were many unit tests in the `nameserver` that have been removed for no reason and, as I already mentioned, some parts of this DNS server (like the recursive resolver) are completely untested (except for some trivial integration test). So this will never LGTM until code coverage is increased to something similar to what we currently have.

tomwilkie:
Other than testing, do you have anything other feedback?

nameserver/dns.go
```diff
@@ -1,117 +1,272 @@
...
 
-// shuffleAnswers reorders answers for very basic load balancing
-func shuffleAnswers(answers []dns.RR) []dns.RR {
-	if len(answers) > 1 {
-		for i := range answers {
-			j := rand.Intn(i + 1)
-			answers[i], answers[j] = answers[j], answers[i]
+func (d *DNSServer) handleRecursive(client *dns.Client, defaultMaxResponseSize int) func(dns.ResponseWriter, *dns.Msg) {
+	return func(w dns.ResponseWriter, req *dns.Msg) {
+		d.ns.debugf("recursive cdrequest: %+v", *req)
```
inercia: I'm not sure you can just send the same request we get from the client to the upstream server. This could lead to a conflict if two clients send a request with the same message ID. I would make a copy...

tomwilkie: This is what the old code did, therefore its not a regression.  It is
however worth making a bug for this,

nameserver/dns.go
```diff
@@ -1,117 +1,272 @@
...
+		return
+	}
+
+	for i := range *answers {
+		j := rand.Intn(i + 1)
+		(*answers)[i], (*answers)[j] = (*answers)[j], (*answers)[i]
+	}
+}
+
+func truncateResponse(response *dns.Msg, maxSize int) {
```
inercia: I think you don't need all of this. If a message is truncated, I think that the client is not going to try to read it (and I think we will send just a minimal response, with the TC bit set). So I think you could just do
```
if response.Len() > maxSize {
    response.Truncated = true
}
```

nameserver/dns.go
```diff
@@ -1,117 +1,272 @@
...
+func (d *DNSServer) handleRecursive(client *dns.Client, defaultMaxResponseSize int) func(dns.ResponseWriter, *dns.Msg) {
+	return func(w dns.ResponseWriter, req *dns.Msg) {
+		d.ns.debugf("recursive cdrequest: %+v", *req)
+		for _, server := range d.upstream.Servers {
+			response, _, err := client.Exchange(req, fmt.Sprintf("%s:%s", server, d.upstream.Port))
+			if err != nil {
+				d.ns.debugf("network error trying %s (%s)", server, err)
+				continue
+			}
+			if response != nil && response.Rcode != dns.RcodeSuccess {
```
inercia: I know this comes from the old code, but I think it is not really necessary. If we get a NXDomain error _and_ the response is authoritative, we wouldn't need to go to the next fallback server. In fact, I think we should continue with the iteration _iff_ the fallback server is unreachable...

nameserver/dns.go
tomwilkie: If the client sends us a request with the EDNS headers set we should reply with as many answers as we can fit in that buffer.

nameserver/dns.go
tomwilkie: Addresses in 3589939

nameserver/dns.go
inercia: But, when using UDP and no EDNS, we should not spend CPU&bandwidth on a message that will not be processed. So I think the `truncateResponse` should take that into account, and return the minimum packet depending on the protocol and capabilities of the client...

tomwilkie: The truncate algorithm is O(logn), so its not CPU expensive by any measure.  And its its a UDP response with no EDNS, that largest response will be 512 bytes, so its not expensive on the network either.

However, the cost of plumbing through all the special cases is something I'd rather avoid.

inercia: The cost of the UDP case is insignificant. Just pass the protocol to the `truncateResponse` and:
```
if proto == UDP && response.Len() > maxSize {
    response.Truncated = true
    return
}
```

prog/weaver/main.go
```diff
@@ -70,6 +77,11 @@ func main() {
 	mflag.StringVar(&ipsubnetCIDR, []string{"#ipsubnet", "#-ipsubnet", "-ipalloc-default-subnet"}, "", "subnet to allocate within by default, in CIDR notation")
 	mflag.IntVar(&peerCount, []string{"#initpeercount", "#-initpeercount", "-init-peer-count"}, 0, "number of peers in network (for IP address allocation)")
 	mflag.StringVar(&apiPath, []string{"#api", "-api"}, "unix:///var/run/docker.sock", "Path to Docker API socket")
+	mflag.StringVar(&dnsDomain, []string{"-dns-domain"}, nameserver.DefaultDomain, "local domain to server requests for")
+	mflag.IntVar(&dnsPort, []string{"-dns-port"}, 53, "port to listen on for DNS requests")
```
inercia: These values should not be hardcoded here: we should use constants (eg, `defaultDNSPort`)

prog/weaver/main.go
```diff
@@ -70,6 +77,11 @@ func main() {
...
+	mflag.IntVar(&dnsTTL, []string{"-dns-ttl"}, 30, "TTL for DNS request from our domain")
```
inercia: Ditto.

I think it does not make sense to use such a long TTL anymore. We should use a tiny value now (eg, 1 second), so that clients could always have up-to-date information...

prog/weaver/main.go
```diff
@@ -262,13 +290,15 @@ func determineQuorum(initPeerCountFlag int, peers []string) uint {
 	return quorum
 }
 
-func handleHTTP(router *weave.Router, httpAddr string, allocator *ipam.Allocator, defaultSubnet address.CIDR, docker *docker.Client) {
+func handleHTTP(router *weave.Router, httpAddr string, allocator *ipam.Allocator, defaultSubnet address.CIDR, docker *docker.Client, ns *nameserver.Nameserver, dnsserver *nameserver.DNSServer) {
```
inercia: Hmmm, don't you think this method has too many arguments...?

tomwilkie: Yes I think this might be worth refactoring at some point in the future.  But lets not role it into this change please.

prog/weaver/main.go
tomwilkie: Constants == hardcoded, there is not difference.  I think they should only be named-constants if reused elsewhere.  However, to save argument, I've done the move in 1882363

prog/weaver/main.go
tomwilkie: 1 second TTL in 1882363

nameserver/dns.go
tomwilkie: #1113

tomwilkie:
Merged coverage report from both integration and unit test: https://circle-artifacts.com/gh/weaveworks/weave/1724/artifacts/0/tmp/circle-artifacts.PPiWyIL/coverage.html

- dns.go 87.6%
- entry.go 92.6%
- http.go 52.0%
- nameserver.go 96.2%

http.go is dragger down as its a short file with of unexercised error handling, and has the GET handler for /name/{hostname} which is currently unused and should be removed (will do that now)

tomwilkie:
@inercia how are we getting on with this review?  Are we getting close to an LGTM?

PR #1012, opened 2015-06-25
Update README for 1.1 release

awh:
Merge as part of 1.1 release.

PR #1066, opened 2015-07-03
introduce 'Proxy' as 'Docker API Proxy'

### LINKED ISSUE #1019: rename 'proxy'
'proxy' is both too generic and too specific (for some users it means 'http proxy'). 'relay' has been suggested as an alternative name.

rade:
in order to reduce potential confusion about its role.

Closes #1019.

### DIFF
--- a/site/troubleshooting.md
+++ b/site/troubleshooting.md
@@ -142,7 +142,7 @@ is waiting for a while before connecting again.
 There may also be further sections for the
 [IP address allocator](ipam.html#troubleshooting),
 [weaveDNS](weavedns.html#troubleshooting), and
-[proxy](proxy.html#troubleshooting).
+[Weave Docker API Proxy](proxy.html#troubleshooting).
 
 ### <a name="list-attached-containers"></a>List attached containers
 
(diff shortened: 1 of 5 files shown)


site/troubleshooting.md
```diff
@@ -142,7 +142,7 @@ is waiting for a while before connecting again.
 There may also be further sections for the
 [IP address allocator](ipam.html#troubleshooting),
 [weaveDNS](weavedns.html#troubleshooting), and
-[proxy](proxy.html#troubleshooting).
+[Weave Docker API Proxy](proxy.html#troubleshooting).
```
rade: I'm calling it the "*Weave* Docker API Proxy" here because I reckon there isn't enough context here to tell that this is a weave component and not a docker component.

rade:
I've put this on `latest_release_doc_updates`. Merging this requires care; as well as propagating the change to `1.0` and `master` (while watching out for any extra proxy references that may merit changing), we also need to merge this into #1012.

Too onerous for a reviewer, IMO, so I suggest I do the merging myself, post review.

squaremo:
Changes look fine. Do we want to update the language in the guides too? (Doesn't need to be done here)

PR #1067, opened 2015-07-03
[proxy] print all stderr output from callWeave to the proxy logs, even if successful

### LINKED ISSUE #1056: Proxy should not swallow any messages that weave sends to stderr
The weave script may emit warnings or errors as it goes about its business, e.g. attaching a new container to the weave network.  Currently it seems these messages are swallowed.

At minimum they should go in the weaveproxy log.

Injecting them as output to be presented by the `docker` cli might be attractive, but potentially dangerous for users that are expecting very specific output from their containers.

Where should they go for clients that call via the Docker remote API?

paulbellamy:
Fixes #1056

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -57,9 +57,9 @@ func NewProxy(c Config) (*Proxy, error) {
 	p.client = client
 
 	if !p.WithoutDNS {
-		dockerBridgeIP, err := callWeave("docker-bridge-ip")
+		dockerBridgeIP, stderr, err := callWeave("docker-bridge-ip")
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf(string(stderr))
 		}
 		p.dockerBridgeIP = string(dockerBridgeIP)
 	}
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -18,13 +18,16 @@ var (
 	weaveWaitEntrypoint = []string{"/w/w"}
 )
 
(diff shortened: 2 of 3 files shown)


rade:
We *don't* want to log what comes out of stdout. We want to log a warning when anything comes out of stderr.

proxy/proxy.go
```diff
@@ -57,9 +57,9 @@ func NewProxy(c Config) (*Proxy, error) {
 	p.client = client
 
 	if !p.WithoutDNS {
-		dockerBridgeIP, err := callWeave("docker-bridge-ip")
+		dockerBridgeIP, stderr, err := callWeave("docker-bridge-ip")
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf(string(stderr))
```
rade: So previously this was returning `err`. What's the rationale for choosing one over the other? Should we perhaps return a combination of both?

Same applies at the other `callWeave` site.

paulbellamy: `err` from a command.Run() is not particularly useful. It will just say something like "exit code: 1". Which is why we should return stderr in both places.

rade:
It would be good to (manually) test this with the warning spat out by #1053, since that is what prompted filing of the issue.

rade:
> It would be good to (manually) test this with the warning spat out by #1053, since that is what prompted filing of the issue.

Have done this. All ok.

In future please create PR branches on weaveworks/weave.

PR #1069, opened 2015-07-03
Fix integration tests for OSX

2opremio:
My impressive first pull request. I tried to make architecture independent changes, but can somebody test it in Linux before merging?

### DIFF
--- a/test/500_weave_multi_cidr_test.sh
+++ b/test/500_weave_multi_cidr_test.sh
@@ -28,7 +28,7 @@ assert_zone_records() {
 
     records=$(weave_on $HOST status | grep "^$CID") || true
     # Assert correct number of records exist
-    assert "echo $records | grep -oE '\b([0-9]{1,3}\.){3}[0-9]{1,3}\b' | wc -l" $#
+    assert "echo $records | grep -oE '\b([0-9]{1,3}\.){3}[0-9]{1,3}\b' | wc -l | tr -d '[:space:]'" $#
 
     # Assert correct records exist
     for ADDR; do
--- a/test/config.sh
+++ b/test/config.sh
@@ -60,16 +60,16 @@ upload_executable() {
 remote() {
     rem=$1
     shift 1
-    "$@" > >(while read line; do echo -e "\e[0;34m$rem>\e[0m $line"; done)
+    "$@" > >(while read line; do echo -e $'\e[0;34m'"$rem>"$'\e[0m'" $line"; done)
 }
 
 colourise() {
-    [ -t 0 ] && echo -ne '\e['$1'm' || true
+    [ -t 0 ] && echo -ne $'\e['$1'm' || true
     shift
     # It's important that we don't do this in a subshell, as some
     # commands we execute need to modify global state
     "$@"
-    [ -t 0 ] && echo -ne '\e[0m' || true
+    [ -t 0 ] && echo -ne $'\e[0m' || true
 }
 
 whitely() {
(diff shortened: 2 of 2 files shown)


2opremio:
Oh, I see you are using Travis, I guess that implicitly checks I didn't break anything in Linux.

bboreham:
If you do the change in a branch in weaveworks/weave then CircleCI will run even more tests.

2opremio:
@bboreham Thanks, should I recreate the pull request in a weaveworks/weave branch?

bboreham:
It's not essential.  Next time.

2opremio:
@bboreham Is there some sort of  "review process"? Do I need a +1/LGTM from a team member to go ahead and merge?

bboreham:
Per the [development process](https://github.com/weaveworks/weave/wiki/WorkingOnWeave) (which I just updated), the reviewer merges.

I'll act as reviewer.

test/500_weave_multi_cidr_test.sh
```diff
@@ -28,7 +28,7 @@ assert_zone_records() {
 
     records=$(weave_on $HOST status | grep "^$CID") || true
     # Assert correct number of records exist
-    assert "echo $records | grep -oE '\b([0-9]{1,3}\.){3}[0-9]{1,3}\b' | wc -l" $#
+    assert "echo $records | grep -oE '\b([0-9]{1,3}\.){3}[0-9]{1,3}\b' | wc -l | tr -d '[[:space:]]'" $#
```
bboreham: Are two square-brackets necessary?  My reading suggests a single pair will do.

test/config.sh
```diff
@@ -60,16 +60,16 @@ upload_executable() {
 remote() {
     rem=$1
     shift 1
-    "$@" > >(while read line; do echo -e "\e[0;34m$rem>\e[0m $line"; done)
+    "$@" > >(while read line; do echo -e $"\e[0;34m$rem>\e[0m $line"; done)
```
bboreham: Why do we want this `$` ?  (I read the man-page and still don't know)
"A double-quoted string preceded by a dollar sign ($) will cause the string to be translated according to  the  current  locale."

test/500_weave_multi_cidr_test.sh
2opremio: You are right, they are not.

test/config.sh
2opremio: You are right again, they should be single quotes, but it also works with double-quotes for some reason.

For refernce: OSX's echo doesn't interpret escape characters (`\e`). By prepending `$`to a string, bash will 'encode' the escaped characters before passing them on to echo. 

```
   Words  of  the  form  $'string' are treated specially.  The word expands to string, with
   backslash-escaped characters replaced as specified by the ANSI  C  standard.   Backslash
   escape sequences, if present, are decoded as follows:
          \a     alert (bell)
          \b     backspace
          \e     an escape character
          \f     form feed
          \n     new line
          \r     carriage return
          \t     horizontal tab
          \v     vertical tab
          \\     backslash
          \'     single quote
          \nnn   the  eight-bit  character whose value is the octal value nnn (one to three
                 digits)
          \xHH   the eight-bit character whose value is the hexadecimal value  HH  (one  or
                 two hex digits)
          \cx    a control-x character

   The expanded result is single-quoted, as if the dollar sign had not been present.

   A  double-quoted string preceded by a dollar sign ($) will cause the string to be trans‐
   lated according to the current locale.  If the current locale is C or POSIX, the  dollar
   sign  is  ignored.  If the string is translated and replaced, the replacement is double-
   quoted.
```

2opremio:
@bboreham I addressed your comments and rebased with master

PR #1070, opened 2015-07-03
Change default IP allocation range from 10.128/10 to 10.32/12

### LINKED ISSUE #1036: default iprange clashes with DigitalOcean Private Networking
A user ran into
````
network 10.128.0.0/10 would overlap with route 10.128.0.0/16
Default -iprange 10.128.0.0/10 overlaps with existing route.
````
Further investigation by them revealed that

> It looks to be set up for my DigitalOcean's eth1 interface, which is for the (optional) private IP of a droplet. I suspect anyone who has a DigitalOcean droplet with Private Networking enabled may have this route by default.

If confirmed then we may want to consider changing the default iprange to avoid this clash.

That unfortunately is a backward incompatible change. *sigh*

### PR DESCRIPTION
**Pull request by bboreham**
To avoid clashing with Digital Ocean's and GCE's default private networks.
Fixes #1036 

(note this PR is raised against master; will re-raise if it was supposed to be against 1.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ can talk to each other...
 In the container started on `$HOST1`...
 
     root@a1:/# ping -c 1 -q a2
-    PING a2.weave.local (10.160.0.2) 56(84) bytes of data.
+    PING a2.weave.local (10.40.0.2) 56(84) bytes of data.
     --- a2.weave.local ping statistics ---
     1 packets transmitted, 1 received, 0% packet loss, time 0ms
     rtt min/avg/max/mdev = 0.341/0.341/0.341/0.000 ms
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by rade** [comment]
Looks like our new choice clashes with CircleCI. Don't know whether that is due to CircleCI itself or our specific CircleCI setup. @tomwilkie any clues?

**Comment by tomwilkie** [comment]
Id be surprised if it's a circle thing - must clash with gce.

I'll investigate tomorrow, but we should make this default smarter - for
instance, include it in the paxos first stage or something.

**Comment by tomwilkie** [comment]
Yeah on my GCE VM I have an IP address 10.240.96.113, which is in 10.192/10 I think

```
eth0      Link encap:Ethernet  HWaddr 42:01:0a:f0:60:71  
          inet addr:10.240.96.113  Bcast:10.240.96.113  Mask:255.255.255.255
          UP BROADCAST RUNNING MULTICAST  MTU:1460  Metric:1
          RX packets:7887697 errors:0 dropped:0 overruns:0 frame:0
          TX packets:8218029 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:9258627273 (9.2 GB)  TX bytes:7594574193 (7.5 GB)
```

bboreham:
what needs doing on this? I changed the range again.

rade:
ah, didn't see. No /11?

bboreham:
The even number seemed aesthetically much better, to my eyes. And halves the chance of something conflicting...

PR #1072, opened 2015-07-04
pass common `launch` args to all three components

### LINKED ISSUE #1043: pass common `weave launch` parameters to all components
Following #1009, three containers are started with one command.  I found it a bit surprising that extra arguments only get passed to the router, not to weavedns or proxy.

Currently, `-debug` seems to be the only one that would require this treatment, though fixing #329 would introduce another flag that needs to go to weavedns.

rade:
ATM that's just --loglevel=...

Closes #1043.

### DIFF
--- a/weave
+++ b/weave
@@ -946,6 +946,18 @@ proxy_addr() {
 # launch helpers
 ######################################################################
 
+common_launch_args() {
+    args=""
+    while [ $# -gt 0 ] ; do
+        case "$1" in
+            --log-level=*)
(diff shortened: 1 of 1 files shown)


rade:
> @squaremo removed their assignment 38 minutes ago

Why?

squaremo:
> @squaremo removed their assignment 38 minutes ago
> 
> Why?

Because I wasn't actively looking at it, and me being assigned dissuades anyone else from doing so.

PR #1073, opened 2015-07-04
make container aliveness check part of the ipam & dns API

### LINKED ISSUE #971: Checking IDs against containers should be explicitly optional
Not assigning an IP because a container is not running breaks the Docker plugin, because it cannot supply a container ID, only an endpoint ID.

rade:
...rather than conducting it implicitly when an ID "looks like" a container ID.

Closes #971.

### DIFF
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -1,19 +1,9 @@
 package docker
 
 import (
-	"regexp"
-
 	"github.com/fsouza/go-dockerclient"
 	. "github.com/weaveworks/weave/common"
 )
-
-// Regexp that matches container IDs
-var containerIDregexp = regexp.MustCompile("^[a-f0-9]+$")
-
-// IsContainerID returns True if the string provided is a valid container id
-func IsContainerID(idStr string) bool {
-	return containerIDregexp.MatchString(idStr)
-}
 
 // An observer for container events
 type ContainerObserver interface {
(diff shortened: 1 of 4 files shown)


inercia:
What is the rationale for this? I think that we check the aliveness _iff_ the container id looks like a container id, right?

rade:
> What is the rationale for this?

See the discussion in #971. The purpose is simply to avoid the "looks like a container ID" semantics, which was an implicit part of the API, and instead make it explicitly the callers responsibility to request an ID is treated as a container ID and checked for aliveness.

inercia:
Maybe  _"looks like a container ID"_ is not enough reason for checking for aliveness, but server should check that a valid ID has been provided anyway...

rade:
> server should check that a valid ID has been provided anyway

What does "valid ID" mean? And would the purpose of such a check be?

inercia:
Your code could happily check with Docker for aliveness of any container name (eg, `weave:expose`) as long as `check-alive` is `true`. I think that kind of checks should be caught earlier and save the query to Docker...

rade:
So you are saying that if a user makes a mistake, and asked to check the aliveness of something that isn't a container id, then we should catch that and suppress the check with docker? And do what? Add the record? Or not? It's exactly this kind of not-immediately obvious semantics that prompted @squaremo to file the issue.

squaremo:
Travis seems to have failed for a spurious reason. Restarted that build.

squaremo:
> squaremo removed their assignment just now

@rade If I'm looking into a PR or issue, I will assign myself. If you would like me to look at something, please invite me to do so by @-mentioning me, or by email, or even -- gasp -- by talking to me.

PR #1074, opened 2015-07-05
Enable ssh's forward agent in Vagrant image

2opremio:
This allows reusing the ssh keys in the host.


2opremio:
I just saw there is a mechanism for loading local configurations from Vagrant.local and I am not sure that ssh forwarding will be used by most people so I am not sure if this pull request makes sense anymore

Issue #68, opened 2014-09-18
make Java InetAddress.getLocalHost() (similar to hostname -i) return the weave IP address


bboreham:
At present, a container started in the way described in the weave README will have two IP addresses: one assigned by Docker and one assigned through weave.  So `hostname -I` will return something like:

    root@9b903541c5a7:/# hostname -I
    172.17.0.72 10.0.0.61 

however, `hostname -i` returns the Docker-assigned IP:

    root@9b903541c5a7:/# hostname -i
    172.17.0.72

I would prefer `hostname -i` to return the weave address

rade:
As of docker 1.4.1 you can add entries to /etc/hosts with `--add-host=name:IP`. Alas those entries get appended, which is of no use to resolving this issue.

qxo:
weave run with --add-host=name:IP --net=none will solve the problem!
weave should add these argument as default :)
eg:
```shell
ip=10.2.1.26;weave run  --with-dns $ip/24 --net=none --add-host=a1:$ip -ti -h a1  ubuntu

rade:
`--net=none` disables docker networking, i.e. you are left with *just* weave networking. That may be fine for some users, but it's not something we'd want to have as the default behaviour.

rade:
Also, if you are running with `--net=none` then the problem described in this issue doesn't manifest in the first place.

rade:
`/etc/hosts` [*is* writable](http://stackoverflow.com/questions/19414543/how-can-i-make-etc-hosts-writable-by-root-in-a-docker-container/25613983#25613983):

````
$ weave launch && weave launch-dns && weave launch-proxy
47921135b7b4943de2a1f56ff44a2060206e4ee8c7fc136a8e237958b710529a
4f19482849dc1e3166506b7da1aef5f9aebe6615f6c944eb1fae72dedf3bc496
4bc93553a81228a9253e58e48d21aebc582e4587b10ddded1e865b0ff2a94df6
$ eval $(weave proxy-env)
$ docker run --name=foo -ti ubuntu
root@foo:/# hostname -i
172.17.0.63
root@foo:/# echo > /etc/hosts
root@foo:/# hostname -i
10.128.0.2
````

I also tried prepending an entry and using docker links at the same time, to see whether docker's /etc/hosts updating stomps on the file. It doesn't; it just updated the entry for the link and left my modifications alone. The only gotcha I can see is that because the file is mounted, there is no atomic way to update it.

[...]

PR #1079, opened 2015-07-06
[proxy] weavewait rewrites /etc/hosts so weave ip is returned first from 'hostname -i'

paulbellamy:
Partial fix for #68. Won't help with dynamic `attach/detach` or the weave script.

### DIFF
--- a/prog/weavewait/main.go
+++ b/prog/weavewait/main.go
@@ -1,31 +1,55 @@
 package main
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
+	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"os/signal"
+	"sort"
+	"strings"
 	"syscall"
 
-	"github.com/weaveworks/weave/net"
+	weavenet "github.com/weaveworks/weave/net"
 )
 
 func main() {
 	if len(os.Args) <= 1 {
 		os.Exit(0)
 	}
-	args := os.Args[1:]
+
+	var (
+		args         = os.Args[1:]
+		notInExec    = true
+		rewriteHosts = true
+	)
 
 	if args[0] == "-s" {
+		notInExec = false
+		rewriteHosts = false
 		args = args[1:]
-	} else {
+	}
+
+	if args[0] == "-h" {
+		rewriteHosts = false
+		args = args[1:]
+	}
+
+	if notInExec {
 		usr2 := make(chan os.Signal)
 		signal.Notify(usr2, syscall.SIGUSR2)
 		<-usr2
 	}
 
-	_, err := net.EnsureInterface("ethwe", -1)
+	iface, err := weavenet.EnsureInterface("ethwe", -1)
 	checkErr(err)
+
+	if rewriteHosts {
+		updateHosts(iface)
+	}
 
 	binary, err := exec.LookPath(args[0])
 	checkErr(err)
@@ -38,4 +62,80 @@ func checkErr(err error) {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
+}
+
+// Prepend the addresses to /etc/hosts, removing the docker address
+func updateHosts(iface *net.Interface) {
+	addrs, err := iface.Addrs()
+	checkErr(err)
+	if len(addrs) == 0 {
+		return
+	}
+	hostname, err := os.Hostname()
+	checkErr(err)
+
+	hosts := parseHosts()
+
+	// Remove existing ips pointing to our hostname
+	toRemove := []string{}
+	for ip, addrs := range hosts {
+		for _, addr := range addrs {
+			if addr == hostname {
+				toRemove = append(toRemove, ip)
+				break
+			}
+		}
+	}
+	for _, ip := range toRemove {
+		delete(hosts, ip)
+	}
+
+	// Add the weave ip(s)
+	for _, addr := range addrs {
+		if addr, ok := addr.(*net.IPNet); ok {
+			ip := addr.IP.String()
+			hosts[ip] = append(hosts[ip], hostname)
+		}
+	}
+
+	writeHosts(hosts)
+}
+
+func parseHosts() map[string][]string {
+	f, err := os.Open("/etc/hosts")
+	checkErr(err)
+	defer f.Close()
+	ips := map[string][]string{}
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		// Remove any comments
+		if i := strings.IndexByte(line, '#'); i != -1 {
+			line = line[:i]
+		}
+
+		fields := strings.Fields(line)
+		if len(fields) > 0 {
+			ips[fields[0]] = fields[1:]
+		}
+	}
+	checkErr(scanner.Err())
+	return ips
+}
+
+func writeHosts(contents map[string][]string) {
+	ips := []string{}
+	for ip := range contents {
+		ips = append(ips, ip)
+	}
+	sort.Strings(ips)
+
+	buf := &bytes.Buffer{}
+	for _, ip := range ips {
+		if addrs := contents[ip]; len(addrs) > 0 {
+			fmt.Fprintf(buf, "%s\t%s\n", ip, strings.Join(addrs, " "))
+		}
+	}
+	checkErr(ioutil.WriteFile("/etc/hosts", buf.Bytes(), 644))
 }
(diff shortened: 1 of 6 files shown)


rade:
I think this should close #68 and we should open a new issue (or issues) covering the remaining cases.

@bboreham would you mind reviewing this?

paulbellamy:
I've tried this with Java's `InetAddress.getLocalHost()`, and it returns the weave address.

```bash
$ weave launch
$ docker $(weave proxy-config) run -ti --rm my-java-app
2c80b6ca84b7/10.128.0.2
$ docker run -ti --rm my-java-app
fc551c2b55a7/172.17.0.5
```

prog/weavewait/main.go
```diff
@@ -39,3 +45,33 @@ func checkErr(err error) {
...
+	addrs, err := iface.Addrs()
+	checkErr(err)
+	if len(addrs) == 0 {
+		return
+	}
+	hostname, err := os.Hostname()
+	checkErr(err)
+	var oldHosts, newHosts bytes.Buffer
+	for _, addr := range addrs {
+		fmt.Fprint(&newHosts, strings.Split(addr.String(), "/")[0], "\t", hostname, "\n")
```
bboreham: Might be simpler to type-check addr as an `IPNet` and fetch the address directly; if we encounter some other type of address at this point we don't really want to go string-manipulating it.

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -19,4 +19,7 @@ COMMITTED_IMAGE=$(proxy docker_on $HOST1 commit c1)
 assert_raises "proxy docker_on $HOST1 run --name c2 $COMMITTED_IMAGE"
 assert "entrypoint c2" "$(entrypoint $COMMITTED_IMAGE)"
 
+# Check containers get their weave IP prepended to /etc/hosts
```
bboreham: Maybe better to comment the effect rather than the mechanism (which isn't apparent here nor important)

prog/weavewait/main.go
bboreham: Also I would question the value of putting the ethwe ipv6 address into `/etc/hosts`

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -19,4 +19,7 @@ COMMITTED_IMAGE=$(proxy docker_on $HOST1 commit c1)
 assert_raises "proxy docker_on $HOST1 run --name c2 $COMMITTED_IMAGE"
 assert "entrypoint c2" "$(entrypoint $COMMITTED_IMAGE)"
 
+# Check weave IP is first ip returned, so java (etc) prefer weave.
+assert "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' $BASE_IMAGE hostname -i | cut -d' ' -f1" "10.2.1.1"
```
bboreham: In the presence of multiple addresses, `hostname -i` does not seem to guarantee any particular order.  By changing the first entry (added by `weavewait`) to 200.128.0.2 I can make it come second.

paulbellamy:
As @bboreham points out above, afaict `hostname -i` makes no guarantees about order. Nor does Java.

If we want to add the `ip hostname` into `/etc/hosts` we could use docker's `--add-hosts`, but as `hostname -I` already returns the weave IP, I don't see the point of that.

rade:
`--add-host` was considered in #68. It is of no use. The whole point of #68 is make `hostname -i` return the weave address. first.

We could *remove* the host entry for the docker ip.

paulbellamy:
Yeah. removing the docker ip seems a bit naughty..

rade:
> Yeah. removing the docker ip seems a bit naughty.

Meh. The docker ip should still show up with `hostname -I`. Provided that both inbound and outbound docker links still work that is fine, I reckon. And we can add a flag for disabling that behaviour.

paulbellamy:
Tried out links with the docker IP removed from /etc/hosts, and they still work fine. Not a big fan of adding yet another flag, but...

prog/weavewait/main.go
```diff
@@ -1,32 +1,54 @@
...
 	}
-	args := os.Args[1:]
+
+	var (
+		args         = os.Args[1:]
+		notInExec    = true
+		rewriteHosts = true
+	)
 
 	if args[0] == "-s" {
```
tomwilkie: Seems like this is getting to the stage where it might need 'proper' argument parsing, usage etc.

prog/weavewait/main.go
```diff
@@ -39,3 +61,43 @@ func checkErr(err error) {
...
+	checkErr(err)
+	if strings.Contains(oldHosts.String(), newHosts.String()) {
+		return
+	}
+	_, err = f.Seek(0, 0)
+	checkErr(err)
+	written, err := io.Copy(f, &newHosts)
+	checkErr(err)
+	// discard the first line, if it is the docker host
+	firstLine, err := oldHosts.ReadString('\n')
```
tomwilkie: This seems to discard the first line if it isn't 127.0.0.1?

paulbellamy: correct. We remove the docker ip from /etc/hosts, so that it doesn't show up in `hostname -i`.

prog/weavewait/main.go
```diff
@@ -39,3 +61,43 @@ func checkErr(err error) {
 		os.Exit(1)
 	}
 }
+
+// Prepend the addresses to /etc/hosts, removing the docker address
+func updateHosts(iface *net.Interface) {
```
tomwilkie: I think it might be best to try a different approach with this function.  This current version is too hard to follow and maintain IMO.

I would construct, in memory, what you want the file to be, maybe as a [][2]string.  Read the current contents, remove the entries you don't want, add the entries you do, sort(?) then write out to disk.

prog/weavewait/main.go
paulbellamy: Yeah. Pushed that off until next time it changes.

prog/weavewait/main.go
```diff
@@ -39,3 +63,79 @@ func checkErr(err error) {
...
+		if addr, ok := addr.(*net.IPNet); ok {
+			ip := addr.IP.String()
+			hosts[ip] = append(hosts[ip], hostname)
+		}
+	}
+
+	writeHosts(hosts)
+}
+
+func parseHosts() map[string][]string {
```
tomwilkie: This returns a map of ip -> []hostname, right?

I wonder if you made this hostname -> []ip, the code above might be simpler?

paulbellamy: The code above would be simpler, but writeHosts would be more complicated.

tomwilkie: Do you have to output the ips in sorted order?

tomwilkie:
LGTM

prog/weavewait/main.go
```diff
@@ -39,3 +63,79 @@ func checkErr(err error) {
...
+		line := scanner.Text()
+
+		// Remove any comments
+		if i := strings.IndexByte(line, '#'); i != -1 {
+			line = line[:i]
+		}
+
+		fields := strings.Fields(line)
+		if len(fields) > 0 {
+			ips[fields[0]] = fields[1:]
```
bboreham: I think it would be safer to append here, in case someone has a hosts file with the same IP address twice.

prog/weavewait/main.go
```diff
@@ -39,3 +63,79 @@ func checkErr(err error) {
 		os.Exit(1)
 	}
 }
+
+// Prepend the addresses to /etc/hosts, removing the docker address
```
bboreham: I don't think it does prepend.

prog/weavewait/main.go
```diff
@@ -39,3 +63,79 @@ func checkErr(err error) {
...
+	}
+	sort.Strings(ips)
+
+	buf := &bytes.Buffer{}
+	for _, ip := range ips {
+		if addrs := contents[ip]; len(addrs) > 0 {
+			fmt.Fprintf(buf, "%s\t%s\n", ip, strings.Join(addrs, " "))
+		}
+	}
+	checkErr(ioutil.WriteFile("/etc/hosts", buf.Bytes(), 644))
```
bboreham: Since you have removed all pre-existing comments and re-ordered the lines, it might be a good idea to add one new comment like "# modified by weave", for the benefit of puzzled users.

test/610_proxy_wait_for_weave_test.sh
bboreham: Now we are removing the weave IP the test is safe, but the comment in the test still refers to checking the "first ip".  I think it should be the only IP.

PR #1080, opened 2015-07-06
rename proxy-(env|config) to (env|config)

### LINKED ISSUE #1014: `weave env|config`
Following #962, it feels a bit strange to mention the proxy in the step following `launch`. I suggest we either rename `weave proxy-[env|config]` to `weave env|config`, or just add the latter. One thing this hinges on is error behaviour. IMO it is reasonable for `weave proxy-[env|config]` to fail, as it currently does, when the proxy isn't running. But that's perhaps not so desirable for `weave env|config`, where returning nothing might make more sense?

paulbellamy:
Keep error message to stderr, as the proxy should be launched with
'weave launch'. Or would you rather have no output at all?

Fixes #1014

### DIFF
--- a/weave
+++ b/weave
@@ -48,8 +48,8 @@ usage() {
     echo "weave hide          [<addr> ...]"
     echo "weave ps            [<container_id> ...]"
     echo "weave status"
-    echo "weave proxy-env"
-    echo "weave proxy-config"
+    echo "weave env"
+    echo "weave config"
     echo "weave version"
     echo "weave stop"
     echo "weave stop-router"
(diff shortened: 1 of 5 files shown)


rade:
> Or would you rather have no output at all?

Let's keep the output to stderr. I can envisage some users invoking `weave env` before `weave launch`, so it is helpful to provide a clue that this may not give the expected result.

rade:
The `README.md` changes need to branched off `weave_1.1_readme_updates`, so they become part of #1012.

Looks like it would be quite easy to keep `proxy-(env|config)` around for backward compatibility, writing a deprecation warning to stderr and executing the new (non-failing) behaviour. So let's do that.

weave
```diff
@@ -1128,10 +1127,18 @@ case "$COMMAND" in
         launch_proxy "$@"
         echo $PROXY_CONTAINER
         ;;
+    env)
+        proxy_addr "export DOCKER_HOST="
+        ;;
     proxy-env)
+        echo "proxy-env is deprecated, please use 'weave env'" >&2
```
rade:
````
env|proxy-env)
    [ "$COMMAND" = "env" ] || echo "$COMMAND is deprecated, please use 'weave env'" > &2
    proxy_addr "export DOCKER_HOST="
    ;;
````

PR #1084, opened 2015-07-06
Consistent option naming

### LINKED ISSUE #602: Inconsistent use of `-` vs `--` for long options in weave script
For example `-password` vs `--with-dns`. `--` is the convention for long options on GNU/Linux, we should support it and deprecate the other forms.

### LINKED ISSUE #1035: role of iprange not clear
A user reports "My worry about using the -iprange was that maybe it would restrict the ip range I could use weave with."

Looking at [our docs](http://docs.weave.works/weave/latest_release/ipam.html#range), it is indeed not absolutely clear that iprange is used for automatic allocation only.

squaremo:
Namely:

 * long options get two hyphens at the start; e.g., `--password`; and,
 * options made from multiple whole words are hyphenated; e.g., `--no-discovery`; and,
 * some options that are applied to DNS but sound universal have `dns` prepended, e.g., `--dns-port`.

There are exceptions; "ipalloc" is treated as one word for hyphenation; the `--tls*` options are left as-is since they correspond to Docker options; some niche options don't get hyphenated e.g., `--bufsz`; some options that are likely to disappear are left alone.

In all cases, the old version of an option will work, while outputting a deprecation warning. Mostly these will end up in container logs. Renamed options that are mentioned in the script usage, or otherwise specially-treated in the script, have a deprecation notice output to the console.

Closes #602 and closes #1035.

### DIFF
--- a/prog/weavedns/main.go
+++ b/prog/weavedns/main.go
@@ -1,12 +1,12 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"net"
 	"os"
 	"strconv"
 
+	"github.com/docker/docker/pkg/mflag"
 	"github.com/miekg/dns"
 	. "github.com/weaveworks/weave/common"
 	"github.com/weaveworks/weave/common/docker"
@@ -42,28 +42,28 @@ func main() {
 		err             error
 	)
 
-	flag.BoolVar(&justVersion, "version", false, "print version and exit")
-	flag.StringVar(&ifaceName, "iface", "", "name of interface to use for multicast")
-	flag.StringVar(&apiPath, "api", "unix:///var/run/docker.sock", "path to Docker API socket")
-	flag.StringVar(&domain, "domain", weavedns.DefaultLocalDomain, "local domain (ie, 'weave.local.')")
-	flag.IntVar(&wait, "wait", -1, "number of seconds to wait for interfaces to come up (0=don't wait, -1=wait forever)")
-	flag.IntVar(&dnsPort, "dnsport", weavedns.DefaultServerPort, "port to listen to DNS requests")
-	flag.StringVar(&httpIfaceName, "httpiface", "", "interface on which to listen for HTTP requests (defaults to empty string which listens on all interfaces)")
-	flag.IntVar(&httpPort, "httpport", weavedns.DefaultHTTPPort, "port to listen to HTTP requests")
-	flag.IntVar(&cacheLen, "cache", weavedns.DefaultCacheLen, "cache length")
-	flag.IntVar(&ttl, "ttl", weavedns.DefaultLocalTTL, "TTL (in secs) for responses for local names")
-	flag.BoolVar(&watch, "watch", true, "watch the docker socket for container events")
-	flag.StringVar(&logLevel, "log-level", "info", "logging level (debug, info, warning, error)")
+	mflag.BoolVar(&justVersion, []string{"#version", "-version"}, false, "print version and exit")
+	mflag.StringVar(&ifaceName, []string{"#iface", "-iface"}, "", "name of interface to use for multicast")
+	mflag.StringVar(&apiPath, []string{"#api", "-api"}, "unix:///var/run/docker.sock", "path to Docker API socket")
+	mflag.StringVar(&domain, []string{"#domain", "-domain"}, weavedns.DefaultLocalDomain, "local domain (ie, 'weave.local.')")
+	mflag.IntVar(&wait, []string{"#wait", "-wait"}, -1, "number of seconds to wait for interfaces to come up (0=don't wait, -1=wait forever)")
+	mflag.IntVar(&dnsPort, []string{"#dnsport", "#-dnsport", "-dns-port"}, weavedns.DefaultServerPort, "port to listen to DNS requests")
+	mflag.StringVar(&httpIfaceName, []string{"#httpiface", "#-httpiface", "-http-iface"}, "", "interface on which to listen for HTTP requests (empty string means listen on all interfaces)")
+	mflag.IntVar(&httpPort, []string{"#httpport", "#-httpport", "-http-port"}, weavedns.DefaultHTTPPort, "port to listen to HTTP requests")
+	mflag.IntVar(&cacheLen, []string{"#cache", "-cache"}, weavedns.DefaultCacheLen, "cache length")
+	mflag.IntVar(&ttl, []string{"#ttl", "-ttl"}, weavedns.DefaultLocalTTL, "TTL (in secs) for responses for local names")
+	mflag.BoolVar(&watch, []string{"#watch", "-watch"}, true, "watch the docker socket for container events")
+	mflag.StringVar(&logLevel, []string{"-log-level"}, "info", "logging level (debug, info, warning, error)")
 	// advanced options
-	flag.IntVar(&negTTL, "neg-ttl", 0, "negative TTL (in secs) for unanswered queries for local names (0=same value as -ttl")
-	flag.IntVar(&refreshInterval, "refresh", weavedns.DefaultRefreshInterval, "refresh interval (in secs) for local names (0=disable)")
-	flag.IntVar(&maxAnswers, "max-answers", weavedns.DefaultMaxAnswers, "maximum number of answers returned to clients (0=unlimited)")
-	flag.IntVar(&relevantTime, "relevant", weavedns.DefaultRelevantTime, "life time for info in the absence of queries (in secs)")
-	flag.IntVar(&udpbuf, "udpbuf", weavedns.DefaultUDPBuflen, "UDP buffer length")
-	flag.IntVar(&timeout, "timeout", weavedns.DefaultTimeout, "timeout for resolutions (in millisecs)")
-	flag.BoolVar(&cacheDisabled, "no-cache", false, "disable the cache")
-	flag.StringVar(&fallback, "fallback", "", "force a fallback server (ie, '8.8.8.8:53') (instead of /etc/resolv.conf values)")
-	flag.Parse()
+	mflag.IntVar(&negTTL, []string{"#neg-ttl", "-neg-ttl"}, 0, "negative TTL (in secs) for unanswered queries for local names (0=same value as --ttl")
+	mflag.IntVar(&refreshInterval, []string{"#refresh", "-refresh"}, weavedns.DefaultRefreshInterval, "refresh interval (in secs) for local names (0=disable)")
+	mflag.IntVar(&maxAnswers, []string{"#max-answers", "#-max-answers", "-dns-max-answers"}, weavedns.DefaultMaxAnswers, "maximum number of answers returned to clients (0=unlimited)")
+	mflag.IntVar(&relevantTime, []string{"#relevant", "-relevant"}, weavedns.DefaultRelevantTime, "life time for info in the absence of queries (in secs)")
+	mflag.IntVar(&udpbuf, []string{"#udpbuf", "#-udpbuf", "-dns-udpbuf"}, weavedns.DefaultUDPBuflen, "UDP buffer length for DNS")
+	mflag.IntVar(&timeout, []string{"#timeout", "#-timeout", "-dns-timeout"}, weavedns.DefaultTimeout, "timeout for resolutions (in millisecs)")
+	mflag.BoolVar(&cacheDisabled, []string{"#no-cache", "-no-cache"}, false, "disable the cache")
+	mflag.StringVar(&fallback, []string{"#fallback", "#-fallback", "-dns-fallback"}, "", "force a fallback server (ie, '8.8.8.8:53') (instead of /etc/resolv.conf values)")
+	mflag.Parse()
 
 	if justVersion {
 		fmt.Printf("weave DNS %s\n", version)
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -5,7 +5,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/pborman/getopt"
+	"github.com/docker/docker/pkg/mflag"
 	. "github.com/weaveworks/weave/common"
 	"github.com/weaveworks/weave/proxy"
 )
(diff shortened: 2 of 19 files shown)


rade:
So no `--option=` except for the tls args?

squaremo:
> So no --option= except for the tls args?

The mflag library will in general treat `--option=value` and `--option value` as the same. The exception, as usual, is boolean flags, where the `--option bool` variety is ruled out.

rade:
> The mflag library will in general treat `--option=value` and `--option value` as the same

Yes, but the script doesn't.

squaremo:
> Yes, but the script doesn't.

Oh I see. Are you saying it did before and it doesn't now?

rade:
> Are you saying it did before and it doesn't now?

No, it didn't, but it seems odd for the exe to support it (if it did so previously then that was by accident) but not the script. Same goes for any short options.

squaremo:
> No, it didn't, but it seems odd for the exe to support it (if it did so previously then that was by accident) but not the script.

The script doesn't in general support it because no-one thought to implement it (probably because it's a bit of a pain, unless you happen to be messing around with the arguments anyway, as in the case of `--tls*`).

The exes *do* support that style because the mflag package happens to support that. As does the flag package. In other words, nothing has changed on that front.

rade:
> The script doesn't in general support it because no-one thought to implement

I actually didn't know it worked.

> probably because it's a bit of a pain

Only very slightly. I see no reason why the `launch-router` arg processing couldn't handle this in very much the same way as `proxy_args` and `dns_args.

squaremo:
OK, I agree that would be nice, but I'm not trying to address that here, only to rename the options in a backwards-compatible way.

rade:
@paulbellamy would you mind reviewing this?

bboreham:
Is this also supposed to close #1035?

squaremo:
> Is this also supposed to close #1035?

That was the thing I actually wanted to fix, yes good point.

weave
```diff
@@ -26,12 +26,13 @@ CIDR_REGEXP="$IP_REGEXP/[0-9]{1,2}"
...
-    echo "weave launch        [-password <password>] [-nickname <nickname>] [-nodiscovery]"
-    echo "                        [-iprange <cidr> [-ipsubnet <cidr>]]"
-    echo "                        [-initpeercount <count>] <peer> ..."
-    echo "weave launch-router [-password <password>] [-nickname <nickname>] [-nodiscovery]"
-    echo "                        [-iprange <cidr> [-ipsubnet <cidr>]]"
-    echo "                        [-initpeercount <count>] <peer> ..."
+    echo "weave launch        [--password <password>] [--nickname <nickname>] [--no-discovery]"
+    echo "                        [--ipalloc-range <cidr> [--ipalloc-default-subnet <cidr>]]"
+    echo "                        [--init-peer-count <count>] <peer> ..."
+    echo "weave launch-router [--password <password>] [--nickname <nickname>]"
```
rade: Three of the above lines are >80 wide.

weave
```diff
@@ -1093,6 +1093,61 @@ DOCKER_VERSION_PATCH=$(echo "$DOCKER_VERSION" | cut -d. -f 3)
...
+                deprecation_warning "[-]-initpeercount" "--init-peer-count"
+                [ ${1#--} == "initpeercount" ] && shift
+                ;;
+            --init-peer-count)
+                shift
+                ;;
+            -no-default-ipam|--no-default-ipam)
+                deprecation_warning "[-]-no-default-ipam" "--no-default-ipalloc"
+                ;;
+        esac
```
rade: I wouldn't bother with catching `--<oldoption>` as deprecated. None of our docs referred to `--` options. The only exception is `--no-default-ipam` where the opposite is the case.

That way the deprecation notice doesn't contain the unwieldy `[-]-` notation.

squaremo: Yeah those are ugly. Good point.

weave
squaremo: Argh

weave
rade: The above deprecation handles `--password=*` and `--iprange=`. Yet the old code in the script doesn't. Hence #1087. That's a little inconsistent.

rade: Why is this intercepting `--<new-option>`? After all, we are not intercepting all options here, only documented ones, so this won't prevent accidental parsing of option values as options, which I don't think is a problem anyway (plus there is other option parsing code in the script which doesn't go through that trouble).

prog/weavedns/main.go
```diff
@@ -42,28 +42,28 @@ func main() {
...
+	mflag.IntVar(&negTTL, []string{"#neg-ttl", "-neg-ttl"}, 0, "negative TTL (in secs) for unanswered queries for local names (0=same value as -ttl")
+	mflag.IntVar(&refreshInterval, []string{"#refresh", "-refresh"}, weavedns.DefaultRefreshInterval, "refresh interval (in secs) for local names (0=disable)")
+	mflag.IntVar(&maxAnswers, []string{"#max-answers", "#-max-answers", "-dns-max-answers"}, weavedns.DefaultMaxAnswers, "maximum number of answers returned to clients (0=unlimited)")
+	mflag.IntVar(&relevantTime, []string{"#relevant", "-relevant"}, weavedns.DefaultRelevantTime, "life time for info in the absence of queries (in secs)")
+	mflag.IntVar(&udpbuf, []string{"#udpbuf", "#-udpbuf", "#-dns-udp-buf"}, weavedns.DefaultUDPBuflen, "UDP buffer length for DNS")
+	mflag.IntVar(&timeout, []string{"#timeout", "#-timeout", "-dns-timeout"}, weavedns.DefaultTimeout, "timeout for resolutions (in millisecs)")
+	mflag.BoolVar(&cacheDisabled, []string{"#no-cache", "-no-cache"}, false, "disable the cache")
+	mflag.StringVar(&fallback, []string{"#fallback", "#-fallback", "-dns-fallback"}, "", "force a fallback server (ie, '8.8.8.8:53') (instead of /etc/resolv.conf values)")
+	mflag.Parse()
```
rade: oh how much more readable this would be if gofmt didn't stomp on vertical alignment:
````
	mflag.BoolVar(  &justVersion,     []string{"#version",                      "-version"},         false,                           "print version and exit")
	mflag.StringVar(&ifaceName,       []string{"#iface",                        "-iface"},           "",                              "name of interface to use for multicast")
	mflag.StringVar(&apiPath,         []string{"#api",                          "-api"},             "unix:///var/run/docker.sock",   "path to Docker API socket")
	mflag.StringVar(&domain,          []string{"#domain",                       "-domain"},          weavedns.DefaultLocalDomain,     "local domain (ie, 'weave.local.')")
	mflag.IntVar(   &wait,            []string{"#wait",                         "-wait"},            -1,                              "number of seconds to wait for interfaces to come up (0=don't wait, -1=wait forever)")
	mflag.IntVar(   &dnsPort,         []string{"#dnsport",     "#-dnsport",     "-dns-port"},        weavedns.DefaultServerPort,      "port to listen to DNS requests")
	mflag.StringVar(&httpIfaceName,   []string{"#httpiface",   "#-httpiface",   "-http-iface"},      "",                              "interface on which to listen for HTTP requests (empty string means listen on all interfaces)")
	mflag.IntVar(   &httpPort,        []string{"#httpport",    "#-httpport",    "-http-port"},       weavedns.DefaultHTTPPort,        "port to listen to HTTP requests")
	mflag.IntVar(   &cacheLen,        []string{"#cache",                        "-cache"},           weavedns.DefaultCacheLen,        "cache length")
	mflag.IntVar(   &ttl,             []string{"#ttl",                          "-ttl"},             weavedns.DefaultLocalTTL,        "TTL (in secs) for responses for local names")
	mflag.BoolVar(  &watch,           []string{"#watch",                        "-watch"},           true,                            "watch the docker socket for container events")
	mflag.StringVar(&logLevel,        []string{                                 "-log-level"},       "info",                          "logging level (debug, info, warning, error)")
	// advanced options
	mflag.IntVar(   &negTTL,          []string{"#neg-ttl",                      "-neg-ttl"},         0,                               "negative TTL (in secs) for unanswered queries for local names (0=same value as -ttl")
	mflag.IntVar(   &refreshInterval, []string{"#refresh",                      "-refresh"},         weavedns.DefaultRefreshInterval, "refresh interval (in secs) for local names (0=disable)")
	mflag.IntVar(   &maxAnswers,      []string{"#max-answers", "#-max-answers", "-dns-max-answers"}, weavedns.DefaultMaxAnswers,      "maximum number of answers returned to clients (0=unlimited)")
	mflag.IntVar(   &relevantTime,    []string{"#relevant",                     "-relevant"},        weavedns.DefaultRelevantTime,    "life time for info in the absence of queries (in secs)")
	mflag.IntVar(   &udpbuf,          []string{"#udpbuf",      "#-udpbuf",      "#-dns-udp-buf"},    weavedns.DefaultUDPBuflen,       "UDP buffer length for DNS")
	mflag.IntVar(   &timeout,         []string{"#timeout",     "#-timeout",     "-dns-timeout"},     weavedns.DefaultTimeout,         "timeout for resolutions (in millisecs)")
	mflag.BoolVar(  &cacheDisabled,   []string{"#no-cache",                     "-no-cache"},        false,                           "disable the cache")
	mflag.StringVar(&fallback,        []string{"#fallback",    "#-fallback",    "-dns-fallback"},    "",                              "force a fallback server (ie, '8.8.8.8:53') (instead of /etc/resolv.conf values)")
````
....which makes it much easier to spot the odd ones. What's going on with `udpbuf`?

File: weave (hunk 2 above)
Comment (squaremo) [inline weave]: > Why is this intercepting --<new-option>?

As in `--ipalloc-default-subnet)` ? Yes, just to skip over the next argument. I think it's still worth attempting accidental parsing of values as options, even if it can't be infallible.

Comment (squaremo) [inline weave]: > I wouldn't bother with catching --<oldoption> as deprecated. None of our docs referred to -- options.

and

> The above deprecation handles --password=* and --iprange=. Yet the old code in the script doesn't.

They aren't documented, but since they are understood (I think?), typos or misreadings may well have ended up in scripts. I agree that the `[-]-` thing is ugly though, so maybe I'll remove that and rely on people to interpret the deprecation message.

File: prog/weavedns/main.go (hunk 3 above)
Comment (squaremo) [inline prog/weavedns/main.go]: Och I know. Those bits are a mess. Piiiiike!

Umm. The udpbuf line is just wrong. The last option should just be `-dns-udpbuf` (if it carries over into #1065 at all).

File: weave [hunk 4]
```diff
@@ -1093,6 +1093,61 @@ DOCKER_VERSION_PATCH=$(echo "$DOCKER_VERSION" | cut -d. -f 3)
...
+        shift
+    done
+}
+
+deprecation_warnings "$@"
```
rade: We really only want to run this for `launch[-router]` (plus a ` -no-default-ipam|--no-default-ipam` check for `launch-proxy`). Otherwise we end up with deprecation warnings for `weave --local run -ti myimage -password foo`

weave
rade: `deprecation_warning` could take `$1` as the first argument and strip off any trailing `=` for display. Reduces the noise somewhat.

rade: That would also address the `[-]-` issue, i.e. we would display whatever option was supplied.

weave
squaremo: Hmmm yeeessss -- that did occur to me, concerned as I am with not accidentally parsing values! I would like to see the docker run arguments after a `--`; but, this hasn't happened yet, and would be a breaking change :(

Perhaps I can just run deprecate_warnings in select branches of the case statement following.

prog/weaveproxy/main.go
```diff
@@ -15,6 +15,23 @@ var (
 	defaultListenAddrs = []string{"tcp://0.0.0.0:12375", "unix:///var/run/weave.sock"}
 )
 
+type listOpts struct {
+	values *[]string
```
bboreham: why is this a pointer to a slice, rather than a slice?

weave
squaremo: Yes, that works nicely!

prog/weaveproxy/main.go
squaremo: Because `ListVar` is handed a pointer to a variable, so the value of the variable can be modified in place.

weave
squaremo: > Perhaps I can just run deprecate_warnings in select branches of the case statement following.

This is what I have done.

weave
```diff
@@ -1105,6 +1104,59 @@ DOCKER_VERSION_PATCH=$(echo "$DOCKER_VERSION" | cut -d. -f 3)
...
+                shift
+                ;;
+            -no-default-ipam|--no-default-ipam)
+                deprecation_warning $1 "--no-default-ipalloc"
+                ;;
+        esac
+        shift
+    done
+}
+
```
rade: These two functions should move up a bit, so that the
````
COMMAND=$1
shift
````
comes just before the `case "$COMMAND" in` again.

File: prog/weaveproxy/main.go (hunk 5 above)
Comment (paulbellamy) [inline prog/weaveproxy/main.go]: In the same vein, why not use `github.com/docker/opts.HostListVar`?

File: weave [hunk 7]
```diff
@@ -1105,6 +1104,59 @@ DOCKER_VERSION_PATCH=$(echo "$DOCKER_VERSION" | cut -d. -f 3)
 COMMAND=$1
 shift 1
 
+deprecation_warning() {
```
paulbellamy: Use this for `proxy-env` and `proxy-config` as well, please.

prog/weaveproxy/main.go
squaremo: Because it's not in `pkg/`, so it can have dependencies on other bits of Docker, which I don't want to drag in.

paulbellamy:
There are remaining references to single-hyphen flags in:
bin/multiweave:33
docs/dockerless.txt:128-129
proc/weavedns/main.go:58 (in the help text)
prog/weaver/main.go:96,149,158
site/ipam.md:140

weave
squaremo: Do those take arguments? They aren't documented :(

prog/weaveproxy/main.go
bboreham: OK, you don't need the struct wrapper; just `type listOpts []string`.  That's what I did in https://github.com/weaveworks/weave/blob/1076-ignore-iface/prog/netcheck/netcheck.go

weave
paulbellamy: They don't take arguments, but do output a deprecation warning, so we should use this for consistency.

squaremo:
> There are remaining references

Nice spotting, thank you :)

prog/weaveproxy/main.go
squaremo: Good point! I was adapting from the package Paul mentions above, and didn't flatten the single-field struct.

squaremo: Actually there's a worse problem here, which is that the listOpts is purely additive, so if you supply an argument it gets _added_ to the default rather than replacing it.

weave
squaremo: Ahh, got it, ta

PR #1088, opened 2015-07-06
detect `--dns-search <arg>`

### LINKED ISSUE #1086: `--dns-search <arg>` not recognised
When the weave script checks for the presence of `docker run` dns options, it recognises `--dns-search=<arg>` but not `--dns-search <arg>`, which is equally valid.

Failure to recognise the option results in weave adding its own `--dns-search=.`, which will cause it to be searched after whatever option got specified. That's unlikely to be disastrous in many cases, but can still be surprising and is contrary to our docs.

rade:
...and not just `--dns-search=<arg>`

Fixes #1086.

### DIFF
--- a/weave
+++ b/weave
@@ -664,8 +664,11 @@ dns_args() {
     WITHOUT_DNS=
     while [ $# -gt 0 ] ; do
         case "$1" in
-            --dns-search|--dns-search=*)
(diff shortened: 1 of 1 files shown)


rade:
gah. wrong branch.

PR #1090, opened 2015-07-06
Don't complain about weave bridge routes clashing

### LINKED ISSUE #1076: `expose` causes network overlap complaint on restart
````
$ weave launch
$ weave expose
10.128.0.2
$ route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.1.254   0.0.0.0         UG    0      0        0 wlan0
10.128.0.0      0.0.0.0         255.192.0.0     U     0      0        0 weave
172.17.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
192.168.1.0     0.0.0.0         255.255.255.0   U     9      0        0 wlan0
192.168.48.0    0.0.0.0         255.255.255.0   U     0      0        0 vboxnet0
$ weave stop
$ weave launch
Network 10.128.0.0/10 overlaps with existing route 10.128.0.0/10 on host.
Default -iprange 10.128.0.0/10 overlaps with existing route on host.
````

Perhaps we should ignore the weave bridge in our overlap calculations.

### PR DESCRIPTION
**Pull request by bboreham**
Added --ignore-iface option to netcheck and use it to ignore weave bridge route

Fixes #1076

### DIFF
--- a/net/route.go
+++ b/net/route.go
@@ -9,13 +9,17 @@ import (
 
 // A network is considered free if it does not overlap any existing
 // routes on this host. This is the same approach taken by Docker.
-func CheckNetworkFree(subnet *net.IPNet) error {
+func CheckNetworkFree(subnet *net.IPNet, ignoreIfName string) error {
 	routes, err := netlink.RouteList(nil, netlink.FAMILY_V4)
 	if err != nil {
 		return err
 	}
 	for _, route := range routes {
 		if route.Dst != nil && overlaps(route.Dst, subnet) {
+			if link, err := netlink.LinkByIndex(route.LinkIndex); err == nil && link.Attrs().Name == ignoreIfName {
+				continue
+			}
+
 			return fmt.Errorf("Network %s overlaps with existing route %s on host.", subnet, route.Dst)
 		}
 	}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by rade** [comment]
- needs to be branched off 1.0
- it should be possible to supply multiple `--ignore-iface` options.

**Comment by bboreham** [comment]
> it should be possible to supply multiple --ignore-iface options.

Why?

**Comment by rade** [comment]
> Why?

It's an obvious generalisation. If you were to write a `netcheck` independently of weave, you'd certainly want this.

**Comment by bboreham** [comment]
Various other changes, starting with #1053, have gone into this code on master and not on 1.0.
Do you want them all on 1.0, or do you want to maintain the separate threads of development?

**Comment by rade** [comment]
> Various other changes, starting with #1053, have gone into this code on master and not on 1.0.

Let's not bother putting this on 1.0 then; until we have a few users complaining about it.

**Comment by bboreham** [comment]
OK; it's in branch 1076-ignore-iface-1.0 when we need it.
(I had already done it by the time I'd come back to read your comment)

File: net/route.go [hunk 1]
```diff
@@ -29,14 +25,31 @@ func overlaps(n1, n2 *net.IPNet) bool {
...
-			return fmt.Errorf("Address %s overlaps with existing route %s on host.", addr, route.Dst)
+		if link, err := netlink.LinkByIndex(route.LinkIndex); err == nil {
+			for _, ignoreIfName := range ignoreIfNames {
+				if link.Attrs().Name == ignoreIfName {
+					continue routesLoop
+				}
+			}
+		}
+		if err := check(route); err != nil {
+			return err
```
rade: sort the interface name slice with `sort.Strings(slice)` and then find elements with `sort.SearchStrings(slice, elem)`. Should work out shorter than the code above.

rade: I wonder whether `ignoreIfNames` might just be a tad confusing. Stick to `ignoreIfaceNames`?

rade:
> OK; it's in branch 1076-ignore-iface-1.0 when we need it.

Have looked at it. Drags in too much. Delete. (though perhaps push the branch to your own repo).

bboreham:
I did it with a `map` instead.

rade:
needs re-basing, by the looks of it.

bboreham:
Rebased

PR #1097, opened 2015-07-07
WIP peer discovery using swarm discovery backends

### LINKED ISSUE #850: use Swarm discovery backends
Weave could be instructed to connect to peers registered in a Swarm backend. For example, we could:
* launch the router with something like `weave launch --with-swarm=etcd://myhost.com/mycluster`.
* start a WeaveSwarm container if not running (in the same way we currently do with WeaveDNS)
* this container would run a daemon that would use the corresponding [Swarm discovery backend](https://github.com/docker/swarm/tree/master/discovery) for sending `/connect`/`/forget` requests to the router.

inercia:
This PR implements a peers discovery mechanism for discovering peers by using Swarm's discovery backends. A new _weavediscovery_ container will monitor peers appearing and leaving the discovery endpoint, and will trigger the corresponding `/connect` and `/forget` in the router. We can join or leave discovery endpoints with the `weave join` and `weave leave` commands.

Note: the `file://` backend is not supported, as _weavediscovery_ does not have access to host files so far...

Closes #850

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,12 +36,11 @@ prog/weaveproxy/weaveproxy
 prog/sigproxy/sigproxy
 prog/weavewait/weavewait
 prog/netcheck/netcheck
+prog/weavediscovery/weavediscovery
 tools/bin
 tools/build
 releases
-.weaver.uptodate
-.weavedns.uptodate
-.weaveexec.uptodate
+*.uptodate
 weave.tar
 prog/weaveexec/weave
 prog/weaveexec/weaveproxy
--- a/Makefile
+++ b/Makefile
@@ -11,24 +11,27 @@ WEAVE_VERSION=git-$(shell git rev-parse --short=12 HEAD)
 
 WEAVER_EXE=prog/weaver/weaver
 WEAVEDNS_EXE=prog/weavedns/weavedns
+WEAVEDISCOVERY_EXE=prog/weavediscovery/weavediscovery
 WEAVEPROXY_EXE=prog/weaveproxy/weaveproxy
 SIGPROXY_EXE=prog/sigproxy/sigproxy
 WEAVEWAIT_EXE=prog/weavewait/weavewait
 NETCHECK_EXE=prog/netcheck/netcheck
 
-EXES=$(WEAVER_EXE) $(WEAVEDNS_EXE) $(SIGPROXY_EXE) $(WEAVEPROXY_EXE) $(WEAVEWAIT_EXE) $(NETCHECK_EXE)
+EXES=$(WEAVER_EXE) $(WEAVEDNS_EXE) $(WEAVEDISCOVERY_EXE) $(SIGPROXY_EXE) $(WEAVEPROXY_EXE) $(WEAVEWAIT_EXE) $(NETCHECK_EXE)
 
 WEAVER_UPTODATE=.weaver.uptodate
 WEAVEDNS_UPTODATE=.weavedns.uptodate
+WEAVEDISCOVERY_UPTODATE=.weavediscovery.uptodate
 WEAVEEXEC_UPTODATE=.weaveexec.uptodate
 
-IMAGES_UPTODATE=$(WEAVER_UPTODATE) $(WEAVEDNS_UPTODATE) $(WEAVEEXEC_UPTODATE)
+IMAGES_UPTODATE=$(WEAVER_UPTODATE) $(WEAVEDNS_UPTODATE) $(WEAVEDISCOVERY_UPTODATE) $(WEAVEEXEC_UPTODATE)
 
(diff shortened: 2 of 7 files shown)


squaremo:
What's the use case for having more than one endpoint?

inercia:
@squaremo Well, it is up to the user. This would allow users to discover and join peers in several URLs, maybe some with a more _static_ nature while some other more _dynamic_ (eg, do `weave join file://some/path` and `weave join etcd://otherserver/some/path`)

squaremo:
OK, but it comes with a cost -- if we don't have a clear use for it, maybe we shouldn't pay the cost?

inercia:
Supporting multiple discovery endpoints is  maybe 100 lines of code (some parts of the `manager.go`). In fact, the whole _discovery_ thing is quite lightweight. I don't think there is a big _cost_...

squaremo:
> Supporting multiple discovery endpoints is maybe 100 lines of code

The cost isn't just LoC. It's complication in the mental model we're asking the user (and future maintainers) to understand, and documentation, for instance. In any case: for something that's unnecessary, any cost is too high!

So, is multiple endpoints necessary? It seems _possible_ that a user would want it. Not sure about plausible, though. Would you want a weave network that spanned two Swarm clusters? Maybe?

inercia:
@squaremo Ok, what do you think about this? when the user `weave join` a discovery _group_, it will automatically disconnect from any previously joined _group_?

squaremo:
> @squaremo Ok, what do you think about this? when the user weave join a discovery group, it will automatically disconnect from any previously joined group?

I think that'd be more surprising than the original scheme ;)

I'm still not sure why would a user would need either variety though. I mean, I could invent a scenario, but I don't recall connecting multiple clusters being mentioned by any users -- maybe it's just not something that would occur as a possibility to people.

Shall we invite more opinions? @errordeveloper @tomwilkie @awh

squaremo:
By the way I should say that I think the idea on the whole is a very nice one @inercia. And I am hoping to steal^W factor out the WeaveClient bit for the plugin and proxy.

bboreham:
Similar to WeaveClient: https://github.com/bboreham/weave/blob/soaktest/weaveapi/weaveapi.go

errordeveloper:
The Swarm discovery has multiple backends and I am not sure how we could recommend either of those backends to the user. The Docker Hub backend is somewhat like "default", but it seems to be still experimental and, personally, I am not quite sure about certain aspects of the approach it takes.

squaremo:
> The Swarm discovery has multiple backends and I am not sure how we could recommend either of those backends to the user.

Unfortunately, I suspect we will have to view them as a fact of life: that's what's available to users of Docker.

errordeveloper:
I was pondering whether external discovery agent is what we should provide as a reference implementation, one that do one thing well. If user needs a different backend, they could go ahead and reimplement the discovery agent based on the reference we provide.

inercia:
@errordeveloper This feature would not only make things easier for users that want to integrate with Swarm (regardless of the current state of Swarm and its backends), but this should also be seen as a discovery mechanism by its own that is, at the same time, _compatible_ with the discovery backends used by Swarm (and I'm not sure but I think this would also be compatible with the new overlay [kv store](https://github.com/docker/libnetwork/blob/master/docs/overlay.md#install-and-bootstrap-kv-store) in Docker)

I don't think we should spend time developing/maintaining any discovery backends (let alone supporting user backends), so that's the reason why we should steal^W reuse Swarm's discovery engine. So maybe this is a win-win...

rade:
I suggest that we initially put this in its own repo and release it as an experimental component, separately from weave. Much as we did with weavedns at the start. Let's get this in the hands of our users and see what they make of it!

rade:
> Note: the `file://` backend is not supported, as weavediscovery does not have access to host files so far...

Thoughts on fixing this? Volume mounts?

inercia:
@rade That was my original plan: to parse the URL in the script and, if the scheme is `file`, then try a a volume mount. However, I think a volume cannot be attached to a running container (maybe I'm wrong), so it would require to restart the container if the user wants to `leave` a `file://` URL and `join` another one...

And regarding moving this to its own repo, you are right: that would be a much better place...

rade:
> it would require to restart the container if the user wants to `leave` a file:// URL and join another `one`...

The whole dynamic join/leave strikes me as rather uncommon.

Bouncing the container should be fine.

inercia:
@rade Could you create a new repo for this, with a name you consider appropriate? I will then close this PR and create a new PR for that repo...

rade:
> Could you please create a new repo for this

https://github.com/weaveworks/discovery

inercia:
Thanks, @rade.

Closing this PR...

Issue #1029, opened 2015-06-26
encrypt handshake


rade:
We should consider encrypting the handshake, perhaps with a key derived from the password via some suitable KDF. This was suggested in [HackerNews](https://news.ycombinator.com/item?id=9507798).

#82 presents one use case. Another is that it allows us to populate the handshake with more information without needing to be concerned about its potential value to an attacker.

Finally, this closes off some possible attack vectors against the handshake, such as #912 (though note that simply encrypting the handshake is not enough to solve that one; the framing needs to change).

dpw:
Why do we need some special encryption for the handshake?  The only things actually required to set up the usual encryption for the TCP connection are the other end's public key and the version negotiation.  
The key pair is randomly generated.  Having version negotiation unencrypted doesn't seem like a big risk (TLS is plagued by downgrade attacks, but that is due to its long history and rampant parameterization, so I don't think the same concerns apply to weave).

rade:
> Why do we need some special encryption for the handshake?

Well, that's the topic of the discussion in that HN thread :)

dpw:
HN-driven development.

rade:
quite :)

rade:
Well, it would be good to avoid multi-phase handshakes. So if we want to permit arbitrary, potentially sensitive, information in the handshake, then it needs to be encrypted.

dpw:
Maybe I'm misunderstanding, but you think that avoiding a round trip when establishing a TCP connection is worth having some special encryption scheme for the handshake?

rade:
Perhaps, yes.

dpw:
How about we do the simple thing first, and elaborate it when a user tells us "hey, it would be really great if weave took a few milliseconds less to establish an encrypted connection"?

rade:
> How about we do the simple thing first

Fair enough. Remind me, what problem are we trying to solve with that? #912? #82? Or something else?

dpw:
Well, the problem I'd really like to solve is to be able to add to the information exchanged when a connection is established without worrying about whether it is sensitive.

Yes on #912: given what will need to change, it looks like it can be addressed for free (the public key is a fixed length blob, so sending it pre-gob is easy).

No on #82: I think this will address the "to prevent spoofing" part of that issue, but there is a bunch of other work needed there.

rade:
@dpw I believe you are working on this. -> assigning

PR #1098, opened 2015-07-07
New connection protocol version with encrypted features map

### LINKED ISSUE #912: DoS vector in protocol message stream
I haven't tested this, but I reckon there is DoS vector in the protocol message stream that attackers can exploit without breaking our crypto.

The weave TCP protocol is a gob stream of messages: a `map[string][string]` for the first message, the handshake, followed by `[]byte` messages. The latter are all encrypted when crypto is enabled. Note that it's the *individual messages* that are encrypted, *not* the stream. And the first message - the handshake - is always unencrypted.

It is therefore possible for an attacker to exploit weaknesses in gob. The following spring to mind:

1. triggering an OOM crash by sending a very large gob message
2. triggering an OOM crash by sending a sequence of octets that triggers a large memory allocation in gob
3. triggering a loop in the gob decoder

I am quite sure 1 is possible. There is a [1GB limit](http://golang.org/src/encoding/gob/decoder.go?h=tooBig#L15) in place but a) that's still quite large, and b) doesn't apply everywhere.

2 is different from 1 because it does not require network capacity and is therefore impossible to guard against with firewalls that enforce capacity limits. And I think it is possible, based on my reading of the gob decoder, i.e. there are [places](http://golang.org/src/encoding/gob/decode.go#L378) where a capacity is read and then an allocation of that capacity is performed.

I don't think 3 is possible, based on the description of gob. But I could well be wrong.

dpw:
Addresses #1029.

Although the new protocol version fixes #912, because compatibility support for the old protocol version is present, we cannot declare #912 fixed by this PR.  Unless there are other overriding factors, I suggest we retain compatibility for one major release, i.e. drop V1 compatibility in 1.2.  I'll create an issue for that if/when this gets merged.


rade:
> I suggest we retain compatibility for one major release, i.e. drop V1 compatibility in 1.2.

That's a good plan since it allows a rolling upgrade from weave 1.0 to 1.1 and then to 1.2.

However, it also does delay the fix to #912 for a fair while. We could add an option to the router to allow manual adjustment of the min supported protocol version. 1.1 users could use that to insist on V2, thus eliminating the #912 DoS vector. They could even do a rolling upgrade of 1.0 that way, i.e. first roll out a normal upgrade to 1.1, and then roll out an upgrade that bumps the min protocol version.

rade:
> We could add an option to the router to allow manual adjustment of the min supported protocol version.

And the default for 1.1 could be to only speak V2, so it's secure by default.

dpw:
> And the default for 1.1 could be to only speak V2, so it's secure by default.

I don't see much point in making the compatibility optional and defaulting to incompatible.  Users won't set the option until they run into the incompatibility, by which point any serious inconvenience that might occur has likely already occurred. "Compatible but not by default" effectively means incompatible.

Given how long this issue has been around, and the relatively low risk that someone will want to DoS a weave deployment at this point, and the high probability that there are other ways to DoS weave if they did want to, I recommend leaving #912 open for one more major release.

If we do make compatibility optional, then it should default to compatible. Paranoid users can then control the risk of #912 if they want. Though any paranoid users should already have firewalls ensuring that weave can only receive from a restricted set of IP addresses.

rade:
> If we do make compatibility optional, then it should default to compatible.

I agree with that *but* also think it is trumped by security. "secure+incompatible" wins over "insecure+compatible".

dpw:
> If we do make compatibility optional, then it should default to compatible.
> 
> I agree with that but also think it is trumped by security. "secure+incompatible" wins over "insecure+compatible".

This is a DoS issue, not a break in weave's encryption.  So as a security issue, it is at the weak end of the spectrum.  This is tacitly recognized by the fact that #912 has been open for weeks with no rush to get it fixed.

In that context, the ordering between "secure+incompatible" and "insecure+compatible" is not so clear. Having an option for the paranoid seems like providing our users with a choice.  Having an option to maintain compatibility seems like such a weak form of compatibility that I doubt it is worth the engineering effort.  Why not just say "1.1 is not compatible with 1.0 due to a security issue"?

rade:
> Why not just say "1.1 is not compatible with 1.0 due to a security issue"?

Could do, but seems a shame given that we *can* make it compatible, and you've already written the code for it.

If we do introduce a min-version option, and make v1 ("compatible but with DoS vector") the default, do you think we can claim to have solved #912?

rade:
>  Users won't set the option until they run into the incompatibility, by which point any serious inconvenience that might occur has likely already occurred.

There will likely be a few things like that in 1.1, i.e. situations where special care, and modifications of scripts, is needed to perform a rolling upgrade from 1.0.

dpw:
> Could do, but seems a shame given that we can make it compatible, and you've already written the code for it.

The compatibility code in this PR is merely a necessary condition for 1.1 to be compatible with 1.0.  Work in other areas may be required for compatibility.

Also, if compatibility were not an objective, I would have removed the vestigial layer of gob encoding that remains in the V2 protocol but is only used to encode `[]byte` items.

> If we do introduce a min-version option, and make v1 ("compatible but with DoS vector") the default, do you think we can claim to have solved #912?

I think we can reasonably claim to have solved #912 with the PR as it stands.
 
I think we can reasonably claim not to have solved #912 until the V1 protocol code is removed.

If we include a min-version option that gives the user a choice between security and compatibility, we'll have to document its implications carefully in the release notes whatever the default.  After that, whether we say #912 is solved or not is academic.

rade:
> Also, if compatibility were not an objective, I would have removed the vestigial layer of gob encoding that remains in the V2 protocol but is only used to encode []byte items.

Interesting. I'm guessing you are talking about replacing layer 4 in #519 with a simple length-prefix encoding with a reasonable size bound?

dpw:
> There will likely be a few things like that in 1.1, i.e. situations where special care, and modifications of scripts, is needed to perform a rolling upgrade from 1.0.

We should keep an eye on whether the end result is worth it.  I assume the desire expressed for compatibility between different versions of weave is because our users wish to be able to mix different versions and have them work together without worrying about it.  If we have a "compatibility" that requires a contorted process on the user's part, then we may find that no-one actually follows that route, deciding that it's easier and safer to do a full deployment instead.

dpw:
> Interesting. I'm guessing you are talking about replacing layer 4 in #519 with a simple length-prefix encoding with a reasonable size bound?

Yes.  With compatibility in mind, it seemed better to keep the v1-to-2 protocol changes focused on a single goal, and leave other changes for future protocol revisions.  But without compatibility, that degobification would be close to trivial.

rade:
The users we have had discussions with are actually quite ok with spending some effort on an upgrade. Their main concern is the ability to perform a *rolling* upgrade. As long as that is possible, even if it involves some modifications to existing config/deployment scripts, they are happy.

There does of course come a point where the contortions required become too cumbersome. Arguably the same is true of our code base.

I think we are close to the point with 1.1 where breaking compatibility is the least worst option.

rade:
@dpw I have cherry-picked the first three commits onto master. Please rebase. Was going to pick the fourth too but found a bug in it. Please fix.

rade:
Thinking about documenting a `--min-version` option has put me off the idea.

dpw:
> Thinking about documenting a --min-version option has put me off the idea.

Already implemented.  We can leave it undocumented, or rip the option out easily enough.

rade:
I have cherry-picked 7d5e6b5 onto master.

rade:
@dpw pls rebase.

dpw:
> @dpw pls rebase.

Done.  But there were no conflicts so I'm not sure why this was requested.

rade:
> Done. But there were no conflicts so I'm not sure why this was requested.

GH thought it needed a rebase. Presumably because of the cherry-pick.

rade:
linter is unhappy.

dpw:
> linter is unhappy.

Good

rade:
> Good

Please fix. What it is complaining about doesn't strike me as onerous.

dpw:
Care to tell me what it is complaining about?  I don't really wish to work out what options or filtering on its output I need to apply to avoid the silly suggestions.

rade:
Look at the travis ci report. `bin/lint` in our repo already filters out the nonsense.

router/connection.go
```diff
@@ -356,6 +380,76 @@ func (conn *LocalConnection) run(actionChan <-chan ConnectionAction, finished ch
...
+
+	remoteConnID, err := strconv.ParseUint(str, 10, 64)
+	if err != nil {
+		return nil, err
+	}
+
+	conn.uid ^= remoteConnID
+	return NewPeer(name, nickName, uid, 0), nil
+}
+
```
rade: I have simplified this in https://github.com/rade/weave/commit/ed1efd2539941641af37d775a25dcaa2e948922f

dpw:
Apparently it now has travis' blessing.

router/protocol.go
```diff
@@ -2,58 +2,329 @@ package router
...
+	Outbound   bool
+}
+
+type ProtocolIntroResults struct {
+	Features   map[string]string
+	Receiver   TCPReceiver
+	Sender     TCPSender
+	SessionKey *[32]byte
+	Version    byte
+}
```
rade: should the `Features` type perhaps be `features` (or some suitably renaming of that)? After all we are passing this to `LocalConnection.parseFeatures(features features)`.

router/protocol.go
```diff
@@ -2,58 +2,329 @@ package router
...
+	if pubKey != nil {
+		features["PublicKey"] = hex.EncodeToString(pubKey[:])
+	}
+
+	enc := gob.NewEncoder(params.Conn)
+	dec := gob.NewDecoder(params.Conn)
+
+	// Encode in a separate goroutine to avoid the possibility of
+	// deadlock
+	encodeDone := make(chan error, 1)
```
rade: Is this 1-capacity in order to prevent the goroutine from getting stuck and hanging around forever in the event of an error in `dec.Decode` below? If so, that merits a comment. If not, then make it 0.

dpw: Yes.  Comments added.

router/protocol.go
dpw: This was deliberate.  I wanted to restrict the code in `protocol.go` to dealing with the protocol only, and exclude the interpretation of the data passed over the protocol (this line is subjective, but the result seems quite clean to me).  And I regard the `features` type as just a utility for the sake of `parseFeatures` (and possibly for other future uses of the features in connection-related code).  So it would seem out of place if hoisted into `protocol.go`.

rade: ok. I thought that might be the reason. rationale accepted.

router/connection.go
```diff
@@ -292,7 +309,13 @@ func (conn *LocalConnection) run(actionChan <-chan ConnectionAction, finished ch
 		return
 	}
 
-	conn.Log("completed handshake; using protocol version", conn.version)
+	if conn.SessionKey == nil {
+		conn.Decryptor = NewNonDecryptor()
+	} else {
+		conn.Decryptor = NewNaClDecryptor(conn.SessionKey, conn.outbound)
+	}
```
rade: I think this should be done as part of `DoIntro`, in `setup[No]Crypto`. That way all the crypto setup is in one place. It also means we can remove `conn.SessionKey`.

router/protocol.go
```diff
@@ -2,58 +2,335 @@ package router
...
+			return ErrExpectedNoCrypto
+		}
+
+		rbuf = make([]byte, len(pubKey))
+		if _, err := io.ReadFull(params.Conn, rbuf); err != nil {
+			return err
+		}
+
+		res.setupCrypto(params, gob.NewEncoder(params.Conn),
+			gob.NewDecoder(params.Conn), rbuf, privKey)
```
rade: cosmetic: I'd put the `gob.NewDecoder` on the same line as `gob.NewEncoder`, here and above.

rade:
It took me a while to get my head around this. It's a fair bit of extra code, and the control flow is more involved than previously. Alas I have not been able to come up with anything better. So modulo the two minor issues above, this LGTM.

rade:
btw, I think this should close #912, and we should open a separate issue to remove v1 protocol support.

router/connection.go
rade: Unfortunately `conn.SessionKey` is required by `LocalConnection.ensureForwarders` for creating the encryptors. Which also suggests that having the decryptor initialisation in `LocalConnection` rather than in `ProtocolIntroResults.setup[No]Crypto` is the right thing.

PR #1099, opened 2015-07-07
Proxy autoconfig

### LINKED ISSUE #1055: Weave Docker API Proxy may broaden the set of users that can call Docker
In the default configuration of Docker, starting the weave proxy (as root) moves you from a situation where you must be root to call docker to one where any user can call docker via the proxy's TCP address.

We should warn users of this potential weakening of security, and perhaps not listen on TCP by default.

paulbellamy:
Use the docker daemon's `-H` flags as the proxy's default.

No support for auto-configuring TLS. For that we would have to do the procfs and flag parsing in the weave script (so that we could mount the certs into the proxy container).

This is possibly a bit linux-specific, as the layout of procfs is not standardized.

Fixes #1055

### DIFF
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -11,8 +11,7 @@ import (
 )
 
 var (
-	version            = "(unreleased version)"
-	defaultListenAddrs = []string{"tcp://0.0.0.0:12375", "unix:///var/run/weave.sock"}
+	version = "(unreleased version)"
 )
 
 type listOpts struct {
@@ -50,7 +49,7 @@ func main() {
 
 	mflag.BoolVar(&justVersion, []string{"#version", "-version"}, false, "print version and exit")
 	mflag.StringVar(&logLevel, []string{"-log-level"}, "info", "logging level (debug, info, warning, error)")
-	ListVar(&c.ListenAddrs, []string{"H"}, defaultListenAddrs, "addresses on which to listen")
+	ListVar(&c.ListenAddrs, []string{"H"}, nil, "addresses on which to listen")
 	mflag.StringVar(&c.HostnameMatch, []string{"-hostname-match"}, "(.*)", "Regexp pattern to apply on container names (e.g. '^aws-[0-9]+-(.*)$')")
 	mflag.StringVar(&c.HostnameReplacement, []string{"-hostname-replacement"}, "$1", "Expression to generate hostnames based on matches from --hostname-match (e.g. 'my-app-$1')")
 	mflag.BoolVar(&c.NoDefaultIPAM, []string{"#-no-default-ipam", "-no-default-ipalloc"}, false, "do not automatically allocate addresses for containers without a WEAVE_CIDR")
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -36,14 +36,21 @@ The first form is more convenient, however you can only pass proxy
 related configuration arguments to `launch-proxy` so if you need to
 modify the default behaviour you will have to use the latter.
 
-By default, the proxy listens on /var/run/weave.sock and port 12375, on
-all network interfaces. This can be adjusted with the `-H` argument, e.g.
+By default, the proxy decides where to listen based on how the
+launching client connects to docker. If the launching client connected
+over a unix socket, the proxy will listen on /var/run/weave.sock. If
+the launching client connected over TCP, the proxy will listen on port
+12375, on all network interfaces. This can be adjusted with the `-H`
+argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:9999
 
-Multiple -H arguments can be specified. If you are working with a remote
-docker daemon, then any firewalls inbetween need to be configured to permit
-access to the proxy port.
+When launching the proxy via TLS, `-H` and/or [TLS options](#tls) are
+required.
+
+Multiple `-H` arguments can be specified. If you are working with a
+remote docker daemon, then any firewalls inbetween need to be
+configured to permit access to the proxy port.
 
 All docker commands can be run via the proxy, so it is safe to adjust
 your `DOCKER_HOST` to point at the proxy. Weave provides a convenient
--- a/weave
+++ b/weave
@@ -17,6 +17,8 @@ BASE_EXEC_IMAGE=$DOCKERHUB_USER/weaveexec
 EXEC_IMAGE=$BASE_EXEC_IMAGE:$IMAGE_VERSION
 PROXY_HOST=${PROXY_HOST:-$(echo "${DOCKER_HOST#tcp://}" | cut -s -d: -f1)}
 PROXY_HOST=${PROXY_HOST:-127.0.0.1}
+DOCKER_CLIENT_HOST=${DOCKER_CLIENT_HOST:-$DOCKER_HOST}
+DOCKER_CLIENT_TLS_VERIFY=${DOCKER_CLIENT_TLS_VERIFY:-$DOCKER_TLS_VERIFY}
 
 # Define some regular expressions for matching addresses.
 # The regexp here is far from precise, but good enough.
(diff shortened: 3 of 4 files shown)


paulbellamy:
per f2f discussion w/ @rade:

let's add TLS support via a binary where the weave script can do:
```
if [ -x $SOME_BINARY ]; then
  PROXY_VOLUMES=$($SOME_BINARY)
fi
```

This way, --local will work, but not support TLS autoconfiguration.

rade:
`command_exists $SOME_BINARY && $($SOME_BINARY) || true`

weave
```diff
@@ -949,6 +949,12 @@ proxy_args() {
     done
 }
 
+proxy_autoconfig() {
+    if command_exists proxyvolumes ; then
+        PROXY_VOLUMES=${PROXY_VOLUMES:-$(proxyvolumes "$@")}
+    fi
+}
```
rade: I *think* this should work:

`! command_exists proxyvolumes || PROXY_VOLUMES=${PROXY_VOLUMES:-$(proxyvolumes "$@")}`

rade: Also, I'd make this part of `proxy_args`.

proxy/flags.go
```diff
@@ -0,0 +1,210 @@
...
+	for {
+		i := bytes.IndexByte(b, 0)
+		if i == -1 {
+			break
+		}
+		strs = append(strs, string(b[0:i]))
+		b = b[i+1:]
+	}
+	return strs
+}
```
rade: `strings.Split(string(b[:len(b)-1]), "\000")`?

rade: (and looking at the call site, you could actually just use `strings.Split(string(b), "\000")`, I think)

test/675_proxy_autoconfig_test.sh
```diff
@@ -0,0 +1,53 @@
...
+  --make-pidfile \
+  -- \
+    -d \
+    -H unix:///var/run/docker.sock \
+    -H tcp://0.0.0.0:2375 \
+    --tlsverify \
+    --tlscacert $PWD/tls/ca.pem \
+    --tlscert   $PWD/tls/$HOST1.pem \
+    --tlskey    $PWD/tls/$HOST1-key.pem \
+    2>&1 >/dev/null
```
rade: won't this break, now that #1125 has been merged? Also perhaps we should add the tls args using `sed` as happens in #1125. That way we are less susceptible to changes in the docker configuration. Or, perhaps even better, why not configure the docker with tls enabled all the time, in the vagrantfile?

paulbellamy: Surprisingly, works fine.

I'm a bit averse to adding tls all the time, as we normally use HTTP to talk to it so we would constantly need to configure the client to use tls.

rade: agreed; I had forgotten that if one enables docker tls then one *must* speak it.

paulbellamy:
doesn't work with systemd as docker launches with: `-H fd://`

paulbellamy:
per f2f discussion w/ @rade:

Configuring based on the daemon is a bit of a hack, so let's configure where to listen based on the client.
The three cases are:

1) They connect via unix socket. We should listen on unix:///var/run/weave.sock
2) They connect via plain TCP. We should listen on tcp://0.0.0.0:12375
3) They connect via TLS. We should throw an error telling them to configure the proxy manually.

site/proxy.md
```diff
@@ -36,13 +36,18 @@ The first form is more convenient, however you can only pass proxy
...
+over a unix socket, the proxy will listen on /var/run/weave.sock. If
+the launching client connected over TCP, the proxy will listen on port
+12375, on all network interfaces. This can be adjusted with the `-H`
+argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:9999
 
-Multiple -H arguments can be specified. If you are working with a remote
-docker daemon, then any firewalls inbetween need to be configured to permit
+If you are connecting via TLS, you will need to specify `-H`. Multiple
```
rade: You should only have to specify the TLS options. Also, please make this a link to the tls section.

prog/weaveproxy/main.go
```diff
@@ -85,3 +87,44 @@ func main() {
...
+func loadDefaultListenAddr(c *proxy.Config) {
+	if len(c.ListenAddrs) > 0 {
+		return
+	}
+
+	tlsEnabled := (os.Getenv("DOCKER_CLIENT_TLS_VERIFY") != "")
+
+	clientArgs := strings.Fields(os.Getenv("DOCKER_CLIENT_ARGS"))
+	filteredArgs := []string{}
+	for i := 0; i < len(clientArgs); i++ {
```
rade: `i, clientArg := range clientArgs`?

rade: argh; can't do that because of the 'skip'. nevermind.

prog/weaveproxy/main.go
```diff
@@ -85,3 +87,44 @@ func main() {
...
+			if !strings.Contains(clientArgs[i], "=") {
+				if i++; i < len(clientArgs) {
+					filteredArgs = append(filteredArgs, clientArgs[i])
+				}
+			}
+		}
+	}
+
+	if clientTLSEnabled && !(c.TLSConfig.Enabled || c.TLSConfig.Verify) {
+		log.Fatalln("Cannot autoconfigure the proxy when using TLS. Please specify -H")
```
rade: error message needs updating.

prog/weaveproxy/main.go
```diff
@@ -85,3 +87,44 @@ func main() {
...
+			if !strings.Contains(clientArgs[i], "=") {
+				if i++; i < len(clientArgs) {
+					filteredArgs = append(filteredArgs, clientArgs[i])
+				}
+			}
+		}
+	}
+
+	if clientTLSEnabled && !(c.TLSConfig.Enabled || c.TLSConfig.Verify) {
+		log.Fatalln("Cannot autoconfigure proxy listen address when launching via TLS")
```
rade: That error message is still misleading. We *do* autoconfigure the proxy listen address. What we *don't* do is configure TLS.

prog/weaveproxy/main.go
```diff
@@ -85,3 +87,44 @@ func main() {
...
+	f.StringVar(&host, []string{"H", "-host"}, host, "")
+	if err := f.Parse(filteredArgs); err != nil {
+		log.Fatalf("Error configuring proxy: %s", err)
+	}
+
+	c.ListenAddrs = []string{"tcp://0.0.0.0:12375"}
+	if host == "" || strings.HasPrefix(host, "unix://") {
+		c.ListenAddrs = []string{"unix:///var/run/weave.sock"}
+	}
+}
```
rade: Most of the code in this whole function should be moved to the script. Passing configuration params to the proxy via env vars is gross. Also, we can do much better error handling in the script, i.e. the user won't just be told that proxy died.

weave
```diff
@@ -935,6 +974,22 @@ proxy_args() {
         esac
         shift
     done
+
+    if [ -z "$PROXY_HOST" ]; then
+      if [ -n "$CLIENT_TLS_ENABLED" -a -z "$PROXY_TLS_ENABLED" ]; then
+        echo "When launching proxy via TLS, -H and/or TLS options are required." >&2
```
rade: "the proxy"

site/proxy.md
```diff
@@ -36,14 +36,18 @@ The first form is more convenient, however you can only pass proxy
...
+argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:9999
 
-Multiple -H arguments can be specified. If you are working with a remote
-docker daemon, then any firewalls inbetween need to be configured to permit
-access to the proxy port.
+Multiple -H arguments can be specified. If you are working with a
+remote docker daemon, then any firewalls inbetween need to be
+configured to permit access to the proxy port.
```
rade: I think we should say something about the TLS case and refer to the TLS section.

weave
```diff
@@ -895,6 +900,25 @@ show_addrs() {
...
+            CLIENT_TLS_ENABLED=1
+            ;;
+          -H|--host)
+            DOCKER_CLIENT_HOST="$2"
+            shift
+            ;;
+          -H=*|--host=*)
+            DOCKER_CLIENT_HOST="${1#*=}"
+            ;;
+        esac
```
rade: I'd put the tls case after the host cases, for consistency with `proxy_args`.

weave
rade: ditch the nesting; make it a single condition.

rade:
rest LGTM

PR #1104, opened 2015-07-09
Don't wait to claim addresses from IPAM

### LINKED ISSUE #1083: `weave launch` hangs on IPAM claim
@pacaletx encountered the following problem attempting to dynamically grow a weave 0.11.2 network ([IRC transcript](https://botbot.me/freenode/weavenetwork/2015-07-06/?msg=43773908&page=1)):

On `$HOST1`
````
weave launch -iprange 10.2.3.0/24 -initpeercount 1
weave launch-dns "10.9.1.1/24" -debug
weave launch-proxy --with-dns --with-ipam
````

Later, attempt to add `$HOST2`
````
weave launch -iprange 10.2.3.0/24 -initpeercount 2
````
with the intent of following up with `launch-dns`/`launch-proxy` and `weave connect $HOST1`. This is not possible as the `weave launch` on $HOST2 hangs:

````
Allocator range [10.2.3.0-10.2.4.0) awaiting consensus: Nodes known: 1, Quorum size: 2
Pending requests:
  Claim 1f38974edf82148f65b1a721003a89f708e922a2846a280e8f5bdb87353882e8 -> 10.2.3.1
Allocator default subnet: 10.2.3.0/24
````

@bboreham diagnosed that $HOST2 had extant application containers with `ethwe` interfaces addressed inside the IPAM range; `weave launch` was therefore waiting on IPAM quorum to reclaim these addresses, a state which would never be reached as the two routers were not yet connected. User confirmed that a `weave connect $HOST1` in a separate session on $HOST2 allowed the `weave launch` to complete.

Possible workarounds:

* Ensure no application containers are running when weave is launched
* Specify IP address to second launch
* Run `weave connect` asynchronously

all of which are inconvenient in other ways.

Finally, it was not clear why `weave launch` was stuck (similar concerns are echoed [here](https://github.com/weaveworks/weave/issues/1023#issuecomment-118781742))

### PR DESCRIPTION
**Pull request by bboreham**
If an error is immediately apparent, e.g. we have already allocated that address on this host, or we have been told via gossip that another host owns that range, then it will go back to the caller.
If an error is possible but not detectable yet, then the claim is left pending and the caller is told it's ok (nil error return).

Another change in behaviour is that we stop tracking whether the caller has closed the http socket before we managed to reply properly; since this happens naturally on the first 'nil' return and we don't expect to be processing the request very long anyway.

Fixes #1083

### DIFF
--- a/ipam/claim.go
+++ b/ipam/claim.go
@@ -9,10 +9,9 @@ import (
 )
 
 type claim struct {
-	resultChan       chan<- error
-	ident            string
-	addr             address.Address
-	hasBeenCancelled func() bool
+	resultChan chan<- error
+	ident      string
+	addr       address.Address
 }
 
 func (c *claim) sendResult(result error) {
@@ -29,11 +28,6 @@ func (c *claim) sendResult(result error) {
 
 // Try returns true for success (or failure), false if we need to try again later
 func (c *claim) Try(alloc *Allocator) bool {
-	if (c.hasBeenCancelled)() {
-		c.Cancel()
-		return true
-	}
-
 	if !alloc.ring.Contains(c.addr) {
 		// Address not within our universe; assume user knows what they are doing
 		alloc.infof("Ignored address %s claimed by %s - not in our universe\n", c.addr, c.ident)
--- a/ipam/claim.go
+++ b/ipam/claim.go
@@ -3,6 +3,7 @@ package ipam
 import (
 	"fmt"
 
+	"github.com/weaveworks/weave/common"
 	"github.com/weaveworks/weave/ipam/address"
 	"github.com/weaveworks/weave/router"
 )
@@ -12,6 +13,18 @@ type claim struct {
 	ident            string
 	addr             address.Address
 	hasBeenCancelled func() bool
+}
+
+func (c *claim) sendResult(result error) {
+	// Make sure we only send a result once, since listener stops listening after that
+	if c.resultChan != nil {
+		c.resultChan <- result
+		close(c.resultChan)
+		c.resultChan = nil
+	}
+	if result != nil {
+		common.Error.Println("[allocator] " + result.Error())
+	}
 }
 
 // Try returns true for success (or failure), false if we need to try again later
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by bboreham** [comment]
@rade would you like to review this?

**Comment by rade** [comment]
I've looked at it superficially. I'd like @tomwilkie to take a closer look.

Can you explain what the new behaviour is? i.e. what errors, if any, and when, are making it back the http client?

**Comment by bboreham** [comment]
If an error is immediately apparent, e.g. we have already allocated that address on this host, or we have been told via gossip that another host owns that range, then it will go back to the caller.
If an error is possible but not detectable yet, then the claim is left pending and the caller is told it's ok (nil error return).

File: ipam/claim.go [hunk 1]
```diff
@@ -3,28 +3,35 @@ package ipam
...
-		return true
+func (c *claim) sendResult(result error) {
+	// Make sure we only send a result once, since listener stops listening after that
+	if c.resultChan != nil {
+		c.resultChan <- result
+		close(c.resultChan)
+		c.resultChan = nil
+	}
+	if result != nil {
+		common.Error.Println("[allocator] " + result.Error())
```
tomwilkie: This error could do with a little more context.  "Delayed claim failed for %s: %v" or such like.

tomwilkie:
I believe the code does what it says it should.  I'm also guessing no one can think of a better way of solving this problem - it seems like a shame to swallow responses like this.

ipam/claim.go
bboreham: As written, this line is used in both the delayed and non-delayed cases.  Could infer 'delayed' from `resultChan == nil`, I guess.

tomwilkie: I think you just need an else

bboreham:
The responses go in the weave container log.  I pondered putting them in `weave status` too, but am unclear if a user with this problem is any more likely to look there.

ipam/claim.go
bboreham: I felt it was a good idea to log all failed claims.

awh:
I've reproduced the original report (0.11.2) on 1.0.1, and confirmed it is fixed by this PR.

I will handle the manual merge to master once the branch merge is in.

PR #1105, opened 2015-07-09
Code coverage for smoke tests.

### LINKED ISSUE #673: coverage reporting for smoke tests
It would be great to get coverage reports for the golang code exercised by our smoke tests.

tomwilkie:
Fixes #673

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ WEAVE_EXPORT=weave.tar
 WEAVEEXEC_DOCKER_VERSION=1.3.1
 DOCKER_DISTRIB=prog/weaveexec/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
-COVERAGE_MODULES=./common,./common/docker,./ipam,./ipam/paxos,./ipam/ring,./ipam/space,./net,./net/address,./router
+COVERAGE_MODULES=$(shell go list -f '{{join .Deps "\n"}}' ./prog/weaver | grep "weaveworks" | paste -s -d,)
 NETGO_CHECK=@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
 	rm $@; \
 	echo "\nYour go standard library was built without the 'netgo' build tag."; \
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,14 @@ WEAVEEXEC_DOCKER_VERSION=1.3.1
 DOCKER_DISTRIB=prog/weaveexec/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
 COVERAGE_MODULES=./common,./common/docker,./ipam,./ipam/paxos,./ipam/ring,./ipam/space,./net,./net/address,./router
+NETGO_CHECK=@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
+	rm $@; \
+	echo "\nYour go standard library was built without the 'netgo' build tag."; \
+	echo "To fix that, run"; \
+	echo "    sudo go clean -i net"; \
+	echo "    sudo go install -tags netgo std"; \
+	false; \
+}
 
 all: $(WEAVE_EXPORT)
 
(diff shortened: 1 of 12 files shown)


test/140_weave_local_test.sh
```diff
@@ -4,15 +4,15 @@
...
-run_on $HOST1 sudo weave --local launch --ipalloc-range 10.2.5.0/24
+run_on $HOST1 sudo COVERAGE=$COVERAGE weave --local launch --ipalloc-range 10.2.5.0/24
 assert_raises "docker_on $HOST1 ps | grep weave"
 
-run_on $HOST1 sudo weave --local run 10.2.6.5/24 -ti --name=c1 $SMALL_IMAGE /bin/sh
+run_on $HOST1 sudo COVERAGE=$COVERAGE weave --local run 10.2.6.5/24 -ti --name=c1 $SMALL_IMAGE /bin/sh
 assert_raises "exec_on $HOST1 c1 $CHECK_ETHWE_UP"
 
-run_on $HOST1 sudo weave --local run             -ti --name=c2 $SMALL_IMAGE /bin/sh
+run_on $HOST1 sudo COVERAGE=$COVERAGE weave --local run             -ti --name=c2 $SMALL_IMAGE /bin/sh
```
rade: I'd be tempted to introduce a `weave_local_on`, analogous to `weave_on`.

test/config.sh
```diff
@@ -196,6 +205,19 @@ start_suite() {
...
     whitely assert_end
+    for host in $HOSTS; do
+        weave_stop_router $host
+    done
+}
+
+collect_coverage() {
+    host=$1
+    rm -f cover.router.prof
+    docker_on $host cp weave:/home/weave/cover.router.prof . || true
```
paulbellamy: Why not copy this directly to the coverage dir/name?

Makefile
```diff
@@ -43,7 +44,25 @@ travis: $(EXES)
...
+endif
+	@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
+		rm $@; \
+		echo "\nYour go standard library was built without the 'netgo' build tag."; \
+		echo "To fix that, run"; \
+		echo "    sudo go clean -i net"; \
+		echo "    sudo go install -tags netgo std"; \
+		false; \
+	}
+
```
rade: The duplication with the code below is annoying. @dpw can you think of quick way of reducing that?

Makefile
```diff
@@ -35,6 +35,7 @@ WEAVE_EXPORT=weave.tar
 WEAVEEXEC_DOCKER_VERSION=1.3.1
 DOCKER_DISTRIB=prog/weaveexec/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
+COVERAGE_MODULES=./common,./common/docker,./ipam,./ipam/paxos,./ipam/ring,./ipam/space,./net,./net/address,./router
```
rade: This list is rather prone to become outdated. Can we compute it?

test/config.sh
tomwilkie: docker cp takes a directory.  Its painful.

testing/cover/cover.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"fmt"
+	"os"
+	"sort"
+
+	"golang.org/x/tools/cover"
+)
+
+func merge(p1, p2 *cover.Profile) *cover.Profile {
```
paulbellamy: Are overlapping blocks impossible?

testing/util.go
```diff
@@ -50,3 +51,17 @@ func RunWithTimeout(t *testing.T, d time.Duration, f func()) {
 	}()
 	<-ch
 }
+
+// TrimTestArgs finds the first -- in os.Args and trim all args before that
+func TrimTestArgs() {
```
paulbellamy: So, `flag.Args()` ?

Makefile
tomwilkie: We can computer all directories with go files in, but that would include the proxy and the nameserver.  Can't think of a way of computing all modules used by the router.

paulbellamy:
It would be nice not to have a bunch of test gubbins littered around the `weave` script.

Makefile
paulbellamy: `go list -f '{{join .Deps "\n"}}' ./router | grep "weaveworks"`

rade: Could we at least only list top-level dirs? @dpw no doubt can come up with a nifty one liner to turn such a list into a comma-separate list of all those dirs and their sub-dirs.

testing/cover/cover.go
tomwilkie: Not that I'm aware off.  Same source should produce same blocks.

tomwilkie:
@paulbellamy re:weave script, yeah I agree, but I can't think of a better way.  Any ideas?

Makefile
dpw:
```
space:=$(subst x, ,x)
comma:=,
COVERAGE_MODULES=$(subst $(space),$(comma),$(shell find ./common ./ipam ... -type d))
```

(I'll leave implementing `find -type d` in pure make as an exercise for the reader.)

testing/cover/cover.go
paulbellamy: But what about sub-blocks?

testing/util.go
tomwilkie: Yeah I wish - for some reason it trims a little too much off.  I could investigate more if you like?

testing/cover/cover.go
tomwilkie: None of the go coverage code I've read would indicate that, but I'm not 100% sure TBH.

weave
```diff
@@ -1037,6 +1042,10 @@ launch_router() {
             exit 1
         fi
     fi
+    COVERAGE_ARGS=""
+    if [ -n "$COVERAGE" ] ; then
+        COVERAGE_ARGS="-test.coverprofile=/home/weave/cover.router.prof --"
+    fi
```
rade: I'd put this code in the top-level, and then in `ask_version` reset `COVERAGE_ARGS=` if the container is not the router. Removes the duplication.

PR #1110, opened 2015-07-10
[proxy] rewrote chunked response handler

### LINKED ISSUE #1103: Weave proxy unexpectedly closes the connection in chunked responses
In an ECS-optimized AMI in Amazon's EC2, `amazon-ecs-init prestart` fails when using the weave proxy (`DOCKER_HOST=unix:///vart/run/weave.sock`):

```
[root@ip-XXX-XX-XX-XX ec2-user]# DOCKER_HOST=unix:///var/run/weave.sock /usr/libexec/amazon-ecs-init pre-start
2015-07-08T16:25:11Z [INFO] pre-start
2015-07-08T16:25:11Z [ERROR] could not check if Agent is loaded: persistent connection closed
```

However, it works just fine when using Docker directly. 

After further inspection it seems like weave proxy is sending a `Connection: close` header.

Using Weave Proxy:

```
GET /v1.15/images/json?all=1 HTTP/1.1
Host:
User-Agent: go-dockerclient

HTTP/1.1 200 OK
Connection: close
Content-Type: application/json
Date: Wed, 08 Jul 2015 16:07:12 GMT
Transfer-Encoding: chunked
```

Using Docker directly:

```
GET /v1.15/images/json?all=1 HTTP/1.1
Host:
User-Agent: go-dockerclient

HTTP/1.1 200 OK
Content-Type: application/json
Date: Wed, 08 Jul 2015 16:09:01 GMT
Transfer-Encoding: chunked
```

In case it helps, here's what seems to be the code making the  `/v1.15/images/json?all=1`  request: https://github.com/aws/amazon-ecs-init/blob/master/ecs-init/docker/docker.go#L53

paulbellamy:
1) We cannot send "Connection: close", because the fsouza docker client
   expects the tcp socket to stay open between requests.

2) Because we cannot force-close the connection, we can't hijack the
   connection (because go's net/http doesn't let use un-hijack it).

3) Because we need to maintain the individual chunking of messages (for
   docker-py), we can't just copy the response body, as Go will remove and
   re-add the chunking willy-nilly.

Therefore, we have to read each chunk one-by-one, and flush the
ResponseWriter after each one.

Fixes #1103

Seems like quite a large change for a patch release. So, should probably consider merging into master instead.

### DIFF
--- a/proxy/chunked.go
+++ b/proxy/chunked.go
@@ -1,0 +1,202 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// The wire protocol for HTTP's "chunked" Transfer-Encoding.
+
+// Package internal contains HTTP internals shared by net/http and
+// net/http/httputil.
+package proxy
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+)
+
+const maxLineLength = 4096 // assumed <= bufio.defaultBufSize
+
+var ErrLineTooLong = errors.New("header line too long")
+
+// NewChunkedReader returns a new chunkedReader that translates the data read from r
+// out of HTTP "chunked" format before returning it.
+// The chunkedReader returns io.EOF when the final 0-length chunk is read.
+//
+// NewChunkedReader is not needed by normal applications. The http package
+// automatically decodes chunking when reading response bodies.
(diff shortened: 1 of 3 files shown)


rade:
reviewing this would be easier if there was a commit that introduces the bits of code from `net/http/internal` *unaltered*, and then a separate commit that changes them.

paulbellamy:
Going to change this to merge into master instead of 1.0, as it is a large change.

PR #1111, opened 2015-07-10
Mitigate entropy depletion

### LINKED ISSUE #1037: entropy depletion DoS attack
The following potential problem has been identified by a user:

> Weave's session key generation is currently prone to a trivial entropy exhaustion attack. I can send 7 bytes and have Weave read 32 bytes from /dev/urandom, due to GenerateKeyPair using crypto/rand's Reader directly. Consider a cprng, or rate limiting the incoming connections beyond what you have now. It is possible to reduce the entropy of a machine with an exposed Weave node to 10% of normal levels with minimal effort.

awh:
This PR adds adaptive connection acceptance rate limiting in the form of a lazy token bucket implementation, replacing the constant 1ms sleep in use before. The bucket contains a maximum of 100 tokens, replenished at a rate of one every 100ms; this allows a burst of up to 100 connection attempts without any delay, limiting to 10 per second thereafter under sustained load. Quiescence allows the bucket to refill back to its maximum, the intended effect being that initial network formation and partition healing are not hindered at all, but a sustained denial of service will be.

Testing reveals that this does not address the reported issue entirely - even at the lower bound of 10 requests per second there may not be enough entropy gathered on a headless system to keep pace. Nevertheless, after careful investigation and research, we have concluded that of the two options available to us (the other being implementing and seeding our own CSPRNG in a similar fashion to OpenSSL) this is by far the safest. We have been convinced that the impact of such an attack is limited to slowing `/dev/random` readers, so we have recorded our rationale along with links to supporting expert opinion in our cryptography design document.

In closing, it is noteworthy that OpenSSH can be used to deplete entropy in exactly the same fashion, as each forked process consumes 256 bits from `/dev/urandom`.

Closes #1037.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -14,16 +14,11 @@ import (
 )
 
 const (
-	macMaxAge      = 10 * time.Minute     // [1]
-	tcpAcceptDelay = 1 * time.Millisecond // [2]
+	macMaxAge = 10 * time.Minute // [1]
 )
 
 // [1] should be greater than typical ARP cache expiries, i.e. > 3/2 *
 // /proc/sys/net/ipv4_neigh/*/base_reachable_time_ms on Linux
-
-// [2] time to wait between accepting tcp connections. This guards
-// against brute-force attacks on the password when encryption is in
-// use. It is also a basic DoS defence.
 
 type LogFrameFunc func(string, []byte, *EthernetDecoder)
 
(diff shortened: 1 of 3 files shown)


awh:
@dpw @rade would appreciate a review - thankyou

awh:
@dpw I've reworked your example to fix the drift issue and teased things apart a little to improve readability.

awh:
Moving to 1.0 branch.

PR #1112, opened 2015-07-10
[proxy] rewrote chunked response handler

### LINKED ISSUE #1103: Weave proxy unexpectedly closes the connection in chunked responses

paulbellamy:
1) We cannot send "Connection: close", because the fsouza docker client
expects the tcp socket to stay open between requests.

2) Because we cannot force-close the connection, we can't hijack the
connection (because go's net/http doesn't let use un-hijack it).

3) Because we need to maintain the individual chunking of messages (for
docker-py), we can't just copy the response body, as Go will remove and
re-add the chunking willy-nilly.

Therefore, we have to read each chunk one-by-one, and flush the
ResponseWriter after each one.

Fixes #1103

Merging into master instead of 1.0, as it is definitely not a "patch" change. Replacing #1110.

### DIFF
--- a/proxy/proxy_intercept.go
+++ b/proxy/proxy_intercept.go
@@ -2,16 +2,29 @@ package proxy
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"strconv"
 
 	"github.com/fsouza/go-dockerclient"
 	. "github.com/weaveworks/weave/common"
+)
+
+const (
+	maxLineLength = 4096 // assumed <= bufio.defaultBufSize
+	maxChunkSize  = bufio.MaxScanTokenSize
+)
+
+var (
+	ErrChunkTooLong           = errors.New("chunk too long")
+	ErrInvalidChunkLength     = errors.New("invalid byte in chunk length")
+	ErrLineTooLong            = errors.New("header line too long")
+	ErrMalformedChunkEncoding = errors.New("malformed chunked encoding")
 )
 
 func (proxy *Proxy) Intercept(i interceptor, w http.ResponseWriter, r *http.Request) {
@@ -71,7 +84,7 @@ func (proxy *Proxy) Intercept(i interceptor, w http.ResponseWriter, r *http.Requ
 }
 
 func doRawStream(w http.ResponseWriter, resp *http.Response, client *httputil.ClientConn) {
-	down, downBuf, up, rem, err := hijack(w, client)
+	down, downBuf, up, remaining, err := hijack(w, client)
 	if err != nil {
 		http.Error(w, "Unable to hijack connection for raw stream mode", http.StatusInternalServerError)
 		return
@@ -96,7 +109,7 @@ func doRawStream(w http.ResponseWriter, resp *http.Response, client *httputil.Cl
 
 	upDone := make(chan struct{})
 	downDone := make(chan struct{})
-	go copyStream(down, io.MultiReader(rem, up), upDone)
+	go copyStream(down, io.MultiReader(remaining, up), upDone)
 	go copyStream(up, downBuf, downDone)
 	<-upDone
 	<-downDone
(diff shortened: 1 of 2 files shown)


proxy/proxy_intercept.go
```diff
@@ -117,29 +116,29 @@ func copyStream(dst io.Writer, src io.Reader, done chan struct{}) {
...
+	var err error
+	chunks := NewChunkedReader(io.MultiReader(rem, up))
+	for chunks.Next() && err == nil {
+		_, err = io.Copy(w, chunks.Chunk())
+		wf.Flush()
+	}
+	if err == nil {
+		err = chunks.Err()
+	}
+	if err != nil {
```
tomwilkie: Else?

rade: that would behave differently.

tomwilkie: Ah yes sorry.

proxy/proxy_intercept.go
```diff
@@ -117,29 +116,29 @@ func copyStream(dst io.Writer, src io.Reader, done chan struct{}) {
...
-	// stopping at the end of the chunked section.
-	rawResponseBody := io.MultiReader(rem, up)
-	if _, err := io.Copy(ioutil.Discard, httputil.NewChunkedReader(io.TeeReader(rawResponseBody, down))); err != nil {
-		http.Error(w, "Error copying chunked response body", http.StatusInternalServerError)
-		return
+
+	var err error
+	chunks := NewChunkedReader(io.MultiReader(rem, up))
+	for chunks.Next() && err == nil {
+		_, err = io.Copy(w, chunks.Chunk())
```
peterbourgon: Probably better to `io.Copy` into the `wf` for consistency's sake.

proxy/proxy_intercept.go
```diff
@@ -117,29 +116,29 @@ func copyStream(dst io.Writer, src io.Reader, done chan struct{}) {
...
-		http.Error(w, "Unable to hijack response stream for chunked response", http.StatusInternalServerError)
+	wf, ok := w.(http.Flusher)
+	if !ok {
+		http.Error(w, "Error forwarding chunked response body: flush not available", http.StatusInternalServerError)
 		return
 	}
+
+	w.WriteHeader(resp.StatusCode)
+
+	up, rem := client.Hijack()
```
peterbourgon: I guess `rem` here means.. remaining?

proxy/chunked_test.go
```diff
@@ -0,0 +1,106 @@
...
+		e := "malformed chunked encoding"
+		if r.Err() == nil || r.Err().Error() != e {
+			t.Errorf("chunk reader errored %q; want %q", r.Err(), e)
+		}
+		data, err := ioutil.ReadAll(r.Chunk())
+		if len(data) != 0 {
+			t.Errorf("chunk should have been empty. got %q", string(data))
+		}
+		if err != nil {
+			t.Logf(`data: "%s"`, data)
```
peterbourgon:
```go
t.Logf("data: %q", data)
```

proxy/chunked_test.go
```diff
@@ -0,0 +1,106 @@
...
+		if r.Next() {
+			t.Errorf("Expected failure when reading chunks, but got one")
+		}
+		e := "malformed chunked encoding"
+		if r.Err() == nil || r.Err().Error() != e {
+			t.Errorf("chunk reader errored %q; want %q", r.Err(), e)
+		}
+		data, err := ioutil.ReadAll(r.Chunk())
+		if len(data) != 0 {
+			t.Errorf("chunk should have been empty. got %q", string(data))
```
peterbourgon: I don't think you need the `string(...)` here?

proxy/chunked.go
```diff
@@ -0,0 +1,146 @@
...
+			err = io.ErrUnexpectedEOF
+		} else if err == bufio.ErrBufferFull {
+			err = ErrLineTooLong
+		}
+		return nil, err
+	}
+	if len(p) >= maxLineLength {
+		return nil, ErrLineTooLong
+	}
+	return trimTrailingWhitespace(p), nil
```
peterbourgon: A reason you don't use one of these options?

```go
strings.TrimSpace(p)
strings.TrimRight(p, " \n\t\r")
```

squaremo:
To check my understanding of what's changed: we no longer hijack the downstream connection to the client, just the upstream connection to docker. Docker is fine with the connection being closed arbitrarily (or we are assuming that); some clients are not.

Is there a reason we have to parse the hijacked upstream, rather than using ChunkedReader as before?

> 1) We cannot send "Connection: close", because the fsouza docker client
> expects the tcp socket to stay open between requests.

Just to be clear, this is a bug in that client, right? It should deal with the HTTP connection being closed arbitrarily.

paulbellamy:
> Docker is fine with the connection being closed arbitrarily (or we are assuming that); some clients are not.

Yes

> Is there a reason we have to parse the hijacked upstream, rather than using ChunkedReader as before?

ChunkedReader does not expose individual chunks. It just smashes them together into a bytestream. The downstream client (docker-py in particular) expects json objects and http chunks to align.

> Just to be clear, this is a bug in that client, right? It should deal with the HTTP connection being closed arbitrarily.

Yes

squaremo:
> Is there a reason we have to parse the hijacked upstream, rather than using ChunkedReader as before?
> 
> ChunkedReader does not expose individual chunks. It just smashes them together into a bytestream. The downstream client (docker-py in particular) expects json objects and http chunks to align.

Ahh I beg your pardon, I had it the wrong way around. The ChunkedReader was to gate the straight-through byte copy wasn't it.

paulbellamy:
`net/http/httputil.ChunkedReader` just smashes all the bytes together, yes. Our scanner lets us flush after each chunk.

PR #1115, opened 2015-07-10
Run unit tests in circle and include their coverage in the report.  

tomwilkie:
Report coveralls.io from Circle.

### DIFF
--- a/test/units.sh
+++ b/test/units.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+go get github.com/weaveworks/weave/testing/cover
+
+if [ -n "$COVERDIR" ] ; then
+	coverdir="$COVERDIR"
+else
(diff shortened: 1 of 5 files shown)


test/units.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+go get github.com/weaveworks/weave/testing/cover
```
rade: It seems very strange to refer to the weave gh repo here. Why can't this refer to the local version?

tomwilkie: My understanding is that go get will use the local code, and install the
binary in you bin dir.

rade: > My understanding is that go get will use the local code

Yes, that is my understanding too. But it still looks weird.

tomwilkie: I'll add a build target for it then.

PR #1120, opened 2015-07-12
Various improvements to unit test running

### LINKED ISSUE #998: add race detector Makefile target and run that on CI

tomwilkie:
- units.sh should exit on error.
- Add race detector to unit test runner (fixes #998)
- Use local testing/cover/cover tool in gen_coverage_report.sh and units.sh
- Use the scheduler to parallelise the unit tests.
- Make test run id a string and differentiate between unit runs and integration runs in the scheduler.
- Pipe circle env vars through to build container so we can detect when to use the scheduler
- Add python requests to the weave build container.
- Quote test names in urls, make flask handle url segments with slashes in.


test/gen_coverage_reports.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-set -ex
+set -eux
```
rade: please don't add 'u'. I took it out because otherwise I cannot run `gen_coverage` locally since it barfs about undefined $CIRCLECI.

tomwilkie:
Sorry, bad merge.  I've removed it now.

test/units.sh
```diff
@@ -1,28 +1,58 @@
...
-mkdir -p $coverdir
 fail=0
 
-for dir in $(find . -type f -name '*_test.go' | xargs -n1 dirname | grep -v prog | sort -u); do
-	go get -t -tags netgo $dir
-	output=$(mktemp $coverdir/unit.XXXXXXXXXX)
-	if ! go test -cpu 4 -tags netgo -covermode=count -coverprofile=$output $dir ; then
-		fail=1
-	fi
+TESTDIRS=$(find . -type f -name '*_test.go' | xargs -n1 dirname | grep -v prog | sort -u)
```
awh: `TESTDIRS=$(find . -type f -name '*_test.go' -exec dirname {} \; | grep -v prog | sort -u)` is slightly more efficient

dpw:
Why is the weave-build container being overloaded to run tests? If you want a container to run tests, then introduce one for that purpose.  And layer it on top of a common base with weave-build, if that makes sense.  But weave-build is for building weave, and only for that.  Please clean this up, and copy me on PRs that touch weave-build in future.

PR #1125, opened 2015-07-13
Upgrade smoke-tests to use 'overlay' storage and Ubuntu 15.04

bboreham:
The 'overlay' driver is [claimed](http://jpetazzo.github.io/assets/2015-03-03-not-so-deep-dive-into-docker-storage-drivers.html#57) to be the best choice right now, but requires Linux kernel 3.18.

You can feel the difference - using `vfs` (which our Vagrant set-up did because it didn't install the extra bits for `aufs`) the machine would bog down when tests were running, but with `overlay` you rarely notice this.

I will post some timings to illustrate the difference soon.

### DIFF
--- a/test/Vagrantfile
+++ b/test/Vagrantfile
@@ -14,10 +14,10 @@ require '../vagrant-common.rb'
 def configure_docker(host, hostname, ip)
   pkgs = %w(lxc-docker ethtool)
 
-  host.vm.box = "ubuntu/ubuntu-14.10-amd64"
-  host.vm.box_url = "https://cloud-images.ubuntu.com/vagrant/utopic/current/utopic-server-cloudimg-amd64-vagrant-disk1.box"
+  host.vm.box = "ubuntu/ubuntu-15.04-amd64"
+  host.vm.box_url = "https://cloud-images.ubuntu.com/vagrant/vivid/current/vivid-server-cloudimg-amd64-vagrant-disk1.box"
 
-  host.vm.hostname = hostname
+  host.vm.provision :shell, :inline => "hostnamectl set-hostname "+hostname
   host.vm.network "private_network", ip: ip
 
   host.vm.synced_folder ".", "/vagrant", disabled: true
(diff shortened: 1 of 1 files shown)


tomwilkie:
We should do the same for circle runs. Needs updates to gce.sh and a new
template generating.

rade:
it's less important for circle since that at least uses aufs.

bboreham:
Sample timings for `make run-smoketests`, after all images have been pre-downloaded:

Driver | Run 1 | Run 2
---------|----------|----------
VFS (current) | 23m47s
AUFS | 8m45s | 5m48s
Overlay | 6m19s | 6m33s

Not sure if this is conclusive between AUFS and Overlay.  Maybe the 8-minute one was an outlier.

errordeveloper:
LGTM

2opremio:
This change breaks the smoke-tests (in my local installation at least).

Using current master (907de67a4044b72fc2a9935ddf0ef80a78ebf222), after getting excited about the performance benefits of this PR and recreating the test VMs:

    weave/test $ ./setup.sh
    Copying weave images, scripts, and certificates to hosts, and
    prefetch test images
    Post http://192.168.48.11:2375/v1.19/images/load: dial tcp 192.168.48.11:2375: i/o timeout. Are you trying to connect to a TLS-enabled daemon without TLS?

It seems like `192.168.48.11:2375` cannot be accessed externally anymore when switching to ubuntu 15.

From my host:

    $ docker -H tcp://192.168.48.11:2375 ps
    Cannot connect to the Docker daemon. Is 'docker -d' running on this host?`

From my vagrant dev machine:

     $ docker -H tcp://192.168.48.11:2375 ps
     Get http://192.168.48.11:2375/v1.19/containers/json: dial tcp 192.168.48.11:2375: i/o timeout. Are you trying to connect to a TLS-enabled daemon without TLS?



Reverting to 499d65e7158987a40a5895f3a6dd6afd39e6c03f and recreating the VMs fixes it

bboreham:
Sorry, there's a fix in 6d17fe5df5e663d1cf693a17405eb92da60002f8

PR #1126, opened 2015-07-13
[proxy] More flexible container name -> hostname derivation

### LINKED ISSUE #1018: more flexible container name -> hostname derivation
See the discussion in #1006.

The idea proposed there is to derive the hostname from the container name in the proxy via a configurable regexps. Defaults to `".*"` and could be set to `"^$"` for disabling hostname derivation (i.e. if the regexps matches nothing then we shouldn't set the hostname). Should support sub-matches, e.g. `"aws-[0-9]+-\(.*\)"`.

We *could* have an env-var equivalent for `weave run`. Overkill?

2opremio:
Provides a new "--hostname=substitution" flag to WeaveProxy, where
'substitution' is a sed-style substitution command. 'substitution' will be
applied to container names at launch-time, as a means to control what hostnames will
be registered in WeaveDNS.

The substitution is implemented with
https://golang.org/pkg/regexp/#Regexp.ReplaceAllString . Thus, references to
regexp substitution groups should be prepended with a dollar sign.

For instance, if we provide "--hostname='/aws-[0-9]+-(.*)/my-app-$1/'", running
a container named 'aws-12798186823-foo' through WeaveProxy will lead to WeaveDNS
registering 'my-app-foo'.

Closes #1018


tomwilkie:
we'll need a few lines in the docs too, please.

2opremio:
@tomwilkie Sure, I considered it but it wasn't clear to me were to include it.  https://github.com/weaveworks/weave/tree/master/docs only seems to include architectural stuff.

2opremio:
OK, I found `/site` which I hadn't payed attention to until now. I will add something there first thing tomorrow.

2opremio:
@tomwilkie Done, I will squash the commit once the PR is good to go.

2opremio:
Rebased with master (and squashed the documentation commit) to make coveralls happy

tomwilkie:
You should fetch the coverage.txt from your branch and diff it with master to see where this -5% is coming from.

weave
```diff
@@ -36,7 +36,7 @@ weave launch-router [--password <password>] [--nickname <nickname>]
                       [--no-discovery] [--init-peer-count <count>] <peer> ...
 weave launch-dns    [<addr>]
 weave launch-proxy  [-H <endpoint>] [--with-dns | --without-dns]
-                      [--no-default-ipalloc] [--no-rewrite-hosts]
+                      [--no-default-ipalloc] [--no-rewrite-hosts] [--hostname]
```
rade: That should be `--hostname <figureoutwhattoputhere>`

2opremio:
@tomwilkie Thanks, will do. I suspect it's coming from `proxy/` not having any tests until now.

proxy/proxy.go
```diff
@@ -15,20 +16,25 @@ import (
...
+	substitutionReplacementName = "replacement"
 )
 
 var (
-	containerCreateRegexp = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/create$")
-	containerStartRegexp  = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/[^/]*/(re)?start$")
-	execCreateRegexp      = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/[^/]*/exec$")
+	containerCreateRegexp = regexp.MustCompile(`^(/v[0-9\.]*)?/containers/create$`)
+	containerStartRegexp  = regexp.MustCompile(`^(/v[0-9\.]*)?/containers/[^/]*/(re)?start$`)
+	execCreateRegexp      = regexp.MustCompile(`^(/v[0-9\.]*)?/containers/[^/]*/exec$`)
```
rade: so those regexps were wrong previously? e.g. they would match `/v\/foo`? If so, raise a bug and fix on the 1.0 branch.

proxy/proxy_test.go
```diff
@@ -0,0 +1,59 @@
...
+	if replacementStr != "my/app/$1" {
+		t.Error("Unexpected replacement string", replacementStr)
+	}
+
+	hostname := patternRegexp.ReplaceAllString("aws/17651276812/name", replacementStr)
+	if hostname != "my/app/name" {
+		t.Error("Unexpected replacement hostname", hostname)
+	}
+
+}
```
rade: use `testify`'s `require`, as we do elsewhere in our unit tests.

proxy/proxy.go
2opremio: I don't think there were wrong, I just wanted to use backquotes for all of them (for uniformity). Maybe it's my lack of golang knowledge, but aren't they equivalent? (as I understand backquotes don't require escaping) http://play.golang.org/p/SUiW2R0B_P

proxy/proxy.go
```diff
@@ -119,6 +132,31 @@ func (proxy *Proxy) ListenAndServe() {
...
+	namedSubmatches := findStringNamedSubmatch(substitutionRegexp, hostname)
+	if len(namedSubmatches) < 1 {
+		errStr := fmt.Sprintf("Incorrect hostname expression: %s", hostname)
+		return nil, "", errors.New(errStr)
+	}
+
+	escapedPattern := namedSubmatches[substitutionPatternName]
+	escapedReplacement := namedSubmatches[substitutionReplacementName]
+	unescapedPattern := unescapeBackSlashes(escapedPattern)
+	unescapedReplacement := unescapeBackSlashes(escapedReplacement)
```
rade: I'd just make this
````
pattern := unescapeBackSlashes(namedSubmatches[substitutionPatternName])
replacement :=  unescapeBackSlashes(namedSubmatches[substitutionReplacementName])
````

File: proxy/proxy.go (hunk 2 above)
Comment (rade) [inline proxy/proxy.go]: Ah, I completely missed quote->backquote change.

File: proxy/proxy.go [hunk 5]
```diff
@@ -119,6 +132,31 @@ func (proxy *Proxy) ListenAndServe() {
...
+
+	patternRegexp, err := regexp.Compile(unescapedPattern)
+	if err != nil {
+		errStr := fmt.Sprintf("Incorrect hostname pattern '%s': %s", escapedPattern, err.Error())
```
paulbellamy: You probably mean `hostname` not `escapedPattern` here.

proxy/proxy.go
2opremio: I, at the very least,  need `escapedPattern` to report an error if `unescapedPattern` doesn't compile (read below).

proxy/proxy.go
2opremio: My intention is to show `escapedPattern` to the user, which is the specific part of the substitution command which failed to compile. I could show the full hostname, but `Compile`'s error will refer exclusively to the pattern.

Thanks about `fmt.Errorf`, I will use it instead.

proxy/proxy.go
rade: fair enough; leave as is then.

tomwilkie:
Alternatively, instead of requireing a sed-style regex in the for /foo/bar/, how about we just have --hostname-match=foo and --hostname-replacement=bar?  It would save us a bunch of code (using regexes to parse regexes...)  Default for --hostname-match should be (.*) and --hostname-replacement should be $1

Also, I wouldn't bother including this in the weave usage text; this feature is meant for integrators only.  Just include it in the site/ docs.

2opremio:
@tomwilkie I personally like the `/regexp/substitution/` better, @rade also seems to agree https://github.com/weaveworks/weave/issues/1018 

Also, I am not sure it makes sense not to document it in usage if it's in the docs. If we want it to be hidden from normal users then it shouldn't appear in the site.

tomwilkie:
@rade what do you think?  I suggested this as I'm not sure the parsing of the sed-style regexps is worth the extra code.

proxy/proxy_test.go
2opremio: Done

2opremio:
I pushed a commit which addressed all the code comments (except for @tomwikie 's  design remark). Also, I pushed a change to surround relevant regexps with `^` and `$`. 

Please make sure to squash the changes before merging.

test/680_proxy_hostname_derivation.sh
```diff
@@ -0,0 +1,36 @@
...
+weave_on $HOST1 launch-dns 10.2.254.1/24
+weave_on $HOST1 launch-proxy --hostname '/^[^-]+-(?P<appname>[^-]*)-[^-]+$/$appname/'
+
+start_container -e WEAVE_CIDR=$C1/24 --name=$CNAME1
+start_container -e WEAVE_CIDR=$C2/24 --name=$CNAME2
+start_container -e WEAVE_CIDR=$C3/24 --name=$CNAME3
+
+check() {
+  assert "proxy exec_on $HOST1 $1 getent hosts $2 | tr -s ' '" "$3 $2"
+}
```
rade: I've just merged #1050, which extends `assert_dns_a_record` to do what you want here.

rade:
>  I'm not sure the parsing of the sed-style regexps is worth the extra code.

Agreed, and I am not sold on the UI benefits either.

So two options it is, with the defaults as you suggest. But I do think they should be documented in the usage.

2opremio:
OK, I will rework the PR to use those two options.

test/680_proxy_hostname_derivation.sh
2opremio: Will do

2opremio:
@rade @paulbellamy @tomwilkie please review

proxy/proxy.go
```diff
@@ -15,11 +15,13 @@ import (
...
-	defaultCertFile = "cert.pem"
-	dockerSock      = "/var/run/docker.sock"
-	dockerSockUnix  = "unix://" + dockerSock
+	defaultCaFile               = "ca.pem"
+	defaultKeyFile              = "key.pem"
+	defaultCertFile             = "cert.pem"
+	dockerSock                  = "/var/run/docker.sock"
+	dockerSockUnix              = "unix://" + dockerSock
+	substitutionPatternName     = "pattern"
+	substitutionReplacementName = "replacement"
```
paulbellamy: These are unused now

weave
```diff
@@ -37,6 +37,7 @@ weave launch-router [--password <password>] [--nickname <nickname>]
 weave launch-dns    [<addr>]
 weave launch-proxy  [-H <endpoint>] [--with-dns | --without-dns]
                       [--no-default-ipalloc] [--no-rewrite-hosts]
+                      [[--hostname-match <regexp>] [--hostname-replacement <replacement>]]
```
rade: no need for the double brackets; the options can be set independently.

proxy/proxy.go
```diff
@@ -29,19 +31,22 @@ var (
...
+	WithoutDNS          bool
 }
 
 type Proxy struct {
 	Config
-	client         *docker.Client
-	dockerBridgeIP string
+	client              *docker.Client
+	dockerBridgeIP      string
+	hostnameMatchRegexp *regexp.Regexp
```
paulbellamy: Where's `HostnameReplacement`? How does this compile?

site/proxy.md
```diff
@@ -130,6 +130,35 @@ automatically if it is running at the point when they are started -
 see the [weaveDNS usage](weavedns.html#usage) section for an in-depth
 explanation of the behaviour and how to control it.
 
+Typicaly, the proxy will pass on container names as-is to [weaveDNS](weavedns.html)
```
paulbellamy: Typically

site/proxy.md
```diff
@@ -130,6 +130,35 @@ automatically if it is running at the point when they are started -
...
+for registration. However, there are situations in which the final container
+name is out of the user's control (e.g. when using Docker orchestrators which
+append control/namespacing identifiers to the original container names).
+
+For those situations, the proxy provides two flags: `--hostname-match <regexp>`
+and `--hostname-replacement <replacement>`. When launching a container, its name
+matched against regular expression `<regexp>`. Then, based on that match,
```
paulbellamy: is matched

2opremio:
@tomwilkie It seems there are GCP quota problems in circleci https://circleci.com/gh/weaveworks/weave/1835

tomwilkie:
Yes I've just cleared out some stale VMs.  VMs aren't cleaned up if you
click cancel.

weave
```diff
@@ -37,6 +37,7 @@ weave launch-router [--password <password>] [--nickname <nickname>]
 weave launch-dns    [<addr>]
 weave launch-proxy  [-H <endpoint>] [--with-dns | --without-dns]
                       [--no-default-ipalloc] [--no-rewrite-hosts]
+                      [--hostname-match <regexp>] [--hostname-replacement <replacement>]
```
rade: sorry, but `weave` usage needs to fit into 80 columns.

2opremio:
@rade Thanks. Out of curiosity, is there a reason why fix-commits are not squashed before merging? (Commits like ded3d6b40bdab5871b86731bd28cfdb0c3a67fe5 don't make any sense out of context) I added extra commits to make clear what was being changed, with the intention of squashing them before merging. Should I had kept amending instead?

rade:
I didn't realise you wanted to squash some commits. Better leave a note in the PR next time.

2opremio:
@rade Sorry, I was expecting the merger to check the commits before going ahead. This is probably not the forum for this discussion but, to avoid these situations, wouldn't it make sense to let the author merge the code himself after getting a "go ahead"?

rade:
> I was expecting the merger to check the commits before going ahead.

I did. I'm not bothered by the non-squashed commits; There is enough context to trace them back to this PR.

> This is probably not the forum for this discussion

Correct.

2opremio:
@rade ok, fair enough :)

PR #1131, opened 2015-07-14
Quarantine Prototype

tomwilkie:
An early prototype of how quarantine might work. 

Questions:
- quarantine / unquarantine are hard to spell, type and just don't seem right.  Please suggest alternatives
- currently one can quarantine peers and containers.  To quarantine a peer/container means to temporarily remove the container from WeaveDNS.  We could extend this to also mean remove it from routing... thoughts?
- quarantines are records, <s>gossiped</s> phone tree'd around the cluster.  I went for this over just tombstoning the DNS entry so the burden of remembering the original mapping was not placed on the caller.
- I would like to extend the quarantine code to things like "all endpoints for foo.weave.local in the bar datacenter".  Weave would need to know about DCs, but this could map nicely to EC2.  Thoughts?

Example usage:
```
$ docker run -di --hostname foo.weave.local ubuntu /bin/bash
f18a16ce6e4c4882f38f3f44641ae2a2dc1468196b66829c5bc00b89495cb333
$ nslookup foo.weave.local 172.17.42.1
Server:		172.17.42.1
Address:	172.17.42.1#53

Name:	foo.weave.local
Address: 10.32.0.1

$ ./weave quarantine-container f18a16ce6e4c488 1h
3582c57bed56e2ea
$ nslookup foo.weave.local 172.17.42.1
Server:		172.17.42.1
Address:	172.17.42.1#53

*** Can't find foo.weave.local: No answer

$ ./weave quarantines
ID               Container    Peer              Duration
3582c57bed56e2ea f18a16ce6e4c                   2015-07-14 13:54:38 +0000 UTC

$ ./weave unquarantine 3582c57bed56e2ea
$ nslookup foo.weave.local 172.17.42.1
Server:		172.17.42.1
Address:	172.17.42.1#53

Name:	foo.weave.local
Address: 10.32.0.1
```

TODO:
- [ ] smoke tests
- [ ] delete old entries
- [ ] make unquarantine take a peer / container, and remove all records for that peer / container

### DIFF
--- a/nameserver/http.go
+++ b/nameserver/http.go
@@ -4,11 +4,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/miekg/dns"
 
 	"github.com/weaveworks/weave/net/address"
+	weaverouter "github.com/weaveworks/weave/router"
 )
 
 func (n *Nameserver) badRequest(w http.ResponseWriter, err error) {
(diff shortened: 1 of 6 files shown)


awh:
I've just rebased this on master; the old branch is backed up as `quarantine-backup`.

PR #1142, opened 2015-07-14
Copy dns requests and give them a fresh id when doing recursive queries.

### LINKED ISSUE #1113: Investigate if its safe to reuse request for recursive DNS
I'm not sure you can just send the same request we get from the client to the upstream server. This could lead to a conflict if two clients send a request with the same message ID. I would make a copy...

tomwilkie:
Fixes #1113

### DIFF
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -227,9 +227,9 @@ func (d *DNSServer) handleRecursive(client *dns.Client, defaultMaxResponseSize i
 		}
 
 		for _, server := range d.upstream.Servers {
-			reqCopy := *req
+			reqCopy := req.Copy()
 			reqCopy.Id = dns.Id()
-			response, _, err := client.Exchange(&reqCopy, fmt.Sprintf("%s:%s", server, d.upstream.Port))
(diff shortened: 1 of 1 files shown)


tomwilkie:
@inercia would you mind reviewing please?

nameserver/dns.go
```diff
@@ -227,7 +227,9 @@ func (d *DNSServer) handleRecursive(client *dns.Client, defaultMaxResponseSize i
 		}
 
 		for _, server := range d.upstream.Servers {
-			response, _, err := client.Exchange(req, fmt.Sprintf("%s:%s", server, d.upstream.Port))
+			reqCopy := *req
```
inercia: There is a `Copy()` method for messages...

tomwilkie: Ah good catch!  I was looking for it and I didn't find one.  Do you know where I could find package level docs for the dns stuff?

inercia: https://godoc.org/github.com/miekg/dns ?

inercia:
LGTM. Will merge once the CI tests pass...

nameserver/dns.go
tomwilkie: Sweet, thanks.

PR #1147, opened 2015-07-15
Basic performance test suite

### LINKED ISSUE #1119: Performance test suite
It would be nice to have some performance test suite for the Weave router. The results obtained could be useful for a) optimizing the router, b) tracking performance regressions, c) showing to users the real performance limits of the router 

An initial version of this test suite would launch a couple of iperf containers and get some raw performance numbers. It would really have many things in common with our integration tests, so maybe it should be part of them...

inercia:
Fixes #1119 

This PR adds an initial performance test that runs `iperf` between two hosts. It generates a iperf log that is stored in `test/performance/*.log`. Performance tests can be run with the rest of the integration test by doing a `PERFORMANCE=1 ./run_all.sh`

The underlying objective of this PR is to determine how to deal with performance tests: to determine what we really need, what to do with the results, etc...

How it works:
- Performance tests generate `*.csv` files that are archieved with the rest of the artifacts.
- A Python script tries to get the `*.csv` files from the last _n_ builds for each performance tests from CircleCI.
- The Python script generates a peformance graph with the `*.csv`s
- In contrast with other CI systems like Jenkins, CircleCI cannot provide a constant URL for accessing the latest build (and its artifacts), so we have to store the performance graph in an external service. In this case, perfomance graphs are plotted with the help of [Plotly](www.plotly.com) and stored [here](https://plot.ly/~inercia)

TODO:
- [x] Include the performance tests in the CircleCI builds
- [x] Archive the artifacts (test/performance/*.log)
- [x] Generate some graphs and link them from some place in our docs
- [ ] Create a _weaveworks_ user in Plotly for generating performance graphs. Remove @inercia's credentials (will do this once this is merged in master)
- [ ] Skip branches other than master (will do this once this is merged in master)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@ Vagrantfile.local
 test/tls/*.pem
 test/coverage
 test/coverage.*
+test/performance
--- a/bin/circleci/perf-tests
+++ b/bin/circleci/perf-tests
@@ -1,0 +1,69 @@
+#!/bin/sh
+
+IN_DIR=$SRCDIR/test/performance
+OUT_DIR=$CIRCLE_ARTIFACTS/
+
+export PERFORMANCE=1
+
+###########################################################################
+
+log()   { echo "$1" >&2 ;            }
+abort() { log "ABORT: $1" ; exit 0 ; }
+fatal() { log "FATAL: $1" ; exit 1 ; }
+
+[ -n "$CIRCLE_NODE_INDEX"  ] || fatal "No CIRCLE_NODE_INDEX defined"
+[ $CIRCLE_NODE_INDEX -eq 0 ] || abort "Not executing in CircleCI node $CIRCLE_NODE_INDEX"
+
+case "$1" in
+	pre)
+		log "Installing dependencies..."
+		sudo pip install plotly
+		sudo pip install requests
+	;;
+
+    start)
+		[ -z "$SRCDIR" ] && abort "No SRCDIR defined"
+
+		cd $SRCDIR/test
+		eval $(./gce.sh hosts)
+		$SRCDIR/test/run_all.sh $SRCDIR/test/*_perf.sh
+    ;;
+
+    stop)
+    ;;
+
+    post)
+		[ -n "$CIRCLE_ARTIFACTS" ] || fatal "no artifacts folder in environment"
+		[ -n "$PLOTLY_USERNAME" ]  || fatal "no PlotLy user name in environment"
+		[ -n "$PLOTLY_API_KEY" ]   || fatal "no PlotLy API key in environment"
+		[ -d $IN_DIR ]             || abort "no performance directory found"
+		[ -d $OUT_DIR ]            || { log "creating $OUT_DIR" ; mkdir -p $OUT_DIR ; }
+
+		log "Processing performance logs in $IN_DIR"
+		for csv_file in $IN_DIR/*.csv ; do
+		    [ -f $csv_file ] || continue
+
+		    name=$(basename $csv_file .csv)
+		    name_wo_ext=$(basename $csv_file)
+
+		    log "Generating graph for $name_wo_ext [branch: $CIRCLE_BRANCH]"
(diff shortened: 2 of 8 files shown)


inercia:
@rade @tomwilkie any suggestion on what we could do with the performance tests results? should I generate a graph and upload it to somewhere?

tomwilkie:
You could stick the raw data it in the circle artifacts (see last few lines of circle.yml, this is what we do for the coverage data)

For extra points you could post it to an appengine app that plots results over time...

inercia:
@tomwilkie Would you please take a look at this PR?

bin/circleci/do-coverage
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+[ $CIRCLE_NODE_INDEX -eq 0 ] || exit 0
```
tomwilkie: Please leave all this in the circle.yml, at least for this change.

test/config.sh
```diff
@@ -214,5 +215,47 @@ collect_coverage() {
     mv cover.router.prof $(mktemp -u ./coverage/integration.XXXXXXXX) || true
 }
 
+################################
+# performance utils
+################################
```
tomwilkie: As these are only used by one test, can we put them in that test?  config.sh is getting too big.

bin/circleci/do-perf-graphs
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
```
tomwilkie: Can the functionality of this be rolled into do-perf-graphs.py?

test/config.sh
inercia: I can move this code to a different place, but my plan is to add more performance tests, so they need to be in a common place... `perf.sh` maybe?

bin/circleci/do-coverage
inercia: Done. I think it would be a good idea to avoid the _code-in-YAML_ and move things to this kind of auxiliary scripts...

test/config.sh
tomwilkie: Yes that sounds like a good idea.  I've been meaning to do that for the dns tests.

bin/circleci/do-coverage
tomwilkie: Yes maybe in a future change we could do this; in which case it would probably not be lots of little files, but perhaps one big file with a switch statement.

bin/circleci/do-perf-graphs
inercia: Not sure it is worth the trouble, especially as I would prefer to keep the `[ $CIRCLE_NODE_INDEX -eq 0 ]` check out of `circle.yml` (and the Python script is not a good place either), and iterating over a list of `*.csv`s in Python is annoying...

inercia:
On hold until #1437 is fixed...

rade:
> On hold until #1437 is fixed...

why? It's annoying to have that failure, but it shouldn't stop us from getting work done.

inercia:
Because the only remaining thing is to get the CircleCI build fixed

rade:
> Because the only remaining thing is to get the CircleCI build fixed

I see. That's not what the [TODO](https://github.com/weaveworks/weave/pull/1147#issue-95167510) says.

test/config.sh
inercia: Moved to `perf.sh`.

faddat:
I'd really like to see this merged-- since I  use some pretty unusual topologies that include users, datacenters, mobile devices, and stuff we shall just call "other" having a way to test it all would be great-- in fact, I've got to say that this seems like an easy integration candidate for Weave.scope-- imagine the pretty pictures!

awh:
@tomwilkie does it make sense to rework this to use the new BI stuff instead of plotly?

tomwilkie:
Could do, but bi.weave.works isn't public, and this could be.

bboreham:
Closing due to inactivity

PR #1153, opened 2015-07-15
[proxy] should not attempt weavewait in on weave container's execs

### LINKED ISSUE #1143: `docker exec` fails on non-weave containers
````
$ docker run --name=foo -dit ubuntu
4f2e5cbb280b17a1a9e3f8ff8baff46b7aa23e605b1f9d1eac72ad72fca7b9bf
$ weave launch
$ eval $(weave env)
$ docker exec foo ls
exec: "/w/w": stat /w/w: no such file or directory
````

We shouldn't be attempting weavewait unless the container has the requisite volume mount.

paulbellamy:
Fixes #1143

### DIFF
--- a/proxy/create_exec_interceptor.go
+++ b/proxy/create_exec_interceptor.go
@@ -29,7 +29,9 @@ func (i *createExecInterceptor) InterceptRequest(r *http.Request) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


rade:
CI failed.

PR #1164, opened 2015-07-16
fix detection of host netns when weave script runs inside container

### LINKED ISSUE #1162: `weave attach` fails to complain about `--net=host`
split out from #1157...

````
$ weave launch
$ docker run --net=host --name=foo -dti ubuntu
917ea31ec6df7d5536a1c1dd173a2d1ea5f24742a86608bced4e928f3dbed55b
$ weave attach foo
10.32.0.1
````
This is supposed to complain; instead we end up with an `ethwe` in the host network namespace:
````
$ ip link show ethwe
45: ethwe: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 65535 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
    link/ether f2:33:6e:fb:67:aa brd ff:ff:ff:ff:ff:ff
````
which is bad.

The issue appears to arise due to executing inside the weavexec container; we get the expected error message when running with `--local`...

````
$ weave launch
$ docker run --net=host --name=foo -dti ubuntu
1c91f506678dc7dc25e49ac9e08cc67b1bd00e36f787c5f7c60881ca4c86f527
$ WEAVE_DEBUG=1 weave attach foo
...
+ readlink /hostproc/23784/ns/net
+ readlink /hostproc/6/ns/net
+ [ -h /hostproc/23784/ns/net -a -h /hostproc/6/ns/net -a net:[4026531968] =  ]
...
10.32.0.1
$ weave detach foo
10.32.0.1
$ sudo WEAVE_DEBUG=1 weave --local attach foo
...
+ readlink /proc/23784/ns/net
+ readlink /proc/24051/ns/net
+ [ -h /proc/23784/ns/net -a -h /proc/24051/ns/net -a net:[4026531968] = net:[4026531968] ]
+ echo Container is running in the host network namespace, and therefore cannot be
+ echo connected to weave. Perhaps the container was started with --net=host.
...
````

rade:
$$ refers to the pid in the container's process namespace, so using it as a reference into /hostproc is entirely wrong.

Fixes #1162.

### DIFF
--- a/weave
+++ b/weave
@@ -359,7 +359,7 @@ netnsenter() {
 }
 
 connect_container_to_bridge() {
-    if [ -h "$PROCFS/$CONTAINER_PID/ns/net" -a -h "$PROCFS/$$/ns/net" -a "$(readlink $PROCFS/$CONTAINER_PID/ns/net)" = "$(readlink $PROCFS/$$/ns/net)" ] ; then
+    if [ -h "$PROCFS/$CONTAINER_PID/ns/net" -a -h "/proc/$$/ns/net" -a "$(readlink $PROCFS/$CONTAINER_PID/ns/net)" = "$(readlink /proc/$$/ns/net)" ] ; then
         echo "Container is running in the host network namespace, and therefore cannot be" >&2
         echo "connected to weave. Perhaps the container was started with --net=host." >&2
         return 1


tomwilkie:
Might this be worthy of an integration test?

rade:
> Might this be worthy of an integration test?

That would be a departure from our current "we don't test failure cases in the weave script" approach ;)

tomwilkie:
Fair enough, but given weave regressed on this once...

PR #1165, opened 2015-07-16
[proxy] do not weave attach containers in '--net=host|container:<name|id>'

### LINKED ISSUE #1157: If you start a container with --net=host via the proxy, I end up with a ethwe in my host net namespace, and you cannot restart weave
```
$ weave launch
$ eval $(weave env)
$ docker run --net=host --name=foo -dti ubuntu /bin/sh
$ export DOCKER_HOST=
$ weave stop
$ weave launch
Network 10.32.0.0/12 overlaps with existing route 10.32.0.0/12 on host.
Default --ipalloc-range 10.32.0.0/12 overlaps with existing route on host.
```

paulbellamy:
Fixes #1157

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -191,7 +191,12 @@ func (proxy *Proxy) listen(protoAndAddr string) (net.Listener, string, error) {
 	return listener, fmt.Sprintf("%s://%s", proto, addr), nil
 }
 
-func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config) ([]string, bool) {
+func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *docker.HostConfig) ([]string, bool) {
+	if hostConfig != nil &&
+		hostConfig.NetworkMode != "" &&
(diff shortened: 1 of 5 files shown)


proxy/proxy.go
```diff
@@ -191,7 +191,12 @@ func (proxy *Proxy) listen(protoAndAddr string) (net.Listener, string, error) {
 	return listener, fmt.Sprintf("%s://%s", proto, addr), nil
 }
 
-func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config) ([]string, bool) {
+func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *docker.HostConfig) ([]string, bool) {
+	if hostConfig != nil &&
+		hostConfig.NetworkMode != "" &&
+		hostConfig.NetworkMode != "bridge" {
+		return nil, false
```
tomwilkie: Can we log something so explain why we're not attaching?

rade:
needs to be re-branched off 1.0

PR #1166, opened 2015-07-16
Make 210_dns_cross_hosts_test.sh excercise gossip dns tombstones.

tomwilkie:
Also, make gce.sh less verbose, and do the env dance correctly.


rade:
Has this improved coverage?

tomwilkie:
By 0.06% so no, not really.  But not everything can be measured by coverage; this was part of tracking down the issues @2opremio was having.

PR #1167, opened 2015-07-16
[proxy] do not weave attach containers in '--net=host|container:<name|id>'

### LINKED ISSUE #1157: If you start a container with --net=host via the proxy, I end up with a ethwe in my host net namespace, and you cannot restart weave

paulbellamy:
Fixes #1157

Replaces #1165

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -2,6 +2,7 @@ package proxy
 
 import (
 	"crypto/tls"
+	"errors"
 	"net"
 	"net/http"
 	"net/url"
@@ -22,6 +23,10 @@ var (
 	containerCreateRegexp = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/create$")
 	containerStartRegexp  = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/[^/]*/(re)?start$")
 	execCreateRegexp      = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/[^/]*/exec$")
+
+	ErrInvalidNetworkMode = errors.New("--net option")
+	ErrWeaveCIDRNone      = errors.New("WEAVE_CIDR=none")
+	ErrNoDefaultIPAM      = errors.New("--no-default-ipam option")
 )
 
 type Proxy struct {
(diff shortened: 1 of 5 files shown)


proxy/proxy.go
```diff
@@ -126,3 +129,14 @@ func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config) ([]string, bool)
...
+func validNetworkMode(hostConfig *docker.HostConfig) bool {
+	return networkMode(hostConfig) == "bridge"
+}
+
+func networkMode(hostConfig *docker.HostConfig) string {
+	if hostConfig == nil || hostConfig.NetworkMode == "" {
+		return "bridge"
+	}
+	return hostConfig.NetworkMode
+}
```
rade: inline this function.

proxy/start_container_interceptor.go
```diff
@@ -21,7 +21,11 @@ func (i *startContainerInterceptor) InterceptResponse(r *http.Response) error {
 		return err
 	}
 
-	cidrs, ok := i.proxy.weaveCIDRsFromConfig(container.Config)
+	if !validNetworkMode(container.HostConfig) {
+		Debug.Printf("Ignoring container %s with --net=%s", container.ID, networkMode(container.HostConfig))
+		return nil
+	}
```
rade: So now we are checking twice, both here and in `weaveCIDRsFromConfig`. Not great.

I suggest changing the `ok` return of `weaveCIDRsFromConfig` to a messsage (or error?) instead, which we can then log.

rade: And I'd log at 'info' level.

rade:
LGTM. Would you mind merging this yourself, and propagate to master, since it will involve some conflict resolving.

PR #1170, opened 2015-07-16
Re-arrange weaveexec Dockerfile to improve download times

bboreham:
Combining multiple ADDs on the same line reduces the number of layers, and moving the stuff that doesn't change earlier in the file improves the chance it will re-use that layer from cache.

Empirically, the ADD lines are never cached, so they should come last.


rade:
I am not especially keen on earlier layers - entrypoint, in this case - depending on later layers. But meh.

PR #1172, opened 2015-07-17
Use Boostrap styles for warnings in docs

### LINKED ISSUE #1061: improve blockquote rendering
http://docs.weave.works/weave/latest_release/weavedns.html#usage - the markdown blockquote in that is rendered rather oddly.

@davkal

davkal:
Instead of using `> ` for warnings in the docs, we'll use 
```html
<div class="alert alert-warning">
Crazy stuff happening...
</div>
```

Fixes #1061

### DIFF
--- a/site/_layouts/default.html
+++ b/site/_layouts/default.html
@@ -19,7 +19,7 @@
 
      <!-- CSS
        –––––––––––––––––––––––––––––––––––––––––––––––––– -->
-    <link rel="stylesheet" type="text/css" href="assets/custom.css">
+    <link rel="stylesheet" type="text/css" href="assets/custom.css?v=1">
 
      <!-- Favicon
        –––––––––––––––––––––––––––––––––––––––––––––––––– -->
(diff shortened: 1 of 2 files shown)


rade:
I don't really like this. Makes the markdown less readable. Can't we have some CSS for rendering markdown blockquotes?

davkal:
Ok, I mean, abusing blockquote that way isnt exactly clean, either. What if someone really wants to quote?
But I can be convinced by the fact that mixing html and markdown is a tinge dirtier, still. I'll try a CSS based solution.

rade:
> What if someone really wants to quote?

Ideally the resulting style would work for both quotes and "important info". So shouldn't be too shrill. No red, bold, flashing blobs :)

davkal:
screen shot 2015-07-17 at 14 15 39
Going with the bootstrap alert style, but overriding blockquote styling.

rade:
Looks fine to me.

davkal:
Cool, I just pushed the fix. Original implementation is in website repo, copied compiled css over to here and added `?v=1` to css link as a browser cache buster.

Ready for review.

PR #1174, opened 2015-07-17
More helpful error message when --ip-alloc-range mismatched

### LINKED ISSUE #1128: "IP Allocator with different subnet detected" error could be more helpful
I just connected a weave 1.0 and weave-master and got the following in the latter's logs:
````
ERRO: 2015/07/14 09:15:10.658636 ->[192.168.48.11:6783|8e:70:be:57:9b:a5(host1)]: connection shutting down due to error: IP Allocator with different subnet detected
````
which would be a lot more helpful if it a) told me what the ranges were (both ours and theirs), and b) how to fix it (i.e. supply `--ipalloc-range`).

The latter might be a bit tricky, since we don't really want `ring.go`, where that error arises, to have to know about command line options.

### PR DESCRIPTION
**Pull request by bboreham**
Fixes #1128

I did an additional check at the `Allocator` level rather than modifying the check inside `Ring`, to keep the latter unaware of why `Merge` has been called.

### DIFF
--- a/net/address/address.go
+++ b/net/address/address.go
@@ -23,6 +23,17 @@ func (r Range) String() string             { return fmt.Sprintf("[%s-%s)", r.Sta
 func (r Range) Overlaps(or Range) bool     { return !(r.Start >= or.End || r.End <= or.Start) }
 func (r Range) Contains(addr Address) bool { return addr >= r.Start && addr < r.End }
 
+func (r Range) AsCIDRString() string {
+	prefixLen := 32
+	for size := r.Size(); size > 1; size = size / 2 {
+		if size%2 != 0 { // Size not a power of two; cannot be expressed as a CIDR.
+			return r.String()
+		}
+		prefixLen--
+	}
+	return CIDR{Start: r.Start, PrefixLen: prefixLen}.String()
+}
+
 type CIDR struct {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: net/address/address.go [hunk 1]
```diff
@@ -23,6 +23,17 @@ func (r Range) String() string             { return fmt.Sprintf("[%s-%s)", r.Sta
 func (r Range) Overlaps(or Range) bool     { return !(r.Start >= or.End || r.End <= or.Start) }
 func (r Range) Contains(addr Address) bool { return addr >= r.Start && addr < r.End }
 
+func (r Range) AsCIDRString() string {
+	prefixLen := 32
+	for size := r.Size(); size > 1; size = size / 2 {
+		if size%2 != 0 { // Size not a power of two; cannot be expressed as a CIDR.
+			return fmt.Sprintf("[Range cannot be expressed in CIDR format] %s", r)
```
rade: I'd just return `r.String()` here.

bboreham: done

rade:
> I did an additional check at the `Allocator` level rather than modifying the check inside `Ring`, to keep the latter unaware of why `Merge` has been called.

Fair enough, but that means you really ought to change the error message and error name in the `Ring` code.

PR #1176, opened 2015-07-17
Make Gossip DNS reject gossip messages if clockskew > tombstoneTimeout / 2

### LINKED ISSUE #1156: IPAM rejects updates with large clock skew, but it doesn't really care about the clock, whereas DNS doesn't and does
This is a hold-over from when we had tombstones.

tomwilkie:
Fixes #1156 

5mins max clock skew seems too small to me; I suggest we make it bigger (and therefore make the DNS tombstone timeout larger).

### DIFF
--- a/nameserver/nameserver.go
+++ b/nameserver/nameserver.go
@@ -20,10 +20,13 @@ const (
 	// we delete entries for disconnected peers.  Therefore they just need to hang
 	// around to account for propagation delay through gossip.  10 minutes sounds
 	// long enough.
-	tombstoneTimeout = time.Minute * 10
+	tombstoneTimeout = time.Minute * 30
 
 	// Used by prog/weaver/main.go and proxy/create_container_interceptor.go
 	DefaultDomain = "weave.local."
+
+	// Maximum age of acceptable gossip messages (to account for clock skew)
+	gossipWindow = int64(tombstoneTimeout/time.Second) / 2
 )
 
 // Nameserver: gossip-based, in memory nameserver.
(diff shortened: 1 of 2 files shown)


nameserver/nameserver.go
```diff
@@ -208,12 +214,16 @@ func (n *Nameserver) OnGossipUnicast(sender router.PeerName, msg []byte) error {
...
-	var entries Entries
-	if err := gob.NewDecoder(bytes.NewReader(msg)).Decode(&entries); err != nil {
+	var gossip GossipData
+	if err := gob.NewDecoder(bytes.NewReader(msg)).Decode(&gossip); err != nil {
 		return nil, nil, err
 	}
 
-	if err := entries.check(); err != nil {
+	if gossip.Timestamp < now()-int64(gossipWindow/time.Second) {
+		return nil, nil, fmt.Errorf("Refusing old gossip message: %d", gossip.Timestamp)
```
bboreham: Think it would be worth telling the user the apparent difference in time; it's quite hard from a 64-bit int to see whether your clock is just a bit slow or your machine is set to 1970.

nameserver/nameserver.go
```diff
@@ -210,12 +218,16 @@ func (n *Nameserver) OnGossipUnicast(sender router.PeerName, msg []byte) error {
...
-	var entries Entries
-	if err := gob.NewDecoder(bytes.NewReader(msg)).Decode(&entries); err != nil {
+	var gossip GossipData
+	if err := gob.NewDecoder(bytes.NewReader(msg)).Decode(&gossip); err != nil {
 		return nil, nil, err
 	}
 
-	if err := entries.check(); err != nil {
+	if delta := gossip.Timestamp - now(); delta > gossipWindow || delta < -gossipWindow {
+		return nil, nil, fmt.Errorf("Refusing old gossip message (delta %d): %d", delta, gossip.Timestamp)
```
bboreham: This will get printed out as "connection shutting down due to error: Refusing old gossip message ...", so it doesn't really explain what happened.

How about "clock skew of %v detected" ?

bboreham:
Apart from one nit this LOKTM

tomwilkie:
@bboreham I've addressed your comment and after discussion with @rade increase the tombstone timeout on DNS to 30mins (allow a max clockskew of 15mins)

tomwilkie:
Test failure is flakey TestNameservers so can safely be ignored.  I'm looking into it.

Issue #1171, opened 2015-07-16
getent failures in 210_dns_cross_hosts_test.sh


bboreham:
Getting errors fairly often, but not every time.  For example:

````
$ ./210_dns_cross_hosts_test.sh 
Resolve names across hosts
cff20f82331df0d70d2fb11e92eff15be65e2a5a5ac408c0e952ba8bb761920a
8404284b1069c09ab4b8e2e316c111878e3626cf45a5476cdb49aa3a7f2c4cc8
921e10ad338c21f7602b05bd834801844fdb571d6486e10d92ff39564d2912ce
28d009291e4ada0f5444dd3c2948db135d953f2ed187175f057100ddd753e6bc
827ca410d18cffc7ac0c91167ae0f1d60f77f33604c89ed5405a77e0cc06df30
test #32 "exec_on 192.168.48.11 baz getent hosts bar.weave.local" failed:
	program terminated with code 0 instead of 2
test #42 "exec_on 192.168.48.11 baz getent hosts bar.weave.local" failed:
	program terminated with code 0 instead of 2
2 of 67 tests failed in 102.807s.
$ ./210_dns_cross_hosts_test.sh 
Resolve names across hosts
adc9137baa3f12b20b4368c7ff1d77999b429ff5f885040e4329893a8ddd3627
200a7c2d47acd8a11764a7a0c4b85ee0cdc13b2b9800904122cd26cf4747e9f2
f4d52eae68cb653cf41562a1bee1190fc777b6908ebc2bbcaf6b4cbf82097e18
513b6063b9159d5653c2f3f934d833c479f97076ead8a7def813ca4a5aea7d42
a09583b06c421755065b48e1f7ce37bad80e7fbbcf6b0cc22db5d886d4234770
test #52 "exec_on 192.168.48.11 baz getent hosts bar.weave.local" failed:
	program terminated with code 0 instead of 2
1 of 67 tests failed in 27.148s.
````

bboreham:
Similar in test 260:

````
$ ./260_dns_removal_test.sh
Automatic DNS record removal on container death
576fc4e50dd43944ceebb9d543ee3c1520baf56742fd20c2856daf2a0dc79b4b
069d90844c156838983427a938e4f5c23a9de44dac722b22aac91e5cecc6d775
3bb9c34d2fe6321e58466febb5e8bc40256c4de6ff572845391a646a28a2fab0
10.2.0.43
[...]

PR #1180, opened 2015-07-17
Don't swallow errors in the docker client.

tomwilkie:
Might be part of #1171

### DIFF
--- a/nameserver/nameserver.go
+++ b/nameserver/nameserver.go
@@ -120,7 +120,7 @@ func (n *Nameserver) ReverseLookup(ip address.Address) (string, error) {
 	return match.Hostname, nil
 }
 
-func (n *Nameserver) ContainerDied(ident string) error {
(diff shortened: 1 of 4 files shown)


nameserver/nameserver.go
```diff
@@ -130,10 +130,12 @@ func (n *Nameserver) ContainerDied(ident string) error {
 		return false
 	})
 	n.Unlock()
-	if n.gossip != nil {
-		return n.gossip.GossipBroadcast(entries)
+	if n.gossip != nil && len(*entries) > 0 {
+		err := n.gossip.GossipBroadcast(entries)
+		if err != nil {
+			n.errorf("Failed to broadcase container death: %v", err)
```
rade: "broadcast"

rade: also prob worth mentioning the ident.

rade:
rest lgtm

PR #1189, opened 2015-07-17
remove outermost gob

rade:
...replacing it with length-prefix encoding.

Addresses part of #519.


dpw:
> Looks like it might end up with two identical lines in doIntroV1 and doIntroV2.

Ah, I see - they diverge again in the later commit, so no point in unifying them.

rade:
@dpw I have addressed all your feedback. Will squash the commits once you are happy.

dpw:
That was all, squash away.

rade:
> squash away

Done.

Issue #1158, opened 2015-07-15
TestNameservers is flaky


tomwilkie:
Very occasionally (1 in 15 runs so far)

```
--- FAIL: TestNameservers-4 (2.77s)
        Error Trace:    nameserver_test.go:95
			nameserver_test.go:149
			nameserver_test.go:191
	Error:  	Not equal: "51.51.253.7" (expected)
			        != "" (actual)
		
FAIL
coverage: 60.2% of statements
```

paulbellamy:
This test has 5^1000 codepaths, and no output as to which one it picked.

tomwilkie:
I can't get this to flake:

```
vagrant@ubuntu-14:~/src/github.com/weaveworks/weave/nameserver$ while true; do go test -cpu 4 -run TestNameserver; done
PASS
ok  	github.com/weaveworks/weave/nameserver	5.699s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.307s
PASS
ok  	github.com/weaveworks/weave/nameserver	4.947s
PASS
ok  	github.com/weaveworks/weave/nameserver	4.859s
PASS
ok  	github.com/weaveworks/weave/nameserver	4.897s
PASS
ok  	github.com/weaveworks/weave/nameserver	6.186s
PASS
ok  	github.com/weaveworks/weave/nameserver	6.016s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.198s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.485s
PASS
ok  	github.com/weaveworks/weave/nameserver	6.347s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.380s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.129s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.835s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.351s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.790s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.917s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.395s
PASS
ok  	github.com/weaveworks/weave/nameserver	6.054s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.713s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.684s
PASS
ok  	github.com/weaveworks/weave/nameserver	5.204s
PASS
[...]

PR #1190, opened 2015-07-17
add some debugging output to a flakey test

paulbellamy:
Should help diagnose #1158, if we ever manage to reproduce it.


tomwilkie:
Could we just have each of the functions append a string to a list say what its done?   And then dump the list if it fails?  Seems like that would be less code.

PR #1191, opened 2015-07-17
Don't delete record of IP allocation immediately after container death

### LINKED ISSUE #1047: Retain IP address for container across restart
[from IRC](https://botbot.me/freenode/weavenetwork/2015-06-29/?msg=43178991&page=1)

If you use `docker restart` then weave should arrange that the container retains its IP 
address.

(Note that [restart](https://docs.docker.com/docker/reference/api/docker_remote_api_v1.14/#restart-a-container) is indeed a Docker daemon command)

This could be viewed as a sub-case of #599, IP address mobility.

bboreham:
This fixes #1047, without any special reference to the restart command, because the same container ID will get back the same address as long as you start/attach it within the timeout window of 5 seconds.

Perhaps the timeout should be configurable?

### DIFF
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -14,6 +14,7 @@ import (
 type ContainerObserver interface {
 	ContainerStarted(ident string)
 	ContainerDied(ident string)
+	ContainerDestroyed(ident string)
 }
 
 type Client struct {
@@ -82,11 +83,11 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
 		for event := range events {
 			switch event.Status {
 			case "start":
-				id := event.ID
-				ob.ContainerStarted(id)
+				ob.ContainerStarted(event.ID)
 			case "die":
-				id := event.ID
-				ob.ContainerDied(id)
+				ob.ContainerDied(event.ID)
+			case "destroy":
+				ob.ContainerDestroyed(event.ID)
 			}
 		}
 	}()
--- a/ipam/allocate.go
+++ b/ipam/allocate.go
@@ -26,6 +26,8 @@ func (g *allocate) Try(alloc *Allocator) bool {
 	}
 
 	if addr, found := alloc.lookupOwned(g.ident, g.r); found {
+		// If we had heard that this container died, resurrect it
+		delete(alloc.dead, g.ident) // delete is no-op if key not in map
 		g.resultChan <- allocateResult{addr, nil}
 		return true
 	}
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -21,7 +21,7 @@ const (
 	msgRingUpdate
 	msgSpaceRequestDenied
 
-	paxosInterval = time.Second * 5
+	tickInterval  = time.Second * 5
 	MinSubnetSize = 4 // first and last addresses are excluded, so 2 would be too small
 )
 
@@ -53,7 +53,8 @@ type Allocator struct {
 	pendingClaims    []operation                  // held until we know who owns the space
 	gossip           mesh.Gossip                  // our link to the outside world for sending messages
 	paxos            *paxos.Node
-	paxosTicker      *time.Ticker
+	paxosActive      bool
+	ticker           *time.Ticker
 	shuttingDown     bool // to avoid doing any requests while trying to shut down
 	isKnownPeer      func(mesh.PeerName) bool
 	now              func() time.Time
@@ -77,12 +78,14 @@ func NewAllocator(ourName mesh.PeerName, ourUID mesh.PeerUID, ourNickname string
 func (alloc *Allocator) Start() {
 	actionChan := make(chan func(), mesh.ChannelSize)
 	alloc.actionChan = actionChan
+	alloc.ticker = time.NewTicker(tickInterval)
 	go alloc.actorLoop(actionChan)
 }
 
 // Stop makes the actor routine exit, for test purposes ONLY because any
 // calls after this is processed will hang. Async.
 func (alloc *Allocator) Stop() {
+	alloc.ticker.Stop()
 	alloc.actionChan <- nil
 }
 
@@ -501,19 +504,17 @@ func (alloc *Allocator) SetInterfaces(gossip mesh.Gossip) {
 
 func (alloc *Allocator) actorLoop(actionChan <-chan func()) {
 	for {
-		var tickChan <-chan time.Time
-		if alloc.paxosTicker != nil {
-			tickChan = alloc.paxosTicker.C
-		}
-
 		select {
 		case action := <-actionChan:
 			if action == nil {
 				return
 			}
 			action()
-		case <-tickChan:
-			alloc.propose()
+		case <-alloc.ticker.C:
+			if alloc.paxosActive {
+				alloc.propose()
+			}
+			alloc.tryPendingOps()
 		}
 
 		alloc.assertInvariants()
@@ -525,18 +526,16 @@ func (alloc *Allocator) actorLoop(actionChan <-chan func()) {
 
 // Ensure we are making progress towards an established ring
 func (alloc *Allocator) establishRing() {
-	if !alloc.ring.Empty() || alloc.paxosTicker != nil {
+	if !alloc.ring.Empty() || alloc.paxosActive {
 		return
 	}
 
+	alloc.paxosActive = true
 	alloc.propose()
 	if ok, cons := alloc.paxos.Consensus(); ok {
 		// If the quorum was 1, then proposing immediately
 		// leads to consensus
 		alloc.createRing(cons.Value)
-	} else {
-		// re-try until we get consensus
-		alloc.paxosTicker = time.NewTicker(paxosInterval)
 	}
 }
 
@@ -549,13 +548,9 @@ func (alloc *Allocator) createRing(peers []mesh.PeerName) {
 
 func (alloc *Allocator) ringUpdated() {
 	// When we have a ring, we don't need paxos any more
-	if alloc.paxos != nil {
+	if alloc.paxosActive {
+		alloc.paxosActive = false
 		alloc.paxos = nil
-
-		if alloc.paxosTicker != nil {
-			alloc.paxosTicker.Stop()
-			alloc.paxosTicker = nil
-		}
 	}
 
 	alloc.space.UpdateRanges(alloc.ring.OwnedRanges())
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -21,8 +21,9 @@ const (
 	msgRingUpdate
 	msgSpaceRequestDenied
 
-	tickInterval  = time.Second * 5
-	MinSubnetSize = 4 // first and last addresses are excluded, so 2 would be too small
+	tickInterval         = time.Second * 5
+	MinSubnetSize        = 4 // first and last addresses are excluded, so 2 would be too small
+	containerDiedTimeout = time.Second * 5
 )
 
 // operation represents something which Allocator wants to do, but
@@ -51,6 +52,7 @@ type Allocator struct {
 	nicknames        map[mesh.PeerName]string     // so we can map nicknames for rmpeer
 	pendingAllocates []operation                  // held until we get some free space
 	pendingClaims    []operation                  // held until we know who owns the space
+	dead             map[string]time.Time         // containers we heard were dead, and when
 	gossip           mesh.Gossip                  // our link to the outside world for sending messages
 	paxos            *paxos.Node
 	paxosActive      bool
@@ -70,6 +72,7 @@ func NewAllocator(ourName mesh.PeerName, ourUID mesh.PeerUID, ourNickname string
 		paxos:       paxos.NewNode(ourName, ourUID, quorum),
 		nicknames:   map[mesh.PeerName]string{ourName: ourNickname},
 		isKnownPeer: isKnownPeer,
+		dead:        make(map[string]time.Time),
 		now:         time.Now,
 	}
 }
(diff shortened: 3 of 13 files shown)


bboreham:
As it stands, this change has the effect that you can starve the IP allocator by starting and stopping a lot of containers in succession.  We could, if we run out of free addresses, start taking the oldest ones from the dead list.

Similarly, we will be unable to donate space to a peer if we have lots of dead ones piled up.

bboreham:
New idea:
Do not retain IP addresses for un-named containers.
Retain named containers' IP addresses for longer than 5 seconds. 5 minutes?
Discard named containers' IP addresses when we see a 'destroy' event.

bboreham:
closed in favour of #1196

bboreham:
Maybe we do want to do it this way after all.

bboreham:
Need to bring docs over from #1196, removing the comment about not working on auto-restart once #1210 is merged.

bboreham:
Brought docs and `weave restart` command in from #1196.
The docs are written assuming #1210 is merged.

ipam/allocator.go
```diff
@@ -75,6 +79,7 @@ func NewAllocator(ourName router.PeerName, ourUID router.PeerUID, ourNickname st
 func (alloc *Allocator) Start() {
 	actionChan := make(chan func(), router.ChannelSize)
 	alloc.actionChan = actionChan
+	alloc.ticker = time.NewTicker(tickInterval)
```
paulbellamy: Should probably stop this when the allocator stops.

ipam/allocator.go
```diff
@@ -230,23 +250,26 @@ func (alloc *Allocator) ContainerDied(ident string) {
...
+		errChan <- alloc.delete(ident)
+	}
+	return <-errChan
+}
 
-		// Also remove any pending ops
-		found = alloc.cancelOpsFor(&alloc.pendingAllocates, ident) || found
-		found = alloc.cancelOpsFor(&alloc.pendingClaims, ident) || found
+func (alloc *Allocator) delete(ident string) error {
+	addrs, found := alloc.owned[ident]
```
paulbellamy: `found` is unused

ipam/allocator.go
bboreham: ok

site/features.md
```diff
@@ -118,6 +118,22 @@ IP addresses of external services the hosts or containers need to
 connect to. The same IP range must be used everywhere, and the
 individual IP addresses must, of course, be unique.
 
+If you restart a container, it will retain the same IP addresses on
+the weave network:
```
tomwilkie: Should we mention the 5 sec timeout?  Retaining IP across stop / start.

ipam/allocator.go
```diff
@@ -51,9 +52,11 @@ type Allocator struct {
 	nicknames        map[router.PeerName]string   // so we can map nicknames for rmpeer
 	pendingAllocates []operation                  // held until we get some free space
 	pendingClaims    []operation                  // held until we know who owns the space
+	dead             map[string]time.Time         // containers we heard were dead, and when
```
tomwilkie: I think my preference would be to turn owned into a map[string]struct, and have that struct contain the list of ips and the time the container was stopped.  This would have less chance of the dead and the owned maps becoming out of sync with future changes.

ipam/allocate.go
```diff
@@ -26,6 +26,10 @@ func (g *allocate) Try(alloc *Allocator) bool {
 	}
 
 	if addr, found := alloc.lookupOwned(g.ident, g.r); found {
+		// Resurrect a container that we heard had died
+		if _, pending := alloc.dead[g.ident]; pending {
```
tomwilkie: You don't need the if; delete is a no-op if the key is missing.

ipam/allocator.go
tomwilkie: Alternatively, you could use a heap, and make the removeDeadContainers significantly cheaper (wouldn't have to iterate through all the recently died) and make reclaiming the oldest dead ip O(1).

tomwilkie:
Minor comments but generally looks fine.

ipam/allocator.go
bboreham: Thank you for the suggestions. 

`dead` is used in just 4 lines at present, so I think the code expansion to implement a heap would be overwhelming. 

`removeDeadContainers` is called once per 5 seconds, and over time there will be a balance between the number of containers it skips over and the number it needs to clear down (which is the number it skipped over last time).  Given that the actual clear-down is more work than the skipping, I do not think it can be made significantly cheaper overall.

site/features.md
bboreham: Do you think we should mention it because if a system is severely overloaded it could take longer than 5 seconds to do the restart?  If we believe that will impact users, we should perhaps go for a longer timeout.

bboreham:
I believe I have addressed all comments.

rade:
> New idea:
> Do not retain IP addresses for un-named containers.
> Retain named containers' IP addresses for longer than 5 seconds. 5 minutes?
> Discard named containers' IP addresses when we see a 'destroy' event.

What happened to that?

bboreham:
I thought we had a conversation where the starvation problem was not considered serious.

rade:
> I thought we had a conversation where the starvation problem was not considered serious.

Did we? In any case, you said in a comment above:

> if a system is severely overloaded it could take longer than 5 seconds to do the restart? If we believe that will impact users, we should perhaps go for a longer timeout.

Surely the longer we make that interval (and I agree that 5s seems awfully short), the more starvation becomes an issue. If, say, we went for a 30s interval, are we safe?

There is also possibly another issue... what happens when a restarting container had IP addresses allocated with `weave attach net:...`? The proxy / weave script won't be asking for those on restart. Does that mean we end up leaking them?

bboreham:
> what happens when a restarting container had IP addresses allocated with `weave attach net:...`? 

The proxy only re-requests based on `WEAVE_CIDR`, so will ignore those other addresses.  

`weave restart` will re-attach all addresses, but sadly neglects to tell IPAM about this, so leaks everything.

Good spot!

If we fixed the above issues, there remains the issue of `weave attach ip:xxx`, which are not stored anywhere so are lost when the process restarts.

rade:
> The proxy only re-requests based on `WEAVE_CIDR`, so will ignore those other addresses.

So those will leak, right?

> If we fixed the above issues, there remains the issue of `weave attach ip:xxx`, which are not stored anywhere so are lost when the process restarts.

I'd argue that the fix should be in the other direction, i.e. restarting should always ditch the additional entries. I could be wrong. We should make a table of all the restart scenarios - 1) `weave restart`, 2) `docker restart` via proxy, 3) `docker stop` followed by `docker start` shortly thereafter, 4) `docker stop` followed by `docker start` a good while later, any others? - and how this affects a) IPAM and non-IPAM addresses specified with WEAVE_CDIR, b) IPAM and non-IPAM addresses specified in `weave attach`, c) hostname-based DNS records, d) additional DNS records.

Whatever we end up with, the IPAM records must be consistent with what the container actually ends up with. Ditto for DNS.

bboreham:
> So those will leak, right?

Yes, when IPAM is informed that the container has come back, it removes the note that it was dead, which means those entries are not cleared out. 

This behaviour would be useful if the user was planning to re-run the `weave attach` command(s) shortly.

I have fixed the `weave restart` case to reclaim everything.  Will work on that table.

bboreham:
Bonus comment: `weave restart` and `docker restart` will generally take 10 seconds if the container is running `/bin/sh`, because it doesn't explicitly handle `SIGTERM`. 

`docker restart` has a `-t` argument to change this interval, but `weave restart` has no such provision.

rade:
We should add a -t option to weave restart. Separate issue.

bboreham:
This is how I think it works:

item | `weave restart` | `docker restart` | `docker stop` followed by `docker start` a good while later | policy restart after reboot
------|----------------------|----------------------|------------------------------------------------------------------------------|------
IPAM addresses specified with `WEAVE_CIDR`        | retained | retained | re-allocated; may change | re-allocated; may change 
non-IPAM addresses specified with `WEAVE_CDIR` | retained | retained | retained | retained
IPAM addresses specified in `weave attach`               | retained | lost | lost | lost
non-IPAM addresses specified in `weave attach`        | retained | lost | lost | lost
hostname-based DNS records                                     | retained | `WEAVE_CIDR` entries retained | `WEAVE_CIDR` entries re-allocated | `WEAVE_CIDR` entries re-allocated 
additional DNS records                                  | lost | lost | lost | lost

With this PR, `docker stop` followed by `docker start` shortly thereafter is treated exactly like `docker restart`.  Also Docker restarting the process itself under a restart policy.

bboreham:
I just tried running 50 containers that exit quickly, and the whole thing hangs.  More work required.

bboreham:
The issue from Friday is now fixed, and this PR is ready for review.

weave
```diff
@@ -735,6 +736,10 @@ echo_addresses() {
     echo $1 $2 $3
 }
 
+echo_just_ips() {
```
awh: There's another in progress PR (#1379) that adds an `echo_ips` function that strips the netmask length from the end. Perhaps `echo_just_ips` should be called `echo_cidrs` instead? Any if you want to go with `echo_just_cidrs` I can update the other PR to be `echo_just_ips`...

common/docker/client.go
```diff
@@ -10,6 +10,10 @@ type ContainerObserver interface {
 	ContainerDied(ident string)
 }
 
+type ContainerObserver2 interface {
```
rade: yuck!

How much hassle is it to add the method to`ContainerObserver` and change all the implementations?

site/features.md
```diff
@@ -118,6 +118,33 @@ IP addresses of external services the hosts or containers need to
...
+    host1$ weave ps a1
+    a1 16:c0:6f:5d:c5:73 10.32.0.3/12
+
+It works the same if you stop and immediately restart:
+
+    host1$ docker stop a1
+    host1$ docker start a1
+
+However, additional addresses added with the `weave attach` command
+will not be retained.
```
rade: Surely addresses from `weave run` aren't retained either. That is one reason for having `weave restart`.

rade:
The leaking of `weave attach`ed IPAM addresses is unfortunate. Losing them would be better. File an issue.

rade:
How does this relate to #401 (part b) and the associated fix in #1210?

bboreham:
#401-b is about when Docker restarts an application container: Docker will send a 'died' then 'started' event.  Previously, weave IPAM would release the address on 'died', so there was no guarantee that the same address would be assigned if re-requested for the same container.

#1210 is what makes weaveproxy re-request the address when it sees the 'started' event, and re-attach the network.

rade:
So what kind of restart is this PR dealing with?

rade:
You said "With this PR, docker stop followed by docker start shortly thereafter is treated exactly like docker restart." How is that different from "when Docker restarts an application container: Docker will send a 'died' then 'started' event."?

bboreham:
It's not different from the point of view of someone who is only listening to the events.
It is different from the point of the view of someone typing `docker start`, `docker restart` or `docker run --restart ...`

#1210 unified all of these by changing from watching the commands go in to watching the events come out.

rade:
I interpreted the table in https://github.com/weaveworks/weave/pull/1191#issuecomment-135785725 as "the current state of affairs". I gather that is not the case, and obviously can't be since the table mentions `weave restart`, which is new in this PR. So the table reflects the state of affairs after this PR. Or perhaps this PR plus #1210. Not sure which. And this PR does appear to be about more than just IP address allocation. In particular, the introduction of `weave restart` allows application containers to be restarted and re-attached to the weave network; something that previously was only possible for containers started via the proxy (and limited at that, since dynamic attachments are forgotten).

rade:
Should also file an issue to retain extra DNS records, at least in some situations.

bboreham:
I have filed the extra issues, rebased and updated the doc as suggested.

bboreham:
Note this PR is currently based off the branch for #1561, so that it gets the retry behaviour and we can try to get test 640 working consistently.

awh:
This looks good, but I am seeing occasional failures of `640_proxy_restart_reattaches_test.sh`. In the cases when it passes, it completes in ~50s; otherwise it times out at ~120s. Instrumentation reveals that the `run_many` loop is hanging after some number of iterations (e.g. 37, 28) in the failure cases; will investigate further.

awh:
`docker events` normally repeats the following every iteration:

```
2016-01-12T12:35:59.388983735Z 322c11c6e0ddd246e8f44e9d2953e84ddfb79a80591429270e2c193677e2b95a: (from alpine) create
2016-01-12T12:35:59.392560828Z 322c11c6e0ddd246e8f44e9d2953e84ddfb79a80591429270e2c193677e2b95a: (from alpine) attach
2016-01-12T12:35:59.410298728Z 322c11c6e0ddd246e8f44e9d2953e84ddfb79a80591429270e2c193677e2b95a: (from alpine) start
2016-01-12T12:35:59.596727907Z 2e23062526708e30f9801ac9faa97b6ab8adcc1cd2c8245d058a1b0421b53778: (from weaveworks/weaveexec:latest) create
2016-01-12T12:35:59.598303701Z 2e23062526708e30f9801ac9faa97b6ab8adcc1cd2c8245d058a1b0421b53778: (from weaveworks/weaveexec:latest) attach
2016-01-12T12:35:59.601584397Z 2e23062526708e30f9801ac9faa97b6ab8adcc1cd2c8245d058a1b0421b53778: (from weaveworks/weaveexec:latest) start
2016-01-12T12:35:59.645652816Z 2e23062526708e30f9801ac9faa97b6ab8adcc1cd2c8245d058a1b0421b53778: (from weaveworks/weaveexec:latest) die
2016-01-12T12:35:59.651364857Z 2e23062526708e30f9801ac9faa97b6ab8adcc1cd2c8245d058a1b0421b53778: (from weaveworks/weaveexec:latest) destroy
2016-01-12T12:35:59.708844581Z 322c11c6e0ddd246e8f44e9d2953e84ddfb79a80591429270e2c193677e2b95a: (from alpine) die
2016-01-12T12:35:59.751633179Z 322c11c6e0ddd246e8f44e9d2953e84ddfb79a80591429270e2c193677e2b95a: (from alpine) destroy
```
However when the test hangs the last output is this:
```
2016-01-12T12:35:59.779496735Z aba7f36e7755112caff1e70771637c37dc04c894edc8bd5162cd9f4069e64d73: (from alpine) create
2016-01-12T12:35:59.782579941Z aba7f36e7755112caff1e70771637c37dc04c894edc8bd5162cd9f4069e64d73: (from alpine) attach
2016-01-12T12:35:59.789890407Z aba7f36e7755112caff1e70771637c37dc04c894edc8bd5162cd9f4069e64d73: (from alpine) start
2016-01-12T12:36:00.094434404Z 378c9d457981165f64ca654f8eec4d2286563ce0b36bda3ec8cea134c0db260d: (from weaveworks/weaveexec:latest) create
2016-01-12T12:36:00.096374780Z 378c9d457981165f64ca654f8eec4d2286563ce0b36bda3ec8cea134c0db260d: (from weaveworks/weaveexec:latest) attach
2016-01-12T12:36:00.099326907Z 378c9d457981165f64ca654f8eec4d2286563ce0b36bda3ec8cea134c0db260d: (from weaveworks/weaveexec:latest) start
2016-01-12T12:36:00.261661612Z 378c9d457981165f64ca654f8eec4d2286563ce0b36bda3ec8cea134c0db260d: (from weaveworks/weaveexec:latest) die
2016-01-12T12:36:00.267873661Z 378c9d457981165f64ca654f8eec4d2286563ce0b36bda3ec8cea134c0db260d: (from weaveworks/weaveexec:latest) destroy
```
In this particular example, the hang occurred on only the second iteration! It looks like the test container isn't dieing for some reason?

awh:
> It looks like the test container isn't dieing for some reason?

The `docker run ... --rm ... /bin/true` process that launched the hanging container persists even after the smoke test aborted.

bboreham:
Is it hanging in `weavewait`?

awh:
> Is it hanging in weavewait?

I'll check. Might be an idea to see if you can reproduce it too - it's failing for me about 20% of the time.

awh:
> Is it hanging in weavewait?

It looks like it is; given that I've seen it hang on the second iteration, it would appear to be unrelated to address exhaustion...

awh:
Let me know when you've rebased on the weavewait race fix and I'll give it another try.

bboreham:
Rebased and succeeded on CircleCI

PR #1193, opened 2015-07-19
Disable dns in multiweave

rade:
Introduce a `--no-dns` option to `weave launch[-router]`. Undocumented, since it's mainly intended for testing. And use that in multiweave, so it works again, post #1065.

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -55,6 +55,7 @@ func main() {
 		peerCount          int
 		apiPath            string
 		peers              []string
+		noDNS              bool
 		dnsDomain          string
 		dnsPort            int
 		dnsTTL             int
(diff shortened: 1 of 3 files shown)


tomwilkie:
If I understand correctly, the only thing stopping multiweave starting is the dns port mapping?  If so, why not disable that (as this change does) for multiweave, but leave all the dns gossip going?   Isn't that what multiweave is suppose to be testing?

rade:
> why not disable that (as this change does) for multiweave, but leave all the dns gossip going?

Didn't realise that was possible. So leave `nameserver.New` and just skip `nameserver.NewDNSServer`?

tomwilkie:
You could do that. Or you could launch both the servers, and just keep the
changes in the weave script to prevent the port mapping.

rade:
> You could do that. Or you could launch both the servers, and just keep the changes in the weave script to prevent the port mapping.

Right. That's a tad obscure if anybody ever does use this option outside testing, since `weave status` would happily report weavedns as running.

rade:
Thinking about it more, I'd rather leave it as is. Disabling DNS should disable all of it. That's the only sensible choice if the option is to be at all useful outside the confines of multiweave.

tomwilkie:
The test failure is #1158

prog/weaver/main.go
```diff
@@ -80,6 +81,7 @@ func main() {
 	mflag.StringVar(&ipsubnetCIDR, []string{"#ipsubnet", "#-ipsubnet", "-ipalloc-default-subnet"}, "", "subnet to allocate within by default, in CIDR notation")
 	mflag.IntVar(&peerCount, []string{"#initpeercount", "#-initpeercount", "-init-peer-count"}, 0, "number of peers in network (for IP address allocation)")
 	mflag.StringVar(&apiPath, []string{"#api", "-api"}, "unix:///var/run/docker.sock", "Path to Docker API socket")
+	mflag.BoolVar(&noDNS, []string{"-no-dns"}, false, "disable DNS server")
```
tomwilkie: Are you keen on --no-dns?  It doesn't really fit with the rest of the --dns-foo flags.  Could it be --dns-disabled or --dns=false?

rade: I am quite keen on it. It fits in with the `--no-*` options in the proxy. and `--no-peer-discovery` here.

PR #1196, opened 2015-07-20
Catch 'restart' command in proxy and retain IP address

### LINKED ISSUE #1047: Retain IP address for container across restart

bboreham:
Fixes #1047, when using the weave Docker API proxy.

When notified of an impending restart, we retain this information until we see a `die` or `destroy` event from Docker.

Note that Docker will do a 'restart' for a stopped container; we cannot do anything there because the IP address is lost when it stops, so we give up in `notify-restart` if `container_id` detects a problem.

Couple of other quiet changes in this PR:
 * avoid printing some proxy error messages twice in the logs
 * reduce time docker waits before killing container from 10 seconds to 1 in smoke-test 640

### DIFF
--- a/proxy/start_container_interceptor.go
+++ b/proxy/start_container_interceptor.go
@@ -13,14 +13,21 @@ type startContainerInterceptor struct{ proxy *Proxy }
 
 func callWeaveAndLog(container *docker.Container, description string, args ...string) error {
 	if _, stderr, err := callWeave(args...); err != nil {
-		return fmt.Errorf("Failed: %s container %s: %s", description, container.ID, string(stderr))
+		return fmt.Errorf("Failed: %s: %s", fmt.Sprintf(description, container.ID), string(stderr))
 	} else if len(stderr) > 0 {
-		Log.Warningf("%s container %s: %s", description, container.ID, string(stderr))
+		Log.Warningf("%s: %s", fmt.Sprintf(description, container.ID), container.ID, string(stderr))
 	}
 	return nil
 }
 
 func (i *startContainerInterceptor) InterceptRequest(r *http.Request) error {
+	if strings.HasSuffix(r.URL.Path, "/restart") {
+		container, err := inspectContainerInPath(i.proxy.client, r.URL.Path)
+		if err != nil {
+			return err
+		}
+		return callWeaveAndLog(container, "Notifying weave of restart of container %s", "notify-restart", container.ID)
+	}
 	return nil
 }
 
@@ -39,7 +46,7 @@ func (i *startContainerInterceptor) InterceptResponse(r *http.Response) error {
 	args := []string{"attach"}
 	args = append(args, cidrs...)
 	args = append(args, "--or-die", container.ID)
-	if err := callWeaveAndLog(container, "Attaching to weave network", args...); err != nil {
+	if err := callWeaveAndLog(container, "Attaching container %s to weave network", args...); err != nil {
 		return err
 	}
 
--- a/proxy/start_container_interceptor.go
+++ b/proxy/start_container_interceptor.go
@@ -1,7 +1,7 @@
 package proxy
 
 import (
-	"errors"
+	"fmt"
 	"net/http"
 	"strings"
 
@@ -10,6 +10,15 @@ import (
 )
 
 type startContainerInterceptor struct{ proxy *Proxy }
+
+func callWeaveAndLog(container *docker.Container, description string, args ...string) error {
+	if _, stderr, err := callWeave(args...); err != nil {
+		return fmt.Errorf("Failed: %s container %s: %s", description, container.ID, string(stderr))
+	} else if len(stderr) > 0 {
+		Log.Warningf("%s container %s: %s", description, container.ID, string(stderr))
+	}
+	return nil
+}
 
 func (i *startContainerInterceptor) InterceptRequest(r *http.Request) error {
 	return nil
@@ -30,11 +39,8 @@ func (i *startContainerInterceptor) InterceptResponse(r *http.Response) error {
 	args := []string{"attach"}
 	args = append(args, cidrs...)
 	args = append(args, "--or-die", container.ID)
-	if _, stderr, err := callWeave(args...); err != nil {
-		Log.Warningf("Attaching container %s to weave network failed: %s", container.ID, string(stderr))
-		return errors.New(string(stderr))
-	} else if len(stderr) > 0 {
-		Log.Warningf("Attaching container %s to weave network: %s", container.ID, string(stderr))
+	if err := callWeaveAndLog(container, "Attaching to weave network", args...); err != nil {
+		return err
 	}
 
 	return i.proxy.client.KillContainer(docker.KillContainerOptions{ID: container.ID, Signal: docker.SIGUSR2})
--- a/site/features.md
+++ b/site/features.md
@@ -118,6 +118,25 @@ IP addresses of external services the hosts or containers need to
 connect to. The same IP range must be used everywhere, and the
 individual IP addresses must, of course, be unique.
 
+If you restart a container, it will retain the same IP addresses on
+the weave network:
+
+    host1$ docker run --name a1 -tdi ubuntu
+    f76b09a9fcfee04551dbb8d951d9a83e7e7d55126b02fd9f44f9f8a5f07d7c96
+    host1$ weave ps a1
+    a1 1e:dc:2a:db:ef:ff 10.32.0.3/12
+    host1$ docker restart a1
+    host1$ weave ps a1
+    a1 16:c0:6f:5d:c5:73 10.32.0.3/12
+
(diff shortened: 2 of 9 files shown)


rade:
I think we should also have a `weave restart` command.

Plus we should document this marvellous new feature somewhere.

bboreham:
Note that the name and addresses are being re-inserted into weaveDNS after restart; there is no special mechanism to retain them.  Possibly costing some extra network traffic.

rade:
> Note that the name and addresses are being re-inserted into weaveDNS after restart

Which kinda makes me wonder exactly what the use case is we are trying to address here.

bboreham:
> Which kinda makes me wonder exactly what the use case is we are trying to address here.

Issue #1047 has a link to the IRC conversation that started it and a Docker issue that shows many people want the IP address to remain the same over a restart.  Although "Docker links" probably make it more critical there.

bboreham:
Also this text from [Kubernetes](https://github.com/GoogleCloudPlatform/kubernetes/blob/release-1.0/docs/services.md#why-not-use-round-robin-dns) summarises a widely-held position re DNS:

* There is a long history of DNS libraries not respecting DNS TTLs and caching the results of name lookups.
* Many apps do DNS lookups once and cache the results.
* Even if apps and libraries did proper re-resolution, the load of every client re-resolving DNS over and over would be difficult to manage.

rade:
Does this help us answer the question of whether the records should be kept in DNS?

bboreham:
> Does this help us answer the question of whether the records should be kept in DNS?

I don't believe so.

rade:
docs?

bboreham:
just pushed docs and smoke tests

site/features.md
```diff
@@ -248,6 +249,29 @@ invocation:
...
+    a1 16:c0:6f:5d:c5:73 10.32.0.3/12
+
+There is also a `weave restart` command, if you are not using the
+Docker API proxy:
+
+    host1$ weave restart b1
+
+Note that if Docker restarts a container under a restart policy it
+will not be re-attached to the weave network.
+
```
rade: I think this should go into `ipam.md`.

test/100_cross_hosts_test.sh
```diff
@@ -27,6 +27,9 @@ start_container $HOST2 net:$SUBNET_2 --name=c6
 C6=$(container_ip $HOST2 c6)
 assert_raises "exec_on $HOST1 c5 $PING $C6"
 
+weave_on $HOST2 restart c6
+assert_raises "exec_on $HOST1 c5 $PING $C6"
+
```
rade: add `# restart retains IP`

site/features.md
bboreham: The statement isn't restricted to those using IPAM; as it happens restart via the proxy worked with manually-assigned addresses already but we didn't document it.

test/640_proxy_restart_reattaches_test.sh
```diff
@@ -9,11 +9,13 @@ NAME=seetwo.weave.local
...
-proxy docker_on $HOST1 run -e WEAVE_CIDR=$C2/24 -dt --name=c2 -h $NAME $SMALL_IMAGE /bin/sh
-proxy docker_on $HOST1 run -e WEAVE_CIDR=$C1/24 -dt --name=c1          $DNS_IMAGE   /bin/sh
+proxy docker_on $HOST1 run -e WEAVE_CIDR="$C2/24 net:10.40.0.0/30" -dt --name=c2 -h $NAME $SMALL_IMAGE /bin/sh
+proxy docker_on $HOST1 run -e WEAVE_CIDR=$C1/24               -dt --name=c1          $DNS_IMAGE   /bin/sh
 
+IP2=$(weave_on $HOST1 ps $2 | grep -o -E '10\.40\.0\.[12]')
 proxy docker_on $HOST1 restart c2
 assert_raises "proxy exec_on $HOST1 c2 $CHECK_ETHWE_UP"
 assert_dns_record $HOST1 c1 $NAME $C2
+assert_dns_record $HOST1 c1 $NAME $IP2
```
rade: why not use a single IP, in the default range? i.e. don't specify any `WEAVE_CIDR`. And call `container_ip` to find it.

proxy/start_container_interceptor.go
```diff
@@ -11,7 +11,24 @@ import (
 
 type startContainerInterceptor struct{ proxy *Proxy }
 
+func callWeaveAndLog(container *docker.Container, description string, args ...string) error {
+	if _, stderr, err := callWeave(args...); err != nil {
+		return fmt.Errorf("Failed: %s container %s: %s", description, container.ID, string(stderr))
```
rade: the original call site in `startContainerInterceptor.InterceptResponse` had this as a warning. Possibly for a good reason. Ask @paulbellamy.

proxy/start_container_interceptor.go
```diff
@@ -11,7 +11,24 @@ import (
...
+	} else if len(stderr) > 0 {
+		Log.Warningf("%s container %s: %s", description, container.ID, string(stderr))
+	}
+	return nil
+}
+
```
rade: Those log messages look garbled to me. Won't we end up with "Failed: Attaching to weave network container 12345: someError", and "Failed: Notifying weave of container restart container 12345: someError"?

proxy/start_container_interceptor.go
```diff
@@ -11,7 +11,24 @@ import (
...
 func (i *startContainerInterceptor) InterceptRequest(r *http.Request) error {
+	if strings.HasSuffix(r.URL.Path, "restart") {
```
rade: hmm "/restart", perhaps? Or some more clever URL path parsing thingy? @paulbellamy

test/640_proxy_restart_reattaches_test.sh
bboreham: I thought it was reasonable to test the case where the container has more than one address.

rade: Meh

proxy/start_container_interceptor.go
paulbellamy: Error is better.

bboreham: Yeah, the previous behaviour was to log the message twice so I thought this would be an improvement.

weave
```diff
@@ -1339,6 +1343,16 @@ case "$COMMAND" in
         when_weave_running with_container_fqdn $CONTAINER put_dns_fqdn $ALL_CIDRS
         echo $RES
         ;;
+    restart)
```
rade: missing from usage.

proxy/start_container_interceptor.go
bboreham: we only get here after the regexp has established it is either 'start' or 'restart' at the very end of the string.

paulbellamy: `/restart` is fine.  Could do a separate path handler, but then you need more refactoring of the interceptor.

proxy/start_container_interceptor.go
bboreham: The second one, with two containers, is an error on my part.  Thanks!

bboreham:
Addressed review comments, rebased and squashed.

proxy/start_container_interceptor.go
rade: The first one still reads wrong to me.

site/features.md
```diff
@@ -118,6 +118,25 @@ IP addresses of external services the hosts or containers need to
...
+    host1$ docker run --name a1 -tdi ubuntu
+    f76b09a9fcfee04551dbb8d951d9a83e7e7d55126b02fd9f44f9f8a5f07d7c96
+    host1$ weave ps a1
+    a1 1e:dc:2a:db:ef:ff 10.32.0.3/12
+    host1$ docker restart a1
+    host1$ weave ps a1
+    a1 16:c0:6f:5d:c5:73 10.32.0.3/12
+
+There is also a `weave restart` command, if you are not using the
+Docker API proxy:
```
rade: Weave Docker API proxy.

weave
```diff
@@ -45,6 +45,7 @@ weave forget        <peer> ...
 weave run           [--with-dns | --without-dns] [<addr> ...]
                       <docker run args> ...
 weave start         [<addr> ...] <container_id>
+weave restart       <container_id>
```
rade: I'd put this after `detach`. Ditto in the code.

bboreham:
Addressed further review comments, rebased and squashed.

bboreham:
Closed in favour of #1191

Issue #598, opened 2015-04-23
[ipam] mixing manual and automatic allocation in the same subnet


rade:
The introduction of IPAM doesn't prevent users from assigning IPs themselves. However, it is currently impossible to mix the two approaches in the same subnet, i.e. place containers in the same subnet which have some IPs allocated automatically and some assigned manually.

We could a) constrain IPAM allocation to a designated portion of the subnet in which containers are placed, (e.g. allocate in a /24, but place the container in a /23), or b) carve out a reserved manual allocation portion of the subnet that IPAM allocates from (e.g. tell IPAM to allocate from a /23 but leave out a specific /24).

errordeveloper:
This is also important for IP address mobility, let's say I start a container and get it's IP automatically in the first place, later find myself in a situation where I must move it to a different host. There could be a need for IP relocation manager, may be, but on the other hand users may be okay with ad-hoc solution to this specialised sort of case. I don't think proposed options A or B would actually cater for this, I think the kind functionality this needs is "random address stealing", which still matches the title of the issue.

rade:
What you are describing is indeed different, and something we've discussed. Please raise a separate issue.

rade:
actually, I'll raise an issue...

rade:
> This is also important for IP address mobility

Different issue. #599

bboreham:
I believe, post #827, this can now be done by starting weave with e.g. `-iprange 10.2.0.0/25`, use IPAM with `net:10.2.0.0/24` and manual addresses like `ip:10.2.0.129/24`

rade:
right. slightly more readable version: `-iprange 10.2.0.0/24` and `net:10.2.0.0/23` and `ip:10.2.1.1/23`.

The problem is that this only works for *one* subnet.

bboreham:
Closed by #1200

PR #1200, opened 2015-07-20
Check manually-assigned addresses with IPAM to see if they clash

### LINKED ISSUE #687: [ipam] refuse manual allocation in ipam range
If a user inadvertently assigns an IP in the range owned by IPAM (as described in #681), weave could display an informative message.

bboreham:
Fixes #687 and #598

One quiet change in this PR:
 * avoid printing error messages from 'claim' twice in the logs

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -215,9 +215,9 @@ func (alloc *Allocator) Lookup(ident string, r address.Range) (address.Address, 
 }
 
 // Claim an address that we think we should own (Sync)
-func (alloc *Allocator) Claim(ident string, addr address.Address) error {
+func (alloc *Allocator) Claim(ident string, addr address.Address, reclaim bool) error {
 	resultChan := make(chan error)
-	op := &claim{resultChan: resultChan, ident: ident, addr: addr}
+	op := &claim{resultChan: resultChan, ident: ident, addr: addr, reclaim: reclaim}
 	alloc.doOperation(op, &alloc.pendingClaims)
 	return <-resultChan
 }
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -215,9 +215,9 @@ func (alloc *Allocator) Lookup(ident string, r address.Range) (address.Address, 
 }
 
 // Claim an address that we think we should own (Sync)
-func (alloc *Allocator) Claim(ident string, addr address.Address, reclaim bool) error {
+func (alloc *Allocator) Claim(ident string, addr address.Address, noErrorOnUnknown bool) error {
 	resultChan := make(chan error)
-	op := &claim{resultChan: resultChan, ident: ident, addr: addr, reclaim: reclaim}
+	op := &claim{resultChan: resultChan, ident: ident, addr: addr, noErrorOnUnknown: noErrorOnUnknown}
 	alloc.doOperation(op, &alloc.pendingClaims)
 	return <-resultChan
 }
--- a/ipam/claim.go
+++ b/ipam/claim.go
@@ -9,12 +9,13 @@ import (
 )
 
 type claim struct {
-	resultChan chan<- error
-	ident      string
-	addr       address.Address
-	reclaim    bool
+	resultChan       chan<- error
+	ident            string
+	addr             address.Address
+	noErrorOnUnknown bool
 }
 
+// Send an error (or nil for success) back to caller listening on resultChan
 func (c *claim) sendResult(result error) {
 	// Make sure we only send a result once, since listener stops listening after that
 	if c.resultChan != nil {
@@ -32,9 +33,7 @@ func (c *claim) sendResult(result error) {
 func (c *claim) Try(alloc *Allocator) bool {
 	if !alloc.ring.Contains(c.addr) {
 		// Address not within our universe; assume user knows what they are doing
-		if c.reclaim {
-			alloc.infof("Ignored address %s claimed by %s - not in our universe", c.addr, c.ident)
-		}
+		alloc.infof("Ignored address %s claimed by %s - not in our universe", c.addr, c.ident)
 		c.sendResult(nil)
 		return true
 	}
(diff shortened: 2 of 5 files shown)


rade:
> I could add some more lines to remove this if desired.

yes, please.

bboreham:
Done

ipam/allocator.go
```diff
@@ -215,9 +215,9 @@ func (alloc *Allocator) Lookup(ident string, r address.Range) (address.Address,
 }
 
 // Claim an address that we think we should own (Sync)
-func (alloc *Allocator) Claim(ident string, addr address.Address) error {
+func (alloc *Allocator) Claim(ident string, addr address.Address, launch bool) error {
 	resultChan := make(chan error)
-	op := &claim{resultChan: resultChan, ident: ident, addr: addr}
+	op := &claim{resultChan: resultChan, ident: ident, addr: addr, launch: launch}
```
squaremo: I do not understand what `launch` mean in these changes

bboreham: I was trying to convey that we want it to behave differently when called at launch of weave compared to later.

squaremo: I think it would be better to name it for how it behaves differently

squaremo:
I think these are two different operations, rather than a variation on one operation (and perhaps that is why I was confused by "launch").

The first is to tell weave about an IP that had been claimed previously. The second is to ask for a particular IP that has not been handed out previously. It's actually the latter that deserves the name "Claim"; the former is more like, well, "Reclaim" or "Recover".

How this is reflected in the HTTP API, I'm not sure -- PUT to a URL seems a good match for both. So a query arg might be appropriate for discriminating.

bboreham:
I renamed 'launch' to 'reclaim'; rebased and squashed.
Also changed the behaviour to output a message and carry on when IPAM is waiting for quorum, rather than hang the caller.  This is more in line with what #687 asks for.

ipam/claim.go
```diff
@@ -12,6 +12,7 @@ type claim struct {
 	resultChan chan<- error
 	ident      string
 	addr       address.Address
+	reclaim    bool
 }
 
 func (c *claim) sendResult(result error) {
```
squaremo: As a drive-by comment: I find it somewhat counter-intuitive that `claim.sendResult(nil)` indicates a successful operation.

ipam/claim.go
```diff
@@ -40,15 +44,23 @@ func (c *claim) Try(alloc *Allocator) bool {
...
+		}
 		return false
 	default:
 		alloc.debugf("requesting address %s from other peer %s", c.addr, owner)
 		err := alloc.sendSpaceRequest(owner, address.NewRange(c.addr, 1))
-		if err != nil { // can't speak to owner right now; figure it out later
-			alloc.infof("Claim %s for %s: %s; will try later.", c.addr, c.ident, err)
-			c.sendResult(nil)
+		if err != nil { // can't speak to owner right now
+			if c.reclaim { // don't hold up launch; figure it out later
```
squaremo: > // don't hold up launch; figure it out later

This comment doesn't make sense -- what does `reclaim` have to do with launch?

bboreham: this code is only called from `weave launch`

squaremo: that's a coincidence

squaremo: Or think of it this way: if reclaim is a general-purpose operation that happens to be called at launch, this comment doesn't belong here. If reclaim is an operation intended _only_ for use during launch, why is it not better adapted to its purpose, e.g., a bulk operation (in other words, why does it look general purpose ..)

bboreham: that's an accident

squaremo: > that's an accident

which bit?

bboreham: that it looks general-purpose.  Certainly the behaviour of apparently agreeing with a claim when we can't answer for sure yet (introduced in #1104) is purely directed at launch-time.

squaremo: OK. As it is, it's a non-sequitur: "if reclaiming, don't hold up launch". Why should you have to know how it's used in the script to be able to make sense of that?

As much as I dislike the idea of holding things up for a comment, I think this is going to cause befuddlement for the next person to look at the bit of code. What I'd rather see is an explanation of why there is a different behaviour for an apparently very similar operation -- or for it to obviously be a different operation.

bboreham:
I decided `noErrorOnUnknown` described the behaviour best.

ipam/claim.go
```diff
@@ -40,15 +43,23 @@ func (c *claim) Try(alloc *Allocator) bool {
 		// success
 	case router.UnknownPeerName:
 		// If our ring doesn't know, it must be empty.
-		alloc.infof("Claim %s for %s: address allocator still initializing; will try later.", c.addr, c.ident)
-		c.sendResult(nil) // don't make the caller wait
+		if c.noErrorOnUnknown {
+			alloc.infof("Claim %s for %s: address allocator still initializing; will try later.", c.addr, c.ident)
+			c.sendResult(nil) // don't make the caller wait
+		} else {
+			c.sendResult(fmt.Errorf("%s is in the range %s, but the allocator is not initialized yet", c.addr, alloc.universe.AsCIDRString()))
```
squaremo: The suggestion of the previous comment is that this will block the caller. Is that the case?

squaremo: Whatever, I'll fix it up in post

squaremo: > Whatever, I'll fix it up in post

#1219

rade:
I struggle to see how this solves #598. Please explain.

bboreham:
Line 866 in the `weave` script informs IPAM of a manually-selected address.  If this can be granted, it is, and hence IPAM will not then allocate that address to something else, which I thought was the main issue of #598.

rade:
ok

bboreham:
I should perhaps have updated the docs, where it says 

> make sure they don’t conflict with any IP ranges in use on the hosts (including those delegated to weave’s automatic IP address allocator)

rade:
Surely http://docs.weave.works/weave/master/head/ipam.html#manual should be updated.

PR #1202, opened 2015-07-20
Upgrade base image of weaveworks/weave-build to ubuntu:15.04

2opremio:
ubuntu:latest runs an old version of the Docker client for which "docker save"
fails when passed multiple images:

"sudo docker save weaveworks/weave:latest weaveworks/weaveexec:latest > weave.tar

Usage: docker save IMAGE

Save an image to a tar archive (streamed to stdout by default)

  -o, --output=""    Write to an file, instead of STDOUT"

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,9 +1,9 @@
-FROM ubuntu
+FROM ubuntu:15.04
 
 RUN apt-get -y update && apt-get -y install --no-install-recommends ca-certificates apt-transport-https
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 36A1D7869245C8950F966E92D8576A8BA88D21E9
 RUN echo deb https://get.docker.io/ubuntu docker main > /etc/apt/sources.list.d/docker.list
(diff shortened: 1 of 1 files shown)


2opremio:
BTW, assuming this gets merged. Is there anything I should know before pushing to DockerHub?

rade:
I don't understand this. We are already updating Docker in that Dockerfile.

2opremio:
@rade The resulting version of Docker is 1.0.1

```
$ docker run --entrypoint docker weaveworks/weave-build version

Client version: 1.0.1
Client API version: 1.12
Go version (client): go1.2.1
Git commit (client): 990021a
```

rade:
> The resulting version of Docker is 1.0.1

But why? Makes no sense given that we install `lxc-docker-1.3.1`

2opremio:
@rade Good point, my bad, sorry. The problem actually was that the image is DockerHub is outdated/overwritten (DockerHub reads "Updated 2 months, 2 weeks ago" but I don't seem to be able to find any further history). 

Now, if you agree @rade, I will close this and simply push to DockerHub. Apologies again.

rade:
> Now, if you agree I will close this and simply push to DockerHub. Apologies again.

Go for it.

Issue #1209, opened 2015-07-22
weaveproxy stub breaks Docker restart policies


bboreham:
The `weavewait` stub, which we use to coordinate network set-up when containers are started via the weave docker API proxy, relies on the proxy sending it a `SIGUSR2`.

As noted at https://github.com/docker/docker/issues/11065, this breaks any docker restart policy, e.g. `docker run --restart=always`.

bboreham:
We could just signal the container directly, but that requires `--pid=host` which means a newer docker

PR #1210, opened 2015-07-22
make proxy attach network on docker start event

### LINKED ISSUE #401: make docker daemon initiated container (re)start work properly

bboreham:
Do network attach on docker event rather than intercepting start and restart commands.

This means it also works when the docker daemon restarts a container.  Fixes #401 and #1258.


bboreham:
This change has a side-effect that 'docker run' via the proxy now returns before the network is set up and the container is running.

rade:
Presumably the container *is* running; just not the application process. Still i can imagine that might trip up some users. And the same is true of `docker start`.

rade:
...and `docker restart`. i.e. it is probably not unreasonable to expect that when these command return successfully that the container is fully up and running.

The question is what might break for users.

bboreham:
It sometimes breaks smoke-tests like 680.  I made a fix.

prog/weavewait/main.go
```diff
@@ -26,6 +26,11 @@ func main() {
 		args = args[1:]
 	}
 
+	if len(args) > 0 && args[0] == "-c" { // tell pid 1 to continue
+		checkErr(syscall.Kill(1, syscall.SIGUSR2))
```
rade: It is only pid 1 if the container has its own process namespace. Which won't be the case if it was started with `--pid=host`. That will be uncommon, but still something we should handle.

Also, I do wonder how safe it is to assume it is pid 1 even when the container *does* have its own process namespace. Does docker guarantee that a) in the docs, or, at least, b) in the code?

bboreham: Having looked at related questions for the issue where sigint is ignored, I do not think the Docker docs are explicit either way. 

Can you see a way to find the effective pid of a container, inside its namespace?

rade: > Can you see a way to find the effective pid of a container, inside its namespace?

Nope.

I think it's worth going through Docker's container startup code to check whether the entrypoint could be executed as anything other than pid 1. Pretty sure the answer is 'no'.

proxy/proxy.go
```diff
@@ -46,25 +48,28 @@ type Config struct {
...
 	}
 
-	client, err := docker.NewClient(dockerSockUnix)
+	client, err := weavedocker.NewClient(dockerSockUnix)
 	if err != nil {
 		return nil, err
 	}
-	p.client = client
+	p.client = client.Client
+	client.AddObserver(p)
```
paulbellamy: Shouldn't this be done right at the end, so it doesn't call us before `p` is "finished"?

proxy/start_container_interceptor.go
```diff
@@ -1,175 +1,26 @@
...
+	container, err := inspectContainerInPath(i.proxy.client, r.URL.Path)
+	if err == nil && containerShouldAttach(container) {
+		i.proxy.createWait(container.ID)
+	}
+	return err
 }
 
 func (i *startContainerInterceptor) InterceptResponse(r *http.Response) error {
-	if r.StatusCode < 200 || r.StatusCode >= 300 { // Docker didn't do the start
+	if r.StatusCode != 201 && r.StatusCode != 204 { // Docker didn't do the start
```
paulbellamy: This is already in with:
 `if r.StatusCode < 200 || r.StatusCode >= 300 { // Docker didn't do the start`

proxy/start_container_interceptor.go
```diff
@@ -1,175 +1,26 @@
...
-func uniqueStrs(s []string) []string {
-	m := map[string]struct{}{}
-	result := []string{}
-	for _, str := range s {
-		if _, ok := m[str]; !ok {
-			m[str] = struct{}{}
-			result = append(result, str)
-		}
+	if err == nil {
+		i.proxy.waitForStart(container.ID)
```
paulbellamy: When a container fails to start won't this leak channels?

prog/weavewait/main.go
paulbellamy: I believe we could get the pid from `Container.State.Pid` when we launch the exec.

proxy/proxy.go
```diff
@@ -220,6 +234,95 @@ func (proxy *Proxy) listen(protoAndAddr string) (net.Listener, string, error) {
...
+		}
+		if len(ips) > 0 {
+			if err := updateHosts(container.HostsPath, container.Config.Hostname, ips); err != nil {
+				return err
+			}
+		}
+	}
+
+	// exec another weavewait to let first weavewait know network is ready
+	config := docker.CreateExecOptions{
```
paulbellamy: How is signalling the process from an exec better than just signalling the container from here?

prog/weavewait/main.go
paulbellamy: Ah, that's the global pid, not the scoped one.

bboreham: I read the Docker container startup code at length; it is quite confusing as there is a lot of duplication between the 'regular start' case and the 'start a process inside an already-existing container' case.

However, it seems pretty clear that they expect the "main" process of the container to have PID 1, as they refer to it as the "init process".

Post #824, we don't attempt to signal in the other case, e.g. when a process is started by `docker exec`.  But this change makes that line of reasoning invalid, so perhaps we should try to do the signal.

errordeveloper:
@paulbellamy if you have a minute, may be you could cherry-pick `--rewrite-inspect` changes on top of this?I got pretty confused last time I tried cherry-picking it myself :confused:

proxy/start_container_interceptor.go
bboreham: > When a container fails to start won't this leak channels?

Yes.  It isn't trivial as there could be multiple parallel requests to start the same container, so we can't just blow away the entry if one of them fails.

We could watch for container-destroyed events and remove the entry there, so they will leak no faster than containers.

proxy/proxy.go
bboreham: > Shouldn't this be done right at the end, so it doesn't call us before p is "finished"?

Good point.

weave
```diff
@@ -74,9 +74,15 @@ EOF
 }
 
 exec_remote() {
+    HOSTS_PATH=$(docker $DOCKER_CLIENT_ARGS run --rm --net=none \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        --entrypoint /bin/sh \
+        $EXEC_IMAGE -c 'docker inspect -f {{.HostsPath}} $(hostname)')
```
bboreham: This change introduces an extra call to Docker on every `weave xxx` invocation; it is necessary so we can bind-mount the containers directory and run `weavehosts` on `weave attach`.  The choice is to optimise run-speed of the proxy over `weave xxx`.

We run `docker` via `docker` in order to guarantee that there is one container we can inspect.

Perhaps a middle-ground approach would be to check if the command is 'attach' before obtaining `CONTAINERS_PATH` ?

paulbellamy: If you're modifying `/etc/hosts` in the weave script, won't that also need this? Either way, doing it only for `attach` and/or `launch-proxy` seems like a good idea.

bboreham: That's sort-of what I meant, but I just realised this is a change too far for this PR, since we are not going to do the edit for `weave run` or `weave attach`.  The reason to defer is it then seems necessary to unwind it in `weave detach`.

bboreham:
OK, I think the coding is done; I'm going to rebase from latest master and take a look at splitting into two PRs: one for #401 and one for #1209 and #1300.

bboreham:
OK, this is rebased and excludes the change to wait for interface via netlink, which will go in a separate PR.

proxy/proxy.go
```diff
@@ -224,6 +247,94 @@ func (proxy *Proxy) listen(protoAndAddr string) (net.Listener, string, error) {
...
+	proxy.Unlock()
+}
+
+func (proxy *Proxy) waitForStart(r *http.Request) {
+	proxy.Lock()
+	wait, found := proxy.waiters[r]
+	proxy.Unlock()
+	if found {
+		Log.Debugf("Wait for start of container %s", wait.ident)
+		<-wait.ch
```
paulbellamy: Race condition with line 286, where `wait.ch = nil`

site/features.md
```diff
@@ -64,6 +64,10 @@ or
 [remote API](https://docs.docker.com/reference/api/docker_remote_api/)
 are attached to the weave network before they begin execution.
 
+Containers started in this way that subsequently restart, either by an
+explicit `docker restart` command or by Docker restart policy, are
+re-attached to the weave network by the weave Docker API proxy.
```
paulbellamy: However, addresses attached after container start will not be re-attached.

test/640_proxy_restart_reattaches_test.sh
```diff
@@ -6,14 +6,29 @@ C1=10.2.0.78
...
+check_attached
+
+# Kill outside of Docker so Docker will restart it
+run_on $HOST1 sudo kill $(docker_on $HOST1 inspect --format='{{.State.Pid}}' c2)
+check_attached
+
+# Restart docker itself
+# - disabled since the commands are different between our Vagrant VMs and GCE
+#run_on $HOST1 sudo systemctl restart docker # for systemd
+#run_on $HOST1 sudo service docker restart # for upstart
```
paulbellamy: Does `sudo /etc/init.d/docker restart` work on both?

bboreham: I don't think anything under 'init' works on a systemd platform.

proxy/proxy.go
bboreham: I don't think they are called from the same thread; `waitForStart` is called from an http-handler thread and notifyWaiters is called from a goroutine started by `docker.Client.AddObserver`.  

It's easy to pull the channel out inside the lock; I'll do that.

test/640_proxy_restart_reattaches_test.sh
bboreham: Worked on systemd locally; on CircleCI I got "Docker is managed via upstart, try using service docker restart"

bboreham:
I have addressed the two most recent review comments, rebased, and commented out the part of the test that will fail until #1209 is addressed.

paulbellamy:
LGTM

awh:
The plan of record is that we're waiting for #1314 before merging this.

bboreham:
Have rebased and adjusted following merge of #1314

proxy/proxy.go
```diff
@@ -224,6 +247,88 @@ func (proxy *Proxy) listen(protoAndAddr string) (net.Listener, string, error) {
...
+			close(wait.ch)
+			wait.ch = nil
+		}
+	}
+	proxy.Unlock()
+}
+
+func (proxy *Proxy) waitForStart(r *http.Request) {
+	var ch chan struct{}
+	proxy.Lock()
```
paulbellamy: `RLock`?

bboreham: Well spotted.  Or I might just change the RWLock to a Lock since we don't need the complications of an RWLock; we have lots of updates and just this one read-only access.

bboreham: I changed the RWMutex to a Mutex.

PR #1217, opened 2015-07-27
document how to run weave when docker access is root-only

### LINKED ISSUE #1186: 'weave env' became less useful when we stopped using tcp socket
This sequence of commands (found in the README) will now fail:

    host2$ weave launch $HOST1
    host2$ eval $(weave env)
    host2$ docker run --name a2 -ti ubuntu
    Post http:///var/run/weave.sock/v1.19/containers/create: dial unix /var/run/weave.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?

If you then try to fix the problem by running as root, 

    host2$ sudo weave launch $HOST1
    host2$ eval $(sudo weave env)
    host2$ sudo docker run --name a2 -ti ubuntu

the third line runs the container without going via the proxy and hence without attaching to weave.

The reason is that sudo does not pick up the environment variable from `weave env`.

rade:
Closes #1186.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,21 +52,29 @@ Say you have docker running on two hosts, accessible to each other as
 `$HOST1` and `$HOST2`, and want to deploy an application consisting of
 two containers, one on each host.
 
-First start weave on $HOST1:
+On $HOST1 we run:
 
     host1$ weave launch && weave launch-dns && weave launch-proxy
-
-this runs the weave router, DNS and Docker API proxy, each in their
-own container. Next we configure our `DOCKER_HOST` environment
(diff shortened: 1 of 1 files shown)


rade:
NB: I branched this of latest_release_doc_updates, since it's an issue with the current release too, though not as severe.

So post merge we need to merge latest_release_doc_updates -> 1.0 -> master -> weave_1.1_readme_updates. All of which will conflict, but should be easy to resolve.

@errordeveloper I'll do the merging myself once you are happy.

errordeveloper:
Yep, looks good :+1:

PR #1222, opened 2015-07-27
[proxy] add --rewrite-inspect flag

### LINKED ISSUE #1199: make proxy work with kubernetes
Kubernetes wants to look at container IP, and the proxy doesn't support this right now.

paulbellamy:
To return the weave ip/mac for docker inspect calls.

~~Blocked on a pr to go-dockerclient being merged: https://github.com/fsouza/go-dockerclient/pull/336~~

@errordeveloper, Could you see if this works for you?

Possibly, fixes #1199?


paulbellamy:
Note, need to check this works with inspecting images..

paulbellamy:
Closed while being tried out. Needs testing for the USR2 change, if it works.

errordeveloper:
Yeah, looks like it does, just testing from scratch now. Will update.

errordeveloper:
Ok, so just for the record, this appears to work, however I am seeing a strange issue which looks like only some `/w/w` processes get `SIGUSR2`. I am not quite sure, but due to the signal having to go via Docker and my overcomited VMs might struggle and hit some edge case where Docker does't end-up sending those signals. To confirm, I will have to trace whether Docker calls `signal()` or what not, as well as read the code. According to the logs, it's supposed to, but I am not entirely convinced with that being a guarantee. I've checked with David, and he say that signals cannot just get dropped, although they may get blocked.

Issue #1215, opened 2015-07-27
FDP can't coexist with new DNS server


dpw:
The new DNS server is integrated into the router, and binds to port 53.  FDP runs the router with `--net=host`, because it needs to perform ODP operations in the default netns.  Put them together, and the router container now tries to bind to port 53 in the default netns, and EADDRINUSE ensues.  So FDP is now likely to only work if you do `weave launch --no-dns`.

Either the DNS server needs to be run on some other port number, or it should be disentangled from the router and moved into its own container once again.  The latter seems preferable, in order to minimize the surface area of code run in a container with `--privileged --net=host`.

rade:
> run on some other port number

It actually just needs to listen on a specific interface/ip, namely the docker bridge ip. That's what the port mappings in the script do, but they are ineffective when running with `--net=host`.

tomwilkie:
We can add a new argument (--dns-listen-address or something) and have the weave script put the docker bridge ip in there.  Will that be enough?

rade:
Why is this still open? I thought #1225 was meant to address this issue?

tomwilkie:
#1225 mostly addresses the issue; @dpw will need to update the weave script in the FDP branch to launch with the new parameter.  I can do this if you like?

rade:
I am quite sure @dpw considers this issue to be pertaining to the current weave code base only, i.e. making any changes to FDP is out of scope.

PR #1225, opened 2015-07-28
Allow users to set dns listen address

tomwilkie:
So we can run the weave router in the host net namespace.

In FDP the weave script will need to set ```--dns-listen-address=<docker bridge ip>:53```

Part of #1215


tomwilkie:
The failure is in 150_connect_forget_test.sh, which is being fixed by #1228

PR #1227, opened 2015-07-28
Extend scheduler to periodically garbage collect stranded VMs

tomwilkie:


test/scheduler/app.yaml
```diff
@@ -7,8 +7,10 @@ threadsafe: true
 handlers:
 - url: .*
   script: main.app
+- url: /tasks/*
+  script: main.app
+  login: admin
```
paulbellamy: This can't be hit, remove please.

PR #1236, opened 2015-07-29
Run integration tests in parallel when possible.

### LINKED ISSUE #1169: make CircleCI more frugal on shard & VM usage
Most of our smoke tests only require one VM. Some two, and an even smaller number need three. Yet our CI setup creates a total of 4 shards x 3 VMs = 12 VMs per test run. We could get away with a lot less if our CircleCI scheduler knew how many VMs are needed by particular tests, and got smarter about the scheduling as a result, e.g. perhaps have one shard with 3 VMs, one with 2 and two with 1.

We could suffix the test file name with `-2`, `-3`, e.g. `100_cross_host_tests-2.sh`. (I'd keep `-1` optional, since it's the common case.

tomwilkie:
Fixes #1169 

The idea here is that most tests only need 1 host; each shard has 3.  Therefore we should be able to run some tests in parallel on different hosts.

If this works, we could reduce the number of CircleCI shards and increase the number of VMs we give them, potentially making the tests shorter and use less resources.


testing/runner/runner.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (t test) run(hosts []string, results chan result) {
+	var out bytes.Buffer
+
+	cmd := exec.Command(t.name)
+	cmd.Env = os.Environ()
+	cmd.Stdout = &out
+	cmd.Stderr = &out
+
+	// repace HOSTS in env
```
paulbellamy: replace

test/run_all.sh
```diff
@@ -25,7 +25,7 @@ check_test_status() {
 # Overwrite assert.sh _assert_cleanup trap with our own
 trap check_test_status EXIT
 
-TESTS="${@:-*_test.sh}"
+TESTS="${@:-$(find . -name '*_test.sh')}"
```
paulbellamy: What's the difference?

tomwilkie: adds ./ to the beginning of the test names, so the runner can find them.

paulbellamy:
Seems like it would be cleaner to just have one test runner which is always used, and decides whether to talk to the scheduler, vs having run_all.sh *and* runner, which sometimes one is used vs the other...

tomwilkie:
Could do; run_all runs a bunch of pretest checks; I take it you're not proposing we write those in go?

I could add a run-sequentially flag to running and then always use runner?

tomwilkie:
Ideas discussed:
- keep the weavewait container running... nope
- use the weave router container, as thats guaranteed to be there and if someone stops it, then... well...
- copy weavewait into the container
  - container is not running, so would need to start with a sleep, copy in, update entrypoint etc.  Unworkable.

tomwilkie:
Another idea - abandon volumes from, have weave script dump weavewait somewhere in the host fs, and mount with a 'proper' volume mount.

tomwilkie:
@paulbellamy PTAL

testing/runner/runner.go
```diff
@@ -0,0 +1,246 @@
...
+			go func() {
+				errored := test.run(testHosts)
+				resultsChan <- result{errored, testHosts}
+			}()
+			outstanding++
+		}
+
+		// Otherwise, wait for the test to finish and return
+		// the hosts to the pool
+		result := <-resultsChan
```
paulbellamy: `resultsChan` could have > 1 things here

PR #1238, opened 2015-07-29
[proxy] only mount /w into weavewaited containers

### LINKED ISSUE #1232: /run shows host system
without weave `docker run -ti --rm postgres:9.3 ls -1 /run` shows `exim3 lock postgresql utmp`.
after running
```
weave launch
weave launch-dns
weave launch-proxy
`weave env`
```
the same command shows the /run of the hostsystem.

i found this, because the default postgres image cant't be started with weave.
it complains:
```
chmod: cannot access ‘/run/postgresql’: No such file or directory
```
when is entrypoint tries to set the permissions.

paulbellamy:
Fixes #1232

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -3,6 +3,7 @@ package proxy
 import (
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
 	"net/url"
@@ -32,9 +33,10 @@ var (
 type Proxy struct {
 	Config
 
-	dial           func() (net.Conn, error)
-	client         *docker.Client
-	dockerBridgeIP string
+	dial            func() (net.Conn, error)
+	client          *docker.Client
+	dockerBridgeIP  string
+	weaveWaitVolume string
 }
 
 type Config struct {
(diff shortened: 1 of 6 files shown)


tomwilkie:
Challenge with this approach is we create a non-running container just for the weavewait volume.  People might delete said container and have things fail mysteriously.  Or they might be running GC scripts that delete if for them (apparently this is common).

I can't think of a better approach however.  Maybe there is something we can do with a data volume?

tomwilkie:
Ideas discussed:
- keep the weavewait container running... nope?
- use the weave router container, as thats guaranteed to be there and if someone stops it, then... well...
- copy weavewait into the container
  - container is not running, so would need to start with a sleep, copy in, update entrypoint etc.  Unworkable.

tomwilkie:
Another idea - abandon volumes from, have weave script dump weavewait somewhere in the host fs, and mount with a 'proper' volume mount.

Another idea from david: use an anonymous data volume

paulbellamy:
@tomwilkie I've changed this to mount /w from the host (from weaveexec's anonymous data volume). The bind is set:
* on container creation
    * where we set the entrypoint
* on container start 
    * in case the proxy anonymous volume has been removed
    * or if the proxy has been restarted/changed-version

PTAL, and I will squash the commits if happy.

Also fixes #1241, as it was being a pain when debugging.

tomwilkie:
Looks good!  Please squash all but "Removed all the imports from "code.google.com"" and I'll take one final quick look.

paulbellamy:
I removed the code to set the binds on container start. It was causing some strange behaviour in tests, and there is no way to override the volumes in the CLI.

proxy/proxy.go
```diff
@@ -139,3 +164,16 @@ func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *dock
...
+	var binds []string
+	for _, bind := range hostConfig.Binds {
+		s := strings.Split(bind, ":")
+		if len(s) >= 2 && s[1] == "/w" {
+			continue
+		}
+		binds = append(binds, bind)
+	}
+	hostConfig.Binds = binds
+	hostConfig.Binds = append(hostConfig.Binds, fmt.Sprintf("%s:/w:ro", proxy.weaveWaitVolume))
```
rade: ditch the `hostConfig.Binds = binds` and simply write "append(binds, ...)" in the next line.

tomwilkie:
LGTM

rade:
> Also fixes #1241

No, it doesn't. @paulbellamy @tomwilkie did either of you actually test that?

PR #1246, opened 2015-07-30
parameterize test/gce.sh so scope can override TEMPLATE_NAME and NUM_HOSTS

paulbellamy:
First part of fixing: https://github.com/weaveworks/scope/issues/334


tomwilkie:
I did not know you could do that.  Assuming is passes tests, LGTM

PR #1253, opened 2015-08-03
[proxy] Handle case where no command is specified.

### LINKED ISSUE #1241: proxy swallows error message when container fails to start
````
$ docker run --rm -it gliderlabs/alpine
Error response from daemon: No command specified
$ weave launch
$ eval $(weave env)
$ docker run --rm -it gliderlabs/alpine
Error response from daemon: ba8ea356fa43646ec9e758af06d4443695483f27152c6ff40ca192ce188a5c01 container is not running.
$ docker logs weaveproxy
...
WARN: 2015/07/30 09:54:09.488245 Attaching container ba8ea356fa43646ec9e758af06d4443695483f27152c6ff40ca192ce188a5c01 to weave network failed: ba8ea356fa43646ec9e758af06d4443695483f27152c6ff40ca192ce188a5c01 container is not running.

WARN: 2015/07/30 09:54:09.488313 Error intercepting response: ba8ea356fa43646ec9e758af06d4443695483f27152c6ff40ca192ce188a5c01 container is not running.
````

### PR DESCRIPTION
**Pull request by paulbellamy**
The docker daemon can't do this because we set /w/w.

Fixes #1241

### DIFF
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -3,6 +3,7 @@ package proxy
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -14,6 +15,10 @@ import (
 )
 
 const MaxDockerHostname = 64
+
+var (
+	ErrNoCommandSpecified = errors.New("No command specified")
+)
 
 type createContainerInterceptor struct{ proxy *Proxy }
 
@@ -103,6 +108,10 @@ func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container *docker.Co
 		if container.Entrypoint == nil {
 			container.Entrypoint = image.Config.Entrypoint
 		}
+	}
+
+	if len(container.Entrypoint) == 0 && len(container.Cmd) == 0 {
+		return ErrNoCommandSpecified
 	}
 
 	if len(container.Entrypoint) == 0 || container.Entrypoint[0] != weaveWaitEntrypoint[0] {
--- a/prog/weavewait/main.go
+++ b/prog/weavewait/main.go
@@ -7,16 +7,14 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/weaveworks/weave/net"
+	weavenet "github.com/weaveworks/weave/net"
+	"github.com/weaveworks/weave/proxy"
 )
 
 func main() {
-	if len(os.Args) <= 1 {
-		os.Exit(0)
-	}
 	args := os.Args[1:]
 
-	if args[0] == "-s" {
+	if len(args) > 0 && args[0] == "-s" {
 		args = args[1:]
 	} else {
 		usr2 := make(chan os.Signal)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by tomwilkie** [comment]
Without the proxy:
```
vagrant@ubuntu-14:~/src/github.com/weaveworks/weave$ docker run gliderlabs/alpine foo
exec: "foo": executable file not found in $PATH
Error response from daemon: Cannot start container ff2e425302332ded5ec18e41cd2f2662d22b3b852ebee078b9e074fd172180b1: [8] System error: exec: "foo": executable file not found in $PATH
vagrant@ubuntu-14:~/src/github.com/weaveworks/weave$ echo $?
1
```

With the proxy:
```
vagrant@ubuntu-14:~/src/github.com/weaveworks/weave$ docker logs hopeful_mayer 
exec: "foo": executable file not found in $PATH
vagrant@ubuntu-14:~/src/github.com/weaveworks/weave$ echo $?
1
```

Not identical, but I don't think there is anything we can do about that.

File: proxy/create_container_interceptor.go [hunk 1]
```diff
@@ -105,7 +110,17 @@ func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container *docker.Co
 		}
```
tomwilkie: How can container.Entrypoint ever be nil?  Its zero-value is an empty list, no?  Or is == nil that same as len(...) == 0?  In which case, we already checked that.

proxy/create_container_interceptor.go
```diff
@@ -105,7 +110,17 @@ func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container *docker.Co
...
 	}
 
+	for _, e := range container.Entrypoint {
+		if e != weaveWaitEntrypoint[0] {
+			return nil
+		}
+	}
```
tomwilkie: What is this trying to achieve?  Why is it checking all entries in container.Entrypoint?

tomwilkie:
The 620_proxy_entrypoint_command_test.sh failures don't looks intermittent, so it looks like something is really broken there.

proxy/create_container_interceptor.go
paulbellamy: The difference between it not being present in the JSON, and it being an empty list in the JSON.

```Go
	var a []string
	var b = []string{}
	fmt.Println(a == nil) // true
	fmt.Println(b == nil) // false
```

proxy/create_container_interceptor.go
rade: This looks just wrong to me. It will return, and hence not insert the `/w/w` when it finds something *other* than `/w/w`.

paulbellamy:
Yeah, not identical. I've had a read through the daemon/client code, but can't quite see why. Can keep digging if we think it's important.

proxy/create_container_interceptor.go
tomwilkie: You are right, this is a stoopid language.

tomwilkie:
LGTM (not merging as its assigned to you)

PR #1254, opened 2015-08-03
Fix dotted imports for logging

dpw:
Many router files do a dotted import of `common` for the sake of `Log`.  This means everything else from `common` ends up in scope, for the sake of a single variable.


squaremo:
NB tested by pushing to a branch in this repo: https://circleci.com/gh/weaveworks/weave/2244

PR #1263, opened 2015-08-05
don't save the python test file when running test/600

paulbellamy:
Otherwise it shows up in git.


tomwilkie:
Ha, I tried to get that to work but didn't for some reason.

PR #1265, opened 2015-08-06
[proxy] rewrite /etc/hosts from outside the container

### LINKED ISSUE #1252: Proxy fails to start containers from scratch or where user is non root
Fails on rewriting /etc/hosts. From scratch has no /etc/hosts, and non-root users have no permission to rewrite it.

Options:
1) Leave /etc/hosts alone if it fails.
2) Output a warning, but continue
3) Fail with a more informative message

paulbellamy:
Avoids problems when container user is non-root. Added a test for this
and the --no-rewrite-hosts flag.

Fixes #1252

### DIFF
--- a/proxy/start_container_interceptor.go
+++ b/proxy/start_container_interceptor.go
@@ -1,8 +1,15 @@
 package proxy
 
 import (
+	"bufio"
+	"bytes"
 	"errors"
+	"fmt"
+	"io/ioutil"
+	"net"
 	"net/http"
+	"os"
+	"sort"
 	"strings"
 
 	"github.com/fsouza/go-dockerclient"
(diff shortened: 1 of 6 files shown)


weave
```diff
@@ -1131,8 +1131,14 @@ launch_proxy() {
     # when launching the weaveproxy container.
     docker_client_args $DOCKER_CLIENT_ARGS
     proxy_args "$@"
+    HOSTS_PATH=$(docker run --rm \
+      -v /var/run/docker.sock:/var/run/docker.sock \
+      --entrypoint /bin/sh \
+      $EXEC_IMAGE -c 'docker inspect -f {{.HostsPath}} $HOSTNAME')
```
rade: It took me a while to figure this out. `$HOSTNAME` gets resolved inside the container, right?

paulbellamy: Correct. Used as a way of finding out the container ID.

test/615_proxy_rewrite_hosts_test.sh
```diff
@@ -0,0 +1,26 @@
...
+  ENV HOME /home/user
+  USER user
+  CMD true
+EOF
+assert "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' non-root hostname -i | cut -d' ' -f1" "10.2.1.1"
+
+# When rewrite hosts is disabled
+weave_on $HOST1 stop-proxy
+weave_on $HOST1 launch-proxy --no-rewrite-hosts
+assert_raises "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' $SMALL_IMAGE hostname -i | cut -d' ' -f1 | grep -v 10.2.1.1"
```
rade: I suggest extracting s.t. like
````
check() {
  assert_raises "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.1/24' $1 hostname -i | cut -d' ' -f1 | grep 10.2.1.1" $2
````
and then call it as
````
check $SMALL_IMAGE
...
check non-root
...
check $SMALL_IMAGE 1
...
````

File: proxy/start_container_interceptor.go [hunk 3]
```diff
@@ -41,5 +48,128 @@ func (i *startContainerInterceptor) InterceptResponse(r *http.Response) error {
...
+	if !i.proxy.NoRewriteHosts {
+		ips, err := weaveContainerIPs(container)
+		if err != nil {
+			return err
+		}
+		if len(ips) > 0 {
+			if err := updateHosts(container.HostsPath, container.Config.Hostname, ips); err != nil {
+				return err
+			}
+		}
```
rade:
````
if err != nil || len(ips) == 0 {
    return err
}
if err := updateHosts(...); err != nil {
    return err
}
````

paulbellamy: This would skip sending USR2 if there are no IPs found.

rade: ah true. never mind.

Issue #1269, opened 2015-08-06
Support alternative shells for `weave env`


paulbellamy:
Certain shells (e.g. fish) don't support "export". For what to do, see `boot2docker shellinit`.

peterbourgon:
It seems that only fish is a unique snowflake.

```go
func printExportUnix(values map[string]string) {
	for name, value := range values {
		switch filepath.Base(os.Getenv("SHELL")) {
		case "fish":
			if value == "" {
				fmt.Printf("    set -e %s\n", name)
			} else {
				fmt.Printf("    set -x %s %s\n", name, value)
			}
		default: // default command to export variables POSIX shells, like bash, zsh, etc.
			if value == "" {
				fmt.Printf("    unset %s\n", name)
			} else {
				fmt.Printf("    export %s=%s\n", name, value)
			}
		}
	}
}
```

https://github.com/boot2docker/boot2docker-cli/blob/master/cmds.go#L218-L233

dpw:
There are lots of weird and wonderful shells out there.  If boot2docker supports fish, I'd guess that means that someone cared enough to contribute fish support, rather than that fish is the only shell that needs special handing.

So I'm sceptical that trying to support non-POSIX shells is worthwhile.  People can always do `sh -c 'eval $(weave env); exec my_favorite_sh'` if they want.

peterbourgon:
@dpw :+1:

paulbellamy:
or use `docker $(weave config) ...`

rade:
> or use `docker $(weave config)` ...

You are assuming here that the shell hasn't invented a different way of passing command line args. ;)

PR #1274, opened 2015-08-10
When giving instructions, prefer set -x to export

peterbourgon:
`export` is a {ba,z,}sh-ism, whereas `set -x` is more generally accepted. Specifically, `export` doesn't work on [fish](http://fishshell.com/).

This changes instructions piped to users, either for manual copy/pasting or for inlining e.g. $(weave env), to use `set -x` instead of `export`.

(Any other places I missed?)

### DIFF
--- a/test/gce.sh
+++ b/test/gce.sh
@@ -139,9 +139,9 @@ function hosts {
 		hosts="$hostname $hosts"
 		args="--add-host=$hostname:$(internal_ip $json $name) $args"
 	done
-	echo export SSH=ssh
-	echo export HOSTS=\"$hosts\"
-	echo export WEAVE_DOCKER_ARGS=\"$args\"
+	echo set -x SSH ssh
+	echo set -x HOSTS \"$hosts\"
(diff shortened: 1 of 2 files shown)


rade:
Is this meant to fix #1269?

peterbourgon:
I didn't see that issue. But I guess it will fix it, yes.

rade:
Have you actually tested this? In POSIX shell, and in bash, `set -x` turns on tracing. It doesn't set/export an env var.

````
$ set -x FOO foo
$ echo $FOO
+ echo
````

peterbourgon:
I did test it, but my test was broken, as I'm (obviously) now getting your behavior. Mea culpa. I guess there's no straightforward fix, we'll have to introspect for the running shell.

dpw:
FWIW, `export` is in POSIX, and always has been.  The original POSIX shell spec didn't support `export VAR=blah`, which is why you will see some scripts doing `VAR=blah ; export VAR`.  But that form was added to POSIX a while ago too.

peterbourgon:
Aye. Fish is deliberately not POSIX compliant.

PR #1289, opened 2015-08-12
Move ListVar into common, so it can be reused elsewhere.

tomwilkie:


dpw:
Can we put this in a specific package please, not common.  If common is to have any valid rationale at all, it should be reserved for things which are genuinely an extremely common requirement.  This does not seem to fit that criterion.

I know that common is already down the slippery slope towards being a dumping ground for stuff that did not seem to fit anywhere else.  But the first step to addressing that is to stop making the problem worse.  And the problem of common is compounded by the fact that it is dot imported by various files.

tomwilkie:
Fair enough - any suggestions?  common/listvar?  common/flags?

dpw:
> Fair enough - any suggestions? common/listvar? common/flags?

It looks like it is specific to the `mflag` package.  Calling it just `common/mflag` would lead to an annoying clash.  So `common/mflagext` or similar.

tomwilkie:
@dpw PTAL

dpw:
> @dpw PTAL

Fine by me.

PR #1290, opened 2015-08-12
Add role label to weave containers so scope can filter

tomwilkie:
As part of https://github.com/weaveworks/scope/issues/337, I'd like to label the weave & scope containers and have them filtered out of the scope view by default.

Let the bike shedding begin...

### DIFF
--- a/prog/weaveexec/Dockerfile
+++ b/prog/weaveexec/Dockerfile
@@ -1,6 +1,7 @@
 FROM gliderlabs/alpine
 
 MAINTAINER Weaveworks Inc <help@weave.works>
+LABEL works.weave.role=system
 
 WORKDIR /home/weave
 VOLUME /w
--- a/prog/weaver/Dockerfile
+++ b/prog/weaver/Dockerfile
@@ -1,5 +1,6 @@
 FROM scratch
 MAINTAINER Weaveworks Inc <help@weave.works>
+LABEL works.weave.role=system
 WORKDIR /home/weave
 ADD ./weaver /home/weave/
 ENTRYPOINT ["/home/weave/weaver"]


tomwilkie:
feedback from @rade - what happens with older versions of docker? Need to support docker 1.3.1

tomwilkie:
To clarify, role=system is something I just made up.

tomwilkie:
@errordeveloper how do kubernetes do labels?  If there an existing schema / topology?

tomwilkie:
@errordeveloper says k8s does something like this:
```
"Labels": {
    "io.kubernetes.pod.name": "kube-system/kube-dns-v8-zwokj"
}
```

tomwilkie:
@rade labels seem to be backward compatible (this imports and runs on docker 1.3.1)

awh:
Is there something contentious about this PR? It's a tiny change but has been open for nine days...

tomwilkie:
@mikeb was against the approach of label containers then filtering based on
label.

squaremo:
> @mikeb was against the approach of label containers then filtering based on label.

https://github.com/weaveworks/scope/issues/337#issuecomment-130349271:

> If you want this to be a general mechanism for hiding containers from display in Scope, I think using labels is a poor choice. Firstly, it's indirect -- you're asking the user to go and change how a container is labelled (which involves recreating the container!), just to hide it in their display. Secondly, the user may not even have control over how containers are created -- that might be done by kubernetes, or marathon, or whatever.

If you want a hook on which to hang some by-default filtering of weave containers, then labels are fine, but you should give them a namespace.

tomwilkie:
> f you want a hook on which to hang some by-default filtering of weave
containers, then labels are fine, but you should give them a namespace.

Thats exactly what we want - the filtering mechanism with be general, but
we need to filter on something, and we'd like that to be a label.  I'll
namespace this one.

tomwilkie:
Failures are unrelated, see #1360

squaremo:
Merged on the understanding that this is not really intended to be a feature for end-users. (If it were, I would be concerned about being stuck with the terms "role" and "system".)

PR #1291, opened 2015-08-12
Docker 1.8 compatibility

### LINKED ISSUE #1287: docker 1.8.0 breaks proxy
From #1285...

````
$ docker logs weaveproxy
INFO: 2015/08/11 21:26:36.897225 weave proxy git-94af02fbf41c
INFO: 2015/08/11 21:26:36.897625 Command line arguments: --tlsverify --tlscacert /home/weave/tls/ca.pem --tlscert /home/weave/tls/cert.pem --tlskey /home/weave/tls/key.pem -H tcp://0.0.0.0:12375
FATA: 2015/08/11 21:26:37.076987 Could not start proxy: Could not find the weavewait volume
$ docker version
Client:
 Version:      1.8.0-rc3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   696147b
 Built:        Fri Aug  7 17:36:36 UTC 2015
 OS/Arch:      darwin/amd64

Server:
 Version:      1.8.0-rc3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   696147b
 Built:        Fri Aug  7 17:36:36 UTC 2015
 OS/Arch:      linux/amd64
````

https://github.com/weaveworks/weave/issues/1285#issuecomment-130085857 shows a new `Mounts` section. And the `Volumes` section, which is what the proxy relies on, is gone.  Also note that it says `"RW": true` on all the mounts, which seems wrong!

### PR DESCRIPTION
**Pull request by paulbellamy**
Fixes #1287

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -82,7 +82,11 @@ func NewProxy(c Config) (*Proxy, error) {
 		Error.Fatalf("Could not configure tls for proxy: %s", err)
 	}
 
-	p.client, err = docker.NewClient(p.DockerAddr)
+	// We pin the protocol version to 1.15 (which corresponds to Docker 1.3.x;
+	// the earliest version supported by weave) in order to insulate ourselves
+	// from breaking changes to the API, as happened in 1.20 (Docker 1.8.0) when
+	// the presentation of volumes changed in inspect.
+	p.client, err = docker.NewVersionedClient(p.DockerAddr, "1.15")
 	if err != nil {
 		return nil, err
 	}
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -25,9 +25,8 @@ var (
 	containerStartRegexp  = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/[^/]*/(re)?start$")
 	execCreateRegexp      = regexp.MustCompile("^(/v[0-9\\.]*)?/containers/[^/]*/exec$")
 
-	ErrInvalidNetworkMode = errors.New("--net option")
-	ErrWeaveCIDRNone      = errors.New("WEAVE_CIDR=none")
-	ErrNoDefaultIPAM      = errors.New("--no-default-ipam option")
+	ErrWeaveCIDRNone = errors.New("WEAVE_CIDR=none")
+	ErrNoDefaultIPAM = errors.New("--no-default-ipam option")
 )
 
 type Proxy struct {
@@ -152,8 +151,9 @@ func (proxy *Proxy) ListenAndServe() error {
 func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *docker.HostConfig) ([]string, error) {
 	if hostConfig != nil &&
 		hostConfig.NetworkMode != "" &&
+		hostConfig.NetworkMode != "default" &&
 		hostConfig.NetworkMode != "bridge" {
-		return nil, ErrInvalidNetworkMode
+		return nil, fmt.Errorf("--net option: %q", hostConfig.NetworkMode)
 	}
 	for _, e := range config.Env {
 		if strings.HasPrefix(e, "WEAVE_CIDR=") {
--- a/test/610_proxy_wait_for_weave_test.sh
+++ b/test/610_proxy_wait_for_weave_test.sh
@@ -29,9 +29,9 @@ assert_raises "proxy docker_on $HOST1 exec c3 true"
 assert_raises "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.2/24' $BASE_IMAGE touch /w/w" 1
 
 # Check only user-specified volumes and /w are mounted
-proxy docker_on $HOST1 create --name c4 -v /tmp/1:/srv1 -v /tmp/2:/srv2 -e 'WEAVE_CIDR=10.2.1.3/24' $BASE_IMAGE $CHECK_ETHWE_UP
-proxy docker_on $HOST1 start -ai c4
-assert "docker_on $HOST1 inspect --format='{{range \$k, \$v := .Volumes}}{{\$k}} {{end}}' c4" "/srv1 /srv2 /w "
+dirs_sans_proxy=$(docker_on $HOST1 run --rm -it $BASE_IMAGE mount | wc -l)
+dirs_with_proxy=$(proxy docker_on $HOST1 run --rm -it -v /tmp/1:/srv1 -v /tmp/2:/srv2 -e 'WEAVE_CIDR=10.2.1.3/24' $BASE_IMAGE mount | wc -l)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: test/610_proxy_wait_for_weave_test.sh [hunk 1]
```diff
@@ -31,7 +31,7 @@ assert_raises "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.2/24' $BASE_IMAG
 # Check only user-specified volumes and /w are mounted
 proxy docker_on $HOST1 create --name c4 -v /tmp/1:/srv1 -v /tmp/2:/srv2 -e 'WEAVE_CIDR=10.2.1.3/24' $BASE_IMAGE $CHECK_ETHWE_UP
 proxy docker_on $HOST1 start -ai c4
-assert "docker_on $HOST1 inspect --format='{{range \$k, \$v := .Volumes}}{{\$k}} {{end}}' c4" "/srv1 /srv2 /w "
+assert "docker_on $HOST1 inspect --format='{{range .Mounts}}{{.Destination}} {{end}}' c4" "/srv1 /srv2 /w "
```
rade: isn't `Mounts` new in 1.8? i.e. are we switching from this test breaking on 1.8 to this test breaking on everything *except* 1.8?

test/run_all.sh
```diff
@@ -36,7 +36,7 @@ echo Running $TESTS
 for t in $TESTS; do
     echo
     greyly echo "---= Running $t =---"
-    (. $t)
+    . $t
```
rade: Is this at all related to this PR? If not, stick it on master.

proxy/proxy.go
```diff
@@ -148,8 +147,9 @@ func (proxy *Proxy) ListenAndServe() error {
 func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *docker.HostConfig) ([]string, error) {
 	if hostConfig != nil &&
 		hostConfig.NetworkMode != "" &&
+		hostConfig.NetworkMode != "default" &&
```
rade: I cannot find a mention of this new mode anywhere in the Docker API or CLI docs. Both list the modes as "bridge, host, and container:<name|id>"

Yet simply starting a container on Docker 1.8 does produce this mode in `docker inspect`. @mavenugo is the code wrong (i.e. "default" shouldn't be there) or the docs?

mavenugo: @rade default was added recently as part of the `default-network-driver` support in order to make a remote network plugin such as weave as the default network driver. an empty network-mode "" is considered seen as `default` instead of assuming it as `bridge` in order to make sure the plugin can be the default.
But if there is no `default-network-driver` mentioned in the daemon, then `bridge` is used by default.
The behavior is the same in the experimental channel right after 1.7 release.

rade: @mavenugo right, that's what I thought. So it's the Docker API and CLI docs that are wrong then.

test/610_proxy_wait_for_weave_test.sh
rade: Looks like CircleCI has confirmed that suspicion.

paulbellamy: Yes, but we should rebuild the gce templates to use docker 1.8 also.

rade: I'd rather not have tests that are so version sensitive.

rade: Here's a different way to perform the test:
- start a container w/o the proxy and run s.t. like `find / | wc -l`
- start a container with the proxy and a couple of volume mounts, and run `find / | wc -` in that
- check the counts differ by 3.

errordeveloper:
I've tested 6060451 with the [Docker platfrom guide](http://weave.works/guides/weave-and-docker-platform/chapter4/compose-scalable-swarm-cluster-with-weave.html) and it LGTM.

weave
```diff
@@ -882,7 +882,7 @@ ipam_cidrs_or_die() {
 # TODO: Handle args with spaces
 tls_arg() {
     filename=${3:-$2}
-    PROXY_VOLUMES="$PROXY_VOLUMES -v $1:/home/weave/tls/$filename.pem"
+    PROXY_VOLUMES="$PROXY_VOLUMES -v $1:/home/weave/tls/$filename.pem:ro"
```
rade: I've created #1293 for this since it has nothing to do with 1.8 compatibility and hence shouldn't be in this PR.

test/610_proxy_wait_for_weave_test.sh
```diff
@@ -29,9 +29,9 @@ assert_raises "proxy docker_on $HOST1 exec c3 true"
 assert_raises "proxy docker_on $HOST1 run -e 'WEAVE_CIDR=10.2.1.2/24' $BASE_IMAGE touch /w/w" 1
 
 # Check only user-specified volumes and /w are mounted
-proxy docker_on $HOST1 create --name c4 -v /tmp/1:/srv1 -v /tmp/2:/srv2 -e 'WEAVE_CIDR=10.2.1.3/24' $BASE_IMAGE $CHECK_ETHWE_UP
-proxy docker_on $HOST1 start -ai c4
-assert "docker_on $HOST1 inspect --format='{{range \$k, \$v := .Volumes}}{{\$k}} {{end}}' c4" "/srv1 /srv2 /w "
+dirs_sans_proxy=$(docker_on $HOST1 run --rm -it $BASE_IMAGE mount | wc -l)
+dirs_with_proxy=$(proxy docker_on $HOST1 run --rm -it -v /tmp/1:/srv1 -v /tmp/2:/srv2 -e 'WEAVE_CIDR=10.2.1.3/24' $BASE_IMAGE mount | wc -l)
+assert_raises "[[ '$((dirs_sans_proxy+3))' -eq '$dirs_with_proxy' ]]"
```
rade: `assert "$((dirs_san_proxy+3))" $dirs_with_proxy`?

rade: though admittedly that is the wrong way round.

`assert "$((dirs_with_proxy-3))" $dirs_sans_proxy`

would be better.

proxy/proxy.go
```diff
@@ -82,7 +81,7 @@ func NewProxy(c Config) (*Proxy, error) {
 		Error.Fatalf("Could not configure tls for proxy: %s", err)
 	}
 
-	p.client, err = docker.NewClient(p.DockerAddr)
+	p.client, err = docker.NewVersionedClient(p.DockerAddr, "1.15")
```
rade: I think this deserves a comment:

"We pin the protocol version to 1.15 (which corresponds to Docker 1.3.x; the earliest version supported by weave) in order to insulate ourselves from breaking changes to the API, as happened in 1.20 (Docker 1.8.0) when the presentation of volumes changed in `inspect`."

PR #1307, opened 2015-08-16
Compress recursive answers if they are too big.

### LINKED ISSUE #1306: "cannot unmarshal DNS message" using weavedns to resolve email-smtp.us-east-1.amazonaws.com
This bug shows a similar problem existed with consul dns:

https://github.com/golang/go/issues/11070

Looks like the following is happening:
- recursive dns is responding with a compressed answer, to fit in 512bytes
- miekg dns is uncompressing the answer transparently for us (https://github.com/miekg/dns/issues/216)
- we send the uncompresses answer on to the client, which then fails to parse it as its truncated.

tomwilkie:
Fixes #1306

### DIFF
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -239,6 +239,9 @@ func (d *DNSServer) handleRecursive(client *dns.Client, defaultMaxResponseSize i
 			}
 			d.ns.debugf("response: %+v", response)
 			response.Id = req.Id
+			if response.Len() > getMaxResponseSize(req, defaultMaxResponseSize) {
+				response.Compress = true
+			}
 			if err := w.WriteMsg(response); err != nil {
 				d.ns.infof("error responding: %v", err)
 			}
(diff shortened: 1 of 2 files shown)


rade:
Is it possible to write test for this that isn't horrific?

tomwilkie:
Yeah I think I could probably put together a simple-but-slightly-contrived unit test.  Let me have a go.

tomwilkie:
Actually, there seems to be know way to use the golang dns resolver with your own server - its hard wired to read /etc/resolv.conf, and I don't think rewriting that in a unit test is a good idea.

So we can have a test of the server side bits, but not the actual client error.

rade:
> Is it possible to write test for this that isn't horrific?

Hmm. 70 lines of test code for a 3 line change.

I've fixed some minor niggles and re-based. @tomwilkie I see this is assigned to you - can I merge it?

tomwilkie:
Yeah, test is a bit long.  I can probably make it shorter with some aggressive refactoring.  But I doubt its worth it.

Go ahead, merge away.

PR #1312, opened 2015-08-17
Add sudo apt-get update to circle.yml

### LINKED ISSUE #1311: CircleCI builds fail in package installation
e.g. https://circleci.com/gh/weaveworks/weave/2448

Got the following message in the CircleCI UI:

> If you recently started seeing "Unable to locate package" errors when installing apt packages in your build, this may be caused by our temporary measure to mitigate the outage of the ubuntu mirror that we are using. In order to fix this, please run "apt-get update" before installing any packages in your build.
> 
> We are using an ubuntu mirror that became inaccessible last Friday evening, PDT.
>
> We have implemented a temporary fix that replaces the broken repository with a good one. However, the fix requires you to run "apt-get update" because the apt metadata has been changed.
>
> We are very sorry for the trouble. We are still investigating the issue with the old u mirror and will update you as soon as we can.

tomwilkie:
Fixes #1311

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -26,6 +26,7 @@ dependencies:
         docker save weaveworks/weave-build >$WEAVE_BUILD;
       fi
   post:
+    - sudo apt-get update
     - sudo apt-get install bc jq
     - pip install requests
     - curl https://sdk.cloud.google.com | bash


inercia:
LGTM.

rade:
Shouldn't this have gone onto the 1.0 branch? i.e. won't builds there still fail?

inercia:
yes, @rade, I think this should be backported too...

rade:
> I think this should be backported to

*sigh*. We try to avoid backporting. Anyway, too late now. Just make sure you merge 1.0 back into master afterwards, so that master is always ahead of 1.0.

rade:
backported in d2205424915a9cbd998d52236eea4577c2c297d1, merged back into master in 8c9679b19c1dba647b1bd375f54114a862c399a7.

PR #1313, opened 2015-08-17
Try to compress too long replies (when forwarding).

### LINKED ISSUE #1306: "cannot unmarshal DNS message" using weavedns to resolve email-smtp.us-east-1.amazonaws.com

inercia:
Fixes #1306 for 1.0.2

### DIFF
--- a/nameserver/server_test.go
+++ b/nameserver/server_test.go
@@ -3,12 +3,14 @@ package nameserver
 import (
 	"encoding/binary"
 	"fmt"
+	"math/rand"
 	"net"
 	"testing"
 	"time"
 
 	"github.com/miekg/dns"
 	. "github.com/weaveworks/weave/common"
+	"github.com/weaveworks/weave/ipam/address"
 	wt "github.com/weaveworks/weave/testing"
 )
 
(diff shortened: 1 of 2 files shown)


nameserver/server_test.go
```diff
@@ -72,6 +75,18 @@ func TestUDPDNSServer(t *testing.T) {
...
+			header := makeHeader(m, &q, 300)
+			for {
+				if m.Len() > minUDPSize {
+					m.Compress = true
+					wt.AssertTrue(t, m.Len() <= minUDPSize, fmt.Sprintf("length: %d", m.Len()))
+					break
+				}
+				ip := address.Address(rand.Uint32()).IP4()
+				m.Answer = append(m.Answer, &dns.A{Hdr: *header, A: ip})
+			}
```
rade:
```
for m.Len() <= minUDPSize {
...
}
m.Compress = true
wt.AssertTrue(...)
```

inercia: Fixed. This was copy&pasted from #1307, so the same comment applies there...

rade:
Please rebase, so this gets the CircleCI fix.

PR #1314, opened 2015-08-17
Wait for ethwe via netlink, not via polling or signals

### LINKED ISSUE #1209: weaveproxy stub breaks Docker restart policies

### LINKED ISSUE #1258: containers fail to start correctly when networking options are passed on start
It turns out that the version 1.18 of Docker remote API allows `HostConfig` attribute to be set when starting a container and, moreover, Kubernetes actually happens to use it.

I have added a few of debug message to the proxy like this
```diff
diff --git a/proxy/create_container_interceptor.go b/proxy/create_container_interceptor.go
index e6942a5..e737532 100644
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -48,6 +48,9 @@ func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
                return err
        }
 
+       Log.Debugf("%#v", container)
+       Log.Debugf("%#v", container.HostConfig)
+
        if cidrs, err := i.proxy.weaveCIDRsFromConfig(container.Config, container.HostConfig); err != nil {
                Log.Infof("Ignoring container due to %s", err)
        } else {
diff --git a/proxy/start_container_interceptor.go b/proxy/start_container_interceptor.go
index 7353c07..f3b8e9c 100644
--- a/proxy/start_container_interceptor.go
+++ b/proxy/start_container_interceptor.go
@@ -21,6 +21,9 @@ func (i *startContainerInterceptor) InterceptResponse(r *http.Response) error {
                return err
        }
 
+       Log.Debugf("%#v", container)
+       Log.Debugf("%#v", container.HostConfig)
+
        cidrs, err := i.proxy.weaveCIDRsFromConfig(container.Config, container.HostConfig)
        if err != nil {
                Log.Infof("Ignoring container %s due to %s", container.ID, err)
```

and below is what I got.

```
INFO: 2015/08/04 14:28:53.636380 POST /containers/create?name=k8s_master.1681ebfb_redis-master-djojn_default_148c42d8-3ab5-11e5-8b8d-080027308234_c5875ebf
DEBU: 2015/08/04 14:28:53.636577 proxy.createContainerRequestBody{Config:(*docker.Config)(0xc209112d80), HostConfig:(*docker.HostConfig)(nil), MacAddress:""}
DEBU: 2015/08/04 14:28:53.636592 (*docker.HostConfig)(nil)
INFO: 2015/08/04 14:28:53.636598 Creating container with WEAVE_CIDR ""
```

we just create a container is expecting it to be wanting attached to weave

```
INFO: 2015/08/04 14:28:53.696899 GET /containers/json
DEBU: 2015/08/04 14:28:53.731191 Response from target: 200 OK map[Content-Length:[1183] Content-Type:[application/json] Date:[Tue, 04 Aug 2015 14:28:53 GMT]]
DEBU: 2015/08/04 14:28:53.745209 Response from target: 201 Created map[Date:[Tue, 04 Aug 2015 14:28:53 GMT] Content-Length:[90] Content-Type:[application/json]]
INFO: 2015/08/04 14:28:53.763274 POST /containers/33a47c80f51a37f574fc11921b7285839fd003380b01466fcfd24a25c17d37bb/start
```

we are now about to start it

```
INFO: 2015/08/04 14:28:53.833493 GET /containers/json
DEBU: 2015/08/04 14:28:53.843715 Response from target: 200 OK map[Content-Type:[application/json] Date:[Tue, 04 Aug 2015 14:28:53 GMT] Content-Length:[1562]]
DEBU: 2015/08/04 14:28:53.845850 &docker.Container{ID:"33a47c80f51a37f574fc11921b7285839fd003380b01466fcfd24a25c17d37bb", Created:time.Time{sec:63574295333, nsec:711737800, loc:(*time.Location)(0xf82480)}, Path:"/w/w", Args:[]string{"/entrypoint.sh", "redis-server"}, Config:(*docker.Config)(0xc209113200), State:docker.State{Running:true, Paused:false, Restarting:false, OOMKilled:false, Pid:5560, ExitCode:0, Error:"", StartedAt:time.Time{sec:63574295333, nsec:840536648, loc:(*time.Location)(0xf82480)}, FinishedAt:time.Time{sec:0, nsec:0, loc:(*time.Location)(0xf82480)}}, Image:"0ff407d5a7d9ed36acdf3e75de8cc127afecc9af234d05486be2981cdc01a38c", Node:(*docker.SwarmNode)(nil), NetworkSettings:(*docker.NetworkSettings)(0xc209645620), SysInitPath:"", ResolvConfPath:"/var/lib/docker/containers/d42839cdb477cec9decb0981cf972f40268b80ea3c0ca0a8ec5878baddef0cb6/resolv.conf", HostnamePath:"/var/lib/docker/containers/d42839cdb477cec9decb0981cf972f40268b80ea3c0ca0a8ec5878baddef0cb6/hostname", HostsPath:"/var/lib/docker/containers/d42839cdb477cec9decb0981cf972f40268b80ea3c0ca0a8ec5878baddef0cb6/hosts", Name:"/k8s_master.1681ebfb_redis-master-djojn_default_148c42d8-3ab5-11e5-8b8d-080027308234_c5875ebf", Driver:"overlay", Volumes:map[string]string{"/data":"/var/lib/docker/vfs/dir/2137d563501b2eef62c7ce701719056a99c9337d9e26b7c0ca30420192b243e4", "/dev/termination-log":"/var/lib/kubelet/pods/148c42d8-3ab5-11e5-8b8d-080027308234/containers/master/33a47c80f51a37f574fc11921b7285839fd003380b01466fcfd24a25c17d37bb", "/hostproc":"/proc", "/var/run":"/run", "/w":"/var/lib/docker/vfs/dir/4c11cc5c5962cfa4b67e8ea74c0efa542ef05e686f98cc6b8ea756be191ad9a7"}, VolumesRW:map[string]bool{"/data":true, "/dev/termination-log":true, "/hostproc":false, "/var/run":false, "/w":false}, HostConfig:(*docker.HostConfig)(0xc209134b40), ExecIDs:[]string(nil), RestartCount:0, AppArmorProfile:""}
DEBU: 2015/08/04 14:28:53.845965 &docker.HostConfig{Binds:[]string{"/var/lib/kubelet/pods/148c42d8-3ab5-11e5-8b8d-080027308234/containers/master/33a47c80f51a37f574fc11921b7285839fd003380b01466fcfd24a25c17d37bb:/dev/termination-log"}, CapAdd:[]string(nil), CapDrop:[]string(nil), ContainerIDFile:"", LxcConf:[]docker.KeyValuePair(nil), Privileged:false, PortBindings:map[docker.Port][]docker.PortBinding(nil), Links:[]string(nil), PublishAllPorts:false, DNS:[]string{"10.1.0.3", "10.0.2.3"}, DNSSearch:[]string{"default.svc.kube.local", "svc.kube.local", "kube.local"}, ExtraHosts:[]string(nil), VolumesFrom:[]string(nil), NetworkMode:"container:d42839cdb477cec9decb0981cf972f40268b80ea3c0ca0a8ec5878baddef0cb6", IpcMode:"container:d42839cdb477cec9decb0981cf972f40268b80ea3c0ca0a8ec5878baddef0cb6", PidMode:"", UTSMode:"", RestartPolicy:docker.RestartPolicy{Name:"", MaximumRetryCount:0}, Devices:[]docker.Device(nil), LogConfig:docker.LogConfig{Type:"json-file", Config:map[string]string(nil)}, ReadonlyRootfs:false, SecurityOpt:[]string(nil), CgroupParent:"", Memory:0, MemorySwap:-1, CPUShares:2, CPUSet:"", CPUQuota:0, CPUPeriod:0, Ulimits:[]docker.ULimit(nil)}
```

so now we are actually getting the `HostConfig.NetworkMode` passed in, hence the container is actually being ignored. However, `/w/w` is already there and waits for `SIGUSR2`...

```
INFO: 2015/08/04 14:28:53.845986 Ignoring container 33a47c80f51a37f574fc11921b7285839fd003380b01466fcfd24a25c17d37bb due to --net option
DEBU: 2015/08/04 14:28:53.845997 Response from target: 204 No Content map[Date:[Tue, 04 Aug 2015 14:28:53 GMT]]
INFO: 2015/08/04 14:28:53.855651 GET /containers/33a47c80f51a37f574fc11921b7285839fd003380b01466fcfd24a25c17d37bb/json
DEBU: 2015/08/04 14:28:53.857430 Calling weave[--local ps 33a47c80f51a37f574fc11921b7285839fd003380b01466fcfd24a25c17d37bb]
DEBU: 2015/08/04 14:28:53.916636 Response from target: 200 OK map[Date:[Tue, 04 Aug 2015 14:28:53 GMT] Content-Type:[application/json]]
```

The entry point was meant to be `/entrypoint.sh redis-server`, which is expect to turn into `redis-server` process, instead this is what appears in the process table

```
core@core-03 ~ $ ps -ef | grep redis-server
root      5560  1038  0 14:28 ?        00:00:00 /w/w /entrypoint.sh redis-server
core     19344  6012  0 14:36 pts/0    00:00:00 grep --colour=auto redis-server
```

### LINKED ISSUE #1300: Race between weavewait and proxy
The proxy is coded to start client containers via `weavewait`, then send them a `SIGUSR2` when everything is ready.  However the container start is asynchronous, so it is possible that the proxy manages to send the signal before `weavewait` is listening for it, which will result in a hang.

Comments in #1258 suggest this has been seen in real life.

bboreham:
Also move the implementation of proxy rewriting /etc/hosts into the `weave` script, to guarantee it gets done before the interface is marked up.

Fixes #1209. Fixes #1258. Fixes #1300.

### DIFF
--- a/net/if.go
+++ b/net/if.go
@@ -2,30 +2,24 @@ package net
 
 import (
 	"fmt"
-	"math"
 	"net"
-	"time"
+	"syscall"
+
+	"github.com/vishvananda/netlink/nl"
 )
 
-const (
-	sleepTime = 100 * time.Millisecond
-)
-
-// Wait `wait` seconds for an interface to come up. Pass zero to check once
-// and return immediately, or a negative value to wait indefinitely.
-func EnsureInterface(ifaceName string, wait int) (iface *net.Interface, err error) {
-	if iface, err = findInterface(ifaceName); err == nil || wait == 0 {
-		return
+// Wait for an interface to come up.
+func EnsureInterface(ifaceName string) (*net.Interface, error) {
+	s, err := nl.Subscribe(syscall.NETLINK_ROUTE, syscall.RTNLGRP_LINK)
+	if err != nil {
+		return nil, err
 	}
-	i := int64(math.MaxInt64)
-	if wait > 0 {
-		i = int64(time.Duration(wait) * time.Second / sleepTime)
+	defer s.Close()
+	if iface, err := findInterface(ifaceName); err == nil {
+		return iface, nil
 	}
-	for ; err != nil && i > 0; i-- {
-		time.Sleep(sleepTime)
-		iface, err = findInterface(ifaceName)
-	}
-	return
+	waitForIfUp(s, ifaceName)
+	return findInterface(ifaceName)
 }
 
 func findInterface(ifaceName string) (iface *net.Interface, err error) {
@@ -36,4 +30,32 @@ func findInterface(ifaceName string) (iface *net.Interface, err error) {
 		return iface, fmt.Errorf("Interface %s is not up", ifaceName)
 	}
 	return
+}
+
+func waitForIfUp(s *nl.NetlinkSocket, ifaceName string) error {
+	for {
+		msgs, err := s.Receive()
+		if err != nil {
+			return err
+		}
+		for _, m := range msgs {
+			switch m.Header.Type {
+			case syscall.RTM_NEWLINK: // receive this type for link 'up'
+				ifmsg := nl.DeserializeIfInfomsg(m.Data)
+				attrs, err := syscall.ParseNetlinkRouteAttr(&m)
+				if err != nil {
+					return err
+				}
+				name := ""
+				for _, attr := range attrs {
+					if attr.Attr.Type == syscall.IFA_LABEL {
+						name = string(attr.Value[:len(attr.Value)-1])
+					}
+				}
+				if ifaceName == name && ifmsg.Flags&syscall.IFF_UP != 0 {
+					return nil
+				}
+			}
+		}
+	}
 }
--- a/net/if.go
+++ b/net/if.go
@@ -26,7 +26,7 @@ func ensureInterface(s *nl.NetlinkSocket, ifaceName string) (*net.Interface, err
 	if iface, err := findInterface(ifaceName); err == nil {
 		return iface, nil
 	}
-	waitForIfUp(s, ifaceName)
+	netlinkReceiveUntil(s, matchIfName(ifaceName))
 	iface, err := findInterface(ifaceName)
 	return iface, err
 }
@@ -48,7 +48,7 @@ func EnsureInterfaceAndMcastRoute(ifaceName string) (*net.Interface, error) {
 	if CheckRouteExists(ifaceName, dest) {
 		return iface, err
 	}
-	waitForRoute(s, ifaceName, dest)
+	netlinkReceiveUntil(s, matchRoute(ifaceName, dest))
 	return iface, err
 }
 
@@ -62,30 +62,39 @@ func findInterface(ifaceName string) (iface *net.Interface, err error) {
 	return
 }
 
-func waitForIfUp(s *nl.NetlinkSocket, ifaceName string) error {
+func netlinkReceiveUntil(s *nl.NetlinkSocket, f func(syscall.NetlinkMessage) (bool, error)) error {
 	for {
 		msgs, err := s.Receive()
 		if err != nil {
 			return err
 		}
 		for _, m := range msgs {
-			switch m.Header.Type {
-			case syscall.RTM_NEWLINK: // receive this type for link 'up'
-				ifmsg := nl.DeserializeIfInfomsg(m.Data)
-				attrs, err := syscall.ParseNetlinkRouteAttr(&m)
-				if err != nil {
-					return err
-				}
-				name := ""
-				for _, attr := range attrs {
-					if attr.Attr.Type == syscall.IFA_LABEL {
-						name = string(attr.Value[:len(attr.Value)-1])
-					}
-				}
-				if ifaceName == name && ifmsg.Flags&syscall.IFF_UP != 0 {
-					return nil
-				}
+			if done, err := f(m); done {
+				return err
 			}
 		}
+	}
+}
+
+func matchIfName(ifaceName string) func(m syscall.NetlinkMessage) (bool, error) {
+	return func(m syscall.NetlinkMessage) (bool, error) {
+		switch m.Header.Type {
+		case syscall.RTM_NEWLINK: // receive this type for link 'up'
+			ifmsg := nl.DeserializeIfInfomsg(m.Data)
+			attrs, err := syscall.ParseNetlinkRouteAttr(&m)
+			if err != nil {
+				return true, err
+			}
+			name := ""
+			for _, attr := range attrs {
+				if attr.Attr.Type == syscall.IFA_LABEL {
+					name = string(attr.Value[:len(attr.Value)-1])
+				}
+			}
+			if ifaceName == name && ifmsg.Flags&syscall.IFF_UP != 0 {
+				return true, nil
+			}
+		}
+		return false, nil
 	}
 }
--- a/net/if.go
+++ b/net/if.go
@@ -15,11 +15,41 @@ func EnsureInterface(ifaceName string) (*net.Interface, error) {
 		return nil, err
 	}
 	defer s.Close()
+	iface, err := ensureInterface(s, ifaceName)
+	if err != nil {
+		return nil, err
+	}
+	return iface, err
+}
+
+func ensureInterface(s *nl.NetlinkSocket, ifaceName string) (*net.Interface, error) {
 	if iface, err := findInterface(ifaceName); err == nil {
 		return iface, nil
 	}
 	waitForIfUp(s, ifaceName)
-	return findInterface(ifaceName)
+	iface, err := findInterface(ifaceName)
+	return iface, err
+}
+
+// Wait for an interface to come up and have a route added to the multicast subnet.
+// This matches the behaviour in 'weave attach', which is the only context in which
+// we expect this to be called.  If you change one, change the other to match.
+func EnsureInterfaceAndMcastRoute(ifaceName string) (*net.Interface, error) {
+	s, err := nl.Subscribe(syscall.NETLINK_ROUTE, syscall.RTNLGRP_LINK, syscall.RTNLGRP_IPV4_ROUTE)
+	if err != nil {
+		return nil, err
+	}
+	defer s.Close()
+	iface, err := ensureInterface(s, ifaceName)
+	if err != nil {
+		return nil, err
+	}
+	dest := net.IPv4(224, 0, 0, 0)
+	if CheckRouteExists(ifaceName, dest) {
+		return iface, err
+	}
+	waitForRoute(s, ifaceName, dest)
+	return iface, err
 }
 
 func findInterface(ifaceName string) (iface *net.Interface, err error) {
--- a/proxy/create_exec_interceptor.go
+++ b/proxy/create_exec_interceptor.go
@@ -35,7 +35,7 @@ func (i *createExecInterceptor) InterceptRequest(r *http.Request) error {
 		Log.Infof("Leaving container %s alone because %s", container.ID, err)
 	} else if hasWeaveWait {
 		Log.Infof("Exec in container %s with WEAVE_CIDR \"%s\"", container.ID, strings.Join(cidrs, " "))
-		cmd := append(weaveWaitEntrypoint, "-s")
+		cmd := weaveWaitEntrypoint
 		options.Cmd = append(cmd, options.Cmd...)
 	}
 
--- a/proxy/create_exec_interceptor.go
+++ b/proxy/create_exec_interceptor.go
@@ -35,8 +35,7 @@ func (i *createExecInterceptor) InterceptRequest(r *http.Request) error {
 		Log.Infof("Leaving container %s alone because %s", container.ID, err)
 	} else if hasWeaveWait {
 		Log.Infof("Exec in container %s with WEAVE_CIDR \"%s\"", container.ID, strings.Join(cidrs, " "))
-		cmd := weaveWaitEntrypoint
-		options.Cmd = append(cmd, options.Cmd...)
+		options.Cmd = append(weaveWaitEntrypoint, options.Cmd...)
 	}
 
 	if err := marshalRequestBody(r, options); err != nil {
--- a/weave
+++ b/weave
@@ -1446,14 +1446,28 @@ case "$COMMAND" in
     attach)
         collect_cidr_args "$@"
         shift $CIDR_ARG_COUNT
-        if [ "$1" = "--or-die" ] ; then
-          ATTACH_TYPE="_or_die"
-          shift
-        fi
+        while [ $# -gt 0 ]; do
+            case "$1" in
+                --or-die)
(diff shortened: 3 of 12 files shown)


net/if.go
```diff
@@ -2,29 +2,24 @@ package net
...
-// Wait `wait` seconds for an interface to come up. Pass zero to check once
-// and return immediately, or a negative value to wait indefinitely.
-func EnsureInterface(ifaceName string, wait int) (iface *net.Interface, err error) {
-	if iface, err = findInterface(ifaceName); err == nil || wait == 0 {
-		return
+// Wait for an interface to come up.
+func EnsureInterface(ifaceName string) (iface *net.Interface, err error) {
+	s, err := nl.Subscribe(syscall.NETLINK_ROUTE, syscall.RTNLGRP_LINK)
+	if err != nil {
+		return nil, err
```
rade: just 'return' will do here. Alternatively, ditch the named returns altogether.

net/if.go
```diff
@@ -37,3 +32,32 @@ func findInterface(ifaceName string) (iface *net.Interface, err error) {
...
+					return err
+				}
+				name := ""
+				for _, attr := range attrs {
+					if attr.Attr.Type == syscall.IFA_LABEL {
+						name = string(attr.Value[:len(attr.Value)-1])
+					}
+				}
+				up := ifmsg.Flags&syscall.IFF_UP != 0
+				if ifaceName == name && up {
```
rade: Inline the previous line.

proxy/create_exec_interceptor.go
```diff
@@ -35,7 +35,7 @@ func (i *createExecInterceptor) InterceptRequest(r *http.Request) error {
 		Log.Infof("Ignoring container %s due to %s", container.ID, err)
 	} else if hasWeaveWait {
 		Log.Infof("Exec in container %s with WEAVE_CIDR \"%s\"", container.ID, strings.Join(cidrs, " "))
-		cmd := append(weaveWaitEntrypoint, "-s")
+		cmd := weaveWaitEntrypoint
```
rade: inline into the line below

weave
```diff
@@ -1382,14 +1382,28 @@ case "$COMMAND" in
...
+                    ;;
+            esac
+            shift
+        done
         [ $# -eq 1 ] || usage
         CONTAINER=$(container_id $1)
         create_bridge
         ipam_cidrs$ATTACH_TYPE $CONTAINER $CIDR_ARGS
+        if [ -n "$REWRITE_HOSTS" ] && command_exists weavehosts ; then
+            PATH_AND_NAME=$(docker inspect -f '{{.HostsPath}} {{.Config.Hostname}}' $CONTAINER)
```
rade: shouldn't we have similar error handling to `container_id` here?

weave
```diff
@@ -1382,14 +1382,28 @@ case "$COMMAND" in
...
+            weavehosts $PATH_AND_NAME $ALL_CIDRS
```
rade: What happens when the container dies during this? We need to make sure we produce an intelligible error message for that. As we do elsewhere, e.g. in `container_id`.

bboreham: You get the container ID echoed to `stdout` and the message "Container [long_hex_number] not running."  This message comes from the `with_container_netns` on the next line, i.e. these additions have not changed anything.

Docker does not mind you inspecting or writing the filesystem of a container that has exited.  Aha!, you say, what if you add `--rm`, but that is only possible when running via the proxy, in which case the container would still be in `weavewait` at this point.

bboreham:
There are some remaining problems with this.  We cannot set up everything on the interface before marking it 'up': Linux does not allow a route to be added for a down interface.  This leads to two races: the container may exit, destroying the interface, before we try the `route add` and `arping`, and the container may send a multicast packet before we have added the route.

Per discussion yesterday, I am going to pursue waiting for the route to be added, and re-arranging the weave script so that is the last thing that happens.

weave
bboreham: On the basis that `container_id` would detect most errors four lines earlier, and `docker inspect` is ok with inspecting exited containers, I think it's ok as is.

bboreham:
I have addressed the latest review comments, and commented out the part of the test that will fail until #1209 is fixed.

squaremo:
@rade Are you looking at this (as your comments suggest)? If so please assign yourself :)

bboreham:
I have submitted https://github.com/vishvananda/netlink/pull/46 to push the subscribe-to-netlink implementation upstream; if that PR is accepted then this one would be considerably simplified.

bboreham:
No sign of movement on https://github.com/vishvananda/netlink/pull/46; shall we merge as-is and fix up if/when it gets accepted?

rade:
> shall we merge as-is and fix up if/when it gets accepted?

Yes.

net/if.go
```diff
@@ -37,3 +61,40 @@ func findInterface(ifaceName string) (iface *net.Interface, err error) {
 	}
 	return
 }
+
+func netlinkDo(s *nl.NetlinkSocket, f func(syscall.NetlinkMessage) (bool, error)) error {
```
rade: `netlinkReceiveUntil`?

rade:
LGTM modulo the naming of `netlinkDo`.

bboreham:
Renamed, rebased, squashed (a bit).

rade:
Tests...

> Fixes #1209.

A test for this is being added in #1210. It doesn't make much sense to add a test here, since fixing #1209 in itself does not provide much useful functionality - the container restarts now, but w/o being attached to weave.

Have you checked this manually though?

> Fixes #1258.

Needs to be tested manually at least, and deserves an automated test IF one can be constructed w/o too much horror.

> Fixes #1300.

No easy way to write a test. And the new code quite obviously avoids the problem. However, it is worth manually testing, by inserting a `sleep` in weavewait, whether a) the reported problem is indeed real (i.e. breaks on master), and b) whether the new code does do the right thing when the everything is set up already (especially since that situation takes a different path through the new code).

errordeveloper:
>> Fixes #1258.

>Needs to be tested manually at least, to check that a) that issue is real (i.e. we can get things to break on master), and

I can test with master, if that's really needed.

> b) is fixed here.

I have tested this manually earlier, and Kubernetes woks fine with initial version of this that lives in a combined [`1199-proxy-kubernetes-2`](https://github.com/weaveworks/weave/tree/1199-proxy-kubernetes-2) branch.

The thing is that it's this plus #1210 that really fix the issue with Kubernetes.

Let me recap the cascade of issues, which I hope should clarify things a bit more.

1. `/w/w` would appear in each container and wait for the signal, which it never received, as on start proxy decides that that container doesn't belong to it due to `--net=container:<id>` and I filed #1258
2. @paulbellamy reworked the logic of sending the signal
3.  some containers would still have `/w/w` hanging around due to #1300 

It is doable to test each case separately and confirm the observations are the same, but I would also need to improve the testing procedure, as it's been taking me at least an hour to get results for each build the way I was doing it. I would also welcome anyone else to run my tests, but that could take even longer, I'm happy to explain once I speed it up.

>Deserves an automated test IF one can be constructed w/o too much horror.

Can be done, although it's really the combination of #1258 and #1300 that caused trouble and we cannot reliably test for #1300, while #1258 is a kind of weird beast due to it's backwards compatibility nature.

rade:
> I have tested this manually earlier, and Kubernetes works fine

I want to see this tested independently of Kubernetes.

errordeveloper:
> I want to see this tested independently of Kubernetes.

I've tried Python client and it actually doesn't allow `HostConfig` on start, the [Go one still does](https://godoc.org/github.com/fsouza/go-dockerclient#Client.StartContainer).

rade:
Could we test this with just curl and some json?

bboreham:
Testing for #1209:

On master, status is "Exited (137)":

````
$ sudo weave launch
$ sudo docker -H $DOCKER_HOST run -tid --restart=always ubuntu
178d375b4c5b2f1dfac5c95220b83741b92daf3993b169f5e1c12bce22706ed5
$ sudo kill -9 `sudo docker inspect --format='{{.State.Pid}}' 178d`
$ sudo docker ps -a
CONTAINER ID        IMAGE                         COMMAND                CREATED             STATUS                       PORTS                                                                                            NAMES
178d375b4c5b        ubuntu                        "/w/w /bin/bash"       32 seconds ago      Exited (137) 8 seconds ago                                                                                                    clever_bartik       
...
````

On branch 1209-wait-iface-netlink, status is "Up 3 seconds"

````
$ sudo weave launch
$ sudo docker -H $DOCKER_HOST run -tid --restart=always ubuntu
4134f5b514a74cf4cfc6aaa90de0ad62b2288402863c67542753dd055bf0c04b
$ sudo kill -9 `sudo docker inspect --format='{{.State.Pid}}' 4134`
bryan@bbvlx1:~/.../github.com/weaveworks/weave$ sudo docker ps -a
CONTAINER ID        IMAGE                         COMMAND                CREATED             STATUS                       PORTS                                                                                            NAMES
4134f5b514a7        ubuntu                        "/w/w /bin/bash"       18 seconds ago      Up 3 seconds                                                                                                                  sleepy_feynman      
...
````

**Comment by errordeveloper** [comment]
This seems to do the basic thing for me, didn't fire it agains the proxy yet:
```bash
POST="curl --silent --request POST --header Content-Type:application/json 192.168.48.11:2375"

r=$($POST/containers/create --data-binary @create_redis.json --data \
'{
  "Hostname": "",
  "Domainname": "",
  "User": "",
  "AttachStdin": false,
  "AttachStdout": true,
  "AttachStderr": true,
  "Tty": false,
  "OpenStdin": false,
  "StdinOnce": false,
  "Env": null,
  "Image": "redis",
  "WorkingDir": "",
  "NetworkDisabled": false,
  "HostConfig": {
    "NetworkMode": "bridge"
  }
}' | jq -r .Id)

$POST/containers/$r/start

t=$($POST/containers/create --data-binary @create_tester.json --data \
'{
  "Hostname": "",
  "Domainname": "",
  "User": "",
  "AttachStdin": false,
  "AttachStdout": true,
  "AttachStderr": true,
  "Tty": true,
  "OpenStdin": false,
  "StdinOnce": false,
  "Env": null,
  "Image": "busybox",
  "WorkingDir": "",
  "NetworkDisabled": false
}' | jq -r .Id)

$POST/containers/$t/start --data "$(printf '{ "NetworkMode": "container:%s" }' $r)"
```

bboreham:
I have added a smoke-test for #1258 that fails on master and passes on this branch.

bboreham:
I have manually tested #1300 - adding a 4-second sleep breaks completely on master and works fine on this branch.

I believe that completes the additional tests requested.

PR #1318, opened 2015-08-17
don't reconnect to ourself

### LINKED ISSUE #1305: Stop trying to connect to yourself and spamming the logs
I start weave and pass every peer the same set of IPs, which happens to include their own.  The logs are full of messages like this:

```
INFO: 2015/08/16 10:28:26.594151 ->[172.31.13.96:6783] attempting connection
INFO: 2015/08/16 10:28:26.594664 ->[172.17.42.1:50625] connection accepted
ERRO: 2015/08/16 10:28:26.595323 ->[172.31.13.96:6783|fe:09:5f:88:03:b4(ip-172-31-13-96)]: connection shutting down due to error: Cannot connect to ourself
ERRO: 2015/08/16 10:28:26.595593 ->[172.17.42.1:50625|fe:09:5f:88:03:b4(ip-172-31-13-96)]: connection shutting down due to error: Cannot connect to ourself
INFO: 2015/08/16 10:33:44.324598 ->[172.31.13.96:6783] attempting connection
INFO: 2015/08/16 10:33:44.324989 ->[172.17.42.1:50638] connection accepted
ERRO: 2015/08/16 10:33:44.325565 ->[172.31.13.96:6783|fe:09:5f:88:03:b4(ip-172-31-13-96)]: connection shutting down due to error: Cannot connect to ourself
ERRO: 2015/08/16 10:33:44.325803 ->[172.17.42.1:50638|fe:09:5f:88:03:b4(ip-172-31-13-96)]: connection shutting down due to error: Cannot connect to ourself
```

Could we either remember that the IP pointed to ourself, and stop trying to connect to it, or only log errors if they are different from last time?

rade:
When we discover that a connection target address leads to ourself we set the retry time to the zero time.Time value - the only special time.Time value. We use that marker to a) suppress subsequent
connection attempts, and b) display the next retry as 'never' in `weave status|report`.

Closes #1305.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -2,6 +2,7 @@ package router
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
@@ -27,6 +28,8 @@ const (
 	TieBreakLost
 	TieBreakTied
 )
+
+var ErrConnectToSelf = errors.New("Cannot connect to ourself")
 
 type RemoteConnection struct {
 	local         *Peer
@@ -437,7 +440,7 @@ func (conn *LocalConnection) registerRemote(remote *Peer, acceptNewPeer bool) er
 	}
 
 	if conn.remote == conn.local {
-		return fmt.Errorf("Cannot connect to ourself")
+		return ErrConnectToSelf
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


rade:
I decided against introducing another connection state because the five we have is too many already. The downside is that, as is the case now, in the summary information displayed by `weave status` one cannot tell the difference between connections that failed because they are self connections, and those that failed for a different reason.

We could exclude these failed self connections altogether from `weave status connections` and the count in `weave status` but that would make it impossible for users to discover that they have accidentally supplied a self address.

PR #1319, opened 2015-08-18
Expose dns server on tcp as well as udp

paulbellamy:
Issue was discovered when testing scope with master vs 1.0 branches.

@inercia, Any reason not to do this? AFAICT it initializes the same thing on tcp and udp, but just doesn't expose the tcp.

### DIFF
--- a/weave
+++ b/weave
@@ -1079,7 +1079,7 @@ case "$COMMAND" in
         # additional parameters, such as resource limits, to docker
         # when launching the weave container.
         DNS_CONTAINER=$(docker run --privileged -d --name=$DNS_CONTAINER_NAME \
-            -p $DOCKER_BRIDGE_IP:53:53/udp \
+            -p $DOCKER_BRIDGE_IP:53:53/udp -p $DOCKER_BRIDGE_IP:53:53/tcp \
             -e WEAVE_CIDR=none \
             -v /var/run/docker.sock:/var/run/docker.sock \
             $WEAVEDNS_DOCKER_ARGS $DNS_IMAGE -iface $CONTAINER_IFNAME -httpiface "$DNS_HTTP_IFNAME" "$@")


inercia:
@paulbellamy This is odd. We should definitively expose both ports.

inercia:
LGTM.

rade:
@paulbellamy please describe what problem this is solving. Presumably something to do with using weavedns outside of weave.

tomwilkie:
The new scope script uses the open tcp port 53 to detect weave DNS, as we can't use the presence of the weavedns container anymore.

squaremo:
> The new scope script uses the open tcp port 53 to detect weave DNS

That seems like it would be rather fragile.

squaremo:
>> The new scope script uses the open tcp port 53 to detect weave DNS
> 
> That seems like it would be rather fragile.

Checking that _something_ is listening on port 53 may give you false positives. If the next thing you do is to attempt a WeaveDNS operation -- e.g., adding an entry -- with the expectation that it could fail if WeaveDNS is not _actually_ running, then why don't you just attempt that operation in the first place?

rade:
Surely there is a real bug here, affecting weave. I would expect DNS resolution to be unable to retrieve large answers. Alas we have a test for that - 295_dns_large_response.sh - which succeeds. Something doesn't add up here.

tomwilkie:
The port mapping was missing from the 1.0 branch, but present on master, where the 295_dns_large_responses.sh test exists.  So the presence of the test doesn't imply there is definitely a bug.

Also, 1.0 only ever gave 'small' answers, so its unlikely anyone ever hit this.

rade:
> Also, 1.0 only ever gave 'small' answers, so its unlikely anyone ever hit this.

Except for recursive queries.

rade:
#1333.

The change lgtm. We could add a test, along the lines of #1333, but that is superfluous on master, so let's not bother.

PR #1325, opened 2015-08-18
deal with `weave --help|help` locally

### LINKED ISSUE #1322: weave --help should be more helpful when Docker is inaccessible
$ weave --help
    Post http:///var/run/weave.sock/v1.19/containers/create: dial unix /var/run/weave.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?

I think at some point in the past it used to say "weave must be run as root"

rade:
thus making them unaffected by docker breakages

Fixes #1322.


bboreham:
lgtm.

rade:
This is ready, I think. @bboreham please take another look.

bboreham:
lgtm again

PR #1298, opened 2015-08-13
Add a min docker version check

achanda:
Bailing out early is better UX
Properly fixes #1297

PR #1326, opened 2015-08-18
better docker version check

### LINKED ISSUE #576: dealing with old dockers in the weavexec era

rade:
Fixes #576. Supersedes #1298.

### DIFF
--- a/weave
+++ b/weave
@@ -11,6 +11,20 @@ else
 fi
 IMAGE_VERSION=${VERSION:-$IMAGE_VERSION}
 
+# - Docker versions prior to 1.2.0 cannot parse `--dns-search=.`
+# - Docker versions prior to 1.3.0 fail to include the domain when
+#   setting the hostname, so lookups for non-fq names fail.
+# - Docker version 1.3.0 contains the severe bug
+#   https://github.com/docker/docker/issues/8632
+# - The weavexec image embeds a Docker 1.3.1 client. Misleading
+#   version mismatch errors are produced if the user is running with a
+#   client *and* server with a lower version.
+# - Docker will complain if you run a client against a server older
+#   than that client.
+#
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -11,6 +11,18 @@ else
...
 
+# - Docker versions prior to 1.2.0 cannot parse `--dns-search=.`
+# - Docker versions prior to 1.3.0 fail to include the domain when
+#   setting the hostname, so lookups for non-fq names fail.
+# - Docker version 1.3.0 contains the severe bug
+#   https://github.com/docker/docker/issues/8632
+# - The weavexec image embeds a Docker 1.3.1 client. Misleading
+#   version mismatch errors are produced if the user is running with a
+#   client *and* server with a lower version.
+#
```
bboreham: Is there one more premise: that Docker will complain if you run a client against a server older than that client?

bboreham:
Apart from quibbling over the reasoning in a comment, looks ok to me.

PR #1327, opened 2015-08-18
`weave env --restore`

### LINKED ISSUE #1288: stopping weave results in error and breaks subsequent docker invocations
Technically not a bug, but most users will see it that way. Certainly annoys the hell out of me...

````
$ weave launch
$ eval $(weave env)
...do some weaving...
$ weave stop
error deleting container: Cannot connect to the Docker daemon. Is 'docker -d' running on this host?
Cannot connect to the Docker daemon. Is 'docker -d' running on this host?
$ docker ps
Cannot connect to the Docker daemon. Is 'docker -d' running on this host?
````

The fix is to `export DOCKER_HOST=<previous value>` prior to `weave stop`, which a) is non-obvious, and b) somewhat difficult if you don't know / didn't save the previous value.

### PR DESCRIPTION
**Pull request by rade**
`weave env` stashes the existing DOCKER_HOST in ORIG_DOCKER_HOST, and `weave env --restore` sets DOCKER_HOST to $ORIG_DOCKER_HOST.

Note that `weave env --restore` always executes locally, which means it will work when the env has a 'broken' DOCKER_HOST.

Closes #1288.

### DIFF
--- a/weave
+++ b/weave
@@ -41,7 +41,7 @@ weave launch-proxy  [-H <endpoint>] [--with-dns | --without-dns]
                       [--no-default-ipalloc] [--no-rewrite-hosts]
                       [--hostname-match <regexp>]
                       [--hostname-replacement <replacement>]
-weave env
+weave env           [--restore]
 weave config
 weave connect       [--replace] [<peer> ...]
 weave forget        <peer> ...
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rade** [comment]
@bboreham would appreciate some feedback here on what this does and how it does it.

Possible points of contention...

- `weave env --restore`

File: weave [hunk 1]
```diff
@@ -178,12 +178,19 @@ if [ "$1" = "run" ] ; then
...
+    true
+elif [ "$1" = "--local" ] ; then
+    shift
+else
     exec_remote "$@"
     exit $?
 fi
 
-shift 1
+if [ "$1" = "env" -a "$2" = "--restore" ] ; then
```
bboreham: could just put this ahead of the `if [ "$1" = "run" ]` and remove the extra shenanigans.

rade: no, because `weave --local env --restore` needs to work.

rade:
When `weave env --restore` is invoked without a `DOCKER_HOST_ORIG`, weave prints a warning on stderr and exits with 1. Alternatives I've considered: a) silent exit 0, b) silent exit 1, c) warning and exit 0. Anybody want to argue for a different choice than the one I made?

rade:
@bboreham I think this is ready modulo docs and squashing. Would appreciate some feedback first.

bboreham:
The implementation lgtm.

I am unclear how prevalent the set-up is where setting `DOCKER_HOST` in your regular environment just works.  It doesn't do anything for me because the sockets are protected so I have to `sudo` every command.

rade:
The new command also works in situations where `DOCKER_HOST` starts off unset/null. i.e. `weave env --restore` sets it to back to blank. (NB: minor wart: it is not unsetting DOCKER_HOST, only setting it to null).

rade:
Umm. This wasn't ready to merge. Still needs docs!

PR #1330, opened 2015-08-19
plug GossipSender leak

### LINKED ISSUE #1329: memory leak when IPAM fails with "Received invalid state update!"
From https://github.com/weaveworks/weave/issues/1178#issuecomment-132323364

Running
```
host1$ sudo weave launch-router
host1$ sudo weave run -ti ubuntu
host2$ sudo weave launch-router
host2$ sudo weave run -ti ubuntu
host1$ sudo weave connect host2
```
results in an ever increasing memory usage of the two weave routers.

rade:
GossipChannel.sendDown is the (only) place which creates GossipSenders. It has two call sites, Send and SendDown, only the former of which performed a gc of existing senders. Therefore, in configurations with

- a lot of connection churn, and
- SendDown invocations for these connections, and
- few or no Send invocations

we can leak GossipSenders.

The most likely scenario for this to occur is a peer that, while having no other connections, is making/accepting connections which get torn down post handshake, e.g. because IPAM detected a clash.

Fixes #1329.

### DIFF
--- a/router/gossip_channel.go
+++ b/router/gossip_channel.go
@@ -113,17 +113,32 @@ func (c *GossipChannel) Send(srcName PeerName, data GossipData) {
 	connections := c.ourself.Connections()
 	c.Lock()
 	defer c.Unlock()
-	// GC - randomly (courtesy of go's map iterator) pick some
-	// existing entries and stop&remove them if the associated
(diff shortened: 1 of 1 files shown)


bboreham:
lgtm.

PR #1331, opened 2015-08-19
clarify test 160

bboreham:
The point of the last few lines of test 160 is to see whether starting weave in a state where it thinks it should talk to a currently-stopped peer will hang it.  This change makes this clearer (but does not run the launch command in a `timeout` wrapper because `weave_on` is a shell function and it's too complicated to extract it out for `timeout` to call).

### DIFF
--- a/test/160_claim_3_test.sh
+++ b/test/160_claim_3_test.sh
@@ -32,6 +32,7 @@ stop_router_on $HOST1
 stop_router_on $HOST2
 
 # Now make host1 attempt to claim from host2, when host2 is stopped
+# the point being to check whether host1 will hang trying to talk to host2
 weave_on $HOST2 launch-router
 # Introduce host3 to remember the IPAM CRDT when we stop host2
 weave_on $HOST3 launch-router $HOST2


test/160_claim_3_test.sh
```diff
@@ -32,10 +32,11 @@ stop_router_on $HOST1
 stop_router_on $HOST2
 
 # Now make host1 attempt to claim from host2, when host2 is stopped
+# checking whether host1 will hang trying to talk to host2
 weave_on $HOST2 launch-router
 # Introduce host3 to remember the IPAM CRDT when we stop host2
 weave_on $HOST3 launch-router $HOST2
 stop_router_on $HOST2
-weave_on $HOST1 launch-router $HOST3
+assert_raises "DOCKER_HOST=tcp://$HOST1:$DOCKER_PORT timeout 30 $WEAVE launch-router $HOST3"
```
rade: The unravelling of of `weave_on` is rather unfortunate. Would a sub-shell work?

bboreham: Indeed, it was a bit finicky to get right.
On the basis that `timeout` is an external command, I could only see it working if `weave_on` were extracted to a separate shell script.

rade: Hmm. I'm not sure this is worthwhile. I'd leave the code alone and only add the explanatory comment.

PR #1349, opened 2015-08-21
Warn on detection of REJECT rule in FORWARD chain

### LINKED ISSUE #1266: Weave on a single centos7 host, Containers cant resolve each other
I have a single weave host, with two containers running on it.
I started weave
`weave launch && weave launch-dns && weave launch-proxy`
 and then started the containers with docker commands, 
`docker run -d -ti -h webnode1 --name webnode1 -p 80:80 docker.io/httpd:latest`
`docker run -d -ti -h webnode2 --name webnode2 -p 8080:80 docker.io/httpd:latest`

but i am not able to ping them from one another and then the traffic to the outside world is not passed through.

`docker exec webnode1 ping -c 5 webnode2` results in unknown host webnode2
Running the latest version of weave 1.0.1, docker is also latest version.

awh:
Closes #1266.

### DIFF
--- a/weave
+++ b/weave
@@ -155,6 +155,18 @@ enforce_docker_bridge_addr_assign_type() {
     fi
 }
 
+check_forwarding_rules() {
+    if run_iptables -C FORWARD -j REJECT --reject-with icmp-host-prohibited > /dev/null 2>&1; then
+        cat >&2 <<EOF
+WARNING: existing iptables rule
+
+    '-A FORWARD -j REJECT --reject-with icmp-host-prohibited'
+
+will block name resolution via weaveDNS - please reconfigure your firewall.
+EOF
+    fi
+}
+
 is_cidr() {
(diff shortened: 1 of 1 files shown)


awh:
Suggestions for a better error message welcome...

rade:
This should tell the user a) *exactly* what the offending rule is, b) what the likely effect is on weave.

awh:
Message updated.

awh:
@rade latest revision:
````
$ ./weave launch
WARNING: existing iptables rule

    '-A FORWARD -j REJECT --reject-with icmp-host-prohibited'

will block name resolution via weaveDNS - please reconfigure your firewall.
````

alex-sherwin:
You should give an example command of removing the rule in the warning..

joserivca:
so it was fixed? im still having the problem!

PR #1350, opened 2015-08-21
Add GET /ip handler to return all IPAM mappings.

tomwilkie:
Scope needs to know the IP addresses of containers.

### DIFF
--- a/ipam/http.go
+++ b/ipam/http.go
@@ -1,6 +1,7 @@
 package ipam
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


rade:
> Scope needs to know the IP addresses of containers.

...but what you have implemented only returns the automatically allocated addresses.

What's wrong with `weave ps`?

tomwilkie:
I don't want to repeatedly invoke a shell script; the idea would be for IPAM to also record the IP addresses of non-allocated containers.  Its already told about them by the weave script and the proxy, so 'all' thats needed is for IPAM to not ignore the requests, as it does in https://github.com/weaveworks/weave/blob/master/ipam/claim.go#L34

What do you think?

rade:
>  the idea would be for IPAM to also record the IP addresses of non-allocated containers.

IPAM should not get involved in anything outside its range. Introduces an artificial coupling.

> I don't want to repeatedly invoke a shell script

That doesn't strike me as all that bad.

squaremo:
> I don't want to repeatedly invoke a shell script
> 
> That doesn't strike me as all that bad.

NB in the case of e.g., the docker plugin, the interfaces are not named (exactly) `"ethwe"`; i.e., `weave ps` assumes interfaces are assigned by the weave script. It might be a reasonable assumption for the minute, but probably won't remain so.

tomwilkie:
Using the output of ```weave ps``` was a pain.  Had to include the weave script and the docker cli in the scope image.  And have to parse the output of the script.  Yuk.

https://github.com/weaveworks/scope/pull/386

rade:
> Using the output of weave ps was a pain

doesn't look all that painful.

>  Had to include the weave script and the docker cli in the scope image

That is somewhat annoying, but otoh it means you can now also invoke any other `weave` commands.

Can we close this?

bboreham:
I need a function like this for debugging, reasonably often.

PR #1351, opened 2015-08-24
[proxy] make sure we close net/http.Response.Body, so we don't leak memory

paulbellamy:


tomwilkie:
This looks fine to me, but I don't know the code/problem well enough to thoroughly review.

paulbellamy:
TL;DR of the issue is that in certain cases a `net/http.Response.Body` actually holds a reference to a goroutine, which can leak. e.g. https://github.com/golang/go/blob/master/src/net/http/client.go#L408

paulbellamy:
It's not an issue with incoming requests, because the http server closes those when the handler returns.

squaremo:
We do `client.Close()`, is that not enough? (If it's not enough, why _doesn't_ the body get closed in the branches where the connection is hijacked)

rade:
Closed because this isn't actually an issue? Or closed because the fix is wrong?

paulbellamy:
Closed because @squaremo is correct, we do `client.Close()` already, so wasn't actually an issue.

PR #1352, opened 2015-08-24
Small improvements in features doc

bboreham:
Escape a minus character to stop it being taken as a bullet, and clarify which example we are talking about.


squaremo:
Looks fine, ta

PR #1355, opened 2015-08-24
Pass docker bridge ip to dns server, and filter it out of upstream dns ips.

### LINKED ISSUE #1343: DNS crashes if host resolv.conf contains docker0 bridge ip
We are trying to resolve container dns names from host by adding weavedns ip (docker0 ip) to /etc/resolv.conf. This seems to create circular dependency inside weavedns because it sees docker0 ip as an upstream server.

tomwilkie:
Flag is called ```--dns-effective-listen-address```, and is automatically populated by the ```weave``` script.

Fixes #1343 

Note we don't actually validate the contents of this field (ie check its a real IP address).  We can do, but I didn't see the point.

### DIFF
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -38,8 +38,18 @@ type DNSServer struct {
 	udpClient *dns.Client
 }
 
-func NewDNSServer(ns *Nameserver, domain string, address string, ttl uint32, clientTimeout time.Duration) (*DNSServer, error) {
+func filter(ss []string, s string) []string {
+	for i := 0; i < len(ss); {
+		if ss[i] == s {
+			ss = append(ss[:i], ss[i+1:]...)
+			continue
+		}
+		i++
+	}
+	return ss
+}
 
+func NewDNSServer(ns *Nameserver, domain, address, effectiveAddress string, ttl uint32, clientTimeout time.Duration) (*DNSServer, error) {
 	s := &DNSServer{
 		ns:        ns,
 		domain:    dns.Fqdn(domain),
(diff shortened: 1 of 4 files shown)


weave
```diff
@@ -1162,7 +1162,7 @@ launch_router() {
         -v /var/run/docker.sock:/var/run/docker.sock \
         -e WEAVE_PASSWORD \
         -e WEAVE_CIDR=none \
-        $WEAVE_DOCKER_ARGS $IMAGE $COVERAGE_ARGS --iface $CONTAINER_IFNAME --port $CONTAINER_PORT --name "$PEERNAME" --nickname "$(hostname)" --ipalloc-range "$IPRANGE" "$@")
+        $WEAVE_DOCKER_ARGS $IMAGE $COVERAGE_ARGS --iface $CONTAINER_IFNAME --port $CONTAINER_PORT --name "$PEERNAME" --nickname "$(hostname)" --ipalloc-range "$IPRANGE" --dns-effective-listen-address "$DOCKER_BRIDGE_IP" "$@")
```
bboreham: Maybe this could go into `$DNS_PORT_MAPPING`, to keep like things together and so it will be blanked on `--no-dns`?

tomwilkie: Good idea.

tomwilkie: Actually DNS_PORT_MAPPING are args for docker, these are args for the router itself.

Is it important to blank it when --no-dns?

bboreham: Ah.  I don't imagine it would be important to blank it; just that it will show up in logs for people who add `--no-dns`.

nameserver/dns.go
```diff
@@ -52,6 +62,9 @@ func NewDNSServer(ns *Nameserver, domain string, address string, ttl uint32, cli
 	if s.upstream, err = dns.ClientConfigFromFile(etcResolvConf); err != nil {
 		return nil, err
 	}
+	if s.upstream != nil {
```
inercia: Maybe `if s.upstream != nil && len(effectiveAddress) > 0`, for saving some cycles...

inercia:
LGTM.

PR #1359, opened 2015-08-25
Get proxy to re-attach weave router if it restarts

bboreham:
This is a follow-on to #1210, created as a separate PR for clarity.

Fixes part (a) in #401

You get a bit of noise in the logs on a total restart (e.g. a reboot), because the router isn't necessarily up and listening before we start re-attaching other containers.  But it all works out in the end, because the router attach re-does all the parts that went wrong.

### DIFF
--- a/weave
+++ b/weave
@@ -512,6 +512,10 @@ launch() {
         echo "Perhaps you are running the docker daemon with container networking disabled (-b=none)." >&2
         return 1
     fi
+    # No-op if already attached
+    if netnsenter ip link show $CONTAINER_IFNAME >/dev/null 2>&1 ; then
+        return 0
+    fi
     connect_container_to_bridge &&
         netnsenter ethtool -K eth0 tx off >/dev/null &&
         netnsenter ip link set $CONTAINER_IFNAME up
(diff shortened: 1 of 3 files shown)


weave
```diff
@@ -1177,6 +1181,17 @@ launch_router() {
         -e WEAVE_PASSWORD \
         -e WEAVE_CIDR=none \
         $WEAVE_DOCKER_ARGS $IMAGE $COVERAGE_ARGS --iface $CONTAINER_IFNAME --port $CONTAINER_PORT --name "$PEERNAME" --nickname "$(hostname)" --ipalloc-range "$IPRANGE" --dns-effective-listen-address "$DOCKER_BRIDGE_IP" "$@")
+}
+
+# Recreate the parameter values that are set when the router is first launched
+fetch_router_args() {
+    IPRANGE=$(docker inspect -f '{{.Args}}' $1 | grep -o -e '-ipalloc-range [^ ]*')
```
paulbellamy: Won't this miss `--ipalloc-range 192.whatever`?

weave
```diff
@@ -1177,6 +1181,17 @@ launch_router() {
...
+    # This doesn't get back the right syntax, but it's enough for our purposes
+    DNS_PORT_MAPPING=$(docker inspect -f '{{with index .HostConfig.PortBindings "53/udp"}}{{.}}{{end}}' $1)
+}
```
rade: Isn't the use of  `IPRANGE` and `DNS_PORT_MAPPING` in `attach_router()` purely an optimisation? And one that these days covers edge cases only? Could we get rid of it, i.e. *always* call `ipam_reclaim` and `put_dns_fqdn` for all weave-networked containers?

weave
bboreham: The args are dumped as a sequence of name/value pairs, all space-separated.  I don't see that an extra dash would break it, if that's what you meant.

Since, in this code, we only need the presence/absence of `IPRANGE` and `DNS_PORT_MAPPING`, it is enough.  Having the script call an http interface on the router would also be fine, but would need to be implemented.

Currently, the proxy does not call the router at all; that seems a step too far to introduce here.

weave
```diff
@@ -1345,14 +1360,24 @@ case "$COMMAND" in
...
         ;;
     launch-router)
         deprecation_warnings "$@"
         check_not_running $CONTAINER_NAME $BASE_IMAGE
         launch_router "$@"
+        attach_router $CONTAINER_NAME
         echo $ROUTER_CONTAINER
         ;;
+    attach-router)
+        deprecation_warnings "$@"
```
rade: that `deprecation_warnings ...` shouldn't be there.

weave
paulbellamy: I was more concerned about the `-key=value` vs `-key value`, not the number of dashes.

I agree about having the proxy call the router.

weave
bboreham: `ipam_reclaim` would get 401s if `IPRANGE` is blank, since IPAM would be disabled.

`DNS_PORT_MAPPING` doesn't have this problem since DNS can't be disabled.

rade: DNS *can* be disabled.

bboreham: I stand corrected.

bboreham:
Replaced by #1518

PR #1360, opened 2015-08-25
Pin the version of docker-py image in testing

bboreham:
At the time of writing, all smoke-test runs are failing because of a bug in docker-py.
This change pins the version to (a) work round this and (b) avoid surprises in future.
We should update the pinned version periodically to ensure we keep up with the Docker API.

The image we use, `joffrey/docker-py`, has very few tags, and no recent ones work with weave's smoke-tests, so I created a new image `weaveworks/docker-py:pinned`, built from https://github.com/docker/docker-py/commit/8ba85ac26f08521a96aab7be1e358a4aec709d32

### DIFF
--- a/test/600_proxy_docker_py.sh
+++ b/test/600_proxy_docker_py.sh
@@ -3,6 +3,9 @@
 
 . ./config.sh
 
+# pin the version to avoid being blind-sided by incompatible changes
+DOCKER_PY=weaveworks/docker-py:pinned
+
 docker_py_test() {
     SHARD=$1
     TOTAL_SHARDS=$2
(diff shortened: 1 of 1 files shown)


rade:
This is a recipe for us missing important changes to the Docker API that break the proxy, so negates half of the purpose of running this test suite in the first place (the other is to catch breaking changes in the proxy).

So I don't think we should do this. Hopefully the [root cause](https://github.com/docker/docker-py/pull/738#issuecomment-135316933) of these failures can be addressed instead.

bboreham:
My impression is that they do run CI, but it was not covering the same cases we rely on.
A subsequent change, https://github.com/docker/docker-py/commit/b5f1e64e8f18b313019fd47a431115cf881873e8, broke our use of the test in a different way, and I couldn't quickly figure out how to recover from that.  I thought I'd let their changes settle down and then get it back in shape.

rade:
> My impression is that they do run CI, but it was not covering the same cases we rely on.

We are not doing anything unusual; just run the integration test suite. That should be made part of the docker-py CI.

PR #1361, opened 2015-08-26
Pin version of docker-py image

bboreham:
Parallel change to #1360, for the 1.0 branch.

The file has changed a lot (including its name) on master, so a merge of this change to master would be equivalent to re-implementing #1360.


rade:
This will still need to be merged into master, since master must always be
strictly ahead of 1.0. If you merge #1360 first then merging this will
simply entail throwing away any changes.

PR #1365, opened 2015-08-27
ignore interface deletion errors during `weave reset`

### LINKED ISSUE #1356: Weave reset fails
When running /weave --local reset the command fails with the following error message: 

Cannot find device "vethwepl14191@if322"

Looking at the weave code you always try to escape such error by adding `|| true` at the end so we don't understand why we are stuck on that error.

rade:
These can occur when a container dies during `weave reset`.

Fixes #1356.


squaremo:
I'm not sure that more smothering errors is good in general. But here, I guess the worst that can happen is that the deletion fails because of a reason other than it not existing, and there's a bit of garbage left behind.

PR #1367, opened 2015-08-27
Fix docker api proxy smoke-tests after docker-py change (on 1.0 branch)

bboreham:
Docker-py was changed upstream to run using a test harness Pytest.

Also I have seen some failures due to the 'busybox' image not being found on the test host.

### DIFF
--- a/test/600_proxy_docker_py_test.sh
+++ b/test/600_proxy_docker_py_test.sh
@@ -5,6 +5,7 @@
 start_suite "Run docker-py test suite against the proxy"
 
 docker_on $HOST1 pull joffrey/docker-py >/dev/null
+docker_on $HOST1 pull busybox >/dev/null
 
 weave_on $HOST1 launch-proxy --no-default-ipam
 
--- a/test/600_proxy_docker_py_test.sh
+++ b/test/600_proxy_docker_py_test.sh
@@ -5,6 +5,7 @@
 start_suite "Run docker-py test suite against the proxy"
 
 docker_on $HOST1 pull joffrey/docker-py >/dev/null
+# workaround for https://github.com/docker/docker-py/issues/745
 docker_on $HOST1 pull busybox >/dev/null
 
 weave_on $HOST1 launch-proxy --no-default-ipam
--- a/test/600_proxy_docker_py_test.sh
+++ b/test/600_proxy_docker_py_test.sh
@@ -13,7 +13,7 @@ if docker_on $HOST1 run \
     -e DOCKER_HOST=tcp://172.17.42.1:12375 \
     -v /tmp:/tmp \
     -v /var/run/docker.sock:/var/run/docker.sock \
-    joffrey/docker-py python tests/integration_test.py ; then
+    joffrey/docker-py py.test tests/integration_test.py ; then
     assert_raises "true"
 else
     assert_raises "false"
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -45,6 +45,7 @@ func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
 		return err
 	}
 	r.Body.Close()
+	r.Body = ioutil.NopCloser(bytes.NewReader(body))
 
 	container := createContainerRequestBody{}
 	if err := json.Unmarshal(body, &container); err != nil {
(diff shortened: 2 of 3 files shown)


bboreham:
I have no idea why particular test is failing on 1.0 and not on master:

````
>       self.assertEqual(container_log_config['Config'], {})
E       AssertionError: None != {}
````

**Comment by rade** [comment]
> `self.assertEqual(container_log_config['Config'], {})`

So that's in [test_valid_no_config_specified](https://github.com/docker/docker-py/blame/db1a93fd27a7627300c1fd54040d2f727ada78a4/tests/integration_test.py#L305), which was introduced [recently](https://github.com/docker/docker-py/commit/9937a31a88b9a5de248bd631cc9eb1a83c48f56f).

**Comment by rade** [comment]
@paulbellamy can you look into that "breaks on 1.0, fine on master" issue? I've spent a few minutes checking whether anything significant has changed between 1.0 and master in how we manipulate the HostConfig (which is where the log config resides) on create, but it all looks pretty similar.

**Comment by paulbellamy** [comment]
@bboreham it fails/passes on 1.0/master without your changes? or with your changes?

**Comment by paulbellamy** [comment]
I believe it is because in the `createContainerInterceptor` we decode, modify, and re-encode the request. Because go has limited nil/zero semantics the go dockerclient encodes the request slightly differently. Sending a `null` instead of `{}`.

To completely avoid this we would have to refrain from parsing the request using the go docker client, and just parse it as a set of maps (which would all-around be safer anyway). However that is a bigger change than should go in 1.0.

**Comment by rade** [comment]
> I believe it is because in the `createContainerInterceptor` we decode, modify, and re-encode the request.

That's exactly the kind of thing I was looking for in the proxy code, but afaict we do that in both 1.0 and master, so it doesn't explain the difference in behaviour.

**Comment by paulbellamy** [comment]
There is no change. It also fails on master, but master's test matching regex is wrong.

**Comment by rade** [comment]
> master's test matching regex is wrong

File an issue.

> To completely avoid this we would have to refrain from parsing the request using the go docker client, and just parse it as a set of maps (which would all-around be safer anyway).

Is that what docker itself does?

I do rather wonder whether the test is too strict, or even just wrong. Note that it actually sends `None` and expects to get back `{}`. Which is weird.

**Comment by paulbellamy** [comment]
Even when you say `None` python still sends `{}`.

**Comment by paulbellamy** [comment]
It's a "bug" in go's `encoding/json`:
```Go
// decode the json into a null map
var m map[string]string
json.Unmarshal([]byte("{}"), m)
bs, _ := json.Marshal(m)
fmt.Printf("%q\n", bs) // => "null"

// decode the same json into an existing map
m = make(map[string]string)
json.Unmarshal([]byte("{}"), m)
bs, _ = json.Marshal(m)
fmt.Printf("%q\n", bs) // => "{}"
```

i.e. the result of parsing a json map depends on whether the thing you are decoding into is nil, which it is in go-dockerclient.

But, the distinction is rarely important.

rade:
So how do the docker client and docker itself differ in decoding the request?

paulbellamy:
Hold on actually. Needs more research..

paulbellamy:
Ok, have added a temporary fix for these tests, where we don't replace the request body unless we modify it. Have opened #1371 to deal with the underlying issue, which is a bigger change than I would like for 1.0.

test/600_proxy_docker_py_test.sh
```diff
@@ -5,6 +5,7 @@
 start_suite "Run docker-py test suite against the proxy"
 
 docker_on $HOST1 pull joffrey/docker-py >/dev/null
+docker_on $HOST1 pull busybox >/dev/null
```
rade: probably worth adding a comment that this is a workaround for https://github.com/docker/docker-py/issues/745

rade:
@paulbellamy @bboreham I have made a couple of small tweaks. Please check they are ok. Then squash and merge and do the 1.0->master merge.  All LGTM :)

test/600_proxy_docker_py_test.sh
```diff
@@ -13,7 +15,7 @@ if docker_on $HOST1 run \
     -e DOCKER_HOST=tcp://172.17.42.1:12375 \
     -v /tmp:/tmp \
     -v /var/run/docker.sock:/var/run/docker.sock \
-    joffrey/docker-py python tests/integration_test.py ; then
+    joffrey/docker-py py.test tests/integration_test.py ; then
```
bboreham: I think this should have been updated in line with https://github.com/weaveworks/weave/commit/54833195edf044b8f1f97c7d9de6757c52edccda

bboreham: Sorry, ignore me, 1.0 works differently.

PR #1370, opened 2015-08-28
Looser docker-py test matcher for parallelization

paulbellamy:
Previous one was only catching TestFoo, not FooTest. Python uses both.

~~Will expose a broken the test.~~

Found in https://github.com/weaveworks/weave/pull/1367

Depends on #1367

### DIFF
--- a/test/600_proxy_docker_py.sh
+++ b/test/600_proxy_docker_py.sh
@@ -13,13 +13,10 @@ docker_py_test() {
     docker_on $HOST1 pull busybox >/dev/null
     # Get a list of the tests for use to shard
     docker_on $HOST1 pull joffrey/docker-py >/dev/null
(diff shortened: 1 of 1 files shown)


bboreham:
According to the [docs](http://pytest.org/latest/goodpractises.html#test-discovery), `FooTest` shouldn't match.

The test that @rade noted was `test_valid_no_config_specified`, for which we would just want to change `Test` to  `[Tt]est` in the regex.

paulbellamy:
Right, ok. Probably worth fixing in docker-py then. The regex is matching class names, not the method names.

bboreham:
On closer inspection, the test is using [pytest unittest integration](http://pytest.org/latest/unittest.html#support-for-unittest-testcase-integration-of-fixtures), so maybe matching `class .*TestCase` is closer to what we want.

PR #1372, opened 2015-08-28
Support draft and pre-releases in `bin/release`

### LINKED ISSUE #880: Update `bin/release` to create releases in draft state
From `bin/release`:
````
    echo "== Creating GitHub release $RELEASE_NAME $VERSION"
    # This cannot be done as a draft because of a bug in
    # github-release: https://github.com/aktau/github-release/issues/7
    github-release release \
        --user $GITHUB_USER \
        --repo weave \
        --tag $LATEST_TAG \
        --name "$RELEASE_NAME $VERSION" \
        --description "$RELEASE_DESCRIPTION"
````
aktau/github-release#7 has been addressed, so we can update our release script accordingly.

### LINKED ISSUE #909: make the release script support non-mainline releases
The release script contains various sanity checks for latest_release, and updates latest_release.  That's great... until you want to do a non-mainline release.

It also assumes that release tags start with 'v'.  That's conventional for mainline releases, but not necessarily what you want for preview releases etc.

I'm sure we all wish for preview releases to be a rarity, but if they do happen from time to time, it would be nice to use the release script for them with less brute force involved.

### PR DESCRIPTION
**Pull request by awh**
This PR refactors the `bin/release` script to have three steps:

````
bin/release build
bin/release draft [--pre-release]
bin/release publish
````

The second step creates the GitHub release in draft state, optionally marked as pre-release; the final step removes the draft state (making the release visible to the outside world) and updates latest in {Git,Docker}Hub only if the pre-release attribute is missing.

## Caveats
* The previous implementation used `git-describe` with a `--match=v*` to find the most recent version tag. To support non-mainline releases I've had to remove all the 'v' related logic with three consequences:
 * The version in the changelog must match the exact tag name. For mainline releases it must include the 'v' prefix, which is not currently the case
 * With no prefix to search for, the new implementation uses `git tag --points-at HEAD` and expects to find a single tag there which it assumes is the version tag. This is slightly less flexible that before, because it means the `HEAD` of your local branch must be the tagged commit that you want to build
 * The `latest_release` tag must point at the version tag, not directly at the commit object (e.g. you *must* now use `git tag -af latest_release v2.0.0` whereas before you could have had `v2.0.0` checked out and done `git tag -af latest_release`). The current release instructions describe the first method anyway, so probably not a big deal

## Outstanding Issues
* Something of a showstopper - the `github-release edit` command used to remove the draft status resets the description field. I can think of a several solutions to this, none of which are very appealing:
 * Read the description with `github release info` first and then supply it back as an argument to `github-release edit`. Note that the output of `info` isn't really amenable to reliable parsing, particularly in the face of the freeform multiline description field
 * Manipulate the GitHub API directly to remove the draft status. Not straightforward due to the need to manually search through the (potentially paginated) list of releases to find the release ID; if we are willing to do this, begs the question of why we're bothering with `github-release`
 * Prompt the user to publish the release manually using the Github UI. We could do this either at the end of `bin/release publish`, or we could require them to do it first and check at the start
* The current release process instructions specify (in bold) that the user updates the `latest_release` tag in the repository cloned into `releases/`, not in the repository in which they created the version tag and are running `bin/release`. This seems odd - does anyone object to the checks being modified so that they all operate on the 'outer' repository? (Related, there's a bug in the current logic anyway - see the 'Fix latest_release hash test' commit in this PR)

@dpw @rade @squaremo particularly interested in your opionions on the caveats and issues - this PR isn't ideal at the moment. Will hold off producing updated release process instructions for the wiki until things are ironed out.

Further explanatory material is available in the individual commit messages.

Fixes #880, fixes #909.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -1,0 +1,131 @@
+# Release Process
+## Prerequisites
+
+* Up-to-date versions of `git`, `go` etc
+* Install GitHub release tool `go get github.com/weaveworks/github-release`
+* Create a [github token for
+  github-release](https://help.github.com/articles/creating-an-access-token-for-command-line-use/);
+set and export `$GITHUB_TOKEN` with this value
+* Update all dependencies with `make update`
+
+## Build Phase
+### Update CHANGELOG.md
+
+* Checkout the branch from which you wish to release
+* Choose an appropriate version tag, henceforth referred to as `$TAG`.
+  Mainline releases use a version number (e.g. `TAG=v2.0.0`), whereas
+  pre-releases get a descriptive name (e.g. `TAG=feature-preview-20150902`)
+* Add a changelog entry for the new tag at the top of `CHANGELOG.md`.
+  The first line must be a markdown header of the form `## Release
+  $TAG`
+
+Commit the changelog update:
+
+    git commit -m "Add release $TAG" CHANGELOG.md
+
+### Create Version Tag
+
+Next you must tag the changelog commit with `$TAG`
+
+    git tag -a -m "Release $TAG" $TAG
+
+### Execute Build
+
+You are now ready to perform the build. If you have skipped the
+previous steps (e.g. because you're doing a rebuild), you must ensure
+that `HEAD` points to the tagged commit. You may then execute
+
+    bin/release build
+
+This has the following effects:
+
+* `git tag --points-at HEAD` is used to determine `$TAG` (hence the
+  `HEAD` requirement)
+* Your *local* repository is cloned into `releases/$TAG`
+* `CHANGELOG.md` is checked to ensure it has an entry for `$TAG`
+* Distributables injected with `$TAG` are built
+* Tests are executed
+
+## Draft Phase
+### Push Version Tag Upstream
+
+First you must push your version tag upstream, so that an associated
+GitHub release may be created:
+
+    git push git@github.com:weaveworks/weave $TAG
+
+N.B. if you're testing the release process, push to your fork
+instead!
+
+### Create Draft Release
+
+You're now ready to draft your release notes:
+
+    bin/release draft [--pre-release]
+
+This has the following effects:
+
+* A [release](https://help.github.com/articles/about-releases) is
+  created in GitHub for `$TAG`. This release is in the draft state, so
+  it is only visible to contributors
+* The `weave` script is uploaded as an attachment to the release
+* If `--pre-release` is specified, the release will have the
+  pre-release attribute set (this affects the way GitHub displays the
+  release and modifies the behaviour of the publish phase)
+
+Navigate to https://github.com/weaveworks/weave/releases, 'Edit' the
+draft and input the release notes. When you are done make sure you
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by awh** [comment]
I've looked at fixing `github-release edit` so that it only populates the `PATCH` JSON for those properties that have been specified, but the command line parser they use isn't up to it (it can't distinguish between not specifying a string vs specifying an empty string for example). It wouldn't be hard to add a `github-release publish` that just flips the bit we require though, and it does align with the GitHub UI operations so would probably be acceptable upstream as a PR.

**Comment by awh** [comment]
I have [forked `github-release`](https://github.com/weaveworks/github-release), implemented `github-release publish` and updated `bin/release` to use it.

**Comment by awh** [comment]
After discussion with @squaremo I have modified the script so that all tag checking operations happen on the *outer* repository. Since both currently outstanding issues are now addressed, I've included another commit which adds `docs/release-process.md` outlining the new release process.

**Comment by awh** [comment]
Have tested both mainline and pre-release paths; this PR is ready for review.

File: docs/release-process.md [hunk 1]
```diff
@@ -0,0 +1,128 @@
...
+There's a few things that can go wrong.
+
+ * If the build is wonky, e.g., the tests don't pass, you can delete
+   the directory in `./releases/`, fix whatever it is, move the
+   version tag (which should still be only local) and have another go.
+ * If the DockerHub pushes fail (which sadly seems to happen a lot),
+   you can just run `./bin/release publish` again.
+ * If it turns out you released the wrong thing, you can delete the
+   release on GitHub and start again; it's the only thing that has to
+   be manually deleted.
```
rade: Surely that is not correct. Images will have been published on dockerhub and tagged as 'latest', which means they will get picked up by the snaphot/master weave script. And the latest_release tag in gh will point at the just-released version, which means our published download URLs will pick that up.

docs/release-process.md
```diff
@@ -0,0 +1,128 @@
...
+draft and input the release notes. When you are done make sure you
+'Save draft' (and not 'Publish release'!).
+
+Once the release notes have passed review, proceed to the publish
+phase.
+
+## Publish Phase
+### Move/Force Push `latest_release` Tag
+
+This step is only required for mainline (non pre-release) releases:
```
rade: s/is only required/must only be performed

docs/release-process.md
awh: I copied that section from the existing wiki page. It's instructions on how to overwrite a bad release with a fresh one of the same name, not how to 'undo'...

rade: I know it's a copy. Needs to be made clear that these aren't undo instructions.

awh:
Update documentation.

rade:
LGTM

I suggest we
- get rid of https://github.com/weaveworks/weave/wiki/Release-Process
- add a link to the new docs to https://github.com/weaveworks/weave/wiki and https://github.com/weaveworks/weave/wiki/TestingReleaseProcess (or perhaps incorporate the latter as a separate section in the new docs)

Issue #1371, opened 2015-08-28
We shouldn't be modifying unintended fields when creating containers/execs


paulbellamy:
The side effect of decoding/encoding the request body with go-dockerclient, is that it can encode it differently than the client had meant to. This caused a failure in https://github.com/weaveworks/weave/pull/1367, and could generally cause mangling of requests (e.g. if the fields present in go-dockerclient disagree with the docker server).

We should use a less invasive way of modifying the incoming requests.

Specific instance of this bug is the failure in [test_valid_no_config_specified ](https://github.com/docker/docker-py/blame/db1a93fd27a7627300c1fd54040d2f727ada78a4/tests/integration_test.py#L305)

rade:
I am reluctant to make such a significant change in 1.1 so close to a release, though I see you are already half-doing that in #1376.

Would updating the go docker client to the latest version - which iirc happens anyway during the release process - ameliorate this issue?

paulbellamy:
I agree it's a bit big for 1.1.

Updating the go docker client may temporarily help, but won't solve other issues (like with volume drivers). The client makes a few simplifying assumptions about the format of the data which are approriate for a client, but not specific enough for the server.

rade:
> won't solve other issues (like with volume drivers)

Oh. Really? The most recent go docker client ignores the volume driver settings?

paulbellamy:
Ah, no, looking again, it does handle them. It's more that longer-term when the API changes the go client (being stricter) will break more often.

rade:
could somebody please check that #1378 gets fixed when running with a proxy built with the latest golang docker client?

bboreham:
On it.

bboreham:
It does not get fixed; go-dockerclient does not yet know anything about the new `VolumeDriver` parameter.

bboreham:
So I submitted https://github.com/fsouza/go-dockerclient/pull/367

peter-edge:
@bboreham let me know if there's anything else missing on the volumes side of go-dockerclient, I think we might be both trying to accomplish similar things https://github.com/fsouza/go-dockerclient/pull/362

bboreham:
@peter-edge ok, but we're reacting to a user issue #1378; I eyeballed the Docker change to look for other missing fields but don't have any good basis to know for sure.

errordeveloper:
[...]

PR #1376, opened 2015-09-01
[proxy] add --rewrite-inspect flag

### LINKED ISSUE #117: Expose weave network to docker inspect
I believe that the zettio-weave project needs to sort out how to use its run, attach and detach subcommands to inject and maintain  information about the weave bridge into the docker inspect <container> output.

https://github.com/hesco/hesco-weave#to-do-tasks-for-other-projects

paulbellamy:
To return the weave ip/mac for docker inspect calls.

Fixes #117, and #1199 (possibly? Anything else outstanding for that?)

### DIFF
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -29,6 +29,7 @@ func main() {
 	mflagext.ListVar(&c.ListenAddrs, []string{"H"}, nil, "addresses on which to listen")
 	mflag.StringVar(&c.HostnameMatch, []string{"-hostname-match"}, "(.*)", "Regexp pattern to apply on container names (e.g. '^aws-[0-9]+-(.*)$')")
 	mflag.StringVar(&c.HostnameReplacement, []string{"-hostname-replacement"}, "$1", "Expression to generate hostnames based on matches from --hostname-match (e.g. 'my-app-$1')")
+	mflag.BoolVar(&c.RewriteInspect, []string{"-rewrite-inspect"}, false, "Rewrite 'inspect' calls to return the weave network settings (if attached)")
 	mflag.BoolVar(&c.NoDefaultIPAM, []string{"#-no-default-ipam", "-no-default-ipalloc"}, false, "do not automatically allocate addresses for containers without a WEAVE_CIDR")
 	mflag.BoolVar(&c.NoRewriteHosts, []string{"-no-rewrite-hosts"}, false, "do not automatically rewrite /etc/hosts. Use if you need the docker IP to remain in /etc/hosts")
 	mflag.StringVar(&c.TLSConfig.CACert, []string{"#tlscacert", "-tlscacert"}, "", "Trust certs signed only by this CA")
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -3,12 +3,15 @@ package proxy
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
 	"regexp"
+	"strings"
 
 	"github.com/fsouza/go-dockerclient"
 	. "github.com/weaveworks/weave/common"
@@ -61,4 +64,33 @@ func inspectContainerInPath(client *docker.Client, path string) (*docker.Contain
 		Log.Warningf("Error inspecting container %s: %v", containerID, err)
 	}
 	return container, err
+}
+
+func weaveContainerIPs(container *docker.Container) (mac string, ips []net.IP, nets []*net.IPNet, err error) {
+	stdout, stderr, err := callWeave("ps", container.ID)
+	if err != nil || len(stderr) > 0 {
+		err = errors.New(string(stderr))
+		return
+	}
+	if len(stdout) <= 0 {
+		return
+	}
+
+	fields := strings.Fields(string(stdout))
+	if len(fields) < 2 {
+		return
+	}
+	mac = fields[1]
+
+	var ip net.IP
+	var ipnet *net.IPNet
+	for _, cidr := range fields[2:] {
+		ip, ipnet, err = net.ParseCIDR(cidr)
+		if err != nil {
+			return
+		}
+		ips = append(ips, ip)
+		nets = append(nets, ipnet)
+	}
+	return
 }
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -105,8 +105,8 @@ func inspectContainerInPath(client *docker.Client, path string) (*docker.Contain
 	return container, err
 }
 
-func weaveContainerIPs(container *docker.Container) (mac string, ips []net.IP, nets []*net.IPNet, err error) {
-	stdout, stderr, err := callWeave("ps", container.ID)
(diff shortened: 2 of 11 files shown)


errordeveloper:
I will test this with Kubernetes once again now.

errordeveloper:
I confirm this is working perfectly well with Kubernetes.

site/proxy.md
```diff
@@ -104,6 +104,12 @@ Multiple IP addresses and networks can be supplied in the `WEAVE_CIDR`
 variable by space-separating them, as in
 `WEAVE_CIDR="10.2.1.1/24 10.2.2.1/24"`.
 
+The docker IP will still be returned by `docker inspect`. If you want
```
bboreham: This is assuming regular `bridge` networking. Do we need to qualify what happens when you run a container with `--net=none` or `host` ?

proxy/inspect_exec_interceptor.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil
+	}
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+	r.Body.Close()
+
+	exec := &docker.ExecInspect{}
```
bboreham: Shouldn't we do this in the maps & strings style, rather than a fixed struct, since we already know that ages poorly?

proxy/inspect_exec_interceptor.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	newBody, err := json.Marshal(exec)
+	if err != nil {
+		return err
+	}
+	r.Body = ioutil.NopCloser(bytes.NewReader(newBody))
+	r.ContentLength = int64(len(newBody))
+	r.TransferEncoding = nil // Stop it being chunked, because that hangs
+
```
bboreham: Maybe the above 7 lines, which are common to `inspectContainerInterceptor`, could be factored out into a function that sets up the reply?  Similar to `marshalRequestBody`, but for `Response`.

prog/weaveproxy/main.go
```diff
@@ -29,6 +29,7 @@ func main() {
 	mflagext.ListVar(&c.ListenAddrs, []string{"H"}, nil, "addresses on which to listen")
 	mflag.StringVar(&c.HostnameMatch, []string{"-hostname-match"}, "(.*)", "Regexp pattern to apply on container names (e.g. '^aws-[0-9]+-(.*)$')")
 	mflag.StringVar(&c.HostnameReplacement, []string{"-hostname-replacement"}, "$1", "Expression to generate hostnames based on matches from --hostname-match (e.g. 'my-app-$1')")
+	mflag.BoolVar(&c.RewriteInspect, []string{"-rewrite-inspect"}, false, "Rewrite inspect calls to return the weave IP (if attached)")
```
bboreham: If we come up with some other reason to rewrite the inspect info, will we want to rename this flag?  Should it be more tightly targeted now?

site/proxy.md
```diff
@@ -104,6 +104,12 @@ Multiple IP addresses and networks can be supplied in the `WEAVE_CIDR`
...
+`docker inspect` to return the weave IP instead, then the proxy must
```
bboreham: We also rewrite the MAC address and IP network prefix length.

proxy/inspect_exec_interceptor.go
paulbellamy: Was planning to do change all the interceptors over to maps&strings in one go, but no reason not to do it first here, I suppose.

site/proxy.md
paulbellamy: It does rewrite those, but will add to the docs, thanks.

paulbellamy:
Note, docker `NetworkSettings` includes:
```json
"NetworkSettings": {
    "Bridge": "",
    "Gateway": "",
    "IPAddress": "",
    "IPPrefixLen": 0,
    "MacAddress": "",
    "PortMapping": null,
    "Ports": null
},
```
This only overwrites the `IPAddress`, `IPPrefixLen`, and `MacAddress`, but not `Gateway`.

bboreham:
Gateway seems to point to the docker bridge IP.  I don't think we should leave that in place.

prog/weaveproxy/main.go
rade: Also, the description is too specific, since we substitute more than just the IP.

Also, the flag should be mentioned in the weave usage.

bboreham:
Changes look good; we are still leaving `gateway` in the `NetworkSettings` - do we want that?

paulbellamy:
Could replace `gateway` with the IP of the weave router... But I'm not sure how correct that is.

bboreham:
This gateway refers to the route that the container has to the rest of the world, so given that we don't add such a route for ethwe I don't think we should claim a fake one.

I'm not so sure now that leaving it is wrong.  That route and gateway still exists.  It just isn't associated with the IP or MAC that we are now returning.

paulbellamy:
If we are returning an IP/MAC for an interface with no gateway, it probably makes most sense not to return a gateway.

site/proxy.md
```diff
@@ -104,6 +104,15 @@ Multiple IP addresses and networks can be supplied in the `WEAVE_CIDR`
 variable by space-separating them, as in
 `WEAVE_CIDR="10.2.1.1/24 10.2.2.1/24"`.
 
+The docker NetworkSettings (including IP address, MacAddress, and
+IPPrefixLen), will still be returned by `docker inspect`. If you want
+`docker inspect` to return the weave NetworkSettings instead, then the
+proxy must be launced with the `--rewrite-inspect` flag. This will
+only substitute in the weave network settings when the container has a
+weave IP.
```
bboreham: Should we document what happens if the container has more than one weave IP address?

(it looks like we return the first one output by `ip addr show dev ethwe`)

proxy/common.go
```diff
@@ -19,6 +22,14 @@ var (
 	weaveWaitEntrypoint = []string{"/w/w"}
 )
 
+type UnmarshalWrongTypeError struct {
+	Field, Expected string
+}
+
+func (e *UnmarshalWrongTypeError) Error() string {
+	return fmt.Sprintf("Wrong type for %s field, expected %s", e.Field, e.Expected)
```
bboreham: This message will be confusing if the real issue is that the field is missing completely from the JSON.

rade:
Is d7d8b86189dc31c772a04ca2a89efb8a24404d3f also fixing #1371?

bboreham:
No, it's avoiding making #1371 worse for this specific data.  Some of the code will be re-usable when we fix #1371 fully.

rade:
Hmm. It feels odd to do things completely differently here.

bboreham:
It seems much more likely that we will spoil someone's day by disrupting what is returned by `docker inspect` when Docker change it, than the return values from `create`, `start`, `restart`, etc.  Borne out by the fact that only one person managed to trip over this in the real world so far.

rade:
@paulbellamy anything left to do here?

paulbellamy:
I believe all of the comments have been addressed, so if the docs look good, then it should be ready to go.

bboreham:
LGTM

Issue #1381, opened 2015-09-03
Obtain container hostname from specified Docker container label


2opremio:
TL;DR Add a new flag to weave proxy `--hostname-from-label` which specifies a label key from which to extract the container's hostname.

We already added `--hostname-match` and `--hostname-replacement` to the proxy, allowing to derive hostnames from transformations of container names (see https://github.com/weaveworks/weave/pull/1126).

For ECS this is not enough since the original specified container names are [mangled destructively](https://github.com/weaveworks/guides/blob/5a70632985c5f1c64dcb807e35cb33cab6a7745e/aws-ecs/README.md#known-issues--limitations), not allowing to retrieve the original name at spawning time.

However, I just found out that ECS adds a label, `com.amazonaws.ecs.container-name`, to every container with the original container name.

We can add a new flag to weave proxy, `--hostname-from-label` to specifiy a label key from which to extract the container's hostname.

rade:
> We can add a new flag to weave proxy, `--hostname-from-label` to specifiy a label key from which to extract the container's hostname.

And this could work in *conjunction* with the existing options, right?

2opremio:
Yes, I don't see a reason why it wouldn't.

rade:
ok. hurry up, let's try to get this in 1.1 :)

2opremio:
I'm on it!

faddat:
Can you give me an example of how I can use this?

faddat:
Nevermind.  I did not know what docker labels could take key+value.  For the next ignorant sucker:

From the docker docs:

label (label=<key> or label=<key>=<value>)

Issue #1212, opened 2015-07-27
Upgrade gce hosts to ubuntu 16


paulbellamy:

bboreham:
Closed by #2694 (possibly earlier, it was difficult to see)

Issue #1161, opened 2015-07-16
weave attach non zero exit code on success


anandkumarpatel:
if a command run with `docker --rm` fails to remove the container after a successful run docker will return a non-zero exit.
when using weave attach I frequently see 
```
Error removing intermediate container 9b63e9a1e9bf: Driver aufs failed to remove root filesystem 9b63e9a1e9bf1aaa97d943b4c082eaa152e38d55bf1884bc28d88de1cc891f1a: rename /docker/aufs/mnt/9b63e9a1e9bf1aaa97d943b4c082eaa152e38d55bf1884bc28d88de1cc891f1a /docker/aufs/mnt/9b63e9a1e9bf1aaa97d943b4c082eaa152e38d55bf1884bc28d88de1cc891f1a-removing: device or resource busy
```
and the command return non-zero so our tools think the command failed, however the attach actually did go through.
the above error was coming from this line due to the `--rm` bug
https://github.com/weaveworks/weave/blob/master/weave#L74

docker issue https://github.com/docker/docker/issues/9665#issuecomment-121759275

info
```
root@zapdos:/home/zapdos#  sudo weave version
weave script 0.11.1
weave router 0.11.1
weave DNS 0.11.1
weave exec 0.11.1
```
```
root@zapdos:/home/zapdos# docker -D version && docker -D info
Client version: 1.6.0
Client API version: 1.18
Go version (client): go1.4.2
Git commit (client): 4749651
OS/Arch (client): linux/amd64
Server version: 1.6.0
Server API version: 1.18
Go version (server): go1.4.2
Git commit (server): 4749651
OS/Arch (server): linux/amd64
Containers: 650
Images: 8317
Storage Driver: aufs
 Root Dir: /docker/aufs
 Backing Filesystem: extfs
 Dirs: 9626
 Dirperm1 Supported: false
Execution Driver: native-0.2
Kernel Version: 3.13.0-29-generic
Operating System: Ubuntu 14.04 LTS
CPUs: 32
Total Memory: 58.81 GiB
Name: zapdos 
ID: HHHH:IIII:MRNF:OYNJ:FZNT:X4EA:ZWQR:TF7G:EEEE:HSP6:RYBS:F3ER
Debug mode (server): false
Debug mode (client): true
Fds: 1111
Goroutines: 3753
System Time: Wed Jul 15 18:54:22 UTC 2015
EventsListeners: 1
Init SHA1: 11117501111dbf64c1111c27861111027c
[...]

PR #1384, opened 2015-09-03
[proxy] Add --hostname-from-label

2opremio:
This allows registering container hostnames based on a label provided when
creating the container.

Implements #1381

### DIFF
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -164,29 +164,43 @@ for registration. However, there are situations in which the final container
 name is out of the user's control (e.g. when using Docker orchestrators which
 append control/namespacing identifiers to the original container names).
 
-For those situations, the proxy provides two flags: `--hostname-match <regexp>`
-and `--hostname-replacement <replacement>`. When launching a container, its name
-is matched against regular expression `<regexp>`. Then, based on that match,
-`<replacement>` will be used to generate a hostname, which will ultimately be
-handed over to weaveDNS for registration.
+For those situations, the proxy provides a few flags: `--hostname-from-label
+<labelkey>`, `--hostname-match <regexp>` and `--hostname-replacement
+<replacement>`. When launching a container, the hostname is initialized to the
+value of the container label with key `<labelkey>`, if `<labelkey>` wasn't
+provided, the container name is used. Additionally, the hostname is matched
+against regular expression `<regexp>`. Then, based on that match,
+`<replacement>` will be used to obtainer the final hostname, which will
+ultimately be handed over to weaveDNS for registration.
 
-For instance, if we launch the proxy using
-`--hostname-match '^aws-[0-9]+-(.*)$'` and `--hostname-replacement 'my-app-$1'`
+For instance, we can launch the proxy using all three flags
 
-    host1$ weave launch-router && weave launch-proxy --hostname-match '^aws-[0-9]+-(.*)$' --hostname-replacement 'my-app-$1'
+    host1$ weave launch-router && weave launch-proxy --hostname-from-label hostname-label --hostname-match '^aws-[0-9]+-(.*)$' --hostname-replacement 'my-app-$1'
     host1$ eval "$(weave env)"
 
-then, running a container named `aws-12798186823-foo` will lead to weaveDNS registering
-hostname `my-app-foo` and not `aws-12798186823-foo`.
+Note how regexp substitution groups should be prepended with a dollar sign
+(e.g. `$1`). For further details on the regular expression syntax please see
+[Google's re2 documentation](https://github.com/google/re2/wiki/Syntax).
+
+
+Then, running a container named `aws-12798186823-foo` without labels will lead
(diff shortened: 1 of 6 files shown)


test/680_proxy_hostname_derivation_test.sh
```diff
@@ -6,23 +6,58 @@ C1=10.2.0.78
...
+weave_on $HOST1 launch-router
+weave_on $HOST1 launch-proxy --hostname-from-label hostname-label --hostname-match '^[^-]+-(?P<appname>[^-]*)-[^-]+$' --hostname-replacement '$appname'
+
+CID1=$(proxy_start_container_with_dns $HOST1 -e WEAVE_CIDR=$C1/24 --label=hostname-label=$CNAME1)
+CID2=$(proxy_start_container_with_dns $HOST1 -e WEAVE_CIDR=$C2/24 --label=hostname-label=$CNAME2)
+CID3=$(proxy_start_container_with_dns $HOST1 -e WEAVE_CIDR=$C3/24 --label=hostname-label=$CNAME3)
+
+assert_dns_a_record $HOST1 $CID1 $FQDN1 $C1
+assert_dns_a_record $HOST1 $CID2 $FQDN3 $C3
+assert_dns_a_record $HOST1 $CID3 $FQDN2 $C2
```
rade: - please introduce some helper(s) to avoid all the repetition
- this should be a single test suite

weave
```diff
@@ -56,6 +56,7 @@ weave launch-proxy  [-H <endpoint>] [--with-dns | --without-dns]
                       [--no-default-ipalloc] [--no-rewrite-hosts]
                       [--hostname-match <regexp>]
                       [--hostname-replacement <replacement>]
+                      [--hostname-from-label <labelkey>]
```
rade: I'd put that first, since it logically precedes the others.

site/proxy.md
```diff
@@ -178,6 +181,18 @@ Note how regexp substitution groups should be prepended with a dollar sign
...
+    host1$ weave launch-router && weave launch-proxy --hostname-from-label hostname-label
+    host1$ eval "$(weave env)"
+
+the hostname will be extracted from the label with key `hostname-label`
+
+    host1$ docker run -ti --label=hostname-label=my-app-foo ubuntu ping my-app-foo
+    PING my-app-foo.weave.local (10.32.0.2) 56(84) bytes of data.
+    64 bytes from my-app-foo.weave.local (10.32.0.2): icmp_seq=1 ttl=64 time=0.031 ms
+    64 bytes from my-app-foo.weave.local (10.32.0.2): icmp_seq=2 ttl=64 time=0.042 ms
+
```
rade: Too elaborate; I suggest we have just one example, showing the use of all three options.

2opremio: I would like to show both how it works with labels and how the container name is used if no label is provided. I cannot do that without spawning two containers.

What I can definitely do is only start the weave proxy once. I have pushed a change with that, let me know if that is still too elaborate.

rade: > I would like to show both how it works with labels and how the container name is used if no label is provided.

And my point was that I don't think that is necessary. Just show all three being used together.

2opremio:
@rade Please let me squash the commits before merging

rade:
LGTM. squash away.

rade:
> LGTM. squash away.

Assuming you fix the CI failures, of course :)

2opremio:
It seems the tests fail because I am hitting a Docker in CircleCI, most probably https://github.com/docker/docker/issues/9665

Two tests fail when destroying the `weaveproxy` container with

    Cannot destroy container [container]: Driver aufs failed to remove root filesystem [filesystemUUID]: device or resource busy


*It Works On My Local Machine™* and rerunning the tests doesn't seem to help. I suspect it's related to starting and stopping the `weaveproxy` multiple times in my tests (which I had to do to satisfy the request from @rade to only have a suite per testfile), mixed with running the tests in parallel. See https://github.com/weaveworks/weave/blob/dns-hostname-from-labels/test/680_proxy_hostname_derivation_test.sh#L41

@tomwilkie Have you run into this one before? Maybe @bboreham ?

rade:
It shouldn't be using aufs. That is bad, bad, bad. We fixed that in #1125 for the VMs, but I guess not for CircleCI. Perhaps that is what the incredibly descriptive #1212 is about?

rade:
See also #1161 and #408.

bboreham:
Correction: #1125 was to fix using _vfs_; aufs is what I was using in the past, and I didn't notice it to be too bad.

But I think it's true that I've seen fewer of those "failed to remove" messages since moving to the 'overlay' driver.  Rebooting the VM was the best solution when I did get those messages.

2opremio:
Including the tests in this PR make it fail systematically, whereas with overlay *It Works On My Local Machine™* every time.

2opremio:
For the record, the problem was circumvented by only starting the router in the test suite, instead of doing it for every test.

PR #1391, opened 2015-09-04
Stop marshalling numbers as floats

### LINKED ISSUE #1390: with `--rewrite-inspect` proxy modifies how integers are marshalled
From testing latest change (32a82f2d0e83) on Kubernetes, one rather [critical issue popped-up](https://gist.github.com/errordeveloper/a13b33a342bed5a9652e#file-inspect-diff). Essentially what happens is that we somehow end-up marshalling integers in scientific notation and the parser on the other end fail with `json: cannot unmarshal number 5.24288e+07 into Go value of type int64`.

bboreham:
Fixes #1390

Note that setting both `--memory` and `--memory-swap` is done to avoid Docker issuing a warning that it can't set the swap limit.

### DIFF
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -50,7 +50,9 @@ func unmarshalRequestBody(r *http.Request, target interface{}) error {
 	}
 	r.Body = ioutil.NopCloser(bytes.NewReader(body))
 
-	return json.Unmarshal(body, &target)
+	d := json.NewDecoder(bytes.NewReader(body))
+	d.UseNumber() // don't want large numbers in scientific format
+	return d.Decode(&target)
 }
 
 func marshalRequestBody(r *http.Request, body interface{}) error {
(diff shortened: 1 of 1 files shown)


rade:
I don't think the test is worth the hassle. remove.

bboreham:
Are you referring to the last `grep`?  Since it works on my machine and fails on CircleCI, I can easily agree with that sentiment.

However, just setting the memory limit (so there is a large number in the output) breaks `docker inspect  -f`, so I think that is worth keeping.

rade:
I am saying we don't need a test for this issue at all. Heresy, I know.

bboreham:
You are saying to remove the whole file `605_proxy_inspect_test.sh` ?

I guess I can put the memory limit into test 610.  I like the idea of having a test that would have caught this error.

rade:
no, I am saying don't add any extra test for this issue. i.e. `605_proxy_inspect_test.sh` should remain unchanged.

bboreham:
Done.

PR #1395, opened 2015-09-08
[proxy] all request modification is done through map[string]interface{}

### LINKED ISSUE #1371: We shouldn't be modifying unintended fields when creating containers/execs

paulbellamy:
That way we are less likely fall victim to go-dockerclient, or the API changing.

Fixes #1371

### DIFF
--- a/proxy/create_container_interceptor.go
+++ b/proxy/create_container_interceptor.go
@@ -17,12 +17,6 @@ var (
 
 type createContainerInterceptor struct{ proxy *Proxy }
 
-type createContainerRequestBody struct {
-	*docker.Config
-	HostConfig *docker.HostConfig `json:"HostConfig,omitempty" yaml:"HostConfig,omitempty"`
-	MacAddress string             `json:"MacAddress,omitempty" yaml:"MacAddress,omitempty"`
-}
-
 // ErrNoSuchImage replaces docker.NoSuchImage, which does not contain the image
 // name, which in turn breaks docker clients post 1.7.0 since they expect the
 // image name to be present in errors.
@@ -35,35 +29,45 @@ func (err *ErrNoSuchImage) Error() string {
 }
 
 func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
-	container := createContainerRequestBody{}
+	container := jsonObject{}
 	if err := unmarshalRequestBody(r, &container); err != nil {
 		return err
 	}
 
-	if cidrs, err := i.proxy.weaveCIDRsFromConfig(container.Config, container.HostConfig); err != nil {
+	hostConfig, err := container.Object("HostConfig")
+	if err != nil {
+		return err
+	}
+
+	networkMode, err := hostConfig.String("NetworkMode")
+	if err != nil {
+		return err
+	}
+
+	env, err := container.StringArray("Env")
+	if err != nil {
+		return err
+	}
+
+	if cidrs, err := i.proxy.weaveCIDRs(networkMode, env); err != nil {
 		Log.Infof("Leaving container alone because %s", err)
 	} else {
 		Log.Infof("Creating container with WEAVE_CIDR \"%s\"", strings.Join(cidrs, " "))
-		if container.HostConfig == nil {
-			container.HostConfig = &docker.HostConfig{}
-		}
-		if container.Config == nil {
-			container.Config = &docker.Config{}
-		}
-		i.proxy.addWeaveWaitVolume(container.HostConfig)
-		if err := i.setWeaveWaitEntrypoint(container.Config); err != nil {
+		if err := i.proxy.addWeaveWaitVolume(hostConfig); err != nil {
 			return err
 		}
-		hostname := r.URL.Query().Get("name")
-		if i.proxy.Config.HostnameFromLabel != "" {
-			if labelValue, ok := container.Labels[i.proxy.Config.HostnameFromLabel]; ok {
-				hostname = labelValue
-			}
+		if err := i.setWeaveWaitEntrypoint(container); err != nil {
+			return err
 		}
-		hostname = i.proxy.hostnameMatchRegexp.ReplaceAllString(hostname, i.proxy.HostnameReplacement)
+		hostname, err := i.containerHostname(r, container)
+		if err != nil {
+			return err
+		}
 		if dnsDomain := i.proxy.getDNSDomain(); dnsDomain != "" {
-			i.setHostname(&container, hostname, dnsDomain)
-			if err := i.proxy.setWeaveDNS(container.HostConfig, container.Hostname, dnsDomain); err != nil {
+			if err := i.setHostname(container, hostname, dnsDomain); err != nil {
+				return err
+			}
+			if err := i.proxy.setWeaveDNS(hostConfig, hostname, dnsDomain); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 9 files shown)


proxy/create_container_interceptor.go
```diff
@@ -76,48 +83,114 @@ func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
...
-func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container *docker.Config) error {
-	if len(container.Entrypoint) == 0 {
-		image, err := i.proxy.client.InspectImage(container.Image)
+func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container map[string]interface{}) error {
+	var entrypoint []string
+	if e, ok := container["Entrypoint"]; ok && e != nil {
+		switch e := e.(type) {
+		case string:
+			entrypoint = []string{e}
+		case []string:
```
bboreham: It looks like the rest of this block duplicates lookupStringArray(); could just call it?

bboreham:
Did you consider using a library like https://github.com/Jeffail/gabs ?

paulbellamy:
In classic go fashion, I did not. That would probably help clean this up, though. Will give it a go, thanks.

proxy/create_container_interceptor.go
```diff
@@ -76,48 +83,114 @@ func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
...
+	return
+}
+
+func (i *createContainerInterceptor) hostnameFromLabel(hostname string, container map[string]interface{}) (string, error) {
+	labels, err := lookupObject(container, "Labels")
+	if err != nil {
+		return "", err
+	}
+
+	labelIface, ok := labels[i.proxy.Config.HostnameFromLabel]
```
bboreham: From here to the end of the function is pretty much lookupString() - could you just call that?

proxy/create_container_interceptor.go
paulbellamy: With a refactor of lookupStringArray to preserve nil vs empty arrays, probably..

paulbellamy:
gabs creates intermediary nodes when looking up, which is good, but returns `interface{}` everywhere, so you still have to do all the typecasting.
jsonq, doesn't create intermediary nodes when looking up, so you have to special-case that error everywhere.

Neither library really shortens the code, but I did clean up the other spots mentioned.

bboreham:
LGTM; please could you rebase against current master.

rade:
Perhaps we should have a type alias for `map[string]interface{}` and turn the lookup functions into methods on that?

proxy/proxy.go
```diff
@@ -365,15 +365,11 @@ func (proxy *Proxy) attach(container *docker.Container) error {
 	return nil
 }
 
-func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *docker.HostConfig) ([]string, error) {
-	netMode := ""
-	if hostConfig != nil {
-		netMode = hostConfig.NetworkMode
+func (proxy *Proxy) weaveCIDRsFromConfig(networkMode string, env []string) ([]string, error) {
```
rade: isn't this function somewhat misnamed now?

proxy/json.go
```diff
@@ -13,12 +13,13 @@ func (e *UnmarshalWrongTypeError) Error() string {
 	return fmt.Sprintf("Wrong type for %s field, expected %s, but got %T", e.Field, e.Expected, e.Got)
 }
 
-// For parsing json objects. Look up (or create) a child object in the given map.
-func lookupObject(m map[string]interface{}, key string) (map[string]interface{}, error) {
-	iface, ok := m[key]
+type jsonObject map[string]interface{}
+
+func (j jsonObject) Object(key string) (jsonObject, error) {
```
rade: This method appears to have lost its comment. Either put it back or get rid of the comments on the other methods.

proxy/json.go
```diff
@@ -27,12 +28,20 @@ func lookupObject(m map[string]interface{}, key string) (map[string]interface{},
...
 	}
 
-	return result, nil
+	return jsonObject(result), nil
 }
 
 // For parsing json objects. Look up (or create) a string in the given map.
-func lookupString(m map[string]interface{}, key string) (string, error) {
-	iface, ok := m[key]
+func (j jsonObject) String(key string, ks ...string) (string, error) {
```
rade: The method comment no longer reflects the behaviour.

rade: It is rather odd/inconsistent that this method operates on a path whereas `Object` and `StringArray` don't.

rade: Looks like there is just one call site that uses a path; so I'd be tempted to remove this functionality and spell it out at that call site instead.

proxy/create_container_interceptor.go
```diff
@@ -74,50 +78,104 @@ func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
...
+func (i *createContainerInterceptor) setWeaveWaitEntrypoint(container jsonObject) error {
+	var entrypoint []string
+	entrypoint, err := container.StringArray("Entrypoint")
+	if err != nil {
+		err, ok := err.(*UnmarshalWrongTypeError)
+		if !ok {
+			return err
+		}
+
+		got, ok := err.Got.(string)
```
rade: So this covers the case of entrypoint being just a string rather than an array? Can that definitely happen? And if so, can it definitely not happen for, say, "Cmd"?

Would there be any harm in dealing with this in `StringArray`?

proxy/json.go
paulbellamy: I'd rather make them all take a path, as it's a cleaner usage interface, but we only use the path on this one, so didn't want to add unused behaviour.

proxy/create_container_interceptor.go
```diff
@@ -74,50 +78,104 @@ func (i *createContainerInterceptor) InterceptRequest(r *http.Request) error {
...
+	}
+	if hostname == "" && name != "" {
 		// Strip trailing period because it's unusual to see it used on the end of a host name
 		trimmedDNSDomain := strings.TrimSuffix(dnsDomain, ".")
 		if len(name)+1+len(trimmedDNSDomain) > MaxDockerHostname {
 			Log.Warningf("Container name [%s] too long to be used as hostname", name)
 		} else {
-			container.Hostname = name
-			container.Domainname = trimmedDNSDomain
+			hostname = name
```
rade: this line seems superfluous

proxy/create_container_interceptor.go
paulbellamy: We could handle it in `StringArray`. It would make it looser in what we accept for `Dns`, `Binds`, and `DnsSearch`, as well. Not sure whether it can happen for "Cmd" (It's never shown in the docs). Will check that.

rade: https://en.wikipedia.org/wiki/Robustness_principle

paulbellamy: Extremely ill-defined input formats lead to integration and backwards-compatibility nightmares. Case in point, the code we're discussing now.

Cmd can also be a string or array, so I've handled that in `StringArray`

proxy/json.go
paulbellamy: Removed on others, as they were somewhat obvious.

proxy/json.go
```diff
@@ -27,12 +28,19 @@ func lookupObject(m map[string]interface{}, key string) (map[string]interface{},
...
+	if len(ks) > 0 {
+		j, err := j.Object(key)
+		if err != nil {
+			return "", err
+		}
+		return j.String(ks[0], ks[1:]...)
+	}
+
+	iface, ok := j[key]
```
rade: As noted before, it is odd that this method performs a traversal, whereas `Object` and `StringArray` don't.  Seems to me that it would be trivial to traverse on all of them - the traversal code could live in `Object` alone, or some helper fun.

rade:
@paulbellamy please review my last commit.

rade:
@paulbellamy rest LGTM, so feel free to merge once you have reviewed/squashed my commit.

paulbellamy:
I had specifically encoded the arguments to the json traversal methods as a list of len >= 1, so that `len(path)-1` could not == -1 and panic.

rade:
meh.

paulbellamy:
Either way, in hindsight the path traversal adds more code than it saves, so I've removed it.

Issue #1410, opened 2015-09-10
proxy launch fails with `use of closed network connection` error


awh:
Launching the weave proxy fails with `Error response from daemon: use of closed network connection`

Broken by [this commit](https://github.com/fsouza/go-dockerclient/commit/edc2bd38d8fa79581371ea78b3c09935877a1ade)

rade:
Presumably this is a bug on go-dockerclient, in which case we should file a bug there. On the off chance that it isn't, we need to get a fix into 1.1.1.

awh:
You can configure the proxy to use `tcp://` instead of `unix://` as a workaround whilst we wait for the issue to be addressed upstream...

awh:
Fixed upstream by https://github.com/fsouza/go-dockerclient/commit/c9ad0ce23f68428421adfc6ced9e6123f54788a5.

PR #1411, opened 2015-09-10
always arp_update new addresses in 'attach'

### LINKED ISSUE #1405: arping not issued for dynamic `attach`, delaying container visibility
Our `attach` code looks like this (in pseudo-code):
```
attach() {
    if the ethwe interface already exists then
       add IPs to ethwe
       return
    create ethwe
    add IPs to ethwe
    arping IPs
}
```

So we don't `arping` if the ethwe already exists.

rade:
Previously we'd only do that when the container had no ethwe at the start. Therefore adding a recycled IP address to an already attached container would not send out an arp update for that IP, which would break connectivity to that IP for containers which had previously accessed that IP.

Fixes #1405.

### DIFF
--- a/test/170_ip_reuse_test.sh
+++ b/test/170_ip_reuse_test.sh
@@ -1,0 +1,31 @@
+#! /bin/bash
+
+. ./config.sh
+
+C1=10.2.1.43
+C2=10.2.1.44
+C3=10.2.1.45
+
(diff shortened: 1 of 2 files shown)


rade:
The CircleCI failure is due to #1410; the test I've added passed, and everything passes for me locally.

PR #1414, opened 2015-09-11
Proxy TLS autoconfiguration

### LINKED ISSUE #1285: simplify and document instructions for weave on a Mac
@squaremo pointed out that boot2docker uses tcp instead of unix sockets to run a proxy. Therefor, the user must set their DOCKER_HOST manually and also needs to run 'weave launch-proxy' after launching weave.

paulbellamy:
Fixes #1285

If there is no `-H` or `--tls*` set by the user, we get docker's TLS args via `/proc/<docker-pid>/cmdline`, and use them.

I'm not really sure of a better way to test, as it requires rebooting the remote docker daemon. So if the test exits abnormally there might be no docker daemon running on that host (or one with TLS), and interacts badly with different tests hosts methods of booting docker. (GCE vs vagrant)

@rade feedback, please?

### DIFF
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -45,8 +45,9 @@ argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:9999
 
-When launching the proxy via TLS, `-H` and/or [TLS options](#tls) are
-required.
+If no TLS or listening interfaces are set, TLS will be autoconfigured
+based on the docker daemon's settings, and listening interfaces will
+be autoconfigured based on your docker client's settings.
 
 Multiple `-H` arguments can be specified. If you are working with a
 remote docker daemon, then any firewalls inbetween need to be
@@ -204,13 +205,20 @@ to the proxy, the specified label has precedence over the container's name.
 
 ## <a name="tls"></a>Securing the docker communication with TLS
 
-If you are
-[connecting to the docker daemon with TLS](https://docs.docker.com/articles/https/),
-you will probably want to do the same when connecting to the
-proxy. That is accomplished by launching the proxy with the same
-TLS-related command-line flags as supplied to the docker daemon. For
-example, if you have generated your certificates and keys into the
-docker host's `/tls` directory, we can launch the proxy with:
+If you are [connecting to the docker daemon with
+TLS](https://docs.docker.com/articles/https/), you will probably want
+to do the same when connecting to the proxy. The proxy will
+automatically detect the docker daemon's TLS configuration, and
+attempt to duplicate it. In the standard auto-detection case you will
+be able to launch a TLS-enabled proxy with:
+
+    host1$ weave launch-proxy
+
+You can also manually configure the proxy's TLS. This is accomplished
+by launching the proxy with the same TLS-related command-line flags as
+supplied to the docker daemon. For example, if you have generated your
+certificates and keys into the docker host's `/tls` directory, we can
+launch the proxy with:
 
     host1$ weave launch-proxy --tls-verify --tls-cacert=/tls/ca.pem \
              --tls-cert=/tls/server-cert.pem --tls-key=/tls/server-key.pem
--- a/weave
+++ b/weave
@@ -995,11 +995,7 @@ tls_arg() {
     PROXY_ARGS="$PROXY_ARGS $1 /home/weave/tls/$3.pem"
 }
 
-proxy_args() {
-    PROXY_VOLUMES=""
-    PROXY_ARGS=""
-    PROXY_TLS_ENABLED=""
-    PROXY_HOST=""
+proxy_parse_args() {
     while [ $# -gt 0 ]; do
         case "$1" in
           -H)
@@ -1042,10 +1038,20 @@ proxy_args() {
         esac
         shift
     done
+}
+
+proxy_args() {
+    PROXY_VOLUMES=""
+    PROXY_ARGS=""
+    PROXY_TLS_ENABLED=""
+    PROXY_HOST=""
+    proxy_parse_args "$@"
 
     if [ -z "$PROXY_HOST" -a -n "$CLIENT_TLS_ENABLED" -a -z "$PROXY_TLS_ENABLED" ] ; then
-      echo "When launching the proxy via TLS, -H and/or TLS options are required." >&2
-      exit 1
+      if ! proxy_parse_args $(docker_tls_args) ; then
+        echo "When launching the proxy via TLS, -H and/or TLS options are required." >&2
+        exit 1
+      fi
     fi
     if [ -z "$PROXY_HOST" ] ; then
       case "$DOCKER_CLIENT_HOST" in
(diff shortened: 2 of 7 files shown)


paulbellamy:
@dpw Two quick questions:
1) Do you know of a better way to get the pid of the docker process? We can't just use `pgrep`, as it only looks in `/proc/`, not in `/hostproc`.
2) Is `/proc/<pid>/comm` as safe as it seems? Processes can modify it, but I wouldn't expect docker to do that.

dpw:
> 1) Do you know of a better way to get the pid of the docker process? We can't just use pgrep, as it only looks in /proc/, not in /hostproc.

No, I don't know of a better way.

> 2) Is /proc/<pid>/comm as safe as it seems? Processes can modify it, but I wouldn't expect docker to do that.o 

It's not really safe.  Apart from the issue of modification, you can symlink the docker executable to whatever name you like.  `/proc/<pid>/exe` seems like an alternative, but I'm not sure it's preferable (no reason to expect docker to be installed as `/usr/bin/docker`.  So I don't think we can fundamentally do better if we have to dig through procfs for this info.

site/proxy.md
```diff
@@ -45,8 +45,9 @@ argument, e.g.
 
     host1$ weave launch-proxy -H tcp://127.0.0.1:9999
 
-When launching the proxy via TLS, `-H` and/or [TLS options](#tls) are
-required.
+If no TLS or listening interfaces are set, TLS will be autoconfigured
+based on the docker daemon's settings, and listening interfaces will
+be autoconfigured based on your docker client's settings.
```
rade: So you could end up with TLS being configured but listening on a domain socket, which presumably doesn't support TLS?

weave
```diff
@@ -1402,6 +1409,7 @@ case "$COMMAND" in
             shift
         done
         [ -n "$SUB_STATUS" ] || echo
+
```
rade: stray change

site/proxy.md
paulbellamy: Yeah. TLS is only used on the TCP socket, so that would be a no-op.

weave
```diff
@@ -1045,7 +1041,18 @@ proxy_args() {
...
+proxy_args() {
+    PROXY_VOLUMES=""
+    PROXY_ARGS=""
+    PROXY_TLS_ENABLED=""
+    PROXY_HOST=""
+    proxy_parse_args "$@"
 
+    if [ -z "$PROXY_HOST" -a -z "$PROXY_TLS_ENABLED" ] ; then
+      proxy_parse_args $(docker_tls_args) || true
```
rade: perhaps we should only do this when CLIENT_TLS_ENABLED?

rade:
Somebody with a Mac needs to test this.

I don't think this fixes all of #1285...

1. The boot2docker instructions aren't referenced from anywhere.
2. I thought Docker toolbox, not boot2docker was the recommended way to run docker on a mac, so our instructions should reflect that.
3. http://weave.works/guides/weave-and-docker-platform/using-weave-with-machine-and-swarm.html needs to be updated.
4. We should probably add a Mac guide to http://weave.works/guides/weave-docker-index.html, which currently only lists CentOS, CoreOS and Ubuntu.

Feel free to create separate issues for some of the above w.r.t. updating docs BUT we must make sure that this PR does indeed allow us to simplify the weave-on-mac instructions as per https://github.com/weaveworks/weave/issues/1285#issuecomment-130171411, or better.

README.md
```diff
@@ -135,6 +135,27 @@ Similarly, in the container started on `$HOST2`...
...
+## OSX
+
+For usage on OSX (e.g. with Docker Machine and/or Boot2Docker) we need
+to make sure that the VM is running and configured before launching
+weave. For example, for a VM named "weave-1" on docker-machine:
+
+    osx$ docker-machine create -d virtualbox weave-1
+    osx$ docker-machine start weave-1
+    osx$ eval "$(docker-machine env weave-1)"
+
```
rade: These instructions should a) go into the Installation section, and b) finish right here (the remainder is unnecessary/duplicate info).

And we need a separate branch for this change, since we must not update the master README with instructions that only work for the next release. Look back at how we handled this for the README changes that were part of the 1.1 release.

squaremo:
OK, I tried this out -- or rather, tried the instructions in PR #1441, which are admirably short :)

(After installing docker toolbox)

    $ docker-machine create -d virtualbox weave-1
    $ docker-machine start weave-1 # seems to be a no-op?
    $ eval $(docker-machine env weave-1)

I can now `docker ps` and get a sensible response. Then I load weave from a tar file, built on this branch:

```bash
$ docker load -i weave.tar
```

    $ ./weave launch
    $ eval $(./weave env)

Now I get

```bash
$ docker ps
An error occurred trying to connect: Get https://192.168.99.100:12375/v1.20/containers/json: tls: oversized record received with length 20527
```

Looking at the weaveproxy logs and inspect output, I see it wasn't started with any TLS arguments; looking at `WEAVE_DEBUG=1 ./weave launch`, it seems the invocation of `docker_tls_args` doesn't return anything. Whaaaa?

OK, let's have a look at what that program is actually doing, or at least what it's looking at.

```bash
$ docker-machine ssh weave-1
# WHALE PICTURE
$ ps aux | grep docker
# ... docker has pid 2230
$ cat /proc/2230/cmdline | tr '\0' ' '
/usr/local/bin/docker daemon -D -g /var/lib/docker -H unix:// -H tcp://0.0.0.0:2376 --label provider=virtualbox --tlsverify --tlscacert=/var/lib/boot2docker/ca.pem --tlscert=/var/lib/boot2docker/server.pem --tlskey=/var/lib/boot2docker/server-key.pem -s aufs
```

Ah. No `-d` flag there, which is what `docker_tls_args` looks for -- apparently it's now `daemon`.

squaremo:
(By the way, that's docker v1.8.2)

rade:
> `$ docker-machine start weave-1 # seems to be a no-op?`

Is it? If so, let's get rid of it.

squaremo:
Using a rebuilt image, my transcript from above works smoothly now. Yay!

paulbellamy:
@rade, If you're happy with the docs, this is ready to go.

rade:
please file issues for points 3 and 4 of https://github.com/weaveworks/weave/pull/1414#issuecomment-140035091

rade:
@paulbellamy btw, this is a huge improvement. Well done!

abuehrle:
Let me know when this is merged so I can update the guides.  Does this now work without grepping for the settings?

rade:
> Let me know when this is merged so I can update the guides. 

See https://github.com/weaveworks/weave/pull/1414#issuecomment-140745507. Some code changes are part of that too.

But you need to be careful here, since obviously the new functionality won't be available until 1.2 is out.

>  Does this now work without grepping for the settings?

Yes! Even better, it means we no longer have to launch the proxy separately. Everything now just works as on a normal/sane platform.

abuehrle:
Thanks, that’s great!   I’ll change the docs in a PR and wait for your release then.  In the meantime, I will include the hack until then.

rade:
> please file issues for points 3 and 4 of https://github.com/weaveworks/weave/pull/1414#issuecomment-140035091

https://github.com/weaveworks/guides/issues/73
https://github.com/weaveworks/guides/issues/74

PR #1425, opened 2015-09-14
Handle cancelled allocation in 'weave attach'

### LINKED ISSUE #1418: mangled error when `weave attach` is cancelled
```
$ weave launch --init-peer-count=2
$ eval $(weave env)
$ docker run -ti ubuntu
^CError response from daemon: invalid CIDR address: Allocate
```

Instrumenting the proxy's `weave attach` invocation yields

```
...
+ curl --connect-timeout 3 -s -S -X POST http://172.17.0.10:6784/ip/db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab?check-alive=true
+ IPAM_CIDRS=Allocate request for db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab cancelled
+ [ Allocate request for db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab cancelled = 404 page not found ]
+ ALL_CIDRS=Allocate request for db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab cancelled
+ [ 0 -gt 0 ]
+ [ -n 1 ]
+ command_exists weavehosts
+ command -v weavehosts
+ docker inspect -f {{.HostsPath}} {{.Config.Hostname}} db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab
+ PATH_AND_NAME=/var/lib/docker/containers/db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab/hosts db1bb81e6866
+ weavehosts /var/lib/docker/containers/db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab/hosts db1bb81e6866 Allocate request for db1bb81e68660dda3bd19e501fec236ff1da6e1d6a63a067a457468ea7492dab cancelled
```

i.e. the IP allocation request returns a message which we subsequently attempt to parse as a CIDR.

Deduced from #1417.

bboreham:
Make IPAM return a simpler message and have the weave script check for that message and exit with status 0, so the proxy does not print any error message.

Fixes #1418

### DIFF
--- a/ipam/http.go
+++ b/ipam/http.go
@@ -35,12 +35,18 @@ func doAllocate(alloc *Allocator, dockerCli *docker.Client, w http.ResponseWrite
 	ident := vars["id"]
 	addr, err := alloc.Allocate(ident, subnet.HostRange(), closedChan)
 	if err != nil {
+		if _, ok := err.(*errorCancelled); ok { // cancellation is not really an error
+			common.Log.Infoln("[allocator]:", err.Error())
+			fmt.Fprint(w, "cancelled")
+			return
+		}
 		badRequest(w, err)
 		return
 	}
 	if r.FormValue("check-alive") == "true" && dockerCli != nil && dockerCli.IsContainerNotRunning(ident) {
 		common.Log.Infof("[allocator] '%s' is not running: freeing %s", ident, addr)
 		alloc.Free(ident, addr)
+		fmt.Fprint(w, "cancelled")
 		return
 	}
 
--- a/ipam/http.go
+++ b/ipam/http.go
@@ -16,6 +16,37 @@ func badRequest(w http.ResponseWriter, err error) {
 	common.Log.Warningln("[allocator]:", err.Error())
 }
 
+func extractCIDR(w http.ResponseWriter, vars map[string]string) (address.CIDR, bool) {
+	cidrStr := vars["ip"] + "/" + vars["prefixlen"]
+	subnetAddr, cidr, err := address.ParseCIDR(cidrStr)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


ipam/http.go
```diff
@@ -16,6 +16,43 @@ func badRequest(w http.ResponseWriter, err error) {
...
+			fmt.Fprint(w, "cancelled")
+			return
+		}
+		badRequest(w, err)
+		return
+	}
+	if r.FormValue("check-alive") == "true" && dockerCli != nil && dockerCli.IsContainerNotRunning(ident) {
+		common.Log.Infof("[allocator] '%s' is not running: freeing %s", ident, addr)
+		alloc.Free(ident, addr)
+		fmt.Fprint(w, "cancelled")
```
bboreham: Note this is a case not called out by #1418, but it seems reasonable to do the same thing - previously it would return blank which would cause an error elsewhere in the weave script.

bboreham:
Wrong branch; replaced by #1426

weave
```diff
@@ -933,6 +935,8 @@ ipam_cidrs() {
             if [ "$CIDR" = "404 page not found" ] ; then
                 echo "IP address allocation must be enabled to use 'net:'" >&2
                 return 1
+            elif [ "$IPAM_CIDRS" = "cancelled" ] ; then
```
rade: shouldn't that be `"$CIDR" = ...`?

bboreham: well spotted

PR #1426, opened 2015-09-14
Handle cancelled allocation in 'weave attach'

### LINKED ISSUE #1418: mangled error when `weave attach` is cancelled

bboreham:
Make IPAM return a simpler message and have the weave script check for that message and exit with status 0, so the proxy does not print any error message.

Fixes #1418

(replacement for #1425 which was against master not 1.1)

### DIFF
--- a/ipam/allocate.go
+++ b/ipam/allocate.go
@@ -59,7 +59,7 @@ func (g *allocate) Try(alloc *Allocator) bool {
 }
 
 func (g *allocate) Cancel() {
-	g.resultChan <- allocateResult{0, fmt.Errorf("Allocate request for %s cancelled", g.ident)}
+	g.resultChan <- allocateResult{0, &errorCancelled{"Allocate", g.ident}}
 }
 
 func (g *allocate) ForContainer(ident string) bool {
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -184,6 +184,15 @@ func hasBeenCancelled(cancelChan <-chan bool) func() bool {
 	}
 }
 
+type errorCancelled struct {
+	kind  string
+	ident string
+}
+
+func (e *errorCancelled) Error() string {
+	return fmt.Sprintf("%s request for %s cancelled", e.kind, e.ident)
+}
+
 // Actor client API
 
 // Allocate (Sync) - get new IP address for container with given name in range
--- a/ipam/claim.go
+++ b/ipam/claim.go
@@ -95,7 +95,7 @@ func (c *claim) deniedBy(alloc *Allocator, owner router.PeerName) {
 }
 
 func (c *claim) Cancel() {
-	c.sendResult(fmt.Errorf("Operation cancelled."))
+	c.sendResult(&errorCancelled{"Claim", c.ident})
 }
 
 func (c *claim) ForContainer(ident string) bool {
--- a/ipam/http.go
+++ b/ipam/http.go
@@ -16,6 +16,35 @@ func badRequest(w http.ResponseWriter, err error) {
 	common.Log.Warningln("[allocator]:", err.Error())
 }
 
+func parseCIDR(w http.ResponseWriter, cidrStr string) (address.CIDR, bool) {
+	subnetAddr, cidr, err := address.ParseCIDR(cidrStr)
+	if err != nil {
+		badRequest(w, err)
+		return address.CIDR{}, false
+	}
+	if cidr.Start != subnetAddr {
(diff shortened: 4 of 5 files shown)


rade:
@bboreham please take a look at my changes. If you are happy I will squash them, though squashing them into your first commit will be tricky, so may just have to squash the lot.

bboreham:
LGTM.  I can do the squash into 2 commits - have already done it on my own branch.

bboreham:
Squashed.

rade:
Two issues:

1) There is an annoying blank line after the "cancelled" warning in the proxy logs:
```
INFO: 2015/09/21 19:54:47.023571 Attaching container 2f47c1e67e1b1c49e5aabbc6f0581b7ed04ada93f64c0d91f27d28cb4564bc8f with WEAVE_CIDR "" to weave network
WARN: 2015/09/21 19:54:48.690007 Attaching container 2f47c1e67e1b1c49e5aabbc6f0581b7ed04ada93f64c0d91f27d28cb4564bc8f to weave network: cancelled

INFO: 2015/09/21 19:54:48.690454 POST /v1.19/containers/2f47c1e67e1b1c49e5aabbc6f0581b7ed04ada93f64c0d91f27d28cb4564bc8f/resize?h=24&w=80
```

2) In https://github.com/weaveworks/weave/issues/1418#issuecomment-140030376 we observed that the cancellation is triggered by container death. Surely this can occur for most of the operations that invoke `ipam_cidrs`. And, furthermore, additional cancellation triggers may be introduced in the future. Yet only in `attach` do we check for the new `$IPAM_CIDRS = "cancelled"`. Elsewhere we simply carry on. And in most cases we don't even look at `$IPAM_CIDRS` at all, only `$ALL_CIDRS` (which will be empty or contain CIDRS obtained prior to cancellation). The only thing saving us here is that the subsequent operations will fail if the container has died. This seems rather fragile. Also, what happens if the container dies just before we call `ipam_cidrs`? I did a quick experiment of adding `docker kill $CONTAINER` just before the `ipam_cidrs` call in `attach`. This causes `docker run` to stall. So we get different behaviour depending on when the container died.

bboreham:
>  I did a quick experiment of adding `docker kill $CONTAINER` just before the `ipam_cidrs` call in `attach`. 
> This causes docker run to stall.

I don't think it does, per se.  This is just the standard user-experience of weave IPAM stalling because it hasn't reached consensus.

rade:
> This is just the standard user-experience of weave IPAM stalling because it hasn't reached consensus.

are you saying that cancelling container startup just before reaching ipam_cidrs results in a stall, but doing so during ipam_cidrs doesn't? That can't be right.

PR #1429, opened 2015-09-14
Use explicit index into operations slice in cancelOpsFor()

### LINKED ISSUE #1424: Cancelling two simultaneous IP allocation requests crashes weave
Due to dumb coding in `alloc.cancelOpsFor()`

To reproduce:

````
$ sudo weave launch --init-peer-count=2 
$ sudo docker run -tdi ubuntu
df7fe26b8d870e7090662b4e36edf988c4d1df0545a6ee1e1bcaf3999ebe2486
$ sudo weave attach df7 &
[1] 32082
$ sudo weave attach df7 &
[2] 32090
$ sudo docker kill df7
[attach commands fail and exit]
$ sudo docker logs weave
...
panic: runtime error: slice bounds out of range

goroutine 16 [running]:
github.com/weaveworks/weave/ipam.(*Allocator).cancelOpsFor(0xc2090d4300, 0xc2090d4360, 0xc2092e3100, 0x40, 0xd261a0)
	/home/bryan/golang/src/github.com/weaveworks/weave/ipam/allocator.go:133 +0x2fa
github.com/weaveworks/weave/ipam.func·004()
	/home/bryan/golang/src/github.com/weaveworks/weave/ipam/allocator.go:249 +0x1b6
...
````

(note this was on a dev branch so line numbers may be slightly off, but the bug is on master)

### PR DESCRIPTION
**Pull request by bboreham**
to avoid mis-indexing when more than one op is removed

Fixes #1424

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -131,9 +131,9 @@ func (alloc *Allocator) cancelOpsFor(ops *[]operation, ident string) bool {
 			found = true
 			op.Cancel()
 			*ops = append((*ops)[:i], (*ops)[i+1:]...)
-			continue
+		} else {
+			i++
 		}
-		i++
 	}
 	return found
 }
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -126,12 +126,14 @@ func (alloc *Allocator) cancelOps(ops *[]operation) {
 // if we found any.
 func (alloc *Allocator) cancelOpsFor(ops *[]operation, ident string) bool {
 	var found bool
-	for i, op := range *ops {
-		if op.ForContainer(ident) {
+	for i := 0; i < len(*ops); {
+		if op := (*ops)[i]; op.ForContainer(ident) {
 			found = true
 			op.Cancel()
 			*ops = append((*ops)[:i], (*ops)[i+1:]...)
+			continue
 		}
+		i++
 	}
 	return found
 }
--- a/ipam/allocator_test.go
+++ b/ipam/allocator_test.go
@@ -240,6 +240,42 @@ func TestCancel(t *testing.T) {
 	}
 }
 
+func TestCancelOnDied(t *testing.T) {
+	const (
+		CIDR       = "10.0.1.7/26"
+		container1 = "abcdef"
+	)
+
+	router := gossip.NewTestRouter(0.0)
+
+	alloc1, subnet := makeAllocator("01:00:00:02:00:00", CIDR, 2)
+	alloc1.SetInterfaces(router.Connect(alloc1.ourName, alloc1))
+
+	alloc1.Start()
+
+	doneChan := make(chan bool)
+	go func() {
+		_, ok := alloc1.Allocate(container1, subnet, nil)
+		doneChan <- ok == nil
+	}()
+
+	// Do another allocation in parallel, to check that is handled correctly
+	go func() {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: ipam/allocator_test.go [hunk 1]
```diff
@@ -240,6 +240,42 @@ func TestCancel(t *testing.T) {
...
+		doneChan <- ok == nil
+	}()
+
+	AssertNothingSent(t, doneChan)
+	time.Sleep(100 * time.Millisecond)
+	AssertNothingSent(t, doneChan)
+
+	alloc1.ContainerDied(container1)
+
+	if <-doneChan || <-doneChan {
```
rade: huh?

ipam/allocator_test.go
```diff
@@ -240,6 +240,42 @@ func TestCancel(t *testing.T) {
...
+	go func() {
+		_, ok := alloc1.Allocate(container1, subnet, nil)
+		doneChan <- ok == nil
+	}()
+
+	// Do another allocation in parallel, to check that is handled correctly
+	go func() {
+		_, ok := alloc1.Allocate(container1, subnet, nil)
+		doneChan <- ok == nil
+	}()
```
rade: you could define the function just once.

ipam/allocator_test.go
rade: ah, I get it now.

ipam/allocator_test.go
```diff
@@ -240,6 +240,42 @@ func TestCancel(t *testing.T) {
...
+
+	AssertNothingSent(t, doneChan)
```
rade: why do we need this assert? Isn't the 2nd one sufficient?

ipam/allocator.go
```diff
@@ -126,12 +126,14 @@ func (alloc *Allocator) cancelOps(ops *[]operation) {
...
-	for i, op := range *ops {
-		if op.ForContainer(ident) {
+	for i := 0; i < len(*ops); {
+		if op := (*ops)[i]; op.ForContainer(ident) {
 			found = true
 			op.Cancel()
 			*ops = append((*ops)[:i], (*ops)[i+1:]...)
+			continue
 		}
+		i++
```
rade: I'd ditch the continue and do the increment in an `else` branch.

bboreham: tut; one line longer...

rade: yes, but more readable.

ipam/allocator_test.go
bboreham: Yes, I guess it is.  The code parallels `TestCancel` which does the same thing, presumably for troubleshooting.

ipam/allocator.go
rade: I've been reading the golang spec and cannot figure out what is going wrong with the original code. There is obviously some issue with modifying the slice while iterating over it. Presumably `i` ends up pointing beyond the end of the slice. But why? And what is the value of `op`?

ipam/allocator_test.go
rade: Ditch it in `TestCancel` on the 1.1 branch then.

ipam/allocator.go
bboreham: The relevant part is:

> The range expression is evaluated once before beginning the loop

The failure symptom is that `(*ops)[i+1:]` is invalid because `i` is now equal to the length of the array, which has been shortened.

PR #1428, opened 2015-09-14
Update docker repo, keys and package name

awh:
Docker 1.8+ uses a new package name (`docker-engine`) and repo configuration. This PR updates the build and smoketest Vagrantfiles with the new settings to get the latest release of Docker (currently 1.8.2)

PR #1435, opened 2015-09-15
Update docker repo, keys and package name

awh:
Replacement for #1428, now targetted at 1.1 branch.

Docker 1.8+ uses a new package name (docker-engine) and repo configuration. This PR updates the build and smoketest Vagrantfiles with the new settings to get the latest release of Docker (currently 1.8.2)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@ require './vagrant-common.rb'
 vm_ip = "172.16.0.3" # arbitrary private IP
 
 pkgs = %w(
-  lxc-docker
+  docker-engine
   aufs-tools
   build-essential
   ethtool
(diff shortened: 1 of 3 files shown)


rade:
What about the [build container](https://github.com/weaveworks/weave/blob/master/build/Dockerfile) and [Ubuntu build instructions](http://docs.weave.works/weave/latest_release/building.html)? Both reference `lxc-docker`.

awh:
> What about the build container and Ubuntu build instructions? Both reference lxc-docker.

The build container is pinned to Docker 1.3.1; is there a reason to change it now, when we didn't for 1.4, 1.5, 1.6 & 1.7? If not, then the package name `lxc-docker` is correct, for that version.

Mentioning @dpw, as he has previously asked to be consulted before changes are made to the build container...

rade:
>  Ubuntu build instructions

Fixed in 2b5c287.

PR #1438, opened 2015-09-15
Fast datapath

dpw:
Smoke tests pass (modulo the one that is broken on master), so I think this is fit for a review pass.

Things still to do:

* The `weave` script changes are still a bit grotty.  But...
* I think there needs to be a "disable it all" option to weave, which disables fastdp, `--net=host`, and the MTU change.
* I haven't modified architecture.txt. To be honest, most of the contents are more like implementation notes that belong as comments in the code.  If the fast datapath code requires additional notes of that nature, I'd be happy to add them as comments. A high-level overview of fast datapath would be a good thing, but I think I'll need to re-write the whole of architecture.txt to match.
* 1.1 compatibility support.  That will be a separate issue.  Short ids is the tricky part here.
* I need to check out a kernel oops I saw when testing with ubuntu 12.04 a few days ago.

### DIFF
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -1,0 +1,1115 @@
+package router
+
+import (
+	"encoding/binary"
+	"fmt"
+	"net"
+	"sync"
+	"time"
+
+	"github.com/weaveworks/go-odp/odp"
+)
+
+// The virtual bridge accepts packets from ODP vports and the router
+// port (i.e. InjectPacket).  We need a map key to index those
+// possibilities:
+type bridgePortID struct {
+	vport  odp.VportID
+	router bool
+}
+
+// A bridgeSender sends out a packet from the virtual bridge
+type bridgeSender func(key PacketKey, lock *fastDatapathLock) FlowOp
+
+// A missHandler handles an ODP miss
+type missHandler func(fks odp.FlowKeys, lock *fastDatapathLock) FlowOp
+
+type FastDatapath struct {
+	dpname string
+
+	// The mtu from the datapath netdev (which should match the
+	// mtus on all the veths hooked up to the datapath).  We
+	// validate that we are able to support that mtu.
+	mtu int
+
+	// The lock guards the FastDatapath state, and also
+	// synchronizes use of the dpif
+	lock             sync.Mutex
+	dpif             *odp.Dpif
+	dp               odp.DatapathHandle
+	deleteFlowsCount uint64
+	missHandlers     map[odp.VportID]missHandler
+	localPeer        *Peer
+	peers            *Peers
+	overlayConsumer  OverlayConsumer
+
+	// Bridge state: How to send to the given bridge port
+	sendToPort map[bridgePortID]bridgeSender
+
+	// How to send to a given destination MAC
+	sendToMAC map[MAC]bridgeSender
+
+	// MACs seen on the bridge recently
+	seenMACs map[MAC]struct{}
+
+	// vxlan vports associated with the given UDP ports
+	vxlanVportIDs    map[int]odp.VportID
+	mainVxlanVportID odp.VportID
+
+	// A singleton pool for the occasions when we need to decode
+	// the packet.
+	dec *EthernetDecoder
+
+	// forwarders by remote peer
+	forwarders map[PeerName]*fastDatapathForwarder
+}
+
+type FastDatapathConfig struct {
+	DatapathName        string
+	Port                int
+	ExpireFlowsInterval time.Duration
+	ExpireMACsInterval  time.Duration
+}
+
+func NewFastDatapath(config FastDatapathConfig) (*FastDatapath, error) {
+	dpif, err := odp.NewDpif()
+	if err != nil {
+		return nil, err
+	}
+
+	success := false
+	defer func() {
+		if !success {
+			dpif.Close()
+		}
+	}()
+
+	dp, err := dpif.LookupDatapath(config.DatapathName)
+	if err != nil {
+		return nil, err
+	}
+
+	iface, err := net.InterfaceByName(config.DatapathName)
+	if err != nil {
+		return nil, err
+	}
+
+	fastdp := &FastDatapath{
+		dpname:        config.DatapathName,
+		mtu:           iface.MTU,
+		dpif:          dpif,
+		dp:            dp,
+		missHandlers:  make(map[odp.VportID]missHandler),
+		sendToPort:    nil,
+		sendToMAC:     make(map[MAC]bridgeSender),
+		seenMACs:      make(map[MAC]struct{}),
+		vxlanVportIDs: make(map[int]odp.VportID),
+		forwarders:    make(map[PeerName]*fastDatapathForwarder),
+	}
+
+	if err := fastdp.deleteVxlanVports(); err != nil {
+		return nil, err
+	}
+
+	if err := fastdp.deleteFlows(); err != nil {
+		return nil, err
+	}
+
+	// We use the weave port number plus 1 for vxlan.  Hard-coding
+	// this relationship may seem dubious, but there is no moral
+	// difference between this and requiring that the sleeve UDP
+	// port number is the same as the TCP port number.  The hard
+	// part would be not adding weaver flags to allow the port
+	// numbers to be independent, but working out how to specify
+	// them on the connecting side.  So we can wait to find out if
+	// anyone wants that.
+	fastdp.mainVxlanVportID, err = fastdp.getVxlanVportID(config.Port + 1)
+	if err != nil {
+		return nil, err
+	}
+
+	// need to lock before we might receive events
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+
+	if _, err := dp.ConsumeMisses(fastdp); err != nil {
+		return nil, err
+	}
+
+	if _, err := dp.ConsumeVportEvents(fastdp); err != nil {
+		return nil, err
+	}
+
+	vports, err := dp.EnumerateVports()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, vport := range vports {
+		fastdp.makeBridgeVport(vport)
+	}
+
+	success = true
+	go fastdp.run()
+	return fastdp, nil
+}
+
+func (fastdp *FastDatapath) Close() error {
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+	err := fastdp.dpif.Close()
+	fastdp.dpif = nil
+	return err
+}
+
+// While processing a packet, we can potentially acquire and drop the
+// FastDatapath lock many times (acquiring it to acceess FastDatapath
+// state, and invoke ODP operations; dropping it to invoke callbacks
+// that may re-enter the FastDatapath).  A fastDatapathLock
+// coordinates this process.
+type fastDatapathLock struct {
+	fastdp *FastDatapath
+	locked bool
+
+	// While the lock is dropped, deleteFlows could be called.  We
+	// need to detect when this happens and avoid creating flows,
+	// because they may be based on stale information.
+	deleteFlowsCount uint64
+}
+
+func (fastdp *FastDatapath) startLock() fastDatapathLock {
+	fastdp.lock.Lock()
+	return fastDatapathLock{
+		fastdp:           fastdp,
+		locked:           true,
+		deleteFlowsCount: fastdp.deleteFlowsCount,
+	}
+}
+
+func (lock *fastDatapathLock) unlock() {
+	if lock.locked {
+		lock.fastdp.lock.Unlock()
+		lock.locked = false
+	}
+}
+
+func (lock *fastDatapathLock) relock() {
+	if !lock.locked {
+		lock.fastdp.lock.Lock()
+		lock.locked = true
+	}
+}
+
+// Bridge bits
+
+type fastDatapathBridge struct {
+	*FastDatapath
+}
+
+func (fastdp *FastDatapath) Bridge() Bridge {
+	return fastDatapathBridge{fastdp}
+}
+
+func (fastdp fastDatapathBridge) String() string {
+	return fmt.Sprint(fastdp.dpname, " (via ODP)")
+}
+
+func (fastDatapathBridge) Stats() map[string]int {
+	return nil
+}
+
+var routerBridgePortID = bridgePortID{router: true}
+
+func (fastdp fastDatapathBridge) StartConsumingPackets(consumer BridgeConsumer) error {
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+
+	if fastdp.sendToPort[routerBridgePortID] != nil {
+		return fmt.Errorf("FastDatapath already has a BridgeConsumer")
+	}
+
+	// set up delivery to the weave router port on the bridge
+	fastdp.addSendToPort(routerBridgePortID,
+		func(key PacketKey, lock *fastDatapathLock) FlowOp {
+			// drop the FastDatapath lock in order to call
+			// the consumer
+			lock.unlock()
+			return consumer(key)
+		})
+	return nil
+}
+
+func (fastdp fastDatapathBridge) InjectPacket(key PacketKey) FlowOp {
+	lock := fastdp.startLock()
+	defer lock.unlock()
+	return fastdp.bridge(routerBridgePortID, key, &lock)
+}
+
+// Ethernet bridge implementation
+
+func (fastdp *FastDatapath) bridge(ingress bridgePortID,
+	key PacketKey, lock *fastDatapathLock) FlowOp {
+	lock.relock()
+	if fastdp.sendToMAC[key.SrcMAC] == nil {
+		// Learn the source MAC
+		fastdp.sendToMAC[key.SrcMAC] = fastdp.sendToPort[ingress]
+		fastdp.seenMACs[key.SrcMAC] = struct{}{}
+	}
+
+	// If we know about the destination MAC, deliver it to the
+	// associated port.
+	if sender := fastdp.sendToMAC[key.DstMAC]; sender != nil {
+		return NewMultiFlowOp(false, odpEthernetFlowKey(key),
+			sender(key, lock))
+	}
+
+	// Otherwise, it might be a real broadcast, or it might
+	// be for a MAC we don't know about yet.  Either way, we'll
+	// broadcast it.
+	mfop := NewMultiFlowOp(false)
+
+	if (key.DstMAC[0] & 1) == 0 {
+		// Not a real broadcast, so don't create a flow rule.
+		// If we did, we'd need to delete the flows every time
+		// we learned a new MAC address, or have a more
+		// complicated selective invalidation scheme.
+		mfop.Add(vetoFlowCreationFlowOp{})
+	} else {
+		// A real broadcast
+		mfop.Add(odpEthernetFlowKey(key))
+	}
+
+	// Send to all ports except the one it came in on. The
+	// sendToPort map is immutable, so it is safe to iterate over
+	// it even though the sender functions can drop the
+	// fastDatapathLock
+	for id, sender := range fastdp.sendToPort {
+		if id != ingress {
+			mfop.Add(sender(key, lock))
+		}
+	}
+
+	return mfop
+}
+
+// Overlay bits
+
+type fastDatapathOverlay struct {
+	*FastDatapath
+}
+
+func (fastdp *FastDatapath) Overlay() Overlay {
+	return fastDatapathOverlay{fastdp}
+}
+
+func (fastdp fastDatapathOverlay) InvalidateRoutes() {
+	log.Debug("InvalidateRoutes")
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+	checkWarn(fastdp.deleteFlows())
+}
+
+func (fastdp fastDatapathOverlay) InvalidateShortIDs() {
+	log.Debug("InvalidateShortIDs")
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+	checkWarn(fastdp.deleteFlows())
+}
+
+func (fastDatapathOverlay) AddFeaturesTo(features map[string]string) {
+	// Nothing needed.  Fast datapath support is indicated through
+	// OverlaySwitch.
+}
+
+func (fastdp fastDatapathOverlay) StartConsumingPackets(localPeer *Peer,
+	peers *Peers, consumer OverlayConsumer) error {
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+
+	if fastdp.overlayConsumer != nil {
+		return fmt.Errorf("FastDatapath already has an OverlayConsumer")
+	}
+
+	fastdp.localPeer = localPeer
+	fastdp.peers = peers
+	fastdp.overlayConsumer = consumer
+	return nil
+}
+
+func (fastdp *FastDatapath) getVxlanVportID(udpPort int) (odp.VportID, error) {
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+
+	if vxlanVportID, present := fastdp.vxlanVportIDs[udpPort]; present {
+		return vxlanVportID, nil
+	}
+
+	vxlanVportID, err := fastdp.dp.CreateVport(
+		odp.NewVxlanVportSpec(fmt.Sprintf("vxlan-%d", udpPort),
+			uint16(udpPort)))
+	if err != nil {
+		return 0, err
+	}
+
+	fastdp.vxlanVportIDs[udpPort] = vxlanVportID
+	fastdp.missHandlers[vxlanVportID] = func(fks odp.FlowKeys, lock *fastDatapathLock) FlowOp {
+		tunnel := fks[odp.OVS_KEY_ATTR_TUNNEL].(odp.TunnelFlowKey)
+		tunKey := tunnel.Key()
+
+		lock.relock()
+		consumer := fastdp.overlayConsumer
+		if consumer == nil {
+			return vetoFlowCreationFlowOp{}
+		}
+
+		srcPeer, dstPeer := fastdp.extractPeers(tunKey.TunnelId)
+		if srcPeer == nil || dstPeer == nil {
+			return vetoFlowCreationFlowOp{}
+		}
+
+		lock.unlock()
+		pk := flowKeysToPacketKey(fks)
+		var zeroMAC MAC
+		if pk.SrcMAC == zeroMAC && pk.DstMAC == zeroMAC {
+			return vxlanSpecialPacketFlowOp{fastdp, srcPeer,
+				&net.UDPAddr{
+					IP:   net.IP(tunKey.Ipv4Src[:]),
+					Port: udpPort,
+				},
+			}
+		}
+
+		key := ForwardPacketKey{
+			SrcPeer:   srcPeer,
+			DstPeer:   dstPeer,
+			PacketKey: pk,
+		}
+
+		var tunnelFlowKey odp.TunnelFlowKey
+		tunnelFlowKey.SetTunnelId(tunKey.TunnelId)
+		tunnelFlowKey.SetIpv4Src(tunKey.Ipv4Src)
+		tunnelFlowKey.SetIpv4Dst(tunKey.Ipv4Dst)
+
+		return NewMultiFlowOp(false, odpFlowKey(tunnelFlowKey),
+			consumer(key))
+	}
+
+	return vxlanVportID, nil
+}
+
+func (fastdp *FastDatapath) extractPeers(tunnelID [8]byte) (*Peer, *Peer) {
+	vni := binary.BigEndian.Uint64(tunnelID[:])
+	srcPeer := fastdp.peers.FetchByShortID(PeerShortID(vni & 0xfff))
+	dstPeer := fastdp.peers.FetchByShortID(PeerShortID((vni >> 12) & 0xfff))
+	return srcPeer, dstPeer
+}
+
+type vxlanSpecialPacketFlowOp struct {
+	fastdp  *FastDatapath
+	srcPeer *Peer
+	sender  *net.UDPAddr
+}
+
+func (op vxlanSpecialPacketFlowOp) Process(frame []byte, dec *EthernetDecoder,
+	broadcast bool) {
+	op.fastdp.lock.Lock()
+	fwd := op.fastdp.forwarders[op.srcPeer.Name]
+	op.fastdp.lock.Unlock()
+
+	if fwd != nil && dec.IsSpecial() {
+		fwd.handleVxlanSpecialPacket(frame, op.sender)
+	}
+}
+
+type fastDatapathForwarder struct {
+	fastdp         *FastDatapath
+	remotePeer     *Peer
+	localIP        [4]byte
+	sendControlMsg func(byte, []byte) error
+	connUID        uint64
+	vxlanVportID   odp.VportID
+
+	lock              sync.RWMutex
+	confirmed         bool
+	remoteAddr        *net.UDPAddr
+	heartbeatInterval time.Duration
+	heartbeatTimer    *time.Timer
+	heartbeatTimeout  *time.Timer
+	ackedHeartbeat    bool
+	stopChan          chan struct{}
+	stopped           bool
+
+	establishedChan chan struct{}
+	errorChan       chan error
+}
+
+func (fastdp fastDatapathOverlay) MakeForwarder(
+	params ForwarderParams) (OverlayForwarder, error) {
+	if params.Crypto != nil {
+		// No encryption suport in fastdp.  The weaver main.go
+		// is responsible for ensuring this doesn't happen.
+		log.Fatal("Attempt to use FastDatapath with encryption")
+	}
+
+	vxlanVportID := fastdp.mainVxlanVportID
+	remoteAddr := params.RemoteAddr
+	if remoteAddr != nil {
+		// The provided address contains the main weave port
+		// number to connect to.  We need to derive the vxlan
+		// port number from that.
+		vxlanRemoteAddr := *params.RemoteAddr
+		vxlanRemoteAddr.Port++
+		remoteAddr = &vxlanRemoteAddr
+
+		var err error
+		vxlanVportID, err = fastdp.getVxlanVportID(remoteAddr.Port)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	localIP, err := ipv4Bytes(params.LocalIP)
+	if err != nil {
+		return nil, err
+	}
+
+	fwd := &fastDatapathForwarder{
+		fastdp:         fastdp.FastDatapath,
+		remotePeer:     params.RemotePeer,
+		localIP:        localIP,
+		sendControlMsg: params.SendControlMessage,
+		connUID:        params.ConnUID,
+		vxlanVportID:   vxlanVportID,
+
+		remoteAddr:        remoteAddr,
+		heartbeatInterval: FastHeartbeat,
+		stopChan:          make(chan struct{}),
+
+		establishedChan: make(chan struct{}),
+		errorChan:       make(chan error, 1),
+	}
+
+	return fwd, err
+}
+
+func ipv4Bytes(ip net.IP) (res [4]byte, err error) {
+	ipv4 := ip.To4()
+	if ipv4 != nil {
+		copy(res[:], ipv4)
+	} else {
+		err = fmt.Errorf("IP address %s is not IPv4", ip)
+	}
+	return
+}
+
+func (fwd *fastDatapathForwarder) logPrefix() string {
+	return fmt.Sprintf("fastdp ->[%s|%s]: ", fwd.remoteAddr, fwd.remotePeer)
+}
+
+func (fwd *fastDatapathForwarder) Confirm() {
+	fwd.lock.Lock()
+	defer fwd.lock.Unlock()
+
+	if fwd.confirmed {
+		log.Fatal(fwd.logPrefix(), "already confirmed")
+	}
+
+	log.Debug(fwd.logPrefix(), "confirmed")
+	fwd.fastdp.addForwarder(fwd.remotePeer.Name, fwd)
+	fwd.confirmed = true
+
+	if fwd.remoteAddr != nil {
+		// have the goroutine send a heartbeat straight away
+		fwd.heartbeatTimer = time.NewTimer(0)
+	} else {
+		// we'll reset the timer when we learn the remote ip
+		fwd.heartbeatTimer = time.NewTimer(MaxDuration)
+	}
+
+	fwd.heartbeatTimeout = time.NewTimer(HeartbeatTimeout)
+	go fwd.doHeartbeats()
+}
+
+func (fwd *fastDatapathForwarder) EstablishedChannel() <-chan struct{} {
+	return fwd.establishedChan
+}
+
+func (fwd *fastDatapathForwarder) ErrorChannel() <-chan error {
+	return fwd.errorChan
+}
+
+func (fwd *fastDatapathForwarder) doHeartbeats() {
+	var err error
+
+	for err == nil {
+		select {
+		case <-fwd.heartbeatTimer.C:
+			if fwd.confirmed {
+				fwd.sendHeartbeat()
+			}
+			fwd.heartbeatTimer.Reset(fwd.heartbeatInterval)
+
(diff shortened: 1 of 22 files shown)


router/peers.go
```diff
@@ -4,14 +4,29 @@ import (
...
+	byShortID map[PeerShortID]ShortIDPeers
+	onGC      []func(*Peer)
+
+	// Called when the mapping from short ids to peers changes
+	onInvalidateShortIDs []func()
+}
+
+type ShortIDPeers struct {
+	// If we know about a single peer with the short id, this is
+	// that peer.  If there is a coliision, this is the peer with
```
awh: collision

router/peers.go
```diff
@@ -31,32 +46,223 @@ type ConnectionSummary struct {
...
+		entry.peer = peer
+		pending.invalidateShortIDs = true
+	} else if peer.Name < entry.peer.Name {
+		// Short ID collision, this peer becomes the principal
+		// peer for the short ID, bumping the previous one
+		// into others.
+
+		if entry.peer == peers.ourself.Peer {
+			// The bumped peer is peers.ourself, so we
+			// need to look foor a new short id
```
awh: for

router/sleeve.go
```diff
@@ -287,9 +291,12 @@ type sleeveForwarder struct {
 	confirmedChan    chan<- struct{}
 	finishedChan     <-chan struct{}
 
+	// listener channels
+	establishedChan chan struct{}
+	errorChan       chan error
```
awh: These should be `<-chan`?

router/sleeve.go
```diff
@@ -363,6 +370,8 @@ func (sleeve *SleeveOverlay) MakeForwarder(params ForwarderParams) (OverlayForwa
 		specialChan:      specialChan,
 		confirmedChan:    confirmedChan,
 		finishedChan:     finishedChan,
+		establishedChan:  make(chan struct{}),
+		errorChan:        make(chan error, 1),
```
awh: It's odd that these two channels are initialised without the use of intermediary variables as with all the other chans in this struct

router/sleeve.go
dpw: Nope.  Look at how those fields get used.

router/sleeve.go
dpw: The difference is due to a very different pattern of usage from the other channels.  Which fundamentally is because the other channels are an internal implementation detail, for communication with an internal goroutine, whereas these two are part of the OverlayForwarder interface.

router/fastdp.go
```diff
@@ -0,0 +1,1115 @@
...
+			sender(key, lock))
+	}
+
+	// Otherwise, it might be a real broadcast, or it might
+	// be for a MAC we don't know about yet.  Either way, we'll
+	// broadcast it.
+	mfop := NewMultiFlowOp(false)
+
+	if (key.DstMAC[0] & 1) == 0 {
+		// Not a real broadcast, so doon't create a flow rule.
```
awh: don't

weave
```diff
@@ -365,33 +372,57 @@ random_mac() {
...
-        # Attempting to set the bridge MTU to a high value directly
-        # fails. Bridges take the lowest MTU of their interfaces. So
-        # instead we create a temporary interface with the desired
-        # MTU, attach that to the bridge, and then remove it again.
-        ip link add name v${CONTAINER_IFNAME}du mtu $MTU type dummy
-        ip link set dev v${CONTAINER_IFNAME}du master $BRIDGE
-        ip link del dev v${CONTAINER_IFNAME}du
-        # Drop traffic from Docker bridge to Weave; it can break subnet isolation
+        # Checking for a specific exit status while set -e is in
+        # effect requires contortions:
```
awh:
````
if docker run --rm --privileged --net=host $IMAGE $COVERAGE_ARGS --create-datapath --datapath=$BRIDGE ; then
    BRIDGE_TYPE=fdp
elif [ $? = 17 ] ; then
    BRIDGE_TYPE=bridge
else
    return 1
fi
````

File: router/sleeve.go (hunk 3 above)
Comment (awh) [inline router/sleeve.go]: D'oh, we need to write to them of course, but return them as read only to the API consumer...

File: router/sleeve.go (hunk 4 above)
Comment (awh) [inline router/sleeve.go]: D'oh again - the intermediaries are passed as arguments to `fwd.run` later on

**Comment by awh** [comment]
Regarding the create/delete datapath & add datapath interface commands added to weaver:

> Previously, the proxy would run the weave script in ways that did not run the weaver executable.  Now they can.

We've started (for better or worse) to develop a stable of little helper executables under `prog/*` - might these functions be moved there? If they were, would it ameliorate the proxy coverage issue?

**Comment by dpw** [comment]
> We've started (for better or worse) to develop a stable of little helper executables under prog/* - might these functions be moved there? If they were, would it ameliorate the proxy coverage issue?

Where there is no functional overlap, I think having a separate helper program is natural.  That's not the case here.  And we probably do want coverage on those helper programs, too.

**Comment by awh** [comment]
I have a couple of questions regarding the short peer ID allocation code:

* Rather than the two fields in `ShortIDPeers`, could you maintain a single sorted slice holding the peers in order of ascending name? You might save some complexity in `addByShortID`/`deleteByShortID` by leveraging the `sort` package, as well as reducing the amount of test code required to check the invariants associated with the existing `peer`/`others` design. (Initially I wondered if you could use `container/heap` for this but you need to be able to remove peers by name)
* At the cost of pre-allocating 4096 `ShortIDPeers` structs and storing them in a free entries map, you could drastically simplify `chooseShortID`; even more so if you rely on Go's map iteration randomisation:

````
func (peers *Peers) chooseShortID() (PeerShortID, bool) {
    for shortID, _ := range peers.freeShortIDs {
        return shortID, true
    }
    return 0, false
}
````
add/delete then move entries between the free/in-use maps; hopefully this would make the mechanism more 'obviously correct' and obviate the need for the complex unit tests.

File: router/peer.go [hunk 7]
```diff
@@ -17,11 +17,24 @@ func ParsePeerUID(s string) (PeerUID, error) {
...
+// Short IDs exist for the sake of fast datapath.  They are 12 bits,
+// randomly assigned, but we detect and recover from collisions.  This
+// does limit us to 4096 peers, but that should be sufficient for a
+// while.
+type PeerShortID uint16
+
+const PeerShortIDBits = 12
+
+func randomPeerShortID() PeerShortID {
+	return PeerShortID(randUint64() & (1<<PeerShortIDBits - 1))
```
awh: * Later on in the FDP commit the hexadecimal literal `0xfff` is used. Perhaps introduce a `MaxPeerShortID` constant initialised to `1<<PeerShortIDBits - 1` and use it here and there?
* Rather trivial, but `randUint64` reads eight bytes from the CSPRNG and then discards > 3/4 of the bits. We could add `randUint16` to `router/utils.go`

router/overlay_switch.go
```diff
@@ -0,0 +1,393 @@
+package router
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+)
+
+// OverlaySwitch selects which overlay to use, from a set of
+// subsidiary overlays.  First, it passes a list of supported overalys
```
awh: overlays

awh:
> Things still to do

Add smoke tests for FDP?

dpw:
> Add smoke tests for FDP?

Such as?  The existing smoke tests exercise FDP.  What functional additions are there to test?

awh:
> What functional additions are there to test?

Fallback behaviour.

dpw:
> > What functional additions are there to test?
> 
> Fallback behaviour.

Already there: `test/105_frag_2_test.sh`

router/overlay_switch.go
```diff
@@ -0,0 +1,393 @@
...
+func (fwd *overlaySwitchForwarder) established(index int) {
+	fwd.lock.Lock()
+	defer fwd.lock.Unlock()
+
+	fwd.forwarders[index].established = true
+
+	// less preferred forwarders are no longer needed
+	fwd.stopFrom(index + 1)
+
+	if !fwd.alreadyEstablished {
```
awh: Do we need this flag in `overlaySwitchForwarder`? We could just set `establishedChan` to nil after we close it, and test for that instead

router/overlay_switch.go
```diff
@@ -0,0 +1,393 @@
...
+	stopChan <-chan struct{}, fwd OverlayForwarder) {
+	establishedChan := fwd.EstablishedChannel()
+loop:
+	for {
+		e := subForwarderEvent{index: index}
+
+		select {
+		case <-establishedChan:
+			e.established = true
+			establishedChan = nil
```
awh: For my own edification - I assume selecting from a `nil` channel just means this case never executes again?

router/fastdp.go
```diff
@@ -0,0 +1,1115 @@
...
+	// Might be nice to delete all the relevant flows here, but we
+	// can just let them expire.
+	fwd.fastdp.removeForwarder(fwd.remotePeer.Name, fwd)
+
+	fwd.lock.Lock()
+	defer fwd.lock.Unlock()
+	fwd.sendControlMsg = func([]byte) error { return nil }
+
+	select {
+	case fwd.errorChan <- nil:
```
awh: I can't fathom the purpose of this - we don't do it in `sleeveForwarder.Stop()` either

router/overlay_switch.go
awh: If so, we could set `establishedChan` to `nil` immediately after we close it in `fastDatapathForwarder.handleHeartbeatAck()`, and eliminate the temporary variable here

dpw: Yes, reading from a nil channel blocks.  The various corner cases for channels seem non-obvious and arbitrary to me.  But it would bee cumbersome to write code that does not depend on them, so it's just necessary to learn them.

router/overlay_switch.go
dpw: The intent of the Overlay API is that `EstablishedChannel` can be called at any point, returning the same channel.

awh:
>>> What functional additions are there to test?

>> Fallback behaviour.

> Already there: test/105_frag_2_test.sh

You're going to have to help me out - that test isn't modified in the PR, and I can't see how the current version tests fallback. Here's what I think will happen:

* Both forwarders will race to establish. They will both succeed, but because fdp is preferred it will end up as the 'best' forwarder and the overlay forwarder will be stopped (probably before it completes PMTU discovery)
* The large ping will get fragmented according to the overlay MTU, and as the VXLAN path is unaffected by the iptables rule it will succeed

What am I missing?

dpw:
> What am I missing?

The test introduces a iptables rule to discard packets > 700 bytes.  fastdp sends a heartbeat packet sized to the overlay MTU (i.e. 1410 bytes).  This won't get through, and so fastdp will never report that it is established (and ultimately it will time out and report an error).  But sleeve will report that it is established, and so be selected as the active overlay.  The test then tests sleeve fragmentation behaviour as before.

awh:
> fastdp sends a heartbeat packet sized to the overlay MTU

Argh I did check for that but I took the comment

````
 587     // the heartbeat payload consists of the 64-bit connection uid
 588     // followed by the 16-bit packet size.
````

literally and missed the `+fwd.fastdp.mtu` in the buffer allocation :smile:

Thanks!

File: router/overlay_switch.go (hunk 9 above)
Comment (awh) [inline router/overlay_switch.go]: > The intent of the Overlay API is that EstablishedChannel can be called at any point, returning the same channel.

And yet both clients of this API call (`LocalConnection.actorLoop` and `monitorForwarder`) use a temporary variable to simulate a nil-return-after-close behaviour for the value of `EstablishedChannel()`. Perhaps the original intent is no longer the most useful behaviour?

Comment (dpw) [inline router/overlay_switch.go]: You seem to be suggesting that your proposal would make things simpler on the client side, perhaps that they could just have `case <-fwd.EstablishedChannel()`?  But that's not the case.

Yes, they both have a temporary `establishedChan` variable which they set to nil.  But they both take action before setting it to nil!  If `EstablishedChannel()` can be nil, they will still need to retain a variable of some kind (essentially to track whether they have already taken the "when established" action).  But there would have to be additional code to manage that variable in the `EstablishedChannel()` returns nil case.  E.g.

    establishedChan := fwd.EstablishedChannel()
    if establishedChan == nil {
        establishedAction()
    }

    for {
        select {
        case <-establishedChan:
            establishedAction()
            establishedChan = nil

            ....


And that's the simplest variant, I think.

So what you are proposing implies additional boilerplate in two files (and a comment against `EstablishedChannel`).  Just to save a single variable?

File: router/fastdp.go [hunk 12]
```diff
@@ -0,0 +1,1115 @@
...
+
+		if err != nil && !odp.IsNoSuchFlowError(err) {
+			log.Warn(err)
+		}
+	}
+}
+
+// The router needs to know which flows are active in order to
+// maintain its MAC->peer table.  We do this by quering the router
+// without an actual packet being involed.  Maybe it's
```
awh: involved

router/fastdp.go
```diff
@@ -0,0 +1,1115 @@
...
+		}
+
+		if err != nil && !odp.IsNoSuchFlowError(err) {
+			log.Warn(err)
+		}
+	}
+}
+
+// The router needs to know which flows are active in order to
+// maintain its MAC->peer table.  We do this by quering the router
```
awh: querying

router/fastdp.go
dpw: I think it shows that I was still acclimatising to this laptop's keyboard when I wrote the bulk of fastdp.go.

dpw:
> I need to check out a kernel oops I saw when testing with ubuntu 12.04 a few days ago.

Now fixed.

router/fastdp.go
```diff
@@ -0,0 +1,1114 @@
...
+	sender  *net.UDPAddr
+}
+
+func (op vxlanSpecialPacketFlowOp) Process(frame []byte, dec *EthernetDecoder,
+	broadcast bool) {
+	op.fastdp.lock.Lock()
+	fwd := op.fastdp.forwarders[op.srcPeer.Name]
+	op.fastdp.lock.Unlock()
+
+	if fwd != nil || dec.IsSpecial() {
```
awh: '&&', not '||'.

awh:
````
INFO: 2015/09/24 16:32:27.694327 Command line options: map[dns-effective-listen-address:172.17.42.1 dns-listen-address:172.17.42.1:53 ipalloc-range:10.2.2.0/24 name:5a:13:35:7e:d4:24 nickname:host1 port:6783 datapath:weave]
INFO: 2015/09/24 16:32:27.694459 Command line peers: []
INFO: 2015/09/24 16:32:27.713558 Communication between peers is unencrypted.
INFO: 2015/09/24 16:32:27.713640 Our name is 5a:13:35:7e:d4:24(host1)
INFO: 2015/09/24 16:32:27.714831 [docker] Using Docker API on unix:///var/run/docker.sock: &[ApiVersion=1.20 GitCommit=0a8c2e3 GoVersion=go1.4.2 Os=linux Arch=amd64 KernelVersion=3.19.0-23-generic BuildTime=Thu Sep 10 19:21:21 UTC 2015 Version=1.8.2]
INFO: 2015/09/24 16:32:27.714862 Assuming quorum size of 1
INFO: 2015/09/24 16:32:27.714999 Sniffing traffic on weave (via ODP)
INFO: 2015/09/24 16:32:27.903108 Discovered local MAC 5a:13:35:7e:d4:24
INFO: 2015/09/24 16:32:28.639248 ->[192.168.48.12:37354] connection accepted
INFO: 2015/09/24 16:32:28.640530 ->[192.168.48.12:37354|b6:d7:ba:d1:81:7c(host2)]: connection ready; using protocol version 2
INFO: 2015/09/24 16:32:28.641298 overlay_switch ->[b6:d7:ba:d1:81:7c(host2)] using fastdp
INFO: 2015/09/24 16:32:28.641476 ->[192.168.48.12:37354|b6:d7:ba:d1:81:7c(host2)]: connection added (new peer)
[signal 0xb code=0x1 addr=0x0 pc=0x4b2280]

goroutine 13 [running]:
github.com/weaveworks/weave/router.(*fastDatapathForwarder).handleVxlanSpecialPacket(0x0, 0xc2092f3290, 0x590, 0xf77, 0xc2092ad770)
	/home/vagrant/src/github.com/weaveworks/weave/router/fastdp.go:606 +0x8d0
github.com/weaveworks/weave/router.vxlanSpecialPacketFlowOp.Process(0xc208070e60, 0xc20805b4a0, 0xc2092ad770, 0xc2092f3290, 0x590, 0xf77, 0xc209173cc0, 0xc208070e00)
	/home/vagrant/src/github.com/weaveworks/weave/router/fastdp.go:419 +0xf3
github.com/weaveworks/weave/router.(*vxlanSpecialPacketFlowOp).Process(0xc2092bede0, 0xc2092f3290, 0x590, 0xf77, 0xc209173cc0, 0x0)
	<autogenerated>:220 +0xd2
github.com/weaveworks/weave/router.(*FastDatapath).send(0xc208070e60, 0x7f613a630dc0, 0xc2092bee20, 0xc2092f3290, 0x590, 0xf77, 0xc2092beda0)
	/home/vagrant/src/github.com/weaveworks/weave/router/fastdp.go:1017 +0x6c0
github.com/weaveworks/weave/router.(*FastDatapath).Miss(0xc208070e60, 0xc2092f3290, 0x590, 0xf77, 0xc2092ad6e0, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/weave/router/fastdp.go:890 +0x5de
github.com/weaveworks/go-odp/odp.func·031(0xc2092bec60, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/go-odp/odp/packet.go:144 +0x228
github.com/weaveworks/go-odp/odp.func·027(0xc2092bec60, 0xc2092bec60, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/go-odp/odp/netlink.go:673 +0x7c
github.com/weaveworks/go-odp/odp.(*NetlinkSocket).Receive(0xc208aad970, 0xc2080cbf38, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/go-odp/odp/netlink.go:579 +0x11c
github.com/weaveworks/go-odp/odp.(*NetlinkSocket).consume(0xc208aad970, 0x7f613a62ec78, 0xc208070e60, 0xc2080cbf98)
	/home/vagrant/src/github.com/weaveworks/go-odp/odp/netlink.go:681 +0x51
github.com/weaveworks/go-odp/odp.DatapathHandle.consumeMisses(0xc2090dc4b0, 0x7f6100000005, 0x7f613a62e7e8, 0xc208070e60, 0xc2090dc500)
	/home/vagrant/src/github.com/weaveworks/go-odp/odp/packet.go:145 +0x96
created by github.com/weaveworks/go-odp/odp.DatapathHandle.ConsumeMisses
	/home/vagrant/src/github.com/weaveworks/go-odp/odp/packet.go:74 +0x67c
````

File: router/fastdp.go (hunk 11 above)
Comment (dpw) [inline router/fastdp.go]: Yeah, that's a relic of a past state of the code that should have been removed.

File: weave (hunk 6 above)
Comment (dpw) [inline weave]: Nice

File: router/peer.go (hunk 7 above)
Comment (dpw) [inline router/peer.go]: > * Later on in the FDP commit the hexadecimal literal 0xfff is used. Perhaps introduce a MaxPeerShortID constant initialised to 1<<PeerShortIDBits - 1 and use it here and there?

I think the two things are distinct.  The `0xfff` is derived from the division of the 24-bit vxlan VNI field into two halves, i.e. it's a statement about how we format things on the wire.  Although the value of `PeerShortIDBits` is obviously derived from that, it's not identical.

> Rather trivial, but randUint64 reads eight bytes from the CSPRNG and then discards > 3/4 of the bits. We could add randUint16 to router/utils.go

Done

File: router/fastdp.go [hunk 15]
```diff
@@ -0,0 +1,1109 @@
...
+func (fastdp *FastDatapath) getVxlanVportID(udpPort int) (odp.VportID, error) {
+	fastdp.lock.Lock()
+	defer fastdp.lock.Unlock()
+
+	if vxlanVportID, present := fastdp.vxlanVportIDs[udpPort]; present {
+		return vxlanVportID, nil
+	}
+
+	vxlanVportID, err := fastdp.dp.CreateVport(
+		odp.NewVxlanVportSpec("vxlan", uint16(udpPort)))
```
awh: It would be useful if the vport name included the UDP port (e.g. 'vxlan-6784') - without this the output of `odp flow list weave` is ambiguous in the face of mixed router ports in the same network.

awh:
@dpw I've a question about OVS behaviour. If I have three nodes in a full mesh, and I send an ethernet broadcast from one of them I end up with a flow like this (broken across lines for clarity):

````
weave
--in-port=vethwepl13226
--eth-src=4a:81:be:6b:45:2d --eth-dst=ff:ff:ff:ff:ff:ff
--set-tunnel-id=0000000000ca2614 --set-tunnel-ipv4-src=192.168.48.12 --set-tunnel-ipv4-dst=192.168.48.11 --set-tunnel-ttl=64 --set-tunnel-df=true
--set-tunnel-id=0000000000217614 --set-tunnel-ipv4-src=192.168.48.12 --set-tunnel-ipv4-dst=192.168.48.13 --set-tunnel-ttl=64 --set-tunnel-df=true
--output=vxlan-6784,vxlan-6784,weave
````

Here there are two set-tunnel actions, one for each peer needing to receive the broadcast, and the vxlan vport appears twice in the output action. How is it that the first output to the vxlan vport is associated with the first set of tunnel parameters, and the second output with the second set? A naive interpretation suggests that the first set of tunnel parameters are overwritten by the second set before we reach the output action, and we then output the packet twice to the same peer...

**Comment by awh** [comment]
>  How is it that the first output to the vxlan vport is associated with the first set of tunnel parameters, and the second output with the second set?

Ahh, it would seem this is an issue with `odp flow list weave` collecting output actions and displaying them out of sequence at the end of the action list...

**Comment by dpw** [comment]
> Ahh, it would seem this is an issue with odp flow list weave collecting output actions and displaying them out of sequence at the end of the action list...

Yes, exactly.

`--output` actions were represented that way due to the limitations of the golang `flags` module.  But it is silly, and other actions are not aggregated like that, which means that you couldn't use the tool to set them.

It's a mess, bu-t the tool was meant for experimenting with OVS, and it's sometimes useful for debugging.  But I wouldn't lean on it too heavily.

If you want to know what the flows created by weave look like, run with `--log-level=debug`. The resulting output is ugly, but (intended to be) accurate.

**Comment by awh** [comment]
Assume a peer `A` listening on our standard vxlan port is instructed to connect to another peer `B` running on a non-standard vxlan port:

````
A (6784) ----------------------> B (9001)
````

In this case, due to the `if remoteAddr != nil` conditional in `MakeForwarder`, the originating peer `A` creates a new vxlan vport for port `9001` primarily so that it can *send* vxlan encapsulated packets to `B`.

On the other hand `B`, not taking the conditional, selects its own main vxlan vport on `9001` for forwarding to `A` - this only works because a side effect of `A` creating a vxlan vport for `9001` is *that it also begins listening for vxlan traffic on this port.*

Have I understood this correctly? If so, I have some further thoughts:

* The behaviour is quite subtle and could probably stand a comment in the code as to why the receiving peer doesn't need to allocate a vxlan vport when it eventually 'learns' the remote address
* As it stands, `B` never actually learns that `A`s main vxlan port is `6784` because the remote port passed to `handleVxlanSpecialPacket` is actually the UDP port used to create `B`s own main vxlan vport in `getVxlanVportID`
* The behaviour is counter-intuitive from a users perspective - they would expect vxlan traffic from `B->A` to be destined for port `6784`. They may even have firewall rules that enforce this expectation, preventing `B` from forwarding to `A`

**Comment by dpw** [comment]
> The behaviour is quite subtle and could probably stand a comment in the code as to why the receiving peer doesn't need to allocate a vxlan vport when it eventually 'learns' the remote address

Ok.

> As it stands, B never actually learns that As main vxlan port is 6784 because the remote port passed to handleVxlanSpecialPacket is actually the UDP port used to create Bs own main vxlan vport in getVxlanVportID

Yes. There is no way for it to learn that. The source port on the vxlan packets is not 6784.

> The behaviour is counter-intuitive from a users perspective - they would expect vxlan traffic from B->A to be destined for port 6784. They may even have firewall rules that enforce this expectation, preventing B from forwarding to A

Then we'll fall back to sleeve.

File: router/fastdp.go [hunk 16]
```diff
@@ -0,0 +1,1109 @@
...
+		switch fop := xfop.(type) {
+		case interface {
+			updateFlowSpec(*odp.FlowSpec)
+		}:
+			fop.updateFlowSpec(&flow)
+		case vetoFlowCreationFlowOp:
+			createFlow = false
+		default:
+			// A foreign flow op, so send the packet the
+			// normal way, decoding the packet lazily.
```
awh: This would be a lot clearer if we were explicit that we're talking about an op from the sleeve forwarder that we have received from the router

router/fastdp.go
```diff
@@ -0,0 +1,1109 @@
...
+			if dec == nil {
+				dec = fastdp.takeDecoder(lock)
+				dec.DecodeLayers(frame)
+				createFlow = false
```
awh: It's obviously really important that we don't create flows if we have accumulated any sleeve forwarder ops - I think this deserves an explanatory comment

awh: This also has a potentially non-obvious performance impact on broadcast traffic from peers which have more than zero sleeve forwarders - do we need to mention this in user facing documentation?

awh:
> If you want to know what the flows created by weave look like, run with --log-level=debug. The resulting output is ugly, but (intended to be) accurate.

I'd really like to see some additions to `weave status`:

`weave status connections` to show which forwarder is in use
`weave status odp` display current state of flows

it would be nice to get these into 1.2 for support purposes, but I don't want to hold up this PR for that - since the review is drawing to a close, how about I create an issue for it and work on it myself whilst you're tidying up the last bits and pieces? @dpw @rade

rade:
> `weave status connections` to show which forwarder is in use

That seems useful.

> `weave status odp` display current state of flows

Perhaps this should just go into `weave report`.

awh:
BTW my WIP 'how it works' doc is [here](https://github.com/awh/weave/blob/fdp-docs/docs/fdp-overview.markdown).

awh:
> That seems useful.

#1505.

router/fastdp.go
dpw: That's not necessarily true (in certain configurations it could be a pcap delivery FlowOp).  And the point of the FlowOp abstraction is you shouldn't care. So the most that seems appropriate is a parenthetical "e.g. a sleeve forwarding FlowOp".

router/fastdp.go
dpw: I think what we should say to users about performance is a conservative set of conditions under which the weave network is supposed to use fastdp (i.e. no NAT or MTU-challenged links between peers, and nothing blocking the relevant ports).  Anything else may cause sleeve to be used  in some cases, leading to a performance hit.  I don't think it would be wise to elaborate further.

router/fastdp.go
awh: I understood that you were being deliberately abstract, nevertheless a concrete example would definitely have helped my understanding so I think the paranthetical is exactly what is needed. Does it make sense to go on with 'send the packet in the normal way'? It's not clear what that might be IMO.

dpw: "send the packet through the FlowOp interface"?

awh: > in certain configurations it could be a pcap delivery FlowOp

Would you mind elaborating on this?

awh: > in certain configurations it could be a pcap delivery FlowOp

For this to be true:

* `-iface` must have been specified in conjunction wtih `-datapath` as per comments in `prog/weaver/main.go`, so that the `InjectPacket` implementation used by `Router.handleForwardedPacket` is the one supplied by `Pcap` (thus yielding a pcap delivery `FlowOp`)
* Encryption must be *disabled*:
    * `FastDatapath.send()` is only invoked from the ODP miss handler
    * Only the vxlan vport miss handler can obtain a pcap delivery `FlowOp` (by invoking `Router.handleForwardedPacket`)
    * The vxlan vport miss handler will only be called if the fast datapath overlay is inluded in the overlay switch, which is not the case if encryption is enabled

Why would you configure things this way?

dpw: > Why would you configure things this way?

Well, at a certain point in fastdp development, I had the code to manage the vxlan tunnel flows, but I hadn't written the full bridge implementation.  But I had something that worked, because I could run with the nascent fastdp Overlay and the pcap-based Bridge (on top of a normal linux bridge netdev).

Even now, it only takes a simple modification to the weave script to run with the fastdp Overlay and pcap-based Bridge (although with the bridge actually being provided by fastdp too).  And it works, demonstrating that the design is general enough to support this with no substantive code changes.  A proper understanding of the fastdp code should not exclude such possibilities.

awh: > "send the packet through the FlowOp interface"?

:+1:

router/local_peer.go
```diff
@@ -243,14 +249,20 @@ func (peer *LocalPeer) handleDeleteConnection(conn Connection) {
...
 func (peer *LocalPeer) broadcastPeerUpdate(peers ...*Peer) {
-	peer.router.Routes.Recalculate()
-	peer.router.TopologyGossip.GossipBroadcast(NewTopologyGossipData(peer.router.Peers, append(peers, peer.Peer)...))
+	// Some tests run without a router.  This should be fixed so
+	// that the relevant part of Router can be easily run in the
+	// context of a test, but that will involve significant
+	// reworking of tests.
+	if peer.router != nil {
+		peer.router.BroadcastTopologyUpdate(append(peers, peer.Peer))
+	}
```
rade: Why did you move `peer.router.Routes.Recalculate()` out of this method and into the call sites? I would have thought that if its moving at all, the logical place for it to reside is the new `router.BroadcastTopologyUpdate(...)` method.

dpw: Because there's a new call site where we don't want to do `peer.router.Routes.Recalculate()`.

awh:
For the avoidance of doubt, other than giving the remaining loose ends the once over I am ready to merge this.

dpw:
> I have a couple of thoughts regarding the short peer ID allocation code:
> 
>     Rather than the two fields in ShortIDPeers, could you maintain a single sorted slice holding the peers in order of ascending name? You might save some complexity in addByShortID/deleteByShortID by leveraging the sort package, as well as reducing the amount of test code required to check the invariants associated with the existing peer/others design. (Initially I wondered if you could use container/heap for this but you need to be able to remove peers by name)

It's true that it would be *slightly* simpler (the simplifications would mainly occur in deleteByShortID).  But I don't think it would make a huge amount of code go away.  We would still need to detect cases where `invalidateShortIDs` must be set.  Similarly for the tests: Ok, they have an added importance in the context of the code as it is. But even with your proposal, I don't think the amount of testing code would be significantly reduced, due to the need to exercise corner cases that would still exist.

Set against that, it seems an over-specified solution to me, and would be less efficient (not in any way that is ever likely to matter, I admit).  If golang was a nice high-level language, I would worry more about that code not being succinct.  Golang is not a nice high-level language.

>     At the cost of pre-allocating 4096 ShortIDPeers structs and storing them in a free entries map, you could drastically simplify chooseShortID; even more so if you rely on Go's map iteration randomisation:

I consider relying on golang's randomisation of the iteration start point to be an anti-pattern on a par with relying on map iteration to be stable in a language that does not guarantee that.

router/fastdp.go
dpw: Done

dpw:
I think I've addressed everything that hasn't otherwise been responded to. There's a squashed-and-rebased branch at https://github.com/weaveworks/weave/tree/fdp-rebase , though I'll push that as fdp for this PR when everyone is happy.

awh:
* There was a comment about defining `const PeerShortIDMask = 1<<PeerShortIDBits - 1` in `peer.go` and using that in `randomPeerShortID`/`extractPeers`. Also `extractPeers` and `tunnelIDFor` don't make use of `PeertShortIDBits` as they should
* We were going to have a paranthetical `(e.g. sleeve forwarder) in `send()`

dpw:
> There was a comment about defining const PeerShortIDMask = 1<<PeerShortIDBits - 1 in peer.go and using that in randomPeerShortID/extractPeers. Also extractPeers and tunnelIDFor don't make use of PeertShortIDBits as they should

Now you mention it, I do recall that discussion.  But I don't see it anywhere on this PR, which seems odd (and I have been careful not to do any rebasing on the fdp branch during this review).  I dug out the relevant github notification email, and the link in it no longer works.  That is rather disappointing.

I also recall that I responded to that suggestion, though maybe you weren't convinced by my response.  Do you have the email for that response?  My point was that the values in extractPeers/tunnelIDFor represent a definition of the encapsulation format, and are not necessarily the same as PeerShortIDBits.  That would leave only a single potential use of PeerShortIDMask.

> We were going to have a paranthetical '(e.g. sleeve forwarder)' in send()

I did overlook this one when going back through all the comments.  Fixed now.

awh:
Found it: https://github.com/weaveworks/weave/pull/1438#discussion_r40575163 - I didn't notice it before, sorry! Accepting that the literal `12`s in `fastdp.go` are only coincidentally equal in value to `PeerShortIDBits`, I might ask why they do not merit their own separate constant from which `0xfff` is derived, but now that I look we seem to have lots of magic literal values sprinkled elsewhere so I suppose it doesn't make a lot of difference :smile:

dpw:
> Found it: #1438 (comment) - I didn't notice it before, sorry!

Thanks. So github has it associated with an "outdated diff", which is why I couldn't find it. But why does github think it is outdated?  There's been no rebasing or other history rewriting.  So I remain puzzled. 

>Accepting that the literal 12s in fastdp.go are only coincidentally equal in value to PeerShortIDBits, I might ask why they do not merit their own separate constant from which 0xfff is derived, but now that I look we seem to have lots of magic literal values sprinkled elsewhere so I suppose it doesn't make a lot of difference :smile:

Yes.  I believe symbolic constants have their place, but many people use them in contexts where I consider it unnecessary.

dpw:
> But why does github think it is outdated? There's been no rebasing or other history rewriting.

Oh, it regards any region touched by a later commit on the branch to be outdated.  I wish there was an "unhide all" button.

Ok, I've found another unresolved item I missed.  I'll have to do another sweep...

dpw:
> Ok, I've found another unresolved item I missed. I'll have to do another sweep...

No, I had addressed it.  Right, I think that's it.

awh:
There's a comment in the `weave` script that mentions a `--no-fastdp` option.

awh:
I think we are there - go ahead and push `fdp-rebase`->`fdp`. What did you want to do about the documentation and `weave status` PRs that I have open against this branch? That was just for receiving comments more than anything else, they can easily be reopened against `master` once this PR is merged if you'd prefer.

dpw:
> I think we are there - go ahead and push fdp-rebase->fdp.

Done

> What did you want to do about the documentation and weave status PRs that I have open against this branch? That was just for receiving comments more than anything else, they can easily be reopened against master once this PR is merged if you'd prefer.

Yes, I think they will have to be rebased against master (or the new fdp branch).

I'm happy to look at the weave status one if @rade doesn't want to (I looked at part of it already).  It would be good to get it merged today.

I will look at the documentation one, sorry I haven't given it any attention so far.  But that doesn't seem like a blocker for the 1.2 release?

rade:
> I will look at the documentation one, sorry I haven't given it any attention so far. But that doesn't seem like a blocker for the 1.2 release?

What piece of documentation do you suggest we link to in the release note item? Also, I would appreciate suggestions on how to describe the FDP feature in the release notes.

awh:
> I consider relying on golang's randomisation of the iteration start point to be an anti-pattern on a par with relying on map iteration to be stable in a language that does not guarantee that.

I guess you're not a fan of `Routes.RandomNeighbours()` and `GossipChannel.gcSenders/relayBroadcast()` then :smile:

rade:
why is the reported coverage so poor?

awh:
> What piece of documentation do you suggest we link to in the release note item?

Perhaps now that we have two overlays, there should be a section in features.md that discusses the encryption/speed tradeoff, port requirements and `WEAVE_NO_FASTDP`?

rade:
> why is the reported coverage so poor?

Most of that seems to be due to `go-odp`. Why is that package showing up in the weave coverage reports?

dpw:
> What piece of documentation do you suggest we link to in the release note item?

That's a good question, but I don't think Adam's doc can be the answer to that.

> Also, I would appreciate suggestions on how to describe the FDP feature in the release notes.

Ok.  Do you have an opinion on how much should we say in the release notes, and how much in the docs?  I would tend towards saying anything substantial in the docs, and then the release notes can just mention the presence of the feature, and say that it is possible to do a rolling upgrade from 1.1 to 1.2 without special measures.

dpw:
> Most of that seems to be due to go-odp. Why is that package showing up in the weave coverage reports?

Shrug.  I don't know what would cause it to be treated differently from any other dependency.

awh:
>  I don't know what would cause it to be treated differently from any other dependency.

Perhaps it's because they're both under the weaveworks github org?

rade:
> Perhaps it's because they're both under the weaveworks github org?

https://github.com/weaveworks/weave/blob/master/Makefile#L57

Looks like a bug; perhaps it should be `weaveworks/weave`. @tomwilkie

awh:
> I'm happy to look at the weave status one if @rade doesn't want to (I looked at part of it already). It would be good to get it merged today.

I've rebased this against the new fdp branch.

tomwilkie:
> Looks like a bug; perhaps it should be weaveworks/weave. @tomwilkie
<https://github.com/tomwilkie>

Yup, that could be considered a bug.  Making it weave/works should fix it.

awh:
Only two things are preventing me from pressing merge:

* I don't know if we're planning to include updated users docs in this PR
* I don't know if `weave status` is getting merged into this branch before it gets merged into master

Thoughts @dpw @rade?

rade:
I've (hopefully) fixed the coverage reporting on master. Somebody please do a re-base.

I am happy to defer user docs and `weave status` to separate issues.

dpw:
> Somebody please do a re-base.

If I rebase this, then @awh will have to rebase #1513, and I'll have to rebase #1522.  If the only impact on the coverage report is to exclude some go-odp files, is a rebase going to tell us anything new?

rade:
> is a rebase going to tell us anything new?

a) whether my fix actually worked
b) whether the decrease in coverage really is (nearly)accounted for by go-odp

dpw:
> > is a rebase going to tell us anything new?
> 
> a) whether my fix actually worked
> b) whether the decrease in coverage really is (nearly)accounted for by go-odp

You could have answered both questions directly by branching your fix off this branch.

Rebased.

awh:
> Rebased.

Samewise.

awh:
> Coverage decreased (-12.06%) to 61.737%

That hasn't gone well :smile:

rade:
There seem to be some quite big coverage holes, e.g.

- {fastdp,overlap_switch,sleeve} InvalidateShortIDs
- fastdp expireMACs,expireFlows,touchFlow

awh:
> There seem to be some quite big coverage holes

That's not the problem - in fact, only `overlay.go` has reduced in coverage due to the addition of eight untested lines. The report is still counting all the lines of the sources files in `odp`, except now it is reporting zero test coverage for them all...

dpw:
> There seem to be some quite big coverage holes, e.g.
> 
>  *   {fastdp,overlap_switch,sleeve} InvalidateShortIDs

Because none of the smoke tests have nearly enough peers to cause short ID collisions.

>  *   fastdp expireMACs,expireFlows,touchFlow

Because we don't run the smoke tests for long enough for flows to be expired (in the same way we never see MAC cache expiries).

At 74% and 78%, the coverage of the two big new files (fastdp.go and overlay_switch.go) is about average.

rade:
Can you think of an easy way to test InvalidateShortIDs?

dpw:
> Can you think of an easy way to test InvalidateShortIDs?

No, nothing that isn't a case of Volkswagening.

dpw:
> Can you think of an easy way to test InvalidateShortIDs?
> 
> No, nothing that isn't a case of Volkswagening.

And note that:

* InvalidateShortIDs is just logging and locking around a call to deleteFlows, which is covered.

* The unit tests verify that InvalidateShortIDs notifications occur when appropriate.

So it seems reasonable to check the code in InvalidateShortIDs by inspection.

rade:
[coverage sorted](https://circleci.com/gh/weaveworks/weave/3211#artifacts).

awh:
@dpw @rade I don't think the current short ID discussion in #1522 has any bearing on this PR - any final comments before I press merge?

rade:
merge it

PR #1445, opened 2015-09-16
Document ECS integration

2opremio:
The intention is to use this as the single authoritative source of ECS AMIs, which right now are scattered around making them difficult to maintain.

@abuehrle Please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ Ensure you are running Linux (kernel 3.8 or later) and have Docker
 
 CoreOS users see [here](https://github.com/fintanr/weave-gs/blob/master/coreos-simple/user-data) for an example of installing weave using cloud-config.
 
+Amazon ECS users see
+[here](https://github.com/weaveworks/guides/blob/master/aws-ecs/LATESTAMIs.md)
+for the latest Weave AMIs and
+[here](http://weave.works/guides/service-discovery-with-weave-aws-ecs.html) to get started with Weave on ECS.
+
 Weave respects the environment variable `DOCKER_HOST`, so you can run
 it locally to control a weave network on a remote host.
 


abuehrle:
Minor edits: not sure if 'out of the box' is totally correct, since software no longer ships in a box :)  but I like it, as it's kind of retro. 
-----
So that Weave Net, Weave Run and Weave Scope can be used 'out of the box', we provide a set of AMIs that are fully compatible with ECS-Optimized Amazon Linux AMIs.

The following are the latest supported Weave AMIs for each region:

2opremio:
Thanks @abuehrle . I included your edits and a final line linking to the ECS guide.

rade:
Is 'Features' really the best place for this? Would have thought that https://github.com/weaveworks/weave#installation is a better place.

2opremio:
@rade The thing is that the AMIs also include Scope so I am not sure if it makes sense to put them in the weave-specific installation instructions. Although, arguably, the weave-specific features page doesn't make sense either.

How about leaving it in the features section and linking that from the Scope/Weave Installation section of the READMEs?

squaremo:
`curl -D - http://docs.weave.works/weave/master/features.html` -> 404

2opremio:
@squaremo I will remove the link in the README and create a separate pull request with the correct link to http://docs.weave.works/weave/master/commit_id/features.html once we merge this.

This is because  http://docs.weave.works/weave/latest_release/features.html cannot be used (this docs haven't been released and we cannot use and we cannot know the commit_id before it gets merged.

squaremo:
> This is because http://docs.weave.works/weave/latest_release/features.html cannot be used since these docs haven't been released and we cannot know the master commit_id before it gets merged.

That indicates to me that it's the wrong place to put that information (surely it will get updated on a schedule independent of weave releases?)

2opremio:
Yep, I think you are right @squaremo It should probably live with the other ECS stuffin guides/ (which at some point should have its own repository). I will add a markdown file there and modify this PR to link to it.

2opremio:
@squaremo I've updated the link in the README which the only remaining change.

2opremio:
Now https://github.com/weaveworks/guides/blob/master/aws-ecs/LATESTAMIs.md is the single source of truth for the latest ECS AMIs

PR #1446, opened 2015-09-18
Add smoke-test for subdomains

bboreham:
Since there wasn't one.

Using ubuntu rather than alpine to get DNS search-paths to work.

### DIFF
--- a/test/225_dns_subdomain_test.sh
+++ b/test/225_dns_subdomain_test.sh
@@ -1,0 +1,25 @@
+#! /bin/bash
+
+. ./config.sh
+
+C1=10.2.54.34
+C2=10.2.54.91
+C3=10.2.54.92
+DOMAIN=foo.bar
+SUBDOMAIN1=sub1.$DOMAIN
+SUBDOMAIN2=sub2.$DOMAIN
+
+start_suite "Resolve names in custom domain with subdomains"
+
+weave_on $HOST1 launch --log-level=debug --dns-domain $DOMAIN.
+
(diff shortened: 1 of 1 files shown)


rade:
We really don't want to add ubuntu images to our tests. So I'm afraid we'll probably have to live w/o a test for this.

(as an aside, any images we *do* add must be added to `TEST_IMAGES` in config.sh)

test/225_dns_subdomain_test.sh
```diff
@@ -0,0 +1,25 @@
...
+SUBDOMAIN1=sub1.$DOMAIN
+SUBDOMAIN2=sub2.$DOMAIN
+
+start_suite "Resolve names in custom domain with subdomains"
+
+weave_on $HOST1 launch --log-level=debug --dns-domain $DOMAIN.
+
+weave_on $HOST1 run --with-dns $C1/24 --name=c1 -h foo.$SUBDOMAIN1 -t $NONBROKENRESOLVER_IMAGE /bin/sh
+weave_on $HOST1 run --with-dns $C2/24 --name=c2 -h foo.$SUBDOMAIN2 -t $NONBROKENRESOLVER_IMAGE /bin/sh
+weave_on $HOST1 run --with-dns $C3/24 --name=c3 -h tre.$SUBDOMAIN1 --dns-search=$SUBDOMAIN2 -t $NONBROKENRESOLVER_IMAGE /bin/sh
```
rade: why not `start_container_with_dns_on` for all the above? That launches containers based on `aanand/docker-dnsutil`, which in turn is based on ubuntu.

bboreham:
Re-done using `start_container_with_dns` as suggested by @rade

PR #1449, opened 2015-09-18
Override sneaky /start HostConfig params

### LINKED ISSUE #1368: container start HostConfig breaks container configuration
As noted in #1258, the Docker remote API allows the HostConfig attributes to be set when _starting_ a container. (They are supposed to be set when _creating_)

The specific case there (Kubernetes) uses `--net=container:(another container)`, so is fixed by #1314, but if someone did the same thing and specified `--net=host` in the start request then `weavewait` would hang because there is no `ethwe` interface.

paulbellamy:
Fixes #1368


test/666_abuse_of_start_test.sh
```diff
@@ -0,0 +1,21 @@
...
+. ./config.sh
+
+start_suite "Abuse of 'start' operation"
+
+weave_on $HOST1 launch
+
+proxy_start_container $HOST1 --name=c1
+proxy docker_on $HOST1 create --name=c2 $SMALL_IMAGE grep "^1$" /sys/class/net/ethwe/carrier
+# Now start c2 with a sneaky HostConfig
+curl -X POST -H Content-Type:application/json -d '{"NetworkMode": "container:c1"}' http://$HOST1:12375/v1.20/containers/c2/start
```
rade: I suggest introducing a `docker_remote_on` in config.sh, which does the `curl`ing and can be invoked with `proxy docker_remote_on`.

test/666_abuse_of_start_test.sh
```diff
@@ -0,0 +1,21 @@
...
+docker_bridge_ip=$(weave_on $HOST1 docker-bridge-ip)
```
rade: This should come right after `weave_on` since it is used by both tests.

test/666_abuse_of_start_test.sh
```diff
@@ -0,0 +1,21 @@
...
+assert "docker_on $HOST1 inspect -f '{{.State.Running}} {{.State.ExitCode}} {{.HostConfig.Dns}}' c2" "false 0 [$docker_bridge_ip]"
+
+# Now start c3 with a mostly null HostConfig
+proxy docker_on $HOST1 create --name=c3 -v /tmp:/hosttmp $SMALL_IMAGE grep "^1$" /sys/class/net/ethwe/carrier
+curl -X POST -H Content-Type:application/json -d '{"Binds":[],"ContainerIDFile":"","LxcConf":null,"Memory":0,"MemorySwap":0,"CpuShares":0,"CpuPeriod":0,"CpusetCpus":"","CpusetMems":"","CpuQuota":0,"BlkioWeight":0,"OomKillDisable":false,"Privileged":false,"PortBindings":null,"Links":[],"PublishAllPorts":false,"Dns":null,"DnsSearch":null,"ExtraHosts":null,"VolumesFrom":null,"Devices":null,"NetworkMode":"","IpcMode":"","PidMode":"","UTSMode":"","CapAdd":null,"CapDrop":null,"RestartPolicy":{"Name":"","MaximumRetryCount":0},"SecurityOpt":null,"ReadonlyRootfs":false,"Ulimits":null,"LogConfig":{"type":"","config":null},"CgroupParent":""}' http://$HOST1:12375/v1.20/containers/c3/start
```
rade: We should trim this to only include necessary fields and ones we care about.

proxy/common.go
```diff
@@ -50,6 +50,10 @@ func unmarshalRequestBody(r *http.Request, target interface{}) error {
 	}
 	r.Body = ioutil.NopCloser(bytes.NewReader(body))
 
+	if len(body) == 0 {
+		return nil
+	}
+
```
rade: why do we suddenly need this?

proxy/common.go
```diff
@@ -62,6 +66,7 @@ func marshalRequestBody(r *http.Request, body interface{}) error {
 	}
 	r.Body = ioutil.NopCloser(bytes.NewReader(newBody))
 	r.ContentLength = int64(len(newBody))
+	r.Header.Set("Content-Type", "application/json")
```
rade: and this?

proxy/proxy.go
```diff
@@ -365,3 +368,54 @@ func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *dock
...
+	}
+	defer resp.Body.Close()
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return
+	}
+
+	return string(b), true
+}
```
rade: Surely this function needs to pay attention to `proxy.With*out*DNS` too. Also, `running` is misleading as a return value name, since it will be set to true if `proxy.WithDNS==true` even when weave isn't running. And perhaps we should ditch that bool return value altogether and instead return "" to indicate that dns settings should be left alone.

proxy/start_container_interceptor.go
```diff
@@ -2,16 +2,41 @@ package proxy
...
-	return err
+
+	hostConfig := &docker.HostConfig{}
+	if err := unmarshalRequestBody(r, &hostConfig); err != nil {
+		return err
+	}
+
+	i.proxy.addWeaveWaitVolume(hostConfig)
+	if dnsDomain, withDNS := i.proxy.getDNSDomain(); withDNS {
+		if err := i.proxy.setWeaveDNS(hostConfig, container.Config.Hostname, dnsDomain); err != nil {
```
rade: should we perhaps use `container.Config.Domainnname` here instead of `dnsDomain`? Or at least check that it is the sname as `dnsDomain`?

rade:
This should be branched off 1.1, not master.

proxy/start_container_interceptor.go
paulbellamy: I don't think so. `dnsDomain` is the (possibly) untrimmed version `weave.local.`, while the `container.Config.Domainname` gets the trailing dot trimmed off. We only set it if the user has not specified `--name`, `--hostname`, or `--dns-search`.

proxy/proxy.go
paulbellamy: re `WithoutDNS`, yep.

I've removed the `running` bool.

test/666_abuse_of_start_test.sh
```diff
@@ -0,0 +1,21 @@
...
+proxy docker_on $HOST1 create --name=c2 $SMALL_IMAGE $CHECK_ETHWE_UP
+proxy docker_api_on $HOST1 POST /v1.20/containers/c2/start '{"NetworkMode": "container:c1"}'
+assert "docker_on $HOST1 inspect -f '{{.State.Running}} {{.State.ExitCode}} {{.HostConfig.Dns}}' c2" "false 0 [$docker_bridge_ip]"
+
+# Start c3 with HostConfig having empty binds and null dns/networking settings
+proxy docker_on $HOST1 create --name=c3 -v /tmp:/hosttmp $SMALL_IMAGE $CHECK_ETHWE_UP
+proxy docker_api_on $HOST1 POST /v1.20/containers/c3/start '{"Binds":[],"Dns":null,"DnsSearch":null,"ExtraHosts":null,"VolumesFrom":null,"Devices":null,"NetworkMode":""}'
+assert "docker_on $HOST1 inspect -f '{{.State.Running}} {{.State.ExitCode}} {{.HostConfig.Dns}}' c3" "false 0 [$docker_bridge_ip]"
+
+end_suite
```
rade: doesn't the `v20` mean we need docker 1.8.0+ for this to work? I'd rather avoid such a dependency.

proxy/proxy.go
```diff
@@ -365,3 +368,57 @@ func (proxy *Proxy) weaveCIDRsFromConfig(config *docker.Config, hostConfig *dock
...
+	}
+	defer resp.Body.Close()
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return
+	}
+
+	return string(b)
+}
```
rade: surely this should return "" when we cannot obtain the domain and `!WithDNS`.

rade: I reckon guarding the `domain` initialisation with
```
if proxy.WithDNS {
    domain = nameserver.DefaultDomain
}
```
would do the trick.

proxy/start_container_interceptor.go
```diff
@@ -2,16 +2,45 @@ package proxy
...
+
 func (i *startContainerInterceptor) InterceptRequest(r *http.Request) error {
 	container, err := inspectContainerInPath(i.proxy.client, r.URL.Path)
-	if err == nil && containerShouldAttach(container) {
+	if err != nil {
+		return err
+	}
+
+	if !containerShouldAttach(container) ||
+		r.Header.Get("Content-Type") != "application/json" {
```
rade: cosmetic: put on previous line

Issue #1396, opened 2015-09-08
fix sleeve wire compatibility with 1.1


dpw:
#991 broke wire compatibility with 1.1.  Fixing that in itself will be fairly straightforward.  But fast datapath will introduce much trickier compatibility issues, so we probably want to fix them all at once.

Details: In 1.1, `LocalConnection handleProtocolMessage` dispatches some UDP-forwarding message types directly.  Post #991, these are combined into a `ProtocolOverlayControlMsg` message type, which gets handed to the forwarder to deal with them.  To restore compatibility, we'll have to reintroduce those legacy message types, and pass the message code down to the forwarder so `sleeveForwarder` can dispatch on it.  And similarly for the control message send side.

dpw:
Some notes about the 1.1 -> 1.2 -> 1.3 compatibility path I envisage, while I have the details in mind.

In 1.1, the sleeve messages on the TCP connection look like:

|Tag|Payload|
|---|-------|
|ConnectionEstablished, FragmentationReceived or PMTUVerified|...|

Post-#991, there is also the hypothetical possibility of other Overlays, which get to define their own message type codes within the payload under a single OverlayControlMsg tag:

|Tag|Payload|
|---|-------|
|OverlayControlMsg|...|

The target state is to use the post-#991/pre-#1451 approach.  But, OverlaySwitch (in #1438) will be used in order to multiplex multiple Overlays.  OverlaySwitch puts a "overlay index" byte at the start of the message, so:

|Tag|Payload|
|---|-------|
|OverlayControlMsg|overlay index, ...|

And in terms of the code in `overlay.go` and `sleeve.go`, the target state is to return things to more or less as they were before #1451.  So in the case of sleeve, the control messages begin with a message type byte.  (fastdp also has a message type byte, even though it only has one message type.)

|Tag|Payload|
|---|-------|
|OverlayControlMsg|overlay index, message type, ...|

In 1.2, `sleeve.go` will need to operate in two modes, as shown in the first and last illustrations above. Which mode is in operation will depend on whether OverlaySwitch is intervening, which is determined by the presence of the `Overlays` feature during connection initiation (no `Overlays` feature indicates a 1.1 router). As a transitional stage, OverlaySwitch will implement the message type byte, passing it to/from the subsidiary Overlay as the `tag` parameter.

[...]

PR #1451, opened 2015-09-21
Restore sleeve compatibility with 1.1

dpw:
Addresses #1396

### DIFF
--- a/router/protocol.go
+++ b/router/protocol.go
@@ -329,10 +329,13 @@ type ProtocolTag byte
 
 const (
 	ProtocolHeartbeat ProtocolTag = iota
-	ProtocolOverlayControlMsg
+	ProtocolConnectionEstablished
+	ProtocolFragmentationReceived
+	ProtocolPMTUVerified
 	ProtocolGossip
 	ProtocolGossipUnicast
 	ProtocolGossipBroadcast
+	ProtocolOverlayControlMsg
 )
 
 type ProtocolMsg struct {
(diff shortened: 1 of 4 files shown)


router/protocol.go
```diff
@@ -329,10 +329,13 @@ type ProtocolTag byte
 
 const (
 	ProtocolHeartbeat ProtocolTag = iota
-	ProtocolOverlayControlMsg
+	ProtocolConnectionEstablished
+	ProtocolFragmentationReceived
+	ProtocolPMTUVerified
```
awh: When we undo this for 1.3, do we not need to leave these three entries intact to avoid renumbering of the subsequent constants? If so perhaps the changes to this file should be pulled into a separate initial commit so we can just `git revert` the remainder...

awh:
The smoketests pass with one host running v1.1.0 and another running this branch. Am ready to merge this once the comment above has been considered.

router/protocol.go
dpw: That's right, the sleeve message type numbering is likely to retain these values forever .  Though what I had in mind for 1.3 was to move the definitions back to sleeve.go, with their old naming but the values set here.

As for structuring things for a 'git revert': One of the benefits of keeping these compatibility PRs distinct is that it will facilitate the task of dropping 1.1 compatibility in 1.3.  But I don't expect it to be quite as a simple as a 'git revert'.  Even if we did the work now to allow that, other changes might be made between now and 1.3 that would present obstacles. So I think we can cross that bridge when we come to it.

PR #1456, opened 2015-09-22
[proxy] fix race condition when waiting for containers to start

paulbellamy:
If ContainerStarted fired before the response handler, then the "waiter"
chan would be nil, and the response handler would hang.


paulbellamy:
Introduced in https://github.com/weaveworks/weave/pull/1210

PR #1460, opened 2015-09-22
Simplify ipam_cidrs function in weave script

bboreham:
Instead of duplicating most of the code, just inject an arg that has the same effect.

We lose a more specific error message when IPAM is disabled, but this seems like a niche interest.

Also, replace the `while [ $# -gt 0 ] ... shift` loop with a `for`.

### DIFF
--- a/weave
+++ b/weave
@@ -909,24 +909,14 @@ ipam_cidrs() {
     shift 1
     ALL_CIDRS=""
     IPAM_CIDRs=""
-    if [ $# -eq 0 ] ; then
-        # If no addresses passed in, get one in the default subnet
-        IPAM_CIDRS=$(call_weave POST /ip/$CONTAINER_ID$CHECK_ALIVE) || return 1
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -909,24 +909,14 @@ ipam_cidrs() {
...
-            echo "$IPAM_CIDRS" >&2
-            return 1
-        fi
-        ALL_CIDRS="$IPAM_CIDRS"
-    fi
-    while [ $# -gt 0 ] ; do
-        if [ "${1%:*}" = "net" ] ; then
-            if [ "$1" = "net:default" ] ; then
+    # If no addresses passed in, get one in the default subnet
+    [ $# -eq 0 ] && set -- net:default
```
rade: I tend to prefer writing these things s.t. the whole expression always succeeds, i.e. `[ $# -gt 0 ] || set -- net:default`.

PR #1462, opened 2015-09-23
DNS lookups should not be case sensitive.

### LINKED ISSUE #1461: DNS shouldn't be case sensitive
"Note that while upper and lower case letters are allowed in domain
names, no significance is attached to the case.  That is, two names with
the same spelling but different case are to be treated as if identical."

https://www.ietf.org/rfc/rfc1035.txt

tomwilkie:
Fixes #1461

### DIFF
--- a/nameserver/entry.go
+++ b/nameserver/entry.go
@@ -5,6 +5,7 @@ import (
 	"encoding/gob"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/weaveworks/weave/net/address"
@@ -23,8 +24,26 @@ type Entry struct {
 }
 
 type Entries []Entry
+type CaseSensitive Entries
+type CaseInsensitive Entries
+type SortableEntries interface {
+	sort.Interface
+	Get(i int) Entry
+}
 
-func (e1 *Entry) equal(e2 *Entry) bool {
+// Gossip messages are sorted in a case sensitive order...
+func (es CaseSensitive) Len() int           { return len(es) }
+func (es CaseSensitive) Swap(i, j int)      { es[i], es[j] = es[j], es[i] }
+func (es CaseSensitive) Get(i int) Entry    { return es[i] }
+func (es CaseSensitive) Less(i, j int) bool { return es[i].less(&es[j]) }
+
+// ... but we store entries in a case insensitive order.
+func (es CaseInsensitive) Len() int           { return len(es) }
+func (es CaseInsensitive) Swap(i, j int)      { es[i], es[j] = es[j], es[i] }
+func (es CaseInsensitive) Get(i int) Entry    { return es[i] }
+func (es CaseInsensitive) Less(i, j int) bool { return es[i].insensitiveLess(&es[j]) }
+
+func (e1 Entry) equal(e2 Entry) bool {
 	return e1.ContainerID == e2.ContainerID &&
 		e1.Origin == e2.Origin &&
 		e1.Addr == e2.Addr &&
@@ -36,6 +55,24 @@ func (e1 *Entry) less(e2 *Entry) bool {
 	switch {
 	case e1.Hostname != e2.Hostname:
 		return e1.Hostname < e2.Hostname
+
+	case e1.Origin != e2.Origin:
+		return e1.Origin < e2.Origin
+
+	case e1.ContainerID != e2.ContainerID:
+		return e1.ContainerID < e2.ContainerID
+
+	default:
+		return e1.Addr < e2.Addr
+	}
+}
+
+func (e1 *Entry) insensitiveLess(e2 *Entry) bool {
+	// Entries are kept sorted by Hostname, Origin, ContainerID then address
+	e1Hostname, e2Hostname := strings.ToLower(e1.Hostname), strings.ToLower(e2.Hostname)
+	switch {
+	case e1Hostname != e2Hostname:
+		return e1Hostname < e2Hostname
 
 	case e1.Origin != e2.Origin:
 		return e1.Origin < e2.Origin
@@ -63,37 +100,37 @@ func (e1 *Entry) String() string {
 	return fmt.Sprintf("%s -> %s", e1.Hostname, e1.Addr.String())
 }
 
-func (es Entries) Len() int           { return len(es) }
-func (es Entries) Swap(i, j int)      { panic("Swap") }
-func (es Entries) Less(i, j int) bool { return es[i].less(&es[j]) }
-
-func (es Entries) check() error {
+func check(es SortableEntries) error {
 	if !sort.IsSorted(es) {
 		return fmt.Errorf("Not sorted!")
 	}
-	for i := 1; i < len(es); i++ {
-		if es[i].equal(&es[i-1]) {
-			return fmt.Errorf("Duplicate entry: %d:%v and %d:%v", i-1, es[i-1], i, es[i])
+	for i := 1; i < es.Len(); i++ {
+		if es.Get(i).equal(es.Get(i - 1)) {
+			return fmt.Errorf("Duplicate entry: %d:%v and %d:%v", i-1, es.Get(i-1), i, es.Get(i))
 		}
 	}
 	return nil
 }
 
-func (es *Entries) checkAndPanic() {
-	if err := es.check(); err != nil {
+func checkAndPanic(es SortableEntries) {
+	if err := check(es); err != nil {
 		panic(err)
 	}
 }
 
+func (es *Entries) checkAndPanic() *Entries {
+	checkAndPanic(CaseInsensitive(*es))
+	return es
+}
+
 func (es *Entries) add(hostname, containerid string, origin router.PeerName, addr address.Address) Entry {
-	es.checkAndPanic()
-	defer es.checkAndPanic()
+	defer es.checkAndPanic().checkAndPanic()
 
 	entry := Entry{Hostname: hostname, Origin: origin, ContainerID: containerid, Addr: addr}
 	i := sort.Search(len(*es), func(i int) bool {
-		return !(*es)[i].less(&entry)
+		return !(*es)[i].insensitiveLess(&entry)
 	})
-	if i < len(*es) && (*es)[i].equal(&entry) {
+	if i < len(*es) && (*es)[i].equal(entry) {
 		if (*es)[i].Tombstone > 0 {
 			(*es)[i].Tombstone = 0
 			(*es)[i].Version++
@@ -107,16 +144,16 @@ func (es *Entries) add(hostname, containerid string, origin router.PeerName, add
 }
 
 func (es *Entries) merge(incoming Entries) Entries {
-	es.checkAndPanic()
-	defer es.checkAndPanic()
+	defer es.checkAndPanic().checkAndPanic()
+
 	newEntries := Entries{}
 	i := 0
 
 	for _, entry := range incoming {
-		for i < len(*es) && (*es)[i].less(&entry) {
+		for i < len(*es) && (*es)[i].insensitiveLess(&entry) {
 			i++
 		}
-		if i < len(*es) && (*es)[i].equal(&entry) {
+		if i < len(*es) && (*es)[i].equal(entry) {
 			if (*es)[i].merge(&entry) {
 				newEntries = append(newEntries, entry)
 			}
@@ -133,8 +170,8 @@ func (es *Entries) merge(incoming Entries) Entries {
 
 // f returning true means keep the entry.
 func (es *Entries) tombstone(ourname router.PeerName, f func(*Entry) bool) Entries {
-	es.checkAndPanic()
-	defer es.checkAndPanic()
+	defer es.checkAndPanic().checkAndPanic()
+
 	tombstoned := Entries{}
 	for i, e := range *es {
 		if f(&e) && e.Origin == ourname {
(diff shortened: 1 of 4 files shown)


nameserver/nameserver.go
```diff
@@ -161,8 +161,9 @@ func (n *Nameserver) PeerGone(peer *router.Peer) {
 func (n *Nameserver) Delete(hostname, containerid, ipStr string, ip address.Address) error {
 	n.Lock()
 	n.infof("tombstoning hostname=%s, container=%s, ip=%s", hostname, containerid, ipStr)
+	lowerHostname := strings.ToLower(hostname)
 	entries := n.entries.tombstone(n.ourName, func(e *Entry) bool {
-		if hostname != "*" && e.Hostname != hostname {
+		if hostname != "*" && strings.ToLower(e.Hostname) != lowerHostname {
```
tomwilkie: I'm not convinced delete should be case insensitive.  Thoughts?

bboreham: Yes, to be consistent. Easier to understand.

tomwilkie:
@inercia would you mind giving this a look please?

rade:
What about the domain? Should that be treated case insensitively too?

Also, what's the motivation for dealing with the case insensitivity in the various lookups/comparisons, instead of simply lower-casing the hostname on entry creation?

tomwilkie:
> What about the domain? Should that be treated case insensitively too?

Domain is included in the hostname here (see https://github.com/weaveworks/weave/blob/master/weave#L844)

> Also, what's the motivation for dealing with the case insensitivity in the various lookups/comparisons, instead of simply lower-casing the hostname on entry creation?

Backwards compatibility.

tomwilkie:
> Also, what's the motivation for dealing with the case insensitivity in the various lookups/comparisons, instead of simply lower-casing the hostname on entry creation?

Actually this will change the sort order, which means it isn't backwards compatible.  Will change to lowercasing on entry creation.

nameserver/nameserver.go
rade: why do we need the `ToLower` here? It's an entry hostname, so surely it should be lower-case already.

rade:
>  Will change to lowercasing on entry creation.

So where does that leave us w.r.t. backward compatibility?

tomwilkie:
> So where does that leave us w.r.t. backward compatibility?

If you are running a mixed 1.1/1.2 cluster
- the 1.1 host will resolve lower case versions of the containers on the weave 1.2 hosts 
- if you are using non-lowercase hostnames/domain on a weave 1.1 hosts, the 1.2 host will not be able to resolve those names.

(1) is probably fine, but (2) looks like a show stopper.  We could solve it by:
- also looking for non-lowercased names in the lookup method (as well as lowercased names)
- going back to original scheme (don't lowercase on the way in, just on search), but keeping the case on sort.  This would also fix (1).

rade:
I reckon add/remove should be case-sensitive, but lookup shouldn't be, i.e. it should return all records that match case-insensitively. I suspect this is how normal DNS servers behave, e.g. you can give them a zone file with mixed case records and they will return case-insensitively matching ones on lookup. Worth checking though.

tomwilkie:
> I reckon add/remove should be case-sensitive

Agreed.

Attempt (3) - only do case insensitive matching on lookup.  Re: backwards compatibility - sort order is preserved.  From a 1.1 host, lookup will be case sensitive, and from a 1.2 host, it will not be.

tomwilkie:
Actually this won't work either; sort on a string puts all the upper case before the lowercase:

http://play.golang.org/p/UXLptUu953

```
func main() {
	a := []string{"aaa", "Aaa", "bbb", "Bbb", "aAa", "aaA", "bBb", "bbB"}
	sort.Strings(a)
	fmt.Println(a)
}
```

```
[Aaa Bbb aAa aaA aaa bBb bbB bbb]
```

Back to drawing board.

rade:
> Actually this won't work either; sort on a string puts all the upper case before the lowercase

The [objective](https://github.com/weaveworks/weave/pull/1462#issuecomment-143664165) is correct. Your implementation is broken.

tomwilkie:
If we change the sort order the gossip messages will fail to merge with 1.1 hosts.  So we can't change the sort order.

We could make lookup O(n), but that sucks.

rade:
> If we change the sort order the gossip messages will fail to merge with 1.1 hosts.

How would this manifest itself?

tomwilkie:
Routers with different sort orders would drop the connection to each other (nameserver.go:218 would fail)

rade:
Why not sort the received data instead of *checking* that it is sorted?

bboreham:
If you're sorting case-sensitively, why would the order be different?

rade:
> If you're sorting case-sensitively, why would the order be different?

It wouldn't be but then you cannot do an efficient case *in*sensitive lookup.

tomwilkie:
> Why not sort the received data instead of checking that it is sorted?

We can do that in 1.2, but the doesn't help with 1.2 sending gossips to 1.1.

rade:
Oh dear. Looks like we are in a bit of pickle here.

dpw:
Why can't you decouple the data structure that you gossip (which needs to be case-sensitively sorted for compatibility) from the data structure that you do lookups in?

tomwilkie:
@dpw we can, or course.  But we haven't done that yet (as it is simpler, with less code, not to).

As part of the original design we discussed putting indexes on the entries data structure to accelerate non-hostname lookups (ie reverse lookups, or delete by container id).  Perhaps, now that the order of the entires doesn't help for lookup, this is needed for lookups too?

tomwilkie:
And a even simpler solution would be to do a case sensitive sort for gossips we send, and a case-insensitive sort on gossips we receive.

dpw:
> @dpw we can, or course. But we haven't done that yet (as it is simpler, with less code, not to).

Ok. I interpreted Matthias "bit of a pickle" comment as suggesting that there was some genuine difficulty, not merely the need to write some code that you'd prefer not to.  As you were then.

tomwilkie:
@rade PTAL

Failures are in 602_proxy_docker_py_test and are unrelated.

nameserver/entry.go
```diff
@@ -63,37 +100,33 @@ func (e1 *Entry) String() string {
...
+	if err := check(es); err != nil {
 		panic(err)
 	}
 }
 
 func (es *Entries) add(hostname, containerid string, origin router.PeerName, addr address.Address) Entry {
-	es.checkAndPanic()
-	defer es.checkAndPanic()
+	checkAndPanic(CaseInsensitive(*es))
+	defer func() { checkAndPanic(CaseInsensitive(*es)) }()
```
rade: If you define
```
func (es *Entries) checkAndPanic() *Entries {
	checkAndPanic(CaseInsensitive(*es))
	return es
}
```
then the above becomes
```
defer es.checkAndPanic().checkAndPanic()
```

which is a trick I learned from https://github.com/davecheney/profile#usage

nameserver/nameserver.go
```diff
@@ -215,7 +217,7 @@ func (n *Nameserver) receiveGossip(msg []byte) (router.GossipData, router.Gossip
 		return nil, nil, fmt.Errorf("clock skew of %d detected", delta)
 	}
 
-	if err := gossip.Entries.check(); err != nil {
+	if err := check(CaseSensitive(gossip.Entries)); err != nil {
```
rade: We should drop this check. That way we switch to sending case-insensitive entries in 1.3, and get rid of most of of the code from this PR at that point.

test/245_dns_case_insensitive_test.sh
```diff
@@ -0,0 +1,31 @@
...
+start_container $HOST1 $C2/24 --name=SeEtWo
+assert_dns_record $HOST1 test seetwo.weave.local $C2
+assert_dns_record $HOST1 test SeeTwo.weave.local $C2
+assert_dns_record $HOST1 test SEETWO.weave.local $C2
+
+start_container $HOST1 $C3/24 --name=seetwo
+assert_dns_record $HOST1 test seetwo.weave.local $C2 $C3
+assert_dns_record $HOST1 test SeeTwo.weave.local $C2 $C3
+assert_dns_record $HOST1 test SEETWO.weave.local $C2 $C3
+assert "exec_on $HOST1 test dig +short seetwo.weave.local A | grep -v ';;' | wc -l" 2
```
rade: why do we need that last check?

tomwilkie: I added this test to expose the sort order issue; if the nameserver wasn't sorting properly, the internal records would be SeEtWo, seeone, seetwo, so a lookup for seetwo would just return len(1), not len(2).

tomwilkie:
Have tested manually - ubuntu-14 is this branch, ubuntu-15 is 1.1:

```
vagrant@ubuntu-14:~/src/github.com/weaveworks/weave$ nslookup searchapp.weave.local 172.17.42.1
Server:		172.17.42.1
Address:	172.17.42.1#53

Name:	searchapp.weave.local
Address: 10.32.0.12
Name:	searchapp.weave.local
Address: 10.40.0.2
Name:	searchapp.weave.local
Address: 10.40.0.3
Name:	searchapp.weave.local
Address: 10.32.0.4
Name:	searchapp.weave.local
Address: 10.32.0.3

vagrant@ubuntu-14:~/src/github.com/weaveworks/weave$ weave status dns
app          10.40.0.7       63acc94a6eed 56:f7:a1:9a:90:ff
app          10.40.0.6       c4078e711876 56:f7:a1:9a:90:ff
app          10.32.0.7       09cf6c314865 ae:23:70:89:80:61
app          10.32.0.8       5a7ac5be5157 ae:23:70:89:80:61
client       10.40.0.10      9b34528aee2e 56:f7:a1:9a:90:ff
client       10.32.0.11      2073da265f48 ae:23:70:89:80:61
elasticsearch 10.32.0.2       661f8e294fef ae:23:70:89:80:61
frontend     10.40.0.8       23d34d5994e1 56:f7:a1:9a:90:ff
frontend     10.40.0.9       52763050e15c 56:f7:a1:9a:90:ff
frontend     10.32.0.10      d34742dd2a09 ae:23:70:89:80:61
frontend     10.32.0.9       ff88aad13e05 ae:23:70:89:80:61
qotd         10.40.0.5       4c0277bfbc51 56:f7:a1:9a:90:ff
qotd         10.32.0.6       0dee827d5b34 ae:23:70:89:80:61
redis        10.40.0.4       14c131aec9e3 56:f7:a1:9a:90:ff
redis        10.32.0.5       038b5dd8171a ae:23:70:89:80:61
searchapp    10.40.0.2       97992c807504 56:f7:a1:9a:90:ff
searchapp    10.40.0.3       fa539363ed99 56:f7:a1:9a:90:ff
searchapp    10.32.0.3       35ec6a8ea094 ae:23:70:89:80:61
Searchapp    10.32.0.12      4e90d538db02 ae:23:70:89:80:61
searchapp    10.32.0.4       940d037096b5 ae:23:70:89:80:61
```

```
vagrant@ubuntu-15:~/src/github.com/weaveworks/scope$ weave status dns
Searchapp    10.32.0.12      4e90d538db02 ae:23:70:89:80:61
app          10.40.0.7       63acc94a6eed 56:f7:a1:9a:90:ff
app          10.40.0.6       c4078e711876 56:f7:a1:9a:90:ff
app          10.32.0.7       09cf6c314865 ae:23:70:89:80:61
app          10.32.0.8       5a7ac5be5157 ae:23:70:89:80:61
client       10.40.0.10      9b34528aee2e 56:f7:a1:9a:90:ff
client       10.32.0.11      2073da265f48 ae:23:70:89:80:61
elasticsearch 10.32.0.2       661f8e294fef ae:23:70:89:80:61
frontend     10.40.0.8       23d34d5994e1 56:f7:a1:9a:90:ff
frontend     10.40.0.9       52763050e15c 56:f7:a1:9a:90:ff
frontend     10.32.0.10      d34742dd2a09 ae:23:70:89:80:61
frontend     10.32.0.9       ff88aad13e05 ae:23:70:89:80:61
qotd         10.40.0.5       4c0277bfbc51 56:f7:a1:9a:90:ff
qotd         10.32.0.6       0dee827d5b34 ae:23:70:89:80:61
redis        10.40.0.4       14c131aec9e3 56:f7:a1:9a:90:ff
redis        10.32.0.5       038b5dd8171a ae:23:70:89:80:61
searchapp    10.40.0.2       97992c807504 56:f7:a1:9a:90:ff
searchapp    10.40.0.3       fa539363ed99 56:f7:a1:9a:90:ff
searchapp    10.32.0.3       35ec6a8ea094 ae:23:70:89:80:61
searchapp    10.32.0.4       940d037096b5 ae:23:70:89:80:61
```

tomwilkie:
Failure in test is the docker py test_792_explicit_port_protocol failure, unrelated.

test/245_dns_case_insensitive_test.sh
```diff
@@ -0,0 +1,31 @@
...
+
+C1=10.2.0.78
+C2=10.2.0.79
+C3=10.2.0.80
+
+start_suite "DNS lookup case (in)sensitivity"
+
+weave_on $HOST1 launch
+
+start_container_with_dns $HOST1        --name=test
```
rade: some odd whitespace there.

test/245_dns_case_insensitive_test.sh
```diff
@@ -0,0 +1,31 @@
...
+
+start_container $HOST1 $C1/24 --name=seeone
+assert_dns_record $HOST1 test seeone.weave.local $C1
+assert_dns_record $HOST1 test SeeOne.weave.local $C1
+assert_dns_record $HOST1 test SEEONE.weave.local $C1
```
rade: What is this set of tests exercising that isn't also exercised by the lot below?

tomwilkie: We need 3 dns entries for this test to properly test the sort order; this provides the dns entry that would split SeEtWo and seetwo if sorting isn't working.

I guess we can remove the ```assert_dns_record ...  seetwo.weave.local $C2 $C3``` ones below?  But I like the completeness.

rade: ok. makes sense.

PR #1464, opened 2015-09-23
Remove 'defer' line that can never execute

bboreham:
The loop entered just after this line has no `break` or `return` statements, so the function never exits.

(barring `panic`, and we do not use `recover`)


rade:
I'd rather keep that line; it's there so we do not need to reason about whether the loop can terminate.

PR #1469, opened 2015-09-25
Run tests in a separate container.

### LINKED ISSUE #1204: Don't run unit tests in build container
https://github.com/weaveworks/weave/pull/1120#issuecomment-123014932

tomwilkie:
Fixes #1204

### DIFF
--- a/test/image/test.sh
+++ b/test/image/test.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+set -e
+
+GOPATH=/home/go
+export GOPATH
+
+WEAVE_SRC=$GOPATH/src/github.com/weaveworks/weave
+
+# Expect that the weave repo will be bind-mounted into $GOPATH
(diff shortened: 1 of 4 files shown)


tomwilkie:
@dpw do you mind reviewing?

test/image/test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+set -e
+
+GOPATH=/home/go
+export GOPATH
+
+WEAVE_SRC=$GOPATH/src/github.com/weaveworks/weave
+
+# Expect that the weave repo will be bind-mounted into $GOPATH
```
dpw: I'm not sure this check is needed here.  It's done in the build container because part of the point of that is to make it easier for people outside of weaveworks to build weave.  But it looks like this container image will be used in a much more particular context, and won't be documented for outside use.

dpw:
Looks generally fine, see the comment.

Issue #1286, opened 2015-08-11
Switch to tools.git


tomwilkie:
Move gce.sh, scheduled and test runner there.

rade:
I have no idea what this issue is about. Please add a bit more detail.

tomwilkie:
I've made https://github.com/weaveworks/tools for a place to put scripts shared between weave, scope and the service.  Things like the lint script, the test script, the cover merge tool etc.  We should make weave.git use these tools, so that improvements are shared between projects.

Here is the scope change to use these scripts: https://github.com/weaveworks/scope/pull/354

tomwilkie:
Next up I want to move the runner and the gee script into build tools.git

PR #1470, opened 2015-09-25
Use test script from tools.git

tomwilkie:
Part of #1286


Makefile
```diff
@@ -44,6 +44,9 @@ NETGO_CHECK=@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
 	echo "    sudo go install -tags netgo std"; \
 	false; \
 }
+ENSURE_TOOLS=@if [ ! -d .tools ]; then \
+	git clone https://github.com/weaveworks/tools.git .tools/;\
+fi
```
rade: That seems like an odd way to do this. Why not just have a rule with `.tools` as a target and the `git clone` as the action?

rade:
@dpw mind taking a look at this?

dpw:
I really don't like the fact that this clones master of the tools repo.  The consequence of that is that a version of codebase that built today might not build tomorrow, due to a change in `tools`.  (And, yes, I don't like golang's package system for exactly that reason.)

So we should specify the version of tools that we clone.  But that raises the question, why not just use git submodules?  I'm no great fan of git submodules, but I'd rather we used it than ad-hoc substitute.  The `git submodule update` invocation can be incorporated into the Makefile so that there is no need to run it explicitly.

Makefile
```diff
@@ -108,8 +108,14 @@ $(WEAVE_EXPORT): $(IMAGES_UPTODATE)
 $(DOCKER_DISTRIB):
 	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
 
-tests: $(COVER_EXE)
-	@test/units.sh
+tests: $(COVER_EXE) .tools
+	.tools/test
+
+lint: .tools
```
rade: this needs adding to `.PHONY`.

rade:
> why not just use git submodules?

indeed.

tomwilkie:
@dpw PTAL

dpw:
I don't see a reason to hide this away in `.tools`.  There seems to be some confusion in the Makefile changes whether it is `tools` or `.tools`.  (Though note that there are some obsolete references to a `tools` directory in `.gitignore` that should have been cleaned up in 795d0edc73).

A separate suggestion, and I can imagine the hassle of fixing this now, but while I'm looking at it: I don't think calling the repo "tools" is great.  Such over-generic names tend to lead to unkempt dumping grounds of unrelated stuff, in my experience.  That already looks like it might be the case: What does the socks thing have in common with the other contents of that repo?  So it might be better to call it build-tools (that seems an apt description of the rest of the contents) and remove the socks thing.

tomwilkie:
- moved .tools -> tools
- renamed tools.git to build-tools.git
- raised separate issue for socks: https://github.com/weaveworks/build-tools/issues/4
- tidy up confusion in Makefile
- removed references to tools from .gitignore

dpw:
Looks ok.  Are the circleci failures significant?

tomwilkie:
Yes in this case they are; need to think on how to fix that.

PR #1471, opened 2015-09-25
Publish latest images from CI on successful builds.

### LINKED ISSUE #1295: publish latest images from CI
Currently, we publish snapshot builds once in a while manually, however it would be much easier if this was done via CI once tests passed. This way some of our users as well as members of the team could test latest code quicker.

tomwilkie:
This time rebuild without instrumentation first.

Fixes #1295

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PUBLISH=publish_weave publish_weaveexec
 
 .DEFAULT: all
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -120,10 +120,12 @@ endif
 
 publish: $(PUBLISH)
 
-clean:
+clean-bin:
```
rade: this needs adding to `.PHONY`

Issue #1472, opened 2015-09-25
"weave launch --log-level=debug" makes "weave env" choke


dpw:
With current master:

```
# weave launch --log-level=debug
# weave env
weaveproxy container is not present. Have you launched it?
```

It works in 1.1.0.

PR #1474, opened 2015-09-25
[proxy] use a container-local unix socket to get the address, instead of logs

### LINKED ISSUE #1453: launching proxy fails with non-default logging driver
I've tested it on latest CoreOS 808.0.0 (alpha) that ships with Docker 1.8.2 and runs it with `--log-driver=journald` and got this

```
Sep 21 19:08:41 core-01 weave[1357]: The weaveproxy container has died. Consult the logs with 'docker logs weaveproxy' for further details.
```
where `docker logs weaveproxy` says:
```
"logs" command is supported only for "json-file" logging driver (got: journald)
```

paulbellamy:
Fixes #1453 and, #1472

### DIFF
--- a/weave
+++ b/weave
@@ -668,7 +668,9 @@ http_call_unix() {
     http_verb="$3"
     url="$4"
     shift 4
+    # NB: requires curl >= 7.40
     output=$(docker exec $container curl -s -S -X $http_verb --unix-socket $socket "$@" http:$url)
+    # in some docker versions, `docker exec` does not fail when the executed command fails
     [ -n "$output" ] || return 1
     echo $output
 }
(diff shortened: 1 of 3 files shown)


weave
```diff
@@ -700,17 +711,19 @@ wait_for_status() {
...
+wait_for_status_unix() {
+    err_msg=$(death_msg $1)
+    while true ; do
+        http_call_unix $1 status.sock GET /status >/dev/null 2>&1 && return 0
+        if ! container_id $1 >/dev/null 2>&1 ; then
+            echo "$err_msg" >&2
+            return 1
+        fi
+        fractional_sleep 3.1
+    done
```
rade: Surely there must be a way to refactor `wait_for_status_unix` and `http_call_unix` to avoid all the duplication with `wait_for_status` and `http_call`.

paulbellamy: None that's actually shorter, unless there's some shell trickery I'm missing.

weave
```diff
@@ -663,6 +663,17 @@ http_call_ip() {
...
 
+http_call_unix() {
+    container="$1"
+    socket="$2"
+    http_verb="$3"
+    url="$4"
+    shift 4
+    output=$(docker exec $container curl -s -S -X $http_verb --unix-socket $socket "$@" http:$url)
+    [[ -n "$output" ]] || return 1
+    echo $output
```
rade: Why do we need all this output capture & checking? Does the `docker exec` not fail when the curl fails? and why do we not have a `--connect-timeout 3` here?

paulbellamy: `docker exec` on docker 1.3 does not fail when the command fails.

`--connect-timeout` is a bit useless with unix sockets, as they fail instantly.

rade: > `docker exec` on docker 1.3 does not fail when the command fails.

oh. in what version of docker did this get fixed?

paulbellamy: Not sure. It's not mentioned in their changelog. Might have been with the move to libcontainer.

rade:
The CircleCI tests are timing out.

Please review my refactor commit. The only behavioural change is the retry reduction. I think that is correct; the 3s connect timeout in the normal status case doesn't usually kick in; if there is nothing listening then curl instantly fails.

paulbellamy:
The retry reduction means we're doing 10 docker execs/second until the proxy starts. that seems a bit excessive to me.

rade:
It's not excessive and exactly what happens with the router. We don't a delay of several seconds in `weave launch`. There was an issue/PR to address precisely that.

paulbellamy:
With the router we're just doing an http request, not a docker exec, though.

rade:
meh

paulbellamy:
I had accidentally used `[[` which broke the `--local` test.

PR #1476, opened 2015-09-29
[proxy] unmount /var/run after boot, so the kernel can GC veth pairs

### LINKED ISSUE #1455: Proxy volume mount of /var/run prevents netns cleanup
The weaveproxy volume mount of `/var/run` captures references to extant Docker network namespaces in `/var/run/docker/netns` due to Docker's use of `MS_REC` (`mount --rbind`) on bind mount creation. This prevents associated veths from being cleaned up on container exit, potentially resulting in duplicate IP address use and other mayhem.

Reproducible on Docker 1.7.1 but not Docker 1.8.2 (**Edit** - no longer true - see below)

paulbellamy:
Change default unix listener to `unix:///var/run/weave/weave.sock`, and always make a symlink from `/var/run/weave.sock` to `/var/run/weave/weave/sock`

Fixes #1455. Pretty ugly hack, but...

Not sure if there's a good way to test this which represents something a user would do/care about, but when I've broken it 690 has failed, so...

### DIFF
--- a/weave
+++ b/weave
@@ -96,7 +96,7 @@ usage() {
 
 exec_remote() {
     docker $DOCKER_CLIENT_ARGS run --rm --privileged --net=host \
-        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v /var/run:/var/run \
         -v /proc:/hostproc \
         -e PROCFS=/hostproc \
         -e DOCKERHUB_USER="$DOCKERHUB_USER" \
@@ -1041,7 +1041,7 @@ proxy_args() {
     if [ -z "$PROXY_HOST" ] ; then
       case "$DOCKER_CLIENT_HOST" in
         ""|unix://*)
-          PROXY_HOST="unix:///var/run/weave.sock"
+          PROXY_HOST="unix:///var/run/weave/weave.sock"
           ;;
         *)
           PROXY_HOST="tcp://0.0.0.0:12375"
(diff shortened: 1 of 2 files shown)


rade:
We can make this more specialised, i.e. simply in the script always create `/var/run/weave` and create a symlink to `/var/run/weave/weave.sock` from `/var/run/weave.sock`.

This should simplify the code a lot.

rade:
@paulbellamy LGTM but the PR summary and description bear little resemblance to the code now. Please fix.

weave
```diff
@@ -96,7 +96,7 @@ usage() {
 
 exec_remote() {
     docker $DOCKER_CLIENT_ARGS run --rm --privileged --net=host \
-        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v /var/run:/var/run \
```
rade: The only reason we need this is because of the symlink creation, correct?

weave
```diff
@@ -1211,6 +1212,7 @@ launch_proxy() {
         --entrypoint=/home/weave/weaveproxy \
         $WEAVEPROXY_DOCKER_ARGS $EXEC_IMAGE $COVERAGE_ARGS $PROXY_ARGS)
     wait_for_log $PROXY_CONTAINER_NAME "proxy listening"
+    ln -sf -T /var/run/weave/weave.sock /var/run/weave.sock
```
rade: This is there just for backward compatibility, right? Please a) add a comment to say so, and b) file an issue to remove this and the `/var/run` volume mount in `exec_remote` for 1.2.

rade: which makes me wonder... we could use a little trick here so we don't need the `/var/run` mount in `exec_remote` even now... run the `ls` in a container (yet another incarnation of the `weavexec` image) that mounts `/var/run/`.

weave
paulbellamy: yes

weave
paulbellamy: We could do, but it's not a huge win when it's being removed in the next version anyway.

PR #1477, opened 2015-09-29
rewrite /etc/hosts in containers

### LINKED ISSUE #1374: Docker 1.8+ breaks resolution of unqualified names
I was able to use Weave 1.0.2 with experimental Docker, but it turns out that short hostnames are not working as expected due to `/etc/hosts` being populated with entries _dynamically_.

I've [used Docker Machine](http://sirile.github.io/2015/07/02/using-docker-18-experimental-with-docker-machine-and-virtualbox-driver-boot2docker.html) to setup a test VM and it gave me docker/docker@5dadfa8 build from Aug 26 17:33:57 UTC 2015.

inercia:
Fix #1374 by rewritting `/etc/hosts` in containers.

* So far we have been rewritting `/etc/hosts` only when attaching from the proxy, but we must also do it when running `weave run`.
* We do not rewrite `/etc/hosts` when running `weave attach`
* When rewritting hosts, we also add any extra hosts provided with `--add-host` arguments.

This PR also removes `weavehosts` and rewrites `/etc/hosts` from the `weave` script.

_Note_: this branch is currently based on `experimental_docker`, so it must be rebased on `master` before merging...


rade:
> When rewritting hosts, we filter out any `X` hostname in `/etc/hosts` where `X.bridge` also exists.

I [thought](https://github.com/weaveworks/weave/issues/1374#issuecomment-141422996) we were going to create an `/etc/hosts` from scratch, rather than filtering.

proxy/proxy.go
```diff
@@ -330,8 +330,14 @@ func (proxy *Proxy) attach(container *docker.Container) error {
 	Log.Infof("Attaching container %s with WEAVE_CIDR \"%s\" to weave network", container.ID, strings.Join(cidrs, " "))
 	args := []string{"attach"}
 	args = append(args, cidrs...)
-	if !proxy.NoRewriteHosts {
-		args = append(args, "--rewrite-hosts")
+	if proxy.NoRewriteHosts {
+		args = append(args, "--no-rewrite-hosts")
+	} else {
+		if container.HostConfig != nil {
```
rade: `else if`

weave
```diff
@@ -813,6 +813,16 @@ dns_args() {
...
 }
 
+rewrite_hosts() {
+    path=$(docker inspect -f '{{.HostsPath}}' $CONTAINER)
+    name=$(docker inspect -f '{{.Config.Hostname}}' $CONTAINER)
+    containers_path="$(dirname $(dirname "$path"))"
+    docker run --rm \
+        -v "$containers_path":"$containers_path" \
+        --entrypoint=/bin/sh \
+        $EXEC_IMAGE -c "/usr/bin/weavehosts $path $name $ALL_CIDRS"
```
rade: why can't we just invoke `weavehosts` directly, as we used to?

weave
```diff
@@ -813,6 +813,16 @@ dns_args() {
     fi
 }
 
+rewrite_hosts() {
+    path=$(docker inspect -f '{{.HostsPath}}' $CONTAINER)
+    name=$(docker inspect -f '{{.Config.Hostname}}' $CONTAINER)
```
rade: using a single `docker inspect`, as we used to, is more efficient.

weave
inercia: Because it depends on the path that leads to this call. `docker run` -> `weaveproxy` -> `weave attach` was ok, because `weaveproxy` is being launched with `-v "$containers_path":"$containers_path"`. But that does not happen when running `weave run` or just `weave attach`: the containers path is not being mounted, so we cannot rewrite `/etc/hosts`

weave
inercia: Ok, I'll use a simple `docker inspect` and then split the output.

proxy/proxy.go
inercia: I'll fix.

weave
rade: I see.

But in that case, couldn't we mount just the specific file instead of the dir?

Also, why is the entrypoint `/bin/sh` instead of `/usr/bin/weavehosts`? Seems like an unnecessary indirection.

inercia: > I see. Do we still need the volume mount in the proxy then?

Not sure but probably not. I'll investigate if there are more usages.

> Also, couldn't we mount just the specific file instead of the dir? 

Not sure about how these Docker mounts work, but probably not, as we `delete("/etc/hosts")` and then `create("/etc/hosts")`, so we probably need the whole dir for creating the new hosts.

> And perhaps mount it in a fixed location - mounting it in some path-which-could-be-anywhere seems kinda dangerous.

If we mount `<some_path>/hosts` at `/tmp/<some_other_path>`, what should `<some_other_path>` be? Not sure that using a fixed path is safer than a _"random"_ path...

> Also, why is the entrypoint `/bin/sh` instead of `/usr/bin/weavehosts`? Seems like an unnecessary indirection.

Probably. I will change it.

rade: > Not sure that using a fixed path is safer than a "random" path..

It is safer because it prevents accidentally mounting over some dir needed by the container.

inercia: Ok. Could you suggest a mount point on the container?

inercia:
@rade Should we modify `/etc/hosts` when the user launches the container with `weave run --without-dns`? The `--without-dns` sounds to me like _"please do not put any of your DNS tricks"_

rade:
> The `--without-dns` sounds to me like _"please do not put any of your DNS tricks"_

agreed.

weave
rade: `/containers`?

weave
```diff
@@ -806,6 +806,19 @@ dns_args() {
     fi
 }
 
+rewrite_hosts() {
+    hosts_path_and_name=$(docker inspect -f '{{.HostsPath}} {{.Config.Hostname}}' $CONTAINER)
+    hosts=$(echo $hosts_path_and_name | cut -f1 -d" ")
+    name=$(echo $hosts_path_and_name | cut -f2 -d" ")
+    containers_path="$(dirname $(dirname "$hosts"))"
```
rade: why not just *one* `dirname`?

weave
```diff
@@ -806,6 +806,19 @@ dns_args() {
...
+    mnt="/containers"
+    mnt_hosts=$(echo $hosts | sed 's|'$containers_path'|'$mnt'|g')
```
rade: related to the `dirname` question... why isn't this just `$mnt/$(basename "$hosts"))`?

weave
inercia: Well, it does the same thing [launch_proxy](https://github.com/weaveworks/weave/blob/master/weave#L1203) does... Only one `dirname` would result in `/containers/[hosts, resolv.conf, ...]`, so `/containers` wouldn't contain _containers_ anymore but only the current container...  I don't mind if you prefer it this way...

weave
inercia: Yes, if we remove the two `dirname`s, this can be simplified...

weave
inercia: Done.

weave
inercia: Done.

weave
rade: >  so `/containers` wouldn't contain containers anymore but only the current container

That is actually very desirable. Unlike the proxy, we only need access to the dir of one container.

Obviously we should change the dir name to `/container`.

weave
```diff
@@ -806,6 +806,19 @@ dns_args() {
...
+    hosts=$(echo $hosts_path_and_name | cut -f1 -d" ")
+    name=$(echo $hosts_path_and_name | cut -f2 -d" ")
+    containers_path="$(dirname "$hosts")"
+    mnt="/container"
+    mnt_hosts="$mnt/$(basename $hosts)"
+    docker run --rm \
+        -v "$containers_path":"$mnt" \
+        --entrypoint=/usr/bin/weavehosts \
+        $EXEC_IMAGE $mnt_hosts $name $ALL_CIDRS
+}
```
rade: The quoting is all over the place here. I'd ditch it all.

bboreham:
Does this change mean we invoke `weavehosts` via Docker every time, even when run from the proxy?  Slowing the proxy down like that would be a shame.

rade:
https://github.com/weaveworks/weave/pull/1477#issuecomment-144032662 still needs answering.

inercia:
> I thought we were going to create an /etc/hosts from scratch, rather than filtering.

We are creating `/etc/hosts` from scratch, but the way we were re-creating it was by copying the contents from the old one. I think it is reasonable to also remove any `X.bridge` name we find there...

rade:
>  but the way we were re-creating it was by copying the contents from the old one

why bother copying anything?

inercia:
> Does this change mean we invoke weavehosts via Docker every time, even when run from the proxy? Slowing the proxy down like that would be a shame.

@bboreham we could run `weavehosts` in two different modes, _with_ or _without_ the container, but that would mean adding some logic depending on how `weave attach` was being invoked... Maybe the proxy could use an undocummented `weave attach --no-weavehosts-container` flag...

rade:
> we could run weavehosts in two different modes

Sounds fiddly. I doubt it's worth the hassle.

inercia:
>> but the way we were re-creating it was by copying the contents from the old one
> why bother copying anything?

Because it is safer: we don't know the contents of `/etc/hosts`.

inercia:
> > we could run weavehosts in two different modes

> Sounds fiddly. I doubt it's worth the hassle.

I think so too.

rade:
> Because it is safer: we don't know the contents of `/etc/hosts`.

That actually makes it *less* safe. The `/etc/hosts` is created by docker. We have no idea what changes they are going to be introducing that breaks weave. So we should create our own instead.

bboreham:
> I doubt it's worth the hassle.

Unless the user's docker has defaulted to vfs...

inercia:
> That actually makes it less safe. The /etc/hosts is created by docker. We have no idea what changes they are going to be introducing that breaks weave. So we should create our own instead.

Are we sure there is nothing there that could have been written by someone else? What if some process in the container adds something to `/etc/hosts` and then we `attach` and replace `/etc/hosts` by our copy? There must be a reason why we have been copying it so far, and @bboreham probably knows it...

rade:
>  There must be a reason why we have been copying it so far

It was The Simplest Thing That Could Possibly Work.

Unfortunately there is no longer *any* Way That Could Possibly Work that doesn't potentially break something.

>  What if some process in the container adds something to `/etc/hosts` and then we attach and replace `/etc/hosts` by our copy? 

Umm. Should we do anything at all on dynamic attach? We don't right now. Dynamically attached containers typically won't have their resolver set to weavedns, so name resolution won't work anyway. NB: We probably *should* rewrite /etc/hosts on `weave run`.

inercia:
> Umm. Should we do anything at all on dynamic attach? 

We are dynamically `attach`ing when using the proxy, right? Can we be sure the contents of `/etc/hosts` have not been modified when the `attach` is performed?

> We probably should rewrite `/etc/hosts` on `weave run`

This PR could affect three different use cases:
1. `weave run`
2. `docker run` && `weave attach`
3. the proxy && `docker run`

So far we have been rewriting `/etc/hosts` only in case _(3)_, when you are suggesting we should only do it in case _(1)_. Is that right?

Do you mean rewriting with filtering, or just writing a fresh version of `/etc/hosts`?

rade:
> We are dynamically attaching when using the proxy, right? Can we be sure the contents of `/etc/hosts` have not been modified when the attach is performed?

Yes, because the container will be waiting in `weavewait` until weave networking has been set up.

> you are suggesting we should only do it in case (1). Is that right?

(1) and (3).

> Do you mean rewriting with filtering, or just writing a fresh version of `/etc/hosts`?

I am arguing that we should do the latter.

inercia:
> > Do you mean rewriting with filtering, or just writing a fresh version of /etc/hosts?

> I am arguing that we should do the latter.

Ok. In that case I think we don't need `weavehosts` at all: we can use some bash scripting...

inercia:
Currently failing on `200_dns_test.sh` with no error messages... unrelated?

rade:
does that test work for you locally?

inercia:
@rade I've just found the error... it's strange it didn't fail in my local machine... :confused:

test/Vagrantfile
```diff
@@ -22,8 +22,8 @@ def configure_docker(host, hostname, ip)
 
   host.vm.synced_folder ".", "/vagrant", disabled: true
 
-  host.vm.provision :shell, :inline => "sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D"
-  host.vm.provision :shell, :inline => "echo deb https://apt.dockerproject.org/repo ubuntu-vivid main > /etc/apt/sources.list.d/docker.list"
+  host.vm.provision :shell, :inline => "sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys --recv-keys 58118E89F3A912897C070ADBF76221572C52609D"
+  host.vm.provision :shell, :inline => "echo deb https://apt.dockerproject.org/repo ubuntu-vivid experimental > /etc/apt/sources.list.d/docker.list"
```
rade: surely this change shouldn't be here.

vagrant-common.rb
```diff
@@ -13,9 +13,9 @@ def install_build_deps(vm, pkgs)
   vm.provision :shell, :inline => <<SCRIPT
 export DEBIAN_FRONTEND=noninteractive
 apt-key adv \
-  --keyserver hkp://p80.pool.sks-keyservers.net:80 \
+  --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys \
   --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
-echo 'deb https://apt.dockerproject.org/repo ubuntu-vivid main' \
+echo 'deb https://apt.dockerproject.org/repo ubuntu-vivid experimental' \
```
rade: surely this change shouldn't be here.

weave
```diff
@@ -107,6 +107,7 @@ exec_remote() {
         -e WEAVE_PASSWORD \
         -e WEAVE_PORT \
         -e WEAVE_CONTAINER_NAME \
+        -e DNS_EXTRA_HOSTS_ARGS \
```
rade: why?

weave
```diff
@@ -238,14 +239,17 @@ elif [ "$1" = "run" -a -z "$IS_LOCAL" ] ; then
...
         DNS_ARGS=$(exec_remote dns-args "$@" || true)
+        DNS_EXTRA_HOSTS_ARGS=$(exec_remote dns-extra-hosts "$@" || true)
+        ATTACH_ARGS="$ATTACH_ARGS --rewrite-hosts"
     fi
     shift $(dns_arg_count "$@")
     collect_cidr_args "$@"
     shift $CIDR_ARG_COUNT
-    CONTAINER=$(docker $DOCKER_CLIENT_ARGS run -e WEAVE_CIDR=none $DNS_ARGS -d "$@")
-    exec_remote attach $CIDR_ARGS --or-die $CONTAINER >/dev/null
+    CONTAINER=$(docker $DOCKER_CLIENT_ARGS run -e WEAVE_CIDR=none $DNS_ARGS $DNS_EXTRA_HOSTS_ARGS -d "$@")
```
rade: why do we have `$DNS_EXTRA_HOSTS_ARGS` in here? They will be in `$@` too, surely.

weave
```diff
@@ -1444,6 +1505,8 @@ case "$COMMAND" in
         echo $RES
         ;;
     attach)
+        extra_hosts=
+        rewrite_hosts=
```
rade: why the lower-case? We use all-caps var names (nearly) everywhere else.

weave
```diff
@@ -238,14 +239,17 @@ elif [ "$1" = "run" -a -z "$IS_LOCAL" ] ; then
     # because we are passing arbitrary arguments on to docker run, and
     # we can't rely on our baked-in docker to support those arguments.
     shift 1
+    ATTACH_ARGS="--or-die"
     if [ "$1" != "--without-dns" ] ; then
         DNS_ARGS=$(exec_remote dns-args "$@" || true)
+        DNS_EXTRA_HOSTS_ARGS=$(exec_remote dns-extra-hosts "$@" || true)
```
rade: it's expensive to make two remote calls here. Can you think of a not-to-hideous way of avoiding that?

weave
```diff
@@ -804,6 +816,51 @@ dns_args() {
...
+
+    cat <<-EOF
+
+# default localhost entries
+127.0.0.1       localhost
+::1             ip6-localhost ip6-loopback
+fe00::0         ip6-localnet
+ff00::0         ip6-mcastprefix
+ff02::1         ip6-allnodes
+ff02::2         ip6-allrouters
```
rade: are the ipv6 entries needed?

weave
```diff
@@ -804,6 +816,51 @@ dns_args() {
     fi
 }
 
+default_etc_hosts() {
+    name=$1
+    shift
+    extra_hosts=$@
```
rade: expansion of `$@` is rather special, so I suggest you drop this line and instead have `for host in "$@"...` below.

weave
```diff
@@ -804,6 +816,51 @@ dns_args() {
...
+    hosts=$(echo $hosts_path_and_name | cut -f1 -d" ")
+    name=$(echo $hosts_path_and_name | cut -f2 -d" ")
+    containers_path=$(dirname $hosts)
+    mnt=/container
+    mnt_hosts=$mnt/$(basename $hosts)
+    contents="$(default_etc_hosts $name $@)"
+    docker run --rm \
+        -v $containers_path:$mnt \
+        --entrypoint=sh \
+        $EXEC_IMAGE -c "echo '$contents' > $mnt_hosts && rm -f $mnt_hosts && echo '$contents' > $mnt_hosts"
```
rade: why is the first `echo` needed?

rade:
need some docs too

test/Vagrantfile
inercia: Yes, as the PR description says, _"this branch is currently based on experimental_docker, so it must be rebased on master before merging..."_, so this change will disappear

vagrant-common.rb
inercia: ditto

weave
inercia: Yes, you are right, I usually use lower-case names for local vars... I'll change it

weave
inercia: I know, but I honestly cannot think of a better way of doing it... I'm open to suggestions :confused:

weave
inercia: This is copy&paste of the standard entries in Ubuntu systems... I think it is safe to leave it as it is...

weave
inercia: I agree. I'll change it.

weave
inercia: It mimics the behaviour described by @awh [here](https://github.com/weaveworks/weave/issues/1374#issuecomment-140118011). I think it could be removed...

inercia: No, it can't... :)

rade: Ah, the first echo is the one that writes to the container fs. So it's the *2nd* echo that seems unnecessary. Unless it causes docker to subsequently explode / spew errors into the log.

weave
inercia: I don't use this anymore. I'll remove.

weave
inercia: Right. Fixed.

weave
inercia: Done, but I don't know what Docker will think about the `hosts` file disappearing...

rade: Try it.

inercia:
> need some docs too

There are no user visible changes, right? I mean that we are just trying to keep things working... And documenting _"oh, by the way, we are going to rewrite your /etc/hosts file"_ maybe will result in more confusion for users...

weave
inercia: Yes, I mean that my Docker daemon seems happy so far...

rade:
> There are no user visible changes, right?

The [proxy docs](http://docs.weave.works/weave/latest_release/proxy.html#ipam) say "automatically rewrite /etc/hosts to replace the docker IP with the container’s weave IP." and document `--no-rewrite-hosts` as the flag to disable that. So that needs updating.

I think we also want a documented `--no-rewrite-hosts` flag for `weave run`.

inercia:
> The [proxy docs](http://docs.weave.works/weave/latest_release/proxy.html#ipam) say "automatically rewrite /etc/hosts to replace the docker IP with the container’s weave IP." and document `--no-rewrite-hosts` as the flag to disable that. So that needs updating.

The behaviour of the proxy when using `--no-rewrite-hosts` has not changed: we will not modify the `/etc/hosts`. Should we do it in a different way?

> I think we also want a documented `--no-rewrite-hosts` flag for `weave run`.

There is no `--no-rewrite-hosts` flag for `weave run`. `weave run` does not rewrite `/etc/hosts` when `--without-dns` has been specified. I will document this.

rade:
> The behaviour of the proxy when using `--no-rewrite-hosts` has not changed

The behaviour w/o the flag has changed, i.e. we do a lot more now than just "replace the docker IP with the container's weave IP".

> There is no `--no-rewrite-hosts` flag for weave run.

I know that.

> `weave run` does not rewrite `/etc/hosts` when `--without-dns` has been specified.

I know that too.

This does not give sufficient control. You may want to suppress rewriting of `/etc/hosts` even when using DNS. Or, coming to think of it, even you may want to rewrite `/etc/hosts` even when not using DNS; though this is less important.

inercia:
> This does not give sufficient control. You may want to suppress rewriting of /etc/hosts even when using DNS. Or, coming to think of it, even you may want to rewrite /etc/hosts even when not using DNS; though this is less important.

I don't see the point in rewriting `/etc/hosts` when not using WeaveDNS (or maybe that should be left for another PR, as the original intention of this one was to fix WeaveDNS when used with _docker-experimental_), but I will add a `--no-rewrite-hosts` for `weave run`.

rade:
> I don't see the point in rewriting `/etc/hosts` when not using WeaveDNS

The original purpose of rewriting `/etc/hosts` was to ensure that the likes of `hostname -i` return the weave IP. That is quite independent of using weaveDNS.

inercia:
Ok, then what do you think about having `weave run [--with-dns | --without-dns] [--no-rewrite-hosts] ...`?. We would overwrite `/etc/hosts` unless `--no-rewrite-hosts`, because deaing wth all the `--with*-dns` x `--*rewrite-hosts` combinations seems a bit confusing...

rade:
yep, that makes sense. It's what the proxy does too (I believe/hope).

inercia:
Build is currently failing with `exec_on host1-3153-0.us-central1-a.positive-cocoa-90213 c1 ping -nq -W 1 -c 1 10.2.1.71" failed: program terminated with code 1 instead of 0`. Unrelated?

inercia:
... and, with a new `push -f`, it works again :confused:

site/proxy.md
```diff
@@ -148,8 +148,11 @@ container with a blank `WEAVE_CIDR`, e.g.
 
 When launching weave-enabled containers, the proxy will automatically
 rewrite `/etc/hosts` to replace the docker IP with the container's
-weave IP. If you need the docker IP to remain in `/etc/hosts`, the
-proxy must be launched with the `--no-rewrite-hosts` flag.
+weave IP. The new `/etc/hosts` files will also contain all the additional
+names that could have been specified with `--add-host` parameters,
+as well as all the localhost entries. In case you prefer to keep `/etc/hosts`
+intact (for example, because you need the docker IP instead of the
+weave IP), the proxy must be launched with the `--no-rewrite-hosts` flag.
```
rade: I suggest moving this into its own section.

Here's the wording Michael and I have come up with:

## <a name="etchosts"></a>Name resolution via `/etc/hosts`

When starting weave-enabled containers, the proxy will automatically replace the container's `/etc/hosts` file, and disable Docker's control over it. The new file contains an entry for the container's hostname and weave IP address, as well as additional entries that have been specified with `--add-host` parameters. This ensures that

- name resolution of the container's hostname, e.g. via `hostname -i`, returns the weave IP address. This is required for many cluster-aware applications to work.
- unqualified names get resolved via DNS, i.e. typically via weaveDNS to weave IP addresses. This is required so that in a typical setup one can simply "ping <container-name>", i.e. without having to specify a `.weave.local` suffix.

In case you prefer to keep `/etc/hosts` under Docker's control (for example, because you need the hostname to resolve to the Docker-assigned IP instead of the weave IP, or you require name resolution for Docker-managed networks), the proxy must be
launched with the `--no-rewrite-hosts` flag.

    host1$ weave launch-router && weave launch-proxy --no-rewrite-hosts

weave
rade: It's also [hard-coded into docker](https://github.com/docker/docker/blob/master/vendor/src/github.com/docker/libnetwork/etchosts/etchosts.go#L25), so let's leave it.

weave
```diff
@@ -810,6 +823,50 @@ dns_args() {
...
+    NAME=$1
+    shift
+    echo "# created by Weave - BEGIN"
+    echo "# container hostname"
+    for CIDR in $ALL_CIDRS ; do
+        echo "${CIDR%/*}    $NAME"
+    done
+    echo
+    echo "# static names added with --add-host"
+    for EXTRA_HOST in $@ ; do
```
rade: should be `"@"`, surely.

weave
```diff
@@ -783,6 +789,13 @@ dns_args() {
             -h|--hostname|--hostname=*)
                 HOSTNAME_SPECIFIED=1
                 ;;
+            --add-host)
+                DNS_EXTRA_HOSTS_ARGS="--add-host=$2 $DNS_EXTRA_HOSTS_ARGS"
+                shift
+                ;;
+            --add-host=*)
+                DNS_EXTRA_HOSTS_ARGS="--add-host=${1#*=} $DNS_EXTRA_HOSTS_ARGS"
+                ;;
```
rade: What's the rationale for doing this in here rather than having a separate function?

As things stand, we will generally invoke `dns_args` twice, which is very expensive since we do all the work twice, including a call to weave (to obtain the domain).

weave
```diff
@@ -242,10 +242,15 @@ elif [ "$1" = "run" -a -z "$IS_LOCAL" ] ; then
         DNS_ARGS=$(exec_remote dns-args "$@" || true)
     fi
     shift $(dns_arg_count "$@")
+    if [ "$1" = "--no-rewrite-hosts" ] ; then
+        shift
+    else
+        DNS_EXTRA_HOSTS_ARGS="--rewrite-hosts $(exec_remote dns-extra-hosts "$@" || true)"
+    fi
```
rade: We need similar logic in the local `run`.

site/proxy.md
inercia: Yes, much better... Done.

weave
inercia: I think it is not strictly necessary, as `$@` will be something like `static.name:1.1.1.1 another.name:2.2.2.2`... but I've surrounded it with `"` anyway

weave
inercia: Yes, that can be _optimized_. I'll fix it.

site/proxy.md
```diff
@@ -271,6 +288,13 @@ specify addresses manually instead:
 
     $ weave run 10.2.1.1/24 -ti ubuntu
 
+`weave run` will rewrite `/etc/hosts` in the same way
+[the proxy does](weavedns.html#usage). In case you prefer to keep
```
rade: shouldn't that link point to #etchosts?

weave
```diff
@@ -242,10 +263,16 @@ elif [ "$1" = "run" -a -z "$IS_LOCAL" ] ; then
         DNS_ARGS=$(exec_remote dns-args "$@" || true)
     fi
     shift $(dns_arg_count "$@")
+    if [ "$1" = "--no-rewrite-hosts" ] ; then
+        shift
+    else
+        dns_extra_hosts_args "$@"
```
rade: nice! you managed to get rid of the 2nd `exec_remote`.

weave
```diff
@@ -40,6 +40,7 @@ IP_REGEXP="[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}"
 CIDR_REGEXP="$IP_REGEXP/[0-9]{1,2}"
 
 usage_no_exit() {
+    [ $# -gt 0 ] && echo "Error: $@"
```
rade: This is an unrelated change. And I don't think it's worth having.

site/proxy.md
inercia: Ups!

weave
inercia: Ok... :confused:

rade:
Please re-base on master, so the Vagrantfiles get reverted.

inercia:
Done. This branch has been re-based on master.

weave
```diff
@@ -1438,11 +1508,25 @@ case "$COMMAND" in
...
+            case "$1" in
+                --no-rewrite-hosts)
+                    REWRITE_HOSTS=
+                    ;;
+                *)
+                    break
+                    ;;
+            esac
+            shift
+        done
```
rade: This needs to happen before the `collect_cidr_args` call, otherwise that will not see CIDRS after a `--no-rewrite-hosts`. Also, there is no need for a loop.

I'd also defer the `dns_extra_hosts_args` call until we need it in the `-n "$REWRITE_HOSTS"` path (only) below.

weave
```diff
@@ -1478,6 +1564,13 @@ case "$COMMAND" in
                 --rewrite-hosts)
                     REWRITE_HOSTS=1
                     ;;
+                --add-host)
+                    DNS_EXTRA_HOSTS="$2 $DNS_EXTRA_HOSTS"
+                    shift
+                    ;;
+                --add-host=*)
+                    DNS_EXTRA_HOSTS="${1#*=} $DNS_EXTRA_HOSTS"
+                    ;;
```
rade: Instead of doing this here, it would be neater do simply invoke `dns_extra_hosts_args "$@"` in the `-n "$REWRITE_HOSTS` branch below.

weave
```diff
@@ -64,7 +64,7 @@ weave connect       [--replace] [<peer> ...]
 weave forget        <peer> ...
 weave status        [targets | connections | peers | dns]
 weave report        [-f <format>]
-weave run           [--with-dns | --without-dns] [<addr> ...]
+weave run           [--with-dns | --without-dns] [--no-rewrite-hosts] [<addr> ...]
```
rade: This is >80 chars wide now. So move the `[<addr> ...]` onto the next line.

inercia: Done.

weave
inercia: We cannot: there is a `[ $# -eq 1 ] || usage` at line 1580.

weave
inercia: Fixed.

weave
```diff
@@ -205,6 +205,40 @@ dns_arg_count() {
     fi
 }
 
+collect_rewrite_hosts_args() {
+    REWRITE_HOSTS_ARG_COUNT=0
+    case "$1" in
+    --rewrite-hosts)
+        REWRITE_HOSTS=1
+        REWRITE_HOSTS_ARG_COUNT=1
+        ;;
```
rade: huh? why do we have a `--rewrite-hosts` all over sudden?

Also, it would be simpler to inline this function at the call site.

inercia: > huh? why do we have a `--rewrite-hosts` all of a sudden?

Because I think it is more user-friendly: it does not make sense to recognize `--no-rewrite-hosts` but to fail with a `usage` when he/she uses `--rewrite-hosts`... I think we are doing the same with `--with-dns`/`--without-dns`

> Also, it would be simpler to inline this function at the call site.

Probably, but I think `collect_rewrite_hosts_args` creates a pattern where we do `collect_*_args` and `shift *_COUNT`, and we can use it in more parts of the code...

rade: I disagree on both points.

This adds a fair bit of complication to the code in this case. Please get rid of it.

inercia: It's being used [here](https://github.com/weaveworks/weave/commit/da9597d9d830ec92b0586e6ee2dc917d885bc241#diff-ed6579eaedd0e67b3ed51c7ee7071044R279) too.

rade: it's even easier to inline there since you can just change the conditional that follows.

weave
```diff
@@ -938,6 +964,50 @@ dns_args() {
...
+    HOSTS=$(echo $HOSTS_PATH_AND_NAME | cut -f1 -d" ")
+    NAME=$(echo $HOSTS_PATH_AND_NAME | cut -f2 -d" ")
+    CONTAINERS_PATH=$(dirname $HOSTS)
+    MNT=/container
+    MNT_HOSTS=$MNT/$(basename $HOSTS)
+    CONTENTS="$(default_etc_hosts $NAME $@)"
+    docker run --rm \
+        -v $CONTAINERS_PATH:$MNT \
+        --entrypoint=sh \
+        $EXEC_IMAGE -c "echo '$CONTENTS' > $MNT_HOSTS && rm -f $MNT_HOSTS && echo '$CONTENTS' > $MNT_HOSTS"
```
bboreham: Writing the file twice deserves a comment.

weave
```diff
@@ -244,10 +264,16 @@ elif [ "$1" = "run" -a -z "$IS_LOCAL" ] ; then
         DNS_ARGS=$(exec_remote dns-args "$@" || true)
     fi
     shift $(dns_arg_count "$@")
+    if [ "$1" = "--no-rewrite-hosts" ] ; then
```
bboreham: We are now fixing a particular ordering of `--with-dns` and `--no-rewrite-hosts` at the beginning of the argument list. In other cases we took care to allow arguments in any order.

weave
bboreham: That could be addressed with a `DNS_EXTRA_HOSTS_COUNT`, analogous to `CIDR_ARG_COUNT`

weave
```diff
@@ -207,6 +207,26 @@ dns_arg_count() {
     fi
 }
 
+dns_extra_hosts_args() {
```
bboreham: What does this function have to do with dns?

weave
```diff
@@ -938,6 +964,50 @@ dns_args() {
     fi
 }
 
+default_etc_hosts() {
```
bboreham: in what sense is this a 'default' ?

weave
```diff
@@ -938,6 +964,50 @@ dns_args() {
...
+    echo "# container hostname"
+    for CIDR in $ALL_CIDRS ; do
+        echo "${CIDR%/*}    $NAME"
+    done
+    echo
+    echo "# static names added with --add-host"
+    for EXTRA_HOST in "$@" ; do
+        NAME=$(echo $EXTRA_HOST | cut -f1 -d":")
+        IP=$(echo $EXTRA_HOST | cut -f2 -d":")
+        echo "$IP     $NAME"
```
bboreham: echo "${EXTRA_HOST#*:}     ${EXTRA_HOST%:*}"

(and remove the two `echo|cut` lines)

test/630_proxy_dns_test.sh
```diff
@@ -6,46 +6,39 @@ C1=10.2.0.78
...
-
-kill_containers() {
+STATIC=static.name
+STATIC_IP=10.9.9.9
+
+do_assert_resolution() {
+  proxy_start_container_with_dns $HOST1 -e WEAVE_CIDR=$C2/24 -dt --name=c2 --add-host $STATIC:$STATIC_IP -h $C2_NAME
+  proxy_start_container_with_dns $HOST1 -e WEAVE_CIDR=$C1/24 -dt --name=c1 --add-host $STATIC:$STATIC_IP
+  $1 $HOST1 c1 $C2_NAME $C2
+  assert_dns_record $HOST1 c1 $STATIC $STATIC_IP
```
bboreham: this (and the entire file) is now mis-named as you are looking for resolution via `/etc/hosts`

weave
```diff
@@ -938,6 +964,50 @@ dns_args() {
...
+ff02::2         ip6-allrouters
+EOF
+
+    echo "# created by Weave - END"
+}
+
+rewrite_etc_hosts() {
+    HOSTS_PATH_AND_NAME=$(docker inspect -f '{{.HostsPath}} {{.Config.Hostname}}' $CONTAINER)
+    HOSTS=$(echo $HOSTS_PATH_AND_NAME | cut -f1 -d" ")
+    NAME=$(echo $HOSTS_PATH_AND_NAME | cut -f2 -d" ")
```
bboreham: could reduce these three lines to

    eval $(docker inspect -f 'HOSTS={{.HostsPath}}; NAME={{.Config.Hostname}}' $CONTAINER)

rade: please don't.

weave
inercia: Yes, maybe it is not really related to DNS... I will rename it to `extra_hosts_args`

weave
rade: Not if we want to accept `--rewrite-hosts` and `--add-host` in any order.

bboreham: I wonder if `dns_extra_hosts_args` would be extended to handle `--rewrite-hosts`; this would simplify the `run` implementation too.  And then you might throw in `--or-die` too.

weave
inercia: Good point. I'll change it.

weave
inercia: Done.

weave
inercia: Renamed to `etc_hosts_contents`

test/630_proxy_dns_test.sh
rade: not just though; the test also attempts to resolve `$C2_NAME`, which will happen via DNS.

weave
inercia: Done.

weave
inercia: I think it is ok to fix the order for this. Otherwise we would have to do some tricks in order to parse the args for both the local `run`: 

    if [ "$1" != "--without-dns" ] ; then
        DNS_ARGS=$(exec_remote dns-args "$@" || true)
    fi
    shift $(dns_arg_count "$@")
    if [ "$1" = "--no-rewrite-hosts" ] ; then
        shift
    else
        extra_hosts_args "$@"
        DNS_EXTRA_HOSTS_ARGS="--rewrite-hosts $DNS_EXTRA_HOSTS_ARGS"
    fi

and the remote `run`

        dns_args "$@"
        shift $(dns_arg_count "$@")
        REWRITE_HOSTS=1
        if [ "$1" = "--no-rewrite-hosts" ] ; then
            REWRITE_HOSTS=
            shift
        fi

PR #1479, opened 2015-09-29
prevent ConnectionMaker.ConnectionTerminated race

### LINKED ISSUE #1478: race can lead to multiple concurrent connection attempts to same target
Apply
```patch
diff --git a/router/connection.go b/router/connection.go
index e3967ec..810c419 100644
--- a/router/connection.go
+++ b/router/connection.go
@@ -360,6 +360,7 @@ func (conn *LocalConnection) shutdown(err error) {
                conn.forwarder.Stop()
        }
 
+       time.Sleep(1 * time.Minute)
        conn.Router.ConnectionMaker.ConnectionTerminated(conn.remoteTCPAddr, err
 }
```

and then
```
host1$ weave launch
host2$ weave launch host1
host1$ weave stop
host2$ weave status connections
-> 192.168.48.11:6783    failed      dial tcp4 192.168.48.11:6783: connection refused, retry: 2015-09-29 13:53:41.831462406 +0000 UTC
host2$ sleep 60
host2$ weave status connections
-> 192.168.48.11:6783    failed      read tcp4 192.168.48.11:6783: connection reset by peer, retry: 2015-09-29 13:55:09.21609408 +0000 UTC
```

This shows an incorrect change of the reported error - the "connection reset by peer" comes from the original failure when we stopped the other peer and stomps on the subsequent "connection refused" error.

However, with slightly different timing, namely the `ConnectionTerminated` occurring when there is an ongoing connection attempt, we can get multiple concurrent connection attempts to the same target. This breaks a pretty important invariant of the `ConnectionMaker`, so there may be other bad effects.

rade:
Fixes #1478.

### DIFF
--- a/router/connection.go
+++ b/router/connection.go
@@ -319,6 +319,7 @@ func (conn *LocalConnection) run(actionChan <-chan ConnectionAction, finished ch
 	if err = conn.initHeartbeats(); err != nil {
 		return
 	}
+	conn.Router.ConnectionMaker.ConnectionCreated(conn)
 
 	go conn.receiveTCP(intro.Receiver)
 	err = conn.actorLoop(actionChan)
@@ -503,7 +504,7 @@ func (conn *LocalConnection) shutdown(err error) {
 	// try to send any more
 	conn.stopForwarders()
 
-	conn.Router.ConnectionMaker.ConnectionTerminated(conn.remoteTCPAddr, err)
+	conn.Router.ConnectionMaker.ConnectionTerminated(conn, err)
 }
 
 // Helpers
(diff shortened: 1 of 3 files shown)


rade:
This doesn't actually fully solve the problem since `ConnectionTerminated` is async.

rade:
I wonder whether the benefit/risk ratio is too low for this to go into 1.1.

bboreham:
I don't think the benefit is high, as a duplicate connection will be torn down, and if a stale error kills a good connection it will re-try shortly.

What is the risk?  Did you see this happen for real?

rade:
> What is the risk?

Simply the complexity of the change, which is in an area that traditionally has seen quite a few problems.

> Did you see this happen for real?

No.

rade:
There's some more tidying up of status reporting that I want to do... -> re-assignig to myself.

rade:
I am going to re-base this on master. closing.

Issue #1475, opened 2015-09-28
Weave 1.0.3 crash during AWS Outage


colebrumley:
During that AWS us-east outage on Aug 25, Weave v1.0.3 running on CoreOS crashed with the following error (full trace [here](https://gist.github.com/colebrumley/31a22a11eb5e4c3e6ca9)):

```
panic: runtime error: growslice: cap out of range [recovered]
	panic: runtime error: growslice: cap out of range

goroutine 4780681 [running]:
encoding/gob.catchError(0xc2328b3350)
	/usr/local/go/src/encoding/gob/error.go:38 +0x96
encoding/gob.encUint8Array(0xc2092d3e60, 0xc24016a600, 0x892760, 0xc21cf60fc0, 0x57)
	/usr/local/go/src/encoding/gob/encode.go:238 +0x17b
encoding/gob.(*Encoder).encodeSingle(0xc2328b32c0, 0xc2328b32f8, 0xc2092db960, 0x892760, 0xc21cf60fc0, 0x57)
	/usr/local/go/src/encoding/gob/encode.go:294 +0x17b
encoding/gob.(*Encoder).encode(0xc2328b32c0, 0xc2328b32f8, 0x892760, 0xc21cf60fc0, 0x57, 0xc208036280)
	/usr/local/go/src/encoding/gob/encode.go:694 +0x20f
encoding/gob.(*Encoder).EncodeValue(0xc2328b32c0, 0x892760, 0xc21cf60fc0, 0x57, 0x0, 0x0)
	/usr/local/go/src/encoding/gob/encoder.go:241 +0x613
encoding/gob.(*Encoder).Encode(0xc2328b32c0, 0x892760, 0xc21cf60fc0, 0x0, 0x0)
	/usr/local/go/src/encoding/gob/encoder.go:168 +0x7f
github.com/weaveworks/weave/router.GobEncode(0xc209461e88, 0x3, 0x3, 0x0, 0x0, 0x0)
	/home/matthias/go/src/github.com/weaveworks/weave/router/utils.go:80 +0x130
github.com/weaveworks/weave/router.func·016(0x7fbcd81a1198, 0xc21cf60ba0)
	/home/matthias/go/src/github.com/weaveworks/weave/router/gossip_channel.go:151 +0x290
github.com/weaveworks/weave/router.(*GossipSender).run(0xc2341d2d20)
	/home/matthias/go/src/github.com/weaveworks/weave/router/gossip.go:59 +0xf6
[...]

Issue #1450, opened 2015-09-18
"growslice: cap out of range" panic in gob encoding


MaximeHeckel:
While trying the `-initpeercount` flag (set to a wrong number to see the effects) one of the weave 1.0.3 routers crashed.

Here are the logs:

```
weave 2015/09/18 16:19:16.475139 ->[192.81.217.207:6783] attempting connection
weave 2015/09/18 16:19:16.477903 ->[192.81.217.207:6783] error during connection attempt: dial tcp4 192.81.217.207:6783: connection refused
weave 2015/09/18 16:19:16.532717 ->[192.241.130.141:6783] attempting connection
weave 2015/09/18 16:19:16.543436 ->[192.241.130.141:6783] error during connection attempt: dial tcp4 192.241.130.141:6783: connection refused
weave 2015/09/18 16:19:18.896839 ->[192.81.217.207:6783] attempting connection
weave 2015/09/18 16:19:18.899240 ->[192.81.217.207:6783] error during connection attempt: dial tcp4 192.81.217.207:6783: connection refused
weave 2015/09/18 16:19:19.505453 ->[192.241.130.141:6783] attempting connection
weave 2015/09/18 16:19:19.509241 ->[192.241.130.141:6783] error during connection attempt: dial tcp4 192.241.130.141:6783: connection refused
panic: runtime error: growslice: cap out of range [recovered]
	panic: runtime error: growslice: cap out of range

goroutine 44 [running]:
encoding/gob.catchError(0xc209bb4630)
	/usr/local/go/src/encoding/gob/error.go:38 +0x96
encoding/gob.encUint8Array(0xc20917dd40, 0xc209b99e00, 0x89b100, 0xc209bc0120, 0x57)
	/usr/local/go/src/encoding/gob/encode.go:238 +0x17b
encoding/gob.(*Encoder).encodeSingle(0xc209bb45a0, 0xc209bb45d8, 0xc209173200, 0x89b100, 0xc209bc0120, 0x57)
	/usr/local/go/src/encoding/gob/encode.go:294 +0x17b
encoding/gob.(*Encoder).encode(0xc209bb45a0, 0xc209bb45d8, 0x89b100, 0xc209bc0120, 0x57, 0xc208034280)
	/usr/local/go/src/encoding/gob/encode.go:694 +0x20f
encoding/gob.(*Encoder).EncodeValue(0xc209bb45a0, 0x89b100, 0xc209bc0120, 0x57, 0x0, 0x0)
[...]

PR #1484, opened 2015-09-30
Sanity-check gossip messages

bboreham:
In #1450 and #1475 we have an error from the Go runtime suggesting that a buffer grew beyond 128GB.  This check will catch any infeasible gossip message being passed through the route seen on those stack traces, so we can narrow down the cause if it happens again.

Choice of limit here is arbitrary; it is not expected that an ordinary user of weave would see even 1MB gossip messages.


bboreham:
Wrong branch

PR #1490, opened 2015-09-30
Output the docker logs when docker-py fails

paulbellamy:
We could also try logging into one of the test boxes and running the test repeatedly there, then grabbing logs.

To help catch https://circleci.com/gh/weaveworks/weave/3037

@tomwilkie and/or @rade, please review.

### DIFF
--- a/test/600_proxy_docker_py.sh
+++ b/test/600_proxy_docker_py.sh
@@ -36,6 +36,9 @@ docker_py_test() {
         assert_raises "true"
     else
         assert_raises "false"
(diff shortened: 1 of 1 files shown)


rade:
A dozen runs, and none of them have triggered the failure we are chasing :(

I suggest we rebase this on 1.1 so it gets run on more occasions.

You did test this, right? i.e. it does produce some sensible output on failure?

paulbellamy:
Running agains 1.1 on: https://github.com/weaveworks/weave/pull/1491

PR #1491, opened 2015-10-01
output the docker logs when docker-py fails (1.1)

paulbellamy:


paulbellamy:
I have seen this branch fail on my dev VM, so it is not unique to circleCI.

rade:
> I have seen this branch fail on my dev VM

i.e. it failed with the error we are chasing?

rade:
We've seen this [fail in CI](https://circleci.com/gh/weaveworks/weave/3076) now too, but the docker log output sheds no light :( If the output isn't useful then there is no point to this PR.

paulbellamy:
Yeah, it failed in the same test (different assertion).

PR #1496, opened 2015-10-01
backoff late connection failures

### LINKED ISSUE #953: Weave can re-try late-failing connections too quickly
Example:

````
[run tests/_cross_hosts_test.sh]
$ weave launch 192.168.48.11
$ docker logs weave
...
weave 2015/06/17 13:18:22.518312 ->[192.168.48.11:6783] attempting connection
weave 2015/06/17 13:18:22.519850 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: completed handshake; using protocol version 1
weave 2015/06/17 13:18:22.520194 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: connection added (new peer)
weave 2015/06/17 13:18:22.559268 ->[192.168.48.12:6783] attempting connection
weave 2015/06/17 13:18:22.559888 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: connection shutting down due to error: IP Allocator with different subnet detected
weave 2015/06/17 13:18:22.560117 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: connection deleted
weave 2015/06/17 13:18:22.560226 Removed unreachable peer e2:4d:c4:b9:c8:ce(host1)
weave 2015/06/17 13:18:22.560320 Removed unreachable peer b6:39:47:1d:24:0a(host2)
weave 2015/06/17 13:18:22.560363 ->[192.168.48.12:6783] connection shutting down due to error during handshake: Found unknown remote name: b6:39:47:1d:24:0a at 192.168.48.12:6783
weave 2015/06/17 13:18:22.560459 ->[192.168.48.11:6783] attempting connection
weave 2015/06/17 13:18:22.563807 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: completed handshake; using protocol version 1
weave 2015/06/17 13:18:22.563932 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: connection added (new peer)
weave 2015/06/17 13:18:22.564742 ->[192.168.48.12:6783] attempting connection
weave 2015/06/17 13:18:22.565402 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: connection shutting down due to error: IP Allocator with different subnet detected
weave 2015/06/17 13:18:22.565566 ->[192.168.48.11:6783|e2:4d:c4:b9:c8:ce(host1)]: connection deleted
weave 2015/06/17 13:18:22.565588 Removed unreachable peer e2:4d:c4:b9:c8:ce(host1)
weave 2015/06/17 13:18:22.565594 Removed unreachable peer b6:39:47:1d:24:0a(host2)
weave 2015/06/17 13:18:22.565674 ->[192.168.48.11:6783] attempting connection
````

rade:
If a connection gets through all the handshake and subsequent checks - i.e. it ends up getting added to the LocalPeer's connection list - but gets terminated within one minute (of the start of the successful connection attempt), then we apply the existing backoff logic to schedule a retry, instead of retrying immediately.

In order to do this we retain the 'Target' data structure for such connections, marking it as 'connected' instead of deleting it (as we did previously).

Fixes #953.

### DIFF
--- a/router/connection_maker.go
+++ b/router/connection_maker.go
@@ -26,10 +26,17 @@ type ConnectionMaker struct {
 	actionChan  chan<- ConnectionMakerAction
 }
 
+type TargetState int
+
+const (
+	TargetWaiting    TargetState = iota // we are waiting to connect there
+	TargetAttempting                    // we are attempting to connect there
+	TargetConnected                     // we are connected to there
+)
+
 // Information about an address where we may find a peer
 type Target struct {
-	attempting  bool          // are we currently attempting to connect there?
-	connected   bool          // are we currently connected to there?
+	state       TargetState
 	lastError   error         // reason for disconnection last time
 	tryAfter    time.Time     // next time to try this address
 	tryInterval time.Duration // retry delay on next failure
@@ -95,7 +102,7 @@ func (cm *ConnectionMaker) ForgetConnections(peers []string) {
 func (cm *ConnectionMaker) ConnectionAborted(address string, err error) {
 	cm.actionChan <- func() bool {
 		target := cm.targets[address]
-		target.attempting = false
+		target.state = TargetWaiting
 		target.lastError = err
 		target.retry()
 		return true
@@ -107,8 +114,7 @@ func (cm *ConnectionMaker) ConnectionCreated(conn Connection) {
 		cm.connections[conn] = void
 		if conn.Outbound() {
 			target := cm.targets[conn.RemoteTCPAddr()]
-			target.attempting = false
-			target.connected = true
+			target.state = TargetConnected
 		}
 		return false
 	}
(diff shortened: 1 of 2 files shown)


rade:
Items for discussion:
- the `ResetAfter` interval, currently set to 1 minute, which determines how long a connection must stay alive for it to be retried immediately on subsequent failure instead of being subject to backoff. Note that this period starts at the time the successful connection attempt was scheduled for. This is easiest in the implementation, but means the period includes a) any delays in CM to attempt the connection, b) the time it takes perform the handshake and other checks.
- the three refactoring commits; they are in a dependency chain but I'd be happy to lop some off the end
- smoke test: I could add one, but it would involve some sleeps and poking the weave status, which is rather distasteful. Note that code coverage of the new code is already excellent.

rade:
>  an extra three lines of coverage?

Actually the behaviour we want to test for - that connections which fail shortly after establishment don't trigger an immediate reconnect - doesn't expand code coverage; the three uncovered lines are for the case that the connection failed later. And we probably don't want to wait that long, depending on what `ResetAfter` interval we settle on.

And even writing a test for the former is tricky; the difference in connection status is that previously we'd cycle through `retrying`, `pending`, `established`, whereas now it's `failed`, `retying`, `pending`, `established`. The duration of the `failed` state will be quite short to start with, and expand as the backoff kicks in. I cannot think of a simple way to write a test here that wouldn't be prone to all sorts of timing-related failures.

PR #1503, opened 2015-10-03
don't allocate IPs on 'detach' and 'hide'

### LINKED ISSUE #860: `weave hide/detach` allocates a new address if called with no preceding `expose/attach`
This is a conscious choice, to make the script simpler, but is strictly unnecessary, and could lead to failure cases, e.g. if ipam is out of addresses.

[note that, as of current state of #827, the same will apply to `detach` with no preceding `attach`]

rade:
Instead of POSTing to IPAM in order to obtain IPs, we GET existing IPs. We ignore 404s, so detaching/hiding remains idempotent.

Fixes #860.

### DIFF
--- a/weave
+++ b/weave
@@ -888,20 +888,33 @@ ipam_reclaim() {
     done
 }
 
-# Call IPAM as necessary to allocate addresses
-# $1 is the full container id and following args are previously parsed CIDR_ARGS.
-# Returns ALL_CIDRS and IPAM_CIDRS
+# Call IPAM as necessary to lookup or allocate addresses
+#
+# $1 is one of 'lookup', 'allocate' or 'allocate_no_check_alive', $2
+# is the full container id. The remaining args are previously parsed
+# CIDR_ARGS.
+#
+# Populates ALL_CIDRS and IPAM_CIDRS
 ipam_cidrs() {
-    CHECK_ALIVE="?check-alive=true"
-    if [ "$1" = "--no-check-alive" ] ; then
-        CHECK_ALIVE=
-        shift 1
-    fi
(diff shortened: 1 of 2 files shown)


weave
```diff
@@ -888,20 +888,33 @@ ipam_reclaim() {
...
+    case $1 in
+        lookup)
+            METHOD=GET
+            CHECK_ALIVE=
+            ;;
+        allocate)
+            METHOD=POST
+            CHECK_ALIVE="?check-alive=true"
+            ;;
+        allocate_unchecked)
```
bboreham: `unchecked` suggests "no check for error" to me.  How about `allocate_no_check_alive` ?

weave
```diff
@@ -910,21 +923,26 @@ ipam_cidrs() {
...
             fi
-            IPAM_CIDRS="$IPAM_CIDRS $CIDR"
-            ALL_CIDRS="$ALL_CIDRS $CIDR"
         else
-            # This is a plain IP address; warn if it clashes but carry on
-            command_exists netcheck && netcheck --ignore-iface=$BRIDGE $arg || true
-            if container_ip $CONTAINER_NAME 2>/dev/null ; then
-                http_call $CONTAINER_IP:$HTTP_PORT PUT /ip/$CONTAINER_ID/${arg%/*} >&2
+            if [ "$METHOD" = "POST" ] ; then
+                # Allocation of a plain IP address; warn if it clashes but carry on
```
bboreham: This is not an _allocation_; the user has passed in a specific IP address.    "assignment" ?

bboreham:
LGTM apart from comments on wording.

PR #1504, opened 2015-10-04
strip off @... from interface names returned by `ip link show`

### LINKED ISSUE #1364: `weave reset` fails with "Cannot find device" on kernel 4.2+
Extracted from #1356...

@MaximeHeckel has seen `weave --local reset` fail with `Cannot find device "vethwepl14191@if322"`. This can happen when a container dies during reset, which is what #1356 is about, but it could also happen if somehow our interface name extraction is broken - that `@if322` in the interface name looks weird.

@MaximeHeckel as @bboreham said, please get us the output of `ip link show` from the host. Also what OS and kernel is that host running? And what version of iptables (`ip -V`)?

rade:
Apparently kernel 4.2 adds these, which breaks `weave reset` when we pass the names to `ip link del'.

Fixes #1364.

### DIFF
--- a/weave
+++ b/weave
@@ -1605,7 +1605,7 @@ case "$COMMAND" in
         conntrack -D -p udp --dport $PORT >/dev/null 2>&1 || true
         destroy_bridge
         for LOCAL_IFNAME in $(ip link show | grep v${CONTAINER_IFNAME}pl | cut -d ' ' -f 2 | tr -d ':') ; do
-            ip link del $LOCAL_IFNAME >/dev/null 2>&1 || true
+            ip link del ${LOCAL_IFNAME%@*} >/dev/null 2>&1 || true
         done
         ;;
     rmpeer)


rade:
@dpw mind taking a look at this? I have not been able to test this since I failed miserably in my attempt to create an Ubuntu 15.10 vagrant VM.

dpw:
Ok, tested with the 15.10 daily.  With master, `weave --local reset` leaves `vethweplNNNN` netdevs for containers on the weave network around.  With this branch, they get cleaned up.

rade:
You also got an "Cannot find device" error on master, right?

dpw:
> You also got an "Cannot find device" error on master, right?

Not visibly: That `ip link del` line you modified has an `>/dev/null 2>&1` on it.  If I try it by hand I see that error.

rade:
Hmm. That's odd, given the user report in #1356.

rade:
Ah, the `>/dev/null 2>&1` got added in 1.0.3.

dpw:
> Hmm. That's odd, given the user report in #1356.

I'd guess they were running a version prior to https://github.com/weaveworks/weave/commit/1df752

rade:
Yes, #1365 was the PR that closed #1356.

PR #1512, opened 2015-10-07
don't stall in ipam on early container death

### LINKED ISSUE #1459: Stall while allocating address for dead container
As noted in https://github.com/weaveworks/weave/pull/1426#issuecomment-142312227, if IPAM is waiting for consensus then an allocation request for a container that has already died will stall.

Currently IPAM checks if the container is alive _after_ allocating an IP address, to avoid the leak if we check, then it dies, then we allocate.

rade:
When a container died before we got into the allocation code (inside the allocator run loop) we would not cancel the allocation since the container death event will have been processed by the allocator before it had any knowledge of the container.

The fix is to move the container-aliveness check inside the allocation code. That way we still detect the death even if the event has already flown past.

We only need to perform the check once, since any subsequent death will result in an event that is processed by the allocator when it has recorded the container in its state.

Fixes #1459.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -175,17 +175,6 @@ func (alloc *Allocator) spaceRequestDenied(sender router.PeerName, r address.Ran
 	}
 }
 
-func hasBeenCancelled(cancelChan <-chan bool) func() bool {
-	return func() bool {
-		select {
-		case <-cancelChan:
-			return true
-		default:
-			return false
-		}
-	}
-}
-
 type errorCancelled struct {
 	kind  string
 	ident string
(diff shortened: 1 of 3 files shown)


ipam/allocator.go
```diff
@@ -199,10 +199,9 @@ func (e *errorCancelled) Error() string {
 
 // Allocate (Sync) - get new IP address for container with given name in range
 // if there isn't any space in that range we block indefinitely
-func (alloc *Allocator) Allocate(ident string, r address.Range, cancelChan <-chan bool) (address.Address, error) {
+func (alloc *Allocator) Allocate(ident string, r address.Range, hasBeenCancelled func() bool) (address.Address, error) {
 	resultChan := make(chan allocateResult)
-	op := &allocate{resultChan: resultChan, ident: ident, r: r,
-		hasBeenCancelled: hasBeenCancelled(cancelChan)}
+	op := &allocate{resultChan: resultChan, ident: ident, r: r, hasBeenCancelled: hasBeenCancelled}
```
bboreham: should be able to remove the `hasBeenCancelled()` function implementation in this file now

ipam/http.go
```diff
@@ -31,7 +31,18 @@ func parseCIDR(w http.ResponseWriter, cidrStr string) (address.CIDR, bool) {
 
 func (alloc *Allocator) handleHTTPAllocate(dockerCli *docker.Client, w http.ResponseWriter, ident string, checkAlive bool, subnet address.CIDR) {
 	closedChan := w.(http.CloseNotifier).CloseNotify()
-	addr, err := alloc.Allocate(ident, subnet.HostRange(), closedChan)
+	addr, err := alloc.Allocate(ident, subnet.HostRange(),
+		func() bool {
+			select {
+			case <-closedChan:
+				common.Log.Infoln("closed")
```
bboreham: bit cryptic.  Does the user get enough information to see what happened?

PR #1513, opened 2015-10-07
Fast datapath weave status

### LINKED ISSUE #1505: Include fast datapath information in `weave status|report`
For the purposes of support and debugging:
* `weave status connections` should display which forwarder is use per connection
* `weave report` should include the flow rules currently in effect and peer short IDs

awh:
Fixes #1505.

````
$ weave status connections
<- 192.168.48.12:39316   established vxlan        d2:80:69:fe:6a:b5(host2)
-> 192.168.48.15:6783    connecting  undetermined 
````

````
$ weave report
{
    "Version": "git-819b0cbccb46",
    "Router": {
        "Protocol": "weave",
        "ProtocolMinVersion": 1,
        "ProtocolMaxVersion": 2,
        "Encryption": false,
        "PeerDiscovery": true,
        "Name": "2a:0e:71:cf:b5:e1",
        "NickName": "host1",
        "Port": 6783,
        "Interface": "weave (via ODP)",
        "CaptureStats": null,
        "MACs": [
            {
                "Mac": "82:72:04:78:39:7c",
                "Name": "2a:0e:71:cf:b5:e1",
                "NickName": "host1",
                "LastSeen": "2015-10-07T17:03:55.129115227Z"
            },
            {
                "Mac": "2a:30:c2:41:98:f3",
                "Name": "d2:80:69:fe:6a:b5",
                "NickName": "host2",
                "LastSeen": "2015-10-07T17:03:55.129053768Z"
            }
        ],
        "Peers": [
            {
                "Name": "2a:0e:71:cf:b5:e1",
                "NickName": "host1",
                "UID": 13490210014865600209,
                "Version": 2,
                "Connections": [
                    {
                        "Name": "d2:80:69:fe:6a:b5",
                        "NickName": "host2",
                        "Address": "192.168.48.12:39316",
                        "Outbound": false,
                        "Established": true
                    }
                ]
            },
            {
                "Name": "d2:80:69:fe:6a:b5",
                "NickName": "host2",
                "UID": 14328624370048807465,
                "Version": 2,
                "Connections": [
                    {
                        "Name": "2a:0e:71:cf:b5:e1",
                        "NickName": "host1",
                        "Address": "192.168.48.11:6783",
                        "Outbound": true,
                        "Established": true
                    }
                ]
            }
        ],
        "UnicastRoutes": [
            {
                "Dest": "2a:0e:71:cf:b5:e1",
                "Via": "00:00:00:00:00:00"
            },
            {
                "Dest": "d2:80:69:fe:6a:b5",
                "Via": "d2:80:69:fe:6a:b5"
            }
        ],
        "BroadcastRoutes": [
            {
                "Source": "d2:80:69:fe:6a:b5",
                "Via": null
            },
            {
                "Source": "2a:0e:71:cf:b5:e1",
                "Via": [
                    "d2:80:69:fe:6a:b5"
                ]
            }
        ],
        "Connections": [
            {
                "Address": "192.168.48.12:39316",
                "Outbound": false,
                "State": "established",
                "Info": "d2:80:69:fe:6a:b5(host2)",
                "OverlayType": "vxlan"
            },
            {
                "Address": "192.168.48.15:6783",
                "Outbound": true,
                "State": "failed",
                "Info": "dial tcp4 192.168.48.15:6783: no route to host, retry: 2015-10-07 17:08:39.096516208 +0000 UTC",
                "OverlayType": "undetermined"
            }
        ],
        "Targets": [
            "192.168.48.15"
        ],
        "OverlayDiagnostics": {
            "fastdp": {
                "Vports": [
                    {
                        "ID": 0,
                        "Name": "weave",
                        "TypeName": "internal"
                    },
                    {
                        "ID": 1,
                        "Name": "vxlan",
                        "TypeName": "vxlan"
                    },
                    {
                        "ID": 2,
                        "Name": "vethwepl7973",
                        "TypeName": "netdev"
                    }
                ],
                "Flows": [
                    {
                        "FlowKeys": [
                            "TunnelFlowKey(id: 0000000000ac8f92, ipv4src: 192.168.48.12, ipv4dst: 192.168.48.11}",
                            "InPortFlowKey{vport: 1}",
                            "EthernetFlowKey{src: 2a:30:c2:41:98:f3, dst: 82:72:04:78:39:7c}"
                        ],
                        "Actions": [
                            "OutputAction{vport: 2}"
                        ],
                        "Packets": 343,
                        "Bytes": 30142,
                        "Used": 28205473
                    },
                    {
                        "FlowKeys": [
                            "EthernetFlowKey{src: 82:72:04:78:39:7c, dst: 2a:30:c2:41:98:f3}",
                            "InPortFlowKey{vport: 2}"
                        ],
                        "Actions": [
                            "SetTunnelAction{id: 0000000000f92ac8, ipv4src: 192.168.48.11, ipv4dst: 192.168.48.12, ttl: 64, df: true}",
                            "OutputAction{vport: 1}"
                        ],
                        "Packets": 343,
                        "Bytes": 30142,
                        "Used": 28205473
                    }
                ]
            },
            "sleeve": null
        }
    },
    "IPAM": {
        "Paxos": null,
        "Range": "[10.32.0.0-10.48.0.0)",
        "DefaultSubnet": "10.32.0.0/12",
        "Entries": [
            {
                "Token": "10.32.0.0",
                "Peer": "2a:0e:71:cf:b5:e1",
                "Version": 3
            },
            {
                "Token": "10.40.0.0",
                "Peer": "d2:80:69:fe:6a:b5",
                "Version": 1
            },
            {
                "Token": "10.47.255.255",
                "Peer": "2a:0e:71:cf:b5:e1",
                "Version": 0
            }
        ],
        "PendingClaims": null,
        "PendingAllocates": null
    },
    "DNS": {
        "Domain": "weave.local.",
        "Address": "172.17.42.1:53",
        "TTL": 1,
        "Entries": [
            {
                "Hostname": "bar.weave.local.",
                "Origin": "d2:80:69:fe:6a:b5",
                "ContainerID": "9deba5cce8a350166622c9a99012789c4ca1e82e25bb90111a51cd3d385625d9",
                "Address": "10.40.0.0",
                "Version": 0,
                "Tombstone": 0
            },
            {
                "Hostname": "foo.weave.local.",
                "Origin": "2a:0e:71:cf:b5:e1",
                "ContainerID": "d12c6ab22bcaf3496deb673cb4d4c8d6ad3c42ab8429b3657d53d2ffc7e1b975",
                "Address": "10.32.0.1",
                "Version": 0,
                "Tombstone": 0
            }
        ]
    }
}
````

### DIFF
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -161,7 +161,12 @@ func HandleHTTP(muxRouter *mux.Router, version string, router *weave.Router, all
 	}
 	muxRouter.Methods("GET").Path("/report").Headers("Accept", "application/json").HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			json, _ := json.MarshalIndent(status(), "", "    ")
+			json, err := json.MarshalIndent(status(), "", "    ")
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				Log.Error("Error during report marshalling: ", err)
+				return
+			}
 			w.Header().Set("Content-Type", "application/json")
 			w.Write(json)
 		})
--- a/router/status.go
+++ b/router/status.go
@@ -1,6 +1,7 @@
 package router
 
 import (
+	"fmt"
 	"time"
 )
 
@@ -170,7 +171,9 @@ func NewLocalConnectionStatusSlice(cm *ConnectionMaker) []LocalConnectionStatus 
 			if conn.Established() {
 				state = "established"
 			}
-			slice = append(slice, LocalConnectionStatus{conn.RemoteTCPAddr(), conn.Outbound(), state, conn.Remote().String()})
+			lc, _ := conn.(*LocalConnection)
+			info := fmt.Sprintf("%-6v %v", lc.forwarder.OverlayType(), conn.Remote())
+			slice = append(slice, LocalConnectionStatus{conn.RemoteTCPAddr(), conn.Outbound(), state, info})
 		}
 		for address, target := range cm.targets {
 			add := func(state, info string) {
--- a/router/status.go
+++ b/router/status.go
@@ -36,6 +36,7 @@ type PeerStatus struct {
 	Name        string
 	NickName    string
 	UID         PeerUID
+	ShortID     PeerShortID
 	Version     uint64
 	Connections []ConnectionStatus
 }
@@ -122,6 +123,7 @@ func NewPeerStatusSlice(peers *Peers) []PeerStatus {
 			peer.Name.String(),
 			peer.NickName,
 			peer.UID,
+			peer.ShortID,
 			peer.Version,
 			connections})
 	})
--- a/router/status.go
+++ b/router/status.go
@@ -22,6 +22,7 @@ type Status struct {
 	BroadcastRoutes    []BroadcastRouteStatus
 	Connections        []LocalConnectionStatus
 	Targets            []string
+	OverlayDiagnostics interface{}
 }
 
 type MACStatus struct {
@@ -80,7 +81,8 @@ func NewStatus(router *Router) *Status {
 		NewUnicastRouteStatusSlice(router.Routes),
 		NewBroadcastRouteStatusSlice(router.Routes),
 		NewLocalConnectionStatusSlice(router.ConnectionMaker),
-		NewTargetSlice(router.ConnectionMaker)}
+		NewTargetSlice(router.ConnectionMaker),
+		router.Overlay.Diagnostics()}
 }
 
 func NewMACStatusSlice(cache *MacCache) []MACStatus {
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -2,6 +2,7 @@ package router
 
 import (
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"net"
 	"sync"
@@ -319,6 +320,77 @@ func (fastdp fastDatapathOverlay) InvalidateShortIDs() {
 func (fastDatapathOverlay) AddFeaturesTo(features map[string]string) {
 	// Nothing needed.  Fast datapath support is indicated through
 	// OverlaySwitch.
+}
+
+type FlowStatus odp.FlowInfo
+
+func (flowStatus *FlowStatus) MarshalJSON() ([]byte, error) {
+	type jsonFlowStatus struct {
+		FlowKeys []string
+		Actions  []string
+		Packets  uint64
+		Bytes    uint64
+		Used     uint64
+	}
+
+	flowKeys := make([]string, 0, len(flowStatus.FlowKeys))
+	for _, flowKey := range flowStatus.FlowKeys {
+		if !flowKey.Ignored() {
+			flowKeys = append(flowKeys, fmt.Sprint(flowKey))
+		}
+	}
+
+	actions := make([]string, 0, len(flowStatus.Actions))
+	for _, action := range flowStatus.Actions {
+		actions = append(actions, fmt.Sprint(action))
+	}
+
+	return json.Marshal(&jsonFlowStatus{flowKeys, actions,
+		flowStatus.Packets, flowStatus.Bytes, flowStatus.Used})
+}
+
+type VportStatus odp.Vport
+
+func (vport *VportStatus) MarshalJSON() ([]byte, error) {
+	type jsonVportStatus struct {
+		ID       odp.VportID
+		Name     string
+		TypeName string
+	}
+
+	return json.Marshal(&jsonVportStatus{vport.ID, vport.Spec.Name(),
+		vport.Spec.TypeName()})
+}
+
+func (fastdp fastDatapathOverlay) Diagnostics() interface{} {
+	lock := fastdp.startLock()
+	defer lock.unlock()
+
+	vports, err := fastdp.dp.EnumerateVports()
+	if err != nil {
+		log.Warn(err)
+	}
+	vportStatuses := make([]VportStatus, 0, len(vports))
+	for _, vport := range vports {
+		vportStatuses = append(vportStatuses, VportStatus(vport))
+	}
+
+	flows, err := fastdp.dp.EnumerateFlows()
+	if err != nil {
+		log.Warn(err)
+	}
+	flowStatuses := make([]FlowStatus, 0, len(flows))
+	for _, flow := range flows {
+		flowStatuses = append(flowStatuses, FlowStatus(flow))
+	}
+
+	return struct {
+		Vports []VportStatus
+		Flows  []FlowStatus
+	}{
+		vportStatuses,
+		flowStatuses,
+	}
 }
 
 func (fastdp fastDatapathOverlay) StartConsumingPackets(localPeer *Peer,
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -662,6 +662,10 @@ func (fwd *fastDatapathForwarder) ControlMessage(tag byte, msg []byte) {
 	}
 }
 
+func (fwd *fastDatapathForwarder) OverlayType() string {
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by awh** [comment]
Requires weaveworks/go-odp#3.

**Comment by dpw** [comment]
The `weave report` output looks good.  But I'm not sure about putting that JSON rendering code into go-odp . Is there anything preventing moving it into weave?

**Comment by rade** [comment]
We should try to achieve greater separation between the control and data planes. This change goes in the wrong direction for that: `OverlayType` shouldn't be on `Connection`. I reckon it shouldn't even be on `LocalConnection`. Perhaps have a more generic `Info` method (see also next point), or make the status reporter drill through to the forwarder instead.

I am not keen on the extra status field and associated column, since it is essentially modal - it's only populated for pending/established connections. I'd add the overlay type to the `Info` field, which is how we deal with other modal info.

Surely the names of overlay types should match the keys in OverlayDiagnostics - in the example we have "vxlan" vs "fastdp".

**Comment by rade** [comment]
Where do short ids show up in this?

**Comment by awh** [comment]
@dpw
> The weave report output looks good.

:+1: Initially I looked at writing JSON marshallers for all the different concrete key and actions types, but then realised it was much easier to leverage your existing `String()` implementations as there's no need for the individual attributes to be modelled in JSON.

> I'm not sure about putting that JSON rendering code into go-odp . Is there anything preventing moving it into weave?

Those method declarations have to be in the same package as the types they're bound to - I'll see if I can declare them in the weave project but in the odp package. Another Golang bugbear - it would be really nice if you could have different marshallers for a given type in scope for different use cases, but the static binding based on package prevents you from doing that. Nor can you provide context to the marshallers without resorting to global state (I wanted to resolve the `InPortFlowKey` vport ID to a name rather than listing all the vports separately, but you can't do that either)

@rade
> We should try to achieve greater separation between the control and data planes.

Is this in preparation for something?

> Perhaps have a more generic Info method (see also next point), or make the status reporter drill through to the forwarder instead.

I actually started out by trying to get at the forwarder, but it's a bit ugly; OTOH there are other parts of `status.go` that are of necessity tied to router internals so I suppose it's not so bad.

>  I'd add the overlay type to the Info field

OK.

> Surely the names of overlay types should match the keys in OverlayDiagnostics - in the example we have "vxlan" vs "fastdp".

Good point, I'll change it.

> why is this suddenly necessary?

It was only necessary whilst developing the marshaller implementations - I can take it out if you like, but it's doing no harm and may be useful again in future if we extend the report again.

> Where do short ids show up in this?

They don't! I'll update the issue and add them in.

**Comment by awh** [comment]
>  I'll see if I can declare them in the weave project but in the odp package

This doesn't work. I tried adding `odp/json.go` with `package odp` and added the directory as a dependency to the `$(WEAVER_EXE)` target in the makefile but it doesn't even get compiled (presumably because the `go get -tags netgo ./prog/weaver` satisfies the dependency for the `odp` package, and `go build ... ./prog-weaver` builds only the main package).

**Comment by awh** [comment]
@rade I've included short peer ID in the report JSON and reworked the connection status implementation as discussed.

File: prog/weaver/http.go [hunk 1]
```diff
@@ -161,7 +161,12 @@ func HandleHTTP(muxRouter *mux.Router, version string, router *weave.Router, all
 	}
 	muxRouter.Methods("GET").Path("/report").Headers("Accept", "application/json").HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			json, _ := json.MarshalIndent(status(), "", "    ")
+			json, err := json.MarshalIndent(status(), "", "    ")
+			if err != nil {
+				Log.Error(err)
```
rade: If we *are* going to leave this in, is there enough context here? i.e. would we be able to tell that things failed to produce a status report?

router/status.go
```diff
@@ -176,7 +181,13 @@ func NewLocalConnectionStatusSlice(cm *ConnectionMaker) []LocalConnectionStatus
 			if conn.Established() {
 				state = "established"
 			}
-			slice = append(slice, LocalConnectionStatus{conn.RemoteTCPAddr(), conn.Outbound(), state, conn.Remote().String()})
+			var info string
+			if lc, ok := conn.(*LocalConnection); ok {
+				info = fmt.Sprintf("%-6v %v", lc.forwarder.OverlayType(), conn.Remote())
+			} else {
+				info = conn.Remote().String()
+			}
```
rade: no need to be defensive here; it will always be a `*LocalConnection`.

prog/weaver/http.go
awh: In the event that the marshalling fails, you see something like this:

````
$ weave report
json: unsupported type: odp.FlowKeys
````

This should only ever happen when a developer is extending the report JSON however.

dpw:
> Those method declarations have to be in the same package as the types they're bound to

Can't you introduce wrapper structs for `FlowInfo` and `Vport` in fastdp.go, and put the `MarshalJSON` method on those?

awh:
> Can't you introduce wrapper structs for FlowInfo and Vport in fastdp.go, and put the MarshalJSON method on those?

I have done so - see what you think. I don't think it's that great, I'd prefer to follow the model established in `status.go` already but this would mean binning off the `OverlayDiagnostics` abstraction and having `NewStatus()` reach inside the overlay switch and extract a typed reference to `FastDatapath` to pass to a `NewFastDatapathStatus()` constructor. Thoughts?

dpw:
> I have done so - see what you think.

Looks ok to me (or rather, it seems like a marginal increase in faff).

> I don't think it's that great, I'd prefer to follow the model established in status.go already but this would mean binning off the OverlayDiagnostics abstraction and having NewStatus() reach inside the overlay switch and extract a typed reference to FastDatapath to pass to a NewFastDatapathStatus() constructor. Thoughts?

I'm having trouble interpreting this, probably due to lack of familiarity with "the model established in status.go already".  But I don't see anything problematic in that area.

awh:
> I'm having trouble interpreting this, probably due to lack of familiarity with "the model established in status.go already". But I don't see anything problematic in that area.

It's a set of wrapper structs and constructors that rely on having access to the concrete types of everything, starting at the router and recursing down from there. I could have added wrapper structs and constructors there, the problem is that there's no clean way to get from `Router` to `FastDatapath` because of the `OverlaySwitch`...

awh:
In essence the `NewStatus()` method in `status.go` would have to cast `Router.Overlay` to an `OverlaySwitch` and then poke around its private `overlays` looking for `FastDatapath` which seems a bit grotty.

dpw:
I'm still not sure what all that machinery would achieve.  Avoiding `interface{}` in one or two places?  I think that may be counterproductive in golang.

awh:
> I'm still not sure what all that machinery would achieve

All the status generating stuff got centralised in `status.go`, but that was only possible because the router had concretely typed references to all the things that needed to have their status reported on. The purpose would be to maintain this existing convention of keeping all the wrapper struct gubbins in `status.go`, at the cost of punching through the `Overlay` abstraction.

> Avoiding interface{} in one or two places?

It's not about avoiding interface{} - it's about `status.go` having to have access to the concrete internals if all the status reading code is to remain centralised there.

dpw:
> All the status generating stuff got centralised in status.go

Got it.  There is a dichotomy between putting all the status generation code in `status.go`, and putting all the fastdp code in `fastdp.go`.  I think I prefer the code as it is currently, but I can see how that might seem like a retrograde step.

Modest proposal: `fastdp_status.go`

awh:
> All the status generating stuff got centralised in status.go

I have a feeling, given the new drive to factor out libweaventual and separate concerns in the router, that this may turn out not to have been a brilliant idea after all - let's leave this as is until the dust settles.

awh:
Rebased.

awh:
Closing to retarget against master.

PR #1517, opened 2015-10-08
`dns-{add,remove}` can add/remove non-container DNS records

### LINKED ISSUE #1385: CLI for adding/removing non-container DNS records
The weaveDNS API permits the addition/removal of records that are not associated with containers. But this functionality is not available in the CLI - `weave dns-add|remove` require a container name/id.

We should remove that restriction. Quite how to express that in the CLI is tbd.

This has been requested by a user.

rade:
...by omitting the container name

We associate non-container dns entries with 'weave:extern' because typically these entries will refer to external (non-weave) IP addresses. 'weave:extern' also happens to be the same length as the existing 'weave:expose' pseudo-container, which is desirable for 'weave status dns' output layout.

Fixes #1385.

### DIFF
--- a/site/weavedns.md
+++ b/site/weavedns.md
@@ -164,6 +164,18 @@ command:
 $ weave dns-remove $C
 ```
 
+By omitting the container name it is possible to add/remove DNS
+records that associate names in the weaveDNS domain with IP addresses
+that do not belong to containers, e.g. non-weave addresses of external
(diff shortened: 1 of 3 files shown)


rade:
Once we are happy with this I will create an issue to address the caveat, namely that records get removed when stopping the peer on which they were added.

PR #1518, opened 2015-10-09
Get proxy to re-attach weave router if it restarts

bboreham:
This is a follow-on to #1210, that fixes part (a) in #401

Replacement for #1359 which was against a different branch.

You get a bit of noise in the logs on a total restart (e.g. a reboot), because the router isn't necessarily up and listening before we start re-attaching other containers.  But it all works out in the end, because the router attach re-does all the parts that went wrong.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -316,7 +316,7 @@ func containerShouldAttach(container *docker.Container) bool {
 }
 
 func containerIsWeaveRouter(container *docker.Container) bool {
-	return strings.HasPrefix(container.Config.Image, "weaveworks/weave")
+	return len(container.Config.Entrypoint) > 0 && container.Config.Entrypoint[0] == weaveEntrypoint
 }
 
 func (proxy *Proxy) createWait(r *http.Request, ident string) {
@@ -393,7 +393,7 @@ func (proxy *Proxy) attach(container *docker.Container, orDie bool) error {
 }
 
 func (proxy *Proxy) attachRouter(container *docker.Container) error {
-	Log.Infof("Attaching weave router container")
+	Log.Infof("Attaching weave router container:", container.ID)
 	args := []string{"attach-router"}
 	if _, stderr, err := callWeave(args...); err != nil {
 		Log.Warningf("Attaching container %s to weave network failed: %s", container.ID, string(stderr))
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -393,7 +393,7 @@ func (proxy *Proxy) attach(container *docker.Container, orDie bool) error {
 }
 
 func (proxy *Proxy) attachRouter(container *docker.Container) error {
-	Log.Infof("Attaching weave router container:", container.ID)
+	Log.Infof("Attaching weave router container: %s", container.ID)
 	args := []string{"attach-router"}
 	if _, stderr, err := callWeave(args...); err != nil {
 		Log.Warningf("Attaching container %s to weave network failed: %s", container.ID, string(stderr))
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -305,12 +305,18 @@ func (proxy *Proxy) ContainerStarted(ident string) {
 	// If this was a container we modified the entrypoint for, attach it to the network
 	if containerShouldAttach(container) {
 		proxy.attach(container, true)
+	} else if containerIsWeaveRouter(container) {
+		proxy.attachRouter(container)
 	}
 	proxy.notifyWaiters(container.ID)
 }
 
 func containerShouldAttach(container *docker.Container) bool {
 	return len(container.Config.Entrypoint) > 0 && container.Config.Entrypoint[0] == weaveWaitEntrypoint[0]
+}
+
+func containerIsWeaveRouter(container *docker.Container) bool {
+	return strings.HasPrefix(container.Config.Image, "weaveworks/weave")
 }
 
 func (proxy *Proxy) createWait(r *http.Request, ident string) {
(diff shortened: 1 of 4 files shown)


bboreham:
Smoke-test was failing; after raising the `sleep` time to allow Docker to restart it has passed.

proxy/proxy.go
```diff
@@ -117,7 +117,8 @@ func NewProxy(c Config) (*Proxy, error) {
 func (proxy *Proxy) AttachExistingContainers() {
 	containers, _ := proxy.client.ListContainers(docker.ListContainersOptions{})
 	for _, cont := range containers {
-		if strings.HasPrefix(cont.Command, weaveWaitEntrypoint[0]) {
+		if strings.HasPrefix(cont.Command, weaveWaitEntrypoint[0]) ||
+			strings.HasPrefix(cont.Command, "/home/weave/weaver") {
```
rade: wouldn't it make more sense to check the image name, since that is how we make the same decision in `ContainerStarted`?

proxy/proxy.go
```diff
@@ -304,6 +307,10 @@ func containerShouldAttach(container *docker.Container) bool {
 	return len(container.Config.Entrypoint) > 0 && container.Config.Entrypoint[0] == weaveWaitEntrypoint[0]
 }
 
+func containerIsWeaveRouter(container *docker.Container) bool {
+	return strings.HasPrefix(container.Config.Image, "weaveworks/weave")
```
rade: This prevents users from fetching their images from a different repo, which is something the weave script permits by setting `DOCKERHUB_USER`.

proxy/proxy.go
bboreham: That would mean making an `inspect` call for every running container.  And the `ContainerStarted` test is actually wrong.

proxy/proxy.go
bboreham: Actually `Config.Image` is a long hex string, so this code is bugged.

weave
```diff
@@ -1194,6 +1198,17 @@ launch_router() {
...
+
+# Recreate the parameter values that are set when the router is first launched
+fetch_router_args() {
+    IPRANGE=$(docker inspect -f '{{.Args}}' $1 | grep -o -e '-ipalloc-range [^ ]*')
+    # This doesn't get back the right syntax, but it's enough for our purposes
+    DNS_PORT_MAPPING=$(docker inspect -f '{{with index .HostConfig.PortBindings "53/udp"}}{{.}}{{end}}' $1)
+}
+
+attach_router() {
+    ROUTER_CONTAINER=$1
```
rade: This shouldn' take the container name as the first arg. `$CONTAINER_NAME` is a global that is used everywhere, including in this function.

weave
```diff
@@ -1194,6 +1198,17 @@ launch_router() {
         -e WEAVE_PASSWORD \
         -e WEAVE_CIDR=none \
         $WEAVE_DOCKER_ARGS $IMAGE $COVERAGE_ARGS --iface $CONTAINER_IFNAME --port $CONTAINER_PORT --name "$PEERNAME" --nickname "$(hostname)" --ipalloc-range "$IPRANGE" --dns-effective-listen-address "$DOCKER_BRIDGE_IP" --docker-api "unix:///var/run/docker.sock" "$@")
+}
```
rade: I reckon this should invoke `attach_router`.

weave
```diff
@@ -1194,6 +1198,17 @@ launch_router() {
...
+
+# Recreate the parameter values that are set when the router is first launched
+fetch_router_args() {
+    IPRANGE=$(docker inspect -f '{{.Args}}' $1 | grep -o -e '-ipalloc-range [^ ]*')
+    # This doesn't get back the right syntax, but it's enough for our purposes
+    DNS_PORT_MAPPING=$(docker inspect -f '{{with index .HostConfig.PortBindings "53/udp"}}{{.}}{{end}}' $1)
+}
```
rade: This shouldn't take the container name as the first arg. Just use `$CONTAINER_NAME`.

rade: Perhaps this should return two "booleans", which we can then feed to `attach_router`.

proxy/proxy.go
bboreham: Not it isn't.  `container.Image` is a hex string; `container.Config.Image` is the image name it was started with.  But your point stands.

proxy/proxy.go
bboreham: Given the point below about wanting to work from different repositories, I think it is best to check the command/entrypoint in both places.

weave
bboreham: ROUTER_CONTAINER was the ID not the name, so overwriting it with the name is definitely wrong.  However we don't strictly need to use the ID inside this function, as you say.

bboreham:
Rebased against master.

weave
```diff
@@ -1367,7 +1371,18 @@ launch_router() {
...
         --docker-api "unix:///var/run/docker.sock" "$@")
-    with_container_netns_or_die $ROUTER_CONTAINER setup_router_iface_$BRIDGE_TYPE
+    attach_router
+}
+
+# Recreate the parameter values that are set when the router is first launched
+fetch_router_args() {
+    IPRANGE=$(docker inspect -f '{{.Args}}' $CONTAINER_NAME | grep -o -e '-ipalloc-range [^ ]*')
+    # This doesn't get back the right syntax, but it's enough for our purposes
+    DNS_PORT_MAPPING=$(docker inspect -f '{{with index .HostConfig.PortBindings "53/udp"}}{{.}}{{end}}' $CONTAINER_NAME)
```
rade: Needs re-basing, at which point you'll discover that this doesn't work in the fastdp case. A better check is to look for `--no-dns` and set the (new on master) NO_DNS_OPT.

weave
```diff
@@ -655,6 +655,10 @@ setup_router_iface_bridge() {
         echo "Perhaps you are running the docker daemon with container networking disabled (-b=none)." >&2
         return 1
     fi
+    # No-op if already attached
+    if netnsenter ip link show $CONTAINER_IFNAME >/dev/null 2>&1 ; then
+        return 0
+    fi
```
rade: surely we need something similar in `setup_router_iface_fastdp`.

bboreham: Well, that exposes a deeper issue.  Using fdp, the `weave` script behaves differently depending on whether you pass a `--password` argument, and we deliberately hide that information from the command-line.

It is sort-of possible to detect which path we went through, from `--datapath weave --iface weave2`, but that's ugly and twisted.

rade: Hmm. I wonder whether there is any harm in always doing the `weave2` magic in fastdp mode.

bboreham: I have been testing this; it looks like the `ip link set down` followed by `up` causes `weaver` to die.

bboreham: Actually, having rebased, the `ip link del weave2` is definitely fatal.

bboreham: If you move the `ip link del weave2` out of the way, and don't make `setup_router_iface` a no-op, you get a "File exists" error on `ip link add name `

bboreham:
Have rebased again and addressed recent comments.

bboreham:
And rebased again.

PR #1520, opened 2015-10-09
[proxy] remove status.sock before binding status.sock

### LINKED ISSUE #1519: Weaveproxy error on restart: "status.sock: bind: address already in use"
Running `weaveproxy` with `--restart=always` and then killing it leads to an endless sequence of errors:

    FATA: 2015/10/09 16:07:08.381199 ListenAndServeStatus failed: listen unix /home/weave/status.sock: bind: address already in use

paulbellamy:
Fixes #1519 

Debating adding a test, as it's caught by the tests in #1518


rade:
no test needed.

Issue #1448, opened 2015-09-18
fastdp compatibility with 1.1


dpw:
Following on from #1396 and #1438:

There's an easy part to this, and a hard part:

The easy part is to making sure that when a version 1.1 router and a #1438 router connect, they fall back to the sleeve Overlay.  #1438 does not do that, but it is trivial to make it do so.

The hard part is fast datapath's reliance on peer short ids. Version 1.1 peers will not generate short ids for themselves, and will not relay short id information.  Fast datapath cannot forward to peers it does not have a short id for.  So it's possible to envisage pathological cases where two groups of version 1.2 peers are connected only through a version 1.1 peer, leading to an effective partition in the overlay network.

(There is also an issue that GOB zeroes out fields when deserializing an old version of a struct.  So the range of `Peer.ShortID` will have to be adjusted from [0, 4095] to [1, 4096] in order to avoid conflating a missing short id with the valid short id of zero.)

I can think of two approaches to supporting rolling upgrades:

The first is to require users to upgrade to 1.2 via an intermediate step, where they start 1.2 in a fastdp disabled mode.  I plan such a mode anyway, so no extra development work will be required.  Once a cluster has been upgraded to 1.2 with fastdp disabled, the user is free to restart routers with fastdp enabled.

The downside of this approach is the potential for confusion if users try to upgrade without following the release notes carefully.  With 1.1, we have seen that debugging issues that might or might not be compatibility bugs is a time consuming process.  Given that this approach means that we genuinely will have a compatibility issue unless the appropriate option is passed, I am concerned about what will ensue.

The second option is to run the fastdp and sleeve overlays in parallel, and use the sleeve overlay whenever we need to forward a packet to (and from) a peer for which we lack a short id.  Currently in #1438, `OverlaySwitch` does not do this:  When they fastdp `OverlayForwarder` confirms, the sleeve `OverlayForwarder` gets stopped.  In the long term, that behaviour is desirable: We don't want to maintain the sleeve forwarder machinery, sending heartbeats and fragmentation tests etc., if it is not needed.  But it would be fairly easy to change that for 1.2.  Then in 1.3, we could change it back.

awh:
> The downside of this approach is the potential for confusion ... we genuinely will have a compatibility issue unless the appropriate option is passed

What if fastdp is disabled by default?

awh:
> Version 1.1 peers ... will not relay short id information

[...]

PR #1522, opened 2015-10-12
Fast datapath compatibility with 1.1

dpw:
This is stacked on top of the fdp-rebase branch, so should not be merged until after #1438.

Addresses #1448.  Compatibility with 1.1 is intended to "just work", so that we won't need to explain any special procedure for doing rolling upgrades to 1.2.

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -196,7 +196,9 @@ func main() {
 	if fastDPOverlay != nil {
 		overlays.Add("fastdp", fastDPOverlay)
 	}
-	overlays.Add("sleeve", weave.NewSleeveOverlay(config.Port))
+	sleeve := weave.NewSleeveOverlay(config.Port)
+	overlays.Add("sleeve", sleeve)
+	overlays.SetCompatOverlay(sleeve)
 	config.Overlay = overlays
 
 	if routerName == "" {
--- a/router/connection.go
+++ b/router/connection.go
@@ -277,7 +277,7 @@ func (features features) Get(key string) string {
 }
 
 func (conn *LocalConnection) parseFeatures(features features) (*Peer, error) {
-	if err := features.MustHave([]string{"PeerNameFlavour", "Name", "NickName", "ShortID", "UID", "ConnID"}); err != nil {
+	if err := features.MustHave([]string{"PeerNameFlavour", "Name", "NickName", "UID", "ConnID"}); err != nil {
 		return nil, err
 	}
 
@@ -293,10 +293,15 @@ func (conn *LocalConnection) parseFeatures(features features) (*Peer, error) {
 
 	nickName := features.Get("NickName")
 
-	shortID, err := strconv.ParseUint(features.Get("ShortID"), 10,
-		PeerShortIDBits)
-	if err != nil {
-		return nil, err
+	var shortID uint64
+	var hasShortID bool
+	if shortIDStr, present := features["ShortID"]; present {
+		hasShortID = true
+		shortID, err = strconv.ParseUint(shortIDStr, 10,
+			PeerShortIDBits)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	uid, err := ParsePeerUID(features.Get("UID"))
@@ -310,7 +315,9 @@ func (conn *LocalConnection) parseFeatures(features features) (*Peer, error) {
 	}
 
 	conn.uid ^= remoteConnID
-	return NewPeer(name, nickName, uid, 0, PeerShortID(shortID)), nil
+	peer := NewPeer(name, nickName, uid, 0, PeerShortID(shortID))
+	peer.HasShortID = hasShortID
+	return peer, nil
 }
 
 func (conn *LocalConnection) registerRemote(remote *Peer, acceptNewPeer bool) error {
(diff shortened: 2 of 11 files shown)


awh:
It'd be nice to have a smoketest to prove the compatibility works - it would be trivial if not for the fact that one of the VMs needs to be downgraded to a release version of weave for the test in question - I'll see if I can think of a way to do that cleanly. No comments otherwise, ready to merge.

rade:
All that +/-1 is rather tedious. Wouldn't it be simpler to just give up one id, i.e. have them range from 1 to 4095?

dpw:
> All that +/-1 is rather tedious. Wouldn't it be simpler to just give up one id, i.e. have them range from 1 to 4095?

That would eliminate the +1/-1 mapping in fastdp.go, but lead to other +1/-1 popping up in the places where we generate random short ids.  Actually, the impact on randomPeerShortID would be worse, because currently it masks a randomUint16().

rade:
> the places where we generate random short ids

afaict there are two functions - `Peers.chooseShortID` and `randomPeerShortyID` - that would be affected. These are *already* affected by the +/-1 change. Yes, the code to produce random numbers in the range 1-4095 is more complicated than for 0-4095, but it would eliminate the +/-1 in the fastdp code.

dpw:
Sounds like a cartoon where they push down on a lump under the carpet, and another lump pops up elsewhere.

There's another approach that might turn out tidier: Add a `HasShortID` bool field to `Peer`, always setting it to true in `NewPeer`.  Any `Peer` with it set to false is one that came from a 1.1 peer.  Then the range of `ShortID` wouldn't need any adjustment.

I can't think of any gotchas right now, I'll look at it more tomorrow.

(And with some care, that `HasShortID` field could be retired eventually.)

rade:
somewhat pertinent... the addition of `ShortID` to `Peer` is unfortunate; it is the *only* field in `Peer` - and topology gossip overall - that is specific to the overlay. It has nothing to do with the control plane as such. I'm trying hard to think of it in generic terms ("there is nothing peculiar about wanting each peer in your system to have a short id"), but it's a struggle ;)

awh:
Can this change not be reverted for 1.3?

dpw:
> somewhat pertinent...

I don't think it's pertinent to this issue - it's certainly not something to be addressed as part of this compat work.  It may be pertinent to the main fastdp issue, so if you want to raise it, I suggest raising it there.  (But that has been in a "for review" state for a couple of weeks now.  And the ShortID code was in something very close to its present state in the fastdp preview a few months ago.  This seems like an odd time to take an interest in it.)

> the addition of ShortID to Peer is unfortunate; it is the only field in Peer - and topology gossip overall - that is specific to the overlay. It has nothing to do with the control plane as such. I'm trying hard to think of it in generic terms ("there is nothing peculiar about wanting each peer in your system to have a short id"), but it's a struggle

I don't greatly disagree.  You probably won't be surprised to hear that there are many aspects of the fastdp code I remain unsatisfied with, and this is one.  It is the something specific to the fastdp overlay that leaks out.  But, my aim in the fastdp work has been that it should become a key part of Weave Net, so it is questionable how much it matters that it is not structured as a pure "bolt on extra" to the rest of the codebase.  The perfect is the enemy of the good; perhaps the ShortID is glaring simply because the overlay code now has a modular structure.  Given how prolonged the merge process for that has been, I'm not sure it would have been viable to extend the scope of the fastdp effort further, and I'm not sure exactly what your intent is.

dpw:
> Can this change not be reverted for 1.3?

Some parts yes, some parts no.  There is no intention that these commits can simply be reverted.  Where we can do things now which make a real difference to the possibility of simplifying the code in later releases, that's good. And I have ideas about how I would go about those simplifications, some of which I described in the associated issue (largely for my own reference; I'm not sure how intelligible the description was to others). But I don't think it is worth doing work now that can just as well, and perhaps better, be done later.

rade:
@dpw I certainly wasn't suggesting that we hold up merging of fdp; I only mentioned the `ShortID` issue in case you could think of a quick fix but also because adding `HasShortID` arguably makes the situation worse.

awh:
> > Wouldn't it be simpler to just give up one id, i.e. have them range from 1 to 4095?

> Sounds like a cartoon where they push down on a lump under the carpet, and another lump pops up elsewhere.

Except that if you did as Matthias suggested, it would eliminate the offset in the VNI to ShortID mapping, which would mean (I think) that the only consequence of completely reverting the reservation of ShortID zero in 1.3 would be that a new peer which happened to choose zero for itself would be unreachable by existing peers until they were restarted.

awh:
> which would mean (I think) that the only consequence of completely reverting the reservation of ShortID zero in 1.3 would be that a new peer which happened to choose zero for itself would be unreachable by existing peers until they were restarted.

Is this right, and if so, do you consider the consequence tolerable as part of a rolling upgrade?

dpw:
> Is this right, and if so, do you consider the consequence tolerable as part of a rolling upgrade?

As you two seem determined to make a fuss about the +1/-1 in fastdp.go, I'm implementing my `HasShortID` proposal, which doesn't add a +1 or a -1 anywhere.

dpw:
Revised.

PR #1525, opened 2015-10-13
Fast datapath weave status

### LINKED ISSUE #1505: Include fast datapath information in `weave status|report`

awh:
This is a replacement for #1513 that is targeted against master.

Fixes #1505.

### DIFF
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -161,7 +161,12 @@ func HandleHTTP(muxRouter *mux.Router, version string, router *weave.Router, all
 	}
 	muxRouter.Methods("GET").Path("/report").Headers("Accept", "application/json").HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			json, _ := json.MarshalIndent(status(), "", "    ")
+			json, err := json.MarshalIndent(status(), "", "    ")
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				Log.Error("Error during report marshalling: ", err)
+				return
+			}
 			w.Header().Set("Content-Type", "application/json")
 			w.Write(json)
 		})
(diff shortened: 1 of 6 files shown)


dpw:
https://github.com/weaveworks/weave/commit/a553691374c2756cc3a92f0565d5afa2d4a1ba35#commitcomment-13720243

awh:
> I'd suggest renaming this method to make it clearer it yields a name for display to the user. E.g. DisplayName.

Done.

dpw:
The only further comment I have is to note that this doesn't put the overlay type into the gossipped connection information as shown in `weave report`.  But even if that's worth doing, it can be a separate issue.

So squash the fixup, and I'm happy to merge.

awh:
> this doesn't put the overlay type into the gossipped connection information as shown in weave report. But even if that's worth doing, it can be a separate issue.

I did consider that, but came to the same conclusion. Fixup squashed.

dpw:
I was wondering whether the number of flows to be shown in `weave report` might be a problem.  It could grow relatively large (easily hundreds, possibly thousands, tens of thousands might start to run into other limits).  But if we allow the JSON describing a flow to be a generous 1000 bytes, it doesn't sound problematic.  And in such cases, there might well also be hundreds or thousands of entries in the peer connections lists, so this is unlikely to present any new issues.

(I'm disappointed to find that the golang JSON encoder always encodes to an internal buffer, you cannot give it a Writer.  Even stranger, `MarshalIndent` works by generating unindented JSON, then parsing it in order to indent it.)

awh:
> I'm disappointed

Disappointed, but not surprised I'll wager :smile:

> MarshalIndent works by generating unindented JSON, then parsing it in order to indent it.

I don't even.

PR #1510, opened 2015-10-06
Fast datapath documentation

awh:
This seemed the best vehicle for receiving comments! @dpw @rade @inercia

PR #1526, opened 2015-10-13
Document fast data path feature

awh:
This is a replacement for #1510 that is targeted against master.

### DIFF
--- a/site/features.md
+++ b/site/features.md
@@ -9,6 +9,7 @@ Weave has a few more features beyond those illustrated by the [basic
 example](https://github.com/weaveworks/weave#example):
 
  * [Virtual ethernet switch](#virtual-ethernet-switch)
+ * [Fast data path](#fast-data-path)
  * [Seamless Docker integration](#docker)
  * [Address allocation](#addressing)
  * [Naming and discovery](#naming-and-discovery)
@@ -54,6 +55,25 @@ and troubleshoot our container network. To put it another way, we can
 now re-use the same tools and techniques when deploying applications
 as containers as we would have done when deploying them 'on metal' in
 our data centre.
+
+### <a name="fast-data-path"></a>Fast data path
+
+Weave automatically chooses the fastest available method to transport
+data between peers. The most performant of these ('fastdp') offers
+near-native throughput and latency but does not support encryption;
+consequently supplying a password will cause the router to fall back
+to a slower mode ('sleeve') that does.
+
+Even when encryption is not in use, certain adverse network conditions
+will cause this fallback to occur dynamically; in these circumstances,
+weave will upgrade the connection back to the fastdp transport without
+user intervention once they abate. You can see which method is in use
+by examining the output of `weave status connections`.
+
+You can also administratively disable fastdp with the
+`WEAVE_NO_FASTDP` environment variable:
+
+    $ WEAVE_NO_FASTDP=true weave launch
 
 ### <a name="docker"></a>Seamless Docker integration
 
@@ -282,7 +302,10 @@ way to generate a random password which satsifies this requirement is
 
     < /dev/urandom tr -dc A-Za-z0-9 | head -c9 ; echo
 
-The same password must be specified for all weave peers.
+The same password must be specified for all weave peers. Note that
+supplying a password will [cause weave to fall back to a slower
+method](#fast-data-path) for transporting data between
+peers.
 
 ### <a name="host-network-integration"></a>Host network integration
 
@@ -449,11 +472,13 @@ Weave can network containers hosted in different cloud providers /
 data centres. So, for example, one could run an application consisting
 of containers on GCE, EC2 and in local data centres.
 
-To enable this, the network must be configured to permit TCP and UDP
-connections to the weave port of the docker hosts. The weave port
-defaults to 6783. This can be overriden by setting `WEAVE_PORT`, but
-it is highly recommended that all peers in a weave network are given
-the same port setting.
+To enable this, the network must be configured to permit connections
+to weave's control and data ports on the docker hosts. The control
+port defaults to TCP 6783, and the data ports to UDP 6783/6784. You
+can override these defaults by setting `WEAVE_PORT` (this is a base
+value - setting `WEAVE_PORT=9000` will result in weave using TCP 9000
+for control and UDP 9000/9001 for data). Note that it is highly
+recommended that all peers be given the same setting.
 
 ### <a name="multi-hop-routing"></a>Multi-hop routing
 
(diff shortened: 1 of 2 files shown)


awh:
@rade @dpw I've added some user facing documentation, would appreciate review.

site/features.md
```diff
@@ -55,6 +56,28 @@ now re-use the same tools and techniques when deploying applications
...
 our data centre.
 
+### <a name="automatic-overlay-selection"></a>Automatic overlay selection
+
+Weave supports two overlay technologies - a custom UDP encapsulation
+('sleeve') featuring optional encryption and underlay PMTU discovery,
+and a VXLAN transport ('fastdp') that leverages the Open vSwitch
+kernel module to provide near-native performance. When connecting to a
+new peer, the router automatically prefers the fastdp overlay under
+the following conditions:
```
rade: That's a lot of technical jargon. Overlay? UDP encapsulation? PMTU discovery? VXLAN? Open vSwitch kernel module? ... I don't think a user should have to know about any of these.

How about we call this section "Optimised data path". Explain in a sentence what we mean by "data path". Explain that normally weave will use a "fast in-kernel data path" or s.t. like that.

site/features.md
```diff
@@ -55,6 +56,28 @@ now re-use the same tools and techniques when deploying applications
...
+
+* UDP port 6784 must be reachable (in addition to TCP/UDP 6783)
```
rade: We need to mention the 6784 in the README (where we mention 6783). Needs a separate PR on a readme-v1.2 (or whatever naming scheme we used previously).

site/features.md
```diff
@@ -282,7 +305,8 @@ way to generate a random password which satsifies this requirement is
 
     < /dev/urandom tr -dc A-Za-z0-9 | head -c9 ; echo
 
-The same password must be specified for all weave peers.
+The same password must be specified for all weave peers. Note that
+supplying a password will disable support for the fastdp overlay.
```
rade: link back to the new section.

site/features.md
```diff
@@ -453,7 +477,9 @@ To enable this, the network must be configured to permit TCP and UDP
 connections to the weave port of the docker hosts. The weave port
 defaults to 6783. This can be overriden by setting `WEAVE_PORT`, but
 it is highly recommended that all peers in a weave network are given
-the same port setting.
+the same port setting. If you wish to use the fastdp overlay, the UDP
+port with value `WEAVE_PORT+1` (6784 by default) must also be
+permitted.
```
rade: link back to the new section.

site/features.md
```diff
@@ -55,6 +56,28 @@ now re-use the same tools and techniques when deploying applications
...
+Weave supports two overlay technologies - a custom UDP encapsulation
+('sleeve') featuring optional encryption and underlay PMTU discovery,
+and a VXLAN transport ('fastdp') that leverages the Open vSwitch
+kernel module to provide near-native performance. When connecting to a
+new peer, the router automatically prefers the fastdp overlay under
+the following conditions:
+
+* UDP port 6784 must be reachable (in addition to TCP/UDP 6783)
+* Underlay network PMTU must be greater than or equal to 1460 bytes
+* Encryption must be disabled
```
rade: "not be enabled" (given that it is an opt-in)

site/features.md
dpw: I don't think we should use the term "overlay" at all.

"VXLAN transport" risks creating the expectation of VXLAN interoperability. At most "... that leverages the Open vSwitch kernel module and VXLAN encapsulation ...".

awh:
I have removed the information content as requested.

site/features.md
awh: OK I will split that out of here once we're happy overall.

site/features.md
```diff
@@ -55,6 +56,20 @@ now re-use the same tools and techniques when deploying applications
 as containers as we would have done when deploying them 'on metal' in
 our data centre.
 
+### <a name="optimised-traffic-transport"></a>Optimised traffic transport
```
rade: s/traffic/data?

dpw: What's wrong with calling this section 'fast datapath'?  If that name is not suitable for a section title, we should find a name we like better and stick to it (fixing 'fastdp' in the status and WEAVE_NO_FASTDP accordingly).

site/features.md
```diff
@@ -55,6 +56,20 @@ now re-use the same tools and techniques when deploying applications
...
+
+Weave automatically chooses the fastest available method to transport
+traffic between peers. The most performant of these ('fastdp') offers
+near-native throughput and latency but does not support encryption;
+consequently supplying a password will cause the router to fall back
+to a slower mode ('sleeve') that does. You can see which method is in
```
rade: "...does not support encryption. The slower method ('sleeve') is selected when [encryption is enabled](#security), or communication via the 'fastdp' method fails for some other reason. You can see..."

site/features.md
rade: I would be fine with "Fast data path". It's perhaps not the most descriptive we can come up with, but it's not far off.

awh:
* 'Optimised traffic transport' becomes 'Optimised data transport'
* Added explanation of sleeve fallback under adverse network conditions
* README changes pulled out in #1541

site/troubleshooting.md
```diff
@@ -161,8 +161,9 @@ The columns are as follows:
     * `pending` - TCP connection up, waiting for confirmation of
        UDP heartbeat
     * `established` - TCP connection and corresponding UDP path are up
- * Info - the remote peer name and nickname for (un)established
-   connections, the failure reason for failed and retrying connection
+ * Info - the traffic transport method, remote peer name and nickname
+   for (un)established connections, the failure reason for failed and
+   retrying connection
```
rade: I know this was wrong before, but there is no such thing as "unestablished". Should say "pending and established". Also, I reckon "failed and retrying" should come before "pending and established", to match the order in Status.

awh:
* 'Optimised data transport' becomes 'Fast data path'

site/troubleshooting.md
awh: Done.

awh:
@dpw did you have any comments on the high level overview doc?

dpw:
> @dpw did you have any comments on the high level overview doc?

I have a fairly strong idea of what a "weave router internals" doc should look like, as I was planning to write one as noted in my "things still to do" list on the fastdp PR.  But I appreciate that you had to figure stuff out as you went along, and I don't see any harm in having that written down.  Your doc might be quite far from what I had in mind though, so I'm not sure I will have much useful feedback.

If you want to collaborate to put together a comprehensive doc, we can work on that.  Or not.  Either is fine by me.  In the former case, I'd suggest splitting the user-facing documentation part out onto a separate PR so we can get that merged promptly.

awh:
I've pulled the overview out of this PR; only updates to the user facing documentation remain.

PR #1528, opened 2015-10-13
[proxy] Fix bug in weave env/config when passing multiple -H flags

### LINKED ISSUE #1527: weave launch-proxy with multiple -H causes invalid weave {config, env} output
What I did:

```
weave launch-proxy -H tcp://0.0.0.0:12375 -H unix:///var/run/weave/weave.sock
weave launch-router
weave env
weave config
```

What I expected:

```
export DOCKER_HOST=tcp://127.0.0.1:12375 ORIG_DOCKER_HOST=
-H=tcp://127.0.0.1:12375
```

What I got:

```
export DOCKER_HOST=tcp://127.0.0.1:12375 unix:///var/run/weave/weave.sock ORIG_DOCKER_HOST=
-H=tcp://127.0.0.1:12375 unix:///var/run/weave/weave.sock
```

From current master.

paulbellamy:
The echo in proxy_addr was not preserving newlines, so the multiple
addrs were all getting returned.

Fixes #1527

### DIFF
--- a/test/650_proxy_env_test.sh
+++ b/test/650_proxy_env_test.sh
@@ -2,6 +2,13 @@
 
 . ./config.sh
 
+CMD="run -e WEAVE_CIDR=10.2.1.4/24 $SMALL_IMAGE $CHECK_ETHWE_UP"
+
+check() {
+  assert_raises "eval '$(weave_on $HOST1 env)' ; docker $CMD"
+  assert_raises "docker $(weave_on $HOST1 config) $CMD"
+}
+
 start_suite "Configure the docker daemon for the proxy"
 
 # No output when nothing running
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -215,9 +215,7 @@ func (proxy *Proxy) ListenAndServeStatus(socket string) {
 }
 
 func (proxy *Proxy) StatusHTTP(w http.ResponseWriter, r *http.Request) {
-	for _, addr := range proxy.normalisedAddrs {
-		fmt.Fprintln(w, addr)
-	}
+	fmt.Fprintln(w, proxy.normalisedAddrs[0])
```
rade: doesn't that mean `weave status` will only show the first listening address?

test/650_proxy_env_test.sh
```diff
@@ -2,6 +2,18 @@
...
+CMD="run -e WEAVE_CIDR=10.2.1.4/24 $SMALL_IMAGE $CHECK_ETHWE_UP"
+
+check() {
+  assert_raises "eval '$(weave_on $HOST1 env)' ; docker $CMD"
+  assert_raises "docker $(weave_on $HOST1 config) $CMD"
+
+  # Check we can use the weave script through the proxy
+  assert_raises "eval '$(weave_on $HOST1 env)' ; $WEAVE version"
+  assert_raises "eval '$(weave_on $HOST1 env)' ; $WEAVE ps"
+  assert_raises "eval '$(weave_on $HOST1 env)' ; $WEAVE launch-router"
```
rade: I know this was there before, but now we invoke this twice.... wouldn't it suffice to have *one* of these three tests, e.g. just the version test?

Issue #1532, opened 2015-10-13
weaveproxy doesn't handle HostConfig json on start request named "HostConfig"


bboreham:
We handle a `HostConfig` structure passed unnamed at top level, i.e.

````
{
   "Memory":0
}
````

, but not something like:

````
"HostConfig": {
   "Memory":0
}
````

Docker appears to handle either, in `ContainerConfigWrapper.getHostConfig()`, but I haven't tested it yet.

rade:
Is this problem present in 1.1.x?

bboreham:
In the sense that passing any kind of HostConfig on `start` broke the proxy, yes.  That was #1368, which is not fixed on the 1.1 branch.

rade:
Fixed by #1534

PR #1534, opened 2015-10-14
Take more care over HostConfig on start request in proxy

### LINKED ISSUE #1531: weaveproxy can overwrite port bindings from docker-compose
To recreate using `docker-compose`, create a file `docker-compose.yml`:

    foo:
      image: gliderlabs/alpine
      command: sleep 100
      ports:
       - "1234:1234"

Start this container via `weaveproxy`:

    DOCKER_HOST=unix:///var/run/weave/weave.sock docker-compose up -d

Check for the published port:

    $ docker inspect -f '{{.HostConfig.PortBindings}}' test_foo_1
    map[]

Cause is that `docker-compose` is sending in an empty JSON request on `start`; we are setting our HostConfig fields on that request, and hence overwriting the `PortBindings`.

This does not happen if you try via `docker create` and `docker start`, since `docker start` does not send any content at all in the request and `startContainerInterceptor.InterceptRequest()` detects that and bypasses setting the HostConfig.

bboreham:
"There are more things in heaven and earth, `HostConfig`, than are dreamt of in your philosophy"

Fixes #1531 and #1532

### DIFF
--- a/proxy/start_container_interceptor.go
+++ b/proxy/start_container_interceptor.go
@@ -12,21 +12,24 @@ func (i *startContainerInterceptor) InterceptRequest(r *http.Request) error {
 		return err
 	}
 
+	// If the client has sent some JSON which might be a HostConfig, add our
+	// parameters back into it, otherwise Docker will consider them overwritten
 	if containerShouldAttach(container) && r.Header.Get("Content-Type") == "application/json" {
 		hostConfig := map[string]interface{}{}
 		if err := unmarshalRequestBody(r, &hostConfig); err != nil {
 			return err
 		}
+		if len(hostConfig) > 0 {
+			i.proxy.addWeaveWaitVolume(hostConfig)
+			if dnsDomain := i.proxy.getDNSDomain(); dnsDomain != "" {
+				if err := i.proxy.setWeaveDNS(hostConfig, container.Config.Hostname, dnsDomain); err != nil {
+					return err
+				}
+			}
 
-		i.proxy.addWeaveWaitVolume(hostConfig)
-		if dnsDomain := i.proxy.getDNSDomain(); dnsDomain != "" {
-			if err := i.proxy.setWeaveDNS(hostConfig, container.Config.Hostname, dnsDomain); err != nil {
+			if err := marshalRequestBody(r, hostConfig); err != nil {
 				return err
 			}
-		}
-
-		if err := marshalRequestBody(r, hostConfig); err != nil {
-			return err
 		}
 	}
 	i.proxy.createWait(r, container.ID)
--- a/proxy/start_container_interceptor.go
+++ b/proxy/start_container_interceptor.go
@@ -2,15 +2,9 @@ package proxy
 
 import (
 	"net/http"
-
-	"github.com/fsouza/go-dockerclient"
 )
 
 type startContainerInterceptor struct{ proxy *Proxy }
-
-type startContainerRequestBody struct {
-	HostConfig *docker.HostConfig `json:"HostConfig,omitempty" yaml:"HostConfig,omitempty"`
-}
 
 func (i *startContainerInterceptor) InterceptRequest(r *http.Request) error {
 	container, err := inspectContainerInPath(i.proxy.client, r.URL.Path)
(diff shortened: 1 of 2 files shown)


test/666_abuse_of_start_test.sh
```diff
@@ -11,7 +11,12 @@ proxy_start_container $HOST1 --name=c1
 # Start c2 with a sneaky HostConfig
 proxy docker_on $HOST1 create --name=c2 $SMALL_IMAGE $CHECK_ETHWE_UP
 proxy docker_api_on $HOST1 POST /containers/c2/start '{"NetworkMode": "container:c1"}'
-assert "docker_on $HOST1 inspect -f '{{.State.Running}} {{.State.ExitCode}} {{.HostConfig.Dns}}' c2" "false 0 [$docker_bridge_ip]"
+assert "docker_on $HOST1 inspect -f '{{.HostConfig.NetworkMode}} {{.State.Running}} {{.State.ExitCode}} {{.HostConfig.Dns}}' c2" "container:c1 false 0 [$docker_bridge_ip]"
+
+# Start c5 with a differently sneaky HostConfig
+proxy docker_on $HOST1 create --name=c5 $SMALL_IMAGE $CHECK_ETHWE_UP
+proxy docker_api_on $HOST1 POST /containers/c5/start '{"HostConfig": {"NetworkMode": "container:c1"}}'
+assert "docker_on $HOST1 inspect -f '{{.HostConfig.NetworkMode}} {{.State.Running}} {{.State.ExitCode}} {{.HostConfig.Dns}}' c5" "container:c1 false 0 [$docker_bridge_ip]"
```
rade: The repetition of the `{{.State.Running}} {{.State.ExitCode}} {{.HostConfig.Dns}}` and `false 0 [$docker_bridge_ip]` is getting a bit tedious. I did bite my tongue when there were only two of these. Now there are three.

bboreham: The third one is different. I have pushed a change to reduce the repetition.

paulbellamy:
If the user sends *both* a top-level HostConfig, *and* a named sub-HostConfig, it looks like we only modify the inner one. Will that override the outer?

rade:
needs rebasing - the new `attach` should presumably go into the `check_...` function.

paulbellamy:
Maybe add a test for cpuset handling, and the case where there are both HostConfigs, but other than that, LGTM.

bboreham:
I looked into `Cpuset`.  Docker only uses that value if you send _both_ `Cpuset` and `HostConfig`.  So, for any caller sending _just_ `Cpuset`, we change the behaviour: previously docker would just ignore the `Cpuset` and now it will use it, and additionally any `HostConfig` parameters supplied in the original create will be blanked out.

However, this undesirable behaviour will happen for any valid JSON that Docker was previously ignoring, and we cannot detect all such requests.  So I propose not to detect any of them.

test/666_abuse_of_start_test.sh
```diff
@@ -4,24 +4,38 @@
 
 start_suite "Abuse of 'start' operation"
 
-weave_on $HOST1 launch
+weave_on $HOST1 launch --log-level=debug
```
rade: I'm guessing that's not meant to be there.

test/666_abuse_of_start_test.sh
```diff
@@ -4,24 +4,38 @@
...
 docker_bridge_ip=$(weave_on $HOST1 docker-bridge-ip)
 proxy_start_container $HOST1 --name=c1
 
+check_hostconfig() {
+    docker_on $HOST1 attach c2 >/dev/null || true # Wait for container to exit
```
rade: on master this now suppresses errors with `>/dev/null 2>&1`. Should do here too.

proxy/start_container_interceptor.go
```diff
@@ -18,25 +18,33 @@ func (i *startContainerInterceptor) InterceptRequest(r *http.Request) error {
...
-	if err := marshalRequestBody(r, hostConfig); err != nil {
-		return err
+		// HostConfig can be sent either unnamed at top level or as a struct named HostConfig
+		hostConfig := params
+		if subParam, found := params["HostConfig"]; found {
+			if typecast, ok := subParam.(map[string]interface{}); ok {
+				hostConfig = typecast
+			}
+		}
+		if len(hostConfig) > 0 {
```
paulbellamy: What about `{"HostConfig":{}}`, looks like that won't trigger this?

bboreham: I think that will leave both pointers as nil so do nothing.  Unless you've tested it?

rade:
@bboreham needs rebase, apparently.

bboreham:
Rebased and squashed into fewer commits

proxy/start_container_interceptor.go
paulbellamy: Have pushed a failing test for it.

PR #1536, opened 2015-10-14
Print request and response json when debugging weaveproxy

### LINKED ISSUE #1515: For debugging, provide a way to see data passing in and out of weaveproxy
Sometimes, to understand what has gone wrong, we would like to see the exact json that went in and out of the proxy.

bboreham:
Fixes #1515 

Simply added to debug-level logging, to avoid the added complexity of a new command-line flag to turn this on.

Thanks to @tomwilkie for the implementation

### DIFF
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -85,6 +85,7 @@ func unmarshalResponseBody(r *http.Response, target interface{}) error {
 	if err != nil {
 		return err
 	}
+	Log.Debugf("->responseBody: %s", body)
 	if err := r.Body.Close(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


tomwilkie:
Worth noting not all commands use unmarshalRequestBody, so I don't think this covers everything.

bboreham:
ok you got me - which ones don't use it?

rade:
surely we want something in `{un}marshalResponseBody` too.

rade:
> which ones don't use it

`NullInterceptor`, and and the no-op `Intercept{Request/Response}` of some other interceptors.

bboreham:
Ones we don't modify don't need to be printed.

PR #1538, opened 2015-10-14
signal mismatches between WEAVE_NO_FASTDP and the bridge type

### LINKED ISSUE #1533: signal mismatches between WEAVE_NO_FASTDP and the bridge type
Someone might reasonably do:
```
weave launch
weave stop
WEAVE_NO_FASTDP=1 weave launch
```
and be surprised that they are still using fastdp.  So we should validate in `create_bridge` that the bridge type and WEAVE_NO_FASTDP match up, and fail if they don't (or just warn?).

dpw:
Fixes #1533.

### DIFF
--- a/weave
+++ b/weave
@@ -406,8 +406,25 @@ create_bridge() {
         # whether it is a bridge instead.
         if [ -d /sys/class/net/$BRIDGE/bridge ] ; then
             BRIDGE_TYPE=bridge
+
+            if [ -n "$LAUNCHING_ROUTER" -a -z "$WEAVE_NO_FASTDP" ] ; then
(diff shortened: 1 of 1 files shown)


dpw:
Note that this intentionally doesn't signal if WEAVE_NO_FASTDP is not set but the bridge is a normal bridge netdev.  The idea being that if someone does `WEAVE_NO_FASTDP=1 weave launch`, that setting should be persistent across weave commands; they should not have to set WEAVE_NO_FASTDP generally.

dpw:
Improved wording, and added a check for the converse case but only in `weave launch` / `weave launch-router`.

PR #1550, opened 2015-10-16
Don't fail in launch-router if weave2 already exists

### LINKED ISSUE #1548: weave2 interface left behind
I observed this testing Kubernetes/Azure guide, which is on CoreOS. I'm pretty sure it's more of general  networkd issue, basically can be fixed in a way similar to #144.

dpw:
And delete weave2 on stop-router.

Fixes #1548.

### DIFF
--- a/weave
+++ b/weave
@@ -645,6 +645,8 @@ router_opts_bridge () {
 setup_router_iface_fastdp() {
     if [ -n "$WEAVE_PASSWORD" ] ; then
         # See router_opts_fastdp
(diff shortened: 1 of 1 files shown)


dpw:
> Surely this also needs to be done on stop. After all, the same RTNETLINK answers: File exists shows up there too.

Good point.

> I'm guessing we don't need this on reset because that destroys the "bridge" and presumably that takes the veth with it. Correct?

Correct.

PR #1551, opened 2015-10-16
Proxy restart shouldn't kill containers

### LINKED ISSUE #1547: proxy shouldn't kill containers when it restarts
image

paulbellamy:
Fixes #1547.

First commit stops it killing containers when restarting the proxy.

Second adds a first pass at an attachment retry mechanism. Thoughts on that? I need to have a think about how to test it.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -116,16 +116,17 @@ func NewProxy(c Config) (*Proxy, error) {
 
 func (proxy *Proxy) AttachExistingContainers() {
 	containers, _ := proxy.client.ListContainers(docker.ListContainersOptions{})
-	for _, apiContainer := range containers {
-		container, err := proxy.client.InspectContainer(apiContainer.ID)
+	for _, c := range containers {
+		container, err := proxy.client.InspectContainer(c.ID)
 		if err != nil {
-			Log.Warningf("Error inspecting container %s: %v", apiContainer.ID, err)
+			if _, ok := err.(*docker.NoSuchContainer); !ok {
+				Log.Warningf("unable to attach existing container %s since inspecting it failed: %v", c.ID, err)
+			}
 			continue
 		}
-		if containerShouldAttach(container) {
+		if containerShouldAttach(container) && (container.State.Running || container.State.Paused) {
 			proxy.attach(container, false)
 		}
-		proxy.notifyWaiters(container.ID)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


rade:
Please split part 2 into a separate issue and PR - I doubt that is going to make it into 1.2.0, whereas part 1 should.

proxy/proxy.go
```diff
@@ -116,10 +116,16 @@ func NewProxy(c Config) (*Proxy, error) {
 
 func (proxy *Proxy) AttachExistingContainers() {
 	containers, _ := proxy.client.ListContainers(docker.ListContainersOptions{})
-	for _, cont := range containers {
-		if strings.HasPrefix(cont.Command, weaveWaitEntrypoint[0]) {
-			proxy.ContainerStarted(cont.ID)
+	for _, apiContainer := range containers {
+		container, err := proxy.client.InspectContainer(apiContainer.ID)
+		if err != nil {
+			Log.Warningf("unable to attach existing container %s since inspecting it failed: %v", apiContainer.ID, err)
```
rade: As I said before, I don't think we should log any warning if the container has simply disappeared.

proxy/proxy.go
```diff
@@ -116,10 +116,16 @@ func NewProxy(c Config) (*Proxy, error) {
 
 func (proxy *Proxy) AttachExistingContainers() {
 	containers, _ := proxy.client.ListContainers(docker.ListContainersOptions{})
-	for _, cont := range containers {
-		if strings.HasPrefix(cont.Command, weaveWaitEntrypoint[0]) {
-			proxy.ContainerStarted(cont.ID)
+	for _, apiContainer := range containers {
```
rade: `apiContainer` sounds strange. `c`.

proxy/proxy.go
```diff
@@ -116,10 +116,16 @@ func NewProxy(c Config) (*Proxy, error) {
...
+		container, err := proxy.client.InspectContainer(apiContainer.ID)
+		if err != nil {
+			Log.Warningf("unable to attach existing container %s since inspecting it failed: %v", apiContainer.ID, err)
+			continue
 		}
+		if containerShouldAttach(container) && (container.State.Running || container.State.Paused) {
+			proxy.attach(container, false)
+		}
+		proxy.notifyWaiters(container.ID)
```
rade: do we need the `notifyWaiters`?

PR #1557, opened 2015-10-19
Add container FQDN to /etc/hosts

### LINKED ISSUE #1497: add container FQDN to /etc/hosts
Is there a way to get weaveDNS to write a containers /etc/hosts file using the container's hostname's FQDN? Not doing so is causing pain in Ambari which wants FQDN's for everything.

inercia:
Fixes #1497

### DIFF
--- a/weave
+++ b/weave
@@ -986,12 +986,16 @@ dns_args() {
 }
 
 etc_hosts_contents() {
-    NAME=$1
+    FQDN=$1
     shift
+    NAME=${FQDN%%.*}
+    HOSTNAMES="$NAME"
+    [ "$NAME" = "$FQDN" -o "$NAME." = "$FQDN" ] || HOSTNAMES="$FQDN $HOSTNAMES"
+
     echo "# created by Weave - BEGIN"
     echo "# container hostname"
     for CIDR in $ALL_CIDRS ; do
-        echo "${CIDR%/*}    $NAME"
+        echo "${CIDR%/*}    $HOSTNAMES"
     done
     echo
     echo "# static names added with --add-host"
@@ -1014,13 +1018,13 @@ EOF
 }
 
 rewrite_etc_hosts() {
-    HOSTS_PATH_AND_NAME=$(docker inspect -f '{{.HostsPath}} {{.Config.Hostname}}' $CONTAINER)
-    HOSTS=${HOSTS_PATH_AND_NAME% *}
-    NAME=${HOSTS_PATH_AND_NAME#* }
+    HOSTS_PATH_AND_FQDN=$(docker inspect -f '{{.HostsPath}} {{.Config.Hostname}}.{{.Config.Domainname}}' $CONTAINER)
+    HOSTS=${HOSTS_PATH_AND_FQDN% *}
+    FQDN=${HOSTS_PATH_AND_FQDN#* }
     CONTAINERS_PATH=$(dirname $HOSTS)
     MNT=/container
     MNT_HOSTS=$MNT/$(basename $HOSTS)
-    CONTENTS="$(etc_hosts_contents $NAME $@)"
+    CONTENTS="$(etc_hosts_contents $FQDN $@)"
     # rewrite /etc/hosts, unlinking the file (so Docker does not modify it again) but
     # leaving it with valid contents...
     docker run --rm \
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -261,14 +261,14 @@ elif [ "$1" = "run" -a -z "$IS_LOCAL" ] ; then
     # we can't rely on our baked-in docker to support those arguments.
     shift 1
     if [ "$1" != "--without-dns" ] ; then
-        DNS_ARGS=$(exec_remote dns-args "$@" || true)
+        eval "$(exec_remote dns-args "$@" 2>/dev/null || true)"
```
rade: no 'eval', please.

weave
```diff
@@ -975,21 +976,20 @@ dns_args() {
         fi
     fi
     if [ -z "$DNS_SEARCH_SPECIFIED" ] ; then
-      if [ -z "$HOSTNAME_SPECIFIED" ] ; then
-        DNS_ARGS="$DNS_ARGS --dns-search=$(dns_domain)"
-      else
-        DNS_ARGS="$DNS_ARGS --dns-search=."
-      fi
+      DNS_DOMAIN=$([ -z "$HOSTNAME_SPECIFIED" ] && dns_domain ||  echo ".")
```
rade: This doesn't seem right; when a user specifies a hostname, it should be that hostname, including any specified domains, which appears in /etc/hosts.

weave
inercia: That's the only solution I find for getting two vars with `exec_remote dns-args`. Otherwise, I would have to do 

```
DNS_ARGS=$(exec_remote dns-args)
DNS_DOMAIN=$(exec_remote dns-domain)
```

Any idea?

rade: I don't understand why you need to pass back the domain in the first place. `weave attach` can work out what the container's configured domain is. It already does that when running `with_container_fqdn`.

rade: To put it another way, why isn't this entire PR simply a matter of adding `{{.Config.Domainname}}` to the `docker inspect` in `rewrite_etc_hosts`, and using the returned info to create the entries?

inercia: Ok, I will get the domain from the container.

inercia: I'm trying to get it working with `docker inspect` and `{{.Config.Domainname}}`, but it does not return the `weave.local` (or whatever the user has configured) domain...

rade: That's weird since it is exactly what `with_container_fqdn` does, which too is called in attach.

inercia: I don't see anything related to the domain when I do

```
$ WEAVE_DEBUG=1 ./weave run -ti  --add-host=static.name:1.1.1.1   --add-host=other.name:2.2.2.2  ubuntu /bin/bash 
[...]
+ with_container_fqdn 07095cd2a0f461b981bf9036ce4cefc0ed52c65273cbf1aae30e69991c67d3ee put_dns_fqdn 10.32.0.1/12
+ CONTAINER=07095cd2a0f461b981bf9036ce4cefc0ed52c65273cbf1aae30e69991c67d3ee
+ COMMAND=put_dns_fqdn
+ shift 2
+ docker inspect --format={{.Config.Hostname}}.{{.Config.Domainname}} 07095cd2a0f461b981bf9036ce4cefc0ed52c65273cbf1aae30e69991c67d3ee
+ CONTAINER_FQDN=07095cd2a0f4.
+ echo 07095cd2a0f4.
+ grep -Eq [^.]+\..+
+ show_addrs 10.32.0.1/12
+ addrs=
+ addrs= 10.32.0.1
+ echo 10.32.0.1
+ echo 07095cd2a0f461b981bf9036ce4cefc0ed52c65273cbf1aae30e69991c67d3ee
07095cd2a0f461b981bf9036ce4cefc0ed52c65273cbf1aae30e69991c67d3ee
```

rade: And neiher should you, since you didn't specify a container name or hostname.

inercia: Ok, so when a user launches `./weave run -ti -h test2.weave.local ...`, he/she should get 

```
# container hostname
10.32.0.1    test2.weave.local test2
```

but if he/she launches `./weave run -ti -h test2 ...`, he/she will get

```
# container hostname
10.32.0.1    test2
```

Is that right? Shouldn't we add `test2.weave.local` too?

rade: no we shouldn't add `test2.weave.local`. Just like we don't register the container in weavedns in that case.

inercia: Ok.

weave
```diff
@@ -986,12 +986,15 @@ dns_args() {
 }
 
 etc_hosts_contents() {
-    NAME=$1
+    FQDN=$1
     shift
+    NAME=${FQDN%%.*}
+    HOSTNAMES=$([ "$NAME" = "$FQDN" -o "$NAME." = "$FQDN" ] && echo "$NAME" || echo "$FQDN $NAME")
```
rade: We should make the logic here and in `with_container_fqdn` consistent. I kinda prefer this version since it doesn't require grep.

Don't like the sub-shell though.

```
NAME=${FQDN%%.*}
HOSTNAMES="$NAME"
[ "$NAME" = "$FQDN" -o "$NAME." = "$FQDN" ] || HOSTNAMES="$FQDN $HOSTNAMES"
```
should work.

inercia: Done.

PR #1561, opened 2015-10-20
Proxy retry attaching containers

### LINKED ISSUE #1556: Proxy should retry attaching containers, after restarting
In case attaching them fails.

paulbellamy:
First experimental pass at having the proxy retry attaching containers if the attach fails.

Nice to haves:
* Detect if there are causes where a container will *never* successfully attach, and handle those.
* Some sort of back-off
* Should also combine multiple requests to attach the same container (by name or id).

Fixes #1556

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
 	. "github.com/weaveworks/weave/common"
@@ -27,6 +28,9 @@ const (
 
 	weaveSock     = "/var/run/weave/weave.sock"
 	weaveSockUnix = "unix://" + weaveSock
+
+	initialInterval = 2 * time.Second
+	maxInterval     = 1 * time.Minute
 )
 
 var (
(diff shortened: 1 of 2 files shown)


bboreham:
I rebased against master and updated to match recent changes.

bboreham:
An example where we can be sure it's never going to attach is when `InspectContainer` returns `NoSuchContainer`.  I think that's more than a nice to have.

bboreham:
Actually when `InspectContainer` returns `NoSuchContainer` the code will not retry, because it's not returned as an error from `attach()`.

bboreham:
Currently the code will abandon retry on various conditions such as container not running.  I cannot think of any further conditions (once we get as far as calling `weave attach`) that should result in abandonment.  Doubtless we will find out in the field.

It will back off to one retry per minute which hopefully isn't too taxing on the machine.

proxy/proxy.go
```diff
@@ -527,3 +557,47 @@ func (proxy *Proxy) updateContainerNetworkSettings(container jsonObject) error {
...
+		if len(pendingRetries) > 0 && nextRetry == nil {
+			nextRetry = time.After(pendingRetries[0].tryAfter.Sub(now))
+		}
+		select {
+		case <-nextRetry:
+			nextRetry = nil
+			a = pendingRetries[0]
+			pendingRetries = pendingRetries[1:]
+		case a = <-proxy.attachContainer:
+			for i, j := range pendingRetries { // remove any existing entry for same id
```
paulbellamy: We might accidentally remove `nextRetry`.

paulbellamy:
The downside of using a `chan attachJob` is that attaching becomes single-threaded. I am debating if something in the rough vain of this would be more performant (and possibly simpler):
```Go
proxy.jobs = map[string]struct{ attachJob, *time.Timer }{
  containerID: struct{
    attachJob: ...,
    Timer: time.AfterFunc(func() {
      err := proxy.attach(containerID, attachJob.OrDie())
      ...
    }),
  },
}

...

func (proxy *Proxy) Stop() {
  for _, job := range proxy.jobs {
    job.Stop()
  }
}
```

proxy/proxy.go
bboreham: You mean "the entry which was previously scheduled to be tried next"? Yes. It seemed innocuous to try the next one at that time.

It seems that `pendingRetries` could be left empty, which would be bad.

Perhaps best to set `nextRetry` to nil if an entry is taken out of the list, so it is re-computed next time round the loop.

paulbellamy:
@bboreham Thoughts on https://github.com/weaveworks/weave/commit/1c7ff83810aad36e26fc75571ad5031d9da95a1e?

proxy/proxy.go
```diff
@@ -72,10 +75,64 @@ type Proxy struct {
...
+
+	j := &attachJob{id: id, tryInterval: initialInterval}
+	j.timer = time.AfterFunc(time.Duration(0), func() {
+		if err := proxy.attach(id, false); err != nil {
+			j.reschedule()
+			return
+		}
+		proxy.notifyWaiters(id, nil)
+	})
+	proxy.attachJobs[id] = j
```
bboreham: I found it quite hard to figure out where the "retry loop" is.  The payload of `time.AfterFunc` gets called, calls `reschedule`, which calls `timer.Reset`, which causes the payload to be called again later.

It might be slightly easier to follow if the `Reset` call was inside the payload.

bboreham:
Should we be re-enabling test 640 or something?

proxy/proxy.go
paulbellamy: Yeah, inlining it does make it easier to follow, thanks. One thing I noticed while doing that, is that on line 92, in the case where an attach already exists, it short-circuits the current "retry timer", but doesn't reset the interval back down to 2 seconds. I'm debating if it should...

paulbellamy:
640 still fails sporadically for me, so left it off...

paulbellamy:
also, rebased.

bboreham:
@paulbellamy is there any more to be done on this?

paulbellamy:
Should be gtg.

bboreham:
I think I'll squash the commits.

Issue #1552, opened 2015-10-16
`weave --local create-bridge` fails when docker isn't running


rade:
```
$ weave --local create-bridge
Post http:///var/run/docker.sock/v1.20/containers/create: dial unix /var/run/docker.sock: no such file or directory.
* Are you trying to connect to a TLS-enabled daemon without TLS?
* Is your docker daemon up and running?
```

This used to work and is important for folks that want to tell docker to use the weave bridge. Does that even work in principle with fastdp? I'm guessing not, in which case we should perhaps spit out an error saying so, with advice to set `WEAVE_NO_FASTDP` (which makes the above work). May turn out to be more trouble than it's worth, in which case this becomes just a "backward compatibility breaking change" for the release notes.

rade:
@dpw @awh thoughts?

dpw:
I'll deal with it later today.

dpw:
I have the "attach docker0 to the weave network" thing I mentioned working.  It needs a bit more polish, so I guess it's too late for 1.2.0.  But it's simple enough that it seems like potential 1.2.1 material if we get any feedback on this.

rade:
Closed by #1562.

PR #1562, opened 2015-10-20
'weave --local create-bridge' is not compatible with fastdp

dpw:
Addresses #1552.

### DIFF
--- a/weave
+++ b/weave
@@ -1635,6 +1635,16 @@ case "$COMMAND" in
     # intentionally undocumented since it assumes knowledge of weave
     # internals
     create-bridge)
+        # This subcommand may be run without the docker daemon, but
+        # fastdp needs to run the router container to setup the ODP
+        # bridge, hence:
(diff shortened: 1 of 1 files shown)


rade:
Hmm. So `weave create-bridge` doesn't do something useful even when docker *is* running?

dpw:
Whether `weave create-bridge` does something useful is questionable.  It creates something, but it's not a normal bridge, and you can't do anything with it that doesn't involve further weave commands that would do `create_bridge` anyway.  So I can't think of a reason why someone would do `weave create-bridge` rather than `weave create-bridge --local`.  Can you?

rade:
>  I can't think of a reason why someone would do `weave create-bridge` rather than `weave create-bridge --local`.  Can you?

I can't. Though perhaps the comment stating "create-bridge is run without the docker daemon" could do with some rephrasing, since create-bridge may be run *with* the docker daemon.

awh:
@dpw would you like to reword the comment before I merge?

dpw:
Done

PR #1573, opened 2015-10-22
Make inspect-rewrite wait for start

### LINKED ISSUE #1549: weaveproxy --rewrite-inspect race at container startup
If you call `docker inspect` on a container right after it starts, you will get the docker IP address until `weave attach` has finished setting up the weave IP address.

This is on both 1.1 and on master.

Weaveproxy could note which containers are still in initialization and hold up any `inspect` requests until initialization has finished.

bboreham:
Fixes #1549

This seems difficult to write a smoke-test for, as `docker start` won't return until the start is finished; the issue shows up if you react to the 'start' event and do an inspect straight away.

### DIFF
--- a/proxy/inspect_container_interceptor.go
+++ b/proxy/inspect_container_interceptor.go
@@ -22,19 +22,20 @@ func (i *inspectContainerInterceptor) InterceptResponse(r *http.Response) error 
 		return err
 	}
 
-	if err := updateContainerNetworkSettings(container); err != nil {
+	if err := updateContainerNetworkSettings(container, i.proxy); err != nil {
 		Log.Warningf("Inspecting container %s failed: %s", container["Id"], err)
 	}
 
 	return marshalResponseBody(r, container)
 }
 
-func updateContainerNetworkSettings(container jsonObject) error {
+func updateContainerNetworkSettings(container jsonObject, proxy *Proxy) error {
 	containerID, err := container.String("Id")
 	if err != nil {
 		return err
 	}
 
+	proxy.waitForStartByIdent(containerID)
 	mac, ips, nets, err := weaveContainerIPs(containerID)
 	if err != nil || len(ips) == 0 {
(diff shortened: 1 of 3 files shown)


rade:
provided this doesn't get more complex, I think it is safe to stick this in 1.2.1.

@bboreham is there anything left to do here?

bboreham:
@rade nothing left from me.

rade:
@paulbellamy pls review. If you are happy with it then lets rebase on 1.2.

proxy/inspect_container_interceptor.go
```diff
@@ -22,19 +22,20 @@ func (i *inspectContainerInterceptor) InterceptResponse(r *http.Response) error
 		return err
 	}
 
-	if err := updateContainerNetworkSettings(container); err != nil {
```
paulbellamy: I'd prefer this as a method on proxy, than taking proxy as an arg.

rade: agreed. especially since it's already shared by multiple interceptors.

proxy/proxy.go
```diff
@@ -357,6 +357,23 @@ func (proxy *Proxy) waitForStart(r *http.Request) {
 	}
 }
 
+// If some other operation is waiting for a container to start, join in the wait
+func (proxy *Proxy) waitForStartByIdent(ident string) {
```
paulbellamy: Should probably return out an error, like `waitForStart`?

proxy/inspect_container_interceptor.go
bboreham: also move the implementation to proxy.go?

paulbellamy: please

bboreham:
Updated in line with review comments and rebased on 1.2.  Will create a new PR to merge into 1.2.

PR #1580, opened 2015-10-23
[proxy] dynamically mount -H dirs into the proxy when launching

### LINKED ISSUE #1579: weave launch-proxy -H unix:///*.sock only works for mounted dirs (e.g. /var/run/weave)
We only mount `/var/run/weave` into the proxy container, so only sockets under that dir will work, at the moment.

paulbellamy:
This lets you use -H unix://*.sock outside /var/run/weave.

The giant gotcha is that if you use `-H /var/run/foo.sock`, you will reintroduce the bug with
namespaces not being GCed. We could make passing a socket in `/var/run`, an error...

Still working on the test...

Fixes #1579

### DIFF
--- a/weave
+++ b/weave
@@ -1277,12 +1277,11 @@ host_arg() {
       exit 1
     fi
     if [ "$host" = "/var/run" ]; then
-      echo "WARNING: When launching the proxy, unix sockets directly in /var/run can lead to a bug with net namespace cleanup." >&2
+      echo "WARNING: Proxying sockets whose immediate parent directory is /var/run is unreliable." >&2
       echo "To work around this, put the unix socket in a subdirectory of /var/run, and symlink it into place." >&2
-      echo "For more info, see https://github.com/weaveworks/weave/issues/1455" >&2
     fi
     PROXY_VOLUMES="$PROXY_VOLUMES -v $host:/host$host"
-    PROXY_ARGS="$PROXY_ARGS -H /host$host/$(basename ${PROXY_HOST#unix://})"
+    PROXY_ARGS="$PROXY_ARGS -H unix:///host$host/$(basename ${PROXY_HOST#unix://})"
   else
     PROXY_ARGS="$PROXY_ARGS -H $1"
   fi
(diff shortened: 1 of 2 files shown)


rade:
> We could make passing a socket in `/var/run`, an error

warning.

weave
```diff
@@ -1266,17 +1266,37 @@ tls_arg() {
...
+      echo "When launching the proxy, unix sockets must be specified as an absolute path." >&2
+      exit 1
+    fi
+    if [ "$host" = "/var/run" ]; then
+      echo "WARNING: When launching the proxy, unix sockets directly in /var/run can lead to a bug with net namespace cleanup." >&2
+      echo "To work around this, put the unix socket in a subdirectory of /var/run, and symlink it into place." >&2
+      echo "For more info, see https://github.com/weaveworks/weave/issues/1455" >&2
+    fi
+    PROXY_VOLUMES="$PROXY_VOLUMES -v $host:/host$host"
+    PROXY_ARGS="$PROXY_ARGS -H /host$host/$(basename ${PROXY_HOST#unix://})"
```
awh: Should this not say `... -H unix:///host$host...`? I see the proxy has some autodetection magic, but we're being explicit about the protocol elsewhere in this script...

weave
```diff
@@ -1266,17 +1266,37 @@ tls_arg() {
...
+  if [ "$PROXY_HOST" != "${PROXY_HOST#unix://}" ]; then
+    host=$(dirname ${PROXY_HOST#unix://})
+    if [ "$host" = "${host#/}" ]; then
+      echo "When launching the proxy, unix sockets must be specified as an absolute path." >&2
+      exit 1
+    fi
+    if [ "$host" = "/var/run" ]; then
+      echo "WARNING: When launching the proxy, unix sockets directly in /var/run can lead to a bug with net namespace cleanup." >&2
+      echo "To work around this, put the unix socket in a subdirectory of /var/run, and symlink it into place." >&2
+      echo "For more info, see https://github.com/weaveworks/weave/issues/1455" >&2
```
rade: This is not helpful. "bug in namespace cleanup" means nothing to the vast majority of users. Plus it suggest that bug is in weave. And linking to an issue that is *closed* conveys the impression that we've actually fixed that bug.

Suggestion: describe the actual user visible problems this is causing.

PR #1585, opened 2015-10-23
Check weaveDNS subdomain on REST PUT

### LINKED ISSUE #1516: DNS entries created for names outside domain
```
$ weave launch
$ weave run -ti --name=foo -h foo.bar ubuntu
55d19a8b1e547c5a5b40fa5c3110cd95fd03db834ced021a6a0991ec0853e272
$ docker exec foo hostname -d
bar
$ weave status dns
foo.bar.     10.32.0.1       55d19a8b1e54 d2:34:fa:c1:1c:77
```

This is wrong. According to [the weaveDNS docs](http://docs.weave.works/weave/latest_release/weavedns.html#usage):

>  if both hostname and container name are specified at the same time the hostname takes precedence; in this circumstance *if the hostname is not in the weaveDNS domain the container will not be registered*, but will still use weaveDNS for resolution.

I am quite sure this used to work correctly; I'm guessing it got broken in the switch to gossipDNS.

As an aside, note that
```
$ weave dns-lookup foo.bar.
```
doesn't work (and neither should it), since resolution only looks up entries in the weaveDNS domain, delegating the rest upstream.

awh:
Fixes #1516.

### DIFF
--- a/nameserver/http.go
+++ b/nameserver/http.go
@@ -35,6 +35,11 @@ func (n *Nameserver) HandleHTTP(router *mux.Router, dockerCli *docker.Client) {
 			return
 		}
 
+		if !dns.IsSubDomain(n.domain, hostname) {
+			n.infof("Ignoring registration %s %s %s (not a subdomain of %s)", hostname, ipStr, container, n.domain)
+			return
+		}
+
 		if err := n.AddEntry(hostname, container, n.ourName, ip); err != nil {
 			n.badRequest(w, fmt.Errorf("Unable to add entry: %v", err))
 			return


nameserver/http.go
```diff
@@ -35,6 +35,11 @@ func (n *Nameserver) HandleHTTP(router *mux.Router, dockerCli *docker.Client) {
 			return
 		}
 
+		if !dns.IsSubDomain(n.domain, hostname) {
+			n.infof("Ignoring name %s, not in %s", hostname, n.domain)
```
rade: Did we used to log this previously, i.e. pre-gossip?

If we really *do* want to log, should we provide more context, e.g. the container and ip?

awh: The message is identical to the one used in 1.0. More context won't hurt though!

awh:
@rade now with added contexts.

rade:
small enough fix to go into 1.2.1. please re-base.

awh:
GitHub YUNO modify PR target branch

PR #1595, opened 2015-10-27
Adding no-dns flag to Help.

enxebre:


rade:
We only document options in the help text that are explained in our main documentation.

What scenario did you encounter where `--no-dns` was useful?

enxebre:
I am just using consul-dns.
Out of cusiosity, if you dont want to document it why dont you just deprecate it?

Aside from this PR are you planning to support weave integration via libnetwork plugin once docker 1.9 get released? so I guess https://github.com/weaveworks/docker-plugin would be the way to go?

Are you planning to integrate with mesos via https://github.com/mesosphere/net-modules or is that something you are not considering?

Thanks!

rade:
> I am just using consul-dns.

Does that *require* turning off weaveDNS?

As an aside, I'm keen to hear about the most important features of consul-dns that you find missing in weaveDNS. Let's not pollute this issue though, so if you don't mind, please post your "top consul-dns features missing in weaveDNS" on [weave-users](https://groups.google.com/a/weave.works/forum/#!forum/weave-users).

>  Out of cusiosity, if you dont want to document it why dont you just deprecate it?

The flag may well be useful in some circumstances.

Re your other questions, please ask on [weave-users](https://groups.google.com/a/weave.works/forum/#!forum/weave-users).

bboreham:
The `--no-dns` flag is not well-tested and we have had no specific use-cases from users to suggest it is widely required, so we are not going to advertise its existence.

Thank you for your contribution, though; @enxebre

PR #1608, opened 2015-10-29
When container is started in host networkMode, we mount a noop weavewait

### LINKED ISSUE #1600: Containers get stuck on weavewait when NetworkMode=host is provided at start
Specific case of https://github.com/weaveworks/weave/issues/1258 which is closed but may be worth reviewing (it was closed with *It's too intrusive a change for 1.0.2.*)

Spawning a container which provides `NetworkMode=host` at start results in weavewait waiting indefinitely.

To reproduce:

```
$ DOCKER_HOST=unix:///var/run/weave/weave.sock docker create --name foo busybox echo foo
$ curl -v --unix-socket /var/run/weave/weave.sock -XPOST http:/v1.15/containers/foo/start -H "Content-Type: application/json" --data-binary '{"HostConfig": { "NetworkMode": "host"} }'
$ ps auxf | grep 'echo foo'
root     16605  0.0  0.0   4880  1808 ?        Ssl  14:41   0:00  \_ /w/w echo foo
```

The weave proxy seems to recognize the host network mode ...
```
INFO: 2015/10/28 14:39:44.275603 POST /v1.15/containers/foo/start
INFO: 2015/10/28 14:39:44.343674 Leaving container c8af2e50a472eaf274b2e2fbd0b468debb814f4fde90634e161c78ad16ca71df alone because the container was created with the '--net=host'
```

... but it doesn't really *leave the container alone*.

In an offline discussion, @paulbellamy mentioned this happens because by the time the container starts it was already created with the weavewait entrypoint.

I experienced this when trying to run Kubernetes with Weave (which may not be a sane thing to do but that's besides the point).

Related: https://github.com/kubernetes/kubernetes/issues/12263

paulbellamy:
Fixes #1600

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -70,6 +70,7 @@ type Proxy struct {
 	dockerBridgeIP      string
 	hostnameMatchRegexp *regexp.Regexp
 	weaveWaitVolume     string
+	weaveWaitNoopVolume string
 	normalisedAddrs     []string
 	waiters             map[*http.Request]*wait
 }
@@ -106,7 +107,7 @@ func NewProxy(c Config) (*Proxy, error) {
 		return nil, err
 	}
 
-	if err = p.findWeaveWaitVolume(); err != nil {
+	if err = p.findWeaveWaitVolumes(); err != nil {
 		return nil, err
 	}
 
@@ -135,23 +136,31 @@ func (proxy *Proxy) Dial() (net.Conn, error) {
 	return net.Dial("unix", dockerSock)
 }
 
-func (proxy *Proxy) findWeaveWaitVolume() error {
+func (proxy *Proxy) findWeaveWaitVolumes() error {
+	var err error
+	if proxy.weaveWaitVolume, err = proxy.findVolume("/w"); err != nil {
+		return err
+	}
+	proxy.weaveWaitNoopVolume, err = proxy.findVolume("/w-noop")
+	return err
+}
+
+func (proxy *Proxy) findVolume(v string) (string, error) {
 	container, err := proxy.client.InspectContainer("weaveproxy")
 	if err != nil {
-		return fmt.Errorf("Could not find the weavewait volume: %s", err)
+		return "", fmt.Errorf("Could not find the weavewait volume: %s", err)
 	}
 
 	if container.Volumes == nil {
-		return fmt.Errorf("Could not find the weavewait volume")
+		return "", fmt.Errorf("Could not find the weavewait volume")
 	}
 
-	volume, ok := container.Volumes["/w"]
+	volume, ok := container.Volumes[v]
 	if !ok {
-		return fmt.Errorf("Could not find the weavewait volume")
+		return "", fmt.Errorf("Could not find the weavewait volume")
 	}
 
-	proxy.weaveWaitVolume = volume
-	return nil
+	return volume, nil
 }
 
 func (proxy *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
@@ -425,7 +434,7 @@ func (proxy *Proxy) attachRouter(container *docker.Container) error {
 
 func (proxy *Proxy) weaveCIDRs(networkMode string, env []string) ([]string, error) {
 	if networkMode == "host" || strings.HasPrefix(networkMode, "container:") {
-		return nil, fmt.Errorf("the container was created with the '--net=%s'", networkMode)
+		return nil, fmt.Errorf("the container has '--net=%s'", networkMode)
 	}
 	for _, e := range env {
 		if strings.HasPrefix(e, "WEAVE_CIDR=") {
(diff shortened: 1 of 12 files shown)


proxy/proxy.go
```diff
@@ -424,6 +433,14 @@ func (proxy *Proxy) weaveCIDRs(networkMode string, env []string) ([]string, erro
 }
 
 func (proxy *Proxy) addWeaveWaitVolume(hostConfig jsonObject) error {
+	return proxy.addVolume(hostConfig, proxy.weaveWaitVolume, "/w", "ro")
+}
+
+func (proxy *Proxy) addWeaveWaitNoopVolume(hostConfig jsonObject) error {
+	return proxy.addVolume(hostConfig, proxy.weaveWaitNoopVolume, "/w", "ro")
+}
```
rade: ditch those functions.

test/666_abuse_of_start_test.sh
```diff
@@ -36,12 +44,25 @@ assert "docker_on $HOST1 inspect -f '{{.HostConfig.MemorySwap}}' c4" "-1"
 # Start c6 with both named and unnamed HostConfig
 proxy docker_on $HOST1 create --name=c6 $SMALL_IMAGE $CHECK_ETHWE_UP
 proxy docker_api_on $HOST1 POST /containers/c6/start '{"NetworkMode": "container:c2", "HostConfig": {"NetworkMode": "container:c1"}}'
-check_hostconfig c6 container:c1
+check_dns c3
```
rade: c6?

test/666_abuse_of_start_test.sh
```diff
@@ -8,25 +8,33 @@ weave_on $HOST1 launch
...
+    docker_on $HOST1 attach c2 >/dev/null 2>&1 || true # Wait for container to exit
+    assert "docker_on $HOST1 inspect -f '{{.HostConfig.NetworkMode}} {{.State.Running}} {{.State.ExitCode}}' $1" "$2 false 0"
 }
 
 # Start c2 with a sneaky HostConfig
 proxy docker_on $HOST1 create --name=c2 $SMALL_IMAGE $CHECK_ETHWE_UP
 proxy docker_api_on $HOST1 POST /containers/c2/start '{"NetworkMode": "container:c1"}'
-check_hostconfig c2 container:c1
+check_dns c2
+check_networkmode c2 container:c1
```
rade: It's annoying that this ends up doing two 'attach' and two 'inspect'. I'd just have a single function, with an optional param to indicate whether to check dns.

test/666_abuse_of_start_test.sh
```diff
@@ -8,25 +8,33 @@ weave_on $HOST1 launch
 docker_bridge_ip=$(weave_on $HOST1 docker-bridge-ip)
 proxy_start_container $HOST1 --name=c1
 
-check_hostconfig() {
+check_dns() {
     docker_on $HOST1 attach c2 >/dev/null 2>&1 || true # Wait for container to exit
```
rade: eek. that looks like a bug. should be `$1` instead of `c2`, surely.

rade:
you accidentally checked in `prog/weavewait/weavewait_noop`.

proxy/proxy.go
```diff
@@ -423,7 +432,7 @@ func (proxy *Proxy) weaveCIDRs(networkMode string, env []string) ([]string, erro
 	return nil, nil
 }
 
-func (proxy *Proxy) addWeaveWaitVolume(hostConfig jsonObject) error {
+func (proxy *Proxy) addVolume(hostConfig jsonObject, source, target, mode string) error {
```
rade: This should just be a function, not a method on `Proxy`.

proxy/proxy.go
```diff
@@ -495,3 +508,14 @@ func (proxy *Proxy) getDNSDomain() (domain string) {
...
+func (proxy *Proxy) networkModeIsAttachable(networkMode string) (bool, error) {
+	for strings.HasPrefix(networkMode, "container:") {
+		container, err := proxy.client.InspectContainer(strings.TrimPrefix(networkMode, "container:"))
+		if err != nil || container.HostConfig == nil {
+			return false, err
+		}
+		networkMode = container.HostConfig.NetworkMode
+	}
+	return networkMode != "host", nil
+}
```
rade: There is a problem with this....

The parent container could be restarted with a different network mode. This is made worse by the fact that the reference can be a name, so you could actually change the parent completely.

My experiments indicate that this does not affect a running container. But does when the container is restarted. Which, if it happens w/o going through the proxy, e.g. due a restart policy, means we can end up with the wrong weavewait. Not much we can do about that, I suppose, but probably merits a comment.

Another edge case is that the parent container may not exist. Looks like we error at that point; shouldn't we leave that to docker? Though it is of course possible that the parent may appear by the time the request reaches docker. hmm.

rade:
When we deal with network modes at container creation time, we do not insert a weavewait entrypoint when the mode is `container:...`. Even though that container could be weave networked.

Here, at start, we mount weavewait (instead of weavewait-noop) when the mode is `container:...` unless the parent is on the host network.

That is inconsistent.

Ideally in both cases we'd want to weavewait if the parent container is weave networked. But we have no reliable way of figuring that out, I think.

So perhaps our logic should simply be, "if the network mode is `host` or `container:...`" then don't weavewait.

rade:
Another problem: the volume mount swivelling won't alter the entrypoint, which is all the proxy looks at when deciding whether to attach. Hmm. Except it also looks at the NetworkMode. arghhh, this is all hideously complicated.

proxy/proxy.go
paulbellamy: A restart policy won't change the networkMode, though. Only an api stop/start with the hostconfig provided.

We could do these checks every time a container is started, and send a hostConfig if needed (regardless of whether the user sent one).

paulbellamy:
per f2f discussion:

I agree we should simplify the logic to just not weavewait if `NetworkMode` is `host` or `container:...`

paulbellamy:
Eugh, but that also means we won't set up any dns on these container either...

rade:
anything left to do here? The failures should get cured with a rebase.

paulbellamy:
Nope, should be ready to go.

PR #1609, opened 2015-10-29
trying a fix for the intermittent failure in test/640

### LINKED ISSUE #1588: (spurious?) failures in 640_proxy_restart_reattaches_test
https://circleci.com/gh/weaveworks/weave/3387
https://circleci.com/gh/weaveworks/weave/3378
https://circleci.com/gh/weaveworks/weave/3375

paulbellamy:
May fix #1588 

Based on the systemctl error, it looks like sudo can't find the systemctl on it's path, so we could pass in the absolute path to it.

Not sure why this would only happen sometimes on GCE?

Waiting for this to run on circle a couple times...

### DIFF
--- a/test/640_proxy_restart_reattaches_test.sh
+++ b/test/640_proxy_restart_reattaches_test.sh
@@ -31,7 +31,7 @@ sleep 1
 check_attached
 
 # Restart docker itself, using different commands for systemd- and upstart-managed.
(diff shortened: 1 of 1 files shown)


rade:
iirc @bboreham says that the systemctl whining has nothing to do with the error, i.e. it is expected.

rade:
though I would say that the test really should suppress that sort of noise.

Issue #1610, opened 2015-10-29
Not sorted DNS data in outbound gossip merge


bboreham:
While investigating #1603, a similar issue was found:

- If gossip data is queued outbound, and we have more to send, `GossipData.Merge()` gets called.  This checks that `g.Entries` is sorted case-sensitively, then calls `g.Entries.merge()` which checks that it is sorted case-insensitively.  With entries of mixed case, it is impossible to satisfy both checks.

(not seen in the wild - requires blocking-back of outbound data together with mixed-case names)

bboreham:
Updated description; this code will fail faster than I thought.

rade:
Fixed by #1615.

PR #1611, opened 2015-10-29
Hold DNS entries sorted case-insensitively until just before sending

### LINKED ISSUE #1603: periodic panic: Not sorted! running 1.2
```
panic: Not sorted!

goroutine 222 [running]:
github.com/weaveworks/weave/nameserver.checkAndPanic(0x7fdaf108e090, 0xc821bd0680)
	/home/bryan/golang/src/github.com/weaveworks/weave/nameserver/entry.go:117 +0x6d
github.com/weaveworks/weave/nameserver.(*GossipData).Encode(0xc82183d620, 0x0, 0x0, 0x0)
	/home/bryan/golang/src/github.com/weaveworks/weave/nameserver/entry.go:246 +0xc7
github.com/weaveworks/weave/router.(*GossipChannel).sendDown.func1(0x7fdaf1915510, 0xc82183d620)
	/home/bryan/golang/src/github.com/weaveworks/weave/router/gossip_channel.go:159 +0x69
github.com/weaveworks/weave/router.(*GossipSender).run(0xc8218dbd40)
	/home/bryan/golang/src/github.com/weaveworks/weave/router/gossip.go:61 +0xf1
created by github.com/weaveworks/weave/router.NewGossipSender
	/home/bryan/golang/src/github.com/weaveworks/weave/router/gossip.go:50 +0xf3
```

bboreham:
Instead of sorting when creating the outbound `GossipData` object, we sort in `Encode()`.
Fixes #1603 and #1610

### DIFF
--- a/nameserver/entry.go
+++ b/nameserver/entry.go
@@ -241,6 +241,15 @@ func (g *GossipData) Merge(o router.GossipData) {
 	}
 }
 
+func (g *GossipData) Decode(msg []byte) error {
+	if err := gob.NewDecoder(bytes.NewReader(msg)).Decode(g); err != nil {
+		return err
+	}
+
+	sort.Sort(CaseInsensitive(g.Entries))
+	return nil
+}
+
 func (g *GossipData) Encode() [][]byte {
 	// Make a copy so we can sort: all outgoing data is sent in case-sensitive order
 	g2 := GossipData{Timestamp: g.Timestamp, Entries: make(Entries, len(g.Entries))}
--- a/nameserver/entry.go
+++ b/nameserver/entry.go
@@ -233,8 +233,8 @@ type GossipData struct {
 }
 
 func (g *GossipData) Merge(o router.GossipData) {
-	checkAndPanic(CaseSensitive(g.Entries))
-	defer func() { checkAndPanic(CaseSensitive(g.Entries)) }()
+	checkAndPanic(CaseInsensitive(g.Entries))
+	defer func() { checkAndPanic(CaseInsensitive(g.Entries)) }()
 	other := o.(*GossipData)
 	g.Entries.merge(other.Entries)
 	if g.Timestamp < other.Timestamp {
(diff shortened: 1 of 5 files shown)


nameserver/entry.go
```diff
@@ -233,8 +233,8 @@ type GossipData struct {
 }
 
 func (g *GossipData) Merge(o router.GossipData) {
-	checkAndPanic(CaseSensitive(g.Entries))
-	defer func() { checkAndPanic(CaseSensitive(g.Entries)) }()
+	checkAndPanic(CaseInsensitive(g.Entries))
+	defer func() { checkAndPanic(CaseInsensitive(g.Entries)) }()
```
rade: I don't think we need this check. Instead `Entries.merge` should check that its argument is correctly sorted (in addition to the existing pre/post check on the method receiver).

rade: which also means we can inline `check` in `checkAndPanic` and that in turn in `Entries.checkAndPanic`.

rade: actually, ignore that last bit.

rade:
needs to be branched off 1.2

rade:
I reckon the gossip data decoding code should be moved near the encoding code. e.g.
```
func (g *GossipData) Decode(msg []byte) error {
	if err := gob.NewDecoder(bytes.NewReader(msg)).Decode(g); err != nil {
		return err
	}
	sort.Sort(CaseInsensitive(g.Entries))
	return nil
}
```
and then call it from `Nameserver.receiveGossip` like this:
```
	var gossip GossipData
	if err := gossip.Decode(msg); err != nil {
		return nil, nil, err
	}
	if delta := gossip.Timestamp - now(); delta > gossipWindow || delta < -gossipWindow {
	}
        ...
```

(I *think* that works)

rade:
the unit tests failed.

nameserver/entry.go
bboreham: Just clarifying for my own understanding: `GossipData.Merge` didn't need to check `g.Entries` because `Entries.merge()` already did that.  But `Entries.merge()` also relies on its `incoming` argument being sorted, so it should check that.

Note that `Entries.merge()` is called in two ways: for internally-generated outgoing data and for incoming gossip from another peer.  In the latter case we re-sort the incoming data so it should pass the test. We have traditionally preferred to raise an error and drop the connection when bad data is received from another peer, as opposed to panicking.

rade: Correct.

> We have traditionally preferred to raise an error and drop the connection when bad data is received from another peer, as opposed to panicking.

That would entail adding error returns to a whole bunch of methods. And I don't think that is justifiable here. We could, alternatively, not bother checking.

bboreham: As I say, we have just sorted the data so will not panic merely because a peer has sent un-sorted data.

bboreham:
Wrong branch; replaced by #1615

nameserver/entry_test.go
```diff
@@ -130,3 +130,31 @@ func TestLookup(t *testing.T) {
...
+		Entry{Hostname: "A"},
+		Entry{Hostname: "B"},
+		Entry{Hostname: "c"},
+		Entry{Hostname: "D"},
+		Entry{Hostname: "E"},
+		Entry{Hostname: "f"},
+	}}
+
+	require.Equal(t, expected, g1)
+}
```
rade: Not essential, but... this should look like
```
g1 := mumble("AcDf")
g2 := mumble("BEf")
g1.Merge(&g2)
require.Equal(t, mumble("ABcDEf"), g1)
```
but only if you can come up with a better name for the function ;)

Issue #1599, opened 2015-10-28
router fatal netlink error if kernel does not support openvswitch vxlan vports


dpw:
See https://groups.google.com/a/weave.works/forum/#!topic/weave-users/vSn7vizptt8

You can reproduce this on any distro by doing:

* `rmmod vport_vxlan`
* Temporarily doing `mv /lib/modules/<kernel version>/kernel/net/openvswitch/vport-vxlan.ko{,.disable}`

PR #1619, opened 2015-10-30
Probe whether the kernel has ODP vxlan support

dpw:
It's possible to build that kernel with openvswitch, and vxlan, but
without the vport_vxlan module that glues them together.  And at
least one distro has actually done this.

Addresses #1599.

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -13,7 +13,6 @@ import (
 	"github.com/davecheney/profile"
 	"github.com/docker/docker/pkg/mflag"
 	"github.com/gorilla/mux"
-	"github.com/weaveworks/go-odp/odp"
 
 	. "github.com/weaveworks/weave/common"
 	"github.com/weaveworks/weave/common/docker"
@@ -119,8 +118,12 @@ func main() {
 		os.Exit(0)
 
 	case createDatapath:
-		err := weave.CreateDatapath(datapathName)
-		if odp.IsKernelLacksODPError(err) {
+		err, odp_supported := weave.CreateDatapath(datapathName)
+		if !odp_supported {
+			if err != nil {
+				Log.Error(err)
+			}
+
 			// When the kernel lacks ODP support, exit
 			// with a special status to distinguish it for
 			// the weave script.
--- a/router/odp.go
+++ b/router/odp.go
@@ -1,25 +1,58 @@
 package router
 
 import (
+	"fmt"
+	"net"
+	"syscall"
+
 	"github.com/weaveworks/go-odp/odp"
 )
 
 // ODP admin functionality
 
-func CreateDatapath(dpname string) error {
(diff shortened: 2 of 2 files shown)


prog/weaver/main.go
```diff
@@ -119,8 +118,12 @@ func main() {
 		os.Exit(0)
 
 	case createDatapath:
-		err := weave.CreateDatapath(datapathName)
-		if odp.IsKernelLacksODPError(err) {
+		err, odp_supported := weave.CreateDatapath(datapathName)
+		if !odp_supported {
+			if err != nil {
+				Log.Fatal(err)
+			}
```
awh: Can we use `checkFatal(err)` here?

router/odp.go
```diff
@@ -1,25 +1,57 @@
...
 	if err != nil && !odp.IsDatapathNameAlreadyExistsError(err) {
-		return err
+		return err, true
+	}
+
+	// Pick an ephemeral port number to use in probing for vxlan
+	// support.
+	udpconn, err := net.ListenUDP("udp4", nil)
+	if err != nil {
+		return err, true
```
awh: It feels odd to return `true` here (even though it makes no difference in practice because the caller is always checking `err`)

prog/weaver/main.go
dpw: Hmm.  No, because we want to exit with the distinguished status code.  Which that code doesn't do - `Fatal` should be `Error`.

router/odp.go
dpw: It's significant.  The boolean determines how errors are treated. There are four possibilities:

error|supported|meaning
-------|-------------|------------
nil     | true       | ODP supported, everything fine
!= nil | true      | there was an error, don't really know whether ODP is supported or not
nil     | false    | ODP not supported (so use a normal bridge)
!= nil | false   | ODP not supported (so use a normal bridge), and the error describes why

The (error, bool) encoding seemed like the natural one in go, but I see how it isn't obvious (particularly given the tendency for an error return value to mean that other return values should be ignored - I'd hoped to signal that is not the case here with the unusual ordering of the return values).

prog/weaver/main.go
dpw: Fixed

rade:
This got merged into master, but the issue it is meant to be resolving has 1.2.1 milestone.

rade:
Hmm. But the issue indicates that something *did* get merged into 1.2.1.

@awh please explain what happened here.

rade:
> But the issue indicates that something *did* get merged into 1.2.

Well, the [commit](https://github.com/weaveworks/weave/commit/ccae8d42fd845e3bf25030e4e04486f802a4e6ff) message says that, but it's actually a merge into master. Except it also show up on the 1.2 branch.

What happened?

awh:
> What happened?

I merged it into 1.2 on the command line, then 1.2->master.

PR #1622, opened 2015-11-02
Update unit test for miekg DNS API change

awh:
Fix unit test breakage caused by https://github.com/miekg/dns/commit/2d2c2ebcfc8f39f0393c32b2728e7757abc93761


tomwilkie:
Is there anyway we could have prevented this? :trollface:

bboreham:
Do you think we should also check for `ErrTruncated` in `handleRecursive()`?  ISTM that it's better to return a truncated response than no response at all.  Also more faithful to other DNS implementations.

awh:
> Do you think we should also check for ErrTruncated in handleRecursive()?

Good spot - I have pushed an update.

PR #1624, opened 2015-11-02
Let IPAM pick any peer to transfer ownership to, on reset

### LINKED ISSUE #1593: weave script hangs in IP address allocation after rolling restart
With 1.2.0, if I:

* Start two peers (one just with `weave launch`, one with `weave launch <other's ip>`
* Run a short-lived container on each host in turn: `docker $(weave config) run --name=foo --rm ubuntu ls` (naming the containers seems to be necessary).
* On each host in turn, do `weave reset` and `weave launch` (the latter with the original args).  I.e. at least one router is running throughout.

then further attempt to start containers on the weave network hang (e.g. `weave run -itd ubuntu` on either host).  Seems to be 100% reproducible.  Last few lines with WEAVE_DEBUG=1:

```
+ IPAM_URL=/ip/ab7015b9774d10f7eb365c9f5aa0264d5e6887dc3fb0dc12bc31944d87aefa8b
+ call_weave POST /ip/ab7015b9774d10f7eb365c9f5aa0264d5e6887dc3fb0dc12bc31944d87aefa8b?check-alive=true
+ container_ip weave weave container is not present. Have you launched it? weave container is not running.
+ docker inspect --format={{.State.Running}} {{.NetworkSettings.IPAddress}} weave
+ status=true 
+ CONTAINER_IP=127.0.0.1
+ http_call 127.0.0.1:6784 POST /ip/ab7015b9774d10f7eb365c9f5aa0264d5e6887dc3fb0dc12bc31944d87aefa8b?check-alive=true
+ addr=127.0.0.1:6784
+ http_verb=POST
+ url=/ip/ab7015b9774d10f7eb365c9f5aa0264d5e6887dc3fb0dc12bc31944d87aefa8b?check-alive=true
+ shift 3
+ curl --connect-timeout 3 -s -S -X POST http://127.0.0.1:6784/ip/ab7015b9774d10f7eb365c9f5aa0264d5e6887dc3fb0dc12bc31944d87aefa8b?check-alive=true
```

It's not caused by IPAM consensus woes.  `weave status`:

```
...
       Service: ipam
     Consensus: achieved
         Range: [10.32.0.0-10.48.0.0)
 DefaultSubnet: 10.32.0.0/12
...
```

Nothing obviously interesting in the weave logs.  Last few lines are

```
INFO: 2015/10/27 09:38:39.678852 [allocator]: Allocate request for 973fd8e60a84dc00aa83bef5c4e7c71015b5229c3d9c564586f7e9955699bb57 cancelled
INFO: 2015/10/27 09:41:22.981950 [allocator]: Allocate request for 2ca43780f78dc2809b20ab79401f2d5f7257d37ba0d28b6d3b1ca2e7a0d297fc cancelled
INFO: 2015/10/27 09:42:52.937975 [allocator]: Allocate request for cbb3c83b9b705e327f948d104bcefc5faa954919b408b9e7258452b9c13c43fe cancelled
INFO: 2015/10/27 09:44:52.927980 [allocator]: Allocate request for ab7015b9774d10f7eb365c9f5aa0264d5e6887dc3fb0dc12bc31944d87aefa8b cancelled
```

Due to me killing the hung weave scripts.  So it is getting the HTTP requests, and getting stuck while processing them.

I have my hands full with other stuff today, so I'm not planning to look further into this right now.  Might be good if someone else confirms.

bboreham:
Fixes #1593

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -56,20 +56,20 @@ type Allocator struct {
 	paxosTicker      *time.Ticker
 	shuttingDown     bool // to avoid doing any requests while trying to shut down
 	now              func() time.Time
-	namesFunc        func() map[router.PeerName]struct{}
+	peerValidFunc    func(router.PeerName) bool
 }
 
 // NewAllocator creates and initialises a new Allocator
-func NewAllocator(ourName router.PeerName, ourUID router.PeerUID, ourNickname string, universe address.Range, quorum uint, names func() map[router.PeerName]struct{}) *Allocator {
+func NewAllocator(ourName router.PeerName, ourUID router.PeerUID, ourNickname string, universe address.Range, quorum uint, validPeer func(name router.PeerName) bool) *Allocator {
 	return &Allocator{
-		ourName:   ourName,
-		universe:  universe,
-		ring:      ring.New(universe.Start, universe.End, ourName),
-		owned:     make(map[string][]address.Address),
-		paxos:     paxos.NewNode(ourName, ourUID, quorum),
-		nicknames: map[router.PeerName]string{ourName: ourNickname},
-		now:       time.Now,
-		namesFunc: names,
+		ourName:       ourName,
+		universe:      universe,
+		ring:          ring.New(universe.Start, universe.End, ourName),
+		owned:         make(map[string][]address.Address),
+		paxos:         paxos.NewNode(ourName, ourUID, quorum),
+		nicknames:     map[router.PeerName]string{ourName: ourNickname},
+		now:           time.Now,
+		peerValidFunc: validPeer,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


rade:
I am having 2nd thoughts about this. Assuming that all peers run IPAM is just wrong.

Instead I wonder whether we can tweak the existing nickname gossipping...

1. ensure that every peer running ipam includes themselves in the nickname map
2. don't prune nicknames that refer to valid peers

Then we can choose a transfer peer first from the intersection of peers in the ring and currently valid peers, and, if that fails, secondly, from the intersection of the nickname map keys and the currently valid peers, and, if that fails, thirdly, from the set of peers in the ring, and, if that fails, fourthly, from the set nickname map keys.

in other words we favour alive peers that actively participate in IPAM (i.e. have entries), over alive peers that passively participate in IPAM (i.e. have IPAM enabled but have no entries), over disappeared peers that actively participated in IPAM (i.e. still have entries), over disappeared peers that that passively participated in IPAM.

NB: this also addresses another issue with the current logic: rmpeer only recognises nicknames of (alive or disappeared) *active* IPAM participants.

bboreham:
Note that:

> 1. ensure that every peer running ipam includes themselves in the nickname map

is already there, because peer names are only ever added to `nicknames` structure by each peer including themselves at startup, in `NewAllocator()`; thereafter they are merged from other peers

ipam/allocator.go
```diff
@@ -56,18 +56,20 @@ type Allocator struct {
...
-		nicknames: map[router.PeerName]string{ourName: ourNickname},
-		now:       time.Now,
+		ourName:     ourName,
+		universe:    universe,
+		ring:        ring.New(universe.Start, universe.End, ourName),
+		owned:       make(map[string][]address.Address),
+		paxos:       paxos.NewNode(ourName, ourUID, quorum),
+		nicknames:   map[router.PeerName]string{ourName: ourNickname},
+		now:         time.Now,
+		isKnownPeer: isKnownPeer,
```
rade: nitpick: `now` should come last, since it is not initialised from any of the arguments. (yes, `owned` isn't either, but that is different ;)

ipam/allocator.go
```diff
@@ -282,7 +311,8 @@ func (alloc *Allocator) Shutdown() {
 		alloc.shuttingDown = true
 		alloc.cancelOps(&alloc.pendingClaims)
 		alloc.cancelOps(&alloc.pendingAllocates)
-		if heir := alloc.ring.PickPeerForTransfer(); heir != router.UnknownPeerName {
+		heir := alloc.pickPeerForTransfer()
+		if heir != router.UnknownPeerName {
```
rade: why the change?

rade:
I am worried about possible deadlocks here, seeing as `Peers.Fetch` requires the lock. @bryan have you analysed this and convinced yourself it is safe?

rade:
Note to self: when merging into master, reduce the two definitions of `isKnownPeer` to one.

bboreham:
Deadlocks: that is an interesting question.  The router communicates to IPAM via the gossipper interface, and IPAM calls the router via the `isKnownPeer()` function.  Allocator runs on its own goroutine and doesn't do any explicit locking of its own, though the 'actor' chan can block.

Since the gossip callbacks are made from the router's `receiveTCP` goroutine, while not holding any locks, it seems that these two cannot create a deadly embrace.

PR #1631, opened 2015-11-03
Observers for DNS names

### LINKED ISSUE #1591: weaveDNS API for watching names
weaveDNS could provide an API for _watching_ names, so a client could keep a connection to weaveDNS and be notified when an IP has been added/removed to/from a name. This would be useful for [other systems](https://github.com/inercia/weave-reverse-proxy).

It could be implemented as a regular REST API with some kind of long-polling (we could be inspired by the _etcd_ protocol, but maybe it would be too complicated and it would make things harder on the client side), or by something like Redis, etc

inercia:
This PR implements 
- a simple mechanism for observing names in the DNS database.
- a websockets interface on the nameserver HTTP router for registering clients that want to observe names as well as for receiving updates.

I have taken several _short paths_ in this PR:
- updates are sent not as _diffs_ but as the whole list of IPs for the name being observed. This should not be a problem as we do not expect too many IPs for a name and it is a local TCP connection anyway, and _diffs_ would make things much harder on the client side.
- observers notifications are triggered after every DNS database modification. I didn't want to do invasive changes in the database code, and this shouldn't be a problem as we do not expect many observers anyway. This adds `n*lookup()` to `add`/`remove` operations in the database, where `n` is the number of different hostnames being observed...

Related to inercia/weave-nginx#3 (nginx side of things)

Closes #1591

### DIFF
--- a/nameserver/http.go
+++ b/nameserver/http.go
@@ -3,14 +3,47 @@ package nameserver
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
+	"time"
 
 	"github.com/gorilla/mux"
+	"github.com/gorilla/websocket"
 	"github.com/miekg/dns"
 
 	"github.com/weaveworks/weave/common/docker"
 	"github.com/weaveworks/weave/net/address"
 )
+
+const (
+	// Maximum message size allowed from peer.
+	maxMessageSize = 512
+
+	// Time allowed to write a message to the peer.
+	writeWait = 10 * time.Second
+
+	// Time allowed to read the next pong message from the peer.
+	pongWait = 60 * time.Second
+
+	// Send pings to peer with this period. Must be less than pongWait.
+	pingPeriod = (pongWait * 9) / 10
+)
+
+var websocketsUpgrader = websocket.Upgrader{
+	ReadBufferSize:  1024,
+	WriteBufferSize: 1024,
+	CheckOrigin:     func(r *http.Request) bool { return true },
+}
+
+// observation request
+type ObserveRequest struct {
+	Name string
+}
+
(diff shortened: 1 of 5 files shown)


bboreham:
It's not clear to me: what happens when the listener on the end of the websocket doesn't keep up with the data rate?

inercia:
> It's not clear to me: what happens when the listener on the end of the websocket doesn't keep up with the data rate?

Are you concerned about the server or the client side of that? I think that a high data rate could cause problems on both sides :blush:

On the server side, I can replace the _updates channel_ for a `value`+`mux`, so only the last update will be sent. And some kind of _throttling_ mechanism would definetively make things safer... This is an initial approach to the problem and I didn't want to overengineer it. 

In general, this is not prepared for _a)_ a large number of observers _b)_ too frequent changes in the observed DNS entries. But it should be enough for a simple case (like a load-balancer scenario)..

bboreham:
Doesn't frequent polling also work well for a simple case like a load-balancer?

inercia:
Well, as you probably know, the problem with frequent polling is in the first word: the _frequency_. A long frequency and you will have stale records, a short one and you will be overloading the DNS server.

As I said, the only cases where this could show scalability problems would be with  "_a)_ a large number of observers _b)_ too frequent changes in the observed DNS entries". And both problems would be easy to fix! So, when I talk about a _simple case_, I mean that load-balancing `mywebserver.weave.local`, even to millions of users, would not be a problem (as long as your containers are not dying&restarting) and the LB would receive immediate updates and it would never have stale records...

bboreham:
We haven't progressed this in a year; closing.

Issue #1546, opened 2015-10-16
connect to http control interface at localhost instead of docker ip


rade:
Since fastdp landed, the router is running in the host network namespace. This means that the http control interface is on localhost:6783. There is therefore no need to go through contortions in the weave script to obtain the weave container's docker ip. Furthermore, it also means things like `weave --local expose` will work against a non-dockerized weaver, i.e. the script can invoke IPAM w/o any interactions with docker.

dpw:
I wouldn't recommend adding dependencies on the fact that the router is running in the host network namespace.  fastdp led to that because use of the kernel's openvswitch module strongly suggested that approach.  Having it in a netns does seem preferable, and I would not claim that we have exhausted all avenues to achieve that.

rade:
> Having it in a netns does seem preferable, and I would not claim that we have exhausted all avenues to achieve that.

That very much seems like a nice-to-have. There isn't even an issue for it!

If we do ever bother to work on that, and come up with a solution, then we can put the contortions back at that point.

awh:
At the moment we fall back to running the router in a container netns if `WEAVE_NO_FASTDP` is defined or if the kernel doesn't have ODP support. Are you suggesting that we run the router in the host netns unconditionally?

rade:
> Are you suggesting that we run the router in the host netns unconditionally?

Yes.

rade:
If we do this we can get rid of most of the code in [proxy.getDNSDomain](https://github.com/weaveworks/weave/blob/master/proxy/proxy.go#L475) - we no longer need to inspect the weave container in order to figure out what IP to find the weave control port on.

That code is run for every container created through the proxy, so this is a nice win.

PR #1637, opened 2015-11-04
Bind router HTTP listener to localhost when fastdp enabled

### LINKED ISSUE #1632: with FASTDP, application containers can control weave
```
$ weave reset
$ weave launch
$ weave docker-bridge-ip
172.17.42.1
$ docker $(weave config) run -ti ubuntu
root@cc4d33557cc4:/# nc 172.17.42.1 6784
foo
HTTP/1.1 400 Bad Request
```

Now, the same is possible w/o FASTDP...

```
$ weave reset
$ WEAVE_NO_FASTDP=1 weave launch
$ docker inspect -f '{{.NetworkSettings.IPAddress}}' weave
172.17.0.29
$ docker $(weave config) run -ti ubuntu
root@c0197cdff032:/# nc 172.17.0.29 6784
foo
HTTP/1.1 400 Bad Request
```

but that can be disabled by running docker with `-icc=false`:

```
$ weave reset
$ WEAVE_NO_FASTDP=1 weave launch
$ docker inspect -f '{{.NetworkSettings.IPAddress}}' weave
172.17.0.1
$ docker $(weave config) run -ti ubuntu
root@e98d44350b9c:/# nc 172.17.0.1 6784
foo
(no response)
```

whereas with FASTDP we still get the same result

```
$ weave reset
$ weave launch
$ docker $(weave config) run -ti ubuntu
root@2fc245e5bd53:/# nc 172.17.42.1 6784
foo
HTTP/1.1 400 Bad Request
```

We should document the importance here of setting `-icc=false` (we already do that for other reasons), but more importantly we need to find a way of preventing access to weave from container when running with FASTDP. I am guessing a couple of iptable rules that prevent access from the docker bridge to local weave control. And data port, since otherwise containers can inject traffic on unsecured weaves.

awh:
Prevent containers from accessing the weave router control port when fast datapath is in use.

Fixes #1632.

### DIFF
--- a/site/features.md
+++ b/site/features.md
@@ -219,7 +219,7 @@ NB: By default docker permits communication between containers on the
 same host, via their docker-assigned IP addresses. For complete
 isolation between application containers, that feature needs to be
 disabled by
-[setting `--icc=false`](https://docs.docker.com/articles/networking/#between-containers)
+[setting `--icc=false`](https://docs.docker.com/engine/userguide/networking/default_network/container-communication/#communication-between-containers)
 in the docker daemon configuration. Furthermore, containers should be
 prevented from capturing and injecting raw network packets - this can
 be accomplished by starting them with the `--cap-drop net_raw` option.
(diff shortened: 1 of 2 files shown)


rade:
hmm. so there is still (and always has been) a problem when running w/o fastdp, and `-icc=true`.

We should mention *somewhere* in our docs (the security section, perhaps?), that setting `-icc=false` is needed to prevent application containers from interfering with the weave network. Or we could fix #1546, but that's too much of a change for point release.

awh:
> We should mention somewhere in our docs (the security section, perhaps?)

I'll add a doc update commit to this PR.

bboreham:
Note this PR was merged into master then cherry-picked onto 1.2 as 4dd63d8f72c247513bf6e20c04394ced35023add and 3abb657777be0c8bb67b6979fbdfef3a6bbea0a8

PR #1641, opened 2015-11-05
better convergence when peers restart

### LINKED ISSUE #1554: restarting a peer can lead to permanent global non-convergence and connection churn
On a number of occasions we have seen log files that show connections shutting down with `Multiple peers found with same name: ...` and `Connection appears to be with different version of a peer we already know of` with associated significant levels of connection churn.

To date we considered these to be transitory conditions, caused by delays in peers detecting that one of their connections is stale, and thus still believing they are connected to the old incarnation of a restarted peer. Heartbeat timeouts should resolve that in less than ~1 minute.

However, users report that they have seen these errors persist for far longer than 1 minute.

rade:
Previously when a peer restarted, information about the new incarnation (i.e. with a different UID) was not accepted by other peers (and connections would be dropped) unless all knowledge of the
previous incarnation had been purged. This could result in a lot of connection churn and hence connectivity disruption, and, in some pathological cases, very slow convergence and hence acceptance of the new incarnation into the network.

We now no longer drop connections when encountering different incarnations of a peer. There are two situations when that can happen:

1) on connection establishment

we simply proceed

2) on receipt of gossip

to ensure convergence we

a) treat the UID as an additional discriminator when deciding whether we should update our information about a peer with that which was gossiped. Specifically, we update the information we hold when a) the gossiped version is greater, or b) is the same and the UID is greater.

b) include the UID in the information we update

c) move our own version number beyond any we receive for ourselves, if the received UID differs from ours.

With (a) we establishes a total order of peer information across several incarnations of the same peer. i.e. we consider information to be fresher if it has a higher version, or the same version and higher UID. This may seem somewhat counter intutive, since it will generally treat information about new incarnations as older than old incarnations, since incarnations always start life with version 1. But to do better we'd need to establish a total order of incarnations that matches their temporal occurrence. Which requires some sort of durable state.

So instead we have (c). Through that we learn the highest version number of any old incarnation of ourselves that other peers still hold, and then make sure that our version is greater than that. Essentially we continue where the old incarnations left off. It's as if instead of restarting we had simply changed UIDs. And due to (a) and (b) the information about the new incarnation of ourselves, now with a higher version, will supersede that of the old incarnations.

Fixes #1554.

### DIFF
--- a/router/peers.go
+++ b/router/peers.go
@@ -40,14 +40,6 @@ func (upe UnknownPeerError) Error() string {
 	return fmt.Sprint("Reference to unknown peer ", upe.Name)
 }
 
-type NameCollisionError struct {
-	Name PeerName
-}
-
-func (nce NameCollisionError) Error() string {
-	return fmt.Sprint("Multiple peers found with same name: ", nce.Name)
-}
-
 type PeerNameSet map[PeerName]struct{}
 
 type ConnectionSummary struct {
@@ -68,6 +60,9 @@ type PeersPendingNotifications struct {
 
 	// The local short ID needs reassigning due to a collision
 	reassignLocalShortID bool
+
+	// The local peer was modified
+	localPeerModified bool
 }
 
 func NewPeers(ourself *LocalPeer) *Peers {
@@ -100,7 +95,8 @@ func (peers *Peers) OnInvalidateShortIDs(callback func()) {
 }
 
 func (peers *Peers) unlockAndNotify(pending *PeersPendingNotifications) {
-	broadcastLocalPeer := pending.reassignLocalShortID && peers.reassignLocalShortID(pending)
+	broadcastLocalPeer := (pending.reassignLocalShortID && peers.reassignLocalShortID(pending)) ||
+		pending.localPeerModified
 	onGC := peers.onGC
 	onInvalidateShortIDs := peers.onInvalidateShortIDs
 	peers.Unlock()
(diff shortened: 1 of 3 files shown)


rade:
I have run the test from https://github.com/weaveworks/weave/issues/1554#issuecomment-154040588 against this. No connection drops at all, even after resuming all the gossiping. And all peers end up with the correct info about the restarted peer (new UID, etc).

router/peers.go
```diff
@@ -458,36 +454,36 @@ func (peers *Peers) applyUpdate(decodedUpdate []*Peer, decodedConns [][]Connecti
...
+		case peers.ourself.Peer:
+			if newPeer.UID != peer.UID {
+				// The update contains information about an old
+				// incarnation of ourselves. We increase our version
+				// number beyond that which we received, so our
+				// information supersedes the old one when it is
+				// received by other peers.
+				pending.localPeerModified = peers.ourself.setVersionBeyond(newPeer.Version)
+			}
+		default:
```
bboreham: I think we lost something in the commenting here - it wasn't immediately obvious to me what the distinction between the `newPeer` and the `default` case meant.

I think `newPeer` means we have only just added this record, while `default` means there was one already.

bboreham:
Apart from my comment on the comment, LGTM

router/peers.go
rade: I've re-ordered the clauses and stuck an "existing peer" comment on the default case.

PR #1648, opened 2015-11-08
separate the control and data planes

rade:
This is a preparatory step for extracting the control plane as a generic communication substrate into a separate package ('mesh').

The control plane shouldn't have to know what it is being used for, i.e. in our case creating an overlay network. We therefore introduce a more generic notion of an Overlay - something that hooks into the lifecycle of control plane connections and buddies them up with its own logical (or real) OverlayConnections. For weave we then have a specialised NetworkOverlay which represents a data plane, i.e. it knows about (i.e. has methods for) packet forwarding.

There is a similar separation at the Router, i.e. there is a generic control plane Router, and more specialised NetworkRouter - the latter knows about packet capture and forwarding.

### DIFF
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -151,7 +151,7 @@ type WeaveStatus struct {
 	DNS     *nameserver.Status `json:"DNS,omitempty"`
 }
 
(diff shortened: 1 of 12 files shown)


awh:
Looks good aside from the `LocalConnection.Overlay` naming confusion!

PR #1658, opened 2015-11-09
Upgrade CI to Docker 1.9 and work round issues

bboreham:
Pending a resolution of https://github.com/docker/docker/issues/17823 we need to disable the code-coverage collection code.  Once that is fixed we should revert that commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ releases
 .weaver.uptodate
 .weavedns.uptodate
 .weaveexec.uptodate
-weave.tar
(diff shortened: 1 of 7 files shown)


bboreham:
I should mention this also changes the filesystem driver from `aufs` (the default on Ubuntu with required option installed) to `overlayfs`.  This was originally thought to address the "mqueue" errors, but didn't, however `overlayfs` seems preferred from what I read so I'm going to leave it changed.

PR #1662, opened 2015-11-10
DNS upstream servers report

### LINKED ISSUE #1571: Print out info on upstream servers for weaveDNS debugging
When remote troubleshooting it can be useful to know what weave thinks its upstream servers are.

inercia:
Fixes #1571.

### DIFF
--- a/nameserver/status.go
+++ b/nameserver/status.go
@@ -1,10 +1,11 @@
 package nameserver
 
 type Status struct {
-	Domain  string
-	Address string
-	TTL     uint32
-	Entries []EntryStatus
+	Domain   string
+	Upstream []string
+	Address  string
+	TTL      uint32
+	Entries  []EntryStatus
 }
 
 type EntryStatus struct {
@@ -37,6 +38,7 @@ func NewStatus(ns *Nameserver, dnsServer *DNSServer) *Status {
 
 	return &Status{
 		dnsServer.domain,
+		dnsServer.upstream.Servers,
 		dnsServer.address,
 		dnsServer.ttl,
 		entryStatusSlice}
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -25,6 +25,12 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 		}
 		return count
 	},
+	"upstreamServers": func(servers []string) string {
+		if len(servers) == 0 {
+			return "none"
+		}
+		return strings.Join(servers, ", ")
+	},
 	"printConnectionCounts": func(conns []mesh.LocalConnectionStatus) string {
 		counts := make(map[string]int)
 		for _, conn := range conns {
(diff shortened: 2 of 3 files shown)


rade:
quick comment. the new field should come earlier everywhere - it's part of the configuration rather than dynamic state, and the former should (generally) precede the latter.

inercia:
> quick comment. the new field should come earlier everywhere - it's part of the configuration rather than dynamic state, and the former should (generally) precede the latter.

You mean `Status.Upstream`?

prog/weaver/http.go
```diff
@@ -25,6 +27,12 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 		}
 		return count
 	},
+	"upstreamServers": func(servers []string) string {
+		if len(servers) == 0 {
+			return "WARNING: no servers !!"
```
inercia: Any suggestions on how we should warn users when no servers are configured?

rade:
> You mean `Status.Upstream`?

Yes, but also in the template.

inercia:
> > You mean `Status.Upstream`?

> Yes, but also in the template.

Ok. What do you think about this?:

```
       Service: dns
        Domain: weave.local.
      Upstream: 8.8.8.8, 8.8.4.4
           TTL: 1
       Entries: 0
```

rade:
> What do you think about this?

Looks fine.

inercia:
CircleCI fails on `640_proxy_restart_reattaches_test`... Rebuilding until it passes...

awh:
Docs need updating to reflect the new output

awh:
> Any suggestions on how we should warn users when no servers are configured?

It should just say

```
    Upstream: disabled
```

for consistency with the other fields.

rade:
> `Upstream: disabled`

No. To me `disabled` means the user has taken specific action to disable something.

inercia:
>> Upstream: disabled

> No. To me disabled means the user has taken specific action to disable something.

_"Unavailable"_ ? _"Not configured"_ ?

rade:
neither. "unavailable" implies it's something that might fix it self. "Not configured" implies it's something the user has to configure (which generally they don't). 'none' perhaps, but I defer to @awh.

awh:
I would go with `none`.

nameserver/status.go
```diff
@@ -32,12 +33,15 @@ func NewStatus(ns *Nameserver, dnsServer *DNSServer) *Status {
...
 	}
 
 	return &Status{
 		dnsServer.domain,
+		dnsServer.upstream.Servers,
 		dnsServer.address,
 		dnsServer.ttl,
-		entryStatusSlice}
+		entryStatusSlice,
+	}
```
awh: Did gofmt do this?

inercia: I did it. I remember reading about never ending lines with `}` in Go because it makes `diffs` worse...

awh: Sorry to be a pain, but can you undo it? The entire rest of the status code is written in this style, and *if* we're going to change it we should do so uniformly in a separate PR...

inercia: Ok, I'll remove these changes (including the `import`s in `prog/weaver/http.go`)

inercia: Done

PR #1663, opened 2015-11-10
Add FDP miss counter to reports

### LINKED ISSUE #1592: Add fast datapath miss counter to `weave report`

inercia:
Get some stats for `weave report` when using the FDP. This prints the number of _upcalls_ (ie, misses) as well as the number of flows in the datapath.

Fixes #1592

### DIFF
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -218,6 +218,9 @@ func (fastdp fastDatapathBridge) String() string {
 }
 
 func (fastdp fastDatapathBridge) Stats() map[string]int {
+	lock := fastdp.startLock()
+	defer lock.unlock()
+
 	return map[string]int{
 		"FlowMisses": int(fastdp.missCount),
 	}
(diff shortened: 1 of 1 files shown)


bboreham:
Please rebase so the tests run against Docker 1.9

inercia:
> Please rebase so the tests run against Docker 1.9

Rebased, but fails because of #1675 ...

bboreham:
Oh, sorry, missed that.

bboreham:
Please try again.

awh:
I think we can lose the commit that adds the flow counter, because we already enumerate all flows in detail (see #1525).

awh:
I don't think there's any need to introduce the sync atomic - we're already holding the fastdp lock in the miss handler, you can just take that again in the Stats() method to protect the read...

inercia:
> I don't think there's any need to introduce the sync atomic - we're already holding the fastdp lock in the miss handler, you can just take that again in the Stats() method to protect the read, as is done in the equivalent pcap stats method.

@awh I agree on the `Miss` handler part, but I'm not sure it is worth locking the whole _fasdp_ on `Stats` just for reading a counter. Why not just return it...?

awh:
> Why not just return it...?

Reading a 64-bit value is unlikely to be safe on a 32-bit architecture.

PR #1664, opened 2015-11-10
Add plugin support to weave script

awh:
Implements part of weaveworks/docker-plugin#36.

### DIFF
--- a/site/features.md
+++ b/site/features.md
@@ -11,6 +11,7 @@ example](https://github.com/weaveworks/weave#example):
  * [Virtual ethernet switch](#virtual-ethernet-switch)
  * [Fast data path](#fast-data-path)
  * [Seamless Docker integration](#docker)
+ * [Docker network plugin](#plugin)
  * [Address allocation](#addressing)
  * [Naming and discovery](#naming-and-discovery)
  * [Application isolation](#application-isolation)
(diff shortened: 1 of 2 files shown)


awh:
Thinking about this some more - if we're going to do a prerelease build of weave with plugin support, should we base it off `1.2` instead of `master`? Thoughts please @bboreham @rade

bboreham:
As the plugin is targeted for the 1.3 release we should prerelease off master not 1.2

awh:
I didn't mean that we should commit the changes to the `1.2` branch - we could branch from `1.2`, make a prerelease from there and then merge that branch back into `master` afterwards. If we were for example worried about the large amount of relatively untested change that has gone onto `master` recently...

weave
```diff
@@ -1781,8 +1811,7 @@ EOF
         fi
         collect_cidr_args "$@"
         shift $CIDR_ARG_COUNT
-        CONTAINER=$(docker run -e WEAVE_CIDR=none $DNS_ARGS -d "$@")
-        create_bridge
+        CONTAINER=$(docker run -e WEAVE_CIDR=none $DNS_ARGS -d "$@") create_bridge
```
bboreham: Did you mean to merge these two lines?

weave
```diff
@@ -1705,6 +1724,12 @@ EOF
         launch_proxy "$@"
         echo $PROXY_CONTAINER
         ;;
+    launch-plugin)
+        deprecation_warnings "$@"
```
bboreham: Shouldn't need this for a brand-new feature

weave
```diff
@@ -1760,6 +1785,11 @@ EOF
             echo "       Service: proxy"
             echo "       Address: $PROXY_ADDRS"
         fi
+        if [ -z "$SUB_STATUS" ] && check_running $PLUGIN_CONTAINER_NAME 2>/dev/null ; then
+            echo
+            echo "       Service: plugin"
+            echo "       Address: unix:///run/docker/plugins/weave.sock"
```
bboreham: This information isn't very useful to the user, is it?  How about "Driver name: weave" ?

bboreham:
Do you think we should check for minimum Docker 1.9 when using the plugin?

site/features.md
```diff
@@ -88,6 +89,22 @@ Containers started in this way that subsequently restart, either by an
 explicit `docker restart` command or by Docker restart policy, are
 re-attached to the weave network by the weave Docker API proxy.
 
+### <a name="plugin"></a>Docker network plugin
+
+Alternatively, you can use weave as a Docker plugin. First you must
+configure the Docker daemon to use a KV store via the
```
bboreham: Docker don't use the term "KV store"; they do use "KV service" once.  Maybe just "configure the Docker daemon to use a cluster store as described here"

awh:
> Do you think we should check for minimum Docker 1.9 when using the plugin?

This is a bit tricky, because under normal circumstances (e.g. sans `--local`) `launch-plugin` is happening inside the `weaveexec` container which has a fixed client version. Is it that important?

bboreham:
I don't suppose the version check is too important; the plugin will just sit there with no-one talking to it, and the `docker network create` will fail.

Issue #1661, opened 2015-11-09
nil pointer dereference at fastdp.go:677


dpw:
See https://groups.google.com/a/weave.works/d/msg/weave-users/KBF7Ao_IT8Q/Nn1ZeLINCQAJ

In 1.2.0/1.2.1:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x28 pc=0x4de519]

goroutine 62 [running]:
github.com/weaveworks/weave/router.(*fastDatapathForwarder).sendHeartbeat(0xc8212a98c0)
	/home/bryan/golang/src/github.com/weaveworks/weave/router/fastdp.go:677 +0x5f9
github.com/weaveworks/weave/router.(*fastDatapathForwarder).doHeartbeats(0xc8212a98c0)
	/home/bryan/golang/src/github.com/weaveworks/weave/router/fastdp.go:623 +0xd9
created by github.com/weaveworks/weave/router.(*fastDatapathForwarder).Confirm
	/home/bryan/golang/src/github.com/weaveworks/weave/router/fastdp.go:605 +0x428
```

Cause: `fastDatapathForwarder Forward()` can return nil (as of https://github.com/weaveworks/weave/commit/35141a4aeff748ac57dce24e05172534dc40fb83) but the invocation in `sendHeartbeat` wasn't updated to take account of that.

rade:
What circumstances trigger this? Presumably something relatively rare, since otherwise we would have seen it earlier.

rade:
Is there also a problem in `MultiFlowOp.Add`? It gets invoked with the result of `Forward`, and it seems a nil value would trip up a subsequent invocation of `MultiFlowOp.Process`.

rade:
and I wonder whether `router.relay` (or rather its maze of call sites) is also affected. It returns the result of `Forward`. Perhaps it should check for nil and return `DiscardingFlowOp` instead.

dpw:
> Cause: fastDatapathForwarder Forward() can return nil (as of 35141a4) but the invocation in sendHeartbeat wasn't updated to take account of that.

So this isn't quite as silly as it might first appear.  If we are sending a fastdp heartbeat to a peer, we must have fastdp, and they must have fastdp, so how can `HasShortID` be false for either (note that `HasShortID` remains true even while a short id collision is being resolved)?

Answer: if we are attempting a connection to a peer that we heard about from a 1.1 peer.

[...]

PR #1668, opened 2015-11-11
Handle the case where Forward returns nil

dpw:
Forward can return nil if the remote peer has HasShortID false.  That
can happen if we learn about a 1.2 peer from a 1.1  peer and attempt to
connect to it.

This change just fixes the panic.  HasShortID will never be set, so we'll
fall back to sleeve.  But if that occurs, if can be fixed by a restart of
the affected routers.

Addresses #1661.


rade:
needs to be re-PR'ed against 1.2

PR #1670, opened 2015-11-11
Forbid traffic to the Weave port from containers

### LINKED ISSUE #1634: Block container access to TCP 6783 & UDP 6783/6784
See https://github.com/weaveworks/weave/issues/1632#issuecomment-153714146

To prevent application containers from interfering with the weave control and data planes, especially when those are unencrypted.

inercia:
Fixes #1634

### DIFF
--- a/weave
+++ b/weave
@@ -426,6 +426,13 @@ create_bridge() {
             run_iptables -t filter -I FORWARD -i $DOCKER_BRIDGE -o $BRIDGE -j DROP
         fi
 
+        [ -n "$DOCKER_BRIDGE_IP" ] || docker_bridge_ip
+
+        # forbid traffic to the Weave port from other containers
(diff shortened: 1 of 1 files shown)


bboreham:
Please rebase so the tests run against Docker 1.9

inercia:
> Please rebase so the tests run against Docker 1.9

Blocked by #1675

awh:
Apologies, the original issue was a bit muddled - it stated we needed to block UDP 6783/6784 but we also need to block TCP 6783 too (I've updated the subject to reflect this). This PR blocks TCP & UDP 6783 - can we block `$PORT+1` (the fast datapath vxlan port) as well?

inercia:
@awh So, just to be sure we are on the same page, we need to block `6783/tcp`, `6783/udp` and `6784/udp`...

PR #1672, opened 2015-11-12
Pin docker version in vagrant VM

2opremio:
Otherwise it will always install the latest docker version, which will be
undesirable when new major releases are done, until they are well tested.

For instance, docker 1.9 was killing my VM when spawning a kubernetes cluster,
due to a high CPU consumption, probably due to
https://github.com/docker/docker/issues/17720

[Review on Reviewable](https://reviewable.io/reviews/weaveworks/weave/1672)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@ require './vagrant-common.rb'
 vm_ip = "172.16.0.3" # arbitrary private IP
 
 pkgs = %w(
-  docker-engine
+  docker-engine=1.10.2-0~wily
   aufs-tools
   build-essential
   ethtool


bboreham:
I like the idea but we have moved the unit tests to 1.9 so it's unlikely we would merge a pin to 1.8 into master.

bboreham:
Can you check whether your bug is fixed by Docker 1.9.1?

bboreham:
@2opremio ping; is 1.9.1 a suitable version to pin to?

2opremio:
@bboreham I haven't tested it yet, but docker/docker#17720 is not fixed so I wouldn't think so.

awh:
@2opremio the blocking issue you referenced has been resolved; if you update the PR to use 1.10.1 and extend it to pin the version in `test/Vagrantfile` too I'll merge it.

2opremio:
True, I will get it done on Monday.

2opremio:
I updated the PR, sorry it took this a whole week (it is Monday though :) ). PTAL

PR #1679, opened 2015-11-13
[proxy] add --no-detect-tls flag, and proxy always tries to detect TLS.

### LINKED ISSUE #1539: Proxy doesn't detect TLS settings if told where to listen
If you supply a `-H` argument to launch-proxy, it will forget about setting up TLS. This is a pain, for example, if you want to listen on a unix socket (so containers can be given access to docker) as well as a TCP socket.

Suggestion from @paulbellamy: An argument `--detect-tls` which forces it to detect TLS settings and use them, whatever other arguments you supply.

paulbellamy:
Fixes #1539

### DIFF
--- a/site/proxy.md
+++ b/site/proxy.md
@@ -231,11 +231,11 @@ be able to launch a TLS-enabled proxy with:
 
     host1$ weave launch-proxy
 
-You can also manually configure the proxy's TLS. This is accomplished
-by launching the proxy with the same TLS-related command-line flags as
-supplied to the docker daemon. For example, if you have generated your
-certificates and keys into the docker host's `/tls` directory, we can
-launch the proxy with:
+To disable auto-detection of TLS configuration, you can either pass
+the `--no-detect-tls` flag, or manually configure the proxy's TLS with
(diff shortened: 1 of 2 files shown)


site/proxy.md
```diff
@@ -231,11 +231,13 @@ be able to launch a TLS-enabled proxy with:
...
 
-You can also manually configure the proxy's TLS. This is accomplished
-by launching the proxy with the same TLS-related command-line flags as
-supplied to the docker daemon. For example, if you have generated your
-certificates and keys into the docker host's `/tls` directory, we can
-launch the proxy with:
+To disable auto-detection of TLS configuration, you can either pass
+the `--no-detect-tls` flag, or manually configure the proxy's TLS.
+
+To manually configure the proxy's TLS, launch the proxy with the same
```
rade: That repetition is annoying.

paulbellamy: Am open to a better turn-of-phrase...

rade: Just make it one paragraph...

"... or manually configure the proxy's TLS with the same TLS-related command-line flags..."

Issue #1586, opened 2015-10-23
work with dockers on domain sockets other than unix:///var/run/docker


shanmugakarna:
`weave launch` is not detecting docker socket if `DOCKER_HOST` is set to non-default unix socket 

Docker daemon is listening on `unix:///var/run/docker-real.sock` and `$DOCKER_HOST=unix:///var/run/docker-real.sock`

`docker` commands works fine as expected. But `weave lauch` returns
`Cannot connect to the Docker daemon. Is 'docker -d' running on this host?`

**_Note:_**
_This is to achieve something similar here https://github.com/rancher/rancher/issues/2398 to integrate weave into Rancher_

rade:
`weave launch` *is* detecting the `DOCKER_HOST`. So, for example, it is possible to set it to the tcp port of a remote docker. However, it does expect docker to always (also) listen on `unix:///var/run/docker`.

I see from the referred bug that you want to make the weave proxy listen on `unix:///var/run/docker`. That doesn't actually work either atm. See #1579.

paulbellamy:
Most of this is actually just getting the weave script to work when docker is not listening on `unix:///var/run/docker.sock`, which is quite tricky due to the "running in two places" aspect of the weave script.

bboreham:
Fixed by #1740 and #1689

PR #1689, opened 2015-11-16
Support weave script running against non-standard DOCKER_HOST

paulbellamy:
First part of fixing #1586.

Need to check it works with the plugin, and `DOCKER_HOST=tcp://*` as well. Any docker invocations I've missed, that need `DOCKER_HOST` passed in?

### DIFF
--- a/prog/plugin/main.go
+++ b/prog/plugin/main.go
@@ -49,7 +49,7 @@ func main() {
 	Log.Println("Weave plugin", version, "Command line options:", os.Args[1:])
 
 	// API 1.21 is the first version that supports docker network commands
-	dockerClient, err := docker.NewVersionedClient("unix:///var/run/docker.sock", "1.21")
+	dockerClient, err := docker.NewVersionedClientFromEnv("1.21")
 	if err != nil {
 		Log.Fatalf("unable to connect to docker: %s", err)
 	}
--- a/test/650_proxy_env_test.sh
+++ b/test/650_proxy_env_test.sh
@@ -30,7 +30,7 @@ check
 
 # Check we can use weave env/config with unix -Hs specified
 weave_on $HOST1 stop
-weave_on $HOST1 launch-proxy -H unix:///var/run/weave/weave.sock
+run_on $HOST1 "COVERAGE=$COVERAGE weave launch-proxy -H unix:///var/run/weave/weave.sock"
 assert_raises "run_on $HOST1 'eval \$(weave env) ; docker $CMD'"
 assert_raises "run_on $HOST1 'docker \$(weave config) $CMD'"
 
--- a/weave
+++ b/weave
@@ -96,9 +96,21 @@ usage() {
     exit 1
 }
 
+docker_sock() {
+  if echo "$DOCKER_HOST" | grep -q "^unix://" >/dev/null; then
+    echo "${DOCKER_HOST#unix://}"
+  else
+    echo "/var/run/docker.sock"
+  fi
+}
+
+docker_sock_volume_mount() {
+  echo "-v $(docker_sock):$(docker_sock) -e DOCKER_HOST=unix://$(docker_sock)"
+}
+
 exec_remote() {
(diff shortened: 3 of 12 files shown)


paulbellamy:
Update: It doesn't work with the plugin, as the plugin is hard-coded to `/var/run/docker.sock`

paulbellamy:
Does still work with `DOCKER_HOST=tcp://*`.

awh:
Needs a rebase - another `docker` invocation got added to the makefile when the plugin codebase was merged in

awh:
I was going to re-raise weaveworks/docker-plugin#47 in this repo as we've merged the code bases, but actually that doesn't make sense - the commit from that PR can be added to this one.

weave
```diff
@@ -1489,7 +1507,7 @@ launch_router() {
         --dns-effective-listen-address $DOCKER_BRIDGE_IP \
         ${NETHOST_OPT:+$DNS_ROUTER_OPTS} $NO_DNS_OPT \
         --http-addr $HTTP_IP:$HTTP_PORT \
-        --docker-api "unix:///var/run/docker.sock" "$@")
+        "$@")
```
awh: This change, in conjunction with the new default value for `dockerAPI` in `prog/weaver/main.go`, means that if the user does _not_ have `DOCKER_HOST` set in their environment when they launch the router then it will silently fail to subscribe to container events and hence all the IPAM & DNS tombstoning magics will not be invoked...

awh: Is that analysis correct? Very surprised no smoke tests fail...

bboreham: smoke tests run `weave launch` via `weave_on` which sets `DOCKER_HOST`

paulbellamy: Good catch, will fix.

paulbellamy:
Also, needs the plugin changes, brought over here, since it moved.

awh:
Looking good. Can we smoke test this? The vagrantfiles already cause the docker daemon in the test VMs to listen on TCP as well as the standard docker socket; presumably we can extend them to include a non-standard unix socket as well?

paulbellamy:
Sure, but no guarantees it's actually using the non-standard port.

paulbellamy:
This needs a "rebuild-with-ssh" on circle to debug why test 690 is hanging. Waiting for https://github.com/weaveworks/build-tools/issues/10 to be deployed to fix a bug, for "rebuild-with-ssh" to work.

prog/plugin/main.go
```diff
@@ -41,7 +41,11 @@ func main() {
 
 	Log.Println("Weave plugin", version, "Command line options:", os.Args)
 
-	dockerClient, err := docker.NewClient("unix:///var/run/docker.sock")
+	dockerSocket := "unix:///var/run/docker.sock"
+	if val := os.Getenv("DOCKER_HOST"); val != "" {
+		dockerSocket = val
+	}
+	dockerClient, err := docker.NewClient(dockerSocket)
```
bboreham: there is a `docker.NewClientFromEnv()` that will do this for us.

test/650_proxy_env_test.sh
```diff
@@ -31,7 +31,7 @@ check
 # Check we can use weave env/config with unix -Hs specified
 weave_on $HOST1 stop
 weave_on $HOST1 launch-proxy -H unix:///var/run/weave/weave.sock
-assert_raises "run_on $HOST1 'eval \$(weave env) ; docker $CMD'"
-assert_raises "run_on $HOST1 'docker \$(weave config) $CMD'"
```
paulbellamy: But we should be connecting as `vagrant` user which is in the `docker` group, so it should be available, no?

prog/plugin/main.go
paulbellamy: Rebased and fixed this.

test/650_proxy_env_test.sh
bboreham: Since `launch-proxy` was called on the line above with `$DOCKER_HOST` set to `tcp:xxx`, `weave.sock` doesn't get the docker group ownership.

PR #1695, opened 2015-11-17
prevent stalling all broadcasts when single connection is blocked

### LINKED ISSUE #1793: re-ordering of topology gossip can delay convergence
For a new connection, the following happens...

1. if the remote peer is new, `router.SendAllGossipDown(conn)`
2. `router.BroadcastTopologyUpdate`
3. start processing inbound tcp, in particular forwarding of received topology gossip/broadcast

Topology gossip assumes that the ordering of the messages resulting from the above is preserved for any given outbound connection.

So, for example, if A and B are connected, and then B and C connect, B's topology update (which is emitted by B in (2)) must be sent to A before C's topology update (which is emitted by B in (3)). If the order is not preserved then A will learn about C through C's topology update, which contains the connection from C to B but not from B to C, rendering C unreachable from A and causing it to get garbage-collected immediately. At that point we need to wait for periodic gossip to sort things out and achieve convergence.

Unfortunately re-ordering is possible because broadcasts are processed by a `GossipSender` per source peer. So in the example, B's own broadcast and the broadcast from C that it forwards can get re-ordered.

rade:
See commit messages for details.

This also fixes #1793.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -314,7 +314,7 @@ func (alloc *Allocator) Shutdown() {
 		if heir := alloc.pickPeerForTransfer(); heir != mesh.UnknownPeerName {
 			alloc.ring.Transfer(alloc.ourName, heir)
 			alloc.space.Clear()
-			alloc.gossip.GossipBroadcast(alloc.Gossip())
+			alloc.gossip.GossipBroadcast(alloc.encode())
 			time.Sleep(100 * time.Millisecond)
 		}
 		doneChan <- struct{}{}
@@ -414,13 +414,13 @@ func (alloc *Allocator) OnGossipUnicast(sender mesh.PeerName, msg []byte) error 
 }
 
 // OnGossipBroadcast (Sync)
-func (alloc *Allocator) OnGossipBroadcast(sender mesh.PeerName, msg []byte) (mesh.GossipData, error) {
+func (alloc *Allocator) OnGossipBroadcast(sender mesh.PeerName, msg []byte) error {
 	alloc.debugln("OnGossipBroadcast from", sender, ":", len(msg), "bytes")
 	resultChan := make(chan error)
 	alloc.actionChan <- func() {
 		resultChan <- alloc.update(sender, msg)
 	}
-	return alloc.Gossip(), <-resultChan
+	return <-resultChan
 }
 
 type gossipState struct {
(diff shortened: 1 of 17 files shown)


rade:
I have tested this with

    WEAVE_NO_FASTDP=1 NUM_WEAVES=<N> bin/multiweave launch --conn-limit 100`

measuring how long it takes from the end of that until

    WEAVE_CONTAINER_NAME=weave1 weave status

shows all connections between peers as `established`.

For N=40 there is no difference between master and this branch - both take about 90 seconds. For N=50, to my surprise, this branch takes a lot longer than master. The logs show a fair amount of connection churn due to dropped heartbeats. I think this is simply due to higher CPU load on this branch since it propagates broadcasts better.

The very issue we are fixing here does actually reduce CPU load since a single stalled connection will cause all subsequent broadcasts to accumulate locally before being sent when the connection unblocks. Overall this causes fewer broadcasts to be sent, which in turn means fewer broadcasts are received and processed. And some "flapping" connection may well get removed from the topology broadcast stream through the merging. Which in turn means fewer connection establishments get triggered through peer discovery.

rade:
I've spent some more time analysing the performance with multiweave and believe my above analysis is correct. I have filed #1761 and #1762 to look into the root causes.

Sorry this is just one commit and may look at bit daunting. But there are really just four things going on here:

- ditching of GossipData - this accounts for the bulk of changes across the many files and is largely mechanical
- ditching of the per-broadcaster GossipSenders - this affects gossip_channel.go only
- buffering logic in GossipSender - this is all in gossip.go
- shutting down of GossipSender on send failure - this is a small API change on ProtocolSender and some extra channel signalling in GossipSender

Unfortunately they cannot be teased apart into separate commits while retaining workingness.

rade:
I'd like to get #1773 merged first and then run another set of performance tests here.

rade:
A lot of the changes in here were supplanted by #1826, #1855 and #1856. The remaining substantial change here is the removal of `GossipData`. This is a substantial change in the API and behaviour. It makes the gossip API much neater - and more amenable to remoting - but getting it to behave reasonably under load has proved challenging.

I've [pushed the branch to my own repo](https://github.com/rade/weave/tree/gossip-broadcast-stall), in case we ever want to revive this.

PR #1696, opened 2015-11-17
Automatically symlink /var/run/*.sock sockets to /var/run/weave/*.sock

### LINKED ISSUE #1694: Automatically symlink when proxying unix domain sockets under `/var/run`
The proxy currently generates a warning when you attempt to proxy sockets whose immediate parent is `/var/run`, due to the issue with mount namespaces capturing network namespaces. Users are advised to make a symlink of the target socket in a subdirectory and supply that instead; this advice should be automated.

paulbellamy:
Uses approach 1 to fix #1694

Comments on approach vs 2, would be great.

### DIFF
--- a/test/650_proxy_env_test.sh
+++ b/test/650_proxy_env_test.sh
@@ -28,4 +28,10 @@ weave_on $HOST1 stop
 weave_on $HOST1 launch-proxy -H tcp://0.0.0.0:12375 -H unix:///var/run/weave/weave.sock
 check
 
+# Check we can use weave env/config with unix -Hs specified
+weave_on $HOST1 stop
+weave_on $HOST1 launch-proxy -H unix:///var/run/weave/weave.sock
+assert_raises "run_on $HOST1 'eval \$(weave env) ; docker $CMD'"
+assert_raises "run_on $HOST1 'docker \$(weave config) $CMD'"
+
 end_suite
--- a/weave
+++ b/weave
@@ -1266,17 +1266,36 @@ tls_arg() {
     PROXY_ARGS="$PROXY_ARGS $1 /home/weave/tls/$3.pem"
 }
 
+# TODO: Handle relative paths for args
+# TODO: Handle args with spaces
+host_arg() {
+  PROXY_HOST="$1"
+  if [ "$PROXY_HOST" != "${PROXY_HOST#unix://}" ]; then
+    host=$(dirname ${PROXY_HOST#unix://})
+    if [ "$host" = "${host#/}" ]; then
+      echo "When launching the proxy, unix sockets must be specified as an absolute path." >&2
+      exit 1
+    fi
+    if [ "$host" = "/var/run" ]; then
+      echo "WARNING: Proxying sockets whose immediate parent directory is /var/run is unreliable." >&2
+      echo "To work around this, put the unix socket in a subdirectory of /var/run, and symlink it into place." >&2
+    fi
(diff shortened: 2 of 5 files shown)


test/650_proxy_env_test.sh
```diff
@@ -28,4 +28,10 @@ weave_on $HOST1 stop
 weave_on $HOST1 launch-proxy -H tcp://0.0.0.0:12375 -H unix:///var/run/weave/weave.sock
 check
 
+# Check we can use weave env/config with unix -Hs specified
+weave_on $HOST1 stop
+weave_on $HOST1 launch-proxy -H unix:///var/run/weave/weave.sock
+assert_raises 'run_on '$HOST1' "eval $(weave env) ; docker '$CMD'"'
+assert_raises 'run_on '$HOST1' "docker $(weave config) '$CMD'"'
```
awh: These tests are failing with a peculiar error (`program terminated with code 1 instead of -e`); the `-e` is coming from inside `$CMD` - the quotation is incorrect.

paulbellamy: Yeah, am having some issues with the quotation getting stuff to execute in the right place.

weave
```diff
@@ -1552,6 +1571,9 @@ launch_proxy() {
         --entrypoint=/home/weave/weaveproxy \
         $WEAVEPROXY_DOCKER_ARGS $EXEC_IMAGE $COVERAGE_ARGS $PROXY_ARGS)
     wait_for_status $PROXY_CONTAINER_NAME http_call_unix $PROXY_CONTAINER_NAME status.sock
+    for sock in $PROXY_VAR_RUN_SYMLINKS; do
+      ln -sf -T "/var/run/weave/$sock" "/host/var/run/$sock"
```
awh: The intent is to link from `/var/run/weave` to `/var/run` is that correct? If so, the arguments to `ln` are reversed...

awh:
Can you think of any compelling reasons why we would choose option 2? I think I prefer option 1.

weave
paulbellamy: It's to place a symlink at `/var/run/$sock` pointing to `/var/run/weave/$sock`. and `-T` means "Target".

paulbellamy:
Option 2 makes for more consistent handling of unix sockets. It means we're not special-casing sockets in `/var/run`, they're handled the same as anywhere else. But it is ~30-70 more lines of code (because running docker containers from go is more work than from cli)

paulbellamy:
Ok, test is failing because docker unix socket is only accessible to root on gce.

weave
awh: Sorry I'm being an idiot - we're creating these symlinks to listen on, not connect to...

paulbellamy: Correct.

awh:
> Option 2 makes for more consistent handling of unix sockets. It means we're not special-casing sockets in /var/run, they're handled the same as anywhere else. But it is ~30-70 more lines of code (because running docker containers from go is more work than from cli)

Another idea - how about having the weave script symlink all `unix:///` args (anywhere in the filesystem - no special casing) to `/var/run/weave/weave.sock` and then have the proxy listen to that single socket?

awh:
If someone does
```
$ weave launch-proxy -H unix:///var/run/test.sock
```
would they find the path returned by this
```
$ weave env
export DOCKER_HOST=unix:///var/run/weave/test.sock ORIG_DOCKER_HOST=
```
 surprising?

paulbellamy:
Their level of surprise probably depends on what they expected it to return.

PR #1701, opened 2015-11-18
Raise minimum Docker version to 1.6

### LINKED ISSUE #1683: weave's `docker pull` will stop working on December 7
and hence `weave setup` and `weave launch[-*]`. Because we embed a Docker 1.3 client...

From https://blog.docker.com/2015/10/docker-hub-deprecation-1-5/

> * On November 19, 2015 Docker clients version 1.5 and earlier will not be able to push images to Docker Hub. They will still be able to pull images. And of course the repositories are fully accessible via newer versions of the Docker client.
* On December 7, 2015, pulls via clients 1.5 and earlier will be disabled. Only version 1.6 or later will be supported.

I believe Bryan has seen some deprecation warnings already, so we should upgrade soon so that users aren't bombarded with these confusing messages (which will refer to a docker version they themselves aren't running).

bboreham:
Fixes #1683

### DIFF
--- a/weave
+++ b/weave
@@ -12,9 +12,10 @@ fi
 IMAGE_VERSION=${WEAVE_VERSION:-$IMAGE_VERSION}
 
 # - Docker versions prior to 1.6 cannot pull images from DockerHub
-# - The weavexec image uses an embedded Docker 1.6 client.
-# - Docker will complain if you run a client against a server older
-#   than that client.
+# - The weavexec image embeds a Docker 1.6.2 client. Docker will give
+#   a "client is newer than server error" if the daemon has an older
+#   API version, which could be confusing if the user's Docker client
+#   is correctly matched with that older version.
 #
 # We therefore check that the user's Docker *client* is >= 1.6
 MIN_DOCKER_VERSION=1.6.0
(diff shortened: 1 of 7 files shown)


weave
```diff
@@ -11,19 +11,13 @@ else
...
 IMAGE_VERSION=${WEAVE_VERSION:-$IMAGE_VERSION}
 
-# - Docker versions prior to 1.2.0 cannot parse `--dns-search=.`
-# - Docker versions prior to 1.3.0 fail to include the domain when
-#   setting the hostname, so lookups for non-fq names fail.
-# - Docker version 1.3.0 contains the severe bug
-#   https://github.com/docker/docker/issues/8632
-# - The weavexec image embeds a Docker 1.3.1 client. Misleading
-#   version mismatch errors are produced if the user is running with a
-#   client *and* server with a lower version.
```
rade: that 'misleading version mismatch errors" comment is still relevant.

bboreham: It seems to me to be covered by the next line - the user will see an error.  Whether it's misleading or not is unimportant.

rade: The reason we are checking that the client is greater than our embedded version is because of the misleading error that otherwise arises when they run a client and server of the same but older version.

bboreham: I changed the wording; ptal.

bboreham:
Wrong branch; replaced by #1706

PR #1702, opened 2015-11-19
[WIP] Show some IPAM stats with "weave status ipam"

### LINKED ISSUE #1639: lack of diagnostics when IPAM is stuck but does have consensus
IPAM can get stuck even though it does have consensus, because a) the entire range is owned by a node which is currently dead, or b) the entire range is allocated.

This is hard to diagnose at present. `weave status` reports on the consensus state only. `weave report` has the info, but a) it's not something we expect users to have to look at, b) the info is not particularly digestable, and c) it's not showing nicknames, which, if an entry belongs to a dead node, is crucial for identifying what node is the problem.

See #1638 for an example.

Suggestions:

1. annotate the ipam entries list in the `weave report` with nicknames. Alternatively, include the nickname map as a separate item (though this is noisy and less helpful in the typical case)
2. introduce a `weave status ipam` that shows *summarised* information about IPAM:
    - the allocation range
    - the total number of IPs in the range, the total number of free IPs, the total number of allocated IPs, and some free/allocated percentage figures - yes, these are not necessarily accurate
    - the same figures, for every peer mentioned in the ring
    - NB: I don't think we should list ring entries

inercia:
Fixes #1639

### DIFF
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -60,14 +60,14 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 			}
 			percentageRanges := float32(ips) * 100.0 / float32(status.RangeNumIPs)
 
-			displayName := name + " (" + nickName + ")"
-			fmt.Fprintf(&buffer, "%30s: %8d IPs (%04.1f%% of total) %s\n",
+			displayName := name + " (" + nickName + "):"
+			fmt.Fprintf(&buffer, "%-32s %8d IPs (%04.1f%% of total) %s\n",
 				displayName, ips, percentageRanges, reachableStr)
 		}
 
 		// print the local info first
 		if ourStats := peerStats[router.Name]; ourStats != nil {
-			printOwned(router.Name, "local", true, ourStats.ips)
+			printOwned(router.Name, ourStats.nickname, true, ourStats.ips)
 		}
 
 		// and then the rest
@@ -158,14 +158,16 @@ var statusTemplate = defTemplate("status", `\
        Service: ipam
 {{if .IPAM.Entries}}\
 {{if allIPAMOwnersUnreachable .IPAM}}\
-        Halted: all IPs owned by unreachable peers
+        Status: all IP ranges owned by unreachable peers - use 'rmpeer' if they are dead
+{{else if len .IPAM.PendingAllocates}}\
+        Status: waiting for IP range grant from peers
 {{else}}\
-     Consensus: achieved
+        Status: ready
 {{end}}\
 {{else if .IPAM.Paxos}}\
-     Consensus: waiting (quorum: {{.IPAM.Paxos.Quorum}}, known: {{.IPAM.Paxos.KnownNodes}})
+        Status: awaiting consensus (quorum: {{.IPAM.Paxos.Quorum}}, known: {{.IPAM.Paxos.KnownNodes}})
 {{else}}\
-     Consensus: deferred
+        Status: idle
 {{end}}\
          Range: {{.IPAM.Range}}
  DefaultSubnet: {{.IPAM.DefaultSubnet}}
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -60,8 +60,8 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 			}
 			percentageRanges := float32(ips) * 100.0 / float32(status.RangeNumIPs)
 
-			displayName := name + " (" + nickName + "):"
-			fmt.Fprintf(&buffer, "%-32s %8d IPs (%04.1f%% of total) %s\n",
+			displayName := name + "(" + nickName + ")"
+			fmt.Fprintf(&buffer, "%-37v %8d IPs (%04.1f%% of total) %s\n",
 				displayName, ips, percentageRanges, reachableStr)
 		}
 
@@ -198,7 +198,7 @@ var peersTemplate = defTemplate("peers", `\
 {{.Name}}({{.NickName}})
 {{range .Connections}}\
    {{if .Outbound}}->{{else}}<-{{end}} {{printf "%-21v" .Address}} \
-{{$nameNickName := printf "%v(%v)" .Name .NickName}}{{printf "%-32v" $nameNickName}} \
+{{$nameNickName := printf "%v(%v)" .Name .NickName}}{{printf "%-37v" $nameNickName}} \
 {{if .Established}}established{{else}}pending{{end}}
 {{end}}\
 {{end}}\
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -46,7 +46,7 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 		getNickName := func(peer mesh.PeerName) string {
 			name, found := status.Nicknames[peer]
 			if !found {
-				return fmt.Sprintf("%s", peer)
+				return peer.String()
 			}
 			return name
 		}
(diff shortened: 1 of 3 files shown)


bboreham:
`weave status` gives some IPAM information (same as prior to this PR):

          Service: ipam
        Consensus: achieved
            Range: 10.32.0.0-10.47.255.255
    DefaultSubnet: 10.32.0.0/12

and this PR then repeats some of that information:

````
$ weave status ipam
      Universe: 10.32.0.0-10.47.255.255 (1048576 IPs)
 DefaultSubnet: 10.32.0.0/12
    Containers:
                da001e7156cddd081b1e888ca150d228ec90301d83b65d47ead4a823b891f07b: [10.32.0.1]
    Ownerships:
             (local):  1048576 IPs (100.0% of universe, in   1 ranges) - used:        1 IPs 
````

This is different to `weave status dns` which doesn't repeat anything from `weave status`.

And the name `Universe` is `Range` in `weave status`

**Comment by bboreham** [comment]
Note that two peers can have the same nickname; this happens when you `weave reset` a peer and then launch it again.  So the output needs to distinguish those two instances.

**Comment by rade** [comment]
Please bear in mind the issue (#1639) that this is meant to address. What is the most useful information we can show ?

**Comment by bboreham** [comment]
Currently I have it like this:

````
         Range: 10.32.0.0-10.47.255.255 (1048576 IPs)
 DefaultSubnet: 10.32.0.0/12
    Containers:
                da001e7156cd: [10.32.0.1]
    Ownerships:
fe:e2:23:09:da:02(local):   262146 IPs (25.0% of total) - used: 2 
ce:98:83:03:49:44(host1):   262143 IPs (25.0% of total) - used: 1 
22:d8:76:19:b1:59(host1):   524287 IPs (50.0% of total) - used: 1 - unreachable!
````

So it's possible to see we have two instances of `host1`, one of which is "dead".

**Comment by rade** [comment]
ok. need to think about what to put in the `weave status` summary, e.g. as per https://github.com/weaveworks/weave/issues/1639#issuecomment-154019069.

I am not keen on showing lists mixed with other stuff. Perhaps we should *just* show what you currently have under "Ownerships"?

(I don't think we need "Containers", plus I think it's confusing, since it's local-only)

**Comment by bboreham** [comment]
OK, now we have this warning in the main `weave status`:

````
       Service: ipam
     Consensus: achieved - all IPs owned by unreachable peers!
         Range: 10.32.0.0-10.47.255.255
 DefaultSubnet: 10.32.0.0/12
````

And `weave status ipam` is like this:

````
ce:98:83:03:49:44(host1):   524289 IPs (50.0% of total) - unreachable!
22:d8:76:19:b1:59(host1):   524287 IPs (50.0% of total) - unreachable!
````

**Comment by rade** [comment]
hmm. We may want to revise our terminology here. Sounds a lot like "success - it's completely broken!"

**Comment by bboreham** [comment]
Latest version:

````
       Service: ipam
        Halted: all IPs owned by unreachable peers
         Range: 10.32.0.0-10.47.255.255
 DefaultSubnet: 10.32.0.0/12
````

Maybe "Consensus: halted - ..." would be better for people parsing the output?  It's not really the consensus that is halted.

**Comment by rade** [comment]
perhaps go for the more generic "Status", with values of "idle", "awaiting consensus", "all IPs owned by unreachable peers - waiting for them to come online or be removed with 'rmpeer'", "ready". or something like that.

**Comment by bboreham** [comment]
Note I have changed the implementation recently:
* All data exported from `ipam/status.go` is now strings and numbers, in keeping with the original design.
* `IPAM.Entries` now has the nickname and known-peer status on each entry, so people parsing `weave report` can do the same computation `weave status` does.  This is more verbose.
* I removed `IPAM.Owned` since it is not required for this issue.  It might have been useful for debugging.

**Comment by bboreham** [comment]
The code currently prints the nickname of the local node as `(local)`, which is quite nice but inconsistent with `weave status peers` which just prints its nickname (e.g. derived from the host).

**Comment by rade** [comment]
ditch the `local`.

**Comment by bboreham** [comment]
OK, now we have:

        Status: all IPs owned by unreachable peers - use 'rmpeer' if they are dead
        Status: ready
        Status: awaiting consensus (quorum: X, known: Y)
        Status: idle

("idle" is the old "deferred")

and "local" is gone.

**Comment by rade** [comment]
should we have a state indication about pending requests? Which can happen in all except the idle state, but I think is mainly interesting in the 'ready' state. Perhaps we could split 'ready' into 'ready' and 'waiting for space grant from other peers' (or something like that).

**Comment by bboreham** [comment]
There can be pending allocations (any address in a subnet) and pending claims (one specific address).       The former generally indicates someone is blocked; the latter may or may not (it's ok to continue using an address that is "owned" by a peer that is actually dead).

So I will print a message for pending allocations only.

**Comment by bboreham** [comment]
OK, now we have:

        Status: ready - waiting for space grant from peers

It isn't possible to see exactly who we are waiting for, even using `weave report`, as the PendingAllocates don't give the subnet.  I'll add that.

**Comment by rade** [comment]
> `ready - waiting for space grant from peers`

I would ditch the "ready" bit.

**Comment by rade** [comment]
also "space" is an odd term. "waiting for IP address grant from peers"? or "IP address range". or "IP range". or "IP".

**Comment by bboreham** [comment]
Now `Status: waiting for IP range grant from peers`.

If no further comments I will squash the commits.

**Comment by rade** [comment]
please show an example of the output of the ipam section of `weave status`, and the output of `weave status ipam`.

**Comment by bboreham** [comment]
````
$ weave status ipam
[...]
       Service: ipam
        Status: waiting for IP range grant from peers
         Range: 10.32.0.0-10.47.255.255
 DefaultSubnet: 10.32.0.0/12
[...]

$ weave status ipam
8e:fc:31:88:19:e4 (vagrant-ubuntu-vivid-64):   524415 IPs (50.0% of total) 
f2:b4:4b:07:b3:73 (host1):         524161 IPs (50.0% of total) - unreachable!
````

**Comment by awh** [comment]
Some observations about the sample output:
* Everywhere else we show the name/nickname, there is no separating space inbetween (e.g. we use `f2:b4:4b:07:b3:73(host1)` not `f2:b4:4b:07:b3:73 (host1)`)
* We don't use separating `:`'s elsewhere - that can go IMO
* Unless it was a cut'n'paste snafu, the padding of the name/nickname field has not worked correctly - the IP counts should be aligned vertically

File: prog/weaver/http.go [hunk 1]
```diff
@@ -25,6 +27,60 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
...
+		for peer, stats := range peerStats {
+			if peer != router.Name {
+				printOwned(peer, stats.nickname, stats.reachable, stats.ips)
+			}
+		}
+
+		return buffer.String()
+	},
+	"allIPAMOwnersUnreachable": func(status ipam.Status) bool {
+		for _, entry := range status.Entries {
```
awh: Do we not need to exclude ourselves (e.g. the `router.Name` entry) from this loop?

bboreham: If we have some IPs then all owners are not unreachable.

awh:
Needs buttery base

bboreham:
Rebased

awh:
@bboreham I suspect you missed my [other comments](https://github.com/weaveworks/weave/pull/1702#issuecomment-159893309) about the `weave status ipam` output...

bboreham:
> the padding of the name/nickname field has not worked correctly - the IP counts should be aligned vertically

the alignment is consistent with previous code; it doesn't work if your hostname is over 9 chars long

bboreham:
Updated sample output:

````
$ weave status ipam
4a:95:cd:2b:cb:35(bbvlx1)               524288 IPs (50.0% of total) 
66:9a:14:47:6b:2d(host1)                524288 IPs (50.0% of total) 
````

and updated `weave status peers` to match alignment up to 14-char hostname:

````
$ weave status peers
4a:95:cd:2b:cb:35(bbvlx1)
   -> 192.168.48.11:6783    66:9a:14:47:6b:2d(host1)              established
66:9a:14:47:6b:2d(host1)
   <- 192.168.48.1:40739    4a:95:cd:2b:cb:35(bbvlx1)             established
````

PR #1708, opened 2015-11-20
Connect to gce boxes as vagrant so we can access docker unix socket

paulbellamy:


bboreham:
what is the background to this change?

paulbellamy:
Came up while trying to write a test for https://github.com/weaveworks/weave/issues/1694

bboreham:
So is it needed?  The tests for #1696 seem to have passed.

paulbellamy:
#1696 included this commit.

bboreham:
Ah, I hadn't spotted that #1696 was reverted

PR #1722, opened 2015-11-26
Add fast datapath high level overview

awh:
Document fast datapath.

### DIFF
--- a/docs/fastdp.md
+++ b/docs/fastdp.md
@@ -1,0 +1,168 @@
+# Overlay Method Selection
+
+Weave automatically selects the best working overlay implementation
+when connecting to a new peer. It does this by initiating both
+forwarders in parallel, eventually settling on the most preferred
+established forwarder at which point the other forwarder is shut down,
+assuming it had not failed already. In the event that the remaining
+forwarder fails at a later time (e.g. because of a heartbeat timeout)
+the control plane TCP connection will recycle allowing the selection
+algorithm to run again. By default fast datapath is preferred over
+Sleeve.
+
+# Local Bridging
+
+Unlike the weave bridge netdev used by Sleeve, the OVS datapath has no
+inherent bridging capability. Consequently fastdp implements an
+ethernet bridge in addition to the vxlan overlay, maintaining its own
+forwarding database by learning MAC addresses and dispatching
+broadcast, unicast & multicast traffic accordingly.
+
+# Short Peer IDs
+
+The router relies on having access to the source and destination peer
+when deciding how to forward packets. When the Sleeve overlay is in
+use this information is conveyed directly within the encapsulation by
+including the names of the peers in question, a solution not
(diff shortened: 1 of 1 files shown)


awh:
I shall address David's comments from #1510 in this PR.

awh:
@dpw thanks for comments!
> This sounds definitional, but it seems like a "for instance" to me ... I'd urge not to try to enumerate the keys and actions in created flows according to the context. It's fiddly

I agree that it's fiddly to maintain, but it's also essential for those who are trying to understand _how it actually works now_ - in fact it's the main thing that I wanted to record after discovering it myself, so I'd prefer to leave it as is.

> Unknown for the bridge implementation. Not unknown in terms of MacCache.

Is that an important distinction to draw in this doc?

dpw:
> > This sounds definitional, but it seems like a "for instance" to me ... I'd urge not to try to enumerate the keys and actions in created flows according to the context. It's fiddly
> 
> I agree that it's fiddly to maintain, but it's also essential for those who are trying to understand how it actually works now - in fact it's the main thing that I wanted to record after discovering it myself, so I'd prefer to leave it as is.

I disagree that it is essential.  Rather the opposite: it is significant that the full keys are left implicit, and generated as a side effect of handling packets, is one of the most important design elements.  That design freed me from having to think about what the full rules looked like, and I'd encourage everyone else to understand the code in the same terms (particularly when it comes to extending it in the future). 

Giving an example of what the generated rules look like seems ok, as context.  Beyond that encourages misunderstanding.

>   >  Unknown for the bridge implementation. Not unknown in terms of MacCache.
> 
> Is that an important distinction to draw in this doc?

Would it take more than a few additional words to avoid the risk of confusion?

awh:
> I disagree that it is essential ... I'd encourage everyone else to understand the code in the same terms (particularly when it comes to extending it in the future).

I think the disagreement comes from the fact that I am trying to capture the _way_ in which fastdp configures the datapath, not just _how_ it does it. The document is as much of interest to people who want to treat the router as a black box but understand what effects it has on the network configuration of the machine its running on, as much as it is to a future maintainer of the code itself IMO... does it make sense in that context? Would you prefer to split that out into a separate document?

> Would it take more than a few additional words to avoid the risk of confusion?

No, assuming the reader already knows the difference between the bridge MAC table and the router cache...

rade:
Why is this in 1.4/master? Surely it should be on `latest_release_doc_updates`.

awh:
> Why is this in 1.4/master? Surely it should be on latest_release_doc_updates.

It's developer documentation in `weave/docs`, and is not linked from site.

rade:
> not linked from site

oh, I thought this was going to be an extension of "How it works", for folks interested in that kind of stuff.

awh:
> oh, I thought this was going to be an extension of "How it works", for folks interested in that kind of stuff.

The plan was for it to evolve into that, but even trying to get these minimal internal docs merged is proving an uphill battle :smile:

PR #1728, opened 2015-11-26
Docker ipam plugin

### LINKED ISSUE #1676: Docker IPAM plugin
Docker has a pluggable address allocation framework - It may be possible to provide our IPAM implementation as a provider. Related to weaveworks/docker-plugin#43

bboreham:
Allocate addresses from weave's pool.

May need to be more sensitive to when docker is asking for the "global pool" and "local pool".

Fixes #1676


rade:
weaveapi -> api? Makes it more awkward for imports, I guess. Still...

bboreham:
Note that we can't `docker network create` using the embedded docker 1.6 client; the `network` command only came in with docker 1.9.

bboreham:
I have renamed `weaveapi` to `api`.

bboreham:
Note this PR is based in part on the branch for #1730; either can be merged first but the other will need to be rebased.

Also #1738 builds upon this branch; it should be merged after this one.

bboreham:
Rebased.

awh:
I have some generals concerns about `configureWeaveClient` (which I am assuming is there to account for the fact that the router may not be up when the plugin is launched):

1. It duplicates what `haveWeaveClient` in the watcher is doing to some extent
2. I don't think it's safe to cache the weave API client in either the watcher or the IPAM driver, because the docker IP of the weave router container may change (at least until #1746 is resolved).

Is there a compelling reason not to resolve the router container IP every time we need to do an API operation? We could encapsulate that behind the API itself, addressing the duplication and the caching issue at the same time.

bboreham:
`configureWeaveClient` is used differently, in that being unable to contact weave is a hard error for IPAM but something we let pass in the DNS watcher.

That said, I agree there could be more commonality, and moving the caching up seems reasonable.

bboreham:
I was thinking the docker IP of the weave router wouldn't change without a `weave reset`, which means the plugin is restarted.  But this may not be true in light of the intermediary bridge?

awh:
```
vagrant@vagrant-ubuntu-vivid-64:~/weave$ WEAVE_NO_FASTDP=1 ./weave launch
vagrant@vagrant-ubuntu-vivid-64:~/weave$ docker inspect -f '{{.NetworkSettings.IPAddress}}' weave
172.17.0.2
vagrant@vagrant-ubuntu-vivid-64:~/weave$ ./weave stop
vagrant@vagrant-ubuntu-vivid-64:~/weave$ docker run -tid ubuntu
74cb1561ab32657284b50c30a23daf03ea32bab0a4ee3f104b46f49413e09752
vagrant@vagrant-ubuntu-vivid-64:~/weave$ WEAVE_NO_FASTDP=1 ./weave launch
vagrant@vagrant-ubuntu-vivid-64:~/weave$ docker inspect -f '{{.NetworkSettings.IPAddress}}' weave
172.17.0.3
```

PR #1730, opened 2015-11-27
Fix DNS integration in Weave docker plugin

### LINKED ISSUE #1725: Fix weaveDNS integration
Registration of containers with weaveDNS is broken due to the changes in the container JSON to accommodate the possibility of multiple networks. Beware that go-dockerclient does not support the new format at the time of writing.

Replaces weaveworks/docker-plugin#54

bboreham:
Includes a refactoring to create an `api` package which we use to call the `RegisterWithDNS()` function.

I took out the `ContainerDied` handler since WeaveDNS is also monitoring that event; I don't think the previous implementation would have worked on `docker rm -f` anyway, as it relied on being able to `inspect` the container.

Fixes #1725

Docs need to be updated to explain how to use this feature (`-h foo.weave.local` etc)


awh:
> Docs need to be updated to explain how to use this feature (-h foo.weave.local etc)

Is that happening in this PR @bboreham ?

PR #1732, opened 2015-11-29
update Peer.HasShortID as part of topology merge

### LINKED ISSUE #1731: [fastdp] inefficiency and packet loss in mixed version networks due to lack of HasShortID merging
Topology gossip merge pays no attention to / does not update `Peer.HasShortID`. This can cause problems in networks containing a mixture of pre-1.2 and 1.2+ weave versions:

- When we first hear about a peer indirectly via an old-version weave that doesn't support ShortID, we will create a peer record with ShortID=0 and HasShortID=false. If we subsequently get an update about that peer from a new-version weave, containing a ShortID and HasShortID=true (because the peer is in fact a new-version peer), we update the ShortID of the peer record, but not the HasShortID. This in turn means we will continue to use the sleeve overlay when routing to that peer, even though we could use fastdp.

- Say we have heard about a peer including its ShortID and HasShortID=true. If we subsequently receive an update about that peer from on old-version weave, we set ShortID=0 (the only value that can be received from an old-version peer), yet leave HasShortID=true. i.e. we've just wiped out the known ShortID and now claim that 0 is the genuine ShortID of that peer, which is a lie. As a result we will route packets for that peer either to the wrong peer or into the void, both of which result in packet loss. Furthermore, when we gossip our incorrect peer record to other peers we contaminate them with the same false information, thus breaking the routing from them to the peer in the same way.

@dpw please confirm that my assessment of the consequences is correct.

rade:
This was missing.

There are two parts to this change:

1. updating the Peers.byName mapping when HasShortID changed (in addition to the existing update on ShortID change). We need to be careful about the sequencing here - both ShortID and HasShortID must be set between the calls to deleteByShortID and addByShortID, since the former is meant to delete the existing entry (hence use the old values) and the latter add the new entry (hence use the new values).

2. update a peer when the received info has the same version and UID but the current record has HasShortID=false, and the received info has HasShortID=true. This caters for the case where we receive an update about a weave 1.2+ peer first via a pre-1.2 peer (which does not transmit ShortIDs, and always sets HasShortID=false) and subsequently via a 1.2+ peer. The former update will force us to communicate with that peer over the 'sleeve' overlay. We want to accept the latter update, even though the version is the same, in order to update the ShortID and HasShortID, so we can switch to fastdp.

Fixes #1731.

### DIFF
--- a/mesh/peers.go
+++ b/mesh/peers.go
@@ -476,16 +476,20 @@ func (peers *Peers) applyUpdate(decodedUpdate []*Peer, decodedConns [][]Connecti
 			newUpdate[name] = void
 		default: // existing peer
 			if newPeer.Version < peer.Version ||
-				(newPeer.Version == peer.Version && newPeer.UID <= peer.UID) {
+				(newPeer.Version == peer.Version &&
+					(newPeer.UID < peer.UID ||
+						(newPeer.UID == peer.UID &&
+							(!newPeer.HasShortID || peer.HasShortID)))) {
 				continue
 			}
 			peer.Version = newPeer.Version
 			peer.UID = newPeer.UID
 			peer.connections = makeConnsMap(peer, connSummaries, peers.byName)
 
-			if newPeer.ShortID != peer.ShortID {
+			if newPeer.ShortID != peer.ShortID || newPeer.HasShortID != peer.HasShortID {
 				peers.deleteByShortID(peer, pending)
 				peer.ShortID = newPeer.ShortID
+				peer.HasShortID = newPeer.HasShortID
 				peers.addByShortID(peer, pending)
 			}
 			newUpdate[name] = void


rade:
I haven't done any testing of this. We do have coverage of the new/changed code, but presumably only because we've added to conjunctions which are satisfied by earlier clauses.

rade:
> I haven't done any testing of this.

Have done some now...

1. built the current 1.3 head and deployed that to my vagrant VMs
2. downloaded weave 1.1.2 onto my desktop. NB: do *not* build this; it won't work when compiled with go1.5

Then...
```
host1$ weave launch
host2$ weave launch
deskt$ weave launch $HOST1 $HOST2
deskt$ weave status connections
-> 192.168.48.12:6783    established ee:53:20:87:88:d2(host2)
-> 192.168.48.11:6783    established ae:d6:8c:ad:be:df(host1)
host1$ weave status connections
<- 192.168.48.1:45937    established sleeve ca:a3:0f:5b:eb:b5(xps)
<- 192.168.48.12:38852   established sleeve ee:53:20:87:88:d2(host2)
host2$ weave status connections
-> 192.168.48.11:6783    established sleeve ae:d6:8c:ad:be:df(host1)
<- 192.168.48.1:53376    established sleeve ca:a3:0f:5b:eb:b5(xps)
```
So host1 and host2 discover each other via deskt, but only establish a sleeve overlay connection.

Now with this branch...
```
deskt$ weave status connections
-> 192.168.48.11:6783    established 32:fa:db:63:78:97(host1)
-> 192.168.48.12:6783    established e6:fc:ae:d4:c3:11(host2)
host1$ weave status connections
<- 192.168.48.1:56893    established sleeve 56:a9:e4:95:db:2f(xps)
<- 192.168.48.12:38864   established fastdp e6:fc:ae:d4:c3:11(host2)
host2$ weave status connections
<- 192.168.48.1:48767    established sleeve 56:a9:e4:95:db:2f(xps)
-> 192.168.48.11:6783    established fastdp 32:fa:db:63:78:97(host1)
```
where host1 and host2 managed to establish a fastdp overlay connection.

So this test confirms the 1st problem and fix identified in #1731. I cannot think of an easy way of testing the 2nd.

PR #1736, opened 2015-11-30
remove UnknownPeerError for quicker convergence and efficiency

rade:
When encountering an unknown peer in received gossip, rather than ignoring the entire gossip we create a placeholder Peer. This has version 0 and uid 0, which means it will be replaced by any complete peer information subsequently received via gossip.

Not ignoring the gossip results in quicker convergence.

We also no longer need to broadcast about the remote peer when adding a connection, since we only did that in order to reduce the occurrence of `UnknownPeerError`s. Not needing to do this is is more efficient, since the broadcast was strictly redundant.

### DIFF
--- a/mesh/local_peer.go
+++ b/mesh/local_peer.go
@@ -157,7 +157,7 @@ func (peer *LocalPeer) handleAddConnection(conn Connection) error {
 	}
 
 	peer.router.Routes.Recalculate()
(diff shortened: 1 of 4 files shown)


rade:
There is one user-visible consequence of this change: occasionally, and usually briefly, `weave status` et al will show peers without a nickname.

PR #1738, opened 2015-11-30
"mesh mode" plugin

### LINKED ISSUE #1727: Local scope plugin
Replaces weaveworks/docker-plugin#43; see there for description and commentary.

bboreham:
...so users don't need to set up a kv store

FIxes #1727

### DIFF
--- a/prog/plugin/main.go
+++ b/prog/plugin/main.go
@@ -20,26 +20,22 @@ import (
 
 var version = "(unreleased version)"
 
-const (
-	meshNetworkName = "weave"
-)
-
 func main() {
 	var (
-		justVersion   bool
-		address       string
-		nameserver    string
-		meshAddress   string
-		logLevel      string
-		createDefault bool
+		justVersion     bool
+		address         string
+		nameserver      string
+		meshAddress     string
+		logLevel        string
+		meshNetworkName string
 	)
 
 	flag.BoolVar(&justVersion, "version", false, "print version and exit")
 	flag.StringVar(&logLevel, "log-level", "info", "logging level (debug, info, warning, error)")
 	flag.StringVar(&address, "socket", "/run/docker/plugins/weave.sock", "socket on which to listen")
 	flag.StringVar(&nameserver, "nameserver", "", "nameserver to provide to containers")
-	flag.StringVar(&meshAddress, "meshsocket", "", "socket on which to listen for mesh mode")
-	flag.BoolVar(&createDefault, "create-default-network", false, "create default network")
+	flag.StringVar(&meshAddress, "meshsocket", "/run/docker/plugins/weavemesh.sock", "socket on which to listen in mesh mode")
+	flag.StringVar(&meshNetworkName, "mesh-network-name", "weave", "network name to create in mesh mode")
 
 	flag.Parse()
 
@@ -48,9 +44,6 @@ func main() {
 		os.Exit(0)
 	}
 
-	if createDefault && meshAddress == "" {
-		meshAddress = "/run/docker/plugins/weavemesh.sock"
-	}
 	SetLogLevel(logLevel)
 
 	Log.Println("Weave plugin", version, "Command line options:", os.Args[1:])
@@ -78,8 +71,8 @@ func main() {
 		defer meshListener.Close()
 	}
 
-	if createDefault {
-		createNetwork(dockerClient, meshAddress)
+	if meshNetworkName != "" {
+		createNetwork(dockerClient, meshNetworkName, meshAddress)
 	}
 
 	sigChan := make(chan os.Signal, 1)
@@ -133,16 +126,16 @@ func listenAndServe(dockerClient *docker.Client, address, nameserver string, end
 	return listener, nil
 }
 
-func createNetwork(dockerClient *docker.Client, address string) {
-	if _, err := dockerClient.Client.NetworkInfo(meshNetworkName); err == nil {
-		Log.Printf("Docker network '%s' already exists", meshNetworkName)
+func createNetwork(dockerClient *docker.Client, networkName, address string) {
+	if _, err := dockerClient.Client.NetworkInfo(networkName); err == nil {
+		Log.Printf("Docker network '%s' already exists", networkName)
 	} else if _, ok := err.(*go_docker.NoSuchNetwork); ok {
 		driverName := strings.TrimSuffix(address, ".sock")
 		if i := strings.LastIndex(driverName, "/"); i >= 0 {
 			driverName = driverName[i+1:]
 		}
 		options := go_docker.CreateNetworkOptions{
-			Name:           meshNetworkName,
+			Name:           networkName,
 			CheckDuplicate: true,
 			Driver:         driverName,
 			IPAM:           go_docker.IPAMOptions{Driver: driverName},
--- a/prog/plugin/main.go
+++ b/prog/plugin/main.go
@@ -6,8 +6,11 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
+	"github.com/docker/libnetwork/ipamapi"
+	go_docker "github.com/fsouza/go-dockerclient"
 	. "github.com/weaveworks/weave/common"
 	"github.com/weaveworks/weave/common/docker"
 	ipamplugin "github.com/weaveworks/weave/plugin/ipam"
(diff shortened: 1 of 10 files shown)


bboreham:
I have rebased this branch to include the plugin DNS change, because otherwise that was breaking the smoke-tests.

bboreham:
It appears that some sequence of events with this plugin can render Docker permanently inoperable, something like:

```
weave launch
docker run --net=weave alpine /bin/sh
weave reset
servicectl restart docker
```

Then we see this kind of thing in `syslog`:

````
Dec  8 16:04:54 host1 systemd[1]: Starting Docker Application Container Engine...
Dec  8 16:04:54 host1 docker[1481]: time="2015-12-08T16:04:54.139044663Z" level=info msg="API listen on /var/run/docker.sock"
Dec  8 16:04:54 host1 docker[1481]: time="2015-12-08T16:04:54.139303415Z" level=info msg="API listen on [::]:2375"
Dec  8 16:04:54 host1 docker[1481]: time="2015-12-08T16:04:54.140789621Z" level=info msg="Firewalld running: false"
Dec  8 16:04:54 host1 docker[1481]: time="2015-12-08T16:04:54.176508426Z" level=warning msg="Unable to locate plugin: weavelocal, retrying in 1s"
...
Dec  8 16:06:09 host1 docker[1481]: time="2015-12-08T16:06:09.191201043Z" level=warning msg="Unable to locate plugin: weavelocal, retrying in 1s"
Dec  8 16:06:10 host1 docker[1481]: time="2015-12-08T16:06:10.191407692Z" level=warning msg="Unable to locate plugin: weavelocal, retrying in 2s"
Dec  8 16:06:12 host1 docker[1481]: time="2015-12-08T16:06:12.191611653Z" level=warning msg="Unable to locate plugin: weavelocal, retrying in 4s"
Dec  8 16:06:16 host1 docker[1481]: time="2015-12-08T16:06:16.191822460Z" level=warning msg="Unable to locate plugin: weavelocal, retrying in 8s"
Dec  8 16:06:24 host1 docker[1481]: time="2015-12-08T16:06:24.192042550Z" level=warning msg="Failed deleting endpoint 4875a70d685a01f15e3a02f0f5722a6d4f65894a47cc2bb35eb7361e8b4eb61d: failed to get endpoint from store during Delete: could not find endpoint 4875a70d685a01f15e3a02f0f5722a6d4f65894a47cc2bb35eb7361e8b4eb61d: datastore not found for scope \n"
Dec  8 16:06:24 host1 systemd[1]: docker.service start operation timed out. Terminating.
Dec  8 16:06:24 host1 systemd[1]: docker.service: main process exited, code=exited, status=2/INVALIDARGUMENT
````

Looking at the timings, you need around 3 endpoints to get past the 30-second timeout in `systemd` to get into this state.

**Comment by awh** [comment]
I was expecting
```
$ docker network create --driver=weavemesh foo
$ docker run --net=foo ...
```
to yield a container attached to the same weave network as obtained from `docker run --net=weave`, but that's not the case. Is that expected? Is it confusing? Should `docker network create --driver=weavemesh` fail with an error?

**Comment by bboreham** [comment]
> I was expecting [...] to yield a container attached to the same weave network as obtained from `docker run --net=weave`, but that's not the case

I imagine what you got was the same L2 network, but with an IP address in a different subnet.  This is a consequence of using a different IPAM driver.

> Is that expected? 

Yes.

> Is it confusing? 

Maybe.  I cannot think of any practical use for this feature.

> Should docker network create --driver=weavemesh fail with an error?

We could examine the address range passed in, and fail if it doesn't match the Weave IPAM range.

**Comment by awh** [comment]
> I imagine what you got was the same L2 network, but with an IP address in a different subnet. This is a consequence of using a different IPAM driver.

That seems likely - I did get an `ethwe0` interface that looked like it had a docker allocated address.

> I cannot think of any practical use for this feature.

Does this not give us a way to satisfy #1742? You could do `docker network create --driver weavemesh --subnet ... --ip-range ...` on each host, specifying a different IP range on each?

**Comment by bboreham** [comment]
> I did get an `ethwe0` interface

That would be on the `docker_gwbridge` network, which is used to give containers a route to the outside world (and to WeaveDNS).

**Comment by awh** [comment]
There's a separate issue for the multicast route - would it not be better to pull that commit out into a separate PR?

**Comment by awh** [comment]
@rade points out that `dns-args` isn't documented in `weave help`

**Comment by rade** [comment]
What happened to the "specify a different network name for the universal weave network." part of https://github.com/weaveworks/docker-plugin/issues/43#issuecomment-156019318? I guess we can defer that until somebody makes a compelling case for it.

**Comment by bboreham** [comment]
I don't understand the "different name" question. We usurped "weave" as a network name, used in documentation for the last release, for the one we create by default. If you call `docker network create` you can pick any name you like.

**Comment by rade** [comment]
> We usurped "weave" as a network name, used in documentation for the last release, for the one we create by default.

It's *that* name which my comment is referring to.

**Comment by bboreham** [comment]
OIC. As it stands you do it in two steps: `weave launch-plugin` then `docker network create ... different-name`. 

We should document more of this.

**Comment by rade** [comment]
So `weave launch-plugin` doesn't create the `weave` network? I find that surprising since I am expecting `weave launch` to be, in essence `weave launch-router && weave launch-proxy && weave launch-plugin`.

**Comment by bboreham** [comment]
Re https://github.com/weaveworks/weave/pull/1738#commitcomment-14876076, note `grep -P` is not supported by Busybox grep; indeed is described in the GNU grep manpage as "highly experimental"

**Comment by bboreham** [comment]
> I am expecting `weave launch` to be, in essence `weave launch-router && weave launch-proxy && weave launch-plugin`.

I was striving to have `weave launch-plugin` do the same as it did in Weave Net 1.3, i.e. start a global driver and nothing else.  Having it create a local-scope network called "weave" will break anything that follows our instructions for the previous version.

This might be ok; probably the people that have done that can adapt.

**Comment by rade** [comment]
> probably the people that have done that can adapt.

exactly. Let's not worry about backward compatibility for plug-in users just yet.

File: prog/plugin/main.go [hunk 1]
```diff
@@ -19,16 +22,20 @@ var version = "(unreleased version)"
...
+		meshAddress     string
+		logLevel        string
+		meshNetworkName string
 	)
 
 	flag.BoolVar(&justVersion, "version", false, "print version and exit")
 	flag.StringVar(&logLevel, "log-level", "info", "logging level (debug, info, warning, error)")
 	flag.StringVar(&address, "socket", "/run/docker/plugins/weave.sock", "socket on which to listen")
 	flag.StringVar(&nameserver, "nameserver", "", "nameserver to provide to containers")
+	flag.StringVar(&meshAddress, "meshsocket", "/run/docker/plugins/weavemesh.sock", "socket on which to listen for mesh mode")
```
rade: "*in* mesh mode"

prog/plugin/main.go
```diff
@@ -39,51 +46,102 @@ func main() {
 
 	SetLogLevel(logLevel)
 
-	Log.Println("Weave plugin", version, "Command line options:", os.Args)
+	Log.Println("Weave plugin", version, "Command line options:", os.Args[1:])
 
-	dockerClient, err := docker.NewClient("unix:///var/run/docker.sock")
+	dockerClient, err := docker.NewVersionedClient("unix:///var/run/docker.sock", "1.21")
```
rade: this version pinning deserves a comment.

site/plugin.md
```diff
@@ -0,0 +1,92 @@
...
+## Only one network
+
+Although Docker allows users to create multiple networks, weave only
+supports the creation of one network in "mesh" mode. Even with a
+cluster store, all containers are on the same virtual network, but
+Docker will allocate their addresses on different subnets so they
+cannot talk to each other directly.
+
+As a result, for the moment you can have only one network at a time.
+
```
rade: These two sections seem to duplicate (and somewhat elaborate) what's in the driver sections. I suggest merging the former into the latter.

Also, bonus points if you can come up with a better heading than "Advanced Topics".

prog/plugin/main.go
bboreham: In the code or in the PR?  I see it as fixing a simple oversight.  There are no choices for specific number at present; this is the only released version that will run the plugin.

test/config.sh
```diff
@@ -217,8 +224,14 @@ assert_dns_ptr_record() {
 start_suite() {
     for host in $HOSTS; do
         [ -z "$DEBUG" ] || echo "Cleaning up on $host: removing all containers and resetting weave"
+        PLUGIN_ID=$(docker_on $host ps -aq --filter=name=weaveplugin)
+        if [ -n "$PLUGIN_ID" ] ; then
+            rm_containers $host $(docker_on $host ps -aq 2>/dev/null | grep -v $PLUGIN_ID)
+        else
+            rm_containers $host $(docker_on $host ps -aq 2>/dev/null)
+        fi
```
rade: the duplication is annoying

prog/plugin/main.go
rade: > In the code or in the PR?

code.

awh:
The final line of the '`weave` driver' section ("As a result, for the moment you can have only one network at a time." should be removed, given that we've just described how that driver can support multiple networks?

bboreham:
Fixed that 'one network' line.

awh:
Ready to merge - squash as desired.

PR #1740, opened 2015-11-30
[proxy] Symlink sockets to /var/run/weave/weave.sock from inside the proxy

### LINKED ISSUE #1694: Automatically symlink when proxying unix domain sockets under `/var/run`

paulbellamy:
Needs an in-depth review, please.

We copy docker.sock owner to symlinks. Permissions are not copied,
because the permissions will inherit from the target.

Fixes #1694, using option 2 (having the proxy create the symlinks). If the weave script creates the symlinks, `weave env` would break (unless we passed the symlinks in, and that's awkward).

There is a slight change in behaviour of `weave env`. New order of preference is: tcp sockets, custom unix sockets, then `/var/run/weave/weave.sock`.

Most of the awkwardness/code in this is just to run a docker container. :disappointed:

Happy to refactor, any suggestions on ways to clean up the code would be great.

### DIFF
--- a/prog/weaveexec/symlink
+++ b/prog/weaveexec/symlink
@@ -4,7 +4,7 @@
 
 set -e
 
-[ -z "$DEBUG" ] || set -x
+[ -z "$WEAVE_DEBUG" ] || set -x
 
 target="$1"
 user="$(stat -c '%u:%g' "$target")"
--- a/prog/weaveexec/symlink
+++ b/prog/weaveexec/symlink
@@ -1,0 +1,18 @@
+#!/bin/sh
+# This lets us force-symlink a bunch of stuff all at once, so we can minimize
+# running containers.
+
+set -e
+
+[ -z "$DEBUG" ] || set -x
+
+target="$1"
+user="$(stat -c '%u:%g' "$target")"
+mode="$(stat -c '%a' "$target")"
+shift 1
+
+for link in "$@" ; do
+  ln -sfT "$target" "$link"
+done
+chown -h "$user" "$@"
+chmod "$mode" "$@"
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -575,8 +575,8 @@ func (proxy *Proxy) symlink(unixAddrs []string) (err error) {
 		"PATH=/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
 	}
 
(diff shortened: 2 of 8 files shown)


proxy/proxy.go
```diff
@@ -186,7 +192,12 @@ func (proxy *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 func (proxy *Proxy) Listen() []net.Listener {
 	listeners := []net.Listener{}
 	proxy.normalisedAddrs = []string{}
+	unixAddrs := []string{}
 	for _, addr := range proxy.ListenAddrs {
+		if strings.HasPrefix(addr, "unix://") {
```
awh: `proxy.listen` interprets any `protoAndAddr` string that begins with a `/` as a unix domain socket - should we not trap that here too?

prog/weaveexec/symlink
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+# This lets us force-symlink a bunch of stuff all at once, so we can minimize
+# running containers.
+
+set -e
+
+[ -z "$DEBUG" ] || set -x
```
awh: Is this deliberately not `WEAVE_DEBUG`?

proxy/proxy.go
```diff
@@ -527,3 +552,74 @@ func (proxy *Proxy) updateContainerNetworkSettings(container jsonObject) error {
...
+		},
+		HostConfig: &docker.HostConfig{Binds: binds},
+	})
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		err2 := proxy.client.RemoveContainer(docker.RemoveContainerOptions{ID: container.ID})
+		if err == nil {
```
awh: I'm not sure this will have the effect you intend, as you are creating shadows of `err` with `:=` later on

paulbellamy: it *does* work, but is a bit subtle. Will try to clean it up a bit.

bboreham:
Does this fix #1579 ?

paulbellamy:
I believe so, yes.

PR #1755, opened 2015-12-03
Option to disable multicast route

### LINKED ISSUE #1741: Add option to stop weave adding a multicast route
For users that are happy with the default routing of multicast packets, weave adding a route to `224.0.0.0/4` is unhelpful.  We could have yet another environment variable to turn off this behaviour.

awh:
Fixes #1741.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ prog/weaveproxy/weaveproxy
 prog/sigproxy/sigproxy
 prog/weavewait/weavewait
 prog/weavewait/weavewait_noop
+prog/weavewait/weavewait_nomcast
 prog/netcheck/netcheck
 prog/docker_tls_args/docker_tls_args
 prog/plugin/plugin
@@ -56,6 +57,7 @@ prog/weaveexec/weavehosts
 prog/weaveexec/weaveproxy
 prog/weaveexec/weavewait
 prog/weaveexec/weavewait_noop
+prog/weaveexec/weavewait_nomcast
 prog/weaveexec/sigproxy
 prog/weaveexec/netcheck
 prog/weaveexec/docker_tls_args
--- a/Makefile
+++ b/Makefile
@@ -14,13 +14,14 @@ WEAVEPROXY_EXE=prog/weaveproxy/weaveproxy
 SIGPROXY_EXE=prog/sigproxy/sigproxy
 WEAVEWAIT_EXE=prog/weavewait/weavewait
 WEAVEWAIT_NOOP_EXE=prog/weavewait/weavewait_noop
+WEAVEWAIT_NOMCAST_EXE=prog/weavewait/weavewait_nomcast
 NETCHECK_EXE=prog/netcheck/netcheck
 DOCKERTLSARGS_EXE=prog/docker_tls_args/docker_tls_args
 DOCKERPLUGIN_EXE=prog/plugin/plugin
 RUNNER_EXE=tools/runner/runner
 TEST_TLS_EXE=test/tls/tls
 
-EXES=$(WEAVER_EXE) $(SIGPROXY_EXE) $(WEAVEPROXY_EXE) $(WEAVEWAIT_EXE) $(WEAVEWAIT_NOOP_EXE) $(NETCHECK_EXE) $(DOCKERTLSARGS_EXE) $(DOCKERPLUGIN_EXE) $(TEST_TLS_EXE)
+EXES=$(WEAVER_EXE) $(SIGPROXY_EXE) $(WEAVEPROXY_EXE) $(WEAVEWAIT_EXE) $(WEAVEWAIT_NOOP_EXE) $(WEAVEWAIT_NOMCAST_EXE) $(NETCHECK_EXE) $(DOCKERTLSARGS_EXE) $(DOCKERPLUGIN_EXE) $(TEST_TLS_EXE)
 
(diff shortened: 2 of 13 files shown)


rade:
I am not convinced you'd really want to set this on a per container basis in the proxy. All other aspects of container configuration performed by the proxy are controlled by global options, with WEAVE_CIDR the only exception.

I don't see how this is conceptually any different from `--no-rewrite-hosts` and `--without-dns`, both of which disable aspects of weave-specific container munging. And both of which have equivalent `weave run` options; not control via env vars.

awh:
> I am not convinced you'd really want to set this on a per container basis

I disagree. For example, you might want to run one set of Elasticsearch containers that discover each other via multicast over the weave network in conjunction with another set that multicast video traffic over the docker bridge.

> I don't see how this is conceptually any different from --no-rewrite-hosts

They're completely different - the host file rewriting is there to correct a behaviour most people consider to be broken; being able to turn it off is an escape hatch in case there was some unforeseen problem with the mechanism.

> and --without-dns

You're much more likely to want to control the routing of subsets of your hosts (containers in this case) than you are to vary the DNS server.

rade:
> You're much more likely to want to control the routing of subsets of your hosts (containers in this case) than you are to vary the DNS server.

We have had exactly one user half-ask for the former.

A few other data points:

- Env vars are a very obscure way of configuring things. Options are generally much better. They show up in 'usage', for starters.

- `weave run` is configured by options, not env vars. If we really want configurability per-container then we should follow the WEAVE_CIDR example - options/params to `weave run` and an env var when starting containers via the proxy.

Overall I am concerned about the cognitive burden we are inflicting on all our users here, for a feature that approximately nobody needs.

awh:
> `weave run` is configured by options, not env vars

`WEAVE_MTU`? `WEAVE_DOCKER_ARGS`? `WEAVE_PASSWORD`?

> Overall I am concerned about the cognitive burden we are inflicting on all our users here, for a feature that approximately nobody needs.

This is hyperbole. The cognitive burden of not supplying an option you don't know about is zero.

awh:
My mistake, you said `weave run` specifically, to which those env vars do not apply.

rade:
> `WEAVE_MTU`

undocumented.

> `WEAVE_DOCKER_ARGS`

undocumented. and only affects the router, not container configuration.

>  `WEAVE_PASSWORD`

only affects the router, not container configuration.

>  The cognitive burden of not supplying an option you don't know about is zero.

The cognitive burden of reading about an option you don't care about is non-zero. And it is considerably greater than zero when that option works differently to any other option.

awh:
Did you have any thoughts on this @bboreham ?

Lawouach:
> We have had exactly one user half-ask for the former.

I'm not sure if that meant me but if so, I apologise for not formulating the request well enough @rade :)

I don't have a personal need for multicast route control on a per-container basis. However, I would still appreciate if weave didn't force me into its own way of thinking either. So I would indeed expect that multicast routing is a feature I can globally disable. Unless, it is critical to weave to function properly. If not, why so much magic?

bboreham:
I think, since it inconvenienced one person enough to speak up, that means it will inconvenience more, so we should have an option to turn it off.

I am swayed by the argument that our general pattern is "options/params to weave run and an env var when starting containers via the proxy.".

On the other side, each option doubles the size of the space we should test, and we are already somewhat lacking in coverage.

rade:
> I am swayed by the argument that our general pattern is "options/params to weave run and an env var when starting containers via the proxy.".

...or options/params for the proxy, for a "global" switch. As our user says..."I would indeed expect that multicast routing is a feature I can *globally* disable." 

> On the other side, each option doubles the size of the space we should test, and we are already somewhat lacking in coverage.

The same is true of env vars, right?

bboreham:
Yes, it was "on the other side" of the argument for having any option at all.

Lawouach:
I have a feeling you guys are having a discussion which is larger than this issue. You are discussing the way to configure weave in general, mayb not the right place for this?

Anyhow, as I said, I prefer that, if weave makes a default decision for me, it provides a mechanism to revert/prevent this decision if I need to, in particular for multicasting. Unless, having this route is a mandatory feature for weave to function properly (if so, this ought to be documented).

> As our user says..."I would indeed expect that multicast routing is a feature I can globally disable."

I'm fine with a global option indeed. However, I don't represent all users out there, neither do I claim to make decision for your design. If the option is on a container basis, it will suit me too. Whatever works for you will be fine by me. I just assumed a global option is simpler.

If you guys prefer not to add more clutter to your options set, that's fine. It's less pratical for me but I'll manage something out of band.

awh:
Reworked to use `--no-multicast-route` instead of an environment variable:
* `weave run --no-multicast-route ...`
* `weave launch-proxy --no-multicast-route ...`

Lawouach:
Hey @awh, I was testing this and I ran into a problem. Was there a reason you didn't pass that parameter to `weave launch` too? This is what I've been using in my systemd unit as documented here: http://docs.weave.works/weave/latest_release/systemd.html

Anyhow, passing --no-multicast-route to `weave launch` makes it crash in a loop.

rade:
> Was there a reason you didn't pass that parameter to `weave launch` too?

It's a parameter to the proxy and to `weave run`, just like `--no-rewrite-hosts` and `--without-dns`.

PR #1758, opened 2015-12-07
Optional encryption

### LINKED ISSUE #82: Option to encrypt some links but not others
E.g. if you want to encrypt traffic leaving your site, but not pay the overhead for intra-datacenter traffic.

awh:
This PR adds support for optional encryption via the `--trusted-subnets` argument to `weave launch`. Whilst the control plane is always secured when a password is specified, peers will forego authenticated encryption on the overlay (data plane) connection between them only if both consider the other to be located on a trusted subnet.

I recommend that the commits are reviewed individually in sequence; most commit messages contain further explanatory text.

Fixes #82, #1788.

### DIFF
--- a/mesh/connection.go
+++ b/mesh/connection.go
@@ -40,17 +40,19 @@ type RemoteConnection struct {
 type LocalConnection struct {
 	sync.RWMutex
 	RemoteConnection
-	TCPConn      *net.TCPConn
-	version      byte
-	tcpSender    TCPSender
-	SessionKey   *[32]byte
-	heartbeatTCP *time.Ticker
-	Router       *Router
-	uid          uint64
-	actionChan   chan<- ConnectionAction
-	errorChan    chan<- error
-	finished     <-chan struct{} // closed to signal that actorLoop has finished
-	OverlayConn  OverlayConnection
+	TCPConn         *net.TCPConn
+	TrustRemote     bool // is remote on a trusted subnet?
+	TrustedByRemote bool // does remote trust us?
+	version         byte
+	tcpSender       TCPSender
+	SessionKey      *[32]byte
+	heartbeatTCP    *time.Ticker
+	Router          *Router
+	uid             uint64
+	actionChan      chan<- ConnectionAction
+	errorChan       chan<- error
+	finished        <-chan struct{} // closed to signal that actorLoop has finished
+	OverlayConn     OverlayConnection
 }
 
 type ConnectionAction func() error
@@ -94,6 +96,7 @@ func StartLocalConnection(connRemote *RemoteConnection, tcpConn *net.TCPConn, ro
 		RemoteConnection: *connRemote, // NB, we're taking a copy of connRemote here.
 		Router:           router,
 		TCPConn:          tcpConn,
+		TrustRemote:      false,
 		uid:              randUint64(),
 		actionChan:       actionChan,
 		errorChan:        errorChan,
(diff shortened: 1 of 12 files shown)


prog/weaver/http.go
```diff
@@ -27,7 +27,7 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 		}
 		return count
 	},
-	"upstreamServers": func(servers []string) string {
+	"printList": func(servers []string) string {
```
rade: `servers` should be renamed.

site/troubleshooting.md
```diff
@@ -162,8 +166,8 @@ The columns are as follows:
       heartbeat
     * `established` - TCP connection and corresponding UDP path are up
  * Info - the failure reason for failed and retrying connections, or
-   the data transport method, remote peer name and nickname for
-   pending and established connections
+   the data transport method, security level, remote peer name and
+   nickname for pending and established connections
```
rade: The order doesn't match the output. Also, what are the possible values for the "security level"?

rade: perhaps leave out the "plaintext" and just have "sleeve(encrypted)".

rade:
I highly recommend "?w=1" for viewing the diff.

site/troubleshooting.md
awh: Ack on the ordering - good spot - will update. Some things about the field itself:

* `plaintext` and `encrypted` were chosen as the two values because they are the same length
* I'd like the output to contain a positive indication that things are going in the plain, and not rely on e.g. someone noticing the absence of an '(encrypted) affix
* It's deliberately separate from the overlay type so we don't have to tinker with the status output when fastdp supports encryption

awh:
> I highly recommend "?w=1" for viewing the diff.

This is why I submitted the whitespace change in a separate commit and recommended reviewing commits individually...

site/troubleshooting.md
rade: > `plaintext` and `encrypted` were chosen as the two values because they are the same length

We don't make any reference "plaintext" for unencrypted weaves in the docs. And it doesn't sound right anyway, since we are not transmitting "text".

> I'd like the output to contain a positive indication that things are going in the plain

I appreciate that. But for those deployments running w/o a password - no doubt the majority - it is just noise.

> It's deliberately separate from the overlay type so we don't have to tinker with the status output when fastdp supports encryption

There is no tinkering required in what I proposed - the rule is "append `(encrypted)` when the connection is encrypted".

awh: > We don't make any reference "plaintext" for unencrypted weaves in the docs. And it doesn't sound right anyway, since we are not transmitting "text".

It has a well defined meaning - I don't think anyone thinks that cryptography only works on ASCII text.

> But for those deployments running w/o a password - no doubt the majority - it is just noise.

How about printing it only if a password has been supplied?

awh: Actually the phrase 'cleartext' appears to be the most appropriate.

bboreham:
Why did we want to "restore leading whitespace"?

awh:
> Why did we want to "restore leading whitespace"?

`weave status` output was modelled on `systemctl status`, which has a single space margin down the left hand side; the length of the new `TrustedSubnets` field required everything to be moved to the right to maintain it.

awh:
> Why do we run this on the remote host? Do we expect the resolution of the address to be different there?

The weave routers in GCE connect via their private IP addresses, so we need to resolve them on the VMs themselves to get the correct trusted subnets.

> If so, maybe a comment explaining why would be useful to the next maintainer.

Ah go on then :smile:

awh:
I believe I have addressed all comments except the one about trailing curly braces, which I think merits discussion outside of this PR. Will autosquash when you're ready to merge.

bboreham:
LGTM; please squash commits.

awh:
Don't merge - I've found an issue with the overlay switch change.

awh:
> I've found an issue with the overlay switch change.

There was actually no need to stop any forwarders at all - now we just record the overlay name in the subforwarder structure (no actual need to do this, but it seemed neater) and continue; PTAL. Will squash again when ready.

bboreham:
LGTM

mesh/connection.go
```diff
@@ -40,17 +40,19 @@ type RemoteConnection struct {
...
+	SessionKey      *[32]byte
+	heartbeatTCP    *time.Ticker
+	Router          *Router
+	uid             uint64
+	actionChan      chan<- ConnectionAction
+	errorChan       chan<- error
+	finished        <-chan struct{} // closed to signal that actorLoop has finished
+	OverlayConn     OverlayConnection
+	TrustRemote     bool // is remote on a trusted subnet?
+	TrustedByRemote bool // does remote trust us?
```
rade: I would move these earlier, just after `TCPConn`.

mesh/connection.go
```diff
@@ -97,7 +99,8 @@ func StartLocalConnection(connRemote *RemoteConnection, tcpConn *net.TCPConn, ro
 		uid:              randUint64(),
 		actionChan:       actionChan,
 		errorChan:        errorChan,
-		finished:         finished}
+		finished:         finished,
+		TrustRemote:      router.Trusts(connRemote)}
```
rade: ditto.

awh:
I have added #1788 to the 'fixes' list for this PR.

PR #1759, opened 2015-12-08
Move test/tls program into Makefile so it builds inside the build container

bboreham:
Building outside the container was breaking the CircleCI build, because CircleCI was variously running Go 1.4 and 1.5.1 which didn't match the build container.

This also saves ~3 seconds on every run of `test/setup.sh`


bboreham:
Replaced by #1760 - should go on 1.3 branch

PR #1767, opened 2015-12-08
better emulation of connections in gossip tests

### LINKED ISSUE #1765: sporadic TestGossipSurrogate failure
Seen in CI, e.g.

https://circleci.com/gh/weaveworks/weave/3929
https://circleci.com/gh/weaveworks/weave/3894

Reproduced locally as well, if run enough times.

The likely cause is the insufficiently accurate emulation of connections with `mockConnection`, brought to the fore by the reduced gossiping post #1736.

rade:
Real `Connection`s do not process gossip until after `LocalPeer.AddConnection`. `mockChannelConnection` wasn't behaving in that way, which resulted in some sporadic test failures since it would cause some gossip to get processed too early, which in turn would trigger gc of some of the just-received peers, which we would then find missing in the test assertions.

To fix this we introduce a latch on `mockChannelConnection` which closes when the remote end is ready to receive gossip. Effectively this emulates the entering of `LocalConnection.receiveTCP` after the call to `AddConnection` completes.

Unfortunately an open latch will cause `sendPendingGossip` to block. We get around that by abandoning checks in "half-connected" connection states.

Fixes #1765.

### DIFF
--- a/mesh/gossip_test.go
+++ b/mesh/gossip_test.go
@@ -11,7 +11,7 @@ import (
 // TODO test gossip unicast; atm we only test topology gossip and
 // surrogates, neither of which employ unicast.
 
-type mockChannelConnection struct {
+type mockGossipConnection struct {
 	RemoteConnection
 	router *Router
 	dest   *Router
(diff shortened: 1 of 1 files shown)


bboreham:
You've removed all testing of one-sided connections?

Issue #1757, opened 2015-12-07
Lots of copies of weavewait


bboreham:
I have 434 files called `w` under `/var/lib/docker/volumes`, evidently created as part of Weave mounting `weavewait` as `/w/w` in every container run via the proxy.

(a) we should find out why they aren't being cleaned up, and (b) I think this was supposed to be a bind-mount not a copy.

bboreham:
This seems to be because (a) the `weaveexec` image exposes two volumes; (b) when we `docker rm` containers run from this image we don't pass the `-v` flag.

Will need to see what happens if you try to `docker rm -v` the proxy while another container has `/w/w` mounted.

rade:
> Will need to see what happens if you try to `docker rm -v` the proxy

What does the proxy container have to do with it? After all, there is only one of them. Or is the problem occuring on repeated `weave launch; weave stop`?

The main "user" of `weaveexec` surely is the weave script.

bboreham:
I get 2 new `w` files for each `weave launch-proxy`. 

The proxy is the main user of these files; it passes them in to containers that will be attached to the weave network. So if those containers outlive the proxy, then, somehow, the `w` files they are using mustn't disappear. 

The weave script is run with `--rm`, so its `w` files are deleted by Docker. This is wasteful but not terminal.

bboreham:
Note https://github.com/weaveworks/weave/pull/1792#issuecomment-164826803 says these files are currently required for some patterns of use.

PR #1768, opened 2015-12-09
Reduce problem of 'w' files

bboreham:
As noted in #1757, we left some copies of `weavewait`, copied as `w`, lying around.

This PR avoids copies of this file except where needed (by the proxy),
and adds the `-v` flag to `docker rm` to clean up those copies.

Note: this PR is on the 1.3 branch.


awh:
Oops - in the meantime, #1755 inadvertently 'increased the problem of 'w' files' by adding another volume, `/w-nomcast`!

bboreham:
I will open another PR against master

PR #1773, opened 2015-12-10
lazy route calculation

### LINKED ISSUE #1761: route calculation takes O(n^2) time
when running something like `WEAVE_NO_FASTDP=1 NUM_WEAVES=40 bin/multiweave launch --conn-limit 100` and taking a profile snapshot, 2/3rd of the time is spent in route calculation.

routes_profile

Looking at the code, I am pretty sure it is O(n_peers^2). Hopefully there's a better way.

rade:
See commit message for explanation.

Fixes #1761.

### DIFF
--- a/mesh/routes.go
+++ b/mesh/routes.go
@@ -19,6 +19,7 @@ type Routes struct {
 	broadcastAll broadcastRoutes // [1]
 	recalculate  chan<- *struct{}
 	wait         chan<- chan struct{}
+	action       chan<- func()
 	// [1] based on *all* connections, not just established &
 	// symmetric ones
 }
@@ -26,6 +27,7 @@ type Routes struct {
 func NewRoutes(ourself *LocalPeer, peers *Peers) *Routes {
 	recalculate := make(chan *struct{}, 1)
 	wait := make(chan chan struct{})
+	action := make(chan func())
 	routes := &Routes{
 		ourself:      ourself,
 		peers:        peers,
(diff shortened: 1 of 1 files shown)


rade:
I've done some performance testing with
```
WEAVE_NO_FASTDP=1 NUM_WEAVES=n bin/multiweave launch --connlimit 100
```
measuring the time from "connecting" until a) CPU usage drops well below 100% and b) `WEAVE_CONTAINER_NAME=weave1 watch -n 5 weave status` reports all connections in the topology as established. The latter is not a good indicator since it fluctuates wildly between runs; it's basically down to luck with periodic gossip timing and random neighbour selection.

```
|    | master    | branch    |
|----+-----------+-----------|
| 30 | 28  / 86  | 28  / 88  |
| 40 | 82  / 171 | 39  / 93  |
| 45 | 122 / 128 | 58  / 116 |
| 47 | 153 / 267 | 70  / 132 |
| 50 | 226 / 228 | 87  / 190 |
| 53 | 365 / 366 | 112 / 137 |
| 55 | 784 / 785 | 136 / 139 |
```

Doing some curve fitting on the first set of figures, shows that master is actually cubic, and this branch is quadratic.

I've also taken a couple of profiles showing the difference between master
weave-master
and the branch
weave-branch

As is evident from that, we are spending considerably less time in route calculation on the branch.

bboreham:
What analysis have you done to come to the "small price to pay" conclusion re flushing FDP routes?

Would I be right in thinking that rapid churn in a far-off region of the network would result in a lot of FDP flushing, even when we don't have any active connections to that region?

rade:
> Would I be right in thinking that rapid churn in a far-off region of the network would result in a lot of FDP flushing, even when we don't have any active connections to that region?

Active connections have no bearing on whether we flush routes. Not in the old code, and not in the new one.

The flushing decision was based purely on whether route calculation changed *any* routes. In a reasonably homogenous topology, very few topology changes won't result in a route change.

The canonical worst case is a single node connected to another node which in turn is connected to a whole bunch of other nodes. The routes on that single node are essentially constant irrespective of topology changes between the node bunch.

rade:
Another way to look at the FDP route flushing change is that the cost of calculating all routes up front - so we can figure out whether FDP routes need flushing - is far greater than flushing routes, dealing with FDP misses, and calculating the required routes lazily. In almost all cases. About the only exception being when the traffic pattern is such that a node is receiving broadcast traffic from nearly all peers between each successive topology update.

PR #1776, opened 2015-12-11
In the plugin, set up static route for multicast

### LINKED ISSUE #1752: Docker network plugin does not add multicast route
Due to the default route via the `docker_gwbridge` interface, the plugin needs to add a multicast route (see #92).

bboreham:
Fixes #1752

### DIFF
--- a/plugin/net/driver.go
+++ b/plugin/net/driver.go
@@ -21,19 +21,21 @@ const (
 )
 
 type driver struct {
-	version    string
-	nameserver string
-	scope      string
+	version          string
+	nameserver       string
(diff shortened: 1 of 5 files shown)


bboreham:
I added an option to disable, matching #1741.

awh:
This looks fine, but I haven't tested it locally due to the rigmarole of setting up a cluster store - will do so once this is rebased on mesh moad. We should also extent `test/195_no_multicast_route_1_test.sh` to include this...

bboreham:
Rebased and added test

PR #1777, opened 2015-12-11
Remove 'nameserver' option from plugin

bboreham:
We don't seem to be using it. 

Extracted from #1738 to simplify it; this change can be applied at a later date.

### DIFF
--- a/plugin/net/driver.go
+++ b/plugin/net/driver.go
@@ -22,16 +22,14 @@ const (
 
 type driver struct {
 	version          string
-	nameserver       string
 	scope            string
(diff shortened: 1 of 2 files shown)


bboreham:
More thought is required; maybe we _should_ be using this option; maybe it is required in some sub-mode of plugin operation.  

Maybe `-icc=false` ?

bboreham:
Note that, even if we did want this route, Docker 1.10 overrides our choice of DNS server.

awh:
LGTM; needs rebase.

bboreham:
Rebased

PR #1782, opened 2015-12-12
Add a 'json' function for 'weave report -f'

### LINKED ISSUE #1709: `weave report -f` should provide a `json` function
If you do `weave report` on its own, you get JSON output, but if you provide a format template then you only get the raw values, which can be hard to interpret, e.g.:

````
$ weave report -f "{{.DNS}}"
{weave.local. 172.17.42.1:53 1 [{foobar.weave.local. fe:e2:23:09:da:02 da001e7156cddd081b1e888ca150d228ec90301d83b65d47ead4a823b891f07b 10.32.0.1 0 0}]}
````

We could add a `json` function like [Docker has](https://github.com/docker/docker/blob/master/api/client/inspect.go#L18), i.e.:

````
$ weave report -f "{{json .DNS}}"
 {"Domain": "weave.local.","Address": "172.17.42.1:53",...
````

bboreham:
Fixes #1709

### DIFF
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -245,7 +245,13 @@ func HandleHTTP(muxRouter *mux.Router, version string, router *weave.NetworkRout
 
 	muxRouter.Methods("GET").Path("/report").Queries("format", "{format}").HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
-			formatTemplate, err := template.New("format").Parse(mux.Vars(r)["format"])
+			funcs := template.FuncMap{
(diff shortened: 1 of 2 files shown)


rade:
Well that's a bit annoying. This PR failed in the gossip tests, even though it is post #1767, which was supposed to have cured that.

rade:
Surely we should mention this feature in [the docs](http://docs.weave.works/weave/latest_release/troubleshooting.html#weave-report).

bboreham:
Added to docs.

PR #1791, opened 2015-12-15
Handle weave containers in 'restarting' state

### LINKED ISSUE #1778: improve check for "container running"
If a container is restarting, `State.Running` is still `true`.

    $ docker inspect --format='{{json .State}}' weaveplugin
    {"Status":"restarting","Running":true,"Paused":false,"Restarting":true,"OOMKilled":false,"Dead":false,"Pid":0,"ExitCode":2,"Error":"","StartedAt":"2015-12-11T16:30:57.160366578Z","FinishedAt":"2015-12-11T16:30:57.190084967Z"}

so the code in the weave script which examines `State.Running` needs to be amended.

For a container with a restart policy, we could make `check_not_running` rm the restarting container, so:

    weave launch-plugin --foo

creates an infinite restart loop because the argument is unrecognized, but a subsequent

    weave launch-plugin [some-other-args]

rm's the restarting one and starts the new one you asked for.

bboreham:
Fixes #1778

### DIFF
--- a/weave
+++ b/weave
@@ -915,12 +915,9 @@ container_weave_addrs() {
 
 # Check that a container named $1 with image $2 for component $3 is not running
 check_not_running() {
-    case $(docker inspect --format='{{.State.Running}} {{.State.Status}} {{.Config.Image}}' $1 2>/dev/null) in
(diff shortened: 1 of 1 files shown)


bboreham:
Replaced by #1814

PR #1792, opened 2015-12-15
Stop leaking 'w' files

bboreham:
As noted in #1757, we left some copies of `weavewait`, copied as `w`, lying around.

This PR creates an auxilliary container `weavevolumes` to hold the `/w/w` files, so the proxy maps into that instead of creating new volumes every time.

Replaces #1768

### DIFF
--- a/weave
+++ b/weave
@@ -1671,7 +1671,10 @@ launch_proxy() {
     docker_client_args $DOCKER_CLIENT_ARGS
     proxy_args "$@"
     mkdir -p /var/run/weave
-    docker create -v /w -v /w-noop -v /w-nomcast --name=$VOLUMES_CONTAINER_NAME $EXEC_IMAGE >/dev/null 2>&1 || true
+    # Create a data-only container to mount the weavewait files from
+    if ! docker inspect -f ' ' $VOLUMES_CONTAINER_NAME > /dev/null 2>&1 ; then
+       docker create -v /w -v /w-noop -v /w-nomcast --name=$VOLUMES_CONTAINER_NAME --entrypoint=/bin/false $EXEC_IMAGE >/dev/null
+    fi
     PROXY_CONTAINER=$(docker run --privileged -d --name=$PROXY_CONTAINER_NAME --net=host \
         $PROXY_VOLUMES \
         --volumes-from $VOLUMES_CONTAINER_NAME \
--- a/weave
+++ b/weave
@@ -339,6 +339,7 @@ IMAGE=$BASE_IMAGE:$IMAGE_VERSION
 BASE_PLUGIN_IMAGE=$DOCKERHUB_USER/plugin
 PLUGIN_IMAGE=$BASE_PLUGIN_IMAGE:$IMAGE_VERSION
 PLUGIN_CONTAINER_NAME=weaveplugin
+VOLUMES_CONTAINER_NAME=weavevolumes
 
 PROCFS=${PROCFS:-/proc}
 DOCKER_BRIDGE=${DOCKER_BRIDGE:-docker0}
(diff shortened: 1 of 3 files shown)


bboreham:
Cleaning up the `w` files breaks an assumption in test 195, that we can stop and restart the proxy and carry on relying on files mounted by the previous instance.

awh:
> Cleaning up the w files breaks an assumption in test 195, that we can stop and restart the proxy and carry on relying on files mounted by the previous instance.

The assumption specifically is that the following sequence is valid:

1. `weave launch-proxy`
2. `docker run` container A via proxy
3. `docker stop` container A via proxy
4. `weave stop-proxy`
5. `weave start-proxy`
6. `docker start` container A via proxy

Which is completely reasonable, and worked before this PR, volume leak notwithstanding; the only reason the breakage wasn't noticed on first submission is because there was no test that exercised that sequence of steps until the `--no-multicast-route` PR got merged.

In other words, the failure is not due to a bad assumption in `195_no_multicast_route_1_test.sh` - it's because the leak fix has introduced a new problem.

bboreham:
The "new" problem is that the system works if you leave files lying around forever.

I didn't say it was a bad assumption; however we need something creative to make progress.

awh:
> The "new" problem is that the system works if you leave files lying around forever.

To my mind that's the existing problem, which this PR fixes at the cost of introducing a different one. Not sure which is worse though - slowly leaking volumes, or a hard failure in a probably-unfrequently-used sequence of steps.

> I didn't say it was a bad assumption

Yes, sorry, I inferred that erroneously :smile:

> we need something creative to make progress.

That may have to wait until tomorrow :smile:

bboreham:
Maybe we can get round this by using a "volumes-only" container to hold on to `/w/w`, which sticks around until you do `weave reset`

(idea triggered by https://github.com/weaveworks/weave/issues/1859#issuecomment-172204574)

weave
```diff
@@ -2102,6 +2105,7 @@ EOF
         for NAME in $CONTAINER_NAME $PROXY_CONTAINER_NAME ; do
             docker stop  $NAME >/dev/null 2>&1 || true
             docker rm -f $NAME >/dev/null 2>&1 || true
+            docker rm -v $VOLUMES_CONTAINER_NAME
```
rade: this is rather wrong; you don't want this to be inside the loop, where it gets executed multiple times.

bboreham:
Better now.

weave
```diff
@@ -1670,8 +1671,13 @@ launch_proxy() {
     docker_client_args $DOCKER_CLIENT_ARGS
     proxy_args "$@"
     mkdir -p /var/run/weave
+    # Create a data-only container to mount the weavewait files from
+    if ! docker inspect -f ' ' $VOLUMES_CONTAINER_NAME > /dev/null 2>&1 ; then
+       docker create -v /w -v /w-noop -v /w-nomcast --name=$VOLUMES_CONTAINER_NAME --entrypoint=/w/w $EXEC_IMAGE >/dev/null
```
rade: I find the `/w/w` obscure. `/bin/false`?

rade:
This does fix #1757, right?

bboreham:
An excellent question.

bboreham:
In general it does, but our smoke-tests are masking the effect because they remove every container with `docker rm -f` before doing `weave reset`.

bboreham:
I added `-v` to the `docker rm` in `config.sh`.  It's a bit of a special-case because we are going through contortions in `start_suite` to avoid hanging Docker.

In short, yes it fixes #1757.

bboreham:
Maybe this should go into 1.4.3?

rade:
> Maybe this should go into 1.4.3?

I was wondering about that too. The change is quite simple, though it does introduce a new container. I cannot think of any upgrade issues. So I suggest we put this in 1.4.3.

bboreham:
Replaced by #1935 on 1.4 branch

PR #1794, opened 2015-12-16
Warn user if weave network is in use on stop or reset of plugin

bboreham:
If Docker perceives "active endpoints" on a network, it will attempt to contact the driver about them, and if you've shut down the plugin then this will time out.  So if you get into this state the best thing to do is restart the plugin.

This is particularly bad on a restart of Docker (e.g. on host reboot), when Docker may never get going, hence leave you unable to start the plugin.

### DIFF
--- a/weave
+++ b/weave
@@ -1712,7 +1712,14 @@ launch_plugin() {
         $PLUGIN_IMAGE "$@")
 }
 
+remove_plugin_network() {
+    docker run --rm --net=host \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        $PLUGIN_IMAGE --remove-network
+}
+
 stop_plugin() {
+    remove_plugin_network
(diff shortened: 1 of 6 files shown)


weave
```diff
@@ -2146,6 +2153,7 @@ EOF
         ;;
     reset)
         [ $# -eq 0 ] || usage
+        remove_plugin_network || echo "WARNING: docker may hang in this state; re-launch weave and remove active endpoints before shutting down" >&2
```
awh: What does 'remove active endpoints' look like in terms of docker commands? I suspect this phrase will be meaningless to most users

awh:
Ready to merge, save the comment on the warning message.

bboreham:
This PR should have removed the changes made in #1738 to put `docker network rm` into the smoke-test startup.

PR #1802, opened 2015-12-17
Update 'reboots' info

### LINKED ISSUE #1774: Troubleshooting doc "reboots" section is out of date
At http://docs.weave.works/weave/latest_release/troubleshooting.html, there is a heading "Reboots" which does not take into account the proxy reattaching the router, which came in in version 1.2.

Also, "Reboots" does not appear in the page's table of contents.

bboreham:
Fixes #1774


rade:
shouldn't this be on `latest_release_doc_updates`?

PR #1805, opened 2015-12-17
Fix indentation of weave script

bboreham:
I was confused for a while because the `else` lines up with the wrong `if`.

I checked the rest of the script and don't see anything else with the same problem.


rade:
There is really no need for a PR for this kind of stuff.

PR #1809, opened 2015-12-21
Add -installsuffix netgo

### LINKED ISSUE #742: Make line should use -installsuffifx in addition to -netgo
After installing a fresh copy of Go, attempting to build weave gives:

````
Your go standard library was built without the 'netgo' build tag.
To fix that, run
    sudo go clean -i net
    sudo go install -tags netgo std
````

These instructions overwrite Go's libraries, which may be inconvenient for some users.
This can be improved by adding `-installsuffix netgo` each time `-tags netgo` is used in the Makefile to make Go put the files in a different directory.
(Inspired by https://github.com/golang/go/issues/9369#issuecomment-69864440)

bboreham:
So we don't have to delete Go's regular standard library; the netgo and non-netgo variants exist side-by-side.  This eases development for people who switch between working on netgo-requiring and regular Go projexts.

Fixes #742 

Note that `go get` also does a `go install`, which is why that line was failing.  This change adds a speculative `go install` to see if we need to ask the user to do the standard library build under `sudo`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,15 +40,14 @@ WEAVE_EXPORT=weave.tar.gz
 WEAVEEXEC_DOCKER_VERSION=1.6.2
 DOCKER_DISTRIB=prog/weaveexec/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(WEAVEEXEC_DOCKER_VERSION).tgz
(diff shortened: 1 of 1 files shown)


bboreham:
It appears that Go 1.4 does not pass the `--installsuffix` flag from `go test` to the build, so this will fail the automated tests until we move to Go 1.5.

awh:
Needs rebase.

bboreham:
Now we are doing the build inside a container (after #1861) this change doesn't gain anything -> closing.

PR #1812, opened 2015-12-22
Issue a warning on --with-dns option to proxy, because it was removed

### LINKED ISSUE #1810: Warn, don't fail, on `--with-dns` option to the proxy
#1754 was a bit heavy-handed; the option should remain with a warning message

bboreham:
Fixes #1810

### DIFF
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -21,6 +21,7 @@ func main() {
 		justVersion bool
 		logLevel    = "info"
 		c           = proxy.Config{Image: "weaveworks/weaveexec"}
+		withDNS     bool
 	)
 
 	c.Version = version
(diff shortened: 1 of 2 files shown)


prog/weaveproxy/main.go
```diff
@@ -39,6 +40,7 @@ func main() {
 	mflag.BoolVar(&c.TLSConfig.Enabled, []string{"#tls", "-tls"}, false, "Use TLS; implied by --tlsverify")
 	mflag.StringVar(&c.TLSConfig.Key, []string{"#tlskey", "-tlskey"}, "", "Path to TLS key file")
 	mflag.BoolVar(&c.TLSConfig.Verify, []string{"#tlsverify", "-tlsverify"}, false, "Use TLS and verify the remote")
+	mflag.BoolVar(&withDNS, []string{"#-with-dns", "#w"}, false, "option removed")
```
rade: `"(deprecated)"`

prog/weaveproxy/main.go
```diff
@@ -57,6 +59,10 @@ func main() {
 	Log.Infoln("weave proxy", version)
 	Log.Infoln("Command line arguments:", strings.Join(os.Args[1:], " "))
 
+	if withDNS {
+		Log.Warning("--with-dns option has been removed; DNS is on by default")
```
rade: "...has been deprecated; ..."

rade: in fact I'd make the message identical to that in the script, i.e. "--with-dns is deprecated; it is on by default".

bboreham: "deprecated" suggests the behaviour is still there, but will be removed later.  This is not what happened.

rade: Then why does the script say "deprecated"?

bboreham: AFAIK that's a different `--with-dns` option, on `weave run`, where the behaviour did not change.

rade: No, I am not talking about `weave run` but the `--with-dns` case you added to `deprecation_warnings` in this very PR.

bboreham: Ah, overdid the copy'n'paste

bboreham: I fixed that message to say "removed".

PR #1814, opened 2015-12-22
Handle weave containers in 'restarting' state

### LINKED ISSUE #1778: improve check for "container running"

bboreham:
Fixes #1778 

Replacement for #1791 against 1.4 branch

### DIFF
--- a/weave
+++ b/weave
@@ -908,12 +908,9 @@ container_weave_addrs() {
 
 # Check that a container for component $1 named $2 with image $3 is not running
 check_not_running() {
-    case $(docker inspect --format='{{.State.Running}} {{.State.Status}} {{.Config.Image}}' $2 2>/dev/null) in
+    RUN_STATUS=$(docker inspect --format='{{.State.Running}} {{.State.Status}} {{.Config.Image}}' $2 2>/dev/null) || true
+    case ${RUN_STATUS%:*} in
         "true restarting $3")
-            echo "$2 is restarting; you can stop it with 'weave stop-$1'." >&2
-            exit 1
-            ;;
-        "true restarting $3:"*)
             echo "$2 is restarting; you can stop it with 'weave stop-$1'." >&2
             exit 1
             ;;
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -1850,19 +1848,19 @@ EOF
         ;;
     launch)
         deprecation_warnings "$@"
-        check_not_running $CONTAINER_NAME       $BASE_IMAGE
-        check_not_running $PROXY_CONTAINER_NAME $BASE_EXEC_IMAGE
+        check_not_running $CONTAINER_NAME       $BASE_IMAGE      router
+        check_not_running $PROXY_CONTAINER_NAME $BASE_EXEC_IMAGE proxy
```
rade: I'd put the component first.

weave
```diff
@@ -913,21 +913,19 @@ container_weave_addrs() {
...
-# Check that a container named $1 with image $2 is not running
+# Check that a container named $1 with image $2 for component $3 is not running
 check_not_running() {
-    case $(docker inspect --format='{{.State.Running}} {{.Config.Image}}' $1 2>/dev/null) in
-        "true $2")
-            echo "$1 is already running." >&2
+    RUN_STATUS=$(docker inspect --format='{{.State.Running}} {{.State.Status}} {{.Config.Image}}' $1 2>/dev/null) || true
+    case ${RUN_STATUS%:*} in
+        "true restarting $2")
+            echo "$1 is restarting; you can stop it with 'weave stop-$3'" >&2
```
rade: should have a full stop at end.

Also, perhaps we should print the 'stop' instructions for the non-restarting case too?

PR #1817, opened 2015-12-22
Check we have Docker 1.9 or newer before trying to remove a network

### LINKED ISSUE #1815: Failure on `weave stop` when run against Docker <1.9
````
$ weave launch
$ weave stop
FATA: 2015/12/22 12:04:20.337536 unable to connect to docker: API error (400): client is newer than server (client API version: 1.21, server API version: 1.20)
````

Believed to be because the 'network remove' code is run without regard to the possibility the user may have a Docker older than 1.9.

### PR DESCRIPTION
**Pull request by bboreham**
Fixes #1815 and #1816

### DIFF
--- a/weave
+++ b/weave
@@ -1713,9 +1713,11 @@ launch_plugin() {
 }
 
 remove_plugin_network() {
-    docker run --rm --net=host \
-        -v /var/run/docker.sock:/var/run/docker.sock \
-        $PLUGIN_IMAGE --remove-network
+    if [ -z $WEAVE_NO_PLUGIN ] && check_docker_server_api_version 1.21 ; then
+        docker run --rm --net=host \
+               -v /var/run/docker.sock:/var/run/docker.sock \
+               $PLUGIN_IMAGE --remove-network
+    fi
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: weave [hunk 1]
```diff
@@ -1713,9 +1713,11 @@ launch_plugin() {
...
 
 remove_plugin_network() {
-    docker run --rm --net=host \
-        -v /var/run/docker.sock:/var/run/docker.sock \
-        $PLUGIN_IMAGE --remove-network
+    if [ -z $WEAVE_NO_PLUGIN ] && check_docker_server_api_version 1.21 ; then
+        docker run --rm --net=host \
+               -v /var/run/docker.sock:/var/run/docker.sock \
+               $PLUGIN_IMAGE --remove-network
+    fi
```
rade: I find it rather odd that here the version check happens inside the function whereas for `launch_plugin()` it happens at (one of) the call sites.

bboreham: My thinking was that if the user has explicitly called `weave launch-plugin` then we should attempt to do what they asked for; all the other sites happen whether the user asked or not.

PR #1826, opened 2015-12-23
clean up GossipSender

rade:
See commits for explanation.


bboreham:
LGTM; has anything come up on this in the 13 days since last update?

PR #1831, opened 2015-12-24
don't stall gossip broadcasting when there are blocked connections

rade:
See commits for explanation.

Note that this PR is based on #1826.


rade:
replaced by  #1855

PR #1833, opened 2015-12-26
ensure gossip ordering required for fast topology convergence

### LINKED ISSUE #1793: re-ordering of topology gossip can delay convergence

rade:
Fixes #1793.

NB: this is branched off #1832.


rade:
Replaced by #1899.

PR #1847, opened 2016-01-02
extract script utilities into single binary

### LINKED ISSUE #1613: extract non-docker-related script utilities into single binary
We should extract the dp related utilities (`weave --create-datapath|--delete-datapath|--add-datapath-iface`) into a separate binary, together with other non-docker-related utilities used by the script (currently just `netcheck`). And invoke that from the weave script.

This will speed up `weave launch` when there is no bridge, and `weave reset` and container launch since it eliminates the temporary containers we currently need in order to run the dp utilities.

And it helps with #1566.

This came up as part of #1584.

rade:
...so that we a) don't have to start containers for invoking fastdp functions, which is costly, and b) reduce image size by having just a single executable for all utilities.

We fall back to invoking the new `weaveutil` in a container if we cannot find the exe. As a result, `weave --local ...` now runs both netcheck and docker-tls-args.

Fixes #1613.


rade:
btw, for me this reduces `time weave launch` from 2.8s to 1.7s.

PR #1861, opened 2016-01-07
Build in a container by default, using go 1.5 and vendored dependencies.

### LINKED ISSUE #1563: build errors when upgrading go
I upgraded my 'go' installation to version 1.5.1 and attempted to build weave 1.1:

first issue:
````
$ make clean
sudo docker rmi weaveworks/weave weaveworks/weaveexec
Untagged: weaveworks/weave:latest
Deleted: c0d8680c8396b4950e4ef0fa02c772e9ec12cd98279ec56bba793498b529d45b
Deleted: 18fa4a56295df3e03eac75dabce7506bbba18ee4de47ec6b694bbb347a28a70d
Untagged: weaveworks/weaveexec:latest
Deleted: eb43fc07d5d0bdcae2ba232be9e08ebfa3e812acb6db931f484ee40c8dad8e89
Deleted: 62ee4bd7f9d1423bcb8df136d4d112dac855c02aae40eb25ee99d4f798c8c238
Deleted: bee91de7eaf330042695a5a93885e6bb7c489394144e85c207bc2ca5132800fe
Deleted: 8c5d7262fe98bf3f7a14b19987e7812b0da22f9ea7e3007945ae524cfe74c37b
go clean -r ./...
can't load package: tools/socks/main.go:11:2: cannot find package "github.com/armon/go-socks5" in any of:
	/usr/local/go/src/github.com/armon/go-socks5 (from $GOROOT)
	/home/bryan/golang/src/github.com/armon/go-socks5 (from $GOPATH)
make: *** [clean] Error 1
````

worked around by removing `tools` subdirectory that was left from previous master build.

next issue:

````
$ make update
go get -u -f -v -tags netgo ./prog/weaver/ ./prog/sigproxy/ ./prog/weaveproxy/ ./prog/weavewait/ ./prog/weavehosts/ ./prog/netcheck/ ./testing/cover/ ./testing/runner/
github.com/weaveworks/weave (download)
[...]
github.com/vishvananda/netlink (download)
package github.com/weaveworks/weave/testing/cover: no buildable Go source files in /home/bryan/golang/src/github.com/weaveworks/weave/testing/cover
github.com/mgutz/ansi (download)
make: *** [update] Error 1
````

worked around by running `make update` again.

next issue:

````
$ make
go get -tags netgo ./prog/weaver
go install net: open /usr/local/go/pkg/linux_amd64/net.a: permission denied
make: *** [prog/weaver/weaver] Error 1
````

worked around by `sudo go install -tags netgo std`.

### LINKED ISSUE #1656: Vendor dependencies with Go 1.5 and gvt

### LINKED ISSUE #1657: Update containerised build and Vagrantfiles to Go 1.5
Related #1656, #1563

### LINKED ISSUE #1850: enable 'clean', 'test' etc Makefile targets to be run in containerised build
In [Building Weave](http://docs.weave.works/weave/latest_release/building.html) docs, it would be useful to have some info on how to run specific Makefile commands in the not-so-obvious build scenarios, where one doesn't directly run a `make` command.

Specifically, when looking into adding a feature to weave by building it in a container, I was able to build it, but not quite sure how to run other Makefile commands, like `clean` and `tests`.

More verbose dev-ramp-up docs can help in making it easier for occasional contributors to contribute :-)

### PR DESCRIPTION
**Pull request by tomwilkie**
Fixes #1657, fixes #1656, fixes #1850.  I think it also fixes #1563.

This PR:
- makes the default, and only, way of building weave use a container
- make container go version 1.5.2
- uses vendored dependancies
- removes any traces of `go get` to prove the vendored dependancies are all there

TODO:
- [ ] make sure everyone is happy with the approach, then:
- [x] cache build image on circle
- [ ] cache vendored dependancies on circle? (not necessary if we submodule...)
- [x] update http://docs.weave.works/weave/latest_release/building.html

### CODE REVIEW

**Comment by dpw** [comment]
Why the use of gvt?  What's wrong with just GOVENDOREXPERIMENT plus git submodules (or subtrees if you prefer)?

**Comment by tomwilkie** [comment]
> Why the use of gvt? What's wrong with just GOVENDOREXPERIMENT plus git submodules (or subtrees if you prefer)?

This is using GOVENDOREXPERIMENT.

Matthias doesn't want the dependancies checked in (ruling out manually fetching the dependancies or using subtree).  I don't want to use submodules, as they are a pain.

The requirements were to pin a specific version of each dependancy, but have them fetched at build time.  `gvt` does exactly this for us.

**Comment by dpw** [comment]
> Matthias doesn't want the dependancies checked in (ruling out manually fetching the dependancies or using subtree). I don't want to use submodules, as they are a pain.

Could you elaborate on how submodules are a pain?  It's possible to make their use very inconspicuous in this context.

The advantage of using submodules is that there is no need to learn the conventions of any tools beyond go and git.

(I'd guess subtrees might work as well, but I'm not very familiar with them.)

> The requirements were to pin a specific version of each dependancy, but have them fetched at build time. gvt does exactly this for us.

I notice that gvt is only referred to from circle.yml.  Presumably it needs to be run to retrieve the dependencies when doing a non-circle build?   How does that happen?

**Comment by tomwilkie** [comment]
> Could you elaborate on how submodules are a pain? It's possible to make their use very inconspicuous in this context.

I'm assuming your suggest is we subtree in every dependancy?  And the do a git submodule sync/update/whatever in the makefile to ensure they are always on the right version?

We've been using gvt in scope & the service and we've found it massively easier than dealing with submodules.  The submodule for tools has already causes problems for some people (Bryan, I think).

> I notice that gvt is only referred to from circle.yml. Presumably it needs to be run to retrieve the dependencies when doing a non-circle build? How does that happen?

This PR is marked work in progress; we were just discussing how to do that in the net room.  The PR should now include a mechanism (via an uptodate file) to ensure gvt is run correctly.

**Comment by rade** [comment]
You have collapsed the WEAVEUTIL_EXE rule into SIGPROXY_EXE, DOCKEPLUGIN_EXE, etc rule. That has two undesirable consequences:

1.  the comment above that rule is now wrong since it says the programs in question do not import the net package, which weaveutil *does* import.
2. the NETGO_CHECK is no longer run on weaveutil.

**Comment by rade** [comment]
You removed the `travis: $(EXES)` target. I use that a lot when working on weave, since it's much faster than a full build. Happy to see it renamed, but not gone.

**Comment by tomwilkie** [comment]
re: travis: $(EXES) yes I removed that as I removed all the travis build stuff.  Will re-add as `make exes`.

**Comment by rade** [comment]
This also fixes #1850, doesn't it?

http://docs.weave.works/weave/latest_release/building.html needs updating.

**Comment by tomwilkie** [comment]
Just the TestHTTPCancel failure now, which @bboreham is working on.

**Comment by dpw** [comment]
> We've been using gvt in scope & the service and we've found it massively easier than dealing with submodules. 

Well, we've been using submodules for vendoring in ambergreen, and it is so inconspicuous that it is hard for me to see how anything could be massively easier.

> The submodule for tools has already causes problems for some people (Bryan, I think).

Without knowing details I can't respond.  If it goes beyond simple unfamiliarity, then maybe more details would be relevant.  Like git in general, submodules have a learning curve.  But that will be true of any vendoring tool as well.  The difference is that some people are already familiar with submodules, and experience gained with them is widely reusable.

Also, from https://golang.org/s/go15vendor :
> The environment variable also enables fetching of git submodules during “go get”. This is meant to allow experiments to understand whether git submodules are an appropriate and useful way to vendor code.

**Comment by tomwilkie** [comment]
I'll give the submodules a try.

File: Makefile [hunk 1]
```diff
@@ -69,38 +75,54 @@ $(WEAVEWAIT_NOOP_EXE): prog/weavewait/*.go
...
+else
+
+$(VENDOR_UPTODATE):
+	gvt rebuild
+	touch $@
+
+$(WEAVEUTIL_EXE):
+	go build $(BUILD_FLAGS) -o $@ ./$(@D)
+	$(NETGO_CHECK)
+
```
rade: The WEAVEUTIL_EXE build stanza has move gratuitously.

rade:
Linting should be containerised.

Makefile
```diff
@@ -69,38 +75,54 @@ $(WEAVEWAIT_NOOP_EXE): prog/weavewait/*.go
 $(WEAVEWAIT_EXE): prog/weavewait/*.go net/*.go
 $(WEAVEWAIT_NOMCAST_EXE): prog/weavewait/*.go net/*.go
 
+ifeq ($(BUILD_IN_CONTAINER),true)
+
+$(EXES) $(VENDOR_UPTODATE):
+	@mkdir -p $(shell pwd)/.pkg
+	$(SUDO) docker run $(RM) $(RUN_FLAGS) \
+	    -v $(shell pwd):/go/src/github.com/weaveworks/weave \
+		-v $(shell pwd)/.pkg:/go/pkg \
```
rade: surely `.pkg` should be added to .gitignore.

Makefile
```diff
@@ -127,8 +149,16 @@ $(WEAVE_EXPORT): $(IMAGES_UPTODATE)
 $(DOCKER_DISTRIB):
 	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
 
-tests: tools/.git
-	tools/test
+ifeq ($(BUILD_IN_CONTAINER),true)
+tests: $(BUILD_UPTODATE) $(VENDOR_UPTODATE) tools/.git
+	$(SUDO) docker run $(RM) $(RUN_FLAGS) \
+		-v $(shell pwd):/go/src/github.com/weaveworks/weave \
```
rade: why is this not also mounting `.pkg`? Won't it end up building all libs from scratch as a result?

Makefile
```diff
@@ -127,8 +149,16 @@ $(WEAVE_EXPORT): $(IMAGES_UPTODATE)
...
+		-e GOARCH -e GOOS -e CIRCLECI -e CIRCLE_BUILD_NUM -e CIRCLE_NODE_TOTAL -e CIRCLE_NODE_INDEX -e COVERDIR\
```
rade: those circleci env vars are annoying. Nothing else in the makefile makes any reference to circleci.

rade:
I am not keen on eventually having all our main Makefile targets look like:
```make
ifeq $(BUILD_IN_CONTAINER,true)
target:
        $(SUDO) docker run ... target
else
target:
        the real build commands
fi
```

How can we avoid that?

tomwilkie:
> I am not keen on 
I can't think of another way, maybe @dpw can.  I've made it so we only have one big if now, and only two docker runs (ie not one for every target).

> Linting should be containerised.
Done.  Added golint to container.

> The WEAVEUTIL_EXE build stanza has moved gratuitously.
Sorry.  Done.

> why is this not also mounting .pkg? Won't it end up building all libs from scratch as a result?
(for tests) go test --help says:
```
	-i
	    Install packages that are dependencies of the test.
	    Do not run the test.
```
Working on figuring out if this is true or not.

tomwilkie:
> those circleci env vars are annoying. Nothing else in the makefile makes any reference to circleci.

yup.  what can you do?

Makefile
```diff
@@ -68,39 +74,70 @@ $(TEST_TLS_EXE): test/tls/*.go
...
+	$(SUDO) docker run $(RM) $(RUN_FLAGS) \
+	    -v $(shell pwd):/go/src/github.com/weaveworks/weave \
+		-v $(shell pwd)/.pkg:/go/pkg \
+		-e GOARCH -e GOOS \
+		$(BUILD_IMAGE) COVERAGE=$(COVERAGE) WEAVE_VERSION=$(WEAVE_VERSION) $@
+
+tests lint:
+	$(SUDO) docker run $(RM) $(RUN_FLAGS) \
+		-v $(shell pwd):/go/src/github.com/weaveworks/weave \
+		-v $(shell pwd)/.pkg:/go/pkg \
```
rade: presumably this needs a `@mkdir -p $(shell pwd)/.pkg` first, just like in the preceding stanza.

Makefile
```diff
@@ -68,39 +74,70 @@ $(TEST_TLS_EXE): test/tls/*.go
...
+		-e GOARCH -e GOOS -e CIRCLECI -e CIRCLE_BUILD_NUM -e CIRCLE_NODE_TOTAL -e CIRCLE_NODE_INDEX -e COVERDIR \
+		$(BUILD_IMAGE) $@
```
rade: Can't you merge that with the preceding build stanza, i.e. have just one stanza with target `$(EXES) $(VENDOR_UPTODATE) tests lint`?

tomwilkie:
re: `go test -i`

This is interests: https://code.google.com/p/go/issues/detail?id=3896

Basically, sticking a `go test -i $GO_TEST_ARGS $TESTDIRS` in tools/test before we run the tests make the whole thing faster it seems - just getting numbers now.

Makefile
tomwilkie: Yeah have that locally now, thanks.

Makefile
tomwilkie: I can yes, I was just seeing if there was anything we can do about the env vars, but I can't see anything.

tomwilkie:
Maybe not a massive speed up, but a welcome one:
```
# with the go test -i
vagrant@vagrant-ubuntu-vivid-64:~/src/github.com/weaveworks/weave$ rm -rf .pkg
vagrant@vagrant-ubuntu-vivid-64:~/src/github.com/weaveworks/weave$ time make tests
...
real	4m0.493s
user	0m0.032s
sys	0m0.112s

# without the go test -i
vagrant@vagrant-ubuntu-vivid-64:~/src/github.com/weaveworks/weave$ rm -rf .pkg
vagrant@vagrant-ubuntu-vivid-64:~/src/github.com/weaveworks/weave$ time make tests
...
real	5m18.121s
user	0m0.024s
sys	0m0.188s
```

rade:
So this PR contains *five* changes:
1. switch to go 1.5 (#1657)
2. vendoring (#1656)
3. reduce travis role to just doc publishing
4. make all Makefile targets in principle executable in containerised build and remove containerised "clone & build" (#1850)
5. drive containerised build & test & lint from Makefile (#1563)

AFAICT, the only real dependencies are 1->2, and 4->5. Correct?

Mixing all these in a single commit, in a single PR, makes it rather more difficult to review & iterate on these. I shall try to extract a few things into separate master commits / PRs. I have already done that for a couple of small refactors.

rade:
>  3. reduce travis role to just doc publishing

Extracted (and improved) in #1864.

Makefile
```diff
@@ -1,10 +1,14 @@
...
 .DEFAULT: all
-.PHONY: all update tests lint publish $(PUBLISH) clean clean-bin prerequisites build travis run-smoketests
+.PHONY: all update tests lint publish $(PUBLISH) clean clean-bin prerequisites build exes run-smoketests
 
 # If you can use docker without being root, you can do "make SUDO="
-SUDO=sudo
+SUDO=sudo -E
+BUILD_IN_CONTAINER=true
+RM=--rm
+RUN_FLAGS=-ti
```
rade: I don't see why RUN_FLAGS needs to be configurable.

build/build.sh
```diff
@@ -1,45 +1,31 @@
...
-    # will have awkward ownership.  So we switch to a user with the
-    # same user and group IDs as source directory.  We have to set a
-    # few things up so that sudo works without complaining later on.
-    uid=$(stat --format="%u" $WEAVE_SRC)
-    gid=$(stat --format="%g" $WEAVE_SRC)
-    echo "weave:x:$uid:$gid::$WEAVE_SRC:/bin/sh" >>/etc/passwd
-    echo "weave:*:::::::" >>/etc/shadow
-    echo "weave	ALL=(ALL)	NOPASSWD: ALL" >>/etc/sudoers
+# Mount the weave repo:
+#  -v $(pwd):/go/src/github.com/weaveworks/scope
```
rade: scope? huh? The whole comment seems completely out of place.

Makefile
tomwilkie: In certain environments (ie when you're running a command via ssh) docker run won't run with -ti.  I even tried ssh with a tty.

build/build.sh
tomwilkie: Sorry thats a copy and paste error.

tomwilkie:
@dpw do you have a link to the repo where you use submodules for vendoring?

dpw:
https://github.com/squaremo/ambergreen

Although you won't find a lot there to support the use of submodules, because not much is needed.  There's `get_vendor_submodules` in the `Makefile`, which avoids the need to explicitly use of `git submodule` for most purposes, but that's it.

(Just noticed another advantage of submodules: When browsing the repo in github, if you click down into one of the submodules dirs under `vendor/github.com`, you have a link to the specific version of the vendor repo.)

bboreham:
Can we add `--installsuffix` (similar to #1809), or is it undesirable in this new build?

tomwilkie:
> Can we add --installsuffix (similar to #1809), or is it undesirable in this new build?

As all the go runtime we need lives in the build container, I don't think its necessary.  But then I don't think it would be a problem.

> https://github.com/squaremo/ambergreen

@dpw I see - you run `git submodule update --init` if `vendor` has no files in it?  How do you deal with fetching new version of dependancies when they change?

dpw:
> How do you deal with fetching new version of dependancies when they change?

Hasn't happened yet.  I imagine we'd just make the `git submodule --init` unconditional.  I'm not sure why I made it conditional in the first place, maybe I had the mistaken idea that it would always go to the network.

tomwilkie:
@dpw the link you provided also has the following paragraph, right after the one you quoted:

> Note that when “go get” fetches a new dependency it never places it in the vendor directory. In general, moving code into or out of the vendor directory is the job of vendoring tools, not the go command.

I've just confirmed running `go get` doesn't setup submodules in the vendor dir for you, as you suggested.  How did you setup you vendor dir for ambergreen?

dpw:
> I've just confirmed running go get doesn't setup submodules in the vendor dir for you, as you suggested. How did you setup you vendor dir for ambergreen?

Yeah, I thought I did, but clearly I didn't. (I wondered if I found some contorted way to persuade `go get` to do this, but that doesn't seem to be possible).

So I suspect I just did a regular `go get` on the pre-vendor repo, and then a shell one-liner to turn the contents of the GOPATH into a series of `submodule add` commands.

tomwilkie:
@dpw @rade PTAL, let me know what you think to the sub modules. 

And @rade let me know if you want this broken into two prs.

rade:
Say I've started depending on a new library which in turn depends on a bunch of other libs. What do I run to add the new dependencies?

How do I update a specific dependency to a new version? go-dockerclient springs to mind, which we need to update in order to be able to use recent docker features. And what happens to the dependencies of that when I do the update?

How does one update all dependencies to the most recent version? Note that this should also *remove* any dependencies that are no longer needed (as well as, obviously, add any new ones). It's the kind of thing I envisage us doing every now and then, i.e. we'd run such an update, do a bunch of tests and pin any breakages to earlier working versions.

tomwilkie:
The answer to all of these, with the PR as it stands, is manually. Using
git submodule.

It's worth noting that gvt isn't a panacea. The gvt rebuild is slow, and
you have to do it too often. Gvt does allow you to add new dependencies
easily, and recursively. It does have a command to fetch the latest version
of a dependency (or all dependencies). It won't recursively update things,
nor will it remove stuff that's no longer needed.

I don't think any of these are frequent operations, so I think we should
optimise for the frequent things (get specified version of all
dependencies, which is quicker with submodules). We can then either develop
tooling to help with the above, or improve gvt to do shallow fetch etc.

rade:
> The answer to all of these, with the PR as it stands, is manually. Using git submodule.

That's not a great answer. I don't want us to accumulate cruft, or be stuck with ancient versions. Lack of tooling encourages that.

Surely you didn't produce the .gitmodules and associated "Subproject commit" files manually. So in my, possibly incredibly naive view, it should be easy to remove all those artifacts and re-create them from scratch, using whatever horrific script you cooked up to create them in the first place.

tomwilkie:
Yes, I did:

```bash
go list -f '{{join .Deps "\n"}}' ./prog/* | grep -E '^(github.com|k8s.io)' | grep -v 'github.com/weaveworks/weave' | sort -u > deps
cd vendor
cat ../deps | while read repo; do git submodule add -f --depth 1 https://$repo.git $repo; done
```

It choked on some of the golang.org imports, which I fixed up manually.  And it misses some of the test dependancies, which I added manually.  It wouldn't be too hard to put together some scripts which do this properly.

build/Dockerfile
```diff
@@ -1,24 +1,9 @@
...
-RUN curl -sSL https://golang.org/dl/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /usr/local -xz
-ENV PATH /usr/local/go/bin:$PATH
-
-# So that sudo does not trample on PATH
-RUN sed -ie '/secure_path/d' /etc/sudoers
-
+FROM golang:1.5.2
+ENV GO15VENDOREXPERIMENT 1
+RUN apt-get update && apt-get install -y libpcap-dev python-requests time
+RUN go get github.com/FiloSottile/gvt; touch /tmp/.donotremove
```
awh: Can this reference to `gvt` be removed now?

site/building.md
```diff
@@ -12,20 +12,14 @@ Apart from the `weave` shell script, weave is delivered as a set of
...
-does not include static
-libraries](http://fedoraproject.org/wiki/Packaging:Guidelines#Packaging_Static_Libraries).
-So we recommend building under Ubuntu.
+later).
 
-You can also build in a container under any system that supports
-Docker.  And you can run Ubuntu in a VM and build there.  These
-options are described below.
+The only way to build is by using the the build container; the
+Makefile if setup to make this transparent.  This method is
```
awh: is setup

site/building.md
```diff
@@ -56,59 +40,16 @@ Then to actually build, simply do:
...
-to bind the bind your go workspace containing the weave repository to
-`/home/go` inside the container.  No container arguments should be
-passed in this case:
-
-```bash
-$ sudo docker run -v /var/run/docker.sock:/var/run/docker.sock -v <host gopath>:/home/go weaveworks/weave-build
-```
+On a fresh repository, the Makefile will do the following:
+- assemble the build container
+- download specific versions of all the dependancies
```
awh: dependencies

tomwilkie:
Rebased, squashed, and remove the incorrectly vendored golint and gvt source.

bboreham:
I was interested to find that this build is faster for small changes.  For example, after having done a full build:

    $ touch net/route.go
    $ time make prog/weaver/weaver

On my machine the old build takes 6.9s and the new takes 4.1s.

This appears to be because the `go get` in the old Makefile was linking `weaver` to install it in `$GOPATH/bin`, although we didn't need or want that.

bboreham:
As discussed elsewhere, `make update` should be re-introduced.

PR #1864, opened 2016-01-09
Shrink travis usage

rade:


tomwilkie:
We could remove the `travis` target from the Makefile too?  Rename it `exes`?

rade:
> We could remove the `travis` target from the Makefile too?  Rename it `exes`?

Done.

Issue #894, opened 2015-06-11
Rolling restart can leave ipam hung - not if you do it correctly; but need to document this


bboreham:
If you have peers A, B, C running, then you kill A, start a new A', kill B, start a new B', etc., then you will end up with a ring owned by the old A, B, C and none of the running peers have any space.  If they try to ask for any space they will only try to contact the now-gone peers.

(To trigger this symptom requires that the new peers have unique PeerName, i.e. the weave bridge has also been recreated)

This can be worked around by calling `weave rmpeer` on the new peers or (somewhat less reliably) by calling `weave reset` when shutting down each peer.  Or by shutting down all the old ones before starting any new ones.

The weave logs are not very helpful; just a series of messages like:

    weave 2015/06/11 12:20:12.412849 [gossip IPallocation]: unknown relay destination: 02:5e:a2:16:13:16

rade:
This will be resolved by #678. @bboreham did you have some other, separate solution in mind? Or some stop-gap, such as making it more detectable for users what is happening and what they should do about it?

bboreham:
Mostly wanted to document the problem and current mitigations.
Post-#1010 there is a message to tell you there are no peers available to talk to, but I think it's `-debug` level.

rade:
This got fixed in #1624. except for the case where the hosts are restarted w/o rmpeer, i.e. w/o invoking `weave reset`. We should document the need to invoke `weave reset` on a rolling restart, or indeed any controlled node shutdown.

rade:
> We should document the need to invoke `weave reset` on a rolling restart, or indeed any controlled node shutdown.

That is no longer true, post #1866. -> fixed.

PR #1866, opened 2016-01-10
Derive peer name from product_uuid

### LINKED ISSUE #901: retain peer identity across host reboots
ATM the peer's identity is the MAC address we assign to the weave bridge. This changes on reboots since the bridge doesn't survive them. One consequence of that is that we are prone to leaking IP addresses allocated with IPAM. See #678 and #894.

rade:
Fixes #901.

### DIFF
--- a/site/ipam.md
+++ b/site/ipam.md
@@ -176,20 +176,26 @@ for manual allocation.
 
 ## <a name="stop"></a>Stopping and removing peers
 
-You may wish to `weave stop` and re-launch to change some config or to
-upgrade to a new version; provided the underlying protocol hasn't
-changed it will pick up where it left off and learn from peers in the
-network which address ranges it was previously using. If, however, you
-run `weave reset` this will remove the peer from the network so
-if Weave is run again on that node it will start from scratch.
+You may wish to `weave stop` and re-launch to change a peer's
+configuration or upgrade it to a new version. Provided the underlying
+protocol hasn't changed the peer will pick up where it left off and
+learn from other peers in the network which address ranges it was
+previously using.
 
-For failed peers, the `weave rmpeer` command can be used to
-permanently remove the ranges allocated to said peer.  This will allow
-other peers to allocate IPs in the ranges previously owner by the rm'd
-peer, and as such should be used with extreme caution - if the rm'd
-peer had transferred some range of IP addresses to another peer but
-this is not known to the whole network, or if it later rejoins
-the Weave network, the same IP address may be allocated twice.
+The same normally happens when a peer is restarted any other way,
+e.g. as result of a reboot, provided the
+[system-uuid](http://linux.die.net/man/8/dmidecode) didn't change.
+
+If you want to remove a peer from the network, run `weave reset`. This
+removes the ranges allocated to the peer, thus allowing other peers to
+allocate IPs in them.
+
+For a failed peer, the `weave rmpeer` command can be run on any other
+peer to achieve the same result. The command should be used with
+extreme caution - if the rm'd peer had transferred some range of IP
+addresses to another peer but this is not known to the whole network,
+or if it actually had not failed and later rejoins the network, the
+same IP address may be allocated twice.
 
 Assuming we had started the three peers in the example earlier, and
 host3 has caught fire, we can go to one of the other hosts and run:
--- a/weave
+++ b/weave
@@ -432,12 +432,18 @@ arp_update() {
     ! command_exists arping || $3 arping -U -q -I $1 -c 1 ${2%/*}
 }
 
-# Generate a random MAC value
-random_mac() {
+# Generate a MAC value from a stdin containing six space separated
+# 2-digit hexadecimal numbers.
+mac_from_hex() {
     # In the first byte of the MAC, the 'multicast' bit should be
     # clear and 'locally administered' bit should be set.  All other
     # bits should be random.
-    od -txC -An -N6 /dev/urandom | ( read a b c d e f && printf "%02x:$b:$c:$d:$e:$f" $((0x$a & ~1 | 2)) )
+    read a b c d e f && printf "%02x:$b:$c:$d:$e:$f" $((0x$a & ~1 | 2))
+}
+
+# Generate a random MAC value
+random_mac() {
+    od -txC -An -N6 /dev/urandom | mac_from_hex
 }
 
 ######################################################################
@@ -585,10 +591,22 @@ init_bridge() {
 
     ip link add name $BRIDGE type bridge
 
-    # Set a random MAC address on the bridge.  Current Linux
-    # kernels already do this when creating a bridge, but there
-    # are rumours it was not always so.
-    ip link set dev $BRIDGE address $(random_mac)
+    # Derive the bridge MAC from the system (aka bios) UUID, if
+    # present. Elsewhere we in turn derive the peer name from that,
+    # which we want to be stable across reboots but otherwise
+    # unique. The system UUID fits that bill, unlike, say,
+    # /etc/machine-id, which is often identical on VMs created from
+    # cloned filesystems. If we cannot determine the system UUID we
+    # just generate a random MAC.
+    if SYSTEM_UUID=$(cat /sys/class/dmi/id/product_uuid 2>/dev/null) ; then
+        # We salt the input just as a precaution to avoid clashes with
+        # other applications who might have had the bright idea of
+        # generating MACs in the same way.
+        MAC=$(echo "9oBJ0Jmip-$SYSTEM_UUID" | sha256sum | sed 's/\(..\)/\1 /g' | cut -c1-17 | mac_from_hex)
+    else
+        MAC=$(random_mac)
+    fi
+    ip link set dev $BRIDGE address $MAC
 
     # Attempting to set the bridge MTU to a high value directly
     # fails. Bridges take the lowest MTU of their interfaces. So
@@ -1514,12 +1532,9 @@ launch_router() {
     enforce_docker_bridge_addr_assign_type
     create_bridge
     docker_bridge_ip
-    # We set the router name to the bridge mac since that is
-    # stable across re-creations of the containers.
-    #
-    # TODO use the mac of one of the physical host interfaces
-    # (eth0, wlan0, etc) so the name becomes stable across host
-    # restarts.
+    # We set the router name to the bridge MAC, which in turn is
+    # derived from the system UUID (if available), and thus stable
+    # across reboots.
     PEERNAME=$(cat /sys/class/net/$BRIDGE/address)
     # backward compatibility...
     if is_cidr "$1" ; then


rade:
There is a gotcha here for upgrades: `weave stop; <upgrade>; weave launch` will change the peer identity. This is bad since it prevents reclaiming of container IPs.

The obvious fix is that instead of deriving the peer name from the machine id, we continue to derive it from the bridge mac, and instead derive the bridge mac from the machine id (replacing the random generation). That's what I was going to do initially but it's less clean. Deriving MACs from permanent IDs is dodgy since it will result in clashes with any other program doing the same (though I suppose we could get around that by hashing the machine-id with a weave-specific salt). It also prevents us from gaining 2 bits (the bridge MAC must have the 'multicast' bit clear and the 'locally administered' bit set).

I haven't got any better idea though. Thoughts?

bboreham:
Would also fix #894, I think.

rade:
Why is CircleCI failing? I suspect all hosts have the same machine-id. The `docker info` reported by the test runs for each of the hosts shows that the docker daemon has the same ID (`YTMJ:U24E:PHCW:JR6D:HP3T:KP5N:XN7J:SNOR:BKGZ:25OB:WIOI:VWJP`) on all of them, and while the daemon ID is not derived from the machine-id (afaict), this nevertheless indicates that these hosts are all clones.

This is problematic, and I cannot think of a fix.

Presumably some docker tools, e.g. swarm, will be unhappy with multiple daemons having the same ID, so it's not just a problem for weave.

rade:
@tomwilkie is there anything peculiar about how the circle hosts were created?

If this "all the hosts look the same" issue is something that users can easily run into then this whole approach I am pursuing here is dead :(

bboreham:
Nitpicking: they're "GCE hosts" that we happen to create by running a script (weave/test/gce.sh) on CircleCI.

rade:
> This (lack of unique machine-id) is problematic, and I cannot think of a fix.

I switched to using the BIOS ID, which seems to work.

awh:
Copied from https://github.com/weaveworks/weave/issues/901#issuecomment-170854398

> I am really worried that some common way of creating "machines" via some cloning/copying/whatever mechanism might result in ID duplicates.

I think before we merge this we need to ensure that snapshots/clones (or whatever they're called in the respective vendor's terminology) don't duplicate the `product_uuid`:

* [ ] VirtualBox
* [x] DigitalOcean
* [ ] AWS
* [ ] Azure
* [ ] GCE

rade:
@errordeveloper can you help with that?

site/ipam.md
```diff
@@ -175,20 +175,27 @@ for manual allocation.
...
-the Weave network, the same IP address may be allocated twice.
+You may wish to `weave stop` and re-launch to change a peer's
+configuration or upgrade it to a new version. Provided the underlying
+protocol hasn't changed the peer will pick up where it left off and
+learn from other peers in the network which address ranges it was
+previously using.
+
+The same normally happens when a peer is restarted any other way,
+e.g. as result of a reboot, provided the
+[machine-id](http://man7.org/linux/man-pages/man5/machine-id.5.html)
```
awh: `product_uuid`

awh:
@rade does this not add a mandatory requirement for `weave` to be run as root, even if your user can access docker without?

rade:
That requirement is already there. `weave --local` checks that you are running as root, and weaveexec is run with `--privileged` for a reason.

awh:
D'oh, yes of course.

errordeveloper:
I'm not sure why we haven't listed the popular hypervisors that people are likely to use, and I found this one [note out in the wild wild web](https://gist.github.com/ggrandes/9353295).

One useful [note regarding VMWare](http://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1541). So whether you moved or copied a VM on your machine, it will still appear to be unique. It's the same machine if you haven't moved the files.

I'm intending to confirm what's the case with KVM exactly, whilst at it.

So I'm suggesting to append the following to our list:
- [X] VMWare (tested using AppCatalyst, [should apply to all products](http://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1541))
- [ ] KVM
- [ ] VirtualBox

Also, testing these would give us a better coverage for smaller providers that are not covered above, as those are quite likely to run on either VMWare or KVM.

errordeveloper:
The [`systemd-detect-virt`](https://github.com/systemd/systemd/blob/1eb30a4c5c14a3aa11eeb8bbee6e0fcef7eaee29/src/basic/virt.c) tool may be of some interest here also.

errordeveloper:
On VMWare:
```
> docker run -ti alpine cat /sys/class/dmi/id/product_uuid
564D74BB-2119-E8E8-10D0-77D1195A51A2
> sudo cat /sys/class/dmi/id/product_uuid
564D74BB-2119-E8E8-10D0-77D1195A51A2
```

On EC2:
```
> docker run -ti alpine cat /sys/class/dmi/id/product_uuid
EC209BC9-2CD3-41AB-1D79-171974C47FBC
> sudo cat /sys/class/dmi/id/product_uuid
EC209BC9-2CD3-41AB-1D79-171974C47FBC
```

errordeveloper:
Also, regarding `/sys/hypervisor/uuid ` on EC2, these days it appears identical to `/sys/class/dmi/id/product_uuid`, only lowercase.

```
> cat /sys/hypervisor/uuid 
ec27dad0-2b18-1ef6-9650-42cdb21f0495
> sudo cat /sys/class/dmi/id/product_uuid
EC27DAD0-2B18-1EF6-9650-42CDB21F0495
```

errordeveloper:
I have checked, and cloning an instance into an image and starting a new instance from that did result in a new UUID. Also, rebooting instance from within itself resulted in UUID persistence and rebooting it via 'Stop' and 'Start' in the AWS console still resulted in UUID persistence, which @awh said wasn't the case in DO. I haven't waited very long between stopping and starting it again, so not sure if it could potentially end-up on a different physical host or how exactly that works... In any case, please do not that these UUIDs begin with 'EC2', it's kind of amusing.

rade:
> So I'm suggesting to append VMWare and KVM to our list.

Have added these to the list in @awh's comment, so we keep just one list.

rade:
I've tested with VirtualBox (5.0.12) on my ubuntu machine, cloning our host1 vagrant test VMs.

When I first start the VM right after cloning, it has the same product_uuid as the parent. This persists even across reboots. However, when "power off" the VM and restart it, it gets a fresh product_uuid. And this then remains unchanged even across further power-offs.

I tried this with both "Full" and "Linked" clones, and with and without telling it to re-set the MAC addresses. Curiously, when I tell it to change MACs, the MAC doesn't actually change until the first "power off", and then remains the same. It's almost as if the cloning operation remains half-finished until the clone is powered off. Perhaps this is normal behaviour.

The clones aren't going to be of much use if they all have the same MACs, so I don't think the fact that the UUID also remains the same should concern us.

errordeveloper:
@rade those finding are interesting. Personally, I'm not very surprised about VirtualBox doing what you have observed.

errordeveloper:
On GCE rebooting and power cycling persists the UUID. Cloning results in a new UUID.

errordeveloper:
On Azure rebooting and power cycling persists the UUID. I'm still looking for the clone button....

errordeveloper:
So in Azure you can only clone a "classic" VM, and there is no obvious way to do something similar in the new Resource Manager world. Cloned "classic" VM does get a new UUID, as expected.

bboreham:
I tested KVM: an Ubuntu guest automagically got a new `product_uuid` which survived a reboot, and cloning the guest got a different `product_uuid`.

An Alpine guest under KVM got no `product_uuid` unless the `-uuid` flag was used on the command-line to start the VM.

So I would say this is working as desired under KVM.

DanielDent:
Having restarts work reliably for us is very important. Being able to reclaim IPs is as well. It's not necessarily equally important for all of our machines & all of our services, but there are cases where it very much matters to us.

It looks to me like the only issue is having the host peer ID be the same each time weave runs.  I do appreciate the attempts to make weave autoconfigure in most cases. But in some environments autoconfig isn't going to be reliable enough, and one of the cool things with weave is being able to deploy across a heterogeneous environment.

When launching weave, it would be great if we could simply specify the peer ID. This would also provide administrators with an easy way to bring up a replacement machine after a node has permanently failed.

rade:
> When launching weave, it would be great if we could simply specify the peer ID.

You can already do that. `weave launch --name d2:e1:4d:cc:92:15`.

DanielDent:
Thanks @rade , that's awesome. It wasn't clear to me scanning through the weave script or looking at it's help output.

DanielDent:
@rade When launching with --name, are there any restrictions on the name? Will any (psuedo-)random mac-like string work, or are there other considerations? In random_mac(), care is taken so that the multicast bit is clear and locally administered bit is set. Can that be ignored, or should that be taken into consideration?

DanielDent:
For what it's worth, here is the process which ends up running in the container: "/home/weave/weaver --port 6783 --name RANDOM_MAC_HERE --nickname HOSTNAME_HERE --datapath datapath --ipalloc-range IPALLOC_RANGE_HERE--dns-effective-listen-address 172.17.0.1 --dns-listen-address 172.17.0.1:53 --http-addr 127.0.0.1:6784 --name NAME_I_SPECIFIED_HERE --ipalloc-default-subnet DEFAULT_SUBNET_HERE REMOTE_PEERS_HERE"

It seems to come up with the correct peer ID (NAME_I_SPECIFIED_HERE), but I note that the --name flag is being passed twice.

rade:
>  Will any (pseudo-)random mac-like string work?

Yes.

> the `--name` flag is being passed twice.

That's fine. Last one wins, which is why we don't need any special code in the script for this and a few other options (e.g. `--nickname`).

errordeveloper:
With 1.4.4 I have noticed this after rebooting some of my instances in EC2:
```
root@ip-172-20-0-93:~# weave status ipam
9e:32:45:f9:af:d8(ip-172-20-0-93)        32768 IPs (03.1% of total) 
e6:19:24:41:11:36(ip-172-20-0-92)        65535 IPs (06.2% of total) 
82:c8:e0:d9:db:87(ip-172-20-0-76)        65537 IPs (06.3% of total) 
0e:b5:e1:a7:63:ad(ip-172-20-0-91)        65535 IPs (06.2% of total) 
4e:09:9f:f2:81:4e(ip-172-20-0-223)       65535 IPs (06.2% of total) 
3a:ba:ee:58:8c:35(ip-172-20-0-223)      131074 IPs (12.5% of total) - unreachable!
a2:43:08:c6:20:cb(ip-172-20-0-91)        65535 IPs (06.2% of total) - unreachable!
9e:f3:0c:4c:8a:79(ip-172-20-0-167)      131072 IPs (12.5% of total) - unreachable!
e2:3f:01:58:21:11(ip-172-20-0-166)      131072 IPs (12.5% of total) - unreachable!
ca:0b:2f:c1:6f:94(ip-172-20-0-168)      131072 IPs (12.5% of total) - unreachable!
f2:69:4f:49:49:46(ip-172-20-0-152)       65538 IPs (06.3% of total) 
de:99:02:c8:9d:d0(ip-172-20-0-223)       32767 IPs (03.1% of total) - unreachable!
22:82:0f:f7:cd:f4(ip-172-20-0-88)        65536 IPs (06.2% of total) 
```
Am I correct in assuming that this change should fix it?

Issue #1871, opened 2016-01-11
conflict subnet range when 'weave launch'


novice79:
I try weave on Aliyun ubuntu14.04 server, when exec 'weave launch' I got:
Unable to find image 'weaveworks/weave:1.4.1' locally
1.4.1: Pulling from weaveworks/weave
9169fcce814f: Pulling fs layer
125d612ff28d: Pulling fs layer
6cb65287d390: Pulling fs layer
8b05dbda47f6: Pulling fs layer
ba2ba7a02d38: Pulling fs layer
9169fcce814f: Verifying Checksum
9169fcce814f: Download complete
9169fcce814f: Pull complete
ba2ba7a02d38: Verifying Checksum
ba2ba7a02d38: Download complete
6cb65287d390: Verifying Checksum
6cb65287d390: Download complete
8b05dbda47f6: Verifying Checksum
8b05dbda47f6: Download complete
125d612ff28d: Verifying Checksum
125d612ff28d: Download complete
125d612ff28d: Pull complete
6cb65287d390: Pull complete
8b05dbda47f6: Pull complete
ba2ba7a02d38: Pull complete
Digest: sha256:1d16f94e97c3b21cf9ceeb42c2acbc8201c680ceb244f3429d82cde68c0e4be3
Status: Downloaded newer image for weaveworks/weave:1.4.1
Network 10.32.0.0/12 overlaps with existing route 10.0.0.0/8 on host.
ERROR: Default --ipalloc-range 10.32.0.0/12 overlaps with existing route on host.
You must pick another range and set it on all hosts.
root@iZ94txhw7wwZ:~# ifconfig
datapath  Link encap:Ethernet  HWaddr 22:7e:4e:8b:9c:c6  
          UP BROADCAST RUNNING MULTICAST  MTU:1410  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

docker0   Link encap:Ethernet  HWaddr 02:42:e1:21:f7:a8  
          inet addr:192.168.0.1  Bcast:0.0.0.0  Mask:255.255.240.0
          UP BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:1924 errors:0 dropped:0 overruns:0 frame:0
[...]

PR #1872, opened 2016-01-11
Documentation on IP addresses, routes and networks

bboreham:
To provide some more explanation and guidance for people who are not familiar with the terminology.

Inspired by #1871

### DIFF
--- a/site/features.md
+++ b/site/features.md
@@ -115,7 +115,7 @@ Weave detects when a container has exited and releases its
 automatically allocated addresses so they can be re-used.
 
 See the [Automatic IP Address Management](ipam.html) documentation for
-further details.
+further details. We also have a page on [the basics of IP addressing](ip-addresses.html)
 
 Instead of getting weave to allocate IP addresses automatically, it is
 also possible to specify an address and network explicitly, expressed
--- a/site/features.md
+++ b/site/features.md
@@ -115,7 +115,8 @@ Weave detects when a container has exited and releases its
 automatically allocated addresses so they can be re-used.
 
 See the [Automatic IP Address Management](ipam.html) documentation for
-further details. We also have a page on [the basics of IP addressing](ip-addresses.html)
+further details. We also have an explanation of
+[the basics of IP addressing](ip-addresses.html)
 
 Instead of getting weave to allocate IP addresses automatically, it is
 also possible to specify an address and network explicitly, expressed
--- a/site/ip-addresses.md
+++ b/site/ip-addresses.md
@@ -64,8 +64,8 @@ chances of overlap. But it could be that you or your hosting provider
 are using some of these private addresses in the same range, which would
 cause a clash.
 
-Here's an example that came up recently: on `weave launch`, the
-following error message appeared:
+Here's an example: on `weave launch`, the following error message
+can appear:
 
 ````
 Network 10.32.0.0/12 overlaps with existing route 10.0.0.0/8 on host.
--- a/site/ip-addresses.md
+++ b/site/ip-addresses.md
@@ -1,0 +1,88 @@
+---
+title: IP Addresses, routes and networks
+layout: default
+---
+
+# IP Addresses, routes and networks
+
+Weave Net lets you run containers on a private network, and so the IP
+addresses those containers use are insulated from the rest of the
+Internet, and you don't have to worry about them clashing. Except, if
+they actually do clash with some addresses that you'd like the
+containers to talk to.
+
+Some definitions:
+
+- _IP_ is the Internet Protocol, the fundamental basis of network
+   communication between billions of connected devices.
+- The _IP address_ is (for most purposes) the four numbers separated
+  by dots, like `192.168.48.12`. Each number is one byte in size, so can
+  be between 0 and 255.
+- Each IP address lives on a _Network_, which is some set of those
+  addresses that all know how talk to each other. The network address
+  is some prefix of the IP address, like `192.168.48`. To indicate
+  which part of the address indicates the network, we append a slash
+  and then the number of bits in the network prefix, like
+  `/24`. Another example would be `10.4.2.6/8`, where the network
+  prefix is the first 8 bits - `10`. Written out in full, the address
+  of this network is `10.0.0.0/8`. The most common prefix lengths are
+  8, 16 and 24, but there is nothing stopping you using a /9 network
+  or a /26.
+- A _route_ is an instruction for how to deal with traffic destined
+  for somewhere else - it specifies a Network, and a way to talk to
+  that network.  Every device using IP has a table of routes, so for
+  any destination address it looks up that table, finds the right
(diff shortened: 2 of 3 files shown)


site/features.md
```diff
@@ -115,7 +115,7 @@ Weave detects when a container has exited and releases its
 automatically allocated addresses so they can be re-used.
 
 See the [Automatic IP Address Management](ipam.html) documentation for
-further details.
+further details. We also have a page on [the basics of IP addressing](ip-addresses.html)
```
rade: s/a page/an explanation

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+find a match, so uses the default rule.
+
+The default configurations for both weave Net and Docker use [Private
+Networks](https://en.wikipedia.org/wiki/Private_network), whose
+addresses are never used on the public internet, so that reduces the
+chances of overlap. But it could be that you or your hosting provider
+are using some of these private addresses in the same range, which would
+cause a clash.
+
+Here's an example that came up recently: on `weave launch`, the
```
rade: s/that came up recently/

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+10.47. But the user's host already has a route for `10.0.0.0/8`,
+which overlaps, because the first 8 bits are the same. If we went
+ahead and used the default network, then for an address like
+`10.32.5.6` the kernel would never be sure whether this meant the
+weave Net network of `10.32.0.0/12` or the hosting network of
+10.0.0.0/8.
+
+If you're in this situation, the best option is, as the message says,
+to pick a different range. In this example, a different private range
+such as `172.30/16` would work.
```
rade: `172.30/16` is rejected by our CIDR validator: `FATA: 2016/01/12 00:33:52.436450 invalid CIDR address: 172.30/16`. `172.30.0.0/16` works.

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+  by dots, like `192.168.48.12`. Each number is one byte in size, so can
+  be between 0 and 255.
+- Each IP address lives on a _Network_, which is some set of those
+  addresses that all know how talk to each other. The network address
+  is some prefix of the IP address, like `192.168.48`. To indicate
+  which part of the address indicates the network, we append a slash
+  and then the number of bits in the network prefix, like
+  `/24`. Another example would be `10.4.2.6/8`, where the network
+  prefix is the first 8 bits - `10`. Written out in full, the address
+  of this network is `10.0.0.0/8`. The most common prefix lengths are
```
rade: I read "this network" as referring to `10.4.2.6/8`, which is of course not a network. Replacing "the address of this network" with "that network address" would probably avoid that confusion.

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+````
+Network 10.32.0.0/12 overlaps with existing route 10.0.0.0/8 on host.
+ERROR: Default --ipalloc-range 10.32.0.0/12 overlaps with existing route on host.
+You must pick another range and set it on all hosts.
+````
+
+As the message says, the default that weave Net would like to use is
+`10.32.0.0/12` - a 12-bit prefix so all addresses starting with the bit
+pattern 000010100010, or in decimal everything from 10.32 through
+10.47. But the user's host already has a route for `10.0.0.0/8`,
```
rade: I would perhaps say "everything from (and including) 10.32.0.0 to (but excluding) 10.47.0.0".

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+which overlaps, because the first 8 bits are the same. If we went
+ahead and used the default network, then for an address like
+`10.32.5.6` the kernel would never be sure whether this meant the
+weave Net network of `10.32.0.0/12` or the hosting network of
+10.0.0.0/8.
```
rade: single quotes missing.

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+Net will take care of routing the traffic. To talk an external server
+at address `74.125.133.128`, it looks in its routing table, doesn't
+find a match, so uses the default rule.
+
+The default configurations for both weave Net and Docker use [Private
+Networks](https://en.wikipedia.org/wiki/Private_network), whose
+addresses are never used on the public internet, so that reduces the
+chances of overlap. But it could be that you or your hosting provider
+are using some of these private addresses in the same range, which would
+cause a clash.
```
rade: To much "use" in that last paragraph.

site/ipam.md
```diff
@@ -109,7 +109,7 @@ The range parameter is written in
 [CIDR notation](http://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) -
 in this example "/16" means the first 16 bits of the address form the
 network address and the allocator is to allocate container addresses
-that all start 10.2.
+that all start 10.2. [More information on IP addresses and routes](ip-addresses.html)
```
rade: Try turning that into a sentence.

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+`10.32.0.0/12` - a 12-bit prefix so all addresses starting with the bit
+pattern 000010100010, or in decimal everything from 10.32 through
+10.47. But the user's host already has a route for `10.0.0.0/8`,
+which overlaps, because the first 8 bits are the same. If we went
+ahead and used the default network, then for an address like
+`10.32.5.6` the kernel would never be sure whether this meant the
+weave Net network of `10.32.0.0/12` or the hosting network of
+10.0.0.0/8.
+
+If you're in this situation, the best option is, as the message says,
```
rade: "you are"

site/ip-addresses.md
```diff
@@ -0,0 +1,88 @@
...
+- Each IP address lives on a _Network_, which is some set of those
+  addresses that all know how talk to each other. The network address
+  is some prefix of the IP address, like `192.168.48`. To indicate
+  which part of the address indicates the network, we append a slash
+  and then the number of bits in the network prefix, like
+  `/24`. Another example would be `10.4.2.6/8`, where the network
+  prefix is the first 8 bits - `10`. Written out in full, the address
+  of this network is `10.0.0.0/8`. The most common prefix lengths are
+  8, 16 and 24, but there is nothing stopping you using a /9 network
+  or a /26.
```
rade: You may want to give an example here, rather than below. I reckon it's those non-byte-aligned network prefixes that really confuse people.

We may also want to link to one of those CIDR calculators, such as http://www.ipaddressguide.com/cidr (that one can even be embedded).

site/ip-addresses.md
bboreham: it does include 10.47.0.0; it is everything up to 10.47.255.255

bboreham:
I amended all the parts you commented on @rade

PR #1878, opened 2016-01-11

### LINKED ISSUE #1873: Improve visibility of plugin startup problems

````
$ weave launch-plugin --foo
````

````
$ weave status
[...]
        Service: plugin
     DriverName: weave
````

Compare:

````
$ weave launch-router --foo
````

bboreham:
Fixes #1873 


bboreham:
Replaced by #1879 on 1.4 branch

PR #1879, opened 2016-01-11

### LINKED ISSUE #1873: Improve visibility of plugin startup problems

bboreham:
Fixes #1873

### DIFF
--- a/prog/plugin/main.go
+++ b/prog/plugin/main.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"strings"
@@ -100,6 +101,7 @@ func run(dockerClient *docker.Client, address, meshAddress, meshNetworkName, nam
 		}
 	}
 
+	go listenAndServeStatus("/home/weave/status.sock")
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, os.Kill, syscall.SIGTERM)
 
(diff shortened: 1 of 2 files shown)


bboreham:
Test 640 failed because the container is detected as 'restarting' and `container_id` barfs.  This means there was always a race here, whether the restarting container is actually in a state to have a network attached.

That race is now covered by #1561. But not on this branch.

rade:
`State.Status` is new in Docker 1.9. We get away with using it in the current `check_not_running` through sheer luck. Not so in this new code.

bboreham:
Added `wait_for_status` to `launch_plugin`, and rebased

prog/plugin/main.go
```diff
@@ -168,3 +170,19 @@ func createNetwork(dockerClient *docker.Client, networkName, address string) err
...
+	if err != nil {
+		Log.Fatalf("ListenAndServeStatus failed: %s", err)
+	}
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "ok")
+	})
+	if err := (&http.Server{Handler: handler}).Serve(listener); err != nil {
+		Log.Fatalf("ListenAndServeStatus failed: %s", err)
+	}
+}
```
rade: this is c&p from the proxy. And the socket removal+listening bit is also duplicated in the plugin. And we really don't want those `Fatalf`, at least not in the first part.

rade:
This is getting a bit too scary for 1.4, IMO.

bboreham:
Replaced by #1941 on master

PR #1883, opened 2016-01-12
Fix race condition in weavewait

### LINKED ISSUE #1882: race condition in weavewait
`EnsureInterfaceAndMcastRoute` does:

- check for existing interface
- subscribe for new interfaces and check each one
- check for existing multicast route
- subscribe for new routes and check each one

The subscribe should be done first in both cases, to avoid a race-condition where the thing appears in between checking for it and listening for new ones.

bboreham:
Fixes #1882


bboreham:
Wrong branch

PR #1885, opened 2016-01-12
Allow an existing plugin to keep running on `weave launch` if it has the right arguments

### LINKED ISSUE #1869: plugin prevents successful `weave launch` on reboot
The plugin is started with a restart policy of `always`. Consequently, if you have a startup script that runs `weave launch`, such as our published [systemd](http://docs.weave.works/weave/latest_release/systemd.html), it will fail the 2nd time round, since `weave launch` will report `weaveplugin is already running; you can stop it with 'weave stop-plugin'.` and exit with status 1.

bboreham:
Fixes #1869 

This is heavily special-cased for the plugin; much more mechanism needed to generalise it to all components, but we don't start them with a retry policy so the drive isn't there.

I also added `WEAVEPLUGIN_DOCKER_ARGS`, which came up in the discussion.

### DIFF
--- a/weave
+++ b/weave
@@ -1701,11 +1701,15 @@ stop_proxy() {
 }
 
 launch_plugin() {
+    # Set WEAVEPLUGIN_DOCKER_ARGS in the environment in order to supply
+    # additional parameters, such as resource limits, to docker
+    # when launching the weaveplugin container.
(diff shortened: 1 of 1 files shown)


bboreham:
Closed in favour of #1886

PR #1886, opened 2016-01-12
Allow plugin to keep running on 'weave launch'

### LINKED ISSUE #1869: plugin prevents successful `weave launch` on reboot

bboreham:
Fixes #1869, in a simpler way than #1885.

### DIFF
--- a/weave
+++ b/weave
@@ -1689,7 +1689,14 @@ stop_proxy() {
     stop $PROXY_CONTAINER_NAME "Proxy"
 }
 
-launch_plugin() {
+launch_plugin_if_not_running() {
+    retval=0
+    check_not_running plugin $PLUGIN_CONTAINER_NAME $BASE_PLUGIN_IMAGE || retval=$?
+    # If an existing plugin is running (we start it with restart=always), return its ID
+    [ $retval = 1 ] && PLUGIN_CONTAINER=$(container_id $PLUGIN_CONTAINER_NAME) && return 0
+    # Any other kind of error code from check_not_running is a failure.
+    [ $retval -gt 0 ] && return $retval
+
     PLUGIN_CONTAINER=$(docker run --privileged -d --name=$PLUGIN_CONTAINER_NAME \
         --restart=always \
         --net=host \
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -1846,11 +1846,17 @@ EOF
...
         launch_proxy  $COMMON_ARGS
-        plugin_disabled || launch_plugin $COMMON_ARGS
+        # Allow an existing plugin to keep running, because we run it with restart=always
+        if ! plugin_disabled ; then
+            if check_not_running plugin $PLUGIN_CONTAINER_NAME $BASE_PLUGIN_IMAGE ; then
+                launch_plugin $COMMON_ARGS
+            else
+                [ $? -lt 2 ] || exit 1
+            fi
+        fi
```
rade: You could turn this into just
```
plugin_disabled || ! check_not_running ... || launch_plugin ...
```
by making `check_not_running` exit (as it used to) in the cases where you currently return 2 or 3.

Whether that is better is debatable.

awh: In the general case, I prefer for these kind of functions to `return` rather than `exit` as it makes them more flexible...

bboreham: (a) you get a very long line and (b) I prefer to avoid side-effects.

bboreham:
I should change `weave launch-plugin` to also return 0 on already-running.

weave
rade: All true. But you have to admit that the return code checking is gross.

bboreham:
I fixed weave launch-plugin to also return 0 on already-running, which gave the whole thing a different flavour.

rade:
Better. But I worry that some fool (i.e. me) will re-add the "missing" `check_not_running`.

PR #1888, opened 2016-01-13
refactor overlay creation

rade:
There is one behavioural change in this: the routerName, if not specified with --name, is now derived from the datapath network interface MAC (for --datapath) or the pcap interface MAC (for --iface). Previously it was only derived from the latter. So this removes an awkward asymmetry.

This doesn't actually make any difference to weave usage, since the script always supplies a --name.


awh:
Looks good. One very minor comment to be addressed at your discretion, ready to merge either way.

Issue #1889, opened 2016-01-13
Use weaveworks/mesh instead of weaveworks/weave/mesh


peterbourgon:
We are extracting mesh from weave, and so will refactor weave to use a different import path for mesh.

PR #1890, opened 2016-01-13
Change import location for mesh

peterbourgon:
Mesh has moved to github.com/weaveworks/mesh, so we update the import location and remove the old code. Addresses #1889.


awh:
Looks good, but needs rebase and the relevant submodule added (we removed the `go get` from the makefile, so this does not build when merged to master)

peterbourgon:
Yes, I'm sorry I didn't make the WIP prefix; I know there's still some work to do here. Thanks!

awh:
Looks good, but I'd like to double check that nothing has changed in `mesh/` since you created https://github.com/weaveworks/mesh before I merge it. Do you happen to know which weave repo commit you copied from?

awh:
> Do you happen to know which weave repo commit you copied from?

Never mind, `diff` to the rescue :)

PR #1895, opened 2016-01-14
Move gce.sh & related scripts to tools.

### LINKED ISSUE #1286: Switch to tools.git

tomwilkie:
Fixes #1286

I've already moved gce.sh to the build-tools repo such that Scope no longer needs to checkout weave to run the integration tests. This PR removed the scripts from the weave repo and has it use the shared ones in tools, and hopefully is the last PR in the set.


awh:
There's a relative path issue with loading the SSH key required to access the vagrant VMs:
```
~/weave$ make run-smoketests
cd test && ./setup.sh && ./run_all.sh
Copying weave images, scripts, and certificates to hosts, and
  prefetch test images
Warning: Identity file /home/vagrant/weave/tools/integration/insecure_private_key not accessible: No such file or directory.
Warning: Identity file /home/vagrant/weave/tools/integration/insecure_private_key not accessible: No such file or directory.
vagrant@192.168.48.12's password: vagrant@192.168.48.11's password: Warning: Identity file /home/vagrant/weave/tools/integration/insecure_private_key not accessible: No such file or directory.
vagrant@192.168.48.13's password:
```
Previously `insecure_private_key` was sourced from `weave/test`.

tomwilkie:
@awh would you mind testing again?  I can't run the tests locally on my mac (not enough memory for 2 vms).

awh:
It's still broken - `$SSH` in `tools/integration/config.sh` uses `$SSH_DIR` initialised from `$DIR` at the top of that file. I can fix that by swapping the order of `SSH_DIR` initialisation and the sourcing of `tools/integration/config.sh` in `test/config.sh`, but then there are further problems with the value of `$WEAVE` (firstly it is being set erroneously at the bottom of `tools/integration/config.sh`; if I remove that statement, the value being set at the bottom of `test/config.sh` is wrong because `$DIR` has been overwritten)

rade:
@tomwilkie did you forget to re-assign this? Looks like you made a change following @awh's last comment.

A few comments from me on config.sh...

- `upload_executable()` and `docker_api_on()` are generic and hence should be in the tools config.sh
- it's a bit strange that `start_suite()` is in the tools config.sh and `end_suite()` remains here. Moving the latter would drag across a bunch of other stuff, which is undesirable. So perhaps leave the former (and, I'm guessing, `weave_on()`) here, and accept a little bit of code duplication.

rade:
@tomwilkie also, apologies for not getting to this earlier. We should not be sitting on PRs for two months.

bboreham:
It looks like https://github.com/weaveworks/build-tools/commit/b341fd5f5543210a995fd29731b1fe76778d97be needs to be merged into the tools repo before this can be merged into weave.

bboreham:
Also #2131 made a further change to `weave_on` that would need to be merged to `tools`

PR #1901, opened 2016-01-15
move plugin network creation/removal into weaveutil

### LINKED ISSUE #1897: Move network create/remove code into weaveutil
Following on from #1847, there is code in the plugin that creates and removes a docker network; this could move out to the weaveutil executable.

rade:
The main gotcha is that we lose the ability to specify a different mesh network name, or disable it altogether. That was always poorly supported and IMO is not worth losing sleep over until we / our users have a real need for it.

Closes #1897.

### DIFF
--- a/prog/plugin/main.go
+++ b/prog/plugin/main.go
@@ -5,9 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"os/signal"
 	"strings"
-	"syscall"
 
 	"github.com/docker/libnetwork/ipamapi"
 	go_docker "github.com/fsouza/go-dockerclient"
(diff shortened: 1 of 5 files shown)


rade:
btw, `weave stop` time has dropped from 1.5s to 1.1s for me with this change.

bboreham:
> we run the plugin with restart=always, so it will get restarted

Nitpicking: it does not get restarted immediately if stopped by the user, e.g. `docker stop`. It does get restarted on the next reboot.

bboreham:
> CreateNetwork does sometimes(always?) *not* return ErrNetworkAlreadyExists

Tracked at https://github.com/docker/docker/issues/18363

PR #1902, opened 2016-01-15
Document vendored dependency management commands

awh:
Will follow up with an issue to automate update + prune.

### DIFF
--- a/vendor/README.md
+++ b/vendor/README.md
@@ -1,0 +1,26 @@
+
+# Managing Vendored Dependencies with `git submodule`
+
+These operations result in staged but uncommitted changes to your
+branch; you will need to commit them as normal. Execute them in the
+root of your checkout.
+
+## Adding a New Dependency
+
+    ~/weave$ git submodule add https://example.com/organisation/module vendor/example.com/organisation/module
+
+## Update All Dependencies to Upstream `master`
+
+    ~/weave$ git submodule foreach git pull origin master
+
+## Force a Specific Dependency to a Particular Commit
+
+	~/weave$ (cd vendor/example.com/organisation/module && git checkout <commit-ish>)
+	~/weave$ git add vendor/example.com/organisation/module
+
+## Remove a Dependency
+
+	~/weave$ git submodule deinit vendor/example.com/organisation/module
+    ~/weave$ git rm vendor/example.com/organisation/module
+	~/weave$ rm -rf .git/modules/vendor/example.con/organisation/module
+


tomwilkie:
Some bits of these commands might be useful to include in a readme:

```bash
go list -f '{{join .Deps "\n"}}' ./prog/* | grep -E '^(github.com|k8s.io)' | grep -v 'github.com/weaveworks/weave' | sort -u > deps
cd vendor
cat ../deps | while read repo; do git submodule add -f --depth 1 https://$repo.git $repo; done
```

awh:
(This is more relevant to #1903, but you posted the script snippet here)

> Some bits of these commands might be useful to include

Many of the paths we import are packages nested inside git repositories, and cannot themselves be added as submodules - for example the above scripts produce

```
github.com/docker/libnetwork/drivers/remote/api
github.com/docker/libnetwork/ipamapi
github.com/docker/libnetwork/ipams/remote/api
github.com/docker/libnetwork/netlabel
github.com/docker/libnetwork/types
```

whereas what we need to add as the submodule is `github.com/docker/libnetwork`. Coalescing these automatically is going to involve probing successively longer path prefixes until we hit a URL that is valid `.git` repository root :frowning:

rade:
> Coalescing these automatically is going to involve probing successively longer path prefixes until we hit a URL that is valid `.git` repository roo

In case of github, we know that the repositories all live at `github.com/<org>/<project>`

dpw:
I have an imminent solution for this issue, which I hope will be comprehensive.

dpw:
Not quite finished, and it needs a readme, but it's in a useable and useful state:

https://github.com/dpw/vendetta

Without options it will add submodules to satisfy dependencies.  With the `-u` option it will also update submodules used to satisfy dependencies.  And with the `-p` option it will also prune unused submodules under `vendor/`.

I've tested it on the Net and flux repos.  On the Net repo, it will add dependencies to satisfy the `tools/` submodule, which are not currently present.  That might be the right thing to do (the `github.com/armon/go-socks5` dependency is dubious, but then the presence of the `socks` tool in that `tools` repo seems dubious to me).  Another solution would be to add the necessary vendor submodules into `tools`.

awh:
I'm a bit confused by this development - `gvt` was successfully argued against in favour of submodules as the former constituted an unwanted dependency, and now we are using submodules *and* have created our own tool that provides the dependency update function of `gvt`?

dpw:
> I'm a bit confused by this development - gvt was successfully argued against in favour of submodules as the former constituted an unwanted dependency, and now we are using submodules and have created our own tool that provides the dependency update function of gvt?

With gvt, you need to use gvt to retrieve the dependencies.  So you need it to build or do any development.

Vendetta has a more limited role.  Fetching dependencies can be done with `git submodule update --init --recursive`, so you rarely need to use vendetta. I don't anticipate any explicit encouragement to use it in our repos.

But questions like "how do I populate vendor/ in the first place?", "how do I update the submodules under vendor/ ?", and "how do I prune submodules that are no longer needed?" are reasonable ones.  There does not seem to be an existing tool that addresses just that need, and while you can get some of the way with shell scripts, I think they are bound to have shortcomings.  Hence vendetta.

rade:
I think you've just written the README for vendetta :)

dpw:
Vendetta now has a README: https://github.com/dpw/vendetta

PR #1915, opened 2016-01-20
Support subnet and iprange arguments on network create

### LINKED ISSUE #1806: Support --subnet and --ip-range options to `docker network create`
These options are intended to specify which addresses to use for containers.  When using the weave IPAM driver, they are passed to the `RequestPool` entrypoint as `pool` and `subPool` arguments, but these arguments are currently ignored.

bboreham:
Fixes #1806 

It is perhaps cheating to get Docker to store the info we need, smuggled inside the pool name.  But I don't see any reason why it should break.

### DIFF
--- a/plugin/ipam/driver.go
+++ b/plugin/ipam/driver.go
@@ -1,7 +1,9 @@
 package ipamplugin
 
 import (
+	"fmt"
 	"net"
+	"strings"
 
 	"github.com/docker/libnetwork/ipamapi"
 	"github.com/docker/libnetwork/netlabel"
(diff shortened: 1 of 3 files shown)


plugin/ipam/driver.go
```diff
@@ -44,10 +61,33 @@ func (i *ipam) ReleasePool(poolID string) error {
...
+			return
+		}
+		// We are lying slightly to IPAM here: the range is not a subnet
+		if ip, err = i.weave.AllocateIPInSubnet("_", iprange); err != nil {
+			return
+		}
+		ip.Mask = subnet.Mask // fix up the subnet we lied about
+	} else {
+		err = fmt.Errorf("Unrecognized pool ID: %s", poolID)
+	}
```
rade: cosmetic: I'd tried to get that big `else if` block outdented.

It also troubles me that some error returns are logged under debug and some are not. You could ensure that all get logged by moving the return logging into a `defer` at the beginning. And you may want to do the same in `RequestPool`. Or ditch the debug logging of returns.

plugin/ipam/driver.go
```diff
@@ -44,10 +61,33 @@ func (i *ipam) ReleasePool(poolID string) error {
...
+	// If we pass magic string "_" to weave IPAM it stores the address under its own string
+	if poolID == "weavepool" { // old-style
+		ip, err = i.weave.AllocateIP("_")
+	} else if strings.HasPrefix(poolID, "weave-") {
+		dashPos := strings.Index(poolID[6:], "-")
+		if dashPos == -1 {
+			return nil, nil, fmt.Errorf("Unrecognized pool ID: %s", poolID)
+		}
+		subnetStr := poolID[6 : dashPos+6]
+		iprangeStr := poolID[dashPos+7:]
```
rade: using `strings.Split(...)` would make all the above a bit less fiddly, allowing you to get rid of all the indexing and even the `HasPrefix`, i.e. you'd simply split `poolID` on "-", check you get three elements back and that the first is "weave".

bboreham:
Added documentation and squashed commits

site/plugin.md
```diff
@@ -33,32 +33,48 @@ The Weave plugin actually provides *two* network drivers to Docker
...
+[default subnet](ipam.html#subnets) set for the `weave` router.
+
+To create additional networks using the `weavemesh` driver, pick a
+different subnet and run this command on all hosts:
 
-If you do create additional networks using the `weavemesh` driver,
-containers attached to them will be able to communicate with
-containers attached to `weave`; there is no isolation between those
-networks.
+    $ docker network create --driver=weavemesh --ipam-driver=weavemesh --subnet=<subnet> <network-name>
```
berlic: Isn't it a good idea to mention `--ip-range` here?
I got IPs assigned from default subnet without explicit ip-range setting.

rade: @berlic works fine for me. Note that this PR has not made it into a release yet, so if you are running a released version of weave, such as 1.4.4, the above won't work.

berlic: @rade i've built it from master to try multiple wavemesh-networks.

```
root@node:~# weave version
weave script (unreleased version)
weave router git-c5361801c95e
weave proxy  git-c5361801c95e
weave plugin git-c5361801c95e

root@node:~# docker version
Client:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 18:36:33 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 18:36:33 2016
 OS/Arch:      linux/amd64
```

Tried to reproduce it on fresh VMs – no luck: works without explicit ip-range setting, as expected.
May be because of updated docker version (1.10.2) on my fresh install.
Sorry for disturbance.

PR #1916, opened 2016-01-20
Support requesting a specific IP from the plugin

### LINKED ISSUE #1734: Allow specification of specific IP addresses when using plugin
For users who want a static IP address.
Via the weave IPAM driver.
There is an `options` parameter passed in to the IPAM `RequestAddress` call, so presumably we can populate that somewhere in the Docker call.
Also will need to decide whether the request means "give me this address, even if you think it's already in use", "fail if I can't get this address" or "this is my preferred address but I'll take a different one if necessary".

bboreham:
Fixes #1734

### DIFF
--- a/plugin/ipam/driver.go
+++ b/plugin/ipam/driver.go
@@ -78,14 +78,21 @@ func (i *ipam) RequestAddress(poolID string, address net.IP, options map[string]
 	if _, subnet, err = net.ParseCIDR(parts[1]); err != nil {
 		return
 	}
-	if _, iprange, err = net.ParseCIDR(parts[2]); err != nil {
-		return
+	if address != nil { // try to claim specific address requested
+		if err = i.weave.ClaimIP("_", address); err != nil {
+			return
+		}
+		ip = &net.IPNet{IP: address, Mask: subnet.Mask}
+	} else {
+		if _, iprange, err = net.ParseCIDR(parts[2]); err != nil {
+			return
+		}
+		// We are lying slightly to IPAM here: the range is not a subnet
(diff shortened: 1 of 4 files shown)


bboreham:
Rebased

plugin/ipam/driver.go
```diff
@@ -81,11 +81,21 @@ func (i *ipam) RequestAddress(poolID string, address net.IP, options map[string]
...
 	}
-	// We are lying slightly to IPAM here: the range is not a subnet
-	if ip, err = i.weave.AllocateIPInSubnet("_", iprange); err != nil {
-		return
+	// See if we can give preferred address
+	if address != nil {
+		if err := i.weave.ClaimIP("_", address); err == nil {
+			ip = &net.IPNet{IP: address, Mask: subnet.Mask}
+		} else {
+			Log.Debugln("ClaimIP returned", err, "; allocating new address")
```
awh: Is this the right thing to do, or should we fail the `RequestAddress` operation?

bboreham: Well I thought it was, but looking at https://github.com/docker/libnetwork/blob/master/ipam/allocator.go#L492 it seems Docker's implementation does a fail.

It is documented as "preferred IP", so it seems strange to insist.

bboreham:
I clarified the intended meaning, and removed the word "preferred" from Docker's code.

But there is a new problem: requesting the same address twice does not fail.

Currently we can't easily do a smoke-test because the `--ip` flag only comes in with Docker 1.10

bboreham:
Fixed the problem that requesting the same address twice did not fail.

PR #1921, opened 2016-01-21
Make weave script scan veths with prefix 'ethwe'

### LINKED ISSUE #1745: `weave ps` does not show containers connnected by plug-in / double allocation of IPs and loss of DNS entries on restart
Because it's looking for an interface called `ethwe`, and Docker+plugin ends up with `ethwe0`, or possibly some other number if you attach the container to multiple networks.

bboreham:
Fixes #1745

Veths created via the plugin will have a name like 'ethwe0'; previously we only looked at the device named exactly 'ethwe'. 

Covers `weave ps` and reclaim of addresses for IPAM and WeaveDNS.

Code now bails out in `container_ip_addr_show` if the container is in the host netns, otherwise it takes a while pondering all the devices visible at top level.

### DIFF
--- a/weave
+++ b/weave
@@ -833,8 +833,12 @@ setup_router_iface_bridged_fastdp() {
     setup_router_iface_fastdp "$@"
 }
 
+container_in_host_ns() {
+    [ -h "$PROCFS/$CONTAINER_PID/ns/net" -a -h "/proc/self/ns/net" -a "$(readlink $PROCFS/$CONTAINER_PID/ns/net)" = "$(readlink /proc/self/ns/net)" ]
+}
+
 attach() {
-    if [ -h "$PROCFS/$CONTAINER_PID/ns/net" -a -h "/proc/self/ns/net" -a "$(readlink $PROCFS/$CONTAINER_PID/ns/net)" = "$(readlink /proc/self/ns/net)" ] ; then
+    if container_in_host_ns ; then
         echo "Container is running in the host network namespace, and therefore cannot be" >&2
         echo "connected to weave. Perhaps the container was started with --net=host." >&2
         return 1
--- a/weave
+++ b/weave
@@ -899,8 +899,8 @@ detach() {
     netnsenter ip link del $CONTAINER_IFNAME type veth
 }
 
-container_weave_addrs() {
-    netnsenter ip addr show dev $CONTAINER_IFNAME
+container_ip_addr_show() {
+    ! container_in_host_ns && netnsenter ip addr show
 }
 
 ######################################################################
@@ -1031,6 +1031,17 @@ wait_for_status() {
     done
 }
 
+# Parse the output of `ip addr show`, outputing weave addresses of kind $1
+addrs_from_ip_addr_show() {
+    while read first second rest ; do
+        if [ "$first" != "${first%:}" ] ; then
+            iname="${second%:}"
+        elif [ "$first" = "$1" -a \( "$iname" = "weave" -o "$iname" != "${iname#ethwe}" \) ] ; then
+            echo $second
+        fi
+    done
+}
+
 # Call $1 for all containers, passing container ID, all MACs and all IPs
 with_container_addresses() {
     COMMAND=$1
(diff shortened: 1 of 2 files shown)


rade:
We should be able to construct a smoketest for this.

i.e. something along the lines of
```
$ weave launch
$ CID=$(docker run --net=weave -h foo.weave.local -dti alpine /bin/sh)
$ weave ps
[check that there is an entry for the container ; remember the IP address]
$ weave stop-router; weave launch-router
$ weave run -ti alpine /bin/sh
[check that returned IP is different from the one remembered]
[check that there is a DNS entry for foo.weave.local]
```

weave
```diff
@@ -1027,6 +1031,17 @@ wait_for_status() {
...
+# Parse the output of `ip addr show`, outputing weave addresses of kind $1
+addrs_from_ip_addr_show() {
+    while read first second rest ; do
+        if [ "$first" != "${first%:}" ] ; then
+            iname="${second%:}"
+        elif [ "$first" = "$1" -a \( "$iname" = "weave" -o "$iname" != "${iname#ethwe}" \) ] ; then
+            echo $second
+        fi
+    done
+}
```
rade: `ip -4 -o addr show` would avoid the state. Alas we can't get the MAC that way.

bboreham: Thanks; I didn't know about `-o`.

We don't need the MAC for internal purposes; could split into one implementation for `weave ps` and one for reclaims etc.?

rade: > could split into one implementation for `weave ps` and one for reclaims etc

I don't think the code would end up any simpler than what you have now.

weave
```diff
@@ -1035,12 +1050,14 @@ with_container_addresses() {
...
             ADDRS_CMD="ip addr show dev $BRIDGE"
         else
-            ADDRS_CMD="with_container_netns $CONTAINER_ID container_weave_addrs"
+            ADDRS_CMD="with_container_netns $CONTAINER_ID container_ip_addr_show"
         fi
         if CONTAINER_ADDRS=$($ADDRS_CMD 2>/dev/null) ; then
-            CONTAINER_MAC=$(echo "$CONTAINER_ADDRS" | grep -o 'link/ether .*' | cut -d ' ' -f 2)
-            CONTAINER_IPS=$(echo "$CONTAINER_ADDRS" | grep -o 'inet .*'       | cut -d ' ' -f 2)
-            $COMMAND "$CONTAINER_ID" "$CONTAINER_MAC" "$CONTAINER_IPS"
+            CONTAINER_MAC=$(echo "$CONTAINER_ADDRS" | addrs_from_ip_addr_show 'link/ether')
```
rade: Hmm. What happens when a container is attached to multiple weave networks? Presumably it ends up with multiple ethwe* interfaces, no? If so then the above will return multiple MACs, which is not something that the `weave ps` output is designed for.

bboreham: I do not believe this is a regression.

bboreham: OK, it is a regression in the sense that the code was only prepared for one `ethwe`

bboreham: Note that you cannot attach a container to two Weave networks unless you disable Weave's multicast route creation.

bboreham: You get output like this:

````
$ weave ps
weave:expose 22:bd:66:ea:d4:e3
373d3f21edbd ee:eb:46:e2:7a:72 3e:12:be:44:c3:8a 10.38.3.2/16 10.32.0.1/12
````

It doesn't seem unreasonable.

File: test/800_plugin_cross_hosts_2_test.sh [hunk 3]
```diff
@@ -15,7 +15,20 @@ weave_on $HOST2 launch $HOST1
...
+weave_on $HOST1 stop-router
+weave_on $HOST1 launch-router
+
+start_container $HOST1 --name=c3
+
+assert "container_ip $HOST1 c1" "$c1ip"
+assert_raises "container_ip $HOST1 c3 | grep -v $c1ip"
+
+assert "weave_on $HOST1 dns-lookup $C1_NAME" "$c1ip"
+
```
rade: This should all go into a separate test suite - it has nothing to do with cross-host testing and does not require 2 hosts.

Issue #1179, opened 2015-07-17
Potential race between docker events and dns add http API on container restart


tomwilkie:
If docker events are delayed for any reason, and you stop then start a container, the container died event could come from docker after the http call to dns from the proxy, resulting in the entry getting tombstoned just after being created!

rade:
To reproduce, apply the following patch:
```patch
diff --git a/common/docker/client.go b/common/docker/client.go
index 1f2ee0d..ad801c4 100644
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -1,6 +1,8 @@
 package docker
 
 import (
+       "time"
+
        "github.com/fsouza/go-dockerclient"
        . "github.com/weaveworks/weave/common"
 )
@@ -43,6 +45,7 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
                        switch event.Status {
                        case "die":
                                id := event.ID
+                               time.Sleep(10 * time.Second)
                                ob.ContainerDied(id)
                        }
                }
```

and then run
```
$ weave launch
$ eval $(weave env)
$ docker run --name=foo -dti ubuntu
$ docker restart foo; weave ps; weave status dns; sleep 10; \
  docker run --name=bar -dti ubuntu; weave ps; weave status dns
foo
foo
weave:expose 62:bf:f5:df:7e:20
61bb8df955c6 aa:b2:3b:ff:0c:81 10.32.0.1/12
53b5beee6995 02:c7:b7:2e:a0:73
foo          10.32.0.1       61bb8df955c6 62:bf:f5:df:7e:20
c0eb7d505ce4bd8f8caf747706c0eaeb0c23dbc65cb9efa73d9074c11d70d3f5
weave:expose 62:bf:f5:df:7e:20
c0eb7d505ce4 1e:56:0a:97:08:58 10.32.0.1/12
61bb8df955c6 aa:b2:3b:ff:0c:81 10.32.0.1/12
53b5beee6995 02:c7:b7:2e:a0:73
bar          10.32.0.1       c0eb7d505ce4 62:bf:f5:df:7e:20
```
So after restarting foo everything initially looks normal, but ten seconds later, and after we start another container, we can see that foo's IP address has been allocated to the new container - so we now have two containers with the same IP - and the DNS entry for foo has disappeared.

bboreham:
We could ask Docker on every event whether the container is really dead.

Or, given #1191, we could extend that to revivify the entry on `ContainerStarted` callback.

[...]

PR #1922, opened 2016-01-22
In IPAM, resurrect apparently-dead containers on ContainerStarted event

bboreham:
Partial fix for #1179

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -263,7 +263,11 @@ func (alloc *Allocator) removeDeadContainers() {
 	}
 }
 
-func (alloc *Allocator) ContainerStarted(ident string) {}
+func (alloc *Allocator) ContainerStarted(ident string) {
+	alloc.actionChan <- func() {
+		delete(alloc.dead, ident) // delete is no-op if key not in map
+	}
+}
 
 // Delete (Sync) - release all IP addresses for container with given name
 func (alloc *Allocator) Delete(ident string) error {


rade:
> Partial

Please elaborate.

bboreham:
> only for IPAM, not DNS. It also creates a situation where a container that restarts quickly retains all its IPAM [records], but one that gets stopped and started the next day loses all additional records created via weave attach

rade:
There is something weird going on with this. With [the patch from the issue](https://github.com/weaveworks/weave/issues/1179#issuecomment-179820559), I get...

```
$ weave launch
$ eval $(weave env)
$ docker run --name=foo -dti ubuntu
eeb9a94dffce787d1a5685dbf333212342c39f066933a01b774f306f8478fc29
$ docker restart foo; weave ps; weave status dns; sleep 40; docker run --name=bar -dti ubuntu; weave ps; weave status dns
foo
weave:expose d2:e2:3d:cc:92:05
eeb9a94dffce 42:cf:7e:b5:ce:c3 10.32.0.1/12
foo          10.32.0.1       eeb9a94dffce d2:e2:3d:cc:92:05
23fe25e509c8426b7a25a4768215e3584c2932a21d83ce2fd49d9203437e4443
weave:expose d2:e2:3d:cc:92:05
23fe25e509c8 ae:e0:a9:c7:e0:0c 10.32.0.2/12
eeb9a94dffce 42:cf:7e:b5:ce:c3 10.32.0.1/12
bar          10.32.0.2       23fe25e509c8 d2:e2:3d:cc:92:05
```
What happened to foo's DNS entry?

Note that the patch only delays the delivery of `ContainerDied`

bboreham:
The weird thing going on is that the transport of events via `fsouza/go-dockerclient` is done on separate goroutines for each listener (and our IPAM and DNS are two listeners), but the whole thing will lock up if the chan for one listener blocks.

And, it is nondeterministic whether it locks up before or after the DNS event is delivered.

PR #1928, opened 2016-01-26
Persist IPAM ring

### LINKED ISSUE #678: [ipam] don't leak or double-allocate IPs, on normal restart
ATM, when a peer restarts, one of the following scenarios occurs:

1. The peer got gracefully removed from the weave network on shutdown (e.g. `weave reset`), and starts with a new identity. Everybody is happy. No IPs leaked or double-allocated.
2. The peer didn't get removed gracefully, and lost its identity. The ranges it used to own are leaked and must be recovered with `rmpeer`.
3. The peer retained its identity on restart, gets told by other peers (via gossip) what IP ranges it used to owned, and happily lives ever after.
4. The peer retained its identity on restart, but the information it is told via gossip about the IP ranges it owns is out of date, e.g. if it had given away part of its range to some other peer just before dying and that information hasn't propagated through the entire network yet (either because of delays or partitions). As a result the peer may allocate duplicate IPs and/or eventually explode because it gets told more up-to-date, but conflicting, range ownership info.

We really want to eliminate (4). Persisting the ring is one way to accomplish that. This also helps with (2) since it would allow us to retain a peer's identity across most restarts, thus converting such restarts into case (3) or (4).

bboreham:
Fixes #678, except in the case the peer has changed identity (case 2).

Persistence is implemented using [BoltDB](https://github.com/boltdb/bolt/blob/master/README.md).

We use Gob to store the ring when we receive an update (could optimise to only when it changes?).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -55,3 +55,6 @@
 [submodule "vendor/github.com/vishvananda/netns"]
 	path = vendor/github.com/vishvananda/netns
 	url = https://github.com/vishvananda/netns.git
+[submodule "vendor/github.com/boltdb/bolt"]
+	path = vendor/github.com/boltdb/bolt
+	url = https://github.com/boltdb/bolt
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -30,6 +30,8 @@ const (
 )
 
 var (
+	topBucket  = []byte("top")
+	nameIdent  = []byte("peername")
 	ringBucket = []byte("ring")
 )
 
@@ -72,7 +74,7 @@ type Allocator struct {
 
 // NewAllocator creates and initialises a new Allocator
 func NewAllocator(ourName mesh.PeerName, ourUID mesh.PeerUID, ourNickname string, universe address.Range, quorum uint, dbPrefix string, isKnownPeer func(name mesh.PeerName) bool) (*Allocator, error) {
-	db, err := openDB(dbPrefix)
+	db, err := openDB(ourName, dbPrefix)
 	if err != nil {
 		return nil, err
 	}
@@ -90,14 +92,29 @@ func NewAllocator(ourName mesh.PeerName, ourUID mesh.PeerUID, ourNickname string
 	}, nil
 }
 
-func openDB(dbPrefix string) (*bolt.DB, error) {
+func openDB(ourName mesh.PeerName, dbPrefix string) (*bolt.DB, error) {
 	dbPathname := dbPrefix + "ipam.db"
 	db, err := bolt.Open(dbPathname, 0660, nil)
 	if err != nil {
 		return nil, fmt.Errorf("Unable to open %s: %s", dbPathname, err)
 	}
-	// Create all the buckets we are going to need
 	err = db.Update(func(tx *bolt.Tx) error {
+		nameVal := []byte(ourName.String())
+		// top-level bucket has peerName
+		if top := tx.Bucket(topBucket); top == nil {
+			top, err := tx.CreateBucket(topBucket)
+			if err != nil {
+				return err
+			}
+			top.Put(nameIdent, nameVal)
+		} else {
+			if checkPeerName := top.Get(nameIdent); !bytes.Equal(checkPeerName, nameVal) {
+				common.Log.Infof("[allocator] Deleting persisted data for peername %s", checkPeerName)
+				tx.DeleteBucket(ringBucket)
+				top.Put(nameIdent, nameVal)
+			}
+		}
+		// Create all the buckets we are going to need
 		_, err := tx.CreateBucketIfNotExists(ringBucket)
 		return err
 	})
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -30,6 +30,7 @@ const (
 )
 
 var (
+	ringBucket = []byte("ring")
 )
 
 // operation represents something which Allocator wants to do, but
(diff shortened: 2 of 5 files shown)


bboreham:
Status prior to this PR:

item | `weave restart` | `docker restart` | `docker stop` followed by `docker start` a good while later | policy restart after reboot
------|----------------------|----------------------|------------------------------------------------------------------------------|------
IPAM addresses specified with `WEAVE_CIDR`        | retained | retained | re-allocated; may change | re-allocated; may change 
non-IPAM addresses specified with `WEAVE_CIDR` | retained | retained | retained | retained
IPAM addresses specified in `weave attach`               | retained | leaked | leaked | lost
non-IPAM addresses specified in `weave attach`        | retained | lost | lost | lost

Status after this PR (at time of writing):

item | `weave restart` | `docker restart` | `docker stop` followed by `docker start` a good while later | policy restart after reboot
------|----------------------|----------------------|------------------------------------------------------------------------------|------
IPAM addresses specified with `WEAVE_CIDR`        | retained | retained | re-allocated; may change | retained 
non-IPAM addresses specified with `WEAVE_CIDR` | retained | retained | retained | retained
IPAM addresses specified in `weave attach`               | retained | lost | lost | lost
non-IPAM addresses specified in `weave attach`        | retained | lost | lost | lost

Final target of this PR:

item | `weave restart` | `docker restart` | `docker stop` followed by `docker start` a good while later | policy restart after reboot
------|----------------------|----------------------|------------------------------------------------------------------------------|------
IPAM addresses specified with `WEAVE_CIDR`        | retained | retained | retained | retained 
non-IPAM addresses specified with `WEAVE_CIDR` | retained | retained | retained | retained
IPAM addresses specified in `weave attach`               | retained | retained | retained | retained
non-IPAM addresses specified in `weave attach`        | retained | retained | retained | retained

awh:
As far as I understand it a 'bucket' in boltdb is a namespace for a set of key value pairs. I'm curious as to what motivated this arrangement with two buckets:

Bucket | Key | Value
-------|-----|------
$PEERNAME | $PEERNAME | `nil`
"ring" | "ring" | `<serialised ring>`

Instead of:

Bucket | Key | Value
-------|-----|------
"ipam" | "peername" | $PEERNAME
"ipam" | "ring" | `<serialised ring>`

Or

Bucket | Key | Value
-------|-----|------
$PEERNAME | "ring" | `<serialised ring>`

bboreham:
When I added persisting individual address ownership (#1945), I needed another bucket.

So that gives us:

Bucket | Key | Value
-------|-----|------
$PEERNAME | $PEERNAME | `nil`
"ring" | "ring" | `<serialised ring>`
"owned" | `<containerIDs>` | `<addresses>`

If we put everything under a bucket "ipam" or `$PEERNAME` then we've just added an extra level that has to be traversed on every update, for little benefit.

Given that, putting `$PEERNAME` alongside `ring` is an unnecessary coupling.

awh:
> When I added persisting individual address ownership (#1945), I needed another bucket.

Ah, OK, that makes more sense now.

> If we put everything under a bucket "ipam" or $PEERNAME then we've just added an extra level that has to be traversed on every update, for little benefit.

This I don't understand though - how is putting all the keys in one bucket worse than putting each key in its own bucket?

> Given that, putting $PEERNAME alongside ring is an unnecessary coupling.

Fair enough - in this case, wouldn't

Bucket | Key | Value
-------|-----|------
"peername" | "peername"| $PEERNAME
"ring" | "ring" | `<serialised ring>`
"owned" | `<containerIDs>` | `<addresses>`

be more uniform?

bboreham:
> This I don't understand though - how is putting all the keys in one bucket worse than putting each key in its own bucket?

You'd have to make the assumption that a containerID is never "ring" to literally put them all in the same bucket.  I guess that works, but it seems shonky.

Putting the peername under a key would indeed be more uniform, at the cost of a few extra lines of code.

Note that for `PeerName` code simply looks for a bucket with the right ID - there isn't a key or value under it.  (I updated my previous answer to match this)

awh:
>> This I don't understand though - how is putting all the keys in one bucket worse than putting each key in its own bucket?

> You'd have to make the assumption that a containerID is never "ring" to literally put them all in the same bucket. I guess that works, but it seems shonky.

Completely agree, I wasn't suggesting that - I was just requesting clarification on what you meant by "we've just added an extra level that has to be traversed on every update" as it didn't seem to me that we had.

> Putting the peername under a key would indeed be more uniform, at the cost of a few extra lines of code.
> Note that for PeerName code simply looks for a bucket with the right ID - there isn't a key or value under it. (I updated my previous answer to match this)

Got it, I misapprehended the code - seems reasonable. Thinking about the peer name check itself, which is presumably to stop a 'new' peer from causing havoc with an 'old' peer's ring - is silently deleting the old data the right thing to do, or should we refuse to start so that a human can intervene?

bboreham:
The peer name will change on `weave reset` if the machine's unique ID isn't available, or if it was available and it changed.  At least in the first case I don't think we should stop and wait for assistance.

awh:
> At least in the first case I don't think we should stop and wait for assistance.

Does not `weave reset` also remove the router container, thereby obliterating the database?

bboreham:
Ah yes.

bboreham:
A better example would be the peer name will change on host reboot, if a unique ID isn't available. To halt Weave Net at that point would be a large any likely surprising change in behaviour from previous releases.

awh:
One final comment on the database - it's probably a good idea to include a version field to help with schema upgrades.

Needs rebase.

bboreham:
I added a version check as you suggested (failing on major version mismatch).

This required some change to the bucket scheme, so it is now:

Bucket | Key | Value
-------|-----|------
"top" | "version"| $VERSION
 | "peername"| $PEERNAME
"ring" | "ring" | `<serialised ring>`
"owned" | `<containerID>` | `<addresses>`
 | `<containerID>` | `<addresses>`

Also the previous code to delete all buckets in a loop broke the rules of BoltDB, so I have simplified to just delete them one by one.

awh:
Superseded by #1971.

Issue #1932, opened 2016-01-27
Crash in /w/w


tomwilkie:
Not super important, as I'm hacking on proxy/cni integrations, but I hit this.  We should handle more gracefully.

```
Toms-MacBook-Pro:weave twilkie$ docker logs high_brattain
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x458020]

goroutine 1 [running]:
github.com/weaveworks/weave/net.EnsureInterfaceAndMcastRoute.func1(0xdd, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc82008b028, 0x4, 0xfd8, ...)
	/go/src/github.com/weaveworks/weave/net/if.go:56 +0x30
github.com/weaveworks/weave/net.EnsureInterfaceAndMcastRoute(0x55bc08, 0x5, 0x0, 0x0, 0x0)
	/go/src/github.com/weaveworks/weave/net/if.go:69 +0x5ae
main.checkNetwork(0x0, 0x0)
	/go/src/github.com/weaveworks/weave/prog/weavewait/check_network_iface_mcast.go:10 +0x38
main.main()
	/go/src/github.com/weaveworks/weave/prog/weavewait/main.go:20 +0x6b

goroutine 6 [chan send]:
github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink.LinkSubscribe.func2(0xc820014120, 0xc82000a0e0)
	/go/src/github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink/link_linux.go:847 +0x2e6
created by github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink.LinkSubscribe
	/go/src/github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink/link_linux.go:850 +0x107

goroutine 7 [runnable]:
github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink.RouteSubscribe.func1(0xc820014240, 0xc82000a540)
	/go/src/github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink/route_linux.go:305 +0x36
created by github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink.RouteSubscribe
	/go/src/github.com/weaveworks/weave/vendor/github.com/vishvananda/netlink/route_linux.go:307 +0xe6

goroutine 8 [syscall]:
[...]

PR #1933, opened 2016-01-27
[WIP] Hack up the Weave Proxy to talk CNI

tomwilkie:
This was an idea in the pub, we could make the weave proxy talk to CNI plugins so they could work nicer with Docker.

And it works!  This is POC code, but feedback is welcome:

```
$ # setup your CNI network*
$ ./weave launch-proxy --without-dns --no-multicast-route
$ docker run -ti -l cni.network=net1 alpine /bin/sh
/ # ifconfig
...
ethwe     Link encap:Ethernet  HWaddr 86:5C:46:0F:F8:A0  
          inet addr:192.168.0.6  Bcast:0.0.0.0  Mask:255.255.255.255
          inet6 addr: fe80::845c:46ff:fe0f:f8a0%32623/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:7 errors:0 dropped:0 overruns:0 frame:0
          TX packets:6 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:738 (738.0 B)  TX bytes:508 (508.0 B)
...
```

Details:
- This uses @squaremo's libcni
- We're telling the CNI plugin to make an `ethwe` so that the /w/w script works.
- We have to be careful about volume mounting in the CNI configs and plugins into the proxy container; running the proxy not in a container would be a big help with @awh 
- We haven't wired up delete yet.
- `--no-multicast-route` was needed as we hit #1932

Thanks to @tomdee for his help with this.
- How to setup a calico CNI network:

```
$ mkdir -P /etc/cni/net.d
$ cat >/etc/cni/net.d/10-calico-net1.conf <<EOF
{
    "name": "net1",
    "type": "calico",
    "ipam": {
        "type": "calico-ipam"
    }
}
EOF
$ wget -O /etc/cni/plugins/calico https://github.com/projectcalico/calico-cni/releases/download/v1.0.0/calico
$ chmod +x /etc/cni/plugins/calico
$ wget https://github.com/projectcalico/calico-containers/releases/download/v0.15.0/calicoctl
$ chmod +x calicoctl
$ docker run --detach \
    --net=host \
   --name calico-etcd quay.io/coreos/etcd:v2.0.11 \
    --advertise-client-urls "http://127.0.0.1:2379" \
    --listen-client-urls "http://127.0.0.1:2379"
sudo ./calicoctl node
```

### DIFF
--- a/prog/weaveexec/Dockerfile
+++ b/prog/weaveexec/Dockerfile
@@ -1,4 +1,5 @@
-FROM alpine
+#FROM alpine
+FROM jeanblanchard/alpine-glibc
 
 MAINTAINER Weaveworks Inc <help@weave.works>
 LABEL works.weave.role=system
--- a/.gitmodules
+++ b/.gitmodules
@@ -52,3 +52,6 @@
 [submodule "vendor/github.com/weaveworks/mesh"]
 	path = vendor/github.com/weaveworks/mesh
 	url = https://github.com/weaveworks/mesh
+[submodule "vendor/github.com/appc/cni"]
+	path = vendor/github.com/appc/cni
+	url = https://github.com/appc/cni
--- a/proxy/cni.go
+++ b/proxy/cni.go
@@ -1,0 +1,57 @@
+package proxy
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+
+	"github.com/appc/cni/libcni"
+	"github.com/fsouza/go-dockerclient"
+	. "github.com/weaveworks/weave/common"
+)
+
+var (
+	cniScript     = "./cni.sh"
+	cniNetwork    = "cni.network"
+	cniConfPath   = "/etc/cni/net.d"
+	cniPluginPath = []string{"/home/weave", "/usr/bin", "/"}
+	cniIfName     = "ethwe" // /w/w is waiting for an interface of this name
+)
+
+func useCNI(container *docker.Container) bool {
+	_, ok := container.Labels[cniNetwork]
+	return ok
+}
+
+func (proxy *Proxy) attachCNI(container *docker.Container, orDie bool) {
+	network := container.Labels[cniNetwork]
+
+	// read the json config to find the plugin exe
+	conf, err := libcni.LoadConf(cniConfPath, network)
+	if err != nil {
+		Log.Warningf("Attaching container %s using CNI plugin failed: %s", container.ID, string(stderr))
+		return errors.New(string(stderr))
+	}
+
+	// tell plugin to attach container
+	c := RuntimeConf.CNIConfig{Path: cniPluginPath}
+	_, err := c.AddNetwork(conf, libcni.RuntimeConf{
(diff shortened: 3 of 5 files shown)


squaremo:
I like this idea.

pidster:
+1

tomdee:
+1 too!. This was a lot of fun to work on, thanks @tomwilkie 

An issue that we found when developing this is that the calico CNI plugin could return an error for very short lived containers. It would try to perform an operation in the namespace but the namespace no longer existed. 

The container runs the `ww` process before ceding control to the container's real entrypoint. The `ww` process blocks until the veth device is set to "UP" but that isn't enough to detect that the CNI plugin has finished executing. In the case of the calico CNI plugin, it needs to set up some routes in the name space which need to reference the veth, which can't be set until the device is set to up. 

I don't think the weave proxy should be allowing the container's real entrypoint to run until the CNI plugin has completed. Could this be implemented with signals? @tomwilkie I think you indicated that there had been problems with this approach in the past.

rade:
> In the case of the calico CNI plugin, it needs to set up some routes in the name space which need to reference the veth, which can't be set until the device is set to up. 

...we had exactly the same problem in weave, which is why `weavewait` waits for the interface *and a route entry*.

> Could this be implemented with signals?

No.

>  @tomwilkie I think you indicated that there had been problems with this approach in the past.

Exactly. See #1314, which fixed a number of issues: #1209 (very nasty), #1300 (likewise).

tomdee:
Maybe because the proxy is being run with `--no-multicast-route` it's not waiting for a route? 

Either way, for this to be "production ready" I think the waiter needs to wait till the plugin has finished executing.

rade:
> Maybe because the proxy is being run with `--no-multicast-route` it's not waiting for a route? 

That was done because of #1932, which is now fixed.

So I guess the proxy could add the route after the `cni.AddNetwork` call returns.

prog/weaveexec/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM alpine
+#FROM alpine
+FROM jeanblanchard/alpine-glibc
```
rade: why?

tomdee: The calico plugin (and I image a number of other CNI plugins too) have a dependency on glibc.

(Obviously the commented out line should be removed)

rade: but the golang CNI API itself doesn't have such a dependency, and that is all we are invoking here.

rade: Ah, it looks like the CNI API exec's stuff. Is that correct? If so, how is one supposed to get the required executables (and all their dependencies) into the proxy container?

tomwilkie: That is correct.  The changes in the docker file bind mount in /etc/cni/plugins to get the the executables in there.  And they need to be statically compiled to work.

bboreham:
Closing as this is a proof of concept; we're not going to merge it.

danieledwardgeorgehitchcock:
Hi,

Could this PR be split into its own maintained project?

It would be great to be able to use an off the shelf product such as this as an interface between Docker and CNI

PR #1910, opened 2016-01-19
cope with errors during route traversal

rade:
...and bypass `netlink.LinkByIndex`.

Fixes #1909.

PR #1935, opened 2016-01-28
Reduce docker volumes for weavewait to just one set per release

### LINKED ISSUE #1757: Lots of copies of weavewait

bboreham:
Fixes #1757

- Create an auxilliary container 'weavevolumes' to hold the '/w/w'
  files, and map into that instead of creating new volumes every time.
- Show "/bin/false" as its command in `docker ps` instead of
"/home/weave/sigproxy".
- Remove volumes when cleaning up smoketest containers.

Replaces #1792 to change target branch - some discussion on design there.

### DIFF
--- a/prog/weaveexec/Dockerfile
+++ b/prog/weaveexec/Dockerfile
@@ -4,9 +4,6 @@ MAINTAINER Weaveworks Inc <help@weave.works>
 LABEL works.weave.role=system
 
 WORKDIR /home/weave
-VOLUME /w
-VOLUME /w-noop
-VOLUME /w-nomcast
 ENTRYPOINT ["/home/weave/sigproxy", "/home/weave/weave"]
 
 RUN apk add --update \
--- a/test/config.sh
+++ b/test/config.sh
@@ -176,7 +176,7 @@ proxy_start_container_with_dns() {
 rm_containers() {
     host=$1
     shift
-    [ $# -eq 0 ] || docker_on $host rm -f "$@" >/dev/null
+    [ $# -eq 0 ] || docker_on $host rm -f -v "$@" >/dev/null
 }
 
 container_ip() {
--- a/weave
+++ b/weave
@@ -344,7 +344,7 @@ IMAGE=$BASE_IMAGE:$IMAGE_VERSION
 BASE_PLUGIN_IMAGE=$DOCKERHUB_USER/plugin
 PLUGIN_IMAGE=$BASE_PLUGIN_IMAGE:$IMAGE_VERSION
 PLUGIN_CONTAINER_NAME=weaveplugin
-VOLUMES_CONTAINER_NAME=weavevolumes
+VOLUMES_CONTAINER_NAME=weavevolumes-$IMAGE_VERSION
 
 PROCFS=${PROCFS:-/proc}
 DOCKER_BRIDGE=${DOCKER_BRIDGE:-docker0}
@@ -1672,7 +1672,8 @@ launch_proxy() {
     mkdir -p /var/run/weave
     # Create a data-only container to mount the weavewait files from
     if ! docker inspect -f ' ' $VOLUMES_CONTAINER_NAME > /dev/null 2>&1 ; then
-       docker create -v /w -v /w-noop -v /w-nomcast --name=$VOLUMES_CONTAINER_NAME --entrypoint=/bin/false $EXEC_IMAGE >/dev/null
+       docker create -v /w -v /w-noop -v /w-nomcast --name=$VOLUMES_CONTAINER_NAME \
+           --label=weavevolumes --entrypoint=/bin/false $EXEC_IMAGE >/dev/null
     fi
     PROXY_CONTAINER=$(docker run --privileged -d --name=$PROXY_CONTAINER_NAME --net=host \
         $PROXY_VOLUMES \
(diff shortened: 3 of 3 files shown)


rade:
before:
```
$ docker volume ls -q | wc -l
0
$ weave launch; docker volume ls -q | wc -l
6
$ weave stop;   docker volume ls -q | wc -l
6
$ weave launch; docker volume ls -q | wc -l
12
$ weave reset;  docker volume ls -q | wc -l
12
$ weave launch; docker volume ls -q | wc -l
18
```
after
```
$ docker volume ls -q | wc -l
0
$ weave launch; docker volume ls -q | wc -l
6
$ weave stop;   docker volume ls -q | wc -l
6
$ weave launch; docker volume ls -q | wc -l
9
$ weave reset;  docker volume ls -q | wc -l
6
$ weave launch; docker volume ls -q | wc -l
12
```
So this is an improvement, but not a fix. Looks like there is a volume leak in both the proxy and plugin, and this PR fixes the former but not the latter:
```
$ docker volume ls -q | wc -l
0
$ weave launch-proxy >/dev/null; docker volume ls -q | wc -l
3
$ weave   stop-proxy >/dev/null; docker volume ls -q | wc -l
3
$ weave launch-proxy >/dev/null; docker volume ls -q | wc -l
3
$ weave reset;                   docker volume ls -q | wc -l
0
$ weave launch-proxy >/dev/null; docker volume ls -q | wc -l
3
```
```
$ docker volume ls -q | wc -l
0
$ weave launch-plugin >/dev/null; docker volume ls -q | wc -l
3
$ weave   stop-plugin >/dev/null; docker volume ls -q | wc -l
3
$ weave launch-plugin >/dev/null; docker volume ls -q | wc -l
6
$ weave reset;                    docker volume ls -q | wc -l
6
$ weave launch-plugin >/dev/null; docker volume ls -q | wc -l
9
```

Possibly this is a separate issue.

rade:
```
root@xps:/var/lib/docker/volumes# ls -ltR
.:
total 12
drwxr-xr-x 3 root root 4096 Jan 28 17:27 bae39b7a29cbac43dea177694af83c0bacfe26bb7e1df5023bb5840e2e48dd0b
drwxr-xr-x 3 root root 4096 Jan 28 17:27 b7fbce11e5522e49fe7cb7e979ff7b6e6cfb5a4c0a4db65b7bb212efca52ae12
drwxr-xr-x 3 root root 4096 Jan 28 17:27 a577deb53aadc24b22cc3c6f44397b61733592ebcbdf344a6cf0f7992f1c686d

./bae39b7a29cbac43dea177694af83c0bacfe26bb7e1df5023bb5840e2e48dd0b:
total 4
drwxr-xr-x 2 root root 4096 Jan 28 17:27 _data

./bae39b7a29cbac43dea177694af83c0bacfe26bb7e1df5023bb5840e2e48dd0b/_data:
total 2520
-rwxr-xr-x 1 root root 2578952 Jan 28 10:25 w

./b7fbce11e5522e49fe7cb7e979ff7b6e6cfb5a4c0a4db65b7bb212efca52ae12:
total 4
drwxr-xr-x 2 root root 4096 Jan 28 17:27 _data

./b7fbce11e5522e49fe7cb7e979ff7b6e6cfb5a4c0a4db65b7bb212efca52ae12/_data:
total 3164
-rwxr-xr-x 1 root root 3238344 Jan 28 10:25 w

./a577deb53aadc24b22cc3c6f44397b61733592ebcbdf344a6cf0f7992f1c686d:
total 4
drwxr-xr-x 2 root root 4096 Jan 28 17:27 _data

./a577deb53aadc24b22cc3c6f44397b61733592ebcbdf344a6cf0f7992f1c686d/_data:
total 3208
-rwxr-xr-x 1 root root 3284560 Jan 28 10:25 w
```

Based on the file sizes, these are the three flavours of weavewait.

bboreham:
Are you sure you are running a new plugin?

bboreham:
Do `docker history weaveworks/plugin` and check it is built off the same layers as the current `weaveexec`

rade:
False alarm. The problem was https://github.com/weaveworks/weave/commit/c6d0149ba64e75322b5edaa40cee14e043933594.

rade:
Isn't there a problem here with upgrades? Since we only remove the volume on reset, won't a "soft" upgrade of `weave stop; <upgrade>; weave launch` end up stick old-version weavewaits into subsequently launched containers?

I guess weavewait doesn't change all that often, and perhaps we should insist on a `weave reset` when it does. Seems a bit crude though, especially when the change is a bug fix like the one we just did in #1910.

I am actually curious what the current behaviour is: if you have a container that was started via the proxy, then soft-upgrade weave, and then bounce the container, is the new incarnation a) broken (i.e. it references a weavewait entrypoint which no longer exists), b) using the old-version weavewait, or c) using the new-version weavewait?

bboreham:
To your last question, b: it will have the same weavewait as when it was started.

bboreham:
We could attach the weave version number to the `weavevolumes` container name, so the files would leak very slowly under an upgrade scenario such as you describe.

rade:
> We could attach the weave version number to the `weavevolumes` container name, so the files would leak very slowly under an upgrade scenario such as you describe.

Was thinking the same. But we then would want `weave reset` to remove *all* such containers.

PR #1936, opened 2016-01-28
Remove Unix domain sockets on shutdown

### LINKED ISSUE #1934: `docker volume ls` hangs on Docker v1.10 after `weave stop[-plugin]|reset`
When the plugin is launched it creates two Unix domain sockets `/var/run/docker/plugins/weave.sock` and `/var/run/docker/plugins/weavemesh.sock`.  These are not removed when the plugin exits or when the user runs `weave reset`.

Leaving them around when the plugin is not running causes `docker volume ls` to hang for 15 seconds, observed on Docker 1.10-rc2 but not on 1.9.1.

bboreham:
Fixes #1934.

Note part of #1901 will need to be re-done for the merge to master.

### DIFF
--- a/prog/plugin/main.go
+++ b/prog/plugin/main.go
@@ -93,17 +93,23 @@ func main() {
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, os.Kill, syscall.SIGTERM)
 
+	cleanup := func() {
+		globalListener.Close()
+		meshListener.Close()
+		_ = os.Remove(address)
+		_ = os.Remove(meshAddress)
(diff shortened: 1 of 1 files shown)


rade:
This suffers from the scoping issue that I fixed in 31f736b3, which means that the listener vars will always be `nil`. Also note that one or both of the addresses might be "" (and consequently one or both the listeners might be absent) - this isn't handled. Finally, there are a bunch of `Log.Fatal`s that will bypass `cleanup()`.

bboreham:
Equally the program could `panic()` and not run cleanup. I don't think it can be 100%.

PR #1941, opened 2016-01-29

### LINKED ISSUE #1873: Improve visibility of plugin startup problems

bboreham:
Fixes #1873

Replacement for #1879, now on master branch

### DIFF
--- a/net/listen.go
+++ b/net/listen.go
@@ -1,0 +1,18 @@
+package net
+
+import (
+	"fmt"
+	"net"
+	"os"
+)
+
+func ListenUnixSocket(pathname string) (net.Listener, error) {
+	if err := os.Remove(pathname); err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("Error removing existing unix socket: %s", err)
+	}
+	listener, err := net.Listen("unix", pathname)
+	if err != nil {
+		return nil, fmt.Errorf("ListenUnixSocket failed: %s", err)
+	}
+	return listener, nil
+}
--- a/weave
+++ b/weave
@@ -1724,6 +1724,7 @@ launch_plugin_if_not_running() {
         $PLUGIN_IMAGE "$@") ; then
         return 1
     fi
+    wait_for_status $PLUGIN_CONTAINER_NAME http_call_unix $PLUGIN_CONTAINER_NAME status.sock
     util_op create-plugin-network weave weavemesh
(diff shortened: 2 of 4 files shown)


net/listen.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+	"net"
+	"os"
+)
+
+func ListenUnixSocket(pathname string) (net.Listener, error) {
+	if err := os.Remove(pathname); err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("Error removing existing status socket: %s", err)
```
rade: s/status/unix

bboreham:
Fixed cutpasta

weave
```diff
@@ -1286,6 +1280,7 @@ collect_dns_add_remove_args() {
     [ $# -eq 0 -a \( -n "$C" -o \( $IP_COUNT -gt 0 -a -n "$FQDN" \) \) ] || usage
     check_running $CONTAINER_NAME
     if [ -n "$C" ] ; then
+        check_running $C
```
rade: why?

rade:
This surprised me...

```
$ weave launch-router
dc4d47c5899bb063a771415f8290f83c87bfb19de6f82326370cae5e5c141c69
$ weave launch-plugin --foo
The weaveplugin container has died. Consult the container logs for further details.
$ weave launch-plugin
weaveplugin is restarting; you can stop it with 'weave stop-plugin'.
```
Perhaps we should do a `docker kill` in `wait_for_status`?

weave
bboreham: to maintain the previous behaviour; `container_id` used to do this check

bboreham:
> Perhaps we should do a docker kill in wait_for_status?

Done.

weave
rade: `container_id` is called from lots of places. Yet you only changed this one. What makes this one special?

bboreham: all the others are covered by nearby checks

rade:
This works well. Not just for the plugin, but e.g.
```
WEAVE_DOCKER_ARGS="--restart=always" weave launch --foo
```
will now error and stop the container instead of hanging.

Issue #1940, opened 2016-01-29
Remove proxy dependency on weave script


awh:
The proxy currently relies on the weave script for a number of operations; these dependencies should be removed to further #307:
- [ ] attach
- [x] attach-router
- [x] docker-bridge-ip
- [x] ps

PR #1944, opened 2016-02-01
Use netlink instead of shelling out to `weave ps`

awh:
The proxy currently invokes `weave ps` to rewrite docker inspect network settings; this PR retrieves the IPs from the weave interface directly with netlink.

Partially addresses #1940.

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -1,6 +1,11 @@
 package common
 
 import (
+	"fmt"
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+	"net"
+	"runtime"
 	"strings"
 )
 
@@ -17,4 +22,77 @@ func ErrorMessages(errors []error) string {
 		result = append(result, err.Error())
 	}
 	return strings.Join(result, "\n")
+}
+
+func WithNetNS(ns netns.NsHandle, work func() error) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	oldNs, err := netns.Get()
+	if err == nil {
+		defer oldNs.Close()
+
+		err = netns.Set(ns)
+		if err == nil {
+			defer netns.Set(oldNs)
+
+			err = work()
+		}
+	}
+
+	return nil
+}
+
+type NetDev struct {
+	MAC   net.HardwareAddr
+	CIDRs []*net.IPNet
+}
+
+// Search the network namespace of a process for interfaces matching a predicate
+func FindNetDevs(procPath string, processID int, match func(string) bool) ([]NetDev, error) {
+	var netDevs []NetDev
+
+	ns, err := netns.GetFromPath(fmt.Sprintf("%s/%d/ns/net", procPath, processID))
+	if err != nil {
+		return nil, err
+	}
+	defer ns.Close()
+
+	err = WithNetNS(ns, func() error {
+		links, err := netlink.LinkList()
+		if err != nil {
+			return err
+		}
+		for _, link := range links {
+			if match(link.Attrs().Name) {
+				addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
+				if err != nil {
+					return err
+				}
+
+				netDev := NetDev{MAC: link.Attrs().HardwareAddr}
+				for _, addr := range addrs {
+					netDev.CIDRs = append(netDev.CIDRs, addr.IPNet)
+				}
+				netDevs = append(netDevs, netDev)
+			}
+		}
+		return nil
+	})
+
+	return netDevs, err
+}
+
+// Lookup the weave interface of a container
+func GetWeaveNetDevs(procPath string, processID int) ([]NetDev, error) {
+	return FindNetDevs(procPath, processID, func(name string) bool {
+		return strings.HasPrefix(name, "ethwe")
+	})
+}
+
+// Get the weave bridge interface
+func GetBridgeNetDev(procPath, bridgeName string) ([]NetDev, error) {
+	return FindNetDevs(procPath, 1, func(name string) bool {
+		return name == bridgeName
+	})
 }
--- a/prog/weaveutil/addrs.go
+++ b/prog/weaveutil/addrs.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"github.com/fsouza/go-dockerclient"
+	"github.com/weaveworks/weave/common"
+)
+
+func containerAddrs(args []string) error {
+	if len(args) < 2 {
+		cmdUsage("container-addrs", "<procPath> <bridgeName> [containerID ...]")
+	}
+
+	procPath := args[0]
+	bridgeName := args[1]
+
+	c, err := docker.NewVersionedClientFromEnv("1.18")
+	if err != nil {
+		return err
+	}
+
+	for _, containerID := range args[2:] {
+		netDevs, err := getNetDevs(procPath, bridgeName, c, containerID)
+		if err != nil {
+			if _, ok := err.(*docker.NoSuchContainer); ok {
+				continue
+			}
+			return err
+		}
+
+		for _, netDev := range netDevs {
+			fmt.Printf("%12s %s", containerID, netDev.MAC.String())
+			for _, cidr := range netDev.CIDRs {
+				prefixLength, _ := cidr.Mask.Size()
+				fmt.Printf(" %v/%v", cidr.IP, prefixLength)
+			}
+			fmt.Println()
+		}
+	}
+
+	return nil
+}
+
+func getNetDevs(procPath, bridgeName string, c *docker.Client, containerID string) ([]common.NetDev, error) {
+	if containerID == "weave:expose" {
+		return common.GetBridgeNetDev(procPath, bridgeName)
+	}
+
+	container, err := c.InspectContainer(containerID)
+	if err != nil {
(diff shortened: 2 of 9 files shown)


awh:
Decided with @rade to include porting `weave ps` itself to use netlink via `weaveutil` in this PR -> assigning to myself.

bboreham:
It is possible for a container to have more than one weave interface; only one of them from non-plugin operations (called "ethwe"), but there can be others added via `docker network connect`.

weave
```diff
@@ -1041,32 +1037,16 @@ wait_for_status() {
...
-            ADDRS_CMD="ip addr show dev $BRIDGE"
-        else
-            ADDRS_CMD="with_container_netns $CONTAINER_ID container_ip_addr_show"
-        fi
-        if CONTAINER_ADDRS=$($ADDRS_CMD 2>/dev/null) ; then
-            CONTAINER_MAC=$(echo "$CONTAINER_ADDRS" | addrs_from_ip_addr_show 'link/ether')
-            CONTAINER_IPS=$(echo "$CONTAINER_ADDRS" | addrs_from_ip_addr_show 'inet')
-            if [ -n "$CONTAINER_MAC$CONTAINER_IPS" ] ; then
-                $COMMAND "$CONTAINER_ID" "$CONTAINER_MAC" "$CONTAINER_IPS"
+        if CONTAINER_ADDRS=$(util_op container-addrs $CONTAINER_ID) ; then
```
rade: it would be lovely if the `container-addrs` invocation was outside the loop, applied to `$@`.

awh: It would, but then it would be responsible for invoking `$COMMAND` (currently one of `echo_ips`, `echo_addresses`, `echo_cidrs` and `ipam_reclaim`), as as those are all shell functions that rely on unexported variables that's not entirely straightforward.

The first three could be implemented directly as flags to `container-addrs`, removing the need to invoke via `with_container_addresses` at the top level; the last is more tricky, as that logic should probably live in the golang implementation of `attach_router`. I think we'll have to wait until then to clean this up some more.

rade: That's not quite what I had in mind. I was thinking of invoking `container-addrs $@` and then looping over the output. That's probably more trouble than it is worth though.

tomwilkie:
We should use this in scope, where we shell out to weave ps.

weave
```diff
@@ -1041,33 +1037,14 @@ wait_for_status() {
...
-        fi
-        if CONTAINER_ADDRS=$($ADDRS_CMD 2>/dev/null) ; then
-            CONTAINER_MAC=$(echo "$CONTAINER_ADDRS" | addrs_from_ip_addr_show 'link/ether')
-            CONTAINER_IPS=$(echo "$CONTAINER_ADDRS" | addrs_from_ip_addr_show 'inet')
-            if [ -n "$CONTAINER_MAC$CONTAINER_IPS" ] ; then
-                $COMMAND "$CONTAINER_ID" "$CONTAINER_MAC" "$CONTAINER_IPS"
-            fi
+
+    util_op container-addrs $PROCFS $BRIDGE "$@" | while read CONTAINER_ID CONTAINER_MAC CONTAINER_IPS; do
+        if [ -n "$CONTAINER_MAC$CONTAINER_IPS" ] ; then
```
awh: This test is unnecessary now I think

awh:
@bboreham am I right in thinking that prior to this PR, if you have multiple weave interfaces in a container you'll get output something like this?
```
<mycontainer> <iface1mac> <iface2mac> <iface1ip> <iface2ip>
```
This PR will make the output look like this instead, which is better IMO:
```
<mycontainer> <iface1mac> <iface1ip>
<mycontainer> <iface2mac> <iface2ip>
```
Thoughts?

bboreham:
I guess that's better. We didn't document the previous behaviour, or know how anyone would use it.

awh:
> I guess that's better

The existing scheme is ambiguous if an interface has more than one IP, so I'd call it a definite improvement.

awh:
This is now ready for review. Note that I consider the placement of the shared functions in `common` to be a temporary arrangement; these should live in (a) more focused package(s), we'll have a clearer idea what they should look like as #1940 and #307 progress.

awh:
> We should use this in scope, where we shell out to weave ps.

You might want to wait until the code settles down a bit - see my previous comment - perhaps create an issue in weaveworks/scope to track. You'd prefer this to a REST API to retrieve the info you need?

proxy/json.go
```diff
@@ -45,6 +46,25 @@ func (j jsonObject) String(key string) (string, error) {
 	return result, nil
 }
 
+func (j jsonObject) Int(key string) (int, error) {
```
bboreham: Other than detecting when non-numeric characters have been supplied, do we gain anything from implementing this function?

awh: The underlying map value type for JSON number fields is `json.Number`, not `string` - if you try to use the `String()` method above on it you will get an `UnmarhsalWrongTypeError`

common/utils.go
```diff
@@ -18,3 +23,83 @@ func ErrorMessages(errors []error) string {
 	}
 	return strings.Join(result, "\n")
 }
+
+func WithNetNS(ns netns.NsHandle, work func() error) error {
+	errChan := make(chan error)
+
+	go func() {
```
bboreham: why?

common/utils.go
```diff
@@ -18,3 +23,83 @@ func ErrorMessages(errors []error) string {
 	}
 	return strings.Join(result, "\n")
 }
+
+func WithNetNS(ns netns.NsHandle, work func() error) error {
```
bboreham: should PR this upstream

common/utils.go
```diff
@@ -18,3 +23,83 @@ func ErrorMessages(errors []error) string {
...
+			}
+		}
+
+		errChan <- err
+	}()
+
+	return <-errChan
+}
+
+type WeaveInterface struct {
```
bboreham: Is not really Weave-related, nor is it an interface.  Addresses?

common/utils.go
```diff
@@ -18,3 +23,83 @@ func ErrorMessages(errors []error) string {
...
+		}
+		for _, link := range links {
+			if match(link.Attrs().Name) {
+				addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
+				if err != nil {
+					return err
+				}
+
+				weaveInterface := WeaveInterface{}
+				weaveInterface.MAC = link.Attrs().HardwareAddr
```
bboreham: could combine this into previous line

proxy/proxy.go
```diff
@@ -590,11 +591,21 @@ func (proxy *Proxy) updateContainerNetworkSettings(container jsonObject) error {
...
+	if err != nil {
+		return err
+	}
+
 	if err := proxy.waitForStartByIdent(containerID); err != nil {
 		return err
 	}
-	mac, ips, nets, err := weaveContainerIPs(containerID)
-	if err != nil || len(ips) == 0 {
+	weaveInterfaces, err := GetWeaveInterfaces(proxy.ProcPath, pid)
```
bboreham: you could have another variant of this that searched for exactly "ethwe", so you wouldn't have to arbitrarily pick the first answer.

common/utils.go
awh: Faulty reasoning. I was thinking that I didn't want to lock the thread servicing the calling goroutine, but that's bogus for two reasons - that particular goroutine needs to block until the work is done anyway, and other goroutines for other requests will automatically get shifted to another thread by the scheduler as long as max procs is high enough. I'll remove it.

common/utils.go
awh: To `vishvananda/netns`? I can do that async with this PR getting merged, and then can submit another PR here to update the vendored dependency when it makes it upstream.

bboreham: Right

common/utils.go
awh: It is meant to represent an interface (in the network sense), and will probably sprout extra things like MTU in a future PR. `Link`?

common/utils.go
awh: OK

common/utils.go
bboreham: there is `net.Interface`, which has members `HardwareAddr` and `Addrs()`

proxy/proxy.go
awh: I'll add it if someone else needs to do the same thing.

common/utils.go
awh: Looks appealing, but I don't think I can use it - the `Addrs()` function is hardwired behind the scenes to look things up in the default netns via netlink. Furthermore, `net.Interface` is a struct and not an interface, so I can't provide an alternative implementation.

awh:
Comments addressed; will squash when you're ready to merge.

prog/weaveutil/addrs.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"github.com/fsouza/go-dockerclient"
+	"github.com/weaveworks/weave/common"
+)
+
+func containerAddrs(args []string) error {
+	if len(args) < 2 {
+		return fmt.Errorf("usage: container-addrs <procPath> <bridgeName> [containerID ...]")
+	}
```
rade: That is wrong. See how the other commands do this.

bboreham:
LGTM

awh:
Squashified.

PR #1948, opened 2016-02-02
Improve IPAM Ring unit testing, and streamline ring.ClaimForPeers

bboreham:
Begun while investigating #1946; wasn't actually related but contributed here to improve code coverage.

The test for `ring.ClaimForPeers` doesn't do any special checks, relying on `checkInvariants` which is called via `defer`

### DIFF
--- a/ipam/ring/ring_test.go
+++ b/ipam/ring/ring_test.go
@@ -60,13 +60,19 @@ func TestInsert(t *testing.T) {
 
 	ring.Entries.entry(0).Free = 0
 	ring.Entries.insert(entry{Token: dot245, Peer: peer1name})
-	ring2 := New(start, end, peer1name)
-	ring2.Entries = []*entry{{Token: start, Peer: peer1name, Free: 0}, {Token: dot245, Peer: peer1name}}
-	require.Equal(t, ring2, ring)
+	check := []RangeInfo{
+		{Peer: peer1name, Range: address.Range{Start: start, End: dot245}},
+		{Peer: peer1name, Range: address.Range{Start: dot245, End: end}},
+	}
+	require.Equal(t, check, ring.AllRangeInfo())
 
 	ring.Entries.insert(entry{Token: dot10, Peer: peer1name})
-	ring2.Entries = []*entry{{Token: start, Peer: peer1name, Free: 0}, {Token: dot10, Peer: peer1name}, {Token: dot245, Peer: peer1name}}
-	require.Equal(t, ring2, ring)
+	check2 := []RangeInfo{
+		{Peer: peer1name, Range: address.Range{Start: start, End: dot10}},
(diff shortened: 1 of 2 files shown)


ipam/ring/ring_test.go
```diff
@@ -506,6 +506,22 @@ func TestOwner(t *testing.T) {
 
 }
 
+func makePeerName(i int) mesh.PeerName {
+	peer, _ := mesh.PeerNameFromString(fmt.Sprintf("%02d:%02d:00:00:00:00", i/10, i%10))
```
rade: Why 10 rather than 100?

Also, we have code like this in nameserver_test too. and in mesh.peers_test. and in ipam/testutils_test (would need some tweaking), and in TestFuzzRing.

bboreham: I moved it from TestFuzzRing.

rade: Actually, we aren't going to run with more than 100 peers, are we? So I suggest we just make this `fmt.Sprintf("%02d:...:ff", i)` (the `ff` is there because all-zeros is treated specially in some parts of mesh/weave), but also add a panic for when i>100 (so that if anybody changes the numPeers in the tests they get an obvious explosion).

ipam/ring/ring_test.go
```diff
@@ -506,6 +506,22 @@ func TestOwner(t *testing.T) {
...
+	return peer
+}
+
+func TestClaimForPeers(t *testing.T) {
+	const numPeers = 25
+	start, end := ParseIP("10.21.0.1"), ParseIP("10.21.255.255")
```
rade: all the other tests use 10.0.0.x.

rade:
I would expect this test to give better coverage. It is missing two interesting branches - one which is hit when there are fewer addresses than peers, and another which I have no idea how to reach.

ipam/ring/ring_test.go
bboreham: Although we don't have tests that use more than 100 peers at the same time, `TestFuzzRingHard` does add and remove peers while incrementing `nextPeerID`, with the effect that the number seen by `makePeerName` goes over 100.

ipam/ring/ring_test.go
```diff
@@ -506,6 +506,29 @@ func TestOwner(t *testing.T) {
...
+	// Test for a range of peer counts
+	for i := 0; i < numPeers; i++ {
+		peers[i] = makePeerName(i)
+		ring := New(start, end, peers[0])
+		ring.ClaimForPeers(peers[:i+1])
+	}
+	// One more test, when the number of addresses is smaller than the number of peers
+	ring := New(start, ParseIP("10.0.0.10"), peers[0])
+	ring.ClaimForPeers(peers)
+}
```
rade: I suggest a ring size of 8 and peer counts of 1..10.

PR #1951, opened 2016-02-03
Use netlink instead of shelling out to `weave docker-bridge-ip`

awh:
Builds on #1944; partially addresses #1940.

### DIFF
--- a/prog/weaveproxy/main.go
+++ b/prog/weaveproxy/main.go
@@ -12,17 +12,21 @@ import (
 )
 
 var (
-	version             = "(unreleased version)"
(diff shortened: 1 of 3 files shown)


rade:
`docker_bridge_ip` is called in `create_bridge`. `weave --local create-bridge` is meant to work when docker isn't running, so calling `util_op` is just not something we can do here.

awh:
Rebased on revised #1944 and reverted change to `docker-bridge-ip`; needs rebase again prior to merge after #1944 has gone in.

awh:
Rebased on master and ready for review/merge.

Issue #1747, opened 2015-12-02
Support varying Docker daemon configurations across smoke tests


awh:
The current smoke tests run against a static Docker configuration which is provisioned in the test VM vagrant files. This restriction is making certain tests difficult and should be lifted:
* Testing the proxy's ability to connect to non-standard `DOCKER_HOST`
* Testing global vs local mode network plugin

bboreham:
Just to note the configuration is also provisioned in https://github.com/weaveworks/weave/blob/master/test/gce.sh#L79

awh:
Argh!Looks the the GCE images aren't using systemd then?

awh:
I think we can do this with systemd drop-ins once GCE is upgraded #1212.

rade:
We should use the systemd override mechanism in vagrant-common.rb/tweak_docker_daemon too. Currently that just patches the docker.service file. This breaks updating docker with `vagrant provision` since that will stomp on the file and explode since it restarts the daemon before tweak_docker_daemon has been re-run.

rade:
eek. The current tweak_docker_daemon is not idempotent - if you run `vagrant provision` multiple times you end up with the docker daemon attempting to listen on the same ports multiple times!

rade:
Some [further reading](http://askubuntu.com/questions/659267/how-do-i-override-or-configure-systemd-services) suggests that the correct place for the override is `/etc/systemd/system/docker.service.d/override.conf` - that's what gets created if you run `systemctl edit docker`.

rade:
> We should use the systemd override mechanism in vagrant-common.rb/tweak_docker_daemon too.

[done](https://github.com/weaveworks/weave/commit/f08148cabbd103c1f8112bd0b426c75924fea683)

PR #1955, opened 2016-02-04
'attach-bridge' feature to link an existing bridge netdev with the weave bridge

awh:
In order to continue progressing #1940 in an incremental fashion, we need to reimplement `create_bridge` in the weave script with weaveutil. Unfortunately `weave create-bridge`, intended for use when docker _isn't_ running (so that users can configure docker to use weave's bridge), relies on this function; consequently we need to provide another way for such users to continue using weave for L2 bridging between docker bridges. This PR implements `weave attach-bridge` for linking an existing `docker0` bridge netdev with the weave bridge, obtaining the same functionality; furthermore, it has the advantage of working with fast datapath, which the previous `create-bridge` method did not.

### DIFF
--- a/weave
+++ b/weave
@@ -492,6 +492,12 @@ detect_bridge_type() {
         # No bridge/datapath devices configured
         return 1
     fi
+
+    # WEAVE_MTU may have been specified when the bridge was
+    # created (perhaps implicitly with WEAVE_NO_FASTDP).  So take
+    # the MTU from the bridge unless it is explicitly specified
+    # for this invocation.
+    MTU=${WEAVE_MTU:-$(cat /sys/class/net/$BRIDGE/mtu)}
 }
 
 create_bridge() {
@@ -559,12 +565,6 @@ EOF
                 return 1
             fi
         fi
-
-        # WEAVE_MTU may have been specified when the bridge was
-        # created (perhaps implicitly with WEAVE_NO_FASTDP).  So take
-        # the MTU from the bridge unless it is explicitly specified
-        # for this invocation.
-        MTU=${WEAVE_MTU:-$(cat /sys/class/net/$BRIDGE/mtu)}
     fi
 
     [ "$1" = "--without-ethtool" ] || ethtool_tx_off_$BRIDGE_TYPE $BRIDGE
@@ -795,6 +795,23 @@ add_iface_bridge() {
 
 add_iface_bridged_fastdp() {
     add_iface_bridge "$@"
+}
+
+attach_bridge() {
+    bridge="$1"
+    LOCAL_IFNAME=v${CONTAINER_IFNAME}bl$bridge
+    GUEST_IFNAME=v${CONTAINER_IFNAME}bg$bridge
+
+    ip link show $GUEST_IFNAME >/dev/null 2>&1 && return 0
+    if ! ip link add name $LOCAL_IFNAME mtu $MTU type veth peer name $GUEST_IFNAME mtu $MTU ||
+       ! ip link set $LOCAL_IFNAME up ||
+       ! add_iface_$BRIDGE_TYPE $LOCAL_IFNAME ||
+       ! ip link set $GUEST_IFNAME up
+       ! ip link set $GUEST_IFNAME master $bridge ; then
+        ip link del $GUEST_IFNAME >/dev/null 2>&1 || true
+        ip link del $LOCAL_IFNAME >/dev/null 2>&1 || true
+        return 1
+    fi
 }
 
 router_opts_fastdp() {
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -1795,18 +1814,40 @@ case "$COMMAND" in
     # intentionally undocumented since it assumes knowledge of weave
     # internals
     create-bridge)
+        cat 1>&2 <<EOF
+WARNING: 'weave create-bridge' now requires docker to be running, which
+means you can no longer use it to create the weave bridge upfront for use
+as the docker bridge. If you were trying to do so you can now do the following
+instead:
```
rade: "...as the docker bridge. Instead:"

weave
```diff
@@ -1795,18 +1814,40 @@ case "$COMMAND" in
...
+
+ * Let Docker create its default bridge 'docker0'
```
rade: "Launch docker"

weave
```diff
@@ -1795,18 +1814,40 @@ case "$COMMAND" in
...
+ * Launch weave
+ * Execute 'weave attach-bridge docker0' to link the docker and weave bridges
```
rade: no need for the "docker0"

weave
```diff
@@ -1795,18 +1814,40 @@ case "$COMMAND" in
...
+
+Note that 'weave attach-bridge' is compatible with fast datapath!
```
rade: isn't it "fast data path"?

weave
```diff
@@ -1795,18 +1814,40 @@ case "$COMMAND" in
...
             exit 1
         fi
         create_bridge --without-ethtool
         ;;
+    attach-bridge)
+        if detect_bridge_type ; then
+            attach_bridge ${1:-$DOCKER_BRIDGE}
+            run_iptables -t nat -I POSTROUTING -o $BRIDGE -j ACCEPT
+        else
+            echo "Weave bridge not found. Please do 'weave launch' and try again" >&2
```
rade: s/do/run

weave
```diff
@@ -1795,18 +1814,40 @@ case "$COMMAND" in
     # intentionally undocumented since it assumes knowledge of weave
     # internals
     create-bridge)
+        cat 1>&2 <<EOF
```
rade: That is a lot of output. Perhaps we should only show it when docker isn't running?

awh: Isn't the theory that no-one will be calling `create-bridge` while docker is running?

awh:
Addressed comments.

weave
rade: Hmm. So this PR doesn't actually make it impossible to call `create-bridge` when docker isn't running. So I suggest we do issue a deprecation warning instead:

"WARNING: `weave create-bridge` is deprecated. Instead use `weave attach-bridge` to link the docker and weave bridges after both docker and weave have been started. Note that, unlike `weave create-bridge`, `weave attach-bridge is compatible with fast data path."

(modulo formatting)

rade:
Is `weave attach-bridge` idempotent? It should be.

weave
```diff
@@ -797,6 +797,25 @@ add_iface_bridged_fastdp() {
...
+    ip link add name $LOCAL_IFNAME mtu $MTU type veth peer name $GUEST_IFNAME mtu $MTU || return 1
+
+    if ! ip link set $LOCAL_IFNAME up ||
+       ! ip link set $GUEST_IFNAME up ||
+       ! add_iface_$BRIDGE_TYPE $LOCAL_IFNAME ||
+       ! ip link set $GUEST_IFNAME master $bridge ; then
+        ip link del $GUEST_IFNAME >/dev/null 2>&1 || true
+        return 1
+    fi
+}
```
rade: That code looks like it does the same as `setup_router_iface_bridge`, modulo veth pair naming.

awh: Good spot - will refactor.

rade:
I think it's worth up to ~1 hour of our time trying to write a test for this. e.g. check that two containers on different hosts can communicate, and also check the idempotence. Tricky bit is (re)starting docker with custom allocation ranges.

weave
awh: Unfortunately there is a subtle difference - `attach_bridge` also needs to `ip link set ... master ...` on the other end of the veth pair. The code doesn't factor out nicely as a consequence, so I've settled for structuring the code in the same way as `setup_router_iface_bridge`.

awh:
> Tricky bit is (re)starting docker with custom allocation ranges.

This will require #1747, which I have just spent some time investigating, however the best solution I can come up with depends on #1212. 

> I think it's worth up to ~1 hour of our time trying to write a test for this

Not going to happen, given the above.

awh:
Tested with `--fixed-cidr=172.17.1.0/24` and `--fixed-cidr=172.17.2.0/24` on different hosts.

PR #1958, opened 2016-02-05
Use netlink instead of shelling out to `weave attach-router`

### LINKED ISSUE #2112: weave w/o proxy fails to restart on reboot
```
$ weave launch-router
5e6af5ba959a03226eb80401cf3aba615bc38502b31e2b0ad16353e38c12d479
$ weave status

        Version: git-f649d9bed3d0

        Service: router
       Protocol: weave 1..2
           Name: 66:03:80:3a:36:92(host1)
     Encryption: disabled
  PeerDiscovery: enabled
        Targets: 0
    Connections: 0
          Peers: 1
 TrustedSubnets: none

        Service: ipam
         Status: idle
          Range: 10.32.0.0-10.47.255.255
  DefaultSubnet: 10.32.0.0/12

        Service: dns
         Domain: weave.local.
       Upstream: 10.0.2.3
            TTL: 1
        Entries: 0

$ sudo reboot
...

$ weave status

weave container is restarting.

$ docker logs weave |& tail -2
INFO: 2016/03/30 10:30:23.329385 Command line options: map[ipalloc-range:10.32.0.0/12 name:66:03:80:3a:36:92 nickname:host1 port:6783 datapath:datapath dns-effective-listen-address:172.17.0.1 dns-listen-address:172.17.0.1:53 http-addr:127.0.0.1:6784]
FATA: 2016/03/30 10:30:23.329662 netlink error response: no such device
```

This happens because we rely on the proxy to create the bridge etc.

It is rather unexpected.

Until we fix this properly, we should at least note in the docs that for restarts to work on reboot you must have started the proxy. See also #2108.

### LINKED ISSUE #2900: ensure uniqueness of weave bridge MAC when machine IDs are not unique
Currently, if some kind of machine ID is available, we generate a MAC from that and set it on the weave bridge.

We also use this value as the peer name, except where `--name` has been used to override.

This means, in cases where the machine ID is the same across hosts, we have identical bridge MACs and no way to override.

Either adding a `--mac`  option or using `--name` to set the weave bridge MAC would give users a way out of that problem, pending their machine provider making the IDs unique.

awh:
Sub-part of #1940
Fixes #2112
Fixes #2900, because we now set the bridge MAC from --name


rade:
'common' seems like an odd package to put these in. 'net'?

awh:
> 'common' seems like an odd package to put these in.

I couldn't agree more - see https://github.com/weaveworks/weave/pull/1944#issuecomment-179126321

>  'net'?

Giving packages the same name as the standard library strikes me as a Bad Idea (particularly when there's a high chance we want to use things from their - it forces the use of aliasing). I plan to refactor the packages once we can see the wood for the trees...

bboreham:
I think I will pause here for a bit.

Besides the tests failing, still to do are the no-bridged-fastdp initialisation, and we should move `destroy_bridge` into Go too.

bboreham:
BTW, `linkSetUpByName` seems unnecessary because the netlink library we are using will look up by name when you don't know the index.

awh:
> the netlink library we are using will look up by name when you don't know the index.

The atrocious godoc formatting had obscured this fact from me when I wrote it originally :)

bboreham:
Rebased; will need more work to integrate AWSVPC

common/bridge.go
```diff
@@ -175,8 +179,64 @@ func initBridgedFastdp(config *BridgeConfig) error {
 	return nil
 }
 
+// Add a rule to iptables, if it doesn't exist already
+func addIPTablesRule(ipt *iptables.IPTables, table, chain string, rulespec ...string) error {
```
awh: There's an `ipt.AppendUnique` function that does exactly this already

bboreham:
Fixes #2900, because we now set the bridge MAC from `--name`

net/bridge.go
```diff
@@ -108,3 +174,304 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+			return err
+		}
+		// and allow replies back
+		if err = ipt.AppendUnique("filter", "FORWARD", "-o", config.WeaveBridgeName, "-m", "conntrack", "--ctstate", "RELATED,ESTABLISHED", "-j", "ACCEPT"); err != nil {
+			return err
+		}
+	}
+
+	// create a chain for masquerading
+	ipt.NewChain("nat", "WEAVE") // ignoring errors such as 'chain already exists'
```
brb: Could we use `ipt.ClearChain` which:

```
// ClearChain flushed (deletes all rules) in the specified table/chain.
// If the chain does not exist, a new one will be created
```

?

net/bridge.go
```diff
@@ -108,3 +174,304 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+		}
+
+		dockerBridgeIP, err := FindBridgeIP(config.DockerBridgeName, nil)
+		if err != nil {
+			return err
+		}
+
+		// forbid traffic to the Weave port from other containers
+		if err = ipt.AppendUnique("filter", "INPUT", "-i", config.DockerBridgeName, "-p", "tcp", "--dst", dockerBridgeIP.String(), "--dport", fmt.Sprint(config.Port), "-j", "DROP"); err != nil {
+			return err
```
bboreham: Agreed with the point; an idea I had was to make the lower-level iptables library include the args that failed in its error return, then we don't have to write N wrap calls that repeat this info.

prog/weaveutil/bridge.go
```diff
@@ -0,0 +1,47 @@
...
+		MTU:              mtu,
+		Port:             port,
+		Mac:              args[5],
+		NoFastdp:         args[6] != "",
+		NoBridgedFastdp:  args[7] != "",
+		ExpectNPC:        args[9] == "--expect-npc",
+	}
+	procPath := args[8]
+	bridgeType, err := weavenet.CreateBridge(procPath, &config)
+	fmt.Println(bridgeType.String())
```
bboreham: This relates back to "in the weave script we log the corresponding error value to stderr"

bboreham: No it doesn't

net/arp.go
```diff
@@ -1,27 +1,28 @@
 package net
 
+import "path/filepath"
```
bboreham: fixed

net/bridge.go
```diff
@@ -2,25 +2,54 @@ package net
...
+	"github.com/pkg/errors"
 	"github.com/vishvananda/netlink"
+
+	"github.com/weaveworks/weave/common/odp"
 )
 
 type BridgeType int
 
 const (
 	WeaveBridgeName = "weave"
```
bboreham: added ASCII diagram; please let me know if it makes sense

net/bridge.go
```diff
@@ -34,11 +63,48 @@ func DetectBridgeType(weaveBridgeName, datapathName string) BridgeType {
...
+
+		mac, err := RandomMAC()
+		if err != nil {
+			return errors.Wrap(err, "creating random MAC")
+		}
+
+		if err := netlink.LinkSetHardwareAddr(link, mac); err != nil {
+			return errors.Wrapf(err, "setting bridge %s address to %v", bridgeName, mac)
+		}
+		return ErrSetBridgeMac
```
bboreham: changed to return a tuple (whether address was changed, error)

net/bridge.go
```diff
@@ -108,3 +174,304 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
 	return netdev.CIDRs[0].IP, nil
 }
+
+type BridgeConfig struct {
+	DockerBridgeName string
+	WeaveBridgeName  string
+	DatapathName     string
+	NoFastdp         bool
+	NoBridgedFastdp  bool
+	IsAWSVPC         bool
```
bboreham: changed

net/bridge.go
```diff
@@ -108,3 +174,304 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+			return bridgeType, errors.Wrap(err, "configuring iptables")
+		}
+	}
+
+	if config.IsAWSVPC {
+		bridgeType = AWSVPC
+		// Set proxy_arp on the bridge, so that it could accept packets destined
+		// to containers within the same subnet but running on remote hosts.
+		// Without it, exact routes on each container are required.
+		if err := sysctl(procPath, "net/ipv4/conf/"+config.WeaveBridgeName+"/proxy_arp", "1"); err != nil {
```
bboreham: I think there are a lot of these inside and outside this PR; better to do them all in one go?

net/bridge.go
```diff
@@ -108,3 +174,304 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+}
+
+func initBridgedFastdp(config *BridgeConfig) error {
+	if err := initFastdp(config); err != nil {
+		return err
+	}
+	if err := initBridgePrep(config); err != nil {
+		return err
+	}
+	if _, err := CreateAndAttachVeth("vethwe-bridge", "vethwe-datapath", config.WeaveBridgeName, config.MTU, true, func(veth netlink.Link) error {
```
bboreham: changed

net/bridge.go
bboreham: done

prog/weaveutil/bridge.go
bboreham: fixed

net/bridge.go
```diff
@@ -108,3 +174,304 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+		if err != nil {
+			return None, err
+		}
+
+		if err = configureIPTables(config); err != nil {
+			return bridgeType, errors.Wrap(err, "configuring iptables")
+		}
+	}
+
+	if config.IsAWSVPC {
```
bboreham: The old `weave` script would do certain actions on `create_bridge` whether the bridge existed previously or not.  This is one of them.  So moving the code into a place which is only executed to create the bridge changes that behaviour.

Question is, do we need to replicate the old behaviour?

bboreham: Yet more subtleties: AWSVPC is not a distinct type; you can have AWSVPC mode in effect when the underlying bridge is either BridgedFastdp or Bridge.
This capability is not currently useful, but if we did have a hybrid overlay/underlay implementation it would be.

net/bridge.go
```diff
@@ -108,3 +174,304 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+			}
+			if !odpSupported {
+				bridgeType = Bridge
+			}
+		}
+
+		var err error
+		switch bridgeType {
+		case Bridge:
+			err = initBridge(config)
```
bboreham: Created a `Bridge` interface; ptal

weave
```diff
@@ -481,102 +414,50 @@ detect_bridge_type() {
...
-        # and allow replies back
-        add_iptables_rule filter FORWARD -o $BRIDGE -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-
-        # create a chain for masquerading
-        run_iptables -t nat -N WEAVE >/dev/null 2>&1 || true
-        add_iptables_rule nat POSTROUTING -j WEAVE
-    else
-        if [ -n "$LAUNCHING_ROUTER" ] ; then
-            if [ "$BRIDGE_TYPE" = bridge -a -z "$WEAVE_NO_FASTDP" ] &&
-                util_op check-datapath $DATAPATH 2>/dev/null ; then
```
bboreham: I don't think I should have removed this and put nothing in its place.

bboreham: fixed

common/odp/odp.go
```diff
@@ -89,3 +89,13 @@ func AddDatapathInterface(dpname string, ifname string) error {
 	_, err = dp.CreateVport(odp.NewNetdevVportSpec(ifname))
 	return err
 }
+
+func DatapathSupported() bool {
```
bboreham: This PR removed the previous way it was checked, which was creating and then removing a datapath.  I don't actually recall why that was removed, but it is currently related.
Maybe I should just put exactly the same code back?
Why was the shell-script code rewritten to Go code in 24d9a00f471f6c879c7197306748cab1bf310ab7 ?

net/bridge.go
```diff
@@ -2,43 +2,126 @@ package net
...
+	"github.com/coreos/go-iptables/iptables"
+	"github.com/j-keck/arping"
+	"github.com/pkg/errors"
 	"github.com/vishvananda/netlink"
+
+	"github.com/weaveworks/weave/common/odp"
 )
 
-type BridgeType int
+/* There are 3 modes of operation:
```
brb: I guess showing the `weaver` process in the doc is not that important (and might confuse a reader) as drawing the actual configuration for each setup. Maybe something along these lines:

```
There are thee configurations of bridges:

1. Bridge
                 +---------+
(container-veth)-+  weave  +-(vethwe-bridge)--(vethwe-pcap)
                 +---------+

Used by "router/pcap.go:Pcap" bridge. "vethwe-pcap" (end of veth pair) captures
and injects packets.

2. BridgedFastdp

                 +---------+                                    +------------+
(container-veth)-+  weave  +-(vethwe-bridge)--(vethwe-datapath)-+  datapath  +
                 +---------+                                    +------------+

Used by "router/fastdp.go:fastDatapathBridge" bridge. "datapath" (Open vSwitch
datapath) is connected to the "weave" bridge via veth pair.

3. Fastdp

                 +---------+
(container-veth)-+  weave  +
                 +---------+

"weave" is Open vSwitch datapath. Currently not in use due to missing conntrack
support in datapath of old kernel versions (https://github.com/weaveworks/weave/issues/1577).
```

marccarre: Nitpick: `1. Bridge (a.k.a. Sleeve)`

net/bridge.go
```diff
@@ -2,41 +2,132 @@ package net
...
+		return false, errors.Wrapf(err, "reading %q", sysctlFilename)
+	}
+
+	// From include/uapi/linux/netdevice.h
+	// #define NET_ADDR_PERM       0   /* address is permanent (default) */
+	// #define NET_ADDR_RANDOM     1   /* address is generated randomly */
+	// #define NET_ADDR_STOLEN     2   /* address is stolen from other device */
+	// #define NET_ADDR_SET        3   /* address is set using dev_set_mac_address() */
+	// Note the file typically has a newline at the end, so we just look at the first char
+	if addrAssignType[0] != '3' {
```
marccarre: Although this is not something newly introduced in this PR, the reason *why* we set a random MAC even when we have `0` (`NET_ADDR_PERM`) and `2` (`NET_ADDR_STOLEN`) is unclear. I guess this therefore calls for a comment?

N.B.: I am still researching this and reading [this blogpost](https://www.weave.works/anatomy-of-a-docker-networking-heisenbug/).

bboreham: `2` is the case we most want to fix - when the mac address randomly changes as veths attach and detach.

I cannot think why we want to change it for `0` or `1`; maybe those just don't crop up on a bridge?

net/bridge.go
```diff
@@ -108,3 +199,307 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+
+func CreateBridge(procPath string, config *BridgeConfig) (Bridge, error) {
+	bridgeType, err := DetectBridgeType(config.WeaveBridgeName, config.DatapathName)
+	if err != nil {
+		return nil, err
+	}
+
+	if bridgeType == nil {
+		bridgeType = bridgeImpl{}
+		if !config.NoFastdp {
```
marccarre: Nitpick: double negation, decreases readability.

net/bridge.go
```diff
@@ -108,3 +199,307 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+}
+
+func CreateBridge(procPath string, config *BridgeConfig) (Bridge, error) {
+	bridgeType, err := DetectBridgeType(config.WeaveBridgeName, config.DatapathName)
+	if err != nil {
+		return nil, err
+	}
+
+	if bridgeType == nil {
+		bridgeType = bridgeImpl{}
```
marccarre: This function is relatively hard to reason about:
- we do more than "one thing", e.g. create bridges, modify configuration, create datapath, handle special cases related to AWSVPC, configure `iptables` rules, etc.
- the succession of `bridgeType` overrides matches `try_create_bridge`'s logic, but is relatively hard to follow

Suggestion: 
- have simple conditions to choose the bridge implementation
- give more responsibilities to the bridge implementation
- split this function in smaller parts, each responsible for one concern, or delegate to the bridge implementation (e.g. would it make sense to have the creation of the datapath happening as part of the instantiation of the bridge impl.?)

Essentially, L222-244 do this (approximate refactoring, double-negation removed), which could be easier to follow:
```
func (...) defaultBridgeType(...) (...) {
    if config.Fastdp && odp.IsSupported() {
        if config.BridgedFastdp {
            return bridgedFastdpImpl{datapathName: config.DatapathName}, nil
        } else {
            return fastdpImpl{datapathName: config.WeaveBridgeName}, nil
        }
    } else {
        return bridgeImpl{}, nil
    }
}
[...]
if bridgeType == nil {
    bridgeType = (...).defaultBridgeType(...)
    if err := bridgeType.init(config); err != nil {
        return nil, err
    }
    if err := configureIPTables(config); err != nil {
        return bridgeType, errors.Wrap(err, "configuring iptables")
    }
}
[...]
```

net/bridge.go
```diff
@@ -108,3 +199,307 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+	mac, err := net.ParseMAC(config.Mac)
+	if err != nil {
+		return errors.Wrapf(err, "parsing bridge MAC %q", config.Mac)
+	}
+
+	linkAttrs := netlink.NewLinkAttrs()
+	linkAttrs.Name = config.WeaveBridgeName
+	linkAttrs.HardwareAddr = mac
+	if config.MTU == 0 {
+		config.MTU = 65535
```
marccarre: 1. What are the consequences of MTU being equal to `0`?
2. Why do we default to `65535` specifically? Is this the maximum MTU?
3. What will happen when this is done? Will the network try to find the optimal value using ICMP?

bboreham: 1. 0 just means nobody has set it yet.
2. Yes, packet length is a 16-bit value.
3. Sleeve has its own PMTU discovery, and will fragment packets that are too big to transport on the underlying network. However the PMTU value is per path, and we only get to set one MTU on the bridge and on all veths, for all paths, so we go for the biggest possible value.
Note this code is only hit when not using fastdp.  Fastdp cannot deal with fragments, so we have to set the MTU up front, small enough to work on all paths.

marccarre: 3. I guess the dummy interface below lowers it down automatically then...

bboreham: The purpose of the dummy interface is to set a specific value, because just setting it doesn't work, on a bridge.

net/bridge.go
```diff
@@ -108,3 +199,307 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+		   which is needed for the vxlan encryption. */
+		config.MTU = 1376
+	}
+	if err := netlink.LinkSetMTU(datapath, config.MTU); err != nil {
+		return errors.Wrapf(err, "setting datapath %q mtu %d", config.DatapathName, config.MTU)
+	}
+	return nil
+}
+
+func (bf bridgedFastdpImpl) init(config *BridgeConfig) error {
```
marccarre: Makes sense and easy to match to the above diagram :+1:

net/bridge.go
```diff
@@ -108,3 +199,307 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+		return errors.Wrap(err, "setting dummy interface master")
+	}
+	if err := netlink.LinkDel(dummy); err != nil {
+		return errors.Wrap(err, "deleting dummy interface")
+	}
+
+	return nil
+}
+
+func (b bridgeImpl) init(config *BridgeConfig) error {
```
marccarre: Makes sense and easy to match to the above diagram :+1:

net/bridge.go
```diff
@@ -108,3 +199,307 @@ func FindBridgeIP(bridgeName string, subnet *net.IPNet) (net.IP, error) {
...
+func (f fastdpImpl) attach(veth *netlink.Veth) error {
+	return odp.AddDatapathInterface(f.datapathName, veth.Attrs().Name)
+}
+
+func configureIPTables(config *BridgeConfig) error {
+	ipt, err := iptables.New()
+	if err != nil {
+		return errors.Wrap(err, "creating iptables object")
+	}
+	if config.DockerBridgeName != "" {
```
marccarre: 1. Is the case of `config.DockerBridgeName == ""` actually an error case? 
    (given `cmdUsage("create-bridge", "<docker-bridge-name> [...]`)
2. If not, could this mean potential networking / security issues?

marccarre: 1. Is the case of `config.DockerBridgeName == ""` actually an error case? 
    (given `cmdUsage("create-bridge", "<docker-bridge-name> [...]`)
2. If not, could this mean potential networking / security issues?

N.B.: still trying to make sense out of the all the rules. I guess this suggest more comments or wrapping some in well named functions could help explain why we need all of these.

bboreham: We allow Weave Net to be run without connecting to Docker, e.g. from `weave-kube` or (potentially) with Rkt or Mesos.  This is the intended use with a blank bridge name.
There is no security problem when run as expected - containers never connect to the Docker bridge in any of the above cases. However it is possible that a creative user could give a set of options that open the security hole mentioned.

marccarre: Thanks for the explanation.

prog/weaver/main.go
```diff
@@ -453,34 +491,32 @@ func (nopPacketLogging) LogPacket(string, weave.PacketKey) {
 func (nopPacketLogging) LogForwardPacket(string, weave.ForwardPacketKey) {
 }
 
-func createOverlay(datapathName string, ifaceName string, isAWSVPC bool, host string, port int, bufSzMB int, enableEncryption bool) (weave.NetworkOverlay, weave.Bridge) {
+func createOverlay(bridgeType weavenet.Bridge, config weavenet.BridgeConfig, host string, port int, bufSzMB int, enableEncryption bool) (weave.NetworkOverlay, weave.Bridge) {
```
marccarre: Would having this as a method on `weavenet.Bridge` be more appropriate?
I am asking because of `case bridgeType.IsFastdp()` / `case !bridgeType.IsFastdp()`, which seem to hint for this.

prog/weaver/main.go
```diff
@@ -453,34 +491,32 @@ func (nopPacketLogging) LogPacket(string, weave.PacketKey) {
...
 		vpc := weave.NewAWSVPC()
 		overlay.Add("awsvpc", vpc)
 		bridge = weave.NullBridge{}
 		// Currently, we do not support any overlay with AWSVPC
 		ignoreSleeve = true
-	case datapathName != "" && ifaceName != "":
-		Log.Fatal("At most one of --datapath and --iface must be specified.")
-	case datapathName != "":
-		iface, err := weavenet.EnsureInterface(datapathName)
+	case bridgeType == nil:
```
marccarre: This seems to only happen if there was an error in `CreateBridge`, is it still valid to continue despite having such an error and `NullBridge` being a "null object" (i.e. doing nothing)?

prog/weaver/main.go
bboreham: `net` is supposed to be lower-level than `router`, so we can't have a routine in `net` creating objects in `router`.

It is possible that there should be some movement from `net` to `router`...

marccarre: > It is possible that there should be some movement from `net` to `router`...
Ah yes, I was just looking into this, as the two `Bridge` interface seem to overlap, semantically speaking.

prog/weaver/main.go
bboreham: I think that's an unintentional change in behaviour: previously it was possible to run `weaver` with no bridge or datapath.  For example just to act as a DNS server.

PR #1964, opened 2016-02-09
Fix attach-bridge bugs

awh:
#1955 introduced *three* bugs:
* `weave reset` did not clean up veth pairs created by `weave attach-bridge`
* `weave reset` did not clean up iptables rules inserted by `weave attach-bridge`
* `weave attach-bridge` was not idempotent in terms of iptables rules creation


awh:
Is it worth adding an `insert_iptables_rule` helper?

rade:
> Is it worth adding an `insert_iptables_rule` helper?

yes.

rade:
> Is it worth adding an `insert_iptables_rule` helper?

Done.

PR #1966, opened 2016-02-09
Make weaveutil container-addrs observe $PROCFS

### LINKED ISSUE #1965: various `weave --local` functions are broken
$ sudo weave --local ps
    weave:expose 26:09:9e:a0:d6:a5
    no such file or directory

because it is running

    docker run --rm --privileged --net=host -v /var/run/docker.sock:/var/run/docker.sock -v /proc:/hostproc -e PROCFS=/hostproc --entrypoint=/usr/bin/weaveutil weaveworks/weaveexec:latest container-addrs /proc weave weave:expose f438f9a36b72

because `$PROCFS` is `/proc`, correct for non`--local` case and `--local` with `weaveutil` accessible, but incorrect for `--local` which has to run `weaveutil` via Docker

awh:
Fix #1965.

### DIFF
--- a/prog/weaveutil/addrs.go
+++ b/prog/weaveutil/addrs.go
@@ -7,19 +7,19 @@ import (
 )
 
 func containerAddrs(args []string) error {
-	if len(args) < 2 {
-		cmdUsage("container-addrs", "<procPath> <bridgeName> [containerID ...]")
+	if len(args) < 1 {
+		cmdUsage("container-addrs", "<bridgeName> [containerID ...]")
 	}
+	bridgeName := args[0]
 
-	procPath := args[0]
-	bridgeName := args[1]
+	procPath := procDir()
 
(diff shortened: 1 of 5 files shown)


prog/weaveutil/addrs.go
```diff
@@ -4,22 +4,27 @@ import (
 	"fmt"
 	"github.com/fsouza/go-dockerclient"
 	"github.com/weaveworks/weave/common"
+	"os"
 )
 
 func containerAddrs(args []string) error {
 	if len(args) < 2 {
-		cmdUsage("container-addrs", "<procPath> <bridgeName> [containerID ...]")
+		cmdUsage("container-addrs", "<bridgeName> [containerID ...]")
```
rade: you forgot to update the arg count check accordingly. Will fix.

rade:
did some flyby refactoring here - we were determining procfs in two weaveutil subcommands.

PR #1967, opened 2016-02-09
Start existing container, if its image and arguments match

### LINKED ISSUE #1859: make `weave launch` idempotent and start existing containers
Currently, `weave launch` always does a `docker rm` on any previous containers.

If, instead, we did `docker start` on that container, this would fix #1757, and give a straightforward way to fix #901 and #678, by persisting the state into a file inside the container.

Users who wanted to change the launch options would need to do something to make that happen; perhaps `weave reset` before `weave launch-xxx`.

We might want to worry about log-file size too, since Docker carries on the same log-file after a stop/start.

bboreham:
Fixes #1859

[Review on Reviewable](https://reviewable.io/reviews/weaveworks/weave/1967)

### DIFF
--- a/weave
+++ b/weave
@@ -952,10 +952,12 @@ start_if_necessary() {
     name=$2
     image=$3
     shift 3
-    RUN_STATUS=$(docker inspect --format='{{.State.Running}} {{.Config.Image}} {{.Id}}' $name 2>/dev/null) || true
+    image_base=${image%:*}
+    IMAGE_ID=$(docker inspect --format='{{.Id}}' $image)
+    RUN_STATUS=$(docker inspect --format='{{.State.Running}} {{.Image}} {{.Config.Image}} {{.Id}}' $name 2>/dev/null) || true
     CONTAINER_ID=${RUN_STATUS##* }
     case ${RUN_STATUS%:*} in
-        "true $image")
+        "true $IMAGE_ID "*)
             if args_match $CONTAINER_ID "$@" ; then
                 ALREADY_RUNNING_CONTAINER=$CONTAINER_ID
                 echo $CONTAINER_ID
@@ -964,13 +966,17 @@ start_if_necessary() {
                 return 1
             fi
             ;;
-        "false $image")
+        "false $IMAGE_ID "*)
             if args_match $CONTAINER_ID "$@" ; then
                 docker start $CONTAINER_ID
             else
                 docker rm $name >/dev/null
                 docker run -d --name=$name "$@"
             fi
+            ;;
+        true*$image_base)
+            echo "Found another version of $name running. Aborting." >&2
+            return 2
             ;;
         true*)
             echo "Found another running container named '$name'. Aborting." >&2
@@ -1615,7 +1621,7 @@ launch_router() {
     # Set WEAVE_DOCKER_ARGS in the environment in order to supply
     # additional parameters, such as resource limits, to docker
     # when launching the weave container.
-    start_if_necessary router $CONTAINER_NAME $BASE_IMAGE \
+    start_if_necessary router $CONTAINER_NAME $IMAGE \
         --privileged --net=host $(docker_sock_options) \
         -e WEAVE_PASSWORD \
         -e CHECKPOINT_DISABLE \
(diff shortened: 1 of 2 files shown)


bboreham:
There is a slight change in behaviour from removing the lines like `echo $ROUTER_CONTAINER` and moving that functionality earlier, in some cases to the `docker start` that generates the ID: the user will see the ID in cases which subsequently error, compared to previously where they would only see the ID if nothing went wrong.

This change was made because doing the start in a subshell (i.e. `$()`) breaks `set -e`, and it became ugly to either test error codes all over the place or to assign some `CONTAINER_ID` variable and propagate it up.

weave
```diff
@@ -1590,23 +1621,30 @@ launch_router() {
         fi
     fi
 
+    # Generate a hash of the password so we can tell if it has changed.
+    # Salt the hash with the peer name to make it even harder to reverse
+    [ -n "$WEAVE_PASSWORD" ] && PASSWORD_HASH="--password-hash=$(echo $PEERNAME$WEAVE_PASSWORD | sha256sum | cut -c1-32)"
```
bboreham: @awh do you have a view on this, security-wise?

rade:
I reckon we should prune `password-hash` from the arg logging in weave `main()`.

weave
```diff
@@ -1590,23 +1621,30 @@ launch_router() {
...
+
     # Set WEAVE_DOCKER_ARGS in the environment in order to supply
     # additional parameters, such as resource limits, to docker
     # when launching the weave container.
-    ROUTER_CONTAINER=$(docker run --privileged --net=host -d --name=$CONTAINER_NAME \
+    start_if_necessary router $CONTAINER_NAME $IMAGE \
+        --privileged --net=host \
```
rade: we could move the `--privileged --net=host` into `start_if_necessary`, or, rather, a `start_container` function called by `start_if_necessary`.

weave
```diff
@@ -1590,23 +1621,30 @@ launch_router() {
...
 
+    # Generate a hash of the password so we can tell if it has changed.
+    # Salt the hash with the peer name to make it even harder to reverse
+    [ -n "$WEAVE_PASSWORD" ] && PASSWORD_HASH="--password-hash=$(echo $PEERNAME$WEAVE_PASSWORD | sha256sum | cut -c1-32)"
+
     # Set WEAVE_DOCKER_ARGS in the environment in order to supply
     # additional parameters, such as resource limits, to docker
     # when launching the weave container.
-    ROUTER_CONTAINER=$(docker run --privileged --net=host -d --name=$CONTAINER_NAME \
+    start_if_necessary router $CONTAINER_NAME $IMAGE \
```
rade: It's a bit unfortunate that the image gets passed twice. Alas I cannot think of a neat way around that - we could stash all the run args into one arg, but that complicates the call sites.

weave
awh: * Is the hash truncation with `cut` accidental? You need `-c1-64` to get the whole thing
* Agree with @rade that this value should not appear in logs
* Not convinced that `--password-hash` is a good name for the router arg, even though it is undocumented - `--ignored-value`? `--opaque-identity`?

rade:
Why don't we simply use
```
$ docker inspect -f '{{.Config.Env}}' weave
[WEAVE_PASSWORD=foo WEAVE_CIDR=none PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin]
```
to extract the password of an existing router container?

awh:
I am not convinced that we should throw everything away when the launch peer list changes, as will happen when growing or shrinking a cluster.

rade:
> I am not convinced that we should throw everything away when the launch peer list changes, as will happen when growing or shrinking a cluster.

What should happen instead? And what should happen if the peer list got modified with `weave connect|disconnect`? Related, and irrespective of whether we treat change in launch peers specially, how should the initial peer count be determined? Should it be derived from the current list of targets, or the list of launch peers?

This all smells rather bad. Perhaps we should take a step back and revisit the CLI commands. The whole interaction between specified launch peers, specified init-peer-count and connect/disconnect is far from obvious.

rade:
In the interest of not turning this PR into a major redesign of the CLI, I would suggest we stick with the safe, and easily explained option. Which is "if any args are different, throw everything away".

weave
awh: @bboreham ping

awh:
> In the interest of not turning this PR into a major redesign of the CLI, I would suggest we stick with the safe, and easily explained option.

OK - we do need to do this though:

> Perhaps we should take a step back and revisit the CLI commands. The whole interaction between specified launch peers, specified init-peer-count and connect/disconnect is far from obvious.

Reassigning back to @bryan as there are a couple of comments outstanding nonetheless:

https://github.com/weaveworks/weave/pull/1967#discussion_r52900841
https://github.com/weaveworks/weave/pull/1967#issuecomment-184224893

weave
bboreham: The truncation was intentional; some balance between making the command-line unreadable and the probability of accidental clash.  We even discussed making it shorter to make it harder to reverse.  (There is no 'birthday problem' here)

bboreham:
> Why don't we simply ... extract the password 

Or, more generally, compare the entire list of passed-in environment variable.

PR #1968, opened 2016-02-09
"refactor" veth creation

rade:
Invoke the new create_veth from a few places, eliminating code duplication and inconsistencies.

Semantic differences:

- init_bridged_fastdp now skips the veth creation/config if the veth
  already exists. This eliminates routing blips. Though in reality
  something very strange must have happened to end up in this code
  with an existing veth.
- init_bridged_fastdp no longer deletes the $BRIDGE and $DATAPATH when
  veth creation/configuration fails. This is consistent with
  init_bridge, which does not delete $BRIDGE when something goes
  wrong. And it is generally desirable since the init_* functions are
  invoked on potentially existing bridges - e.g. following 'weave
  stop'.
- attach_bridge and setup_router_iface_bridge used to mark one veth
  end 'up', then invoke add_iface_* for it, and then mark the other
  end 'up'. create_veth marks both ends 'up' first. This makes no
  practical difference in these cases.

notable other differences:
- we detect veth presence by checking whether *both* ends are present,
  rather than just one. We deliberately continue to (and fail in) the
  veth creation if one end already exists. This can only happen when
  we've stumbled across a different interface, since a veth end cannot
  exist independently. So it is right to treat this as an error.
- init_bridged_fastdp invokes add_iface_bridge instead of duplicating
  what that does. Also makes for nice coherence with the
  add_iface_fastdp it is invoking too.

### DIFF
--- a/weave
+++ b/weave
@@ -576,6 +576,24 @@ EOF
     configure_arp_cache $BRIDGE
 }
 
+# create veth with ends $1-$2, and then invoke $3..., removing the
+# veth on failure. No-op of veth already exists.
+create_veth() {
+    VETHL=$1
+    VETHR=$2
+    shift 2
+
+    ip link show $VETHL >/dev/null 2>&1 && ip link show $VETHR >/dev/null 2>&1 && return 0
+
+    ip link add name $VETHL mtu $MTU type veth peer name $VETHR mtu $MTU
+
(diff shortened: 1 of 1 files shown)


rade:
@awh says "merge it"

PR #1971, opened 2016-02-12
Persist IPAM ring, take 2

bboreham:
Revised version of #1928, with BoltDB specifics pulled out to a separate package.

We have a very simple persistence interface, just `Load` and `Save` an `interface{}`, with serialization (Gob) hidden behind the abstraction.

Partial fix of #678

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -55,3 +55,6 @@
 [submodule "vendor/github.com/vishvananda/netns"]
 	path = vendor/github.com/vishvananda/netns
 	url = https://github.com/vishvananda/netns.git
+[submodule "vendor/github.com/boltdb/bolt"]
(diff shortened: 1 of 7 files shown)


rade:
It seems odd that the key is supplied as a []byte rather than a string.

awh:
> It seems odd that the key is supplied as a []byte rather than a string.

Perhaps this is some artefact of the implementation (mmap'ed B+Tree) that is leaking out for efficiency purposes?

awh:
>> It seems odd that the key is supplied as a []byte rather than a string.

> Perhaps this is some artefact of the implementation (mmap'ed B+Tree) that is leaking out for efficiency purposes?

D'oh I thought you were talking about the BoltDB API :smile:

PR #1972, opened 2016-02-12
Persist set of peers to connect to

bboreham:
Building on #1971 

We had some discussion about whether it's a good idea to overwrite the command-line parameters with the last saved data.

[Review on Reviewable](https://reviewable.io/reviews/weaveworks/weave/1972)

### DIFF
--- a/router/network_router.go
+++ b/router/network_router.go
@@ -217,11 +217,40 @@ func (router *NetworkRouter) persistPeers() {
 	}
 }
 
-func (router *NetworkRouter) SetInitialPeers(peers []string) []error {
-	// Load is a no-op if persistence entry has never been saved
-	if err := router.db.Load(peersIdent, &peers); err != nil {
-		return []error{err}
+func (router *NetworkRouter) InitiateConnections(peers []string, replace bool) []error {
+	errors := router.ConnectionMaker.InitiateConnections(peers, replace)
+	router.persistPeers()
+	return errors
+}
+
+func (router *NetworkRouter) ForgetConnections(peers []string) {
+	router.ConnectionMaker.ForgetConnections(peers)
+	router.persistPeers()
+}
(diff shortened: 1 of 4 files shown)


awh:
> We had some discussion about whether it's a good idea to overwrite the command-line parameters with the last saved data.

In spite of the fact that in either the pre/post #1967 merge case there will be no persisted values available if the router is relaunched with different arguments, I would still stick an info logging statement on successful load to the effect that the CLI list has been ignored in favour of persisted values - just in case someone uses `--db-prefix`, or we change the behaviour ourselves later...

bboreham:
I added a test to see if the stored value differed from the command-line value.

Also changed the computation of quorum to use the same set as we use for initial connections, which necessitated moving the initialization of `Allocator` after the router.  This seems innocuous as `Allocator` will block trying to send gossip via the router anyway, if you try to use it.

Also I noticed it was only persisting on connect/forget, so I added an initial call to persist.

router/network_router.go
```diff
@@ -204,3 +206,41 @@ func (router *NetworkRouter) relayBroadcast(srcPeer *mesh.Peer, key PacketKey) F
...
+}
+
+func (router *NetworkRouter) SetInitialPeers(peers []string) ([]string, []error) {
+	var storedPeers []string
+	if err := router.db.Load(peersIdent, &storedPeers); err != nil {
+		return nil, []error{err}
+	}
+
+	if !equal(peers, storedPeers) {
+		log.Println("Overriding command-line peer list with stored list:", peers)
```
awh: You're missing the assignment of storedPeers to peers in the event they are unequal

bboreham:
I think I covered the point about logging the override.

sheerun:
This is huge, thank you!

PR #1945, opened 2016-02-01
[WIP] persist IPAM addresses

bboreham:
Following on from #1928, persist the `owned` addresses. `space` can be derived from a combination of `owned` and `ring`.

Note the first 4 commits on this PR are from #1928; they will be rebased out when #1928 is merged.

`owned` is implemented directly in BoltDB's tree-map structure, while we use Gob to store the ring.  This results in a lot more code for the 'owned' operations, but less work at run-time as we allocate and free addresses.

Status prior to this PR:

item | `weave restart` | `docker restart` | `docker stop` followed by `docker start` a good while later | policy restart after reboot
------|----------------------|----------------------|------------------------------------------------------------------------------|------
IPAM addresses specified with `WEAVE_CIDR`        | retained | retained | re-allocated; may change | re-allocated; may change 
non-IPAM addresses specified with `WEAVE_CIDR` | retained | retained | retained | retained
IPAM addresses specified in `weave attach`               | retained | leaked | leaked | lost
non-IPAM addresses specified in `weave attach`        | retained | lost | lost | lost

The default if you don't ask for anything special is "IPAM address specified in `weave attach`"

Status after this PR (at time of writing):

item | `weave restart` | `docker restart` | `docker stop` followed by `docker start` a good while later | policy restart after reboot
------|----------------------|----------------------|------------------------------------------------------------------------------|------
IPAM addresses specified with `WEAVE_CIDR`        | retained | retained | re-allocated; may change | retained 
non-IPAM addresses specified with `WEAVE_CIDR` | retained | retained | retained | retained
IPAM addresses specified in `weave attach`               | retained | lost | lost | lost
non-IPAM addresses specified in `weave attach`        | retained | lost | lost | lost

Final target of this PR:

item | `weave restart` | `docker restart` | `docker stop` followed by `docker start` a good while later | policy restart after reboot
------|----------------------|----------------------|------------------------------------------------------------------------------|------
IPAM addresses specified with `WEAVE_CIDR`        | retained | retained | retained | retained 
non-IPAM addresses specified with `WEAVE_CIDR` | retained | retained | retained | retained
IPAM addresses specified in `weave attach`               | retained | retained | retained | retained
non-IPAM addresses specified in `weave attach`        | retained | retained | retained | retained

PR #1973, opened 2016-02-12
Persist 'owned' addresses in IPAM

bboreham:
New take on #1945, following on from #1971 

Here we are serializing the entire set every time it changes, which is overkill, but we don't expect the set to have more than 1,000 entries on one host and we don't expect it to change more than ~10 times per second.

[Review on Reviewable](https://reviewable.io/reviews/weaveworks/weave/1973)

### DIFF
--- a/test/163_docker_restart_test.sh
+++ b/test/163_docker_restart_test.sh
@@ -1,0 +1,49 @@
+#! /bin/bash
+
+. ./config.sh
+
+check_attached() {
+    for c in $@; do
+        assert_raises "exec_on $HOST1 $c $CHECK_ETHWE_UP"
+    done
+}
+
+wait_for_proxy() {
+    for i in $(seq 1 120); do
+        echo "Waiting for proxy to start"
+        if proxy docker_on $1 info > /dev/null 2>&1 ; then
+            return
+        fi
+        sleep 1
+    done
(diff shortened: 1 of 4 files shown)


bboreham:
I removed the test for addresses added with `weave attach`; it requires serious modification to proxy startup to be able to attach containers started outside of the proxy.  I think we can ship 1.5 without that feature.

test/163_docker_restart_test.sh
```diff
@@ -0,0 +1,49 @@
...
+        sleep 1
+    done
+    echo "Timed out waiting for proxy to start" >&2
+    exit 1
+}
+
+start_suite "Containers get same IP address on restart"
+
+# Remove any persisted data from previous runs
+run_on $HOST1 "sudo rm -f /tmp/test645-*"
```
awh: Should be 163

test/163_docker_restart_test.sh
```diff
@@ -0,0 +1,49 @@
...
+
+WEAVE_DOCKER_ARGS="-v /tmp:/db --restart=always" weave_on $HOST1 launch-router --db-prefix=/db/test645-
```
awh: Should be 163

bboreham:
I have rebased, fixed the test directory name, and moved half the commits out to #2012

PR #1983, opened 2016-02-18
don't complain on relaunch when ODP is not supported

### LINKED ISSUE #1929: Misleading error on second router launch on kernels with no ODP support
The first time you launch the router without specifying `WEAVE_NO_FASTDP` on such a kernel `create_bridge` silently falls back to creating a default bridge netdev for use with PCAP. On the second launch, the following error is received due to the extant bridge:
```
WEAVE_NO_FASTDP is not set, but there is already a bridge present of
the wrong type for fast datapath.  Please do 'weave reset' to remove
the bridge first.
```
This is confusing, because the first launch works without specifying `WEAVE_NO_FASTDP`. The initial launch should probably fail with a recommendation to try again with `WEAVE_NO_FASTDP`, so that the user knows from the outset that it needs to be supplied in future.

rade:
...and WEAVE_NO_FASTDP isn't set.

Fixes #1929.

### DIFF
--- a/weave
+++ b/weave
@@ -536,7 +536,9 @@ create_bridge() {
         add_iptables_rule nat POSTROUTING -j WEAVE
     else
         if [ -n "$LAUNCHING_ROUTER" ] ; then
-            if [ "$BRIDGE_TYPE" = bridge -a -z "$WEAVE_NO_FASTDP" ] ; then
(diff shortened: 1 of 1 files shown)


rade:
This isn't perfect, but hard to beat for conciseness and clarity, and IMO "Good Enough".

The travis failure is unrelated.

PR #1990, opened 2016-02-22
Support non-electing peers via --observer

### LINKED ISSUE #1743: restricted electorate to help initial consensus
Bootstrapping the data-structure used for IPAM communication between peers requires a consensus, to avoid multiple cliques forming independent data structures.  Currently we have a simple majority election - if you have 10 peers then 6 of them must vote to initialize - but this is unhelpful in situations where the number of peers may fluctuate.

As extension of this scheme, designate some peers as _electors_, and say others cannot vote.  Now it suffices to:
* designate a number of electors, say 3-5, for redundancy.
* have the electorate run an election at initial start-up, using the existing scheme.
* ensure that every other peer connects to at least one of the electors at start-up.

This scheme requires some additional administrative burden - the administrator must designate electors, and must keep them in the cluster.  But it retains the fundamental feature of being partition-tolerant once bootstrapped.

awh:
Initial PR to provoke design discussion on the following points:

1. I have used a `gossipState` ring update message where both `Ring` *and* `Paxos` are `nil` as a way for non-electors to poll their peers for a ring - are we happy with that, or would a separate message type be preferable
2. Most of the changes to `allocator.go` (various `alloc.paxos != nil` protections, introduction of `determinePaxos`) could be reverted if `paxos.Node` became an interface with electing and non-electing variants. The latter would implement null versions of `GossipState`/`Consensus`/`Propose`/`Update`/`Think` and `paxos.NewNode` could then be made to return the correct implementation based on the value of `quorum`
3. `--init-peer-count=0` isn't a great UI IMO; I would rather have something explicit like `--disenfranchise` that conflicts with the specification of `--init-peer-count`
4. `weave status`/`weave report` need to be updated; suggestions for how that should look are welcome

Thoughts @rade @dpw @bboreham?

Fixes #1743.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -45,24 +45,24 @@ type operation interface {
 // necessary plumbing.  Runs as a single-threaded Actor, so no locks
 // are used around data structures.
 type Allocator struct {
-	actionChan       chan<- func()
-	ourName          mesh.PeerName
-	universe         address.Range             // superset of all ranges
-	ring             *ring.Ring                // information on ranges owned by all peers
-	space            space.Space               // more detail on ranges owned by us
-	owned            map[string][]address.CIDR // who owns what addresses, indexed by container-ID
-	nicknames        map[mesh.PeerName]string  // so we can map nicknames for rmpeer
-	pendingAllocates []operation               // held until we get some free space
-	pendingClaims    []operation               // held until we know who owns the space
-	dead             map[string]time.Time      // containers we heard were dead, and when
-	db               db.DB                     // persistence
-	gossip           mesh.Gossip               // our link to the outside world for sending messages
-	paxos            *paxos.Node
-	paxosActive      bool
-	ticker           *time.Ticker
-	shuttingDown     bool // to avoid doing any requests while trying to shut down
-	isKnownPeer      func(mesh.PeerName) bool
-	now              func() time.Time
+	actionChan        chan<- func()
+	ourName           mesh.PeerName
+	universe          address.Range             // superset of all ranges
+	ring              *ring.Ring                // information on ranges owned by all peers
+	space             space.Space               // more detail on ranges owned by us
+	owned             map[string][]address.CIDR // who owns what addresses, indexed by container-ID
+	nicknames         map[mesh.PeerName]string  // so we can map nicknames for rmpeer
+	pendingAllocates  []operation               // held until we get some free space
+	pendingClaims     []operation               // held until we know who owns the space
+	dead              map[string]time.Time      // containers we heard were dead, and when
+	db                db.DB                     // persistence
+	gossip            mesh.Gossip               // our link to the outside world for sending messages
+	paxos             paxos.Participant
+	awaitingConsensus bool
+	ticker            *time.Ticker
+	shuttingDown      bool // to avoid doing any requests while trying to shut down
+	isKnownPeer       func(mesh.PeerName) bool
+	now               func() time.Time
 }
 
 // NewAllocator creates and initialises a new Allocator
@@ -73,7 +73,7 @@ func NewAllocator(ourName mesh.PeerName, ourUID mesh.PeerUID, ourNickname string
 		ring:        ring.New(universe.Start, universe.End, ourName),
 		owned:       make(map[string][]address.CIDR),
 		db:          db,
-		paxos:       paxos.NewNode(ourName, ourUID, quorum),
+		paxos:       paxos.NewParticipant(ourName, ourUID, quorum),
 		nicknames:   map[mesh.PeerName]string{ourName: ourNickname},
 		isKnownPeer: isKnownPeer,
 		dead:        make(map[string]time.Time),
@@ -472,6 +472,9 @@ func (alloc *Allocator) encode() []byte {
 	}
 
 	// We're only interested in Paxos until we have a Ring.
+	// Non-electing participants (e.g. observers) return
+	// a nil gossip state in order to provoke a unicast ring
+	// update from electing peers who have reached consensus.
 	if alloc.ring.Empty() {
 		data.Paxos = alloc.paxos.GossipState()
 	} else {
@@ -540,7 +543,7 @@ func (alloc *Allocator) actorLoop(actionChan <-chan func()) {
 			}
 			action()
 		case <-alloc.ticker.C:
-			if alloc.paxosActive {
+			if alloc.awaitingConsensus {
 				alloc.propose()
 			}
 			alloc.removeDeadContainers()
@@ -556,11 +559,11 @@ func (alloc *Allocator) actorLoop(actionChan <-chan func()) {
 
 // Ensure we are making progress towards an established ring
 func (alloc *Allocator) establishRing() {
-	if !alloc.ring.Empty() || alloc.paxosActive {
+	if !alloc.ring.Empty() || alloc.awaitingConsensus {
 		return
 	}
 
-	alloc.paxosActive = true
+	alloc.awaitingConsensus = true
 	alloc.propose()
 	if ok, cons := alloc.paxos.Consensus(); ok {
 		// If the quorum was 1, then proposing immediately
@@ -578,8 +581,8 @@ func (alloc *Allocator) createRing(peers []mesh.PeerName) {
 
 func (alloc *Allocator) ringUpdated() {
 	// When we have a ring, we don't need paxos any more
-	if alloc.paxosActive {
-		alloc.paxosActive = false
+	if alloc.awaitingConsensus {
+		alloc.awaitingConsensus = false
 		alloc.paxos = nil
 	}
 
@@ -668,11 +671,16 @@ func (alloc *Allocator) update(sender mesh.PeerName, msg []byte) error {
 		alloc.nicknames[peer] = nickname
 	}
 
-	// only one of Ring and Paxos should be present.  And we
-	// shouldn't get updates for a empty Ring. But tolerate
-	// them just in case.
-	if data.Ring != nil {
+	switch {
+	// If someone sent us a ring, merge it into ours. Note this will move us
+	// out of the awaiting-consensus state if we didn't have a ring already.
+	case data.Ring != nil:
 		switch err = alloc.ring.Merge(*data.Ring); err {
+		case nil:
+			if !alloc.ring.Empty() {
+				alloc.pruneNicknames()
+				alloc.ringUpdated()
+			}
 		case ring.ErrDifferentSeeds:
 			return fmt.Errorf("IP allocation was seeded by different peers (received: %v, ours: %v)",
 				alloc.annotatePeernames(data.Ring.Seeds), alloc.annotatePeernames(alloc.ring.Seeds))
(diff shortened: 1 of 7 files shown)


awh:
I am experimenting with a variant of (2) the involves some wrappers to avoid polluting the paxos package.

ipam/status.go
```diff
@@ -36,8 +36,8 @@ func NewStatus(allocator *Allocator, defaultSubnet address.CIDR) *Status {
 	}
 
 	var paxosStatus *paxos.Status
-	if allocator.paxosActive {
-		paxosStatus = paxos.NewStatus(allocator.paxos)
+	if allocator.awaitingConsensus && allocator.participant != nil {
+		//paxosStatus = paxos.NewStatus(allocator.paxos)
```
awh: Temporarily disabled

rade:
> I am experimenting with a variant of (2) that involves the use of wrappers to avoid polluting the paxos package.

Why do you think there is pollution? The paxos package seems like a very natural place for `Participant` to live.

ipam/allocator.go
```diff
@@ -708,6 +711,15 @@ func (alloc *Allocator) update(sender mesh.PeerName, msg []byte) error {
...
 
+	if data.Ring == nil && data.Paxos == nil {
+		// Neither Ring nor Paxos present - sender is a non-elector
+		// waiting for a ring update. Send them a copy of ours if we
+		// have one
+		if !alloc.ring.Empty() && sender != mesh.UnknownPeerName {
+			alloc.sendRingUpdate(sender)
+		}
+	}
+
```
rade: I think we want to replace the three top-level conditionals with a three-branch `switch`.

ipam/status.go
```diff
@@ -36,8 +36,10 @@ func NewStatus(allocator *Allocator, defaultSubnet address.CIDR) *Status {
 	}
 
 	var paxosStatus *paxos.Status
-	if allocator.paxosActive {
-		paxosStatus = paxos.NewStatus(allocator.paxos)
+	if allocator.awaitingConsensus && allocator.participant != nil {
+		if node, ok := allocator.participant.(*paxos.Node); ok {
+			paxosStatus = paxos.NewStatus(node)
+		}
```
rade: That's a bit ugly. And I would have thought we probably want to show something in the status for non-electors.

ipam/allocator.go
```diff
@@ -473,8 +473,11 @@ func (alloc *Allocator) encode() []byte {
 	}
 
 	// We're only interested in Paxos until we have a Ring.
+	// Non-electing participants (e.g. observers) return
+	// a nil gossip state in order to provoke a unicast ring
+	// update from electing peers who have reached consensus.
```
rade: I don't think that is sufficient.

When an ip allocation happens on a non-elector, this should prompt the establishment of a ring among the electors, even though none of them themselves may have required an election.

Otherwise IP allocation on a non-elector could be stalled, i.e. when non of the electors require allocation.

ipam/allocator.go
awh: Nice.

ipam/status.go
awh: Agree, it was a bit of a bodge until I got some feedback on (4).

ipam/allocator.go
awh: Excellent point!

awh:
> When an ip allocation happens on a non-elector, this should prompt the establishment of a ring among the electors, even though none of them themselves may have required an election.

This is now achieved by calling `establishRing` if we are polled by a non-elector but have no ring to send back. If we're an elector, we will propose, otherwise it's a no-op.

awh:
@rade simplified `update` as discussed - PTAL.

Will think about a smoke test - needs docs also.

awh:
Take #3.

ipam/allocator.go
```diff
@@ -681,30 +689,41 @@ func (alloc *Allocator) update(sender mesh.PeerName, msg []byte) error {
...
-			if alloc.paxos.Update(data.Paxos) {
-				if alloc.paxos.Think() {
-					// If something important changed, broadcast
-					alloc.gossip.GossipBroadcast(alloc.Gossip())
-				}
+	// If we reach this point we know the sender is either an elector
+	// broadcasting a paxos proposal to form a ring or a non-elector
+	// broadcasting a ring request. If we have a ring already we can just send
+	// it back regardless.
+	case !alloc.ring.Empty() && sender != mesh.UnknownPeerName:
```
rade: This should just be `case !alloc.ring.Empty()`, and the `sender != mesh.UnknownPeerName` moved inside, to guard the `alloc.sendRingUpdate`. In other words, if we have a non-empty ring we stop right here, and ignore any `data.Paxos` below.

rade:
I wonder about the scalability/performance implications of this. In a system of electors, there is a good chance that the first node on which an allocation happens will be the only one that makes a paxos proposal. It is only when there are near-simultaneous first allocations on multiple nodes that we get competing proposals. By contrast, if the first node on which an allocation happens is a non-elector, it will likely trigger proposals from all electors. This seems less than ideal.

ipam/allocator.go
awh: Erk, good spot, thanks

awh:
Some thoughts:
* Is this really a problem at all? In the worst case scenario (100 node elector set proposing simultaneously) how long does consensus take? I have been thinking about using terraform as a way to perform infrequent but more realistic scale testing than can be achieved with multiweave, perhaps testing this with AWS spot instances would be an interesting pilot
* Our tentative set of use-cases suggest you're unlikely to be using non-electors in conjunction with a large set of electors, so even if the above takes a long time it's unlikely to be encountered in practice
* Can be obviated entirely by `weave consense` in the elector set during initial cluster commissioning - was going to recommend that as a best practice anyway
* The simplicity of the current implementation is compelling

The above notwithstanding, how could we improve the solution? We could make `alloc.propose` behave differently for observers:
* Unicast a ring request to a random peer which is _known to be an elector_ (this information is not disseminated at the moment)
* Receiving elector proposes via broadcast to its quiescent peers; they accept the proposal, and form a ring
* If the above happens within 5s, the next time the observer picks a random elector to talk to it'll probably have a ring available

If observers don't know which of its peers are electors, then either we have to live with the fact that we keep retrying every 5s until we hit one (not ideal) or alternatively we could have non-electors that receive ring requests forward them immediately to another random peer (vulnerable to message loops, also not ideal). Consequently I think we would need to add voter/observer status to the connection handshake to improve on what we have right now - is it worth it?

rade:
> Is this really a problem at all?

I have no idea. Probably not at the 100 node scale.

test/125_observers_cant_vote_3_test.sh
```diff
@@ -0,0 +1,20 @@
...
+. ./config.sh
+
+start_suite "Observers can't vote"
+
+# Start network that doesn't have enough peers for consensus
+weave_on $HOST1 launch --init-peer-count=2
+weave_on $HOST2 launch --init-peer-count=0 $HOST1
+
+# Check that an allocation times out
+assert_raises "timeout 10 cat <( start_container $HOST1 )" 124
```
rade: 10s seems a bit long to wait in the "expected to time out" case.

test/126_observers_can_call_election_3_test.sh
```diff
@@ -0,0 +1,23 @@
...
+weave_on $HOST2 launch --init-peer-count=2 $HOST1
+
+# Add an observer
+weave_on $HOST3 launch --init-peer-count=0 $HOST2
+
+# Wait for network to settle; ensure IPAM is idle
+sleep 5
+assert "weave_on $HOST1 status ipam"
+assert "weave_on $HOST2 status ipam"
+assert "weave_on $HOST3 status ipam"
```
rade: huh? How does this "ensure IPAM is idle"?

test/125_observers_cant_vote_3_test.sh
awh: It's the value you chose in `510_detach_hide_no_ipalloc.sh`

test/126_observers_can_call_election_3_test.sh
awh: It's a bit of a cheat - it assumes that if it's not idle, then consensus will have been reached (and hence `status ipam` will output some values)

test/125_observers_cant_vote_3_test.sh
rade: Not true. That is for cases where we don't expect to hit the timeout. Note that U've not complained about any of those.

test/126_observers_can_call_election_3_test.sh
rade: subtle. deserves a comment. or perhaps add '""' to make it clear that we are asserting that the output is empty.

ipam/allocator.go
```diff
@@ -45,24 +45,24 @@ type operation interface {
...
+	ring              *ring.Ring                   // information on ranges owned by all peers
+	space             space.Space                  // more detail on ranges owned by us
+	owned             map[string][]address.Address // who owns what addresses, indexed by container-ID
+	nicknames         map[mesh.PeerName]string     // so we can map nicknames for rmpeer
+	pendingAllocates  []operation                  // held until we get some free space
+	pendingClaims     []operation                  // held until we know who owns the space
+	dead              map[string]time.Time         // containers we heard were dead, and when
+	db                db.DB                        // persistence
+	gossip            mesh.Gossip                  // our link to the outside world for sending messages
+	participant       paxos.Participant
```
bboreham: To my mind, the variable named `participant` could have been left as `paxos`; "participant" doesn't seem to add anything at the point of use to indicate it may be a real Paxos participant or an observer.

ipam/allocator.go
bboreham: Does this still need fixing?

bboreham:
Re point 1., I checked that down-level peers will tolerate a message with both ring and paxos `nil`.

ipam/allocator.go
bboreham: It looks to me that an allocation attempt will call `establishRing()` and will lead to a gossip broadcast, which in the case of an observer will have the nil-nil state that triggers an election in the electorate.

rade:
> Re point 1., I checked that down-level peers will tolerate a message with both ring and paxos nil.

Which point 1? Is this referring to backward compatibility?

ipam/allocator.go
awh: Yep, that had been addressed.

awh:
Thanks @bboreham, I have squashed your commits in. Outstanding design questions:

1. How should `weave status` / `weave report` be updated
2. Choice of command line option `--ipam-observer`? `--not-elector`?

Once these are resolved I can amend the documentation accordingly.

rade:
> Choice of command line option `--ipam-observer`? `--not-elector`?

I prefer non-ipam-specific terminology, on the grounds that we may well have other subsystems using the same mechanism one day.

awh:
> Once these are resolved I can amend the documentation accordingly.

Will do this after the documentation restructuring has been merged.

rade:
> Will do this after the documentation restructuring has been merged.

Please don't defer the documentation changes.

bboreham:
> Which point 1?

This point 1:

> 1. I have used a gossipState ring update message where both Ring and Paxos are nil as a way for non-electors to poll their peers for a ring - are we happy with that, or would a separate message type be preferable

>  Is this referring to backward compatibility?

Yes, I was referring to backward compatibility, new peers talking to older peers.

bboreham:
`--elector=false` ?

rade:
> `--elector=false` ?

The default can't be false, and having boolean options with a true default is exceptionally awkward.

awh:
I have chosen `--observer`. Added docs and rebased on master.

PR #2000, opened 2016-02-25
Allow specification of initial IPAM ring via --ipam-seed

### LINKED ISSUE #1998: Ability to specify initial IPAM ring seed
If I could do this
```
host1$ weave launch --name 00::01 --ipam-seed 00::01,00::02,00::03 $HOST1 $HOST3 $HOST3
host2$ weave launch --name 00::02 --ipam-seed 00::01,00::02,00::03 $HOST1 $HOST3 $HOST3
host3$ weave launch --name 00::03 --ipam-seed 00::01,00::02,00::03 $HOST1 $HOST3 $HOST3
```
it would be possible for me to spin up a cluster of parameterised VM images containing 'pre-concensed' weave nodes that could operate under partition from the outset.

awh:
Fixes #1998.

### DIFF
--- a/db/boltdb.go
+++ b/db/boltdb.go
@@ -9,7 +9,7 @@ import (
 )
 
 type DB interface {
-	Load(string, interface{}) error
+	Load(string, interface{}) (bool, error)
 	Save(string, interface{}) error
 }
 
@@ -50,11 +50,13 @@ func NewBoltDB(dbPathname string) (*BoltDB, error) {
 	return &BoltDB{db: db}, err
 }
 
-func (d *BoltDB) Load(ident string, data interface{}) error {
-	return d.db.View(func(tx *bolt.Tx) error {
+func (d *BoltDB) Load(ident string, data interface{}) (bool, error) {
+	found := true
+	err := d.db.View(func(tx *bolt.Tx) error {
 		b := tx.Bucket(topBucket)
 		v := b.Get([]byte(ident))
 		if v == nil {
+			found = false
 			return nil
 		}
 		reader := bytes.NewReader(v)
@@ -62,6 +64,7 @@ func (d *BoltDB) Load(ident string, data interface{}) error {
 		err := decoder.Decode(data)
 		return err
 	})
+	return found, err
 }
 
 func (d *BoltDB) Save(ident string, data interface{}) error {
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -47,6 +47,7 @@ type operation interface {
 type Allocator struct {
 	actionChan        chan<- func()
 	ourName           mesh.PeerName
+	seed              []mesh.PeerName           // optional user supplied ring seed
 	universe          address.Range             // superset of all ranges
 	ring              *ring.Ring                // information on ranges owned by all peers
 	space             space.Space               // more detail on ranges owned by us
@@ -70,6 +71,7 @@ type Config struct {
 	OurName     mesh.PeerName
 	OurUID      mesh.PeerUID
 	OurNickname string
+	Seed        []mesh.PeerName
 	Universe    address.Range
 	Quorum      uint
 	Db          db.DB
@@ -80,6 +82,7 @@ type Config struct {
 func NewAllocator(config Config) *Allocator {
 	return &Allocator{
 		ourName:     config.OurName,
+		seed:        config.Seed,
 		universe:    config.Universe,
 		ring:        ring.New(config.Universe.Start, config.Universe.End, config.OurName),
 		owned:       make(map[string][]address.CIDR),
@@ -835,31 +838,49 @@ func (alloc *Allocator) persistRing() {
 }
 
 func (alloc *Allocator) loadPersistedData() {
-	checkPeerName := alloc.ourName
-	if err := alloc.db.Load(nameIdent, &checkPeerName); err != nil {
+	var checkPeerName mesh.PeerName
+	nameFound, err := alloc.db.Load(nameIdent, &checkPeerName)
+	if err != nil {
 		alloc.fatalf("Error loading persisted peer name: %s", err)
-		return
 	}
-	if checkPeerName != alloc.ourName {
+	ringFound, err := alloc.db.Load(ringIdent, &alloc.ring)
+	if err != nil {
+		alloc.fatalf("Error loading persisted IPAM data: %s", err)
+	}
+	ownedFound, err := alloc.db.Load(ownedIdent, &alloc.owned)
+	if err != nil {
+		alloc.fatalf("Error loading persisted address data: %s", err)
+	}
+
+	if nameFound {
+		if checkPeerName == alloc.ourName {
+			if ringFound {
+				if len(alloc.seed) != 0 {
+					alloc.infof("Found persisted IPAM data, ignoring supplied IPAM seed")
+				}
+				alloc.space.UpdateRanges(alloc.ring.OwnedRanges())
+			}
+			if ownedFound {
+				for _, cidrs := range alloc.owned {
+					for _, cidr := range cidrs {
+						alloc.space.Claim(cidr.Addr)
+					}
+				}
+			}
+			return
+		}
 		alloc.infof("Deleting persisted data for peername %s", checkPeerName)
 		alloc.persistRing()
 		alloc.persistOwned()
-		return
 	}
-	if err := alloc.db.Load(ringIdent, &alloc.ring); err != nil {
-		alloc.fatalf("Error loading persisted ring data: %s", err)
+
+	if len(alloc.seed) != 0 {
+		alloc.infof("Initialising with supplied IPAM seed")
+		alloc.createRing(alloc.seed)
+	} else {
+		alloc.infof("Initialising via deferred consensus")
 	}
-	if alloc.ring != nil {
-		alloc.space.UpdateRanges(alloc.ring.OwnedRanges())
-	}
-	if err := alloc.db.Load(ownedIdent, &alloc.owned); err != nil {
-		alloc.fatalf("Error loading persisted address data: %s", err)
-	}
-	for _, cidrs := range alloc.owned {
-		for _, cidr := range cidrs {
-			alloc.space.Claim(cidr.Addr)
-		}
-	}
+
 }
 
 func (alloc *Allocator) persistOwned() {
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -91,6 +91,7 @@ func main() {
 		ipsubnetCIDR       string
 		peerCount          int
 		observer           bool
+		ipamSeedStr        string
 		dockerAPI          string
 		peers              []string
 		noDNS              bool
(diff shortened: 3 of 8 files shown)


prog/weaver/main.go
```diff
@@ -106,6 +107,7 @@ func main() {
 	mflag.StringVar(&dbPrefix, []string{"-db-prefix"}, "weave", "pathname/prefix of filename to store data")
 
 	mflag.StringVar(&trustedSubnetStr, []string{"-trusted-subnets"}, "", "Comma-separated list of trusted subnets in CIDR notation")
+	mflag.StringVar(&ipamSeedStr, []string{"-ipam-seed"}, "", "Comma-separated list of peer names used to form initial ring")
```
rade: nothing anywhere in our docs mentions "ring". and rightly so. We need to explain this in terms that don't delve into implementation detail.

awh:
Docs deliberately missing from this PR to avoid clashing with the documentation restructuring - will add some after that has merged.

rade:
> Docs deliberately missing from this PR to avoid clashing with the documentation restructuring - will add some after that has merged.

Please don't defer the doc changes.

awh:
Currently working on test breakage and some docs.

awh:
Rebased, fixed tests, added docs.

bboreham:
I found it striking that we make users go to all the trouble of formatting a peer-name that looks like a MAC.  We could just let them supply a number and interpret that as a peername internally.

ipam/allocator.go
```diff
@@ -804,31 +806,50 @@ func (alloc *Allocator) persistRing() {
...
 	}
-	for _, cidrs := range alloc.owned {
-		for _, cidr := range cidrs {
-			alloc.space.Claim(cidr.Addr)
+
+	if nameFound {
+		if checkPeerName == alloc.ourName {
+			if ringFound {
+				if len(alloc.seed) != 0 {
+					alloc.infof("Found persisted ring data, ignoring supplied IPAM seed")
```
bboreham: maybe "persisted IPAM data", since we don't document "ring"

awh:
> I found it striking that we make users go to all the trouble of formatting a peer-name that looks like a MAC. We could just let them supply a number and interpret that as a peername internally.

Totally agree - I was thinking of following up with another PR that lets us parse IPv6 style zero-runs e.g. ::1, ::2, ::3 - thoughts? It just means adjusting the peer name parses, should be pretty easy - I might have a stab at adding it quickly to this PR if I have time today.

rade:
I don't get it. The peer name does show up in various places in the ui, in mac notation. There is no point having different ways of inputting it if the output remains the same.

awh:
> There is no point having different ways of inputting it if the output remains the same.

Disagree - what's wrong with having a shorthand notation for use on the command line for specifying the values?

rade:
> what's wrong with having a shorthand notation for use on the command line for specifying the values?

Because then you cannot correlate back the values in the UI.

btw, you could just hash any user supplied name into a MAC, which nicely illustrates the problem with this approach.

awh:
> Because then you cannot correlate back the values in the UI.

You can, because there is an extremely simple rule that tells you that ::01 is equivalent to 00:00:00:00:00:01.

> btw, you could just hash any user supplied name into a MAC, which nicely illustrates the problem with this approach.

It doesn't, because humans can trivially reverse the mapping I'm talking about, unlike the hash you're suggesting.

awh:
Here's the kind of scheme I mean (taken from https://en.wikipedia.org/wiki/IPv6_address#Recommended_representation_as_text):

* Leading zeros in each 16-bit field are suppressed. For example, 2001:0db8::0001 is rendered as 2001:db8::1, though any all-zero field that is explicitly presented is rendered as 0.
* "::" is not used to shorten just a single 0 field. For example, 2001:db8:0:0:0:0:2:1 is shortened to 2001:db8::2:1, but 2001:db8:0000:1:1:1:1:1 is rendered as 2001:db8:0:1:1:1:1:1.
* Representations are shortened as much as possible. The longest sequence of consecutive all-zero fields is replaced by double-colon. If there are multiple longest runs of all-zero fields, then it is the leftmost that is compressed. E.g., 2001:db8:0:0:1:0:0:1 is rendered as 2001:db8::1:0:0:1 rather than as 2001:db8:0:0:1::1.

ipam/allocator.go
```diff
@@ -66,9 +67,10 @@ type Allocator struct {
 }
 
 // NewAllocator creates and initialises a new Allocator
-func NewAllocator(ourName mesh.PeerName, ourUID mesh.PeerUID, ourNickname string, universe address.Range, quorum uint, db db.DB, isKnownPeer func(name mesh.PeerName) bool) *Allocator {
+func NewAllocator(ourName mesh.PeerName, ourUID mesh.PeerUID, ourNickname string, seed []mesh.PeerName, universe address.Range, quorum uint, db db.DB, isKnownPeer func(name mesh.PeerName) bool) *Allocator {
```
rade: I think it's time for a Config struct.

rade:
> Here's the kind of scheme I mean

Yes, please!

ipam/allocator.go
awh: 2eeb754454a64026c85280ff2e55af37d04c8e0d

awh:
> Yes, please!

weaveworks/mesh#23

awh:
Will push a final update once the mesh PR is merged.

awh:
Ready for final review.

site/ipam/overview-init-ipam.md
```diff
@@ -22,6 +22,31 @@ The following automatic IP address managment topics are discussed:
...
+outset even under conditions of partition - no consensus is required.
+
+Alternatively, you can let weave determine the seed automatically via
+a consensus algorithm; since you don't need to provide it with a list
+of peer names anymore you can also let weave generate those randomly
+for you too. However, in order for weave to form a single consensus
+reliably you must now instead tell each peer how many peers there are
+in total either by listing them as target peers or via the
+`--init-peer-count` parameter.
+
```
rade: It is not obvious that the paragraphs that follow only apply to the "configuration via consensus" options. Perhaps we should create two sub sections.

rade:
I *think* the `--observer` flag works irrespective of whether the other peers use seeding or consensing.. That isn't obvious at all. That section currently (post #2060) mentions 'consensus', which would need replacing with more generic terminology, perhaps around "initialising IP address allocation" (btw, don't mention 'IPAM' inside the IPAM docs).

We may need to change the option name in the process. `--ipam-observer`, for starters.

`weave consense` is (now even more obviously) misnamed, since (I think) it can be used to initialise ipam on a peer from others that were seeded, not consensed. Or even when the peer itself has been given a `--ipam-seed` but perhaps not containing itself.

Finally, what happens if you specify more than one of `--ipam-seed`, `--init-peer-count`, and `--observer`?

db/boltdb.go
```diff
@@ -50,11 +50,13 @@ func NewBoltDB(dbPathname string) (*BoltDB, error) {
...
 
-func (d *BoltDB) Load(ident string, data interface{}) error {
-	return d.db.View(func(tx *bolt.Tx) error {
+func (d *BoltDB) Load(ident string, data interface{}) (bool, error) {
+	found := true
+	return found, d.db.View(func(tx *bolt.Tx) error {
 		b := tx.Bucket(topBucket)
 		v := b.Get([]byte(ident))
 		if v == nil {
+			found = false
```
rade: you are making a rather big assumption about evaluation here, namely that the `found` in `return found, ...` is evaluated *after* the `View` function has executed. The go spec may well guarantee that, but I'd rather avoid such subtlety.

rade:
The IPAM status terminology - both as documented and as displayed (which is different! #2061) - no longer makes sense, since it assumes the status is mostly driven by consensus, which will no longer be the case post this PR.

awh:
> We may need to change the option name in the process. --ipam-observer, for starters.

https://github.com/weaveworks/weave/pull/1990#issuecomment-191164562

awh:
> you are making a rather big assumption about evaluation here

Negatory, I tested it :)

> The go spec may well guarantee that, but I'd rather avoid such subtlety.

Fair enough!

rade:
> https://github.com/weaveworks/weave/pull/1990#issuecomment-191164562

ha ha. I was wondering how long it would take you to quote that ;)

awh:
> I think the --observer flag works irrespective of whether the other peers use seeding or consensing..

Technically yes - from the point of view of an observer, explicitly seeded peers are indistinguishable from those that were seeded by consensus. This would imply mixed usage of seeding and consensus in the same cluster though, which I wouldn't recommend.

> weave consense is (now even more obviously) misnamed  since (I think) it can be used to initialise ipam on a peer from others that were seeded, not consensed.

See above - you shouldn't (IMO) be using `weave consense` in a seeded cluster.

> Or even when the peer itself has been given a --ipam-seed but perhaps not containing itself.

Even here, `weave consense` doesn't make sense - such a peer is already initialised. Note that `weave consense` does not guarantee that you own a piece of the ring, just that you have a copy of the ring. If it _did_ make such a guarantee it would be useful in this scenario, as you could run it to ensure that a freshly added peer could subsequently make headway under partition.

> what happens if you specify more than one of --ipam-seed, --init-peer-count, and --observer?

Currently `--observer` and `--init-peer-count` are mutually exclusive - I should add `--ipam-seed` to that set.

awh:
> The IPAM status terminology - both as documented and as displayed (which is different! #2061) - no longer makes sense, since it assumes the status is mostly driven by consensus, which will no longer be the case post this PR.

Actually three of the five possible states remain relevant; peers launched with `--ipam-seed` will start in `ready` instead of `idle`. Seems fine to me - not sure how it could be changed.

awh:
> The IPAM status terminology - both as documented and as displayed (which is different! #2061) - no longer makes sense, since it assumes the status is mostly driven by consensus, which will no longer be the case post this PR.

I have raised #2063 to cover.

awh:
> ... Perhaps we should create two sub sections ... Perhaps refer to it as "one-off consensus" in most places.

Docs updated.

> Currently --observer and --init-peer-count are mutually exclusive - I should add --ipam-seed to that set.

Done.

awh:
Remaining concerns to be addressed by #2065.

prog/weaver/main.go
```diff
@@ -96,6 +96,7 @@ func main() {
 		datapathName       string
 		trustedSubnetStr   string
 		dbPrefix           string
+		ipamSeedStr        string
```
rade: This should go higher up, near the other ipam related args.

prog/weaver/main.go
```diff
@@ -134,6 +135,7 @@ func main() {
 	mflag.StringVar(&dbPrefix, []string{"-db-prefix"}, "weave", "pathname/prefix of filename to store data")
 
 	mflag.StringVar(&trustedSubnetStr, []string{"-trusted-subnets"}, "", "Comma-separated list of trusted subnets in CIDR notation")
+	mflag.StringVar(&ipamSeedStr, []string{"-ipam-seed"}, "", "Comma-separated list of peer names amongst which address space is shared initially")
```
rade: ditto

test/127_ipam_seed_2_test.sh
```diff
@@ -0,0 +1,22 @@
...
+assert_raises "timeout 10 cat <( start_container $HOST2 --name c2)"
+
+# Connect routers
+weave_on $HOST2 connect $HOST1
+
+# Check connectivity
+assert_raises "exec_on $HOST1 c1 $PING c2"
+assert_raises "exec_on $HOST2 c2 $PING c1"
+
+end_suite
```
rade: This file needs moving into the 3xx range.

rade:
arg needs adding to weave usage. So does `--observer`.

prog/weaver/main.go
```diff
@@ -467,6 +471,23 @@ func parseTrustedSubnets(trustedSubnetStr string) []*net.IPNet {
...
+func parseIPAMSeed(ipamSeed string) []mesh.PeerName {
+	peerNames := []mesh.PeerName{}
+	if ipamSeed == "" {
+		return peerNames
+	}
+
+	for _, peerNameStr := range strings.Split(ipamSeed, ",") {
+		peerName, err := mesh.PeerNameFromUserInput(peerNameStr)
+		if err != nil {
+			Log.Fatal("Unable to parse IPAM seed: ", err)
```
rade: Does this show the piece of the input it was tripping over? It should.

awh:
@rade all comments addressed, except for the one regarding add `--observer` to the weave script usage - I will push that directly to master.

PR #2002, opened 2016-02-25
Eliminate dot imports

awh:
They're bad, m'kay?

cc @peterbourgon

### DIFF
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -8,7 +8,7 @@ import (
 
 	docker "github.com/fsouza/go-dockerclient"
 
-	. "github.com/weaveworks/weave/common"
+	"github.com/weaveworks/weave/common"
 )
 
 const (
@@ -84,7 +84,7 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
 		for {
 			events := make(chan *docker.APIEvents)
 			if err := c.AddEventListener(events); err != nil {
-				Log.Errorf("[docker] Unable to add listener to Docker API: %s - retrying in %ds", err, retryInterval/time.Second)
+				common.Log.Errorf("[docker] Unable to add listener to Docker API: %s - retrying in %ds", err, retryInterval/time.Second)
 			} else {
 				start := time.Now()
 				for event := range events {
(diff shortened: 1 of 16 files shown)


rade:
Why not a few strategically placed

    var Log = common.Log

instead?

peterbourgon:
Why pollute the global namespace with variables and introduce an additional layer of indirection just to save a few keystrokes?

rade:
those keystrokes you are saving are adding noise for the reader.

peterbourgon:
They don't add noise, rather explicitness. The goal is to minimize ambiguity, not quantity of characters, and `common.Log` is strictly less ambiguous than `Log`.

rade:
`Log` is not ambiguous. Otherwise it wouldn't compile :)

rade:
overall, may I ask what problem we are trying to solve here?

peterbourgon:
> Except for this one case, do not use `import .` in your programs. It makes the programs much harder to read because it is unclear whether a name like Quux is a top-level identifier in the current package or in an imported package.
—https://github.com/golang/go/wiki/CodeReviewComments#import-dot

rade:
And has anybody actually run into this problem when reading the weave code?

peterbourgon:
I have.

rade:
So you read the weave code, came across a statement like

    Log.Println("Our name is", router.Ourself)

and wondered "Where does `Log` come from?". Then what happened?

bboreham:
`common` is a poor name.

`log.Log.Infof` is more explanatory, but stutters a bit. We could repeat all the entry points in a `weave/log` package, to avoid that.

For those bits of code that are really libraries, it might be better to pass in a logger reference on startup, then the caller has control.

awh:
I wanted to force the issue of refactoring `common` by making its use glaringly obvious - can the consequent improvements go in separate PRs?

awh:
@bboreham is there a reason we can't just merge this now? It's a strict refactoring...

bboreham:
Because it makes the code strictly more ugly.

rade:
>  it makes the code strictly more ugly

I was hoping that https://github.com/weaveworks/weave/pull/2002#issuecomment-188861434 might be an acceptable compromise.

awh:
> Because it makes the code strictly more ugly.

https://github.com/weaveworks/weave/pull/2002#issuecomment-191175405

rade:
I have eliminated the bulk of dot imports.

The remainder will be removed with #2074.

-> closing.

Issue #2003, opened 2016-02-28
Multicast is slower than expected in fastdp


bboreham:
Because the receiving weaver is getting called on every packet via the ODP miss handler

Testing with iperf, in debug mode we see lots of these:

 ````
DEBU: 2016/02/28 17:50:43.307011 ODP miss map[20:BlobFlowKey{type: 20, key: 00000000, mask: ffffffff} 3:InPortFlowKey{vport: 1} 2:BlobFlowKey{type: 2, key: 00000000, mask: ffffffff} 16:TunnelFlowKey{id: 0000000000065bb2, ipv4src: 192.168.48.12, ipv4dst: 192.168.48.11, ttl: 64, tpsrc: 50017, tpdst: 6784} 15:BlobFlowKey{type: 15, key: 00000000, mask: ffffffff} 4:EthernetFlowKey{src: 82:43:51:a8:79:d4, dst: 01:00:5e:01:02:03} 6:BlobFlowKey{type: 6, key: 0800, mask: ffff} 7:BlobFlowKey{type: 7, key: 0a200001e001020311000101, mask: ffffffffffffffffffffffff} 10:BlobFlowKey{type: 10, key: d0b51389, mask: ffffffff} 19:BlobFlowKey{type: 19, key: 00000000, mask: ffffffff}] on port 1
DEBU: 2016/02/28 17:50:43.307245 ODP miss map[15:BlobFlowKey{type: 15, key: 00000000, mask: ffffffff} 19:BlobFlowKey{type: 19, key: 00000000, mask: ffffffff} 16:TunnelFlowKey{id: 0000000000065bb2, ipv4src: 192.168.48.12, ipv4dst: 192.168.48.11, ttl: 64, tpsrc: 50017, tpdst: 6784} 3:InPortFlowKey{vport: 1} 6:BlobFlowKey{type: 6, key: 0800, mask: ffff} 7:BlobFlowKey{type: 7, key: 0a200001e001020311000102, mask: ffffffffffffffffffffffff} 4:EthernetFlowKey{src: 82:43:51:a8:79:d4, dst: 01:00:5e:01:02:03} 20:BlobFlowKey{type: 20, key: 00000000, mask: ffffffff} 2:BlobFlowKey{type: 2, key: 00000000, mask: ffffffff}] on port 1
[...]

PR #2008, opened 2016-02-29
Process DiscardingFlowOp correctly in FastDatapath send()

dpw:
The nil FlowOp check in collectFlowOps was not addressed with the
introduction of DiscardingFlowOp.  This meant that DiscardingFlowOps
produced in the router core would get handled by fastdp as if they were
foreign FlowOps, preventing flow creation and so leading to unwarranted
ODP misses.

To fix this, collectFlowOps cannot simply eliminate FlowOps for which
Discards() is true, because it's true for odpFlowKeyFlowOp.  So the
Discards() case has to be handled in FastDatapath send().

### DIFF
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -765,10 +765,10 @@ func (fwd *fastDatapathForwarder) Forward(key ForwardPacketKey) FlowOp {
 	defer fwd.lock.RUnlock()
 
 	if fwd.remoteAddr == nil {
-		// Returning nil would discard the packet, but also
-		// result in a flow rule, which we would have to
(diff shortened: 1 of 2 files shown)


dpw:
Fixes #2003.

bboreham:
LGTM, but I think we may prefer to create a 1.4 patch release.

dpw:
> LGTM, but I think we may prefer to create a 1.4 patch release.

If I tried to follow the formalities, I'd probably forget something important and embarrass myself, so I'll leave that part to your team. This PR is just a way to present the fix.

bboreham:
Thanks; I'll do the merge

Issue #1655, opened 2015-11-09
Additional IP addresses are lost on container restart


bboreham:
As noted in #1191, IP addresses that have been allocated to a container after it started, with `weave attach`, are lost when the container is restarted by Docker.

This applies whether the address is chosen by IPAM or by the user

bboreham:
Currently, via the http interface to IPAM, you can only fetch one address for a container, in a given address range.  If we extended this to fetch all addresses owned by a container then we could have the proxy re-attach all of them on restart.

Would need to extend the information held by `Allocator.owned` to include the subnet.

This would still leave addresses attached by the user outwith IPAM's control.

PR #2012, opened 2016-03-01
Store all addresses+subnets in IPAM

bboreham:
Partial fix for #1655 

The missing piece is having the proxy recover the addresses on restart.  Currently the proxy acts via `weave attach`, so we would need a variant of that to retrieve all addresses from IPAM.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -47,16 +47,16 @@ type operation interface {
 type Allocator struct {
 	actionChan       chan<- func()
 	ourName          mesh.PeerName
-	universe         address.Range                // superset of all ranges
-	ring             *ring.Ring                   // information on ranges owned by all peers
-	space            space.Space                  // more detail on ranges owned by us
-	owned            map[string][]address.Address // who owns what addresses, indexed by container-ID
-	nicknames        map[mesh.PeerName]string     // so we can map nicknames for rmpeer
-	pendingAllocates []operation                  // held until we get some free space
-	pendingClaims    []operation                  // held until we know who owns the space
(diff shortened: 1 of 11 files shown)


bboreham:
One thing: this changes the persistence format of owned addresses, so we perhaps want to get that change into 1.5.0.

ipam/allocator.go
```diff
@@ -50,7 +50,7 @@ type Allocator struct {
 	universe         address.Range                // superset of all ranges
 	ring             *ring.Ring                   // information on ranges owned by all peers
 	space            space.Space                  // more detail on ranges owned by us
-	owned            map[string][]address.Address // who owns what addresses, indexed by container-ID
+	owned            map[string][]address.CIDR // who owns what addresses, indexed by container-ID
```
rade: vertical alignment fail

net/address/address.go
```diff
@@ -46,29 +50,34 @@ func ParseIP(s string) (Address, error) {
...
+func (cidr CIDR) IsSubnet() bool {
+	mask := cidr.Size() - 1
+	return Offset(cidr.Addr)&mask == 0
+}
+
 func (cidr CIDR) Size() Offset { return 1 << uint(32-cidr.PrefixLen) }
 
 func (cidr CIDR) Range() Range {
-	return NewRange(cidr.Start, cidr.Size())
+	return NewRange(cidr.Addr, cidr.Size())
```
bboreham: We should perhaps mask the address down to its network portion, otherwise this only makes sense when called on a CIDR that is a network address.

PR #1979, opened 2016-02-17
[WIP] 1.5 edit and updates to Weave Net User's Guide: WP rollout

abuehrle:
@awh @bboreham created three files for fastdp: how it works, using fastdp and viewing connections. 

Please have a look and make sure I haven't missed anything technically.  Probably still needs a typo sweep, as I am using Eclipse to edit/write.

(Ignore the README.md addition, I added that by mistake and nothings changed but spaces.  The majority of that file will go into the overview, BTW. )

[Review on Reviewable](https://reviewable.io/reviews/weaveworks/weave/1979)

PR #2015, opened 2016-03-02
Reorganization of docs for WordPress

abuehrle:


rade:
What's relationship between this and #1979?

abuehrle:
Bryan suggested I do this, so that he could merge the files.  We are keeping the other PR to preserve the conversation.  I think because I was merging the master into my branch, instead of rebasing or something like that.

bboreham:
This replaces #1979

site/router-topology/router-encapsulation.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Weave Router Encapsulation
+layout: default
+---
+
+
+When the Weave router forwards packets, the encapsulation looks
+something like this:
+
+    +-----------------------------------+
```
bboreham: this diagram looks like site/images/weave-frame-encapsulation-178x300.png ; was it supposed to be replaced?

abuehrle: I didn't think we had the time, but now we do. Yes, it would be great to replace those.  Should we use the one you did for the Docker University?

PR #2024, opened 2016-03-07
Abstract out logging function from weave api package

bboreham:
To reduce dependencies for other projects who decide to use the Weave Net API.  With this change, you have to pass in the logger you want to use when you create the `Client` object.

The `Logger` interface here is a bit arbitrary; the only function we are actually using is `Debugf`, but it seemed a bit weird to just have that one line.

### DIFF
--- a/api/weaveapi.go
+++ b/api/weaveapi.go
@@ -8,8 +8,6 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
-
-	. "github.com/weaveworks/weave/common"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


rade:
@peterbourgon I would like to get your take on this. I haven't seen any go libraries where logging is configured the way proposed here. Have you? Do you reckon it's ok to do this? Are there better approaches?

peterbourgon:
Loggers are dependencies, so a strong :+1: from me to pass a logger into your constructors. 

Plenty of other packages do this [or something like it](https://github.com/coreos/etcd/blob/master/raft/raft.go#L107-L109); the only improvement I could suggest is making it optional, but in order to do that in this usage, we'd need to complect the API, either or introducing a config object with a usable zero value. In either case, probably not worth it.

Issue #1992, opened 2016-02-23
CNI IPAM plugin


rade:
We should make weave easily deployable as a [CNI IPAM plugin](https://github.com/appc/cni/blob/master/SPEC.md#ip-allocation). See also #1991.

bboreham:
Closed in #2030

PR #2030, opened 2016-03-08
Implement CNI plugin

### LINKED ISSUE #1991: CNI network plugin
we should make weave easily deployable as a [CNI network plugin](https://github.com/appc/cni/blob/master/SPEC.md).

bboreham:
Fixes #1991 and #1992.

[CNI](https://github.com/appc/cni#cni---the-container-network-interface) expects to call an executable, which you name in your `/etc/cni/net.d/somenetwork.conf` file.  By default it looks in `/opt/cni/bin`

For the Weave plugin, `weave setup` creates a couple of scripts in `/opt/cni/bin` which run the plugin container with the necessary arguments, and also creates a configuration file `/etc/cni/net.d/10-weave.conf`.

If you do not configure an IPAM `type` in the configuration, the Weave CNI network plugin will call Weave IPAM directly.

If you don't want the overhead of a container execution on every plugin operation you can also use the Weave plugin executable directly, renamed:

- If you name it `weave-net` it acts as a CNI network plugin.
- If you name it `weave-ipam` it acts as a CNI IPAM plugin.

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -48,7 +48,7 @@ type NetDev struct {
 }
 
 // Search the network namespace of a process for interfaces matching a predicate
-func FindNetDevs(processID int, match func(string) bool) ([]NetDev, error) {
+func FindNetDevs(processID int, match func(link netlink.Link) bool) ([]NetDev, error) {
 	var netDevs []NetDev
 
 	ns, err := netns.GetFromPid(processID)
@@ -58,40 +58,57 @@ func FindNetDevs(processID int, match func(string) bool) ([]NetDev, error) {
 	defer ns.Close()
 
 	err = WithNetNS(ns, func() error {
-		links, err := netlink.LinkList()
-		if err != nil {
-			return err
-		}
-		for _, link := range links {
-			if match(link.Attrs().Name) {
-				addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
+		return forEachLink(func(link netlink.Link) error {
+			if match(link) {
+				netDev, err := linkToNetDev(link)
 				if err != nil {
 					return err
 				}
-
-				netDev := NetDev{MAC: link.Attrs().HardwareAddr}
-				for _, addr := range addrs {
-					netDev.CIDRs = append(netDev.CIDRs, addr.IPNet)
-				}
-				netDevs = append(netDevs, netDev)
+				netDevs = append(netDevs, *netDev)
 			}
-		}
-		return nil
+			return nil
+		})
 	})
 
 	return netDevs, err
 }
 
+func forEachLink(f func(netlink.Link) error) error {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return err
+	}
+	for _, link := range links {
+		if err := f(link); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func linkToNetDev(link netlink.Link) (*NetDev, error) {
+	addrs, err := netlink.AddrList(link, netlink.FAMILY_V4)
+	if err != nil {
+		return nil, err
+	}
+
+	netDev := &NetDev{MAC: link.Attrs().HardwareAddr}
+	for _, addr := range addrs {
+		netDev.CIDRs = append(netDev.CIDRs, addr.IPNet)
+	}
+	return netDev, nil
+}
+
 // Lookup the weave interface of a container
 func GetWeaveNetDevs(processID int) ([]NetDev, error) {
-	return FindNetDevs(processID, func(name string) bool {
-		return strings.HasPrefix(name, "ethwe")
+	return FindNetDevs(processID, func(link netlink.Link) bool {
+		return strings.HasPrefix(link.Attrs().Name, "ethwe")
 	})
 }
 
 // Get the weave bridge interface
 func GetBridgeNetDev(bridgeName string) ([]NetDev, error) {
-	return FindNetDevs(1, func(name string) bool {
-		return name == bridgeName
+	return FindNetDevs(1, func(link netlink.Link) bool {
+		return link.Attrs().Name == bridgeName
 	})
 }
--- a/common/utils.go
+++ b/common/utils.go
@@ -117,7 +117,13 @@ func GetWeaveNetDevs(processID int) ([]NetDev, error) {
 	indexes := make(map[int]struct{})
 	err = forEachLink(func(link netlink.Link) error {
 		if link.Attrs().MasterIndex == weaveBridge.Attrs().Index {
-			indexes[link.Attrs().Index] = struct{}{}
+			peerIndex := link.Attrs().ParentIndex
+			if peerIndex == 0 {
+				// perhaps running on an older kernel where ParentIndex doesn't work.
+				// as fall-back, assume the indexes are consecutive
+				peerIndex = link.Attrs().Index - 1
+			}
+			indexes[peerIndex] = struct{}{}
 		}
 		return nil
 	})
@@ -125,9 +131,8 @@ func GetWeaveNetDevs(processID int) ([]NetDev, error) {
 		return nil, err
 	}
 	return FindNetDevs(processID, func(link netlink.Link) bool {
-		// For checking, rely on index number of veth end inside namespace being -1 of end outside
 		_, isveth := link.(*netlink.Veth)
-		_, found := indexes[link.Attrs().Index+1]
+		_, found := indexes[link.Attrs().Index]
 		return isveth && found
 	})
 }
--- a/common/utils.go
+++ b/common/utils.go
@@ -103,8 +103,14 @@ func linkToNetDev(link netlink.Link) (*NetDev, error) {
 // Lookup the weave interface of a container
 func GetWeaveNetDevs(processID int) ([]NetDev, error) {
 	// Bail out if this container is running in the root namespace
-	nsToplevel, _ := netns.GetFromPid(1)
-	nsContainr, _ := netns.GetFromPid(processID)
+	nsToplevel, err := netns.GetFromPid(1)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open root namespace: %s", err)
+	}
+	nsContainr, err := netns.GetFromPid(processID)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open process %d namespace: %s", processID, err)
+	}
 	if nsToplevel.Equal(nsContainr) {
 		return nil, nil
 	}
--- a/common/utils.go
+++ b/common/utils.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"fmt"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 	"net"
@@ -101,8 +102,39 @@ func linkToNetDev(link netlink.Link) (*NetDev, error) {
 
 // Lookup the weave interface of a container
 func GetWeaveNetDevs(processID int) ([]NetDev, error) {
+	// Bail out if this container is running in the root namespace
+	nsToplevel, _ := netns.GetFromPid(1)
+	nsContainr, _ := netns.GetFromPid(processID)
+	if nsToplevel.Equal(nsContainr) {
+		return nil, nil
+	}
+
+	var weaveBridge netlink.Link
+	forEachLink(func(link netlink.Link) error {
+		if link.Attrs().Name == "weave" {
+			weaveBridge = link
+		}
+		return nil
+	})
+	if weaveBridge == nil {
+		return nil, fmt.Errorf("Cannot find weave bridge")
+	}
+	// Scan devices in root namespace to find those attached to weave bridge
+	indexes := make(map[int]struct{})
+	err := forEachLink(func(link netlink.Link) error {
+		if link.Attrs().MasterIndex == weaveBridge.Attrs().Index {
+			indexes[link.Attrs().Index] = struct{}{}
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
 	return FindNetDevs(processID, func(link netlink.Link) bool {
-		return strings.HasPrefix(link.Attrs().Name, "ethwe")
+		// For checking, rely on index number of veth end inside namespace being -1 of end outside
+		_, isveth := link.(*netlink.Veth)
+		_, found := indexes[link.Attrs().Index+1]
+		return isveth && found
 	})
 }
 
--- a/common/utils.go
+++ b/common/utils.go
@@ -109,19 +109,13 @@ func GetWeaveNetDevs(processID int) ([]NetDev, error) {
 		return nil, nil
 	}
 
-	var weaveBridge netlink.Link
-	forEachLink(func(link netlink.Link) error {
-		if link.Attrs().Name == "weave" {
-			weaveBridge = link
-		}
-		return nil
-	})
-	if weaveBridge == nil {
-		return nil, fmt.Errorf("Cannot find weave bridge")
+	weaveBridge, err := netlink.LinkByName("weave")
+	if err != nil {
+		return nil, fmt.Errorf("Cannot find weave bridge: %s", err)
 	}
 	// Scan devices in root namespace to find those attached to weave bridge
 	indexes := make(map[int]struct{})
-	err := forEachLink(func(link netlink.Link) error {
+	err = forEachLink(func(link netlink.Link) error {
 		if link.Attrs().MasterIndex == weaveBridge.Attrs().Index {
 			indexes[link.Attrs().Index] = struct{}{}
 		}
--- a/plugin/ipam/cni.go
+++ b/plugin/ipam/cni.go
@@ -10,10 +10,18 @@ import (
 )
 
 func (i *Ipam) CmdAdd(args *skel.CmdArgs) error {
+	result, err := i.Allocate(args)
+	if err != nil {
+		return err
+	}
+	return result.Print()
+}
+
+func (i *Ipam) Allocate(args *skel.CmdArgs) (*types.Result, error) {
 	// extract the things we care about
 	conf, err := loadIPAMConf(args.StdinData)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	if conf == nil {
 		conf = &ipamConf{}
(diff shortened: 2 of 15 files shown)


bboreham:
Note that 'gateway' and 'ipmasq' functionality are not yet implemented.

rade:
What's the plan for a) release artifacts, and b) docs? I am happy for these to be done in separate PRs, but we should at least have issues for them (plus anything else we can think of to consider this work reasonably complete).

awh:
Could we just have `weave setup` do the `docker cp` to `/opt/cni/bin`? No additional release artifacts are required then.

awh:
Alternatively, could we have specially named symlinks in `/opt/cni/bin` (again perhaps installed by `weave setup` or `weave launch`) that point to the _weave script itself_, which could detect how it has been invoked and `docker run` the CNI plugin executable from one of our containers?

bboreham:
@awh it's an interesting idea.  Requires write access to that directory.

Also we discussed having the 'plugin' (the entry in /opt/cni/bin) be a symlink to the `weave` script, which would `docker run` the plugin binary.  This has the advantage of restricting the number of separately-managed artefacts, also makes subsequent upgrades simpler.

bboreham:
When used with Kubernetes, the pod's ethernet device is named `eth0`, which defeats the logic of `weave ps`.  I.e. `weave ps` will not report the containers as attached to the weave network.

awh:
Looking good @bboreham

plugin/ipam/cni.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return result, nil
+}
+
+func (i *Ipam) CmdDel(args *skel.CmdArgs) error {
+	if err := i.Release(args); err != nil {
+		return err
+	}
+	result := types.Result{}
+	return result.Print()
```
achanda: Is that line necessary? The other plugins in the cni repo do not have this

bboreham: Possibly. Paging @squaremo ?

bboreham: I removed it in 	c40a544.  Thanks @achanda !

errordeveloper:
I've trying to do this for my kubelet container:
```
ADD https://storage.googleapis.com/kubernetes-release/network-plugins/cni-09214926.tar.gz /tmp/cni.tgz
ADD https://raw.githubusercontent.com/weaveworks/weave/issues/1992-cni-plugin/weave /usr/bin/weave
ADD https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 /usr/bin/docker

RUN mkdir -p /opt/cni/bin && tar xzf /tmp/cni.tgz -C /opt/cni/ && rm -f /tmp/cni.tgz && chmod +x /usr/bin/weave && ln -s /usr/bin/weave /opt/cni/bin/weave-net && ln -s /usr/bin/weave /opt/cni/bin/weave-ipam && mkdir -p /etc/cni/net.d && echo '{ "name": "weave", "type": "weave-net" }' > /etc/cni/net.d/10-weave.conf && chmod +x /usr/bin/docker
ENV WEAVE_VERSION=git-f18f10bac090
```

Weave script symlinks seem to work, i.e. execing `weave-ipam` inside of this container returns an error about missing env vars... However, kubelet still spits out complains like this:
```
NetworkPlugin cni failed on the status hook for pod 'redis-master-clm1t' - invalid CIDR address: Device "eth0" does not exist.
```

Not sure what else I'm missing in the containerised situation, please let me know if you have any ideas.

achanda:
@errordeveloper you need to set a bunch of env vars to run the ipam plugins manually. Here are some docs https://github.com/appc/cni/tree/master/plugins/ipam/host-local

errordeveloper:
@achanda thanks! I sort of figured that, and recall even reading the original spec... I was just testing that containerised plugin is executable and does something close to what's desired (i.e. makes calls to the weave router container etc).

errordeveloper:
Ah, I've had Weave proxy socket mounted into this, looks like all is working with proper docker socket now!

errordeveloper:
I believe I got the last thing to fix, which is due to `kube-proxy` defaulting to `iptables` mode, which I discovered doesn't work with Weave as is. Any thoughts on this part? @bboreham did you test end-to-end agains Kubernetes v1.2 or it was v1.1? (The proxy mode change in v1.2).

bboreham:
I think I paused my 1.2 testing to make sure it worked with the current release (1.1). 

But I also thought I tested in iptables mode. What exactly is the symptom you had?

PS: I think my symlink setup needs fixing for the case where weave runs in a container (which is the default)

errordeveloper:
> PS: I think my symlink setup needs fixing for the case where weave runs in a container (which is the default)

Yes, it symlinks to `/home/weave/weave`...

errordeveloper:
> But I also thought I tested in iptables mode. What exactly is the symptom you had?

Services are unreachable on their service IPs.

Well, actually, looks like it doesn't work in the old mode either, while it did before I started testing the plugin.

bboreham:
You did `weave expose` ?

Check the pods are wired up with `weave ps`. 

Basically, just trace through the route then through iptables. If the plugin ran, all the things that are likely to go wrong are on the outside.  I had endless trouble with VirtualBox disabling networks.

errordeveloper:
Ok, looks SkyDNS keeps dies and restarts. I ran `kubectl get events --watch` and spotted this:
```
2016-03-15 22:02:48 +0000 UTC   2016-03-15 22:03:15 +0000 UTC   4         kube-dns-v10-f1xju   Pod                 Normal    InfraChanged   {kubelet kubedev}   Pod infrastructure changed, it will be killed and re-created.
```

errordeveloper:
> You did weave expose ?

Oh, actually I forgot this time, as I diverged from the usual startup sequence! But I did have it in my usual test, so I don't think that's that made it fail last time, will go and verify later, as I'd like the proxy to work in iptables mode (ideally). Now I'll go and see if `weave expose` somehow resolves the current problem...

errordeveloper:
Turns out all my pods are restarting with reason `InfraChanged`, sounds like fun...

errordeveloper:
OK, I've tried creating a single replication controller and it's still complaining about status hook, which causes `InfraChanged` event.
```
 NetworkPlugin cni failed on the status hook for pod 'redis-master-r2kkt' - invalid CIDR address: Device "eth0" does not exist.
```

errordeveloper:
This status call comes with [an exciting TODO note](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/cni/cni.go#L138-L154), by the way.

errordeveloper:
I am completely puzzled where this error comes from, not able to find any combination of similar string anywhere, I mean the `invalid CIDR address: Device "eth0" does not exist` portion of the error message.

rade:
`invalid CIDR address` comes from golang's `net.ParseCIDR`.

errordeveloper:
Looks like https://github.com/kubernetes/kubernetes/issues/22932 is what my issue is.

site/plugin/cni-plugin-how-to.md
```diff
@@ -0,0 +1,77 @@
...
+CNI, the [_Container Network Interface_](https://github.com/appc/cni#cni---the-container-network-interface),
+is a proposed standard for configuring network interfaces for Linux
+application containers.  CNI is supported by
+[Kubernetes](http://kubernetes.io/), an open source container cluster
+manager built by Google.
+
+###Installing the Weave Net CNI plugin
+
+The Weave Net CNI plugin is installed when you run `weave setup`, if
+it your machine has the directores normally used to host CNI plugins.
```
awh: directories

plugin/net/cni.go
```diff
@@ -0,0 +1,158 @@
...
+		return cleanup(errorf("unable to allocate IP address: %s", err))
+	}
+	if result.IP4 == nil {
+		return cleanup(errorf("IPAM plugin failed to allocate IP address"))
+	}
+
+	err = common.WithNetNS(ns, func() error {
+		guest, err := netlink.LinkByName(local.PeerName)
+		if err != nil {
+			return cleanup(err)
```
awh: I think we can just `return err` here as we do in the next conditional as we're cleaning up any err return from WithNetNS

common/utils.go
```diff
@@ -101,8 +102,39 @@ func linkToNetDev(link netlink.Link) (*NetDev, error) {
...
 // Lookup the weave interface of a container
 func GetWeaveNetDevs(processID int) ([]NetDev, error) {
+	// Bail out if this container is running in the root namespace
+	nsToplevel, _ := netns.GetFromPid(1)
+	nsContainr, _ := netns.GetFromPid(processID)
+	if nsToplevel.Equal(nsContainr) {
+		return nil, nil
+	}
+
+	var weaveBridge netlink.Link
```
awh: `weaveBridge := netlink.LinkByName("weave")`

common/utils.go
```diff
@@ -101,8 +102,39 @@ func linkToNetDev(link netlink.Link) (*NetDev, error) {
...
+			indexes[link.Attrs().Index] = struct{}{}
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
 	return FindNetDevs(processID, func(link netlink.Link) bool {
-		return strings.HasPrefix(link.Attrs().Name, "ethwe")
+		// For checking, rely on index number of veth end inside namespace being -1 of end outside
```
awh: Is this safe enough for 1.5? Any thoughts @dpw?

common/utils.go
```diff
@@ -144,3 +145,23 @@ func GetBridgeNetDev(bridgeName string) ([]NetDev, error) {
 		return link.Attrs().Name == bridgeName
 	})
 }
+
+// Configure the ARP cache parameters for the given interface.  This
+// makes containers react more quickly to a change in the MAC address
+// associated with an IP address.
+func ConfigureArpCache(ifname string) error {
```
awh: I made a similar function here https://github.com/weaveworks/weave/pull/1958/commits/0c3f751b492f4991b33887ba9c710da978916063, that does not rely on `ioutil.WriteFile` because it uses O_CREATE and hence won't error if the path does not exist

prog/plugin/main.go
```diff
@@ -26,6 +26,7 @@ var version = "(unreleased version)"
 func main() {
 	var (
 		justVersion      bool
+		cniNet, cniIpam  bool
```
awh: I've never seen us do this anywhere else - why here? Why not `address, meshAddress string`?

common/utils.go
bboreham: Why do we want it to error?

awh: Why would we not want it to error? We expect the file to be there as part of procfs - if it's not, something is wrong

common/utils.go
```diff
@@ -58,40 +60,108 @@ func FindNetDevs(processID int, match func(string) bool) ([]NetDev, error) {
...
+		if link.Attrs().MasterIndex == weaveBridge.Attrs().Index {
+			indexes[link.Attrs().Index] = struct{}{}
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return FindNetDevs(processID, func(link netlink.Link) bool {
+		// For checking, rely on index number of veth end inside namespace being -1 of end outside
```
awh: Is this assumption safe enough to go into 1.5? Thoughts @dpw?

awh:
Looking good!

plugin/net/cni.go
bboreham: This got moved around in a later commit.

bboreham:
I addressed all comments except the consecutive-link-number one.  Cherry-picked your ARP function and rebased.

awh:
> except the consecutive-link-number one

Good news - `link.Attrs().ParentIndex` corresponds to the `IFLA_LINK` attribute which is the peer ifindex for veths!

awh:
```
$ cat pif.go 
package main

import "fmt"
import "os"
import "github.com/vishvananda/netlink"

func main() {
	link, err := netlink.LinkByName(os.Args[1])
	if err == nil {
		fmt.Println("Index", link.Attrs().Index)
		fmt.Println("Master", link.Attrs().MasterIndex)
		fmt.Println("Peer Index", link.Attrs().ParentIndex)
	}
}
```

```
6: datapath: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1410 qdisc noqueue state UNKNOWN mode DEFAULT group default 
    link/ether 82:6f:90:62:11:76 brd ff:ff:ff:ff:ff:ff
7: weave: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1410 qdisc noqueue state UP mode DEFAULT group default 
    link/ether aa:e7:41:ba:ed:ca brd ff:ff:ff:ff:ff:ff
8: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN mode DEFAULT group default 
    link/ether f2:42:3f:0a:ab:48 brd ff:ff:ff:ff:ff:ff
10: vethwe-datapath@vethwe-bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1410 qdisc pfifo_fast master datapath state UP mode DEFAULT group default qlen 1000
    link/ether f6:4a:eb:69:2e:53 brd ff:ff:ff:ff:ff:ff
11: vethwe-bridge@vethwe-datapath: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1410 qdisc pfifo_fast master weave state UP mode DEFAULT group default qlen 1000
    link/ether 0e:0a:2c:06:61:26 brd ff:ff:ff:ff:ff:ff
13: veth7a26e0e@if12: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master docker0 state UP mode DEFAULT group default 
    link/ether 7e:3a:42:13:cb:7d brd ff:ff:ff:ff:ff:ff link-netnsid 0
15: vethwepl2468@if14: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1410 qdisc pfifo_fast master weave state UP mode DEFAULT group default qlen 1000
    link/ether ae:d5:a3:fd:c9:f8 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

```
$ ./pif vethwepl2468
Index 15
Master 7
Peer Index 14
```

```
$ sudo nsenter --target=2468 --net ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
12: eth0@if13: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 02:42:ac:11:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
14: ethwe@if15: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1410 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
    link/ether 3a:71:8a:cd:2d:96 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

awh:
Can we smoke test this without installing kubernetes?

common/utils.go
dpw: No, I don't think it is very safe.  While veth ifindexes will tend to be consecutive, I don't think there is any guarantte about which comes first.  And the might not be consecutive.  Consider the case where two veth pairs are created concurrently.  Also, it looks like it might be possible to specify a ifindex when creating a new dev.  Also, ifindex allocation can wrap.

Netlink provides a way to find out the peer of a veth.  This wasn't always there, but IIRC it's been around for a while. We should at least try that method.

awh:
Also
```
$ sudo nsenter --target=2468 --net ./pif ethwe
Index 14
Master 0
Peer Index 15
```

site/plugin/cni-plugin-how-to.md
```diff
@@ -0,0 +1,77 @@
...
+CNI, the [_Container Network Interface_](https://github.com/appc/cni#cni---the-container-network-interface),
+is a proposed standard for configuring network interfaces for Linux
+application containers.  CNI is supported by
+[Kubernetes](http://kubernetes.io/), an open source container cluster
+manager built by Google.
+
+###Installing the Weave Net CNI plugin
+
+The Weave Net CNI plugin is installed when you run `weave setup`, if
+it your machine has the directories normally used to host CNI plugins.
```
rade: s/it/

site/plugin/cni-plugin-how-to.md
```diff
@@ -0,0 +1,77 @@
...
+
+###Launching Weave Net
+
+To create a network which can span multiple hosts, the Weave peers must be connected to each other, by specifying the other hosts during `weave launch` or via
+[`weave connect`](/site/using-weave/finding-adding-hosts-dynamically.md).
+
+See [Deploying Applications to Weave Net](/site/using-weave/deploying-applications.md#peer-connections) for a discussion on peer connections. 
+
+As well as launching Weave Net, you have to run an extra command to
+configure the Weave bridge:
```
rade: We don't want to give the impression that `weave expose` generally configures the weave bridge. So... "... run an extra command to perform some additional configuration of the Weave bridge...".

common/utils.go
```diff
@@ -58,40 +59,87 @@ func FindNetDevs(processID int, match func(string) bool) ([]NetDev, error) {
...
+}
+
 // Lookup the weave interface of a container
 func GetWeaveNetDevs(processID int) ([]NetDev, error) {
-	return FindNetDevs(processID, func(name string) bool {
-		return strings.HasPrefix(name, "ethwe")
+	// Bail out if this container is running in the root namespace
+	nsToplevel, _ := netns.GetFromPid(1)
+	nsContainr, _ := netns.GetFromPid(processID)
+	if nsToplevel.Equal(nsContainr) {
```
rade: seems dangerous not to check for errors above. Is it really impossible for `netns.GetFromPid(1)` to fail?

weave
```diff
@@ -115,10 +115,35 @@ docker_run_options() {
     echo --privileged --net=host $(docker_sock_options)
 }
 
+docker_exec_cni_options() {
```
rade: The naming is unfortunate, especially given the existence of `docker_run_options`, since both `run` and `exec` are docker commands, and `docker_run_options` is named after the docker command but `docker_exec_cni_options` is not.

bboreham: `cni_volume_options` ?

rade: I avoided mentioning `cni` since `exec_remote()` shouldn't need to know about that.

rade: Also, that way any other commands which might need extra options can simply be handled with a `case` in `exec_remote_docker_run_options`.

bboreham: Note I needed the same `-v` flags in the CNI mini-script.
See 88cb07203aeb7d1fbe2e7d97b33a68b41719b2a7

PR #2039, opened 2016-03-11
Remove Jekyll

awh:
I'm guessing this is obsolete now?


rade:
@awh why did this get closed?

PR #2049, opened 2016-03-16
Retry docker event subscription

### LINKED ISSUE #2016: Weave proxy losing docker event updates
As mentioned at https://github.com/weaveworks/weave/issues/1775#issuecomment-190734573, [these logs](https://gist.github.com/errordeveloper/8058d847572e1786bce0) show a proxy failing to receive Docker events, and subsequent investigation shows the goroutine for `fsouza/go-dockerclient` missing, presumed exited because of unexpected error.

awh:
Fixes #2016.

### DIFF
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -4,10 +4,16 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
 
 	. "github.com/weaveworks/weave/common"
+)
+
+const (
+	InitialInterval = 1 * time.Second
+	MaxInterval     = 60 * time.Second
 )
 
 // An observer for container events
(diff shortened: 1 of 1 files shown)


common/docker/client.go
```diff
@@ -72,21 +78,29 @@ func (c *Client) Info() string {
...
 	go func() {
-		for event := range events {
-			switch event.Status {
-			case "start":
-				id := event.ID
-				ob.ContainerStarted(id)
-			case "die":
-				id := event.ID
-				ob.ContainerDied(id)
+		retryInterval := InitialInterval
```
bboreham: wouldn't we want to reset the retry interval on successful connection?

awh: > wouldn't we want to reset the retry interval on successful connection?

As it stands, `AddEventListener` can only return an error if you try to subscribe the same listener twice - the actual docker socket dial and `GET /events` request happen asynchronously, and the only way we learn about either of them failing is via event channel closure.

Consequently, in this case, 'successful connection' equates to the `for ... range events` _blocking_, which does not give us a handy place to reset the interval. I did give it some thought, the best I could come up with was to record the amount of time we spend in the event loop and if it exceeds the retryInterval when the channel closes we reset. Thoughts?

awh:
> record the amount of time we spend in the event loop and if it exceeds the retryInterval when the channel closes we reset

@bboreham implemented - PTAL

common/docker/client.go
bboreham: It's not bad. Consider if we fail to connect a number of times, and the retry interval goes up to  1 minute.  Then we connect, but it fails again after 30 seconds.  We will now wait 1 minute before trying again.

However, having had various long pauses while we retried, this doesn't seem much more terrible.

I guess the back-off is just to stop killing the machine with rapid reconnects.

common/docker/client.go
```diff
@@ -4,12 +4,18 @@ import (
...
+	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
 
 	. "github.com/weaveworks/weave/common"
 )
 
+const (
+	InitialInterval = 1 * time.Second
+	MaxInterval     = 60 * time.Second
```
rade: 60 seems high. 10? 20?

bboreham: message posted simultaneously with merge

PR #2050, opened 2016-03-16
Bump template name so it will fetch latest Docker

bboreham:
At the time of writing, this gives us Docker 1.10.3.


rade:
This should have been applied to the 1.4 branch. Without it we are seeing [failures](https://circleci.com/gh/weaveworks/weave/5038) in ./601_proxy_docker_py_test.sh:
```
E   APIError: 400 Client Error: Bad Request ("client is newer than server (client API version: 1.22, server API version: 1.21)")
```

I have [cherry-picked the change  onto 1.4](https://github.com/weaveworks/weave/commit/b1ec2b788a5a581b5d3bce55d344b06a6fd88d12) to fix that.

PR #2054, opened 2016-03-17
Fix missed errors in weave script

### LINKED ISSUE #2053: weave script misses many errors
[`set -e` gets disabled in certain contexts](http://david.rothlis.net/shell-set-e/)

I've just combed the weave script for places where that has an impact...

- `configure_veth_bridged_fastdp` (via `create_veth`)
  - if adding the fastdp iface fails but adding the bridge iface
    succeeds then we end up with the datapath and bridge not being
    connected, and hence presumably mysterious overlay connection
    failures
- `configure_veth_attached_bridge` (via `create_veth`)
  - if adding the interface fails but the `ip link set` succeeds, then
    `weave attach-bridge` succeeds even though the bridge hasn't been
    attached.
- `create_veth` (via `with_container_netns ... setup_router_iface_...`)
  - if setting the MTUs fails then this still succeeds, so routers started with `WEAVE_NO_FASTDP=1` can end up exhibiting lower performance or connectivity issues

And then there are some places where the impact is quite light...

- `args_match` (via `start_if_necessary`)
  - if the existing container gets removed just before this is
    invoked, we still end up comparing the inspect result. This is
    fine though, since it won't match.
- `with_container_netns` (via `with_container_netns_or_die`)
  - if the container gets removed (not just stopped) just before this
    is invoked, this will print an error about the missing container
    but continue executing, usually resulting in some obscure
    follow-on error like "cannot open /hostproc//ns/net"
- `ipam_cidrs` (via `ipam_cidrs_or_die`)
  - if there is an ip (rather than net) CIDR, and claiming it fails,
    and it is followed by other CIDRS (which succeed), then the
    surrounding command succeeds. Though note that #2052 is also at play here.
- `http_call_unix` (via `proxy_addrs`)
  - if the `docker exec ... curl ...` fails but also outputs something on stdout, then `proxy_addrs` will return that. This really shouldn't happen.

The last three are also present in 1.4.x.

Strategic placement of `|| return 1` or `|| exit 1` will cure these.

There are another ~20 functions which are suffering from the same problem but which are currently safe because they are not invoked in a context that disables `set -e`.

bboreham:
Fixes #2053

Return failure from various functions when sub-operation fails.
We cannot rely on `set -e` to fail the whole script in contexts where functions are called inside conditionals.

### DIFF
--- a/weave
+++ b/weave
@@ -608,7 +608,7 @@ create_veth() {
 
     ip link show $VETHL >/dev/null 2>&1 && ip link show $VETHR >/dev/null 2>&1 && return 0
 
-    ip link add name $VETHL mtu $MTU type veth peer name $VETHR mtu $MTU
+    ip link add name $VETHL mtu $MTU type veth peer name $VETHR mtu $MTU || return 1
 
     if ! ip link set $VETHL up || ! ip link set $VETHR up || ! "$@" ; then
         ip link del $VETHL >/dev/null 2>&1 || true
@@ -672,8 +672,8 @@ init_bridged_fastdp() {
 }
 
 configure_veth_bridged_fastdp() {
-    add_iface_fastdp $DATAPATH_IFNAME
-    add_iface_bridge $BRIDGE_IFNAME
+    add_iface_fastdp $DATAPATH_IFNAME || return 1
+    add_iface_bridge $BRIDGE_IFNAME || return 1
 }
 
 ethtool_tx_off_fastdp() {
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -948,7 +948,7 @@ args_match() {
     prev_args="$(docker inspect -f '{{.Args}} {{.Config.Env}}' $1)"
     shift
     dummy_container=$(docker create "$@" 2>/dev/null) || return 1
-    dummy_args="$(docker inspect -f '{{.Args}} {{.Config.Env}}' $dummy_container)"
+    dummy_args="$(docker inspect -f '{{.Args}} {{.Config.Env}}' $dummy_container)" || return 1
```
rade: you also need one of these three lines above.

bboreham: > you also need one of these three lines above.
 Fixed, thanks.

PR #2066, opened 2016-03-18
make 'rmpeer' report on size of transferred range, and not fail

### LINKED ISSUE #2058: `weave rmpeer` fails if peer isn't in the ring
```
host1:~$ weave launch && weave consense
host2:~$ weave launch 192.168.48.11 && weave consense
host2:~$ docker $(weave config) run --rm alpine true
host2:~$ docker kill weave
weave
host1:~$ weave rmpeer host2
host1:~$
```
So far so good, but if we leave out the running of the container then the last step produces

```
host1:~$ weave rmpeer host2
No entries for peer found
```

I think that is bad for automation and ops in general, since it means that instead of a simple "peer has gone for good -> run rmpeer somewhere" we have "peer has gone for good -> check whether the peer has any entries in the ring -> run rmpeer somewhere".

(NB: It doesn't actually return a non-zero exit status....
```
host1:~$ echo $?
0
```
but that's just because of #2052)

This issue begs the question of whether rmpeer should be idempotent. That is generally a good idea for ops. But since we do not want to hang on to nicknames and peernames forever that implies rmpeer should succeed if it cannot find the specified peer. The obvious downside is that then there would be no indication to the user when they've mistyped the peer. But such an indication cannot be reliable anyway.

rade:
Closes #2058.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -403,9 +403,9 @@ func (alloc *Allocator) AdminTakeoverRanges(peerNameOrNickname string) address.C
 			return
 		}
 
-		before := alloc.space.NumFreeAddressesInRange(alloc.universe)
+		before := alloc.space.NumFreeAddresses()
 		alloc.space.AddRanges(newRanges)
-		after := alloc.space.NumFreeAddressesInRange(alloc.universe)
+		after := alloc.space.NumFreeAddresses()
 
 		alloc.gossip.GossipBroadcast(alloc.Gossip())
 
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -376,29 +376,40 @@ func (alloc *Allocator) Shutdown() {
 	<-doneChan
 }
 
-// AdminTakeoverRanges (Sync) - take over the ranges owned by a given peer.
+// AdminTakeoverRanges (Sync) - take over the ranges owned by a given
+// peer, and return how much space was transferred in the process.
 // Only done on adminstrator command.
-func (alloc *Allocator) AdminTakeoverRanges(peerNameOrNickname string) error {
-	resultChan := make(chan error)
+func (alloc *Allocator) AdminTakeoverRanges(peerNameOrNickname string) address.Count {
+	resultChan := make(chan address.Count)
 	alloc.actionChan <- func() {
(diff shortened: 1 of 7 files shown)


rade:
I have modelled the message to have an obvious association with the output of `weave status ipam`:
````
host1:~$ weave status ipam
66:03:80:3a:36:92(host1)                524288 IPs (50.0% of total) 
5e:43:87:6b:b6:2d(host3)                524288 IPs (50.0% of total) - unreachable!
host1:~$ weave rmpeer host3
524288 IPs taken over fromhost3
````

File: ipam/allocator.go [hunk 1]
```diff
@@ -376,26 +376,31 @@ func (alloc *Allocator) Shutdown() {
...
 		}
 
-		newRanges, err := alloc.ring.Transfer(peername, alloc.ourName)
-		alloc.space.AddRanges(newRanges)
-		resultChan <- err
+		before := alloc.space.NumFreeAddressesInRange(alloc.universe)
+		alloc.space.AddRanges(alloc.ring.Transfer(peername, alloc.ourName))
+		after := alloc.space.NumFreeAddressesInRange(alloc.universe)
+
+		resultChan <- int(after) - int(before)
```
rade: I hate that cast here. I cannot fathom why `NumFreeAddressesInRange` returns an `address.Offset`. The returned value does not fit any definition of 'offset' that I can think of. @bboreham @tomwilkie 

I suppose I could stick with that pattern and return an `address.Offset` from this function. Though I would still need a `Sub` or `Difference` function for offsets, returning offsets. Which is just weird.

Or perhaps leave the return type as is and have a function `Difference(address.Offset, address,Offset) int`.

site/ipam/stop-remove-peers-ipam.md
```diff
@@ -36,4 +37,4 @@ status`.
  * [Address Allocation with IP Address Management (IPAM)](/site/ipam/overview-init-ipam.md)
  * [Automatic Allocation Across Multiple Subnets](/site/ipam/allocation-multi-ipam.md)
  * [Isolating Applications on a Weave Network](/site/using-weave/isolating-applications.md)
- 
\ No newline at end of file
+
```
bboreham: what happened here?

Issue #1881, opened 2016-01-11
default init-peer-count is off-by-one when launch args include self


errordeveloper:
We currently presume that if only one peer is passed to `weave launch`, that this peer won't be itself.

When automating provisioning, It is often convenient to write less logic possible and just pass a list of all know peers, which works just fine except for the case when this length of this list is 1.

There exists logic to prevent connection to itself, it shouldn't be impossible to correct IPAM expectations on the basis of [`ErrConnectToSelf`](https://github.com/weaveworks/weave/search?utf8=%E2%9C%93&q=ErrConnectToSelf&type=Code), given only one peer is expected.

rade:
Sure it's possible, but it is completely hideous to implement.

- init-peer-count can be set explicitly; to implement this feature we need to remember whether that was done, or the init-peer-count was derived from the arg count, since we only want to "correct" it in the latter case
- the list of destinations to connect to changes dynamically - both through `weave connect|disconnect` and peer discovery; to implement this feature we need to remember what the original launch arguments are, so that when we do detect that we have connected to ourself, we only "correct" the init-peer-count if the destination address was in the original launch args.
- we must perform the correction once only. somehow.

There are probably others.

Seems to me that by far the easiest and reliable way to fix this issue is for the user to set `--init-peer-count` to the length of args minus one. After all, they (usually) know when they are feeding the peer's own address into that.

bboreham:
I think if #1721 were implemented then the desired semantics would fall out.

rade:
> I think if #1721 were implemented then the desired semantics would fall out.

I don't see how that addresses any of the points I raised.

bboreham:
> I don't see how that addresses any of the points I raised.

- It says to do it only in the absence of a specified count.
- It says to pick the count at the time of consensus, which addresses the issue of dynamic changing and performing it only once.

rade:
I see. How would we ensure that consensus is only initiated after we have (potentially) connected to ourself and discovered that we should exclude that IP from the init-peer-count calculation?

bboreham:
> ensure that consensus is only initiated after 

That is harder.

awh:
Closed by #2070.

bboreham:
#2070 was incorrect, sadly.  See https://github.com/weaveworks/mesh/pull/57

bboreham:
Filed #2481 to cover

PR #2070, opened 2016-03-19
Use current target list to compute quorum

### LINKED ISSUE #1721: `weave launch` and `weave connect` should result in same init-peer-count
```
weave launch host1 host2 host3... hostn
```
should result in the same `init-peer-count` as
```
weave launch host1
weave connect host2
weave connect host3 ... hostn
```

i.e. the peer count for consensus should be based on the number of direct targets *at the time the consensus is initiated*.

This behaviour is less surprising than the current one, making the creation of a network with `weave launch` and `weave connect` equivalent, provided no IP address allocation occurs before all `connect`s have been issued.

We should also consider reducing the `init-peer-count` of any pending consensus when `weave disconnect` is invoked.

bboreham:
Do it by calling out into a function which has access to the list, rather than pushing updates in as suggested in #1721.

Fixes #1721 and #1881

The fix for #1881 is racy - if we decide to go for consensus before hearing back that the address refers to ourselves then we will still count ourself.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -64,6 +64,7 @@ type Allocator struct {
 	ticker            *time.Ticker
 	shuttingDown      bool // to avoid doing any requests while trying to shut down
 	isKnownPeer       func(mesh.PeerName) bool
+	quorum            func() uint
 	now               func() time.Time
 }
 
@@ -73,13 +74,20 @@ type Config struct {
 	OurNickname string
 	Seed        []mesh.PeerName
 	Universe    address.Range
-	Quorum      uint
+	IsObserver  bool
+	Quorum      func() uint
 	Db          db.DB
 	IsKnownPeer func(name mesh.PeerName) bool
 }
 
 // NewAllocator creates and initialises a new Allocator
 func NewAllocator(config Config) *Allocator {
+	var participant paxos.Participant
+	if config.IsObserver {
+		participant = paxos.NewObserver()
+	} else {
+		participant = paxos.NewNode(config.OurName, config.OurUID, 1)
+	}
 	return &Allocator{
 		ourName:     config.OurName,
 		seed:        config.Seed,
@@ -87,9 +95,10 @@ func NewAllocator(config Config) *Allocator {
 		ring:        ring.New(config.Universe.Start, config.Universe.End, config.OurName),
 		owned:       make(map[string][]address.CIDR),
 		db:          config.Db,
-		paxos:       paxos.NewParticipant(config.OurName, config.OurUID, config.Quorum),
+		paxos:       participant,
 		nicknames:   map[mesh.PeerName]string{config.OurName: config.OurNickname},
 		isKnownPeer: config.IsKnownPeer,
+		quorum:      config.Quorum,
 		dead:        make(map[string]time.Time),
 		now:         time.Now,
 	}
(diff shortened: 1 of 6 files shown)


bboreham:
Note the third commit requires a change to Mesh - https://github.com/weaveworks/mesh/pull/26

prog/weaver/main.go
```diff
@@ -372,7 +372,8 @@ func createAllocator(router *mesh.Router, ipRangeStr string, defaultSubnetStr st
 		OurUID:      router.Ourself.Peer.UID,
 		OurNickname: router.Ourself.Peer.NickName,
 		Universe:    ipRange.Range(),
-		Quorum:      quorum,
+		IsObserver:  isObserver,
+		GetQuorum:   func() uint { return determineQuorum(peerCount, router) },
```
rade: 'get' is frowned upon in goland. so I'd just leave the field as `Quorum`

rade:
How confident are you that this cannot deadlock? `determineQuorum` is called from the `ipam.Allocator` actor and requires the `mesh.ConnectionMaker` actor to respond.

awh:
@bboreham this PR updates the mesh dependency to point to a commit which is not in the master repository on github - did you mean to push a branch with a change there also?

bboreham:
I don't have write access to weaveworks/mesh.  However I got the green light to merge.

bboreham:
I updated the dependency and rebased.

ipam/allocator.go
```diff
@@ -603,6 +612,7 @@ func (alloc *Allocator) establishRing() {
 	}
 
 	alloc.awaitingConsensus = true
+	alloc.paxos.SetQuorum(alloc.getQuorum())
```
awh: I think we should INFO log the quorum value at this point

awh:
Should we make `paxos.Propose()` assert that quorum > 0 for safety now you don't have to specify it as an argument to `NewNode`?

test/700_status_and_report_test.sh
```diff
@@ -22,6 +22,7 @@ assert_raises "weave_on $HOST1 status peers | grep nicknamington"
 weave_on $HOST1 connect 10.2.2.1
 assert "weave_on $HOST1 status targets" "10.2.2.1"
 assert "weave_on $HOST1 status connections | tr -s ' ' | cut -d ' ' -f 2" "10.2.2.1:6783"
+weave_on $HOST1 forget 10.2.2.1
```
awh: Can do `weave consense` immediately after launch instead of this

awh:
Looking good. Did you see @rade's comment above (now marked as obsolete) about `GetQuroum`?

ipam/allocator.go
bboreham: We do: `determineQuorum` has `Log.Println("Assuming quorum size of", quorum)`

It won't be printed if it's been set by `--init-peer-count`

bboreham:
> How confident are you that this cannot deadlock? `determineQuorum` is called from the `ipam.Allocator` actor and requires the `mesh.ConnectionMaker` actor to respond.

The ipam and mesh actors are quite separate, connected only by gossip and by the call-out functors `isKnownPeer` and `quorum`.  So as long as incoming gossip is on a separate goroutine the risk seems low.

bboreham:
Rebased to pick up test fix.

rade:
re deadlock potential...

The allocator actor thread calls determineQuorum while holding no locks. That in turn calls cm.Targets, which requires the cm actor thread to respond.

The cm actor thread can take RLock on Peers. That is the *only* code which has the potential to stall that thread.

So logically this is no different from the allocator thread attempting to take an RLock on Peers directly.

Such an attempt can stall if and only if another thread has requested, or actually holds, a WLock on Peers. The former can be stalled by any thread holding a Peers RLock.

So the question is whether any thread calls into the allocator (requiring its actor loop to respond) while holding a Peers RLock or WLock. Note that this doesn't have to be a direct call; it could
happen indirectly via some other thread. That is rather difficult to determine.

PR #2080, opened 2016-03-21
Correct IPAM state docs

### LINKED ISSUE #2061: IPAM status documentation out of date
The documentation of the IPAM status is completely out of date.

http://docs.weave.works/weave/latest_release/troubleshooting.html#weave-status says
> Consensus: achieved

but the field is now called `Status` and there is no such value as `achieved`.

http://docs.weave.works/weave/latest_release/ipam.html#initialisation talks about three states - deferred, waiting, achieved. What gets actually displayed is one of the following:

- `all IP ranges owned by unreachable peers - use 'rmpeer' if they are dead`
- `waiting for IP range grant from peers`
- `ready`
- `awaiting consensus (quorum: X, known: Y)`
- `idle`

This got broken in #1702, where the above status display was introduced but none of the wording in the docs was changed. So this has been wrong since 1.4.0.

awh:
Fixes #2061.

### DIFF
--- a/site/ipam.md
+++ b/site/ipam.md
@@ -48,17 +48,6 @@ time, you can give the number of peers like this:
 
     host3$ weave launch --init-peer-count 3 $HOST2
 
-The consensus mechanism used to determine a majority transitions
-through three states: 'deferred', 'waiting' and 'achieved':
-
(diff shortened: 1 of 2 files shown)


rade:
LGTM

Issue #2023, opened 2016-03-06
Stale WeaveDNS entries


bboreham:
As reported at https://github.com/weaveworks/weave/issues/1976#issuecomment-192985795

rade:
One possible reason is #1867. But let's assume we don't hit that...

Say we have a container C on peer X. Now somebody stops X, removes C, and restarts X.

Another peer Y might see the following event stream

- removal of peer X from topology -> Y removes DNS entry for C
- addition of peer X to the topology
- DNS entry for C@X, sent by some other peer as part of regular gossip before that peer had seen the removal of A from the topology.

Note that both #1867 and this issue have one thing in common: a peer that restarts quite quickly. There should be evidence in the logs for that.

rade:
Note that the probability of #1867 or this one occurring rises with the number nodes since it only takes one peer in the network to fail to remove a DNS entries (as in #1867) or re-create it accidentally, as per this issue, for the rest of the peers to become equally contaminated, thanks to gossip & eventual consistency semantics.

rade:
Possible fix: replace the entry PeerName with PeerUID and maintain the following invariant: for a DNS entry to exist, it must belong to a peer with a PeerUID that is present in the topology.

rade:
@bboreham has just proposed an alternative fix: get peers to tombstone entries which they receive and that purport to belong to them but for which they haven't actually got a record.

This means we wouldn't have to change the entry format.

Alas there is a problem. We have no idea what the last version of the entry was. Just because we received an entry from one node doesn't mean another node doesn't hold a higher version of that same entry.

bboreham:
But we would subsequently receive that higher version, and again send out a higher update to contradict it.

rade:
it is fundamental for convergence to maintain the invariant that no two entries must exist which have the same version but are different.

PR #2081, opened 2016-03-21
Override stale dns entries received via gossip

### LINKED ISSUE #1867: DNS entries can leak when a peer restarts quickly
When a peer restarts quickly, not all other peers necessarily see the peer go away, since topology gossip data merging may combine the removal and addition, or the events get re-ordered in transit, with the removal being skipped (since it will have an earlier version) - both of these effectively just end up updating the peer UID and version. If that happens on just a single peer then the DNS entries of the restarted peer are all retained. This is problematic in two ways:

- if any containers died on the peer while the weave router was down, their entries are leaked.
- for surviving containers, the version of the re-created entry may well be lower than the existing one, e.g. if previously the entry had been tombstoned and resurrected a few times. If the last version of the entry known by surviving peers was a tombstone, then this will effectively wipe out the re-created entry.

Steps to reproduce
```
host1$ weave launch --no-discovery
host2$ weave launch --no-discovery $HOST1
host3$ weave launch --no-discovery $HOST2

host1$ weave run --name=foo -ti alpine /bin/sh
host3$ weave dns-lookup foo
10.32.0.1

host3$ sudo iptables -A INPUT -s $HOST2 -j DROP
host1$ weave stop; docker rm -f foo; weave launch --no-discovery $HOST3
host3$ sudo iptables -D INPUT -s $HOST2 -j DROP

host3$ weave dns-lookup foo
10.32.0.1
```

bboreham:
Fixes #1867 and #2023.

Also makes the merging of tombstones stable, where everything else is equal.

### DIFF
--- a/nameserver/entry.go
+++ b/nameserver/entry.go
@@ -105,6 +105,11 @@ func (e1 *Entry) addLowercase() {
 	e1.lHostname = strings.ToLower(e1.Hostname)
 }
 
+func (e *Entry) tombstone() {
+	e.Version++
+	e.Tombstone = now()
+}
+
 func check(es SortableEntries) error {
 	if !sort.IsSorted(es) {
 		return fmt.Errorf("Not sorted!")
@@ -182,8 +187,7 @@ func (es *Entries) tombstone(ourname mesh.PeerName, f func(*Entry) bool) Entries
 	tombstoned := Entries{}
 	for i, e := range *es {
 		if f(&e) && e.Origin == ourname {
-			e.Version++
-			e.Tombstone = now()
+			e.tombstone()
 			(*es)[i] = e
 			tombstoned = append(tombstoned, e)
 		}
@@ -191,6 +195,7 @@ func (es *Entries) tombstone(ourname mesh.PeerName, f func(*Entry) bool) Entries
 	return tombstoned
 }
 
+// note f() may only modify entries such that they remain in order defined by less()
 func (es *Entries) filter(f func(*Entry) bool) {
 	defer es.checkAndPanic().checkAndPanic()
 
@@ -203,6 +208,17 @@ func (es *Entries) filter(f func(*Entry) bool) {
 		i++
 	}
 	*es = (*es)[:i]
+}
+
+func (es Entries) findEqual(e *Entry) (*Entry, bool) {
+	i := sort.Search(len(es), func(i int) bool {
+		return !es[i].insensitiveLess(e)
+	})
+	if i < len(es) && es[i].equal(*e) {
+		return &es[i], true
+	} else {
+		return nil, false
+	}
 }
 
 func (es Entries) lookup(hostname string) Entries {
--- a/nameserver/entry.go
+++ b/nameserver/entry.go
@@ -93,6 +93,9 @@ func (e1 *Entry) merge(e2 *Entry) bool {
 		e1.Version = e2.Version
 		e1.Tombstone = e2.Tombstone
 		return true
+	} else if e2.Version == e1.Version && e2.Tombstone > e1.Tombstone {
+		e1.Tombstone = e2.Tombstone
+		return true
 	}
 	return false
 }
--- a/nameserver/nameserver.go
+++ b/nameserver/nameserver.go
@@ -203,15 +203,32 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
 		return nil, nil, fmt.Errorf("clock skew of %d detected", delta)
 	}
 
-	n.Lock()
-	isKnownPeer := n.isKnownPeer
-	n.Unlock() // Don't hold the lock over call out to code elsewhere
+	// Filter to remove entries from unknown peers, done before we take
+	// the nameserver lock, so we don't have to worry what isKnownPeer locks.
 	gossip.Entries.filter(func(e *Entry) bool {
-		return isKnownPeer(e.Origin)
+		return n.isKnownPeer(e.Origin)
 	})
 
 	n.Lock()
 	defer n.Unlock()
+
+	// Check entries claiming to originate from us against our current data
+	gossip.Entries.filter(func(e *Entry) bool {
+		if e.Origin == n.ourName {
+			if ourEntry, ok := n.entries.findEqual(e); ok {
+				if ourEntry.Version < e.Version ||
+					(ourEntry.Version == e.Version && ourEntry.Tombstone != e.Tombstone) {
+					// Take our version of the data, but make the version higher than the incoming
+					nextVersion := e.Version + 1
+					*e = *ourEntry
+					e.Version = nextVersion
+				}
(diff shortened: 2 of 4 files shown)


nameserver/entry.go
```diff
@@ -105,6 +108,11 @@ func (e1 *Entry) addLowercase() {
 	e1.lHostname = strings.ToLower(e1.Hostname)
 }
 
+func (e *Entry) makeTombstone() {
```
rade: s/makeTombstone/tombstone

nameserver/nameserver.go
```diff
@@ -204,10 +204,30 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
 	}
 
 	n.Lock()
+	isKnownPeer := n.isKnownPeer
+	n.Unlock() // Don't hold the lock over call out to code elsewhere
```
rade: why bother locking at all? isKnownPeer doesn't change.

nameserver/nameserver.go
```diff
@@ -204,10 +204,30 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
...
 
+	// Check entries claiming to originate from us against our current data
 	gossip.Entries.filter(func(e *Entry) bool {
-		return n.isKnownPeer(e.Origin)
+		if e.Origin == n.ourName {
+			if ourEntry, ok := n.entries.findEqual(e); ok {
+				if ourEntry.Version < e.Version ||
+					(ourEntry.Version == e.Version && ourEntry.Tombstone != e.Tombstone) {
+					// Take our version of the data, but make the version higher than the incoming
+					ourEntry.Version = e.Version + 1
```
rade: modifying ourEntry here means the subsequent merge won't treat the entry as changed, which delays convergence. It's also somewhat dirty.

I suggest copying the entry as is and only increasing the version on the copy.

bboreham:
I have covered all the points raised, but note https://github.com/weaveworks/mesh/issues/28 means the data does not propagate until periodic gossip fires.

nameserver/nameserver.go
```diff
@@ -203,11 +203,30 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
...
+	})
+
 	n.Lock()
 	defer n.Unlock()
 
+	// Check entries claiming to originate from us against our current data
 	gossip.Entries.filter(func(e *Entry) bool {
-		return n.isKnownPeer(e.Origin)
+		if e.Origin == n.ourName {
+			n.infof("Received entry like mine: %v", e)
```
rade: you'll either want to take that out or drop the level to 'debug'.

nameserver/nameserver.go
```diff
@@ -203,11 +203,30 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
...
+			if ourEntry, ok := n.entries.findEqual(e); ok {
+				if ourEntry.Version < e.Version ||
+					(ourEntry.Version == e.Version && ourEntry.Tombstone != e.Tombstone) {
+					// Take our version of the data, but make the version higher than the incoming
+					higherVersion := e.Version
+					*e = *ourEntry
+					e.Version = higherVersion + 1
```
rade: `e.Version++`?

bboreham: No, because we've just copied our old version over it, which is at least one behind.

rade: ah, I completely misread what this was doing. maybe, just maybe, doing `higherVersion := e.Version + 1` would be clearer.

nameserver/nameserver.go
```diff
@@ -220,13 +239,15 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
 // merge received data into state and return "everything new I've
 // just learnt", or nil if nothing in the received data was new
 func (n *Nameserver) OnGossip(msg []byte) (mesh.GossipData, error) {
+	n.infof("Received gossip")
```
rade: left over debug logging.

nameserver/nameserver.go
```diff
@@ -220,13 +239,15 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
...
 	newEntries, _, err := n.receiveGossip(msg)
 	return newEntries, err
 }
 
 // merge received data into state and return a representation of
 // the received data, for further propagation
-func (n *Nameserver) OnGossipBroadcast(_ mesh.PeerName, msg []byte) (mesh.GossipData, error) {
+func (n *Nameserver) OnGossipBroadcast(s mesh.PeerName, msg []byte) (mesh.GossipData, error) {
+	n.infof("Received gossip broadcast from %s", s)
```
rade: left over debug logging.

bboreham:
Updated again, with smoke-test working.  See https://github.com/weaveworks/mesh/pull/29

mesh/local_peer.go
```diff
@@ -149,7 +149,10 @@ func (peer *LocalPeer) handleAddConnection(conn Connection) error {
 	}
 	_, isConnectedPeer := peer.router.Routes.Unicast(toName)
 	peer.addConnection(conn)
-	if isConnectedPeer {
+	if isRestartedPeer {
+		conn.Log("connection added (restarted peer)")
+		peer.router.SendAllGossipDown(conn)
+	} else if isConnectedPeer {
```
rade: use a `switch`.

rade: also, you may want to track this as a separate issue, for release notes purposes, since it goes beyond DNS.

nameserver/nameserver.go
```diff
@@ -203,11 +203,29 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
 		return nil, nil, fmt.Errorf("clock skew of %d detected", delta)
 	}
 
+	gossip.Entries.filter(func(e *Entry) bool {
```
rade: deserves a comment as to why we are doing this here rather than as part of the filter below.

nameserver/nameserver.go
```diff
@@ -226,7 +244,7 @@ func (n *Nameserver) OnGossip(msg []byte) (mesh.GossipData, error) {
 
 // merge received data into state and return a representation of
 // the received data, for further propagation
-func (n *Nameserver) OnGossipBroadcast(_ mesh.PeerName, msg []byte) (mesh.GossipData, error) {
+func (n *Nameserver) OnGossipBroadcast(s mesh.PeerName, msg []byte) (mesh.GossipData, error) {
```
rade: superfluous change?

mesh/local_peer.go
bboreham: tracked at #2085

rade:
Let's leave out the mesh changes; too dangerous for 1.4.6.

bboreham:
Removed mesh changes, squashed commits

Issue #2018, opened 2016-03-03
Setup AWS for system tests run by CircleCI


brb:
* Add a shell script which would setup EC2 machines for running `test/*`.
* Add changes to `circle.yml` in order to use AWS for running the tests. 

Blocks #2017

brb:
What's left:

- [ ] Allow to connect to `dockerd` (via tcp) only for trusted parties.
- [ ] Parallelize tests when running on AWS (i.e. extend `scheduler`).
- [ ] Extend `scheduler` to gc EC2 instances (currently it is done by `shutdown -h :30` which is run locally by instances).

bboreham:
You could perhaps enable the TCP connection to docker for just the IP address that the test is connecting from.

Issue #2017, opened 2016-03-03
Add underlay network implementation based on AWS VPC


brb:
In a case when all Weave peers run on AWS within the same VPC, we could rely on AWS VPC for packet routing instead of using custom overlays.

Most of the work is done, the following is missing:

- [x] Cleanup VPC route table upon peer termination.
- [x] Change the block donation algo strategy to the one which tries to find such block that its donation would require minimal amount of changes in the route table.
- [x] Simplify grouping of IP ranges: merge ranges before calling `HandleUpdate`, match only those ranges which differ.
- [x] Error handling for special situations: the 50 routes limit has been reached, etc.
- [x] System tests on the AWS infra.
- [ ] Write docs, guides, etc.
- [x] Simplify system setup (route table detection, setting proxy_arp, etc.)
- [ ] Consider subnet isolation: no container can be reached from a host running it.
- [x] Avoid using `fatal` loglevel when possible.

Nice to have:

- [ ] Make update ops of the route table asynchronous (i.e., move to a separate goroutine).
- [ ] Improve semantics for choosing a peer for donations.

This issue does not cover changes which would be required in order to make the AWS VPC overlay compatible with the existing overlay implementations (in particular, with `sleeve` and `fastdp`).

bboreham:
Enhancements moved to #2322

PR #2091, opened 2016-03-23
[WIP] 2017: AWSVPC

brb:
**NOTE: DO NOT MERGE**

I am opening this PR to get your early feedback.

The PR covers partly #2017 and #2018


ipam/ring/ring.go
```diff
@@ -351,11 +396,24 @@ func (r *Ring) AllRangeInfo() (result []RangeInfo) {
...
 
+	if isCIDRAligned {
+		r.createEntriesCIDR(peers)
+		return
+	}
+	r.createEntries(peers)
+	r.Seeds = peers
+}
+
+// createEntriesCIDR subdivides the [from,to) (CIDR) range for the given peers
```
bboreham: createEntriesCIDR -> createEntries

ipam/allocator.go
```diff
@@ -66,18 +67,22 @@ type Allocator struct {
...
+	OurName       mesh.PeerName
+	OurUID        mesh.PeerUID
+	OurNickname   string
+	Seed          []mesh.PeerName
+	Universe      address.Range
+	IsObserver    bool
+	Quorum        func() uint
+	Db            db.DB
+	IsKnownPeer   func(name mesh.PeerName) bool
+	IsCIDRAligned bool
```
bboreham: I suggest this is made non-optional.
It will make the allocation of ranges less fair in many cases, but that doesn't seem like a big problem, and it will reduce the code bloat.

ipam/allocator.go
```diff
@@ -623,7 +641,12 @@ func (alloc *Allocator) establishRing() {
 
 func (alloc *Allocator) createRing(peers []mesh.PeerName) {
 	alloc.debugln("Paxos consensus:", peers)
-	alloc.ring.ClaimForPeers(normalizeConsensus(peers))
+	alloc.ring.ClaimForPeers(normalizeConsensus(peers), alloc.isCIDRAligned)
+	// We assume that the peer has not possessed any address ranges before
+	err := alloc.monitor.HandleUpdate(nil, alloc.ring.OwnedAndMergedRanges())
```
bboreham: Suggest moving the various `HandleUpdate` calls inside `ringUpdated`.
Though the one in `Shutdown()` needs different handling - we don't want to retry pending ops there.

ipam/monitor/awsvpc.go
```diff
@@ -0,0 +1,265 @@
...
+		if err != nil {
+			return fmt.Errorf("deleteHostRoute failed: %s", err)
+		}
+	}
+
+	return nil
+}
+
+func (mon *AWSVPCMonitor) String() string {
+	return "awsvpc"
```
bboreham: why?

ipam/monitor/awsvpc.go
```diff
@@ -0,0 +1,265 @@
...
+	instancesResp, err := mon.ec2.DescribeInstances(instancesParams)
+	if err != nil {
+		return nil, fmt.Errorf("DescribeInstances failed: %s", err)
+	}
+	if len(instancesResp.Reservations) == 0 ||
+		len(instancesResp.Reservations[0].Instances) == 0 {
+		return nil, fmt.Errorf(
+			"cannot find %s instance within reservations", mon.instanceID)
+	}
+	vpcID := instancesResp.Reservations[0].Instances[0].VpcId
```
bboreham: what's going on here?

.gitignore
```diff
@@ -63,3 +63,6 @@ test/tls/tls
 test/tls/*.pem
 test/coverage
 test/coverage.*
+
+# Ctags
+tags
```
bboreham: Is this relevant to the AWSVPC change?

ipam/monitor/common.go
```diff
@@ -0,0 +1,16 @@
+package monitor
```
bboreham: hate "common" as a name.  how about `monitor.go` ?

ipam/space/space_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
```
bboreham: stray blank line?

weave
```diff
@@ -676,6 +717,7 @@ init_bridge() {
 
     ip link set dev $BRIDGE address $MAC
 
+
```
bboreham: stray blank line?

ipam/monitor/awsvpc.go
brb: Used here: https://github.com/weaveworks/weave/blob/issues/2017-awsvpc-wip/ipam/status.go#L17 which is used to detect whether we use VPC when attaching containers in https://github.com/weaveworks/weave/blob/issues/2017-awsvpc-wip/weave#L1014

ipam/monitor/awsvpc.go
brb: AWS VPC Route Table ID detection. Perhaps, I should comment this method properly.

ipam/space/space_test.go
brb: No really:
```
import (
   $STDLIB_PKGS

   $EXTERNAL_PKGS

   $LOCAL_PKGS
)
```

bboreham: Right, but it's unrelated to this PR. Just do it on master.

weave
```diff
@@ -1717,13 +1773,26 @@ launch_router() {
...
+                ;;
             *)
                 ARGS="$ARGS '$(echo "$1" | sed "s|'|'\"'\"'|g")'"
                 ;;
         esac
         shift
     done
     eval "set -- $ARGS"
+
+    create_bridge
```
bboreham: Moving create_bridge after the arguments are evaluated breaks the use of `DOCKER_BRIDGE_IP` on line 1735.

weave
```diff
@@ -1717,13 +1773,26 @@ launch_router() {
...
+    docker_bridge_ip
```
bboreham: This is actually unnecessary because `create_bridge` calls it, but leaving it where it was would have avoided breaking line 1735.

weave
```diff
@@ -1759,11 +1828,13 @@ launch_router() {
         --ipalloc-range "$IPRANGE" \
         --dns-effective-listen-address $DOCKER_BRIDGE_IP \
         $DNS_ROUTER_OPTS $NO_DNS_OPT \
+        $AWSVPC_ARGS \
         --http-addr $HTTP_ADDR \
         "$@"
     setup_router_iface_$BRIDGE_TYPE
     wait_for_status $CONTAINER_NAME http_call $HTTP_ADDR
     [ -n "$ALREADY_RUNNING_CONTAINER" ] || populate_router
+    [ -z "$AWSVPC" ] || expose
```
bboreham: `>/dev/null` to avoid spurious printing of the address

bboreham:
Need to adjust `weave status` so it doesn't leave the connections as pending:

    Connections: 1 (1 pending)
          Peers: 2 (with 2 pending connections)

Maybe by instantiating awsvpc as a different kind of connection, as discussed.

test/aws.sh
```diff
@@ -0,0 +1,411 @@
...
+        --source-group "$SEC_GROUP_NAME"            \
+        --protocol all
+    $AWSCLI ec2 authorize-security-group-ingress    \
+        --group-name "$SEC_GROUP_NAME"              \
+        --protocol tcp --port 22                    \
+        --cidr "0.0.0.0/0"
+    $AWSCLI ec2 authorize-security-group-ingress    \
+        --group-name "$SEC_GROUP_NAME"              \
+        --protocol tcp --port 2375                  \
+        --cidr "0.0.0.0/0"
```
bboreham: This is too dangerous - allowing Docker access from anywhere on the Internet

brb: Yep, see the open issues of https://github.com/weaveworks/weave/issues/2018

weave
```diff
@@ -952,17 +996,33 @@ attach() {
         connect_container_to_bridge $CONTAINER_IFNAME || return 1
     fi
 
+    if [[ -n "$AWSVPC" && $(echo "$@" | wc -w) -gt 1 ]]; then
+        echo "AWSVPC is enabled, therefore only one address can be attached to" >&2
```
bboreham: I think this restriction shouldn't be necessary.

weave
brb: Do you mean that it breaks this line https://github.com/weaveworks/weave/pull/2091/files#diff-ed6579eaedd0e67b3ed51c7ee7071044R1735 ?

If yes, I don't see how it breaks. `CONTAINER_PORT` is used only when starting the router.

brb: Re redundant `docker_bridge_ip` call: I don't want to introduce too many changes to the `weave` script in this PR, so I'll get rid of it in a separate PR.

weave
brb: It's already done in https://github.com/weaveworks/weave/pull/2091/files#diff-ed6579eaedd0e67b3ed51c7ee7071044R2085

weave
brb: Yep, this is due to a way we detect subnet CIDR (`CIDR` at line 1009). Anyway, fixing it and getting rid of the limitation.

prog/weaver/Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM scratch
+FROM gliderlabs/alpine
```
bboreham: I'd like to discuss this

weave
bboreham: This line breaks: 

    DNS_ROUTER_OPTS="--dns-listen-address $DOCKER_BRIDGE_IP:53"

prog/weaver/main.go
```diff
@@ -245,7 +248,8 @@ func main() {
 		networkConfig.PacketLogging = nopPacketLogging{}
 	}
 
-	overlay, bridge := createOverlay(datapathName, ifaceName, config.Host, config.Port, bufSzMB)
+	overlay, bridge := createOverlay(datapathName, ifaceName, config.Host, config.Port, bufSzMB,
+		useAWSVPC)
```
rade: same line, please

prog/weaver/main.go
```diff
@@ -297,7 +301,8 @@ func main() {
 		defaultSubnet address.CIDR
 	)
 	if ipamConfig.Enabled() {
-		allocator, defaultSubnet = createAllocator(router, ipamConfig, db, isKnownPeer)
+		allocator, defaultSubnet = createAllocator(router, ipamConfig, db,
+			isKnownPeer, useAWSVPC)
```
rade: same line please

prog/weaver/main.go
```diff
@@ -410,7 +421,14 @@ func createOverlay(datapathName string, ifaceName string, host string, port int,
 	return overlay, bridge
 }
 
-func createAllocator(router *weave.NetworkRouter, config ipamConfig, db db.DB, isKnownPeer func(mesh.PeerName) bool) (*ipam.Allocator, address.CIDR) {
+func createAllocator(router *weave.NetworkRouter, config ipamConfig, db db.DB,
+	isKnownPeer func(mesh.PeerName) bool, useAWSVPC bool) (*ipam.Allocator, address.CIDR) {
```
rade: same line please

brb: but it's super long line...

rade: I am aware of that.

Related, always put function args last.

circle.yml
```diff
@@ -36,14 +39,22 @@ test:
...
-        parallel: true
-    - cd $SRCDIR/test; eval $(./gce.sh hosts); export COVERAGE=true; ./run_all.sh:
-        parallel: true
+    #- cd $SRCDIR/test; ./gce.sh setup && eval $(./gce.sh hosts) && ./setup.sh:
+    #    parallel: true
+    #- cd $SRCDIR/test; eval $(./gce.sh hosts); export COVERAGE=true; ./run_all.sh:
+    #    parallel: true
+    #    timeout: 300
+    - cd $SRCDIR/test; ./aws.sh setup && eval $(./aws.sh hosts) && ./setup.sh:
+        parallel: false
```
bboreham: Why `parallel false` ?  This means we only run smoke-tests on one side of the build, which slows down the overall completion.

circle.yml
```diff
@@ -22,7 +23,9 @@ dependencies:
     - bin/setup-circleci-secrets "$SECRET_PASSWORD"
     - "mkdir -p $(dirname $SRCDIR) && cp -r $(pwd)/ $SRCDIR"
     - "cd $SRCDIR; git submodule update --init"
-    - cd $SRCDIR/test; ./gce.sh make_template:
+    #- cd $SRCDIR/test; ./gce.sh make_template:
```
bboreham: Remove commented-out line

brb: Will do that before a final commit(s).

weave
```diff
@@ -654,6 +655,48 @@ EOF
...
+                RCIDR=$(ip route list exact $CIDR proto kernel | head -n1 | cut -d ' ' -f1)
+                [ -n "$RCIDR" ] && ip route del dev $BRIDGE proto kernel $RCIDR
+            fi
+        fi
+        [ -n "$FQDN" ] && when_weave_running put_dns_fqdn_no_check_alive weave:expose $FQDN $CIDR
+    done
+    show_addrs $ALL_CIDRS
+}
+
+report() {
```
bboreham: Why did this move?  Is it left-over from some other change?

brb: Yep, moving it back.

ipam/testutils_test.go
```diff
@@ -128,7 +129,7 @@ type mockDB struct{}
 func (d *mockDB) Load(_ string, _ interface{}) (bool, error) { return false, nil }
 func (d *mockDB) Save(_ string, _ interface{}) error         { return nil }
 
-func makeAllocator(name string, cidrStr string, quorum uint) (*Allocator, address.CIDR) {
+func makeAllocator(name string, cidrStr string, quorum uint, mon monitor.Monitor) (*Allocator, address.CIDR) {
```
bboreham: If you introduce the `WithMonitor` variant here, then you don't have to go round adding the null one in so many places.

net/address/address.go
```diff
@@ -1,12 +1,21 @@
...
 
 	"github.com/weaveworks/weave/common"
 )
 
+const (
+	CIDRMaxPrefixLen = 32
+)
+
+var (
+	ErrCIDRTooSmall = errors.New("CIDR is too small")
```
bboreham: Not used?  (Similar feature went into master)

ipam/monitor/awsvpc.go
```diff
@@ -0,0 +1,258 @@
...
+
+// Helpers
+
+// removeCommon filters out CIDR ranges which are contained in both a and b slices.
+func removeCommon(a, b []address.CIDR) (newA, newB []address.CIDR) {
+	i, j := 0, 0
+
+	for i < len(a) && j < len(b) {
+		switch {
+		case a[i].Start() == b[j].Start() && a[i].End() == b[j].End():
```
bboreham: `a[i].Equal(b[j])` ? (assuming a suitable implementation of `Equal()`)

net/address/address.go
```diff
@@ -91,6 +145,21 @@ func ParseCIDR(s string) (CIDR, error) {
...
+	}
+	return cidrs
+}
+
+func (cidr CIDR) Start() Address {
+	return cidr.Addr
+}
+
+func (cidr CIDR) End() Address {
+	return Add(cidr.Addr, cidr.Size()-1)
```
bboreham: The `End` of a Range is exclusive, so it is confusing to use the same name for an inclusive end.

Since this `End()` is only used to compare against other calls of the same function, it won't matter which you use.

However, I would probably just indirect via `Range()` and get rid of this function.

net/address/address.go
```diff
@@ -23,6 +32,18 @@ func (r Range) Size() Count                { return Length(r.End, r.Start) }
 func (r Range) String() string             { return fmt.Sprintf("%s-%s", r.Start, r.End-1) }
 func (r Range) Overlaps(or Range) bool     { return !(r.Start >= or.End || r.End <= or.Start) }
 func (r Range) Contains(addr Address) bool { return addr >= r.Start && addr < r.End }
+func (r Range) Equals(or Range) bool       { return r.Start == or.Start && r.End == or.End }
+
+// IsCIDR checks whether the range is CIDR.
+func (r Range) IsCIDR() bool {
```
bboreham: This is only called in an `assert` and it seems to me that if you change `ipam.Config.Universe` to a CIDR then you don't ever need to do this check.

prog/weaver/main.go
```diff
@@ -245,7 +248,7 @@ func main() {
 		networkConfig.PacketLogging = nopPacketLogging{}
 	}
 
-	overlay, bridge := createOverlay(datapathName, ifaceName, config.Host, config.Port, bufSzMB)
+	overlay, bridge := createOverlay(datapathName, ifaceName, config.Host, config.Port, bufSzMB, useAWSVPC)
```
bboreham: seems simpler just to put the conditional on `useAWSVPC` here, so you leave `createOverlay()` alone.

prog/weaver/main.go
```diff
@@ -410,7 +419,13 @@ func createOverlay(datapathName string, ifaceName string, host string, port int,
 	return overlay, bridge
 }
 
-func createAllocator(router *weave.NetworkRouter, config ipamConfig, db db.DB, isKnownPeer func(mesh.PeerName) bool) (*ipam.Allocator, address.CIDR) {
+func createAllocator(router *weave.NetworkRouter, config ipamConfig, db db.DB, useAWSVPC bool, isKnownPeer func(mesh.PeerName) bool) (*ipam.Allocator, address.CIDR) {
+
+	var (
+		mon = monitor.Monitor(monitor.NewNullMonitor())
```
bboreham: Create the `monitor` outside and pass it in.

ipam/monitor/awsvpc_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+func ip(s string) address.Address {
+	addr, _ := address.ParseIP(s)
+	return addr
+}
+
+// [start; end]
+func cidr(start, end string) address.CIDR {
+	c := address.Range{Start: ip(start), End: ip(end) + 1}.CIDRs()
+	common.AssertWithMsg(len(c) == 1,
```
bboreham: Why is this an `assert` not a `require` ?

ipam/ring/ring.go
```diff
@@ -374,26 +393,36 @@ func (r *Ring) ClaimForPeers(peers []mesh.PeerName) {
...
-	for i, peer := range peers {
-		if address.Count(i) == remainder {
-			share--
-			if share == 0 {
-				break
-			}
+// createEntries subdivides the [from,to) (CIDR) range for the given peers
+// and creates entries for the subranges.
+// Each entry is CIRD-aligned.
+func (r *Ring) createEntries(peers []mesh.PeerName) {
```
bboreham: Do we really need a separate function for this?

ipam/monitor/awsvpc_test.go
brb: It's used just to validate an input for the tests.

ipam/ring/ring.go
brb: I'd prefer to have a separate function to not overbloat the `ClaimForPeers` method.

ipam/monitor/awsvpc_test.go
bboreham: I can see that; I guess a better answer is that `require` would need the `testing.T` to be passed in.

ipam/monitor/awsvpc_test.go
```diff
@@ -0,0 +1,71 @@
...
+	r0to255     = cidr("10.0.0.0", "10.0.0.255")
+	r0to3       = cidr("10.0.0.0", "10.0.0.3")
+	r2to3       = cidr("10.0.0.2", "10.0.0.3")
+	r12to13     = cidr("10.0.0.12", "10.0.0.13")
+	r18to19     = cidr("10.0.0.18", "10.0.0.19")
+	r22to23     = cidr("10.0.0.22", "10.0.0.23")
+	r24to27     = cidr("10.0.0.24", "10.0.0.27")
+	r128to255   = cidr("10.0.0.128", "10.0.0.255")
+	r1dot0to255 = cidr("10.0.1.0", "10.0.1.255")
+)
```
rade: I am staring at this wondering why those particular cidrs are interesting.

net/address/address_test.go
```diff
@@ -39,3 +39,61 @@ func TestBiggestPow2AlignedRange(t *testing.T) {
...
+	expectedCIDRs := []CIDR{
+		{ip("192.168.1.42"), 31},
+		{ip("192.168.1.44"), 30},
+		{ip("192.168.1.48"), 28},
+		{ip("192.168.1.64"), 26},
+		{ip("192.168.1.128"), 25},
+		{ip("192.168.2.0"), 27},
+		{ip("192.168.2.32"), 29},
+		{ip("192.168.2.40"), 31},
+	}
```
rade: why are these addresses interesting?

router/awsvpc.go
```diff
@@ -0,0 +1,85 @@
...
+
+// FlowOp
+
+type NoopFlowOp struct{}
+
+func (noop NoopFlowOp) Process(frame []byte, dec *EthernetDecoder, broadcast bool) {}
+
+func (noop NoopFlowOp) Discards() bool {
+	return true
+}
```
rade: This type is the same as `router.DiscardingFlowOp`.

ipam/allocator.go
```diff
@@ -472,6 +481,12 @@ func (alloc *Allocator) AdminTakeoverRanges(peerNameOrNickname string) address.C
 		alloc.ringUpdated()
 		after := alloc.space.NumFreeAddresses()
 
+		// TODO(mp): Return error over resultChan
+		err = alloc.monitor.HandleUpdate(oldRanges, alloc.ring.OwnedAndMergedRanges())
+		if err != nil {
+			alloc.errorf("HandleUpdate: %s", err)
```
rade: why not "HandleUpdate failed: %s", as elsewhere?

ipam/monitor/monitor.go
```diff
@@ -0,0 +1,17 @@
+package monitor
+
+import (
+	"github.com/weaveworks/weave/net/address"
+)
+
+// Monitor is an interface for tracking changes in ring allocations.
+type Monitor interface {
```
rade: That is way too generic a name.

ipam/allocator.go
```diff
@@ -867,8 +901,13 @@ func (alloc *Allocator) donateSpace(r address.Range, to mesh.PeerName) {
 		return
 	}
 	alloc.debugln("Giving range", chunk, "to", to)
+	oldRanges := alloc.ring.OwnedAndMergedRanges()
 	alloc.ring.GrantRangeToHost(chunk.Start, chunk.End, to)
 	alloc.persistRing()
+	err := alloc.monitor.HandleUpdate(oldRanges, alloc.ring.OwnedAndMergedRanges())
+	if err != nil {
+		alloc.errorf("HandleUpdate failed: %s", err)
+	}
```
rade: There is a rather repetitive pattern in all the HandleUpdate calls, which is:

1. oldRanges := alloc.ring.OwnedAndMergedRanges()
2. manipulate ring
3. newRanges := alloc.ring.OwnedAndMergedRanges()
4. HandleUpdate(oldRanges, newRanges)

So why not have a higher-order function that does all that and takes thunk representing step 2?

ipam/allocator.go
```diff
@@ -787,15 +815,21 @@ func (alloc *Allocator) update(sender mesh.PeerName, msg []byte) error {
...
+		oldRanges = alloc.ring.OwnedAndMergedRanges()
 		updated, err := alloc.ring.Merge(*data.Ring)
 		switch err {
 		case nil:
 			if updated {
 				alloc.pruneNicknames()
+				err := alloc.monitor.HandleUpdate(oldRanges, alloc.ring.OwnedAndMergedRanges())
+				if err != nil {
+					return err
+				}
```
rade: At all other `HandleUpdate` call sites an error is logged but ignored. But not here. Why?

ipam/allocator.go
```diff
@@ -73,6 +74,7 @@ type Allocator struct {
 	isKnownPeer       func(mesh.PeerName) bool
 	quorum            func() uint
 	now               func() time.Time
+	monitor           monitor.Monitor // monitor tracks changes in address ranges owned by us
```
rade: The fact that you need a comment here is a good indication that either the type name or the var name or both are not as descriptive as they should be.

router/bridge.go
```diff
@@ -23,6 +23,10 @@ type BridgeConsumer func(PacketKey) FlowOp
 
 type NullBridge struct{}
 
+func NewNullBridge() NullBridge {
+	return NullBridge{}
+}
+
```
rade: why?

router/bridge.go
```diff
@@ -36,7 +40,7 @@ func (NullBridge) Interface() *net.Interface {
 }
 
 func (NullBridge) String() string {
-	return "<no bridge networking>"
+	return "no overlay bridge"
```
rade: why?

router/network_overlay.go
```diff
@@ -35,6 +35,10 @@ type OverlayForwarder interface {
 
 type NullNetworkOverlay struct{ mesh.NullOverlay }
 
+func NewNullNetworkOverlay() NullNetworkOverlay {
+	return NullNetworkOverlay{}
+}
+
```
rade: why?

ipam/monitor/awsvpc.go
```diff
@@ -0,0 +1,258 @@
...
+// removeCommon filters out CIDR ranges which are contained in both a and b slices.
+func removeCommon(a, b []address.CIDR) (newA, newB []address.CIDR) {
+	i, j := 0, 0
+
+	for i < len(a) && j < len(b) {
+		switch {
+		case a[i].Equal(b[j]):
+			i++
+			j++
+			continue
```
rade: the `continue` is superfluous.

ipam/monitor/awsvpc.go
```diff
@@ -0,0 +1,258 @@
...
+		case a[i].End() < b[j].End():
+			newA = append(newA, a[i])
+			i++
+		default:
+			newB = append(newB, b[j])
+			j++
+		}
```
rade: I would love to see the 'equal' case as the default, for symmetry.

ipam/monitor/awsvpc.go
```diff
@@ -0,0 +1,258 @@
...
+	common.Log.Debugf("[monitor] "+fmt, args...)
+}
+
+func (mon *AWSVPCMonitor) infof(fmt string, args ...interface{}) {
+	common.Log.Infof("[monitor] "+fmt, args...)
+}
+
+// Helpers
+
+// removeCommon filters out CIDR ranges which are contained in both a and b slices.
```
rade: this really ought to mention that the slices must be sorted.

ipam/monitor/monitor.go
```diff
@@ -0,0 +1,17 @@
...
+	// HandleUpdate is called whenever an address ring gets updated.
+	//
+	// prevRanges corresponds to ranges which were owned by a peer before
+	// a change in the ring, while currRanges to the ones which are currently
+	// owned by the peer.
+	HandleUpdate(prevRanges, currRanges []address.Range) error
+	// String returns a user-friendly name of the monitor.
+	String() string
```
rade: Having a `String()` method here seems overkill.

ipam/ring/ring.go
rade: I agree with @bboreham here; this is overkill, given that the calling function is left with just five simple lines.

weave
```diff
@@ -1596,13 +1601,24 @@ launch_router() {
...
         shift
     done
     eval "set -- $ARGS"
+
+    create_bridge
+    # We set the router name to the bridge MAC, which in turn is
+    # derived from the system UUID (if available), and thus stable
+    # across reboots.
+    PEERNAME=$(cat /sys/class/net/$BRIDGE/address)
+
```
rade: why did this get moved?

ipam/allocator.go
rade: Alternatively, associate the monitor with the ring, and let the ring take care of invoking it as necessary.

rade: I have just noticed an inconsistency: some HandleUpdate calls happen before ringUpdated/persistRing, and some after. Is there a reason?

weave
bboreham: create_bridge needs to know if `--awsvpc` mode is in operation, so it doesn't do the `ethtool tx off`.

router/bridge.go
bboreham: This string comes out in `weave report`: the change makes it accurate in both the AWSVPC case and in the (unusual) case where the router has been started with no bridge.

ipam/allocator.go
```diff
@@ -472,6 +481,12 @@ func (alloc *Allocator) AdminTakeoverRanges(peerNameOrNickname string) address.C
 		alloc.ringUpdated()
 		after := alloc.space.NumFreeAddresses()
 
+		// TODO(mp): Return error over resultChan
```
bboreham: Is this TODO still valid?

ipam/monitor/monitor.go
brb: `LocalRangesTracker` ?

ipam/ring/ring.go
```diff
@@ -348,6 +348,25 @@ func (r *Ring) OwnedRanges() (result []address.Range) {
 	return r.splitRangesOverZero(result)
 }
 
+// OwnedAndMergedRanges does the same as OwnedRanges, except that it returns
+// merged ranges.
+func (r *Ring) OwnedAndMergedRanges() (result []address.Range) {
```
rade: could this *replace* `OwnedRanges`?

bboreham: No, `Allocator.reportFreeSpace` needs to go over each individual range.

rade: really? That's totally non-obvious to me from a quick look at the code.

rade: @bboreham explained why. It's subtle. And really deserves a comment.

I am in two minds as to whether refactoring this code is worthwile - either introducing a helper that takes a boolean, or making this function call `OwnedRanges` and do the merging on the result.

rade: Arguably the merging should be done by `HandleUpdate`; it makes that API safer.

ipam/monitor/monitor.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+// Monitor is an interface for tracking changes in ring allocations.
+type Monitor interface {
+	// HandleUpdate is called whenever an address ring gets updated.
+	//
+	// prevRanges corresponds to ranges which were owned by a peer before
+	// a change in the ring, while currRanges to the ones which are currently
+	// owned by the peer.
+	HandleUpdate(prevRanges, currRanges []address.Range) error
```
rade: This really ought to mention that the ranges must be sorted.

ipam/ring/ring.go
brb: I'd move the merging to `HandleUpdate` not only because of the safety reason, but also because merging is needed only for awsvpc.

ipam/monitor/awsvpc.go
brb: This would introduce the additional case:
```
case a.End > b.End: <..>
case a.End < b.End: <..>
case a.Start != b.Start: <..>
default: <..>
```

ipam/allocator.go
```diff
@@ -85,23 +86,35 @@ type Config struct {
...
 		participant = paxos.NewObserver()
 	} else {
 		participant = paxos.NewNode(config.OurName, config.OurUID, 0)
 	}
 
-	return &Allocator{
+	if config.Tracker != nil {
+		updateCallback = func(prev []address.Range, curr []address.Range) {
+			err := config.Tracker.HandleUpdate(prev, curr)
+			if err != nil {
```
rade: you can save a line here by combining with previous.

ipam/ring/ring.go
```diff
@@ -368,31 +372,34 @@ func (r *Ring) AllRangeInfo() (result []RangeInfo) {
...
-
-		r.Entries.insert(entry{Token: pos, Peer: peer, Free: share})
-		pos += address.Address(share)
+		if share == 1 || len(peers) == 1 {
+			r.Entries.insert(entry{Token: from, Peer: peers[0], Free: share})
+			return
+		}
+		mid := address.Add(from, address.Offset(share/2))
+		subdivide(from, mid, peers[:len(peers)/2])
+		subdivide(address.Add(mid, address.Offset(share%2)), to, peers[len(peers)/2:])
```
rade: eek. I don't think @bboreham or I realised that we'd need a closure here. So I think this ought to be a separate function after all. Sorry.

ipam/testutils_test.go
```diff
@@ -148,11 +153,21 @@ func makeAllocator(name string, cidrStr string, quorum uint) (*Allocator, addres
...
 		Db:          new(mockDB),
 		IsKnownPeer: func(mesh.PeerName) bool { return true },
+		Tracker:     track,
 	}), cidr
 }
 
-func makeAllocatorWithMockGossip(t *testing.T, name string, universeCIDR string, quorum uint) (*Allocator, address.CIDR) {
-	alloc, subnet := makeAllocator(name, universeCIDR, quorum)
+func makeAllocatorWithMockGossip(t *testing.T, name string, universeCIDR string,
+	quorum uint) (*Allocator, address.CIDR) {
```
rade: combine with previous line.

ipam/testutils_test.go
```diff
@@ -148,11 +153,21 @@ func makeAllocator(name string, cidrStr string, quorum uint) (*Allocator, addres
...
+
+	return makeAllocatorWithMockGossipAndTracker(t, name, universeCIDR, quorum,
+		tracker.NewNullTracker())
+}
+
+func makeAllocatorWithMockGossipAndTracker(t *testing.T, name string, universeCIDR string, quorum uint,
+	track tracker.LocalRangeTracker) (*Allocator, address.CIDR) {
```
rade: combine with previous line.

ipam/testutils_test.go
```diff
@@ -214,14 +229,20 @@ func AssertNothingSentErr(t *testing.T, ch <-chan error) {
 }
 
 func makeNetworkOfAllocators(size int, cidr string) ([]*Allocator, *gossip.TestRouter, address.CIDR) {
+	return makeNetworkOfAllocatorsWithTracker(size, cidr, tracker.NewNullTracker())
+}
+
+func makeNetworkOfAllocatorsWithTracker(size int, cidr string,
+	track tracker.LocalRangeTracker) ([]*Allocator, *gossip.TestRouter, address.CIDR) {
```
rade: combine with previous line.

ipam/testutils_test.go
```diff
@@ -214,14 +229,20 @@ func AssertNothingSentErr(t *testing.T, ch <-chan error) {
...
 	gossipRouter := gossip.NewTestRouter(0.0)
 	allocs := make([]*Allocator, size)
 	var subnet address.CIDR
 
 	for i := 0; i < size; i++ {
 		var alloc *Allocator
-		alloc, subnet = makeAllocator(fmt.Sprintf("%02d:00:00:02:00:00", i),
-			cidr, uint(size/2+1))
+		alloc, subnet = makeAllocatorWithTracker(fmt.Sprintf("%02d:00:00:02:00:00", i),
+			cidr, uint(size/2+1), track)
```
rade: combine with previous line.

ipam/ring/ring.go
```diff
@@ -20,10 +20,11 @@ import (
...
 type Ring struct {
-	Start, End address.Address // [min, max) tokens in this ring.  Due to wrapping, min == max (effectively)
-	Peer       mesh.PeerName   // name of peer owning this ring instance
-	Entries    entries         // list of entries sorted by token
-	Seeds      []mesh.PeerName // peers with which the ring was seeded
+	Start, End     address.Address // [min, max) tokens in this ring.  Due to wrapping, min == max (effectively)
+	Peer           mesh.PeerName   // name of peer owning this ring instance
+	Entries        entries         // list of entries sorted by token
+	Seeds          []mesh.PeerName // peers with which the ring was seeded
+	updateCallback func(prev, curr []address.Range)
```
rade: `onUpdate`?

ipam/ring/ring.go
```diff
@@ -576,6 +584,16 @@ func (r *Ring) PeerNames() map[mesh.PeerName]struct{} {
 	return res
 }
 
+func (r *Ring) callUpdateCallback() func() {
```
rade: I would put this near the invariant assertion functions.

And perhaps name it `trackUpdates`, seeing as the function itself isn't actually invoking the callback, the returned function is.

ipam/ring/ring.go
```diff
@@ -524,7 +562,20 @@ func (r *Ring) PickPeerForTransfer(isValid func(mesh.PeerName) bool) mesh.PeerNa
...
+// Does Transfer from a dead peer
+func (r *Ring) AdminTransfer(from, to mesh.PeerName) []address.Range {
+	return r.transfer(from, to, true)
+}
+
+func (r *Ring) transfer(from, to mesh.PeerName, isAdmin bool) []address.Range {
 	r.assertInvariants()
+	defer r.trackUpdates()()
+	if isAdmin {
+		defer r.trackUpdatesOfPeer(from)()
```
brb: Alternatively, I could make this non-optional. It would make `AdminTransfer` obsolete. Opinions? @rade @bboreham

bboreham: Let me see if I follow this:

- In the ordinary mode the "giving up" peer removes its rule and the "taking over" peer adds a new rule.
- In admin mode the "taking over" peer does both actions.

So, you don't really need two different versions, you can just use the second one in both cases.

It seems attractive to say "most of the time, I only modify my own rules", but I cannot see any particular reason to prefer that.

Doing it all on the "taking over" side makes guaranteeing the ordering simpler, perhaps.

brb: Just to recap, this would make `Transfer` to call both:

```go
defer r.trackUpdates()()
defer r.trackUpdatesOfPeer(from)()
```

In the case of `alloc.Shutdown()`, the second defer statement (which is executed before the first defer statement) will do the necessary cleanup (i.e. will remove the rules). Whilst, the first defer statement will call `tracker.HandleUpdate(nil, nil)` resulting in no action.

In the case of `alloc.AdminTakeoverRanges(...)`, the second defer statement will remove non-local rules belonging to the `from` peer and afterwards, the first defer statement will add new rules for the local peer.

bboreham:
merged as #2341

PR #2095, opened 2016-03-24
Support custom docker args for weave-plugin

berlic:
`WEAVEPLUGIN_DOCKER_ARGS` environment variable is used to pass custom args when creating plugin container

Our default docker log-driver is gelf (set via DOCKER_OPTS) and all containers send logs to logstash.
But we want weave logs to stay local and be accessible via `docker logs` command.

There are `WEAVE_DOCKER_ARGS` and `WEAVEPROXY_DOCKER_ARGS` variables, but no variable for plugin container.


bboreham:
Thanks for the contribution!

Issue #2090, opened 2016-03-23
proxy can fail to re-attach restarted containers (640_proxy_restart_reattaches_test failure)


brb:
The output from the build https://circleci.com/gh/weaveworks/weave/5115 :
```
>>> Test ./640_proxy_restart_reattaches_test.sh finished after 84.4 secs with error: exit status 1
Proxy restart reattaches networking to containers
59f0768f7c59fe50b7c231aca844983f724cd75ed1aff57c48287b344ddc7621
f32c5b5654850e6424b3619dcb7fac44dbad3e8b0d63b50f6dfa6173f0cbe22b
c2
weave
sudo: systemctl: command not found
host4-5115-0.us-central1-a.positive-cocoa-90213> docker stop/waiting
host4-5115-0.us-central1-a.positive-cocoa-90213> docker start/running, process 18202
Waiting for proxy to start
Waiting for proxy to start
120348873ce8f40d196db51c4a8457684320a401bcc082460e09fd875280223b
e313062d93bd0638d3bf9943659aad4d15004d76b3d8905e02101c3fa4e21e0d
test #7 "exec_on host4-5115-0.us-central1-a.positive-cocoa-90213 c2 grep ^1$ /sys/class/net/ethwe/carrier" failed:
	program terminated with code 1 instead of 0
test #8 "exec_on host4-5115-0.us-central1-a.positive-cocoa-90213 c1 getent hosts seetwo.weave.local | grep -q '10.32.0.1'" failed:
	program terminated with code 1 instead of 0
test #9 "exec_on host4-5115-0.us-central1-a.positive-cocoa-90213 c1 getent hosts 10.32.0.1 | tr -s ' ' | tr '[:upper:]' '[:lower:]'" failed:
	expected "10.32.0.1 seetwo.weave.local"
	got nothing
3 of 14 tests failed in 83.633s.
```

The test script has a check for the existence of the `systemctl` command. However, for some reasons `systemctl` gets executed regardless the check.

rade:
I think the systemcl issue is a red herring. I reckon on our test VMs that *always* gets printed. Looks like there is something weird going on with user vs sudo envs, which makes systemctl available to the user but not in sudo. BUT it falls back to upstart, which still works, as evidenced by the "docker stop" and "docker start" lines.

I've just committed a change that should get rid of the "command not found" message.

The real cause of the test failure is the failing tests, as indicated.

bboreham:
Now seems to be consistently failing:

````
[...]

PR #2096, opened 2016-03-24
Report errors when weave script talks to weave component via http

### LINKED ISSUE #2052: errors returned by weave http api are ignored by weave script
curl returns a 0 exit status for HTTP errors (as opposed to connectivity, etc errors):
```
$ curl --connect-timeout 3 -s -S -X POST -d peer=1 http://127.0.0.1:6784/connect
lookup 1: invalid domain name
$ echo $?
0
```
And just to check that we are indeed returning an error code:
```
$ curl -D - --connect-timeout 3 -s -S -X POST -d peer=1 http://127.0.0.1:6784/connect
HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Wed, 16 Mar 2016 22:38:50 GMT
Content-Length: 30

lookup 1: invalid domain name
```

The curl behaviour can be changed with `-f`, but that doesn't print the returned response:
```
$ curl -f --connect-timeout 3 -s -S -X POST -d peer=1 http://127.0.0.1:6784/connect
curl: (22) The requested URL returned error: 400 Bad Request
$ echo $?
22
```

**user impact**

- claiming individual IP addresses in `ipam_cidrs` (called from `weave run|start|attach|expose`; `weave attach` is also called by the proxy)

    ```
    $ weave run ip:10.32.0.1/24 -ti alpine /bin/sh
    978c65ca3593a33a28eafc48029a331949a0033d4ae177326bb7eeb27615485a
    $ weave run ip:10.32.0.1/24 -ti alpine /bin/sh
    Unable to claim: address 10.32.0.1/24 is already owned by 978c65ca3593a33a28eafc48029a331949a0033d4ae177326bb7eeb27615485a
    9a59bd62937a5448770c0c7d553a041a51f23b3a06287a7543897355377d70dc
    $ echo $?
    0
    ```
- reclaiming IP addressed in `weave restart`. This will fail in the same way as the example above if the IP got pinched by another container during the restart.
- `weave rmpeer`

    ````
     $ weave rmpeer foo
     Cannot find peer 'foo'
     $ echo $?
     0
    ````
- `weave connect|forget`

    ````
     $ weave connect 1
     lookup 1: invalid domain name
     $ echo $?
     0
    ````
- `weave report`

    ````
    $ weave report -f '{{'
    template: format:1: unexpected unclosed action in command
    $ echo $?
    0
    ````

**unaffected**

- `weave status` - we handle invalid sub-commands specially, and this really shouldn't fail for other reasons
- `weave consense` - can't fail
- reclaiming individual IP addresses for all containers, and the bridge, when the router (re)starts. This generally shouldn't fail with an HTTP error, so this is fine.
- waiting for router|proxy|plugin status on launch. The status call generally shouldn't fail with an HTTP error, so this is fine.
- obtaining proxy address for the likes of `weave env|config`. The status call generally shouldn't fail with an HTTP error, so this is fine.
- allocating IP addresses through `ipam_cidrs` (called from `weave run|start|attach|expose`; `weave attach` is also called by the proxy) and looking up IP addresses through ipam_cidrs (called from `weave detach|hide`). We catch errors here by checking that the output is a CIDR. So this is fine.
- releasing IP addresses in `weave detach|hide` - this should not go wrong, since we first perform a lookup, and hence should only be releasing owned addresses.

I am tentatively putting this into 1.4.6, since this is pretty bad for scripting weave, though I doubt we can think of a low-risk fix for this. @awh suggested replacing curl with a weaveutil function. Or perhaps do the replacement at the slightly higher level of `call_weave`, `wait_for_status` et al.

bboreham:
Fixes #2052.

The output from `curl` is placed in a temporary file, which is sent to stdout if the result is 2xx and stderr otherwise, except for 404 errors which are specifically handled in various places so `call_weave` swallows the "404 not found" message.

The "Call to weave failed." message is removed, in favour of whatever actual error came back from `curl`.  It may be that this makes some troubleshooting harder.

Also fixes an issue introduced by #1200 if the user has disabled IPAM with the (undocumented) `--ipalloc-range=""` whereby each fixed IP address specified would receive a "404" message, and this PR would otherwise have turned that into a failure of the entire command.

### DIFF
--- a/weave
+++ b/weave
@@ -1050,16 +1050,18 @@ http_call() {
     shift 3
     CURL_TMPOUT=/tmp/weave_curl_out_$$
     HTTP_CODE=$(curl -o $CURL_TMPOUT -w '%{http_code}' --connect-timeout 3 -s -S -X $http_verb "$@" http://$addr$url) || return $?
-    if [ "$HTTP_CODE" = "404" ] ; then # Swallow the message for 404
-        retval=4
-    # anything else not beginning with a 2 is an error: send the message to stderr
-    elif [ "$HTTP_CODE" = "${HTTP_CODE#2}" ] ; then
-        [ -f $CURL_TMPOUT ] && cat $CURL_TMPOUT >&2
-        retval=1
-    else
-        [ -f $CURL_TMPOUT ] && cat $CURL_TMPOUT
-        retval=0
-    fi
+    case "$HTTP_CODE" in
+        2??) # 2xx -> not an error; output response on stdout
+            [ -f $CURL_TMPOUT ] && cat $CURL_TMPOUT
+            retval=0
+            ;;
+        404) # treat as error but swallow response
+            retval=4
+            ;;
+        *) # anything else is an error; output response on stderr
+            [ -f $CURL_TMPOUT ] && cat $CURL_TMPOUT >&2
+            retval=1
+    esac
     rm -f $CURL_TMPOUT
     return $retval
 }
--- a/weave
+++ b/weave
@@ -1350,7 +1350,7 @@ check_overlap() {
 # with_container_addresses.
 ipam_reclaim() {
     for CIDR in $3 ; do
-        http_call $HTTP_ADDR PUT /ip/$1/$CIDR?noErrorOnUnknown=true
+        http_call $HTTP_ADDR PUT /ip/$1/$CIDR?noErrorOnUnknown=true || true
     done
 }
 
--- a/weave
+++ b/weave
@@ -1400,7 +1400,8 @@ ipam_cidrs() {
             if [ "$METHOD" = "POST" ] ; then
                 # Assignment of a plain IP address; warn if it clashes but carry on
                 check_overlap $arg || true
-                when_weave_running http_call $HTTP_ADDR PUT /ip/$CONTAINER_ID/$arg || return 1
+                # Abort on failure, but not 4 (=404), which means IPAM is disabled
+                when_weave_running http_call $HTTP_ADDR PUT /ip/$CONTAINER_ID/$arg || [ $? -eq 4 ] || return 1
             fi
             ALL_CIDRS="$ALL_CIDRS $arg"
         fi
--- a/weave
+++ b/weave
@@ -1080,8 +1080,14 @@ http_call_unix() {
 }
 
 call_weave() {
-    check_running $CONTAINER_NAME || return 1
-    http_call $HTTP_ADDR "$@"
+    TMPERR=/tmp/call_weave_err_$$
+    retval=0
+    http_call $HTTP_ADDR "$@" 2>$TMPERR || retval=$?
+    if [ $retval -ne 0 ] ; then
+        check_running $CONTAINER_NAME && cat $TMPERR >&2
+    fi
+    rm -f $TMPERR
+    return $retval
 }
 
 death_msg() {
--- a/weave
+++ b/weave
@@ -1048,7 +1048,20 @@ http_call() {
     http_verb="$2"
     url="$3"
     shift 3
-    curl --connect-timeout 3 -s -S -X $http_verb "$@" http://$addr$url
+    CURL_TMPOUT=/tmp/weave_curl_out_$$
+    HTTP_CODE=$(curl -o $CURL_TMPOUT -w '%{http_code}' --connect-timeout 3 -s -S -X $http_verb "$@" http://$addr$url) || return $?
+    if [ "$HTTP_CODE" = "404" ] ; then # Swallow the message for 404
+        retval=4
+    # anything else not beginning with a 2 is an error: send the message to stderr
+    elif [ "$HTTP_CODE" = "${HTTP_CODE#2}" ] ; then
+        [ -f $CURL_TMPOUT ] && cat $CURL_TMPOUT >&2
+        retval=1
+    else
+        [ -f $CURL_TMPOUT ] && cat $CURL_TMPOUT
+        retval=0
+    fi
+    rm -f $CURL_TMPOUT
+    return $retval
 }
 
 http_call_unix() {
@@ -1066,10 +1079,7 @@ http_call_unix() {
 
 call_weave() {
     check_running $CONTAINER_NAME || return 1
-    if ! http_call $HTTP_ADDR "$@" ; then
-        echo "Call to $CONTAINER_NAME failed." >&2
-        return 1
-    fi
+    http_call $HTTP_ADDR "$@"
 }
 
 death_msg() {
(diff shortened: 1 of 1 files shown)


rade:
I rebased this locally and came across the following problems:

```
$ weave status foo
404 page not found
$
```

```
$ weave launch --no-dns
$ weave dns-args
--dns 172.17.0.1 --dns-search=weave.local.
```

That last one should return nothing post #2101. The code needs adjusting since it currently matches on a "404 ..." response text.

weave
```diff
@@ -1972,20 +1981,10 @@ EOF
...
-                    usage
-                    ;;
-                *)
-                    echo "$status"
-                    ;;
-            esac
-        else
-            res=1
+        call_weave GET $STATUS_URL || res=$?
+        if [ $res = 404 ] ; then
```
rade: I don't think return values can be greater than 255. The [POSIX spec](http://pubs.opengroup.org/onlinepubs/009695399/utilities/return.html) says:
> If the value of n is greater than 255, the results are undefined.

rade: So perhaps we should

- assign the http status to a well-known var, e.g. `HTTP_STATUS`
- return a specific error code if the status is not in the 2xx range.

rade: hah, i've just seen your "148 = 404 mod 256" trick above. `if [ $res -eq 148 ] ; ...` in the above does indeed work. But, as I pointed out, this relies on undefined behaviour.

weave
```diff
@@ -1372,30 +1379,32 @@ ipam_cidrs() {
...
+            IPAM_CIDRS="$IPAM_CIDRS $CIDR"
+            ALL_CIDRS="$ALL_CIDRS $CIDR"
         else
             if [ "$METHOD" = "POST" ] ; then
                 # Assignment of a plain IP address; warn if it clashes but carry on
                 check_overlap $arg || true
-                when_weave_running http_call $HTTP_ADDR PUT /ip/$CONTAINER_ID/$arg >&2 || return 1
+                output=$(when_weave_running http_call $HTTP_ADDR PUT /ip/$CONTAINER_ID/$arg 2>&1) || retval=$?
+                # 148 (=404 mod 256) means IPAM is disabled
+                if [ $retval != 0 -a $retval != 148 ] ; then
```
rade: you really should be doing *numeric*, not string, comparisons here.

rade:
> sometimes generates an extra "404" message on the user's screen under error conditions, e.g. if they mis-spell a `weave status xxx` command.

That needs fixing.

weave
bboreham: Just to underline the "undefined" nature, `weave status foo` gives usage as expected when run with Ubuntu's `/bin/sh` but not with Alpine's

bboreham:
Rebased and addressed comments.

The `dns_args` code probably still needs work: if the `GET` fails other than with a 404 it will carry on.

rade:
I have tweaked the code slightly. Feel free to squash.

Should `ipam_reclaim` be changed? I can never work out what it does in case of failures, since that also depends on the calling context

Does it ignore errors?
Does it bail on the first error?
Does it error IFF the last reclaim fails?

weave
```diff
@@ -1066,10 +1081,7 @@ http_call_unix() {
 
 call_weave() {
     check_running $CONTAINER_NAME || return 1
-    if ! http_call $HTTP_ADDR "$@" ; then
-        echo "Call to $CONTAINER_NAME failed." >&2
-        return 1
-    fi
+    http_call $HTTP_ADDR "$@"
```
rade: The ""Call to $CONTAINER_NAME failed." error message here provided some (though arguably not enough) context. Without it, we can get weave commands to fail with errrors like
```
curl: (7) Couldn't connect to server
```
which isn't particularly illuminating.

OTOH, I guess we don't want to show an error message in some (all?) 404 cases. Hmm.

bboreham:
I changed `ipam_reclaim`: it should attempt to claim as many addresses as it can rather than bailing out on the first failure.

It now never returns a failure status for the overall `weave launch`, which seems justifiable - weave has launched, even though some sub-parts did not complete.

weave
bboreham: Is it easy to generate that failure?  We have just checked that `weave` is running, which is the most obvious case.

rade: That check is racy though...

So how about
```diff
diff --git a/weave b/weave
index 6d8c014..16f97c5 100755
--- a/weave
+++ b/weave
@@ -1080,8 +1080,17 @@ http_call_unix() {
 }
 
 call_weave() {
-    check_running $CONTAINER_NAME || return 1
-    http_call $HTTP_ADDR "$@"
+    TMPERR=/tmp/call_weave_err_$$
+    if ! http_call $HTTP_ADDR "$@" 2>$TMPERR ; then
+        retval=$?
+        if ! check_running $CONTAINER_NAME ; then
+            rm -f $TMPERR
+            return 1
+        fi
+        cat $TMPERR >&2
+        rm -f $TMPERR
+        return $retval
+    fi
 }
 ```

Comment (bboreham) [inline weave]: Change adopted, with some refinements.

Comment (bboreham) [inline weave]: This stuff is surprisingly subtle; in particular after `if ! http_call ...`, `$?` is 0, probably because it is the return code from `!`.

**Comment by rade** [comment]
This is consistently failing in CI in `640_proxy_restart_reattaches_test.sh`. Running that on my machine, I have tracked the failure down to the `check_attached` coming after
```bash
run_on $HOST1 sudo kill -KILL $(docker_on $HOST1 inspect --format='{{.State.Pid}}' c2)
```
`docker logs weaveproxy` shows
```
INFO: 2016/04/04 16:00:50.609756 Attaching container 346c87f29e0e2478eef75e47275826521e30ae6f7c5f110788c1873cc79c543d with WEAVE_CIDR "" to weave network
WARN: 2016/04/04 16:00:51.187364 Attaching container 346c87f29e0e2478eef75e47275826521e30ae6f7c5f110788c1873cc79c543d to weave network failed: Failure during network configuration for container 346c87f29e0e2478eef75e47275826521e30ae6f7c5f110788c1873cc79c543d:
Error: an inet prefix is expected rather than "cancelled".
```
That container is the `c2` container which is supposed to be restarted. Inspecting it shows...
```
        "State": {
            "Status": "exited",
            "Running": false,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": 137,
            "Error": "",
            "StartedAt": "2016-04-04T16:08:23.745210944Z",
            "FinishedAt": "2016-04-04T16:08:24.147338761Z"
        },
```

rade:
> This is consistently failing in CI in `640_proxy_restart_reattaches_test.sh`.

Our current theory is that this is just #2090 in spades.

rade:
I cannot shake off the suspicion that something else, specific to this PR, is going on here. We have had 10 successive failures in CI for this PR, which is way higher than the normal failure rate of `640_proxy_restart_reattaches_test.sh`.

It could just be that the timing has shifted... `call_weave` no longer checks whether weave is alive before calling it. That check, involving a `docker inspect`, was quite expensive. Without that delay, `weave attach` will proceed to the IP allocation much more quickly, increasing the chances that we hit the race described in #2090.

bboreham:
When working to understand this, I found the failure incredibly sensitive - adding one more line of logging before the allocate is attempted would make it succeed every time.

I see it failing due to `IsContainerNotRunning` seeing the `Restarting` flag.

PR #2100, opened 2016-03-28
make `weave dns-args` and `weave run` work with `--no-dns`

### LINKED ISSUE #2099: `weave dns-args` returns garbage when DNS is disabled, also breaks `weave run`
```
$ weave launch --no-dns
$ weave dns-args
--dns 172.17.0.1 --dns-search=404 page not found
$ weave run --name=foo -ti alpine /bin/sh
Unable to find image 'page:latest' locally
Pulling repository docker.io/library/page
docker: Error: image library/page not found.
See 'docker run --help'.
```

rade:
...rather than exploding

i.e. the former returns nothing and the latter leaves naming related container settings untouched, just like happens in the proxy and plugin.

Fixes #2099.

### DIFF
--- a/test/205_dns_disabled.sh
+++ b/test/205_dns_disabled.sh
@@ -1,0 +1,24 @@
+#! /bin/bash
+
+. ./config.sh
+
+C1=10.2.0.87
+
+start_suite "Standard behaviour when weaveDNS is disabled"
+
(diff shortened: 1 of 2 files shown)


rade:
Wrong branch. Replaced by #2101.

Issue #1687, opened 2015-11-15
disabling dns


hd-deman:
`weave launch-router $ips`
but it still run with dns
```
docker logs weave
INFO: 2015/11/15 19:24:27.242122 Command line options: map[docker-api:unix:///var/run/docker.sock dns-listen-address:172.17.0.1:53 http-addr:127.0.0.1:6784 ipalloc-range:10.32.0.0/12 datapath:weave dns-effective-listen-address:172.17.0.1 iface:veth-weave name:xx:xx:xx:xx:xx:80 nickname:johnny-cage port:6783]
INFO: 2015/11/15 19:24:27.242291 Command line peers: [52.29.xxx.xxx 52.29.xxx.xxx 52.29.xxx.xxx 52.29.xxx.xxx 52.28.xxx.xxx]
INFO: 2015/11/15 19:24:27.244608 Communication between peers is encrypted.
INFO: 2015/11/15 19:24:27.245543 Our name is xx:xx:xx:xx:xx:80(johnny-cage)
INFO: 2015/11/15 19:24:27.265093 [docker] Using Docker API on unix:///var/run/docker.sock: &[Version=1.9.0 ApiVersion=1.21 GitCommit=76d6bc9 GoVersion=go1.4.2 Os=linux Arch=amd64 KernelVersion=4.2.0-16-generic BuildTime=Tue Nov  3 17:54:53 UTC 2015]
INFO: 2015/11/15 19:24:27.265262 Assuming quorum size of 4
FATA: 2015/11/15 19:24:27.265983 Unable to start dns server: listen udp 172.17.0.1:53: bind: address already in use
```

hd-deman:
solved by `weave launch-router --no-dns $PEERS`

and `--no-dns`missed in usage docs
```
weave --help
Usage:

weave --help | help
weave setup
weave version
weave launch        [--password <password>] [--nickname <nickname>]
                      [--ipalloc-range <cidr> [--ipalloc-default-subnet <cidr>]]
                      [--no-discovery] [--init-peer-count <count>] <peer> ...
weave launch-router [--password <password>] [--nickname <nickname>]
                      [--ipalloc-range <cidr> [--ipalloc-default-subnet <cidr>]]
                      [--no-discovery] [--init-peer-count <count>] <peer> ...
```

bboreham:
Yes, it does this by default. 

There is an (undocumented) `--no-dns` option to `weave launch-router` that will stop it.

rade:
[...]

PR #2101, opened 2016-03-28
make `weave dns-args` and `weave run` work with `--no-dns`

### LINKED ISSUE #2099: `weave dns-args` returns garbage when DNS is disabled, also breaks `weave run`

rade:
...rather than exploding

i.e. the former returns nothing and the latter leaves naming related container settings untouched, just like happens in the proxy and plugin.

Fixes #2099.

### DIFF
--- a/test/205_dns_disabled.sh
+++ b/test/205_dns_disabled.sh
@@ -1,0 +1,24 @@
+#! /bin/bash
+
+. ./config.sh
+
+C1=10.2.0.87
+
+start_suite "Standard behaviour when weaveDNS is disabled"
+
+weave_on $HOST1 launch --no-dns
+
+assert "weave_on $HOST1 dns-args" ""
+
+C=$(start_container_with_dns $HOST1 $C1/24 --name=c1)
+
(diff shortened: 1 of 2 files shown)


awh:
Is the intent that `--no-dns` is now a user feature, rather than an internal flag to support multiweave? If so, should go in `weave help` and the docs...

rade:
> Is the intent that `--no-dns` is now a user feature, rather than an internal flag to support multiweave? If so, should go in `weave help` and the docs...

I am way [ahead of you](https://github.com/weaveworks/weave/issues/1687#issuecomment-202354523).

awh:
> I am way ahead of you
:thumbsup:

PR #2103, opened 2016-03-28
fixed formatting for wp  and spellcheck

abuehrle:

PR #2106, opened 2016-03-29
Patch up #2103

rade:
@abuehrle says she added the `~~~bash` because otherwise code blocks would render badly on the web site. We should fix that in the CSS, or, if necessary, wordpress translation.

### DIFF
--- a/site/ipam/troubleshooting-ipam.md
+++ b/site/ipam/troubleshooting-ipam.md
@@ -6,20 +6,18 @@ layout: default
 
 The command
 
-~~~bash
     weave status
-~~~
 
 reports on the current status of the weave router and IP allocator:
 
-~~~bash
+````
 ...
-Service: ipam
+       Service: ipam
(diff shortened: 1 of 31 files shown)


site/ipam/troubleshooting-ipam.md
```diff
@@ -13,10 +13,12 @@ The command
 reports on the current status of the weave router and IP allocator:
 
 ~~~bash
-       Service: ipam
+...
+Service: ipam
```
awh: Needs left-pad :smile:

site/ipam/troubleshooting-ipam.md
```diff
@@ -6,20 +6,18 @@ layout: default
...
 The command
 
-~~~bash
     weave status
-~~~
 
 reports on the current status of the weave router and IP allocator:
 
-~~~bash
+````
```
awh: The `Service: ipam` line in this block needs indenting

PR #2116, opened 2016-03-30
add coverage to plugin

rade:


test/830_cni_plugin_test.sh
```diff
@@ -4,12 +4,19 @@
 
 start_suite "Test CNI plugin"
 
+coverage_args() {
+    # no coverage reporting for now; but when the plugin binary has
```
bboreham: Is the beginning of this comment correct?

rade: yes. We are not collecting coverage from CNI command invocations, which is all this test does.

PR #2122, opened 2016-04-04
Update weaveworks/mesh to send gossip to restarted peer

### LINKED ISSUE #2085: Restarted peers do not get initial gossip
As detailed at https://github.com/weaveworks/mesh/issues/28.

bboreham:
Particularly to bring in https://github.com/weaveworks/mesh/pull/29

Note the test now has a `sleep 32` in it, to take us past a periodic gossip interval.  This is a bit nasty, arising from the fact that we don't immediately broadcast corrections to DNS gossip information received.

Fixes #2085

### DIFF
--- a/common/logging.go
+++ b/common/logging.go
@@ -3,6 +3,7 @@ package common
 import (
 	"bytes"
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


rade:
I am not keen on a test with a 32s pause.

Looking at the DNS code, it would appear to be fairly easy to introduce a broadcast of the records that "re-assert a peers authority".

I am happy to see that go into a separate issue, and merge this as-is in the meantime.

bboreham:
> introduce a broadcast

#2133

PR #2124, opened 2016-04-04
make DNS tombstoning idempotent

rade:
so we avoid the unnecessary gossip, and delayed convergence, that otherwise ensues.

### DIFF
--- a/nameserver/entry.go
+++ b/nameserver/entry.go
@@ -109,9 +109,13 @@ func (e1 *Entry) addLowercase() {
 	e1.lHostname = strings.ToLower(e1.Hostname)
 }
 
-func (e1 *Entry) tombstone() {
-	e1.Version++
+func (e1 *Entry) tombstone() bool {
+	if e1.Tombstone > 0 {
+		return false
+	}
 	e1.Tombstone = now()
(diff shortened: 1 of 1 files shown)


nameserver/entry.go
```diff
@@ -190,8 +194,7 @@ func (es *Entries) tombstone(ourname mesh.PeerName, f func(*Entry) bool) Entries
 
 	tombstoned := Entries{}
 	for i, e := range *es {
-		if f(&e) && e.Origin == ourname {
-			e.tombstone()
+		if f(&e) && e.Origin == ourname && e.tombstone() {
```
bboreham: Slight mis-match between the logging and the behaviour now: the sub-functions of `ContainerDied()` and `Delete()` will log "tombstoning entry" when it is already tombstoned.

bboreham:
Apart from logging note, LGTM

nameserver/entry.go
rade: That was the case already.

I did actually investigate what it would take to change the log messages so they are only emitted on the transition to tombstone. It's a non-trivial change, so not worth it.

PR #1954, opened 2016-02-04
Periodically check for up to date versions of weave.

tomwilkie:
Every 6 hours, hit https://checkpoint-api.weave.works looking for current versions of Weave. If a new version is found, log it.

Reuses the same library as Hashicorp's products.

[Review on Reviewable](https://reviewable.io/reviews/weaveworks/weave/1954)

PR #2131, opened 2016-04-05
2089: Add info about new available versions to the `weave status` report

### LINKED ISSUE #2089: Include result of version check in `weave status`
This information is going only into the logs at the moment, and is easy to miss.

brb:
The `AvailableVersion` field denotes the latest available version which is retrieved via go-checkpoint.
If checkpoint is disabled, the field will contain the `<checkpoint disabled>` value.

E.g.:
```bash
ubuntu@host:~$ weave status

         Version: git-b7d49cdeb7db
AvailableVersion: 1.4.4
<..>
```
Fixes #2089

### DIFF
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -7,10 +7,12 @@ import (
 	"net/http"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/weaveworks/mesh"
 
+	"github.com/weaveworks/go-checkpoint"
 	"github.com/weaveworks/weave/ipam"
 	"github.com/weaveworks/weave/nameserver"
 	"github.com/weaveworks/weave/net/address"
@@ -138,7 +140,7 @@ func defTemplate(name string, text string) *template.Template {
 }
 
 var statusTemplate = defTemplate("status", `\
-        Version: {{.Version}}
+        Version: {{.Version}} ({{.VersionCheck}})
 
         Service: router
        Protocol: {{.Router.Protocol}} \
@@ -221,17 +223,51 @@ var dnsEntriesTemplate = defTemplate("dnsEntries", `\
 
 var ipamTemplate = defTemplate("ipamTemplate", `{{printIPAMRanges .Router .IPAM}}`)
 
+type VersionCheck struct {
+	Enabled     bool
+	NewVersion  string
+	NextCheckAt time.Time
+}
+
+func versionCheck() *VersionCheck {
+	v := &VersionCheck{}
+	if checkpoint.IsCheckDisabled() {
+		return v
+	}
+
+	v.Enabled = true
+	v.NewVersion = newVersion.Load().(string)
+	v.NextCheckAt = checker.NextCheckAt()
+
+	return v
+}
+
+func (v *VersionCheck) String() string {
+	switch {
+	case !v.Enabled:
+		return "version check update disabled"
+	case v.NewVersion != "":
+		return fmt.Sprintf("version %s available - please upgrade!",
+			v.NewVersion)
+	default:
+		return fmt.Sprintf("up to date; next check at %s",
+			v.NextCheckAt.Format("2006/01/02 15:04:05"))
+	}
+}
+
 type WeaveStatus struct {
-	Version string
-	Router  *weave.NetworkRouterStatus `json:"Router,omitempty"`
-	IPAM    *ipam.Status               `json:"IPAM,omitempty"`
-	DNS     *nameserver.Status         `json:"DNS,omitempty"`
+	Version      string
+	VersionCheck *VersionCheck              `json:"VersionCheck,omitempty"`
+	Router       *weave.NetworkRouterStatus `json:"Router,omitempty"`
+	IPAM         *ipam.Status               `json:"IPAM,omitempty"`
+	DNS          *nameserver.Status         `json:"DNS,omitempty"`
 }
 
 func HandleHTTP(muxRouter *mux.Router, version string, router *weave.NetworkRouter, allocator *ipam.Allocator, defaultSubnet address.CIDR, ns *nameserver.Nameserver, dnsserver *nameserver.DNSServer) {
 	status := func() WeaveStatus {
 		return WeaveStatus{
 			version,
+			versionCheck(),
 			weave.NewNetworkRouterStatus(router),
 			ipam.NewStatus(allocator, defaultSubnet),
 			nameserver.NewStatus(ns, dnsserver)}
@@ -283,5 +319,4 @@ func HandleHTTP(muxRouter *mux.Router, version string, router *weave.NetworkRout
 	defHandler("/status/peers", peersTemplate)
 	defHandler("/status/dns", dnsEntriesTemplate)
 	defHandler("/status/ipam", ipamTemplate)
-
 }
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/davecheney/profile"
@@ -27,6 +28,8 @@ import (
 )
 
 var version = "(unreleased version)"
+var checker *checkpoint.Checker
+var newVersion atomic.Value
 
 var Log = common.Log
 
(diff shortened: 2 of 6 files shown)


awh:
```
Version: 1.4 (update check disabled)
Version: 1.4 (update check every 6 hours)
Version: 1.4 (version 1.5 available)
```

rade:
Suggested output
```
Version: 1.4.4 (version check disabled)
Version: 1.4.4 (available: 1.4.4 - up to date)
Version: 1.4.4 (available: 1.4.6 - please upgrade!)
```

rade:
Docs need updating. In particular troubleshooting.md#weave-status and the README.md

prog/weaver/http.go
```diff
@@ -218,16 +220,33 @@ var dnsEntriesTemplate = defTemplate("dnsEntries", `\
...
+	IPAM             *ipam.Status               `json:"IPAM,omitempty"`
+	DNS              *nameserver.Status         `json:"DNS,omitempty"`
+}
+
+func VersionCheckInfo() string {
+	if checkpoint.IsCheckDisabled() {
+		return "version check update disabled"
+	}
+	if vsn := availableVersion.Load().(string); vsn != "" {
+		return fmt.Sprintf("version %s available - please upgrade!")
```
bboreham: missing parameter

brb:
Updated.

Please merge https://github.com/weaveworks/go-checkpoint/pull/9 first. Afterwards, I need to change the commit id of the go-checkpoint git submodule and then this PR can be merged.

Q: What's about `weave report`? I've added `VersionCheckInfo` field to it which contains the same info as it is shown in `weave status`, but I'd prefer something more computer friendly. E.g.:

* `.VersionMeta.CheckEnabled :: boolean`
* `.VersionMeta.NewVersion :: string`
* `.VersionMeta.NextCheck :: string`

brb:
Added `.VersionMeta` to `weave report`.

prog/weaver/http.go
```diff
@@ -114,6 +116,21 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 		return "disabled"
 	},
 	"trimSuffix": strings.TrimSuffix,
+	"versionMeta": func(meta *VersionMeta) string {
```
rade: could that simply be the `String()` method on `versionMeta`?

prog/weaver/http.go
```diff
@@ -217,17 +234,52 @@ var dnsEntriesTemplate = defTemplate("dnsEntries", `\
...
+	NextCheckAt  time.Time
+}
+
 type WeaveStatus struct {
-	Version string
-	Router  *weave.NetworkRouterStatus `json:"Router,omitempty"`
-	IPAM    *ipam.Status               `json:"IPAM,omitempty"`
-	DNS     *nameserver.Status         `json:"DNS,omitempty"`
+	Version     string
+	VersionMeta *VersionMeta               `json:"Version,omitempty"`
```
rade: won't that end up with two `Version` keys in the json, i.e. one for the `Version` field and one for the `VersionMeta` field (which you've mapped to the 'Version' field)?

prog/weaver/http.go
```diff
@@ -217,17 +234,52 @@ var dnsEntriesTemplate = defTemplate("dnsEntries", `\
 
 var ipamTemplate = defTemplate("ipamTemplate", `{{printIPAMRanges .Router .IPAM}}`)
 
+type VersionMeta struct {
```
rade: `VersionCheck`?

prog/weaver/main.go
```diff
@@ -68,7 +76,7 @@ func checkForUpdates() {
 	}
 	resp, err := checkpoint.Check(&params)
 	handleResponse(resp, err)
-	checkpoint.CheckInterval(&params, updateCheckPeriod, handleResponse)
+	checker = checkpoint.CheckInterval(&params, updateCheckPeriod, handleResponse)
```
rade: As discussed, we should change `CheckInterval` s.t. it performs a check immediately. That way we don't need the `Check` call, can get rid of the intermediate vars for `handleResponse` and `params`, and, most importantly, don't need to run this function in its own goroutine, thus ensuring that `checker` will always be set, which in turn will simplify the status reporting code.

prog/weaver/http.go
```diff
@@ -114,6 +116,21 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 		return "disabled"
 	},
 	"trimSuffix": strings.TrimSuffix,
+	"versionMeta": func(meta *VersionMeta) string {
+		if meta == nil {
+			return ""
+		}
```
rade: I don't think it can ever be nil.

brb: Theoretically, it can. E.g., when that goroutine https://github.com/weaveworks/weave/blob/issues/2089-vsn-check/prog/weaver/main.go#L194 gets scheduled after the first `weave status|report` request.

prog/weaver/main.go
brb: ACK.

prog/weaver/http.go
rade: but that cannot happen after the other changes we discussed.

prog/weaver/http.go
brb: Looks better. I'm replacing all `[vV]ersionMeta` occurrences by `[vV]ersionCheck`.

prog/weaver/http.go
brb: Typo, thanks. I'm adding tests for this PR.

prog/weaver/http.go
brb: Yes, after changing the `CheckInterval` function, it cannot happen.

prog/weaver/http.go
brb: Yep, thanks.

brb:
Updated.

brb:
@rade ready for review.

bboreham:
Looks like it needs to be rebased to resolve conflicts on master

test/config.sh
```diff
@@ -44,6 +44,8 @@ CHECK_ETHWE_MISSING="test ! -d /sys/class/net/ethwe"
 
 DOCKER_PORT=2375
 
+CHECKPOINT_DISABLE=true
```
bboreham: does this deserve its own variable?

test/config.sh
```diff
@@ -118,7 +120,7 @@ weave_on() {
     host=$1
     shift 1
     [ -z "$DEBUG" ] || greyly echo "Weave on $host:$DOCKER_PORT: $@" >&2
-    CHECKPOINT_DISABLE=true DOCKER_HOST=tcp://$host:$DOCKER_PORT $WEAVE "$@"
+    CHECKPOINT_DISABLE="$CHECKPOINT_DISABLE" DOCKER_HOST=tcp://$host:$DOCKER_PORT $WEAVE "$@"
```
bboreham: I think this deserves a comment; presumably you are aiming at the `weave launch` commands?

bboreham: There is at least one `weave launch` that doesn't go through this path - in test 140.

test/config.sh
bboreham: oh, I get it - you override it in test 700

bboreham:
Apart from comments LGTM

brb:
Rebased.

test/config.sh
brb: This change does not affect the 140 test. The `CHECKPOINT_DISABLE` thing was present prior this PR.

test/config.sh
brb: Yep.

test/config.sh
brb: https://github.com/weaveworks/weave/pull/1954 has introduced `CHECKPOINT_DISABLE` in `weave_on`.

prog/weaver/http.go
```diff
@@ -116,6 +118,19 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
...
+		return "version check update disabled"
+	case v.NewVersion != "":
+		return fmt.Sprintf("version %s available - please upgrade!",
+			v.NewVersion)
+	default:
+		return fmt.Sprintf("up to date; next check at %s",
+			v.NextCheckAt.Format("2006/01/02 15:04:05"))
+	}
+}
+
```
rade: This is out of place. Should come just after the `versionCheck()` function.

prog/weaver/http.go
```diff
@@ -221,17 +236,38 @@ var dnsEntriesTemplate = defTemplate("dnsEntries", `\
...
-	IPAM    *ipam.Status               `json:"IPAM,omitempty"`
-	DNS     *nameserver.Status         `json:"DNS,omitempty"`
+	Version      string
+	VersionCheck *VersionCheck              `json:"VersionCheck,omitempty"`
+	Router       *weave.NetworkRouterStatus `json:"Router,omitempty"`
+	IPAM         *ipam.Status               `json:"IPAM,omitempty"`
+	DNS          *nameserver.Status         `json:"DNS,omitempty"`
+}
+
+func versionCheck() *VersionCheck {
```
rade: this function should come right after the `VersionCheck` type def.

prog/weaver/http.go
brb: Fixed, thanks.

PR #2133, opened 2016-04-06
Immediately broadcast DNS records that "re-assert a peer's authority"

bboreham:
Further improvement to the issue raised at #2081 - don't wait for periodic gossip to get the data corrected on other peers.

Authoritaah

### DIFF
--- a/nameserver/nameserver.go
+++ b/nameserver/nameserver.go
@@ -208,8 +208,8 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
 		return n.isKnownPeer(e.Origin)
 	})
 
+	var overriddenEntries []Entry
 	n.Lock()
-	defer n.Unlock()
 
 	// Check entries claiming to originate from us against our current data
 	gossip.Entries.filter(func(e *Entry) bool {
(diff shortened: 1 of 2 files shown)


nameserver/nameserver.go
```diff
@@ -221,15 +221,20 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
 					nextVersion := e.Version + 1
 					*e = *ourEntry
 					e.Version = nextVersion
+					entriesToBroadcast = append(entriesToBroadcast, *e)
 				}
 			} else { // We have no entry matching the one that came in with us as Origin
 				e.tombstone()
+				entriesToBroadcast = append(entriesToBroadcast, *e)
```
rade: that should only be done with `e.tombstone()` returns true.

bboreham: fixed, thanks

nameserver/nameserver.go
```diff
@@ -221,15 +221,20 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
...
 			}
 		}
 		return true
 	})
 
 	newEntries := n.entries.merge(gossip.Entries)
+	n.Unlock() // unlock before attempting to broadcast
+	n.broadcastEntries(entriesToBroadcast...)
```
rade: There is a subtlety here... the entries we are broadcasting are (potentially updated) entries from the incoming gossip, rather than entries we hold. It just so happens that they also happen to correspond to identical entries in the latter, post merge. Which is rather important. That's the subtlety.

Might deserve a comment.

nameserver/nameserver.go
```diff
@@ -208,8 +208,8 @@ func (n *Nameserver) receiveGossip(msg []byte) (mesh.GossipData, mesh.GossipData
 		return n.isKnownPeer(e.Origin)
 	})
 
+	var entriesToBroadcast []Entry
```
rade: it would be nice if that variable name described what those entries *are* rather than what we end up doing with them.

`newOrUpdatedEntriesOriginatingFromUs`. urgh.

bboreham: `overridenEntries` ?

rade: +1

bboreham: changed.

PR #2134, opened 2016-04-06
Better log and status messages for Paxos observer

### LINKED ISSUE #2063: Peers launched with `--observer` will show a status of 'idle' even after they have sent a ring request
Electing peers transition from 'idle' to 'awaiting consensus' and then to 'ready' - observers however remain in the 'idle' state until they are 'ready', so it's not possible to see that they are actually awaiting the outcome of an election that they have provoked amongst their elector peers.

bboreham:
Fixes #2063

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -767,7 +767,7 @@ func (alloc *Allocator) update(sender mesh.PeerName, msg []byte) error {
 	//   (if we did so on periodic gossip we would force consensus unnecessarily)
 	// * If we are an elector (to avoid a broadcast storm of ring request messages)
 	default:
-		if _, ok := alloc.paxos.(*paxos.Node); ok && sender != mesh.UnknownPeerName {
+		if alloc.paxos.IsElector() && sender != mesh.UnknownPeerName {
 			alloc.establishRing()
 		}
 	}
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -887,8 +887,10 @@ func (alloc *Allocator) loadPersistedData() {
 	if len(alloc.seed) != 0 {
 		alloc.infof("Initialising with supplied IPAM seed")
 		alloc.createRing(alloc.seed)
-	} else {
+	} else if alloc.paxos.IsElector() {
 		alloc.infof("Initialising via deferred consensus")
+	} else {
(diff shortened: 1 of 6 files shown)


ipam/allocator.go
```diff
@@ -887,6 +887,8 @@ func (alloc *Allocator) loadPersistedData() {
 	if len(alloc.seed) != 0 {
 		alloc.infof("Initialising with supplied IPAM seed")
 		alloc.createRing(alloc.seed)
+	} else if _, ok := alloc.paxos.(*paxos.Observer); ok {
+		alloc.infof("Observer - awaiting IPAM data from another peer")
```
awh: Can we have this message in the same format as the two, e.g. 'Initialising via observation'?

ipam/paxos/participant.go
```diff
@@ -7,6 +7,12 @@ type Participant interface {
 	Propose()
 	Think() bool
 	Consensus() (bool, AcceptedValue)
+	Status() *Status
```
awh: Not thrilled about this - the current status code has been carefully separated out of the things on which information is being gathered, and all of the dependencies go from the status code to the observed structures. This introduces a dependency in the other direction

ipam/paxos/participant.go
```diff
@@ -37,3 +43,7 @@ func (observer *Observer) Think() bool {
 func (observer *Observer) Consensus() (bool, AcceptedValue) {
 	return false, AcceptedValue{}
 }
+
+func (observer *Observer) Status() *Status {
+	return &Status{}
```
awh: Again, a bit grim - the Status struct has absolutely no meaning for observers

awh:
I have an alternative proposal:

* Add a ParticipantType() method to the Participant interface returning 'elector' or 'observer'
* Use said type field used to eliminate the ugly type cast checks elsewhere
* Expose said type in the report JSON (`IPAM.ParticipantType` ?) so that templates can switch on that rather than inferring via the absence of `IPAM.Paxos.Quorum`
* Return the paxos status struct & NewStatus method to paxos.go

Thoughts?

bboreham:
We only really need the type to be a binary `IsObserver()` or something like `IsActive()`

bboreham:
I did it differently - PTAL

rade:
troubleshooting-ipam.md needs updating.

PR #2135, opened 2016-04-06
Use a data volume container for persistence of IPAM etc.

### LINKED ISSUE #2031: IPAM persistence data should not be lost on change of target peer list
After #1967 and #1971, the behaviour on shutdown and restart is much more predictable.  But if you have specified a list of peers with `weave launch host1 host2 ...`, then you change that list, `weave` will remove the old container and create a new one, thus destroying the persisted data.

Docker does not let you change the argument list on an existing container.

(It is possible to persist the data outside of the container, but that is more complicated.)

### LINKED ISSUE #2077: Retain persisted information across upgrades
Upgrading weave currently discards persistence unnecessarily - we should keep the data where possible, migrating the schema if necessary.

bboreham:
Instead of persisting inside the `weave` container, use another data-only container `weavedb`.

Fixes #2031 and fixes #2077.

We cannot use the same data-only container as we use for `weavewait`, as it needs to change when Weave is upgraded.  ~~I did choose to use the same image (i.e. `weaveworks/weaveexec`); nothing from the image is used so this should be unimportant.~~

We have to move the location of the DB file(s) to a specific directory so we can mount a volume on top of that directory.  ~~This has the effect of making `docker run weave` fail unless you supply such a volume, or a different `--db-prefix`.~~


awh:
Looks good - I wonder whether we should create a minimal image to serve as the base for this volume container though? Otherwise we'll be keeping an old version (currently ~70mb) of the `weavexec` image lying around. A minimal image is only 28 bytes:

```
vagrant@vagrant-ubuntu-wily-64:~/minvol$ cat Dockerfile 
FROM scratch
ADD . /weavedb
vagrant@vagrant-ubuntu-wily-64:~/minvol$ docker build -t foo/bar .
Sending build context to Docker daemon 2.048 kB
Step 1 : FROM scratch
 ---> 
Step 2 : ADD . /weavedb
 ---> Using cache
 ---> 7367c98030d5
Successfully built 7367c98030d5
vagrant@vagrant-ubuntu-wily-64:~/minvol$ docker images foo/bar
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
foo/bar             latest              7367c98030d5        9 minutes ago       28 B
```

awh:
Do we need to put anything in the docs for this now? I'm going to mention it when talking about persistence in the operational guide, does it need to go anywhere else? (make sure you rebase on recent master if you do because I restructured all the docs for wordepress)

bboreham:
That's a good point about the old image.  I was following the Docker page on volumes where they note that using the same image doesn't cost any disk space because the layers are shared with the containers we do use, but that won't be true after an upgrade.

brb:
Might be irrelevant, but when we do `weave reset`, do we want to get rid of the data container?

awh:
> Might be irrelevant, but when we do weave reset, do we want to get rid of the data container?

Already implemented!

bboreham:
I added a minimal image as suggested.  Don't think you wanted to do `ADD . /weavedb`

Docs: I'll wait and see what you put in the operational guide.

prog/weaver/main.go
```diff
@@ -160,7 +160,7 @@ func main() {
 	mflag.StringVar(&dnsConfig.EffectiveListenAddress, []string{"-dns-effective-listen-address"}, "", "address DNS will actually be listening, after Docker port mapping")
 	mflag.StringVar(&datapathName, []string{"-datapath"}, "", "ODP datapath name")
 	mflag.StringVar(&trustedSubnetStr, []string{"-trusted-subnets"}, "", "comma-separated list of trusted subnets in CIDR notation")
-	mflag.StringVar(&dbPrefix, []string{"-db-prefix"}, "weave", "pathname/prefix of filename to store data")
+	mflag.StringVar(&dbPrefix, []string{"-db-prefix"}, "/weavedb/weave", "pathname/prefix of filename to store data")
```
brb: nit picking and not related to your changes: we should probably change the comment to "pathname of directory to store db file(s)".

bboreham: but it is not a directory

brb: yeah, I see.

brb:
LGTM, just maybe commits can be squashed.

bboreham:
Rebased/squashed.  The three commits cover separate concerns.

awh:
@bboreham and I just had a discussion about how peers specified to `weave launch` will always be ignored after the first launch now we have decoupled the persistence - proposals to fix are

* Use the union of CLI peers and what is persisted
* Support a `--replace` argument to `weave launch`

@rade thoughts?

rade:
I'd like "weave launch x y z" be equivalent to "weave launch; weave connect x; weave connect y; weave connect z" as much as possible.

So I think we want the union.

We already have a "connect --replace" iirc, so adding a "launch --replace" seems logical.

bboreham:
See #2160 for union and `--replace`

PR #2148, opened 2016-04-08
Bind-mount the docker socket for CNI plugin

### LINKED ISSUE #2146: CNI net plugin can't run weave-ipam
The plugins are generated as scripts `/opt/cni/bin/weave-net` and `weave-ipam`.  Both scripts work by `docker run`-ing the Weave plugin container.

Depending on CNI config, the `weave-net` plugin may try to call the `weave-ipam` plugin, which fails as `/var/run/docker.sock` is not bind-mounted.

bboreham:
Fixes #2146


rade:
test?

bboreham:
Test is now passing.  Needed some re-plumbing to get everything in place, including coverage args.

Issue #2139, opened 2016-04-07
TestAllocatorFuzz fails from time to time with "panic: Error doing gossip broadcast: IP allocation was seeded by different peers <..>"


brb:
```bash
panic: Error doing gossip broadcast: IP allocation was seeded by different peers (received: [00:00:00:02:00:00(nick-00:00:00:02:00:00) 03:00:00:02:00:00(nick-03:00:00:02:00:00) 04:00:00:02:00:00(nick-04:00:00:02:00:00)], ours: [00:00:00:02:00:00(nick-00:00:00:02:00:00) 01:00:00:02:00:00(nick-01:00:00:02:00:00) 02:00:00:02:00:00(nick-02:00:00:02:00:00) 03:00:00:02:00:00(nick-03:00:00:02:00:00) 04:00:00:02:00:00(nick-04:00:00:02:00:00)])

goroutine 75 [running]:
github.com/weaveworks/weave/testing/gossip.(*TestRouter).run(0xc8203cab80, 0x40000020000, 0x7f4f09e7b090, 0xc8203e4500, 0xc820319b00)
	/go/src/github.com/weaveworks/weave/testing/gossip/mocks.go:126 +0x4aa
created by github.com/weaveworks/weave/testing/gossip.(*TestRouter).Connect
	/go/src/github.com/weaveworks/weave/testing/gossip/mocks.go:161 +0xa2

<..>

goroutine 81 [chan receive]:
github.com/weaveworks/weave/ipam.(*Allocator).Allocate(0xc8203e4140, 0xc82041e2a0, 0x8, 0xa000107, 0x16, 0xb35200, 0xa34100, 0x0, 0x0)
	github.com/weaveworks/weave/ipam/_test/_obj_test/allocator.go:267 +0x1f9
github.com/weaveworks/weave/ipam.TestAllocatorFuzz.func4(0xc82041e2a0, 0x8)
	/go/src/github.com/weaveworks/weave/ipam/allocator_test.go:451 +0x15e
github.com/weaveworks/weave/ipam.TestAllocatorFuzz.func9(0x320)
	/go/src/github.com/weaveworks/weave/ipam/allocator_test.go:556 +0x146
github.com/weaveworks/weave/ipam.TestAllocatorFuzz.func8.1(0xc8203f5dd0, 0xc8, 0xc8203f5dc0, 0x4)
	/go/src/github.com/weaveworks/weave/ipam/allocator_test.go:546 +0x77
[...]

PR #2149, opened 2016-04-10
stop ipam tests from leaking routers

rade:
Discovered while investigating #2139.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -46,6 +46,7 @@ type operation interface {
 // are used around data structures.
 type Allocator struct {
 	actionChan        chan<- func()
+	stopChan          chan<- struct{}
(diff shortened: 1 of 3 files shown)


ipam/allocator.go
```diff
@@ -116,8 +116,10 @@ func (alloc *Allocator) Start() {
 // Stop makes the actor routine exit, for test purposes ONLY because any
 // calls after this is processed will hang. Async.
 func (alloc *Allocator) Stop() {
-	alloc.ticker.Stop()
-	alloc.actionChan <- nil
+	select {
```
bboreham: shouldn't you amend the comment to say something like "attempts to make the actor routine exit"?

rade: I've fixed the code rather than the comment.

PR #1978, opened 2016-02-17
Operational guide

awh:
An operational guide that describes likely deployment scenarios and administrative tasks.

Fixes #726, fixes #1102

PR #2155, opened 2016-04-12
Rationalise IPAM concepts

### LINKED ISSUE #2065: Rationalise seeding, consensus and observers
Now that we have augmented the IPAM consensus mechanism (in a somewhat piecemeal fashion) with observers and seeding, there is an opportunity to step back and attempt to reverse engineer an encompassing user model that can be used to streamline the overall command syntax, status output and documentation.

Whilst from an implementation perspective there are two 'modes' of initialisation (seeding, and consensus with electors and observers) @rade speculates that it might make more sense to elevate observation for use with either consensus or seeding, resulting in three modes:

* Seeded
* Consensed
* Observed

It is technically possible to mix all three of these in a cluster, but for the sake of sanity we should probably recommend seeded XOR consensus, with optional observers in either case.

Some considerations:

* We need a replacement name for `weave consense` - it does not make sense if it is used on an observer which is connected to seeded peers
* Determine whether `weave consense` should guarantee that a piece of the ring is owned by the calling peer in addition to obtaining a copy of the existing ring. This would make it more useful when adding a new seeded peer which is itself not part of the seed
* Determine whether `--observer` is appropriately named
* Determine whether `--ipam-seed` is appropriately named
* Determine a name for the state in which an observer has made a ring request (#2063) but has yet to receive one - analagous to 'awaiting consensus' for an elector
* Consider a unified argument to `weave launch` that selects one of the three modes
* Consider whether we use IPAM specific terminology (e.g. `--ipam-observer`) or leave things generic in case we have other uses for consensus at a later date

Things that will need updating:

* Documentation
    * IPAM troubleshooting status explanations
* Log messages (e.g. "Unable to parse IPAM seed")
* Flag help (e.g. `--observer` flag help mentions paxos)

awh:
Fixes #2065.

This is all the implementation stuff, just working on the docs now. Comments welcome!

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -56,7 +56,7 @@ type Allocator struct {
 	nicknames         map[mesh.PeerName]string  // so we can map nicknames for rmpeer
 	pendingAllocates  []operation               // held until we get some free space
 	pendingClaims     []operation               // held until we know who owns the space
-	pendingConsenses  []operation               // held until consensus achieved
+	pendingPrimes     []operation               // held while our ring is empty
 	dead              map[string]time.Time      // containers we heard were dead, and when
 	db                db.DB                     // persistence
 	gossip            mesh.Gossip               // our link to the outside world for sending messages
@@ -193,8 +193,8 @@ func (alloc *Allocator) tryOps(ops *[]operation) {
 
 // Try all pending operations
 func (alloc *Allocator) tryPendingOps() {
-	// Unblock pending consenses first
-	alloc.tryOps(&alloc.pendingConsenses)
+	// Unblock pending primes first
+	alloc.tryOps(&alloc.pendingPrimes)
 	// Process existing claims before servicing new allocations
 	alloc.tryOps(&alloc.pendingClaims)
 	alloc.tryOps(&alloc.pendingAllocates)
@@ -223,11 +223,11 @@ func (e *errorCancelled) Error() string {
 
 // Actor client API
 
-// Consense (Sync) - wait for consensus
-func (alloc *Allocator) Consense() {
+// Prime (Sync) - wait for consensus
+func (alloc *Allocator) Prime() {
 	resultChan := make(chan struct{})
-	op := &consense{resultChan: resultChan}
-	alloc.doOperation(op, &alloc.pendingConsenses)
+	op := &prime{resultChan: resultChan}
+	alloc.doOperation(op, &alloc.pendingPrimes)
 	<-resultChan
 }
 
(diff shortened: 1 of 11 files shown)


awh:
Working on docs... this looks (as anticipated) mighty odd:

```
--ipalloc-init=seed:::1
```

We could make the mode/parameter separator be `=`, and document it like this instead?

```
--ipalloc-init seed=::1
```

Even this looks better IMO:

```
--ipalloc-init=seed=::1
```

rade:
I don't like it.

awh:
Will be updating #1978 to bring it in line with these changes.

site/ipam.md
```diff
@@ -89,21 +88,23 @@ through three states: 'deferred', 'waiting' and 'achieved':
...
 
-You do not need to specify an initial peer count to such peers. This
-can be useful to add ephemeral peers to an existing fixed cluster (for
-example in response to a scale-out event) without worrying about
-adjusting initial peer counts accordingly.
+    host4$ weave launch --ipalloc-init observer $HOST3
 
-####<a name="quorum"></a> `--init-peer-count` and How Quorum is Achieved
+You do not need to specify an initial peer count or seed to such
+peers. This can be useful to add ephemeral peers to an existing fixed
```
bboreham: Can we lose the word "ephemeral"?  It's just as useful when the peers last a long time.

bboreham:
LGTM

ipam/http.go
```diff
@@ -83,7 +83,7 @@ func (alloc *Allocator) HandleHTTP(router *mux.Router, defaultSubnet address.CID
 	})
 
 	router.Methods("GET").Path("/consensus").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		alloc.Consense()
+		alloc.Prime()
```
rade: http endpoint should be changed too.

ipam/allocator.go
```diff
@@ -56,7 +56,7 @@ type Allocator struct {
 	nicknames         map[mesh.PeerName]string  // so we can map nicknames for rmpeer
 	pendingAllocates  []operation               // held until we get some free space
 	pendingClaims     []operation               // held until we know who owns the space
-	pendingConsenses  []operation               // held until consensus achieved
+	pendingPrimes     []operation               // held until consensus achieved
```
rade: comment shouldn't mention consensus. "held until ring has been established"?

awh: Established isn't enough - we need to have a copy of it too. Have gone with "held while our ring is empty"

ipam/http.go
awh: `GET /prime` is odd; also note that it doesn't say 'GET /consense'. Am going with `GET /ring`.

PR #2160, opened 2016-04-13
Union command-line peers with persisted peers

bboreham:
If #2135 is merged, we will have the paradoxical situation that the command-line peer list is ignored in favour of a persisted one.  This PR changes the default behaviour to take the union of the persisted list and the command-line list, with a `--replace` option to state you want exactly that new list.


bboreham:
Having second thoughts about this...

awh:
Superseded by #2186/#2187 => closing.

PR #2162, opened 2016-04-13
2034: Add LogHandler for logging REST API calls at server side

### LINKED ISSUE #2034: `--log-level=debug` should log REST API calls

brb:
Fixes #2034

### DIFF
--- a/common/http.go
+++ b/common/http.go
@@ -1,0 +1,18 @@
+package common
+
+import (
+	"net/http"
+)
+
+type loggingHandler struct {
+	next http.Handler
+}
+
+func (l *loggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	Log.Debugf("[http] %s %s", r.Method, r.URL.RequestURI())
+	l.next.ServeHTTP(w, r)
+}
+
+func LoggingHTTPHandler(h http.Handler) http.Handler {
+	return &loggingHandler{next: h}
+}
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -293,7 +293,7 @@ func main() {
 		}
 		router.HandleHTTP(muxRouter)
 		HandleHTTP(muxRouter, version, router, allocator, defaultSubnet, ns, dnsserver)
-		http.Handle("/", muxRouter)
+		http.Handle("/", common.LoggingHTTPHandler(muxRouter))
 		Log.Println("Listening for HTTP control messages on", httpAddr)
 		go listenAndServeHTTP(httpAddr, muxRouter)
 	}


brb:
NB: similar handler exists in the scope project. Probably, at some point it will make sense to create a separate repo for code which can be reused by both projects.

bboreham:
Please take the unnecessary change out of prog/weaver/http.go

common/http.go
```diff
@@ -0,0 +1,54 @@
...
+// ServeHTTP implements the http.Handler interface.
+func (l *logHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	writerWrapper := &logResponseWriter{
+		ResponseWriter: w,
+		status:         http.StatusOK,
+	}
+
+	start := time.Now()
+	l.handler.ServeHTTP(writerWrapper, r)
+	duration := time.Now().Sub(start)
```
bboreham: what is the motivation to include the duration?

prog/weaver/main.go
```diff
@@ -508,7 +512,7 @@ func parseIPAMSeed(ipamSeed string) []mesh.PeerName {
 	return peerNames
 }
 
-func listenAndServeHTTP(httpAddr string, muxRouter *mux.Router) {
+func listenAndServeHTTP(httpAddr string) {
```
bboreham: this looks like an unrelated change; please put it in a separate commit

bboreham:
@awh as requestor in #2034 do you agree this addresses it?

common/http.go
brb: In general, I find it useful when debugging failures happening in long chains of calls due to timeouts / timing issues. Also, it can give us a better idea on how long some requests are taking.

prog/weaver/main.go
brb: Done.

brb:
> Please take the unnecessary change out of prog/weaver/http.go

Shall I push such cosmetic changes (trailing whitespace, unnecessary empty lines, OCD, etc.) directly to master or leave them as they are?

common/http.go
bboreham: Ok, but this points up a new concern: you are logging things after they happen.
In the event of a failure we would prefer to have seen in the logs what the program was about to do.

bboreham:
Minor cosmetic changes can go straight on master.

common/http.go
brb: Good point, but I assume that in such case (when `l.handler.ServeHTTP` panics) you would be able to reconstruct what happened from a stacktrace.

Anyway, I could add two log statements - before and after. Opinions?

bboreham: My preference would be one line before.  I think we can add the duration later if we discover we care.

brb: @awh are you OK with that?

rade: I don't think we should log the duration *or status code* - failures will generally get logged already anyway.

Which means we don't need `logResponseWriter` and this whole PR will shrink down to a handful of lines. Which is nice.

prog/weaver/main.go
rade: I've actually just fixed this on master - 03afb54c - so this commit should be removed.

common/http.go
brb: Done.

brb:
PTAL.

common/http.go
```diff
@@ -0,0 +1,24 @@
...
+// $METHOD $REQUEST_URI
+func LogHandler(h http.Handler) http.Handler {
+	return &logHandler{handler: h}
+}
+
+// ServeHTTP implements the http.Handler interface.
+func (l *logHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	Log.Debugf("%s %s", r.Method, r.URL.RequestURI())
+	l.handler.ServeHTTP(w, r)
+}
```
rade: I'd call this thing `LoggingHTTPHandler` and dispense with *all* the comments above.

prog/weaver/main.go
```diff
@@ -293,7 +293,11 @@ func main() {
 		}
 		router.HandleHTTP(muxRouter)
 		HandleHTTP(muxRouter, version, router, allocator, defaultSubnet, ns, dnsserver)
-		http.Handle("/", muxRouter)
+		var handler http.Handler = muxRouter
+		if logLevel == "debug" {
+			handler = common.LogHandler(muxRouter)
+		}
+		http.Handle("/", handler)
```
rade: Is there any harm in *always* going via the logging handler? We really don't need to be concerned about efficiency here.

That way all of the above collapses to

    http.Handle("/", common.LoggingHTTPHandler(muxRouter))

brb: Apart from efficiency, none.

common/http.go
brb: Done.

rade:
I wonder whether the log lines should have a prefix, e.g. "[http]" or "[api]" or "[http api]", ... to aid in grep'ing.

@bboreham @awh @brb thoughts?

bboreham:
The `LoggingHTTPHandler` constructor could take a parameter which is the prefix, thus removing the obligation on the common part to say something generic.

rade:
@bboreham good idea. Any opinions on what the prefix should be?

bboreham:
Since we're using the same entry-point for router, IPAM and DNS control, it still has to be quite generic.  I would be ok with `"[http]"`.

brb:
Yep, prefixing each log line with "[http]" is a good idea for the sake of grepping.

However, I'm not sure whether it makes sense to extend `LoggingHTTPHandler` by adding the prefix param, because the same instance of the handler is used by all endpoints and it should stay generic.

rade:
>  the same instance of the handler is used by all endpoints and it should stay generic.

Fair point. We can always make the prefix configurable later if we feel the need.

PR #2165, opened 2016-04-14
Store 'checkAlive' flag in IPAM, so we don't delete non-container IDs

### LINKED ISSUE #2164: IPAM persistence deletes anything that's not a container ID on startup
This logic was added in 95cbc8940e4d1d7d2dc7ff0061ebdcad03f3e738, but without considering IDs like `weave:expose` which should not be deleted.

The http API does have a `checkAlive` flag which could be passed to the allocator and used later to avoid this.

bboreham:
Fixes #2164 

Note the `owned` map changes type so breaks compatibility with earlier persistence files

### DIFF
--- a/ipam/allocate.go
+++ b/ipam/allocate.go
@@ -15,6 +15,7 @@ type allocate struct {
 	resultChan       chan<- allocateResult
 	ident            string
 	r                address.CIDR // Subnet we are trying to allocate within
+	isContainer      bool
 	hasBeenCancelled func() bool
 }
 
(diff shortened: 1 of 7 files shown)


awh:
Two thoughts
1) This requires lots of change for little effect - I wonder if we shouldn't revisit your idea of detecting whether the container name is valid (`weave:expose` and `weave:extern` were carefully chosen to be the same length as a short container ID but not be valid container names or IDs; furthermore the `_` addresses which are stored using the CIDR as the container name aren't valid either). See https://github.com/docker/docker/blob/master/utils/names.go
2) If we do go this route, should `checkAlive` really be something like `isContainerAddresss`, at least in the persisted data?

awh:
So it would seem that `checkAlive` was added in the first place due to objections about inferring from magic values, so I guess we shouldn't do that here. Should consider renaming field as per (2) above. Needs rebase.

bboreham:
Renamed and rebased

PR #2166, opened 2016-04-14
Optional restart policy

awh:
All variants of `weave launch` observe `--no-restart` to disable Docker restart policy.


bboreham:
It occurs to me the code would be shorter and more flexible if the interface was:

    WEAVE_RESTART_POLICY='--restart=always' weave launch ...

PR #2172, opened 2016-04-15
Disable persistence of peer lists

bboreham:
Persistence of peer lists gives you a difference in behaviour compared with 1.4 that may lead to unexpected side-effects.  This PR disables the feature until we can think it through some more.


rade:
Some peer persistence code is still lingering around. I reckon the correct approach here is to comment out the loading and saving code (but not the call sites), with a suitable note.

bboreham:
Re-done with commenting-out

PR #2173, opened 2016-04-15
Improve IPAM status wording

awh:
Be more general about what we're waiting for - non-observer peers in the network may be seeded and not using consensus.


site/ipam/troubleshooting-ipam.md
```diff
@@ -34,6 +34,8 @@ the total allocation range and default subnet. Columns are as follows:
       peers are able to communicate amongst themselves successfully.
       The Status may also show '(observer)' instead of the quorum size,
       if this peer has been started with the `--observer` option.
+    * 'priming' - peer is an observer and is waiting to receive IPAM
+      data from seeding or consensus elsewhere in the network
```
rade: you forgot to delete the preceding sentence.

awh: Good spot, thanks

PR #2182, opened 2016-04-16
fix simulation of periodic gossip in TestTransfer

### LINKED ISSUE #2180: failure in TestTransfer
See https://circleci.com/gh/weaveworks/weave/5402
```
--- FAIL: TestTransfer (0.06s)
	Error Trace:	allocator_test.go:361
	Error:		Not equal: 0x20 (expected)
			        != 0x40 (actual)
```

rade:
`GossipBroadcast` in the `TestRouter` only goes one hop. In `TestTransfer` the three peers are connected in a chain, and hence we need to take extra care to ensure propagation to all peers.

Fixes #2180.

### DIFF
--- a/ipam/allocator_test.go
+++ b/ipam/allocator_test.go
@@ -318,10 +318,12 @@ func TestTransfer(t *testing.T) {
 	_, err = alloc3.Allocate("bar", subnet, returnFalse)
 	require.True(t, err == nil, "Failed to get address")
 
+	// simulation of periodic gossip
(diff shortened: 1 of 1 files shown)


bboreham:
The whole premise of `TestTransfer` appears to be out of date.  It talks about "master", and checks that two of the peers will be unable to get space at the beginning.

rade:
>  It talks about "master"

That is a bug

> checks that two of the peers will be unable to get space at the beginning.

no, it doesn't. you are reading the condition wrong.

bboreham:
I sit corrected.

PR #2184, opened 2016-04-17
prevent excessive ring fragmentation due to repeated space requests

### LINKED ISSUE #2181: excessive ring fragmentation
Apply
```patch
diff --git a/ipam/allocator_test.go b/ipam/allocator_test.go
index a1fa0b9..2097f22 100644
--- a/ipam/allocator_test.go
+++ b/ipam/allocator_test.go
@@ -349,6 +349,8 @@ func TestTransfer(t *testing.T) {
        alloc1.gossip.GossipBroadcast(alloc2.Gossip())
        router.Flush()
 
+       common.Log.Infof("\nring 0: %+v\nring 1: %+v\nring 2: %+v", alloc0.ring, alloc1.ring, alloc2.ring)
+
        free1 := alloc1.NumFreeAddresses(subnet.Range())
        free2 := alloc2.NumFreeAddresses(subnet.Range())
 
@@ -369,6 +371,12 @@ func TestTransfer(t *testing.T) {
        alloc0.Stop()
 }
 
+func TestTransfer1000(t *testing.T) {
+       for i := 0; i < 1000; i++ {
+               TestTransfer(t)
+       }
+}
+
 func TestFakeRouterSimple(t *testing.T) {
        const (
                cidr = "10.0.1.7/22"
```
and then run

    GO15VENDOREXPERIMENT=1 go test -v -run TestTransfer1000 ./ipam

(which may fail due to #2180, but that is not our concern here)

and inspect the output. Some test runs produce output like this:
```
ring 0: Ring [10.0.1.7, 10.0.5.7)
  10.0.1.7 -> 00:00:00:02:00:00 (v4)
  10.0.1.39 -> 02:00:00:02:00:00 (v0)
  10.0.1.71 -> 02:00:00:02:00:00 (v0)
  10.0.1.135 -> 02:00:00:02:00:00 (v0)
  10.0.2.7 -> 02:00:00:02:00:00 (v0)
  10.0.3.7 -> 01:00:00:02:00:00 (v5)
  10.0.3.135 -> 02:00:00:02:00:00 (v0)
  10.0.4.7 -> 02:00:00:02:00:00 (v1)
ring 1: Ring [10.0.1.7, 10.0.5.7)
  10.0.1.7 -> 00:00:00:02:00:00 (v4)
  10.0.1.39 -> 02:00:00:02:00:00 (v0)
  10.0.1.71 -> 02:00:00:02:00:00 (v0)
  10.0.1.135 -> 02:00:00:02:00:00 (v0)
  10.0.2.7 -> 02:00:00:02:00:00 (v0)
  10.0.3.7 -> 01:00:00:02:00:00 (v5)
  10.0.3.135 -> 02:00:00:02:00:00 (v0)
  10.0.4.7 -> 02:00:00:02:00:00 (v1)
ring 2: Ring [10.0.1.7, 10.0.5.7)
  10.0.1.7 -> 00:00:00:02:00:00 (v4)
  10.0.1.39 -> 02:00:00:02:00:00 (v0)
  10.0.1.71 -> 02:00:00:02:00:00 (v0)
  10.0.1.135 -> 02:00:00:02:00:00 (v0)
  10.0.2.7 -> 02:00:00:02:00:00 (v0)
  10.0.3.7 -> 01:00:00:02:00:00 (v5)
  10.0.3.135 -> 02:00:00:02:00:00 (v0)
  10.0.4.7 -> 02:00:00:02:00:00 (v1)
```
In other words, a ring with just three peers can get fragmented into 8 entries after just two allocations.

rade:
Fixes #2181.

### DIFF
--- a/ipam/ring/ring.go
+++ b/ipam/ring/ring.go
@@ -238,10 +238,8 @@ func (r *Ring) Merge(gossip Ring) (bool, error) {
 // entries belonging to ourPeer. Returns the merged entries and an
 // indication whether the merge resulted in any changes, i.e. the
 // result differs from the original.
-func (es entries) merge(other entries, ourPeer mesh.PeerName) (entries, bool, error) {
-	var result entries
+func (es entries) merge(other entries, ourPeer mesh.PeerName) (result entries, updated bool, err error) {
 	addToResult := func(e entry) { result = append(result, &e) }
-	var updated bool
 
 	var mine, theirs *entry
 	var previousOwner *mesh.PeerName
@@ -257,7 +255,8 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName) (entries, bool, er
 		case mine.Token > theirs.Token:
 			// insert, checking that a range owned by us hasn't been split
 			if previousOwner != nil && *previousOwner == ourPeer && theirs.Peer != ourPeer {
-				return result, false, errEntryInMyRange(theirs)
+				err = errEntryInMyRange(theirs)
+				return
 			}
 			addToResult(*theirs)
 			updated = true
@@ -268,13 +267,15 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName) (entries, bool, er
 			switch {
 			case mine.Version >= theirs.Version:
 				if mine.Version == theirs.Version && !mine.Equal(theirs) {
-					return result, false, errInconsistentEntry(mine, theirs)
+					err = errInconsistentEntry(mine, theirs)
+					return
 				}
 				addToResult(*mine)
 				previousOwner = &mine.Peer
 			case mine.Version < theirs.Version:
 				if mine.Peer == ourPeer { // We shouldn't receive updates to our own tokens
-					return result, false, errNewerVersion(mine, theirs)
+					err = errNewerVersion(mine, theirs)
+					return
 				}
 				addToResult(*theirs)
 				updated = true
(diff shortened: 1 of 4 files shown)


rade:
I suggest reviewing the individual commits of this PR rather than the whole diff, since the refactoring makes it look far bigger than it actually is.

bboreham:
There can be many reasons why the ring is updated which do not behoove a repeated request for space.  Would it not be better to have the code know when a request is in flight, and not to repeat that request unless we suspect the target is not going to respond?

rade:
> Would it not be better to have the code know when a request is in flight, and not to repeat that request unless we suspect the target is not going to respond?

That is not easy to work out and will necessarily introduce another timeout or similar constant.

I know that this PR does not eliminate repeated space requests. I said that much in #2181 ("the problem could probably be reduced significantly").

ipam/ring/ring.go
```diff
@@ -217,26 +217,47 @@ func (r *Ring) Merge(gossip Ring) (bool, error) {
...
+	}
+	r.Entries = result
+	return updated, nil
+}
+
+// Merge other entries into ours, and complain when that stomps on
+// entries belonging to ourPeer. Returns the merged entries and an
+// indication whether the merge resulted in any changes, i.e. the
+// result differs from the original.
+func (es entries) merge(other entries, ourPeer mesh.PeerName) (entries, bool, error) {
```
bboreham: Perhaps move to `entry.go` for consistency with other `entries` methods?

rade: I considered that, but the mentioning `ourPeer` and the error messages are highly ring specific.

PR #2185, opened 2016-04-17
Restructure docs

### LINKED ISSUE #2183: doc structure is illogical
The top level doc structure is illogical, and contains a few other niggles. Here it is
```
    Introducing Weave Net
    Weave Features
    Installing Weave Net
    Operational Guide
    Using Weave Net – Example
    IP Addresses, Routes and Networks
    Setting Up The Weave Net Docker API Proxy
    Using Fast Datapath
    Using the Weave Net Docker Network Plugin
    Encryption and Weave Net
    How Weave Net Works
    Address Allocation with IP Address Management (IPAM)
    Using WeaveDNS
    Using Weave with Systemd
    Weave Net FAQ
    Troubleshooting Weave Net
    Building Weave Net
```
Issues:
- `Introducing Weave Net` should just be `Introduction`
- `Weave Features` should just be `Features`
- `Installing Weave Net` should just be `Installation`
- `Operational Guide` should be `Operation`
- `Using Weave Net - Example` should be just `Using Weave Net`
- `IP Addresses, Routes and Networks` does not belong at the top level; not sure where it should go; perhaps the Ops Guide or How It Works
- `Using Fast Datapath` should not be at the top level; not sure where it should go; perhaps the Ops Guide
- `Setting Up the Weave Net Docker API Proxy` should just be `Docker API Proxy`
- `Using the Weave net Docker Nework Plugin` should just be `Docker Network Plugin`
- `Encryption and Weave Net` should just be `Encryption` and it should not be at the top level; not sure where it should go; perhaps the Ops Guide
- `How Weave Net Works` should come later
- `Address Allocation with IP Address Management (IPAM)` should just be `IP Address Allocation`
- `Using WeaveDNS` should be `Naming and Discovery`
- `Using Weave with Systemd` does not belong at the top level; Ops Guide is probably the right place for it
- `Weave Net FAQ` should just be `FAQ`
- `Troubleshooting Weave Net` should just be `Troubleshooting`

So, with all the above changes applied, here's how the top level would look like
```
    Introduction
    Features
    Installation
    Operation
    Using Weave Net
    Docker API Proxy
    Docker Network Plugin
    IP Address Allocation
    Naming and Discovery
    How Weave Net Works
    FAQ
    Troubleshooting
    Building Weave Net
```

abuehrle:
I changed the titles, as suggested in #2183, however I don't have the Ops guide, so we'll have to add that in and also move the files that go with that guide, i.e. systemd.md.   

I had broken out the encryption and the fast data path files on their own, and wasn't sure where you wanted those to go.  

I removed hard returns from the files, as the text doesn't wrap nicely in WP if they are present.  

Also, wordepress has a small issue that I will log over on that repo. 

Fixes #2183.


rade:
> I removed hard returns from the files, as the text doesn't wrap nicely in WP if they are present.

Hard returns are useful. AFAIK @awh is planning to remove them in the conversion.

Also, please don't mix completely distinct changes, e.g. in this instance the title changes and hard return changes.

rade:
I noticed there are some other changes too, e.g.

- change the question formatting in the FAQ from `###` to `**`
- remove residual 'bash' tagging on code blocks
- changed some code block formatting in the IPAM troubleshooting section and in the general troubleshooting section

Regarding that last one, it looks like this is actually a wordepress bug that should be fixed, i.e. it isn't rendering code blocks delimited with 3 single quotes.

Are there any other changes than those listed above?

abuehrle:
Yes, I removed an 'obscure' reference to IPAM in the Features Overview section. Also, the file names and the directory is called ipam, which if you don't want any reference to that term will also have to change. I did not change those. 

Re: removing Hard Breaks.  

I thought I would save some of @awh 's time that would be better spent on fixing net bugs.  And perhaps, we should all standardize on how our IDE's are set up, and where hard breaks should go.  I didn't remove all hard breaks, just the ones on paragraphs, which look really ugly in WP, and which should be soft wrapped.

rade:
>  I didn't remove all hard breaks, just the ones on paragraphs, which look really ugly in WP, and which should be soft wrapped.

I am well aware that it renders horribly now, but that is a bug in the markdown->wp conversion, not the source text, which is why we should fix this in wordepress, and why there is an issue for that already - https://github.com/weaveworks/wordepress/issues/7

From [the markdown spec](https://daringfireball.net/projects/markdown/syntax#p):
> Markdown supports “hard-wrapped” text paragraphs. This differs significantly from most other text-to-HTML formatters (including Movable Type’s “Convert Line Breaks” option) which translate every line break character in a paragraph into a `<br/>` tag. When you do want to insert a <br /> break tag using Markdown, you end a line with two or more spaces, then type return.

abuehrle:
So I should be using soft wrapping when I write text?  Ok, I am confused.

rade:
> So I should be using soft wrapping when I write text?

It shouldn't matter to the rendering, according to the markdown spec I quoted above.

awh:
* We use markdown so the docs are readable in both forms - hard line wraps in the source aid this
* Furthermore, diffs are much more effective when paragraphs aren't a single long line
* I already have a local fix for wordepress#7 I just haven't made it available yet - will do so shortly

awh:
Am reworking this into separate commits and reverting the soft wrapping.

awh:
> Am reworking this into separate commits and reverting the soft wrapping.

Done.

rade:
For this to fix #2183, as it claims, the

```
    IP Addresses, Routes and Networks
    Using Fast Datapath
    Encryption and Weave Net
    Using Weave with Systemd
```

sections need to be moved out of the top-level.

If you want to just dump them at the end for now, and raise a separate issue (noting that in #2183) then that is fine by me.

bboreham:
> sections need to be moved out of the top-level.

I have now pushed commits moving each of these.  PTAL.

Issue #2194, opened 2016-04-20
"Unable to claim: address 10.32.0.1/12 is already owned by 10.32.0.1"


bboreham:
````
vagrant@host1:~$ weave launch-router && weave launch-plugin
vagrant@host1:~$ docker run -dit --net=weave alpine sh
e1ee1b7b3c3c9b0919e38d887941a477c5e250da1de8454ade6d542c40818874
vagrant@host1:~$ docker run -dit --net=weave alpine sh
5806650ad2fcc72582ec7c40766eed15296d42e69a7d6852873243f44515453b
vagrant@host1:~$ weave stop-router
vagrant@host1:~$ weave launch-router
6eb736193f004e6ed5eb03c35450ca1e99ca1636ebb8e54b967e29a6bf094964
Unable to claim: address 10.32.0.2/12 already in use
Unable to claim: address 10.32.0.1/12 already in use
````

Oops

bboreham:
This message is informational; since the address is already owned nothing bad happens, but it is probably disturbing to users.

bboreham:
When the re-launch does its "claim" operations, it passes in the container ID, which does not match up to the earlier allocation done by the plugin, which does not know the container ID.

Tentatively, the claim operation could inspect the container, see if it is attached to a plugin network, and do things differently if so.

bboreham:
Or IPAM could watch for Docker 'network attach' events and replace the "_" workaround with the real container ID.  This would require a different path for the plugin's 'ReleaseAddress' operation.

rade:
How about we modify `weaveutil container-addrs` to include the interface name, and then in the weave the script `ipam_reclaim` can subsitute `_` for the container name when the interface is not `$CONTAINER_IFNAME` (ie. `ethwe`).

PR #2196, opened 2016-04-20
Improve handling of `claim _`

bboreham:
This improves the error message from "address 10.32.0.1/12 is already owned by 10.32.0.1" to "address 10.32.0.1/12 already in use".  Also the code as it stands would not have removed the 'owned' entry if it was deleted from the plugin.

It does not address the basic point of #2194, which is that the situation does not deserve a diagnostic. When the re-launch does its "claim" operations, it passes in the container ID, which does not match up to the earlier allocation via the plugin.

### DIFF
--- a/ipam/claim.go
+++ b/ipam/claim.go
@@ -72,14 +72,14 @@ func (c *claim) Try(alloc *Allocator) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


rade:
needs rebase

bboreham:
rebased

brb:
LGTM

Issue #1654, opened 2015-11-09
Additional DNS records are lost on container restart


bboreham:
As noted in #1191, DNS records that have been added with `weave dns-add` are lost when a container restarts, because weaveDNS deletes all records when it sees the container die and they are not recorded anywhere else.

awh:
Is this a dupe of #635?

bboreham:
#635 speaks about weaveDNS being restarted; this is about the container that owns the address being restarted.

inercia:
Any suggestion on how to approach this problem?

bboreham:
Could apply the same technique as #1191 - wait for a bit (or for container destroy) before removing all trace of the record.  The idea doesn't make me happy, however.

PR #2199, opened 2016-04-21
[WIP]: 635: Persist nameserver entries

### LINKED ISSUE #635: [dns] retain extra records on restart
With #566 and #627 we introduced the ability to add records to DNS for names that aren't the container hostname, and IPs that aren't weave IPs. These records are lost on restart of weaveDNS.

I see three ways of addressing this:

1. get weaveDNS to persist the records somehow. Downside: to date we have deliberately avoided any persistence in weave, and becoming a "system of record".

2. attach the extra information to container labels, and populate DNS with that on startup, much like we do for weave IPs and the container hostname. Downside: a) container labels can only be set on container creation, b) `weave expose` names something other than a container (though we could attach the info to the weave container), c) labels are new in Docker 1.6

3. as 2, but using https://github.com/docker/docker/pull/9013. This removes downside 2a. Downside: doesn't exist.

brb:
I'm opening a PR to get your instant feedback on changes I'd like to introduce to the nameserver component to make it retain extra nameserver entries on restarts.

In this PR, there are multiple TODO items (denoted as `TODO(mp)`) which require your attention.

I'll fix related docs once I'm done with the code.

The PR would fix #635 and #1654.


ipam/allocator.go
```diff
@@ -262,7 +262,7 @@ func (alloc *Allocator) Claim(ident string, cidr address.CIDR, noErrorOnUnknown
 func (alloc *Allocator) ContainerDied(ident string) {
 	alloc.actionChan <- func() {
 		if alloc.hasOwned(ident) {
-			alloc.debugln("Container", ident, "died; noting to remove later")
+			alloc.debugln("Container", ident, "died; nothing to remove later")
```
bboreham: this wasn't misspelled - "noting" means "making a note"

nameserver/nameserver.go
```diff
@@ -25,34 +26,109 @@ const (
...
+		n.infof("restore: cannot find any entry in DB")
+		return
+	}
+
+	// lHostname gets lost, because the gob encoder which is used internally by the DB
+	// wrapper does not serialize private fields, so we need to set this field manually
+	n.entries.addLowercase()
+
+	// TODO(mp) optimization: use the output of `docker ps` to restore
+	//			immediately entries of running containers.
```
bboreham: See also https://github.com/weaveworks/weave/blob/master/prog/weaver/main.go#L305

nameserver/nameserver.go
```diff
@@ -25,34 +26,109 @@ const (
...
+	case err != nil:
+		n.errorf("restore: cannot retrieve entries due to: %s", err)
+		return
+	case !ok:
+		n.infof("restore: cannot find any entry in DB")
+		return
+	}
+
+	// lHostname gets lost, because the gob encoder which is used internally by the DB
+	// wrapper does not serialize private fields, so we need to set this field manually
```
bboreham: Not serialising the lower-case version is intentional; it can be recovered from the real hostname so no point using up bytes on it.  I.e. the important thing is the decision not to serialise it, not the gob encoder.

nameserver/nameserver.go
```diff
@@ -25,34 +26,109 @@ const (
...
+				"restore: peer %s DB should not have contained entries of %s peer",
+				n.ourName, e.Origin))
+		}
+		if e.Tombstone == 0 {
+			n.debugf("restore: restoring %v...", e)
+			// We should not tombstone entries for "weave:extern", because there
+			// might be no subsequent AddEntry call for external IPs
+			// TODO(mp) personally, I don't like the idea, that nameserver is aware of"weave:extern"
+			//			IMO, it should not know about this workaround. One way to get
+			//			rid of it, is to introduce e.g. /restore/$CONTAINER_ID API which
```
bboreham: I don't follow this bit.  `weave:extern` addresses are ones that have been added by the user, outside of all containers, so why would a container ID be involved?

nameserver/nameserver.go
```diff
@@ -176,6 +282,23 @@ func (n *Nameserver) deleteTombstones() {
...
+	entries := n.entries.filterCopy(
+		func(e *Entry) bool {
+			if e.Origin == n.ourName {
+				return true
+			}
+			return false
+		})
+	// TODO(mp) optimize: store a single entry per key-val pair.
+	if err := n.db.Save(nameserverIdent, entries); err != nil {
+		n.errorf("saving to DB failed to due: %s", err)
```
bboreham: "to due" -> "due to"

nameserver/nameserver.go
brb: I need to distinguish between local and external (`containerid == weave:extern`) entries, because their recovery differs. So, my changes made Nameserver aware of this special container ID.

nameserver/nameserver.go
brb: I thought that the main reason for `lHostname` being a lower-case (aka private) is backward compatibility.

ipam/allocator.go
brb: Oops, completely misinterpreted the message.

nameserver/nameserver.go
brb: Thanks.

nameserver/nameserver.go
bboreham: ah, so by `/restore/$CONTAINER_ID` you meant `/restore/weave:extern` ?

nameserver/nameserver.go
bboreham: Even if that were true, it would not be driven by gob.  Gob ignores extra fields sent on the wire.

nameserver/nameserver.go
brb: Yep.

nameserver/nameserver.go
```diff
@@ -129,10 +231,13 @@ func (n *Nameserver) ContainerDied(ident string) {
 	entries := n.entries.tombstone(n.ourName, func(e *Entry) bool {
 		if e.ContainerID == ident {
 			n.infof("container %s died; tombstoning entry %s", ident, e.String())
+			// We want to restore all entries if container comes back
+			e.stopped = true
```
bboreham: Will this run even if the entry has previously been tombstoned by the `Delete()` method ?

brb: Noup. The `(Entries) tombstone(...)` function excludes previously tombstoned entries.

bboreham: I'm struggling to see that at https://github.com/weaveworks/weave/pull/2199/files#diff-079f8b38f0b2e637d5cd2138cb73ed4fR200

brb: The line you linked to: `e.tombstone()` returns false for already tombstoned entries => the `n.entries` list is not updated.

nameserver/nameserver.go
bboreham: Unfortunately, on a restart of the WeaveDNS process, there is no particular reason for the weave script to run.

You could pull this knowledge of `"weave:extern"` up to `main()`, which would match the general move of functionality from shell-script to Go.

nameserver/nameserver.go
bboreham: But this function will run, which was my question.  I think you're saying it doesn't matter as the entry isn't stored back into the list.

brb: Yep, because `range` returns a copy of `Entry` which is being modified by the function, and it used to update `n.entries` only if the condition at the linked line holds.

nameserver/nameserver.go
brb: Sounds like a reasonable approach. I would introduce `func (ns *Nameserver) RestoreAll(containerid string)` which would be called by `main()`.

nameserver/nameserver.go
```diff
@@ -176,6 +281,23 @@ func (n *Nameserver) deleteTombstones() {
...
+// NB: we assume that a caller has taken the nameserver' read or write lock.
+func (n *Nameserver) snapshot() {
+	entries := n.entries.filterCopy(
+		func(e *Entry) bool {
+			if e.Origin == n.ourName {
+				return true
+			}
+			return false
+		})
+	// TODO(mp) optimize: store a single entry per key-val pair.
```
brb: @bboreham any input on this?

nameserver/nameserver.go
```diff
@@ -176,6 +281,23 @@ func (n *Nameserver) deleteTombstones() {
 	n.entries.filter(func(e *Entry) bool {
 		return e.Tombstone == 0 || now-e.Tombstone <= int64(tombstoneTimeout/time.Second)
```
bboreham: Should you change this to leave entries with the `stopped` flag?

bboreham:
As an additional idea, you could act on `ContainerDestroyed` to remove the `stopped` flag since that container is not coming back.

nameserver/nameserver.go
brb: Noup, because otherwise the `stopped` entries of e.g. destroyed containers won't get GC'd.

bboreham: I addressed that elsewhere - clear the `stopped` flag on `ContainerDestroyed`.
Do we want the behaviour to depend on how long the container is stopped for?
I guess we do something similar in IPAM.

brb:
Ghm. I was thinking about the following situation:

1. Nameserver stops.
2. A container is destroyed.
3. Nameserver is re-launched.

Please correct me if I'm wrong, but in such situation Nameserver won't receive the `ContainerDestroyed` event for the container which was destroyed at the step (2), so it might leak.

bboreham:
As discussed verbally: 

> 1. Nameserver stops.
> 2. A container is destroyed.
> 3. Nameserver is re-launched.

. 4. Nameserver re-syncs its data with the current list of all containers and removes the destroyed one.

rade:
you've added a whole bunch of binaries.

common/docker/client.go
```diff
@@ -157,14 +157,35 @@ func (pending pendingStarts) finish(id string) {
...
 
 func (c *Client) AllContainerIDs() ([]string, error) {
-	all, err := c.ListContainers(docker.ListContainersOptions{All: true})
+	return c.containerIDs()
+}
+
+// ExistingContainerIDs returns a list of container IDs which either are running or
+// might be started in the future.
+func (c *Client) ExistingContainerIDs() ([]string, error) {
+	return c.containerIDs("running", "paused", "exited", "created", "restarting")
```
bboreham: Why do we need this restriction?  Is it not possible for a "dead" container to restart?

brb: https://github.com/docker/docker/blob/eabf97a9a3e6204a23dce635cf56040b0613ec8f/daemon/delete.go#L99

bboreham: OK, thanks.  Also https://github.com/docker/docker/blob/master/daemon/start.go#L98

Is it worth making this distinction?  Does it do anything apart from let us remove tombstoned entries earlier?

Having an inclusive list like this is fragile - it will break if the set of states is extended.

test/256_dns_persist_2_test.sh
```diff
@@ -0,0 +1,97 @@
...
+CMD="docker run --restart always --name c2 -h $NAME -e WEAVE_CIDR=$C2/24 -td $SMALL_IMAGE /bin/sh"
+run_on $HOST1 "eval \$(weave env) ; $CMD"
+start_container_with_dns $HOST1 $C1/24 --name=c1
+start_container_with_dns $HOST2 --name=c3
+weave_on $HOST1 dns-add c2 -h $NAME1
+weave_on $HOST1 dns-add $IP -h $NAME2
+
+assert_dns_a_record $HOST1 c1 $NAME $C2
+assert_dns_a_record $HOST1 c1 $NAME1 $C2
+assert_dns_a_record $HOST1 c1 $NAME2 $IP
```
bboreham: there seem to be repeated groups of 3 commands, which should be pulled out to a function

prog/weaver/main.go
```diff
@@ -273,10 +273,25 @@ func main() {
...
+			if err != nil {
+				Log.Fatalf("ExistingContainerIDs failed due to: %s", err)
+			}
+		}
+
+		// To restore external DNS entries we should pretend that "weave:extern"
+		// container exists.
+		existingContainerIDs = append(existingContainerIDs, "weave:extern")
+		ns.Start(existingContainerIDs)
+		ns.RestoreEntries("weave:extern")
```
bboreham: Do we need both this and the `append` earlier?

nameserver/nameserver.go
```diff
@@ -174,8 +364,58 @@ func (n *Nameserver) deleteTombstones() {
...
+}
+
+// loadEntries loads and unwraps entries from DB.
+func (n *Nameserver) loadEntries() bool {
+	var dbEntries []dbEntry
+
+	// Sanity check.
+	// Shouldn't happen, because the DNS server is started after restoreFromDB.
+	if len(n.entries) != 0 {
+		n.fatalf("restore: entries list is not empty")
```
bboreham: `common.Assert`?

nameserver/nameserver.go
```diff
@@ -83,12 +176,50 @@ func (n *Nameserver) broadcastEntries(es ...Entry) {
...
+}
+
+func (n *Nameserver) RestoreEntries(containerid string) {
+	var entries Entries
+
 	n.Lock()
-	n.infof("adding entry for %s: %s -> %s", containerid, hostname, addr.String())
-	entry := n.entries.add(hostname, containerid, origin, addr)
+	for i, e := range n.entries {
+		if e.Origin == n.ourName && e.ContainerID == containerid && e.stopped {
```
bboreham: extract function for this and very similar code in `AddEntry`

nameserver/nameserver.go
```diff
@@ -121,20 +252,78 @@ func (n *Nameserver) ReverseLookup(ip address.Address) (string, error) {
...
+		n.Unlock()
+		return
+	}
+	entries := n.containerDied(ident)
+	n.Unlock()
+	n.broadcastEntries(entries...)
+}
+
+func (n *Nameserver) containerDestroyed(ident string) Entries {
+	entries := n.entries.forceTombstone(n.ourName, func(e *Entry) bool {
```
bboreham: why do we need `forceTombstone`?  You don't seem to be changing `e` in the case where you return `false`.

bboreham: oops, ignore me.

nameserver/entry.go
```diff
@@ -23,6 +23,9 @@ type Entry struct {
 	lHostname   string // lowercased (not exported, so not encoded by gob)
 	Version     int
 	Tombstone   int64 // timestamp of when it was deleted
+	// Denotes that the container is stopped; the flag prevents a tombstoned entry
```
bboreham: s/stopped/stopped and may restart/

common/docker/client.go
brb: > Is it worth making this distinction?

Unfortunately we have to, because for `dead` containers there is no corresponding `ContainerDestroyed` event which handler resets the `stopped` flag.

> Having an inclusive list like this is fragile

Totally agree. An alternative approach would be to iterate and inspect each container during the restoration. However, it does not eliminate the dependency on a container state.

prog/weaver/main.go
brb: Yep, it's needed. `ns.restoreFromDB` (called by `ns.Start`) only flags entries of "weave:extern" as `stopped`, while `ns.RestoreEntries` does "un-tombstoning" of the entries.

common/docker/client.go
bboreham: > there is no corresponding `ContainerDestroyed` event

Citation?

nameserver/nameserver_test.go
```diff
@@ -266,3 +332,192 @@ func TestTombstoneDeletion(t *testing.T) {
 	nameserver.deleteTombstones()
 	require.Equal(t, Entries{}, nameserver.entries)
 }
+
+// TestRestore tests whether entries have been restored after the nameserver restart
+// and after manually calling RestoreEntries.
+func TestRestore(t *testing.T) {
```
bboreham: Suggest that, instead of adding lengthy hard-coded test instances, `TestNameservers` is extended to include container ID and containers being stopped, restarted and destroyed.

nameserver/nameserver_test.go
```diff
@@ -266,3 +332,192 @@ func TestTombstoneDeletion(t *testing.T) {
...
+			Addr:        addr5,
+			Hostname:    hostname5,
+			Version:     0,
+			Tombstone:   0,
+			stopped:     false,
+		},
+	})
+
+	time.Sleep(200 * time.Millisecond)
+	// TODO(mp) Check why ns2 entries are not broadcasted to ns1
```
bboreham: any ideas?

nameserver/nameserver.go
```diff
@@ -83,12 +176,50 @@ func (n *Nameserver) broadcastEntries(es ...Entry) {
...
 
-func (n *Nameserver) AddEntry(hostname, containerid string, origin mesh.PeerName, addr address.Address) {
+func (n *Nameserver) AddEntry(hostname, containerid string,
+	origin mesh.PeerName, addr address.Address) {
+
+	var entries Entries
+	var found bool
+
+	n.Lock()
+	// Restore the stopped entries first
```
bboreham: What do we expect to happen here if the container has restarted with a different IP address?

nameserver/nameserver_test.go
brb: I'm in favour of having multiple test cases which test separate features (i.e. unit testing).

However, I see an advantage of extending `TestNameservers` - it does some randomized testing which could test execution paths which were not exploited by `TestRestore`.

nameserver/nameserver.go
brb: Good find. I haven't addressed such situation, so currently, it will restore entries with a previous IP address being set, which is wrong.

bboreham:
There is some relevant discussion at #1179. It seems we have a fundamental problem in the different mechanisms for adding and removing (on `die`) DNS entries.

bboreham:
Closing due to inactivity

PR #2205, opened 2016-04-22
2057: Add allocator.PeerGone

### LINKED ISSUE #2057: unconnected peers hold on to nicknames indefinitely
```
host1:~$ weave launch && weave consense
host2:~$ weave launch 192.168.48.11 && weave consense
host2:~$ weave reset
host1:~$ weave rmpeer host2
No entries for peer found
host1:~$ weave rmpeer host3
Cannot find peer 'host3'
```
The difference in the error message indicates that host2 was still known to host1. The reason is that we only prune nicknames from ipam on receipt of gossip. When a peer is no longer connected to any others, it does not receive any gossip, so any nicknames left at the time will remain forever.

This is of little practical consequence, other than the possibly surprising difference in error message.

However, it's also easy to fix by pruning on a PeerGone event from the mesh.

brb:
The function is registered within the mesh GC and it is responsible
for removing gone nicknames.

Fixes #2057

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -90,11 +90,13 @@ type Config struct {
 // NewAllocator creates and initialises a new Allocator
 func NewAllocator(config Config) *Allocator {
 	var participant paxos.Participant
+
 	if config.IsObserver {
 		participant = paxos.NewObserver()
 	} else {
 		participant = paxos.NewNode(config.OurName, config.OurUID, 0)
 	}
+
 	return &Allocator{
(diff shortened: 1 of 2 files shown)


rade:
I am not convinced the test code is worth having. Not enough bang for buck.

brb:
Added a simplified version of `TestPeerGone`, although personally I don't like that `alloc.nicknames` gets directly modified in the test.

If you are happy with it, I will squash my commits and get rid of the previous version.

PR #2218, opened 2016-04-26
Upgrade vendored libs

bboreham:
Upgrade all the things!

(Except Go.  Wishing to avoid the issue Docker hit where it rejects old Docker clients)

Each library was upgraded to the latest release, the latest commit, or some other appropriate commit.
See commit message for details on which versions were selected.


awh:
Good job @bboreham - I like how you have recorded the versions in the commit message.

PR #2219, opened 2016-04-26
Document adding a route on non-Weave hosts

bboreham:
ping @abuehrle for doc review

### DIFF
--- a/site/using-weave/host-network-integration.md
+++ b/site/using-weave/host-network-integration.md
@@ -38,8 +38,25 @@ Exposed addresses can also be added to weaveDNS by supplying fully qualified dom
     host2$ weave expose -h exposed.weave.local
     10.2.1.132
 
+###<a name="routing"></a>Routing from Another Host
+
+After running `weave expose`, you can use Linux routing to provide
+access to the Weave network from hosts that are not running Weave Net:
+
+    ip route add <network-cidr> via <exposing-host>
+
+Where,
+ * `<network-cidr>` is an IP address range in use on Weave Net,
+for example,  `10.2.0.0/16` or `10.32.0.0/12` and,
+ * `<exposing-host>` is the address of the machine on which you ran `weave expose`.
+
+>**Note:** You must ensure that the [IP subnet used by Weave
+Net](/site/ipam.md#range) does not clash with anything on those other
+hosts.
+
 
 **See Also**
 
  * [Using Weave Net](/site/using-weave.md)
+ * [General information on IP Addresses, Routes and Networks](/site/how-it-works/ip-addresses.md)
  * [Managing Services - Exporting, Importing, Binding and Routing](/site/using-weave/service-management.md)
--- a/site/faq.md
+++ b/site/faq.md
@@ -31,9 +31,9 @@ Yes you can.
 
 For example, you have a Weave network that runs on hosts A, B, C. and you have an additional host, that we'll call P, where neither Weave nor Docker are running.  However, you need to connect from a process running on host P to a specific container running on host B on the Weave network.  Since the Weave network is completely separate from any network that P is connected to, you cannot connect the container using the container's IP address. 
 
-The simplest way to accomplish this would be to run Weave on the host and then run, `weave expose` to expose the network to any running containers.  See [Integrating a Network Host](/site/using-weave/host-network-integration.md) If this is not possible, you could also  expose a port on the container on host B and then connect to it. 
+A simple way to accomplish this would be to run Weave on the host and then run, `weave expose` to expose the network to any running containers. Or you set up a route from P to one of A, B or C. See [Integrating a Network Host](/site/using-weave/host-network-integration.md).
 
-You can read about exposing ports in [Exporting Services](/site/using-weave/service-management.md#exporting).
+Yet another option is to expose a port from the container on host B and then connect to it. You can read about exposing ports in [Exporting Services](/site/using-weave/service-management.md#exporting).
 
 
 **Q: Why am I seeing the same IP address assigned to two different containers on different hosts?**


site/using-weave/service-management.md
```diff
@@ -38,6 +39,17 @@ With the above in place, you are ready to connect to the 'nc' service from anywh
 
 Similar NAT rules to the above can used to expose services not just to the outside world but also other, internal, networks.
 
+###<a name="export-route"></a> Exporting Services to the wider network
+
+If you want hosts that are not running Weave Net to access services
+exposed as above, on each of those machines:
```
rade: No, that's not what this does. The "above" section is making *individual* services accessible to the host network *without altering the network configuration* - individual services end up with an IP/port on the host network.

By contrast, in this section we show how to make *the entire weave network* (or subnets thereof) accessible. Alas it *does* require networking changes (namely the addition of a route).

NB: a prerequisite here is that the subnet in use by weave is 'free' not just on the weave peers (as normal) but also all the hosts that require access to the weave network.

Also, position wise I reckon it's confusing for this to sit between the exporting/importing sections. I'd stick it at the very end.

abuehrle: Should also use title case for the heading, Exporting Services to the Wider Network  -- rule is to capitalize words with four or more letters.

site/using-weave/service-management.md
```diff
@@ -38,6 +39,17 @@ With the above in place, you are ready to connect to the 'nc' service from anywh
...
+
+    ip route add <network-cidr> via <exposing-host>
+
+Where `<network-cidr>` gives the IP address range used by weave,
+e.g. `10.2.0.0/16` or `10.32.0.0/12` and `<exposing-host>` is the
+address of the machine that has run `weave expose`, e.g. `host2` in
+the example above.
```
abuehrle: Where, 

* `<network-cidr> is the IP address range that is used by Weave, for example, 10.2.0.0/16` or `10.32.0.0/12` and, 
* `<exposing-host>` is the address of the machine on which `weave expose` was run. For example, `host2` in the example above <-- not sure which example this is referring to?

rade:
Q4 in https://github.com/weaveworks/weave/blob/master/site/faq.md needs updating.

site/using-weave/service-management.md
bboreham: @rade the previous section does make the entire weave network accessible, from the local host.  But I get your point: my addition is not about "services exposed as above".

bboreham:
Updated; PTAL

site/using-weave/service-management.md
```diff
@@ -10,6 +10,7 @@ This section contains the following topics:
  * [Binding Services](#binding)
  * [Routing Services](#routing)
  * [Dynamically Changing Service Locations](#change-location)
+ * [Exposing the Weave Network More Widely](#export-route)
```
rade: Looking at the current doc structure, I don't think this is the right place for this. It is not about services and instead is more like the "Integrating with the Host Network" section which lives at the same level as this entire doc. So I reckon it should either go into that (and that section should be renamed) or be a separate section at the same level.

@abuehrle what do you think?

site/using-weave/service-management.md
```diff
@@ -106,6 +107,25 @@ Furthermore, as explained in Binding Services, service locations can be dynamica
 For example, you can move the netcat service to `$HOST4:2211`  and it will retain its 10.2.1.3:3322 endpoint on the Weave network.
 
 
+###<a name="export-route"></a>Exposing the Weave Network More Widely
+
+You can use Linux routing to give access to the Weave network from
```
rade: I think it's "Weave Network" @abuehrle?

site/using-weave/service-management.md
```diff
@@ -106,6 +107,25 @@ Furthermore, as explained in Binding Services, service locations can be dynamica
...
+hosts that are not running Weave. First, ensure that the [IP
+subnet](site/how-it-works/ip-addresses.md) used by Weave does not
+clash with anything on those other hosts, then on one Weave host run:
```
rade: I believe we always say "Weave Net", not just "Weave". @abuehrle?

site/using-weave/service-management.md
```diff
@@ -106,6 +107,25 @@ Furthermore, as explained in Binding Services, service locations can be dynamica
...
+
+    weave expose
+
+Then on each non-Weave machine that you want to access the network,
+run:
+
+    ip route add <network-cidr> via <exposing-host>
+
+Where `<network-cidr>` gives the IP address range used by weave,
```
rade: I would make "IP address range used by Weave Net" a hyperlink to (the .md of) https://www.weave.works/documentation/net-1-5-0-ipam/#range

bboreham: On further thought, this doesn't have to map to the IPAM range, just to some subnet in use on Weave Net that you want to route.

bboreham:
I have moved the whole section to `ipam.md` as suggested, fixed the "Weave Net" usage, squashed and rebased.

site/using-weave/host-network-integration.md
```diff
@@ -38,6 +38,20 @@ Exposed addresses can also be added to weaveDNS by supplying fully qualified dom
     host2$ weave expose -h exposed.weave.local
     10.2.1.132
 
+###<a name="routing"></a>Routing from another host
+
```
abuehrle: Title case for the title.  This means only using capital letters for the principal words. Articles, conjunctions, and prepositions do not get capital letters unless they start the title.

site/using-weave/host-network-integration.md
```diff
@@ -38,6 +38,20 @@ Exposed addresses can also be added to weaveDNS by supplying fully qualified dom
...
+Having run `weave expose` as above, you can use Linux routing to give
+access to the Weave Network from hosts that are not running Weave Net:
+
```
abuehrle: After running `weave expose`, you can use Linux routing to provide access to the Weave network from hosts that are not running Weave Net:

site/using-weave/host-network-integration.md
```diff
@@ -38,6 +38,20 @@ Exposed addresses can also be added to weaveDNS by supplying fully qualified dom
...
+    ip route add <network-cidr> via <exposing-host>
+
+Where `<network-cidr>` is an IP address range in use on Weave Net,
+e.g. `10.2.0.0/16` or `10.32.0.0/12` and `<exposing-host>` is the
+address of the machine where you ran `weave expose`.
+
```
abuehrle: Where,
 * `<network-cidr>` is an IP address range in use on Weave Net,
for example,  `10.2.0.0/16` or `10.32.0.0/12` and, 
 * `<exposing-host>` is the address of the machine on which you ran `weave expose`.

bboreham:
Updated per comments.

site/using-weave/host-network-integration.md
```diff
@@ -38,6 +38,21 @@ Exposed addresses can also be added to weaveDNS by supplying fully qualified dom
...
+    ip route add <network-cidr> via <exposing-host>
+
+Where,
+ * `<network-cidr>` is an IP address range in use on Weave Net,
+for example,  `10.2.0.0/16` or `10.32.0.0/12` and,
+ * `<exposing-host>` is the address of the machine on which you ran `weave expose`.
+
+>**Note:** You must ensure that the [IP subnet](site/how-it-works/ip-addresses.md)
+used by Weave Net does not clash with anything on those other hosts.
```
rade: The obvious question a user will have at that point is... "How?". I suggest we turn some of the above test into a hyperlink to ipam.md#range.

bboreham:
Updated the FAQ and "subnet used by Weave Net", and squashed.

PR #2227, opened 2016-04-27
dedup service export docs

### LINKED ISSUE #2207: duplicate service export docs
https://www.weave.works/documentation/net-1-5-0-using-weave/net-1-5-0-service-export/

is a duplicate of

https://www.weave.works/documentation/net-1-5-0-using-weave/net-1-5-0-service-management/#exporting

except that the content is ever so slightly different. And both differ from https://github.com/weaveworks/weave/blob/1.4/site/features.md#service-export. So I have no idea which of the two versions is the "correct" one. Perhaps neither.

@abuehrle

rade:
Fixes #2207 

@abuehrle


abuehrle:
That looks better.  I had toyed with the idea of putting all of the 'managing services' stuff into one topic, hence the duplication.

PR #2228, opened 2016-04-27
tweak 'Using Weave Net' section order

rade:
- move "Managing Services in Weave Net" up so it comes after
  "Integrating a Host Network with Weave Net", to which it is deeply
  related
- move "Adding and Removing Hosts Dynamically" down so it is adjacent
  to the other sections which talk about peer connectivity

@abuehrle


abuehrle:
LGTM

PR #2230, opened 2016-04-27
Tweak titles under "Using Weave Net"

rade:
- strip "Weave" and "Weave Net"
- "Integrating a Host Network with Weave" -> "Integrating with the
  Host Network"
- "Configuring Weave to Explicitly Use an IP Range" -> "Allocating IPs
  in a Specific Range"

@abuehrle


abuehrle:
better. LGTM

PR #2238, opened 2016-04-28
Persist the IPAM ring on `weave reset` and `weave rmpeer`

bboreham:
These cases were missed earlier.
Note this change also makes `AdminTakeoverRanges()` cause pending ops to be retried, which seems appropriate - e.g. we may have taken over a range which covers a pending claim.

Calling `ringUpdated()` instead of `AddRanges()` is less efficient, because we will compute what space has been added instead of using what we already know in `newRanges`, but this is ok as `rmpeer` should not be called rapidly.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -650,8 +650,8 @@ func (alloc *Allocator) establishRing() {
 func (alloc *Allocator) createRing(peers []mesh.PeerName) {
 	alloc.debugln("Paxos consensus:", peers)
 	alloc.ring.ClaimForPeers(normalizeConsensus(peers))
-	alloc.gossip.GossipBroadcast(alloc.Gossip())
 	alloc.ringUpdated()
+	alloc.gossip.GossipBroadcast(alloc.Gossip())
 }
 
 func (alloc *Allocator) ringUpdated() {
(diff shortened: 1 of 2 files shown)


rade:
So, AFAICT...

1. persisting the ring in `Shutdown()` makes no practical difference, since `Shutdown` is only invoked in `weave reset`, which wipes out the store
2. persisting the ring in `AdminTakeOverRanges()` prevents ranges from getting "lost" when `rmpeer` is invoked on a peer which dies soon after
3. retrying pending ops in `AdminTakeOverRanges()` allows pending allocations/claims to proceed immediately, if they were waiting for space in the taken over ranges. Previously such allocations/claims would be retried only after the next timer tick or next gossip (and on master the latter would actually have to result in a ring change).

Questions (not about this PR as such but related):
- Why do we invoke `tryPendingOps()` on the ticker? AFAICT, with the above changes all changes that could cause a pending op to be retried successfully already invoke `tryPendingOps()`
- Why does `createRing()` invoke `ringUpdated()` *after* gossiping? Surely in general we should persist data before communicating it to other peers.

bboreham:
1. It is possible that something (e.g. a host reboot) interrupts the sequence between `Shutdown` and wiping out the store.
2. Yes
3. Yes, this is a fuller description of what I put in the PR description

> Why do we invoke `tryPendingOps()` on the ticker?

It's possible that a message requesting space got lost, e.g. the host we sent it to rebooted.

> Why does `createRing()` invoke `ringUpdated()` after gossiping? 

That looks like an error.

PR #2258, opened 2016-05-05
Publish doc previews for all branches, not just those named issues/*

awh:
Please update github.com/weaveworks/weave/wiki/WorkingOnWeave on merge.


awh:
If this has worked, you should be able to visit https://dev-weavewww.pantheon.io/docs/net/doc-preview-all-branches/introducing-weave/ after the CI build has finished.

PR #2264, opened 2016-05-06
Kill process instead of container, on container restart. 

### LINKED ISSUE #2123: restarting containers get killed if attachment fails
When executing `docker run` or `docker start` against the proxy, and a failure occurs when attaching the container to the weave network, we want the run/start to fail. We do this by `docker kill`ing the container - the `--or-die` variant of `weave attach`. So far so good...

But the very same code is used for re-attaching a restarting container. And `docker kill`ing a restarting container will prevent it from ever restarting. Which is (probably?) wrong.

The same "or die" behaviour is exhibited by `weave restart`.

I wonder whether killing the *process* might have the desired effect of returning an error to the CLI, while keeping docker's restart logic operational.

bboreham:
Fixes #2123 

I shall try to add a smoke-test.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -438,18 +438,20 @@ func (proxy *Proxy) waitForStart(r *http.Request) error {
 	return nil
 }
 
-// If some other operation is waiting for a container to start, join in the wait
-func (proxy *Proxy) waitForStartByIdent(ident string) error {
-	var ch chan error
+func (proxy *Proxy) waitChan(ident string) chan error {
 	proxy.Lock()
+	defer proxy.Unlock()
 	for _, wait := range proxy.waiters {
 		if ident == wait.ident && !wait.done {
-			ch = wait.ch
-			break
+			return wait.ch
 		}
 	}
-	proxy.Unlock()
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -498,7 +501,11 @@ func (proxy *Proxy) attach(containerID string, orDie bool) error {
 		args = append(args, "--no-multicast-route")
 	}
 	if orDie {
-		args = append(args, "--or-die")
+		if killContainer {
+			args = append(args, "--or-die")
+		} else {
+			args = append(args, "--or-kill")
+		}
```
rade: This is either the wrong way round or the naming is really confusing.

bboreham: `--or-die` is the existing behaviour; it kills the container.
The new behaviour is to kill the process.
Maybe `--or-kill-process` ?

bboreham: amended boolean arg to `killProcess`

rade:
> I shall try to add a smoke-test.

yes, please.

bboreham:
Smoke-test added

PR #2265, opened 2016-05-06
make claim block if the IPAM ring is not established

### LINKED ISSUE #2232: Weave giving "Unable to claim" error
This started to occur after upgrading to weave 1.5.0, launching two hosts like this:

```
CHECKPOINT_DISABLE=true weave launch --ipalloc-range 192.99.0.0/16 10.10.200.162 10.10.200.160
```

And then trying to launch a container in that range:

```
docker -H unix:///var/run/weave/weave.sock run -dt --hostname=mycontainer.weave.local -e WEAVE_CIDR=192.99.10.112/16 ubuntu bash
```

Causes an error:
```
docker: Error response from daemon: Unable to claim: 192.99.10.112/16 is in the range 192.99.0.0/16, but the allocator is not initialized yet.
```

brb:
This PR aligns the behaviour of Claim with Allocate which blocks until the ring has been established.

Fixes #2232

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -276,7 +276,7 @@ func (alloc *Allocator) Lookup(ident string, r address.Range) ([]address.CIDR, e
 }
 
 // Claim an address that we think we should own (Sync)
-func (alloc *Allocator) Claim(ident string, cidr address.CIDR, isContainer, noErrorOnUnknown bool) error {
+func (alloc *Allocator) Claim(ident string, cidr address.CIDR, isContainer, noErrorOnUnknown bool, hasBeenCancelled func() bool) error {
 	resultChan := make(chan error)
 	op := &claim{
 		resultChan:       resultChan,
@@ -284,6 +284,7 @@ func (alloc *Allocator) Claim(ident string, cidr address.CIDR, isContainer, noEr
 		cidr:             cidr,
 		isContainer:      isContainer,
 		noErrorOnUnknown: noErrorOnUnknown,
+		hasBeenCancelled: hasBeenCancelled,
 	}
 	alloc.doOperation(op, &alloc.pendingClaims)
 	return <-resultChan
--- a/ipam/allocator_test.go
+++ b/ipam/allocator_test.go
@@ -27,7 +27,7 @@ func (alloc *Allocator) SimplyAllocate(ident string, cidr address.CIDR) (address
 }
 
 func (alloc *Allocator) SimplyClaim(ident string, cidr address.CIDR) error {
-	return alloc.Claim(ident, cidr, true, true)
+	return alloc.Claim(ident, cidr, true, true, returnFalse)
 }
 
 func TestAllocFree(t *testing.T) {
--- a/ipam/claim.go
+++ b/ipam/claim.go
@@ -15,6 +15,7 @@ type claim struct {
 	cidr             address.CIDR
 	isContainer      bool
 	noErrorOnUnknown bool
+	hasBeenCancelled func() bool
 }
 
 // Send an error (or nil for success) back to caller listening on resultChan
@@ -33,6 +34,11 @@ func (c *claim) sendResult(result error) {
 
 // Try returns true for success (or failure), false if we need to try again later
 func (c *claim) Try(alloc *Allocator) bool {
+	if c.hasBeenCancelled() {
+		c.Cancel()
+		return true
+	}
+
 	if !alloc.ring.Contains(c.cidr.Addr) {
 		// Address not within our universe; assume user knows what they are doing
 		alloc.infof("Address %s claimed by %s - not in our range", c.cidr, c.ident)
@@ -51,11 +57,10 @@ func (c *claim) Try(alloc *Allocator) bool {
 		// success
 	case mesh.UnknownPeerName:
 		// If our ring doesn't know, it must be empty.
+		alloc.infof("Claim %s for %s: is in the range %s, but the allocator is not initialized yet; will try later.",
+			c.cidr, c.ident, alloc.universe.AsCIDRString())
 		if c.noErrorOnUnknown {
-			alloc.infof("Claim %s for %s: address allocator still initializing; will try later.", c.cidr, c.ident)
 			c.sendResult(nil)
-		} else {
-			c.sendResult(fmt.Errorf("%s is in the range %s, but the allocator is not initialized yet", c.cidr, alloc.universe.AsCIDRString()))
 		}
 		return false
 	default:
--- a/ipam/http.go
+++ b/ipam/http.go
@@ -38,30 +38,50 @@ func writeAddresses(w http.ResponseWriter, cidrs []address.CIDR) {
 	}
 }
 
-func (alloc *Allocator) handleHTTPAllocate(dockerCli *docker.Client, w http.ResponseWriter, ident string, checkAlive bool, subnet address.CIDR) {
-	closedChan := w.(http.CloseNotifier).CloseNotify()
-	addr, err := alloc.Allocate(ident, subnet, checkAlive,
-		func() bool {
-			select {
-			case <-closedChan:
-				return true
-			default:
-				res := checkAlive && dockerCli != nil && dockerCli.IsContainerNotRunning(ident)
-				checkAlive = false // we check only once; if the container dies later we learn about that through events
-				return res
-			}
-		})
-	if err != nil {
-		if _, ok := err.(*errorCancelled); ok { // cancellation is not really an error
-			common.Log.Infoln("[allocator]:", err.Error())
-			fmt.Fprint(w, "cancelled")
-			return
+func hasBeenCanceled(dockerCli *docker.Client, closedChan <-chan bool, ident string, checkAlive bool) func() bool {
+	return func() bool {
+		select {
+		case <-closedChan:
+			return true
+		default:
+			res := checkAlive && dockerCli != nil && dockerCli.IsContainerNotRunning(ident)
+			checkAlive = false // we check only once; if the container dies later we learn about that through events
+			return res
 		}
-		badRequest(w, err)
+	}
+}
+
+func cancellationErr(w http.ResponseWriter, err error) bool {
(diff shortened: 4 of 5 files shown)


ipam/http.go
```diff
@@ -73,13 +93,10 @@ func (alloc *Allocator) HandleHTTP(router *mux.Router, defaultSubnet address.CID
...
-			noErrorOnUnknown := r.FormValue("noErrorOnUnknown") == "true"
+			ident := vars["id"]
 			checkAlive := r.FormValue("check-alive") == "true"
-			if err := alloc.Claim(vars["id"], cidr, checkAlive, noErrorOnUnknown); err != nil {
-				badRequest(w, fmt.Errorf("Unable to claim: %s", err))
-				return
-			}
-			w.WriteHeader(204)
+			noErrorOnUnknown := r.FormValue("noErrorOnUnknown") == "true"
+			alloc.handleHTTPClaim(w, ident, cidr, checkAlive, noErrorOnUnknown)
```
rade: why did you introduce the `ident` var and move the `noErrorOnUnknown`? It doesn't look like an obvious improvement and makes the diff larger than necessary.

ipam/http.go
```diff
@@ -64,6 +64,26 @@ func (alloc *Allocator) handleHTTPAllocate(dockerCli *docker.Client, w http.Resp
...
+		}
+	}
+	err := alloc.Claim(ident, cidr, isContainer, noErrorOnUnknown, hasBeenCanceled)
+	if err != nil {
+		badRequest(w, fmt.Errorf("Unable to claim: %s", err))
+		return
+	}
+	w.WriteHeader(204)
+}
+
```
rade: Is there a reason why the above is so different from `handleHTTPAllocate`? i.e.

- allocate does a checkAlive inside hasBeenCancelled but claim does not
- allocate suppresses errorCancelled errors but claim does not
- allocate uses an anonymous function for the cancellation check but claim does not

ipam/allocator_test.go
```diff
@@ -532,7 +532,7 @@ func TestAllocatorFuzz(t *testing.T) {
 		addressIndex := rand.Int31n(int32(subnet.Size()))
 		alloc := allocs[allocIndex]
 		addr := address.Add(subnet.Addr, address.Offset(addressIndex))
-		err := alloc.Claim(name, address.MakeCIDR(subnet, addr), true, true)
+		err := alloc.Claim(name, address.MakeCIDR(subnet, addr), true, true, returnFalse)
```
rade: I've just pushed a refactor in daa369b which makes all the changes in this file superfluous, and instead only a single line needs changing.

ipam/allocator.go
```diff
@@ -276,14 +276,17 @@ func (alloc *Allocator) Lookup(ident string, r address.Range) ([]address.CIDR, e
 }
 
 // Claim an address that we think we should own (Sync)
-func (alloc *Allocator) Claim(ident string, cidr address.CIDR, isContainer, noErrorOnUnknown bool) error {
+func (alloc *Allocator) Claim(ident string, cidr address.CIDR, isContainer,
+	noErrorOnUnknown bool, hasBeenCancelled func() bool) error {
```
rade: don't split args over multiple lines

ipam/claim.go
```diff
@@ -55,7 +61,9 @@ func (c *claim) Try(alloc *Allocator) bool {
 			alloc.infof("Claim %s for %s: address allocator still initializing; will try later.", c.cidr, c.ident)
 			c.sendResult(nil)
 		} else {
-			c.sendResult(fmt.Errorf("%s is in the range %s, but the allocator is not initialized yet", c.cidr, alloc.universe.AsCIDRString()))
+			alloc.infof(
+				"%s is in the range %s, but the allocator is not initialized yet",
+				c.cidr, alloc.universe.AsCIDRString())
```
rade: cosmetic: don't split this over multiple lines

also, the message should say "Claim %s for %s: is in the range %s, but ..."

rade: In fact, just move the log message out of the true branch, and have an empty else branch.

test/320_claim_3_test.sh
```diff
@@ -52,4 +52,22 @@ weave_on $HOST3 prime
...
+stop_weave_on $HOST3
+delete_persistence $HOST1 $HOST2 $HOST3
+
+weave_on $HOST1 launch $HOST2
+# Start another container on host1. The starting should block, because host1 is
+# not able to establish the ring due to host2 or host3 being offline.
+CMD="proxy_start_container $HOST1 --name c4 -e WEAVE_CIDR=10.32.0.4/12"
+assert_raises "timeout 5 cat <( $CMD )"  124
+# However, allocation for an external subnet should not block.
+assert_raises "proxy_start_container $HOST1 -e WEAVE_CIDR=10.48.0.1/12"
```
rade: The above is twice making an assumptions about what the default ipalloc range is. It shouldn't. Use `--ipalloc-range $UNIVERSE` when launching weave, as in some other tests.

test/320_claim_3_test.sh
```diff
@@ -52,4 +52,22 @@ weave_on $HOST3 prime
...
 weave_on $HOST1 launch-router $HOST3
 
+stop_weave_on $HOST1
+stop_weave_on $HOST2
+stop_weave_on $HOST3
+delete_persistence $HOST1 $HOST2 $HOST3
+
+weave_on $HOST1 launch $HOST2
+# Start another container on host1. The starting should block, because host1 is
+# not able to establish the ring due to host2 or host3 being offline.
```
rade: there is no host3

test/320_claim_3_test.sh
```diff
@@ -52,4 +52,22 @@ weave_on $HOST3 prime
...
+CMD="proxy_start_container $HOST1 --name c4 -e WEAVE_CIDR=10.32.0.4/12"
+assert_raises "timeout 5 cat <( $CMD )"  124
+# However, allocation for an external subnet should not block.
+assert_raises "proxy_start_container $HOST1 -e WEAVE_CIDR=10.48.0.1/12"
+
+# Launch host2, so that host1 could establish the ring.
```
rade: s/could/can

test/320_claim_3_test.sh
```diff
@@ -52,4 +52,22 @@ weave_on $HOST3 prime
...
+weave_on $HOST2 launch $HOST1
+sleep 2
```
rade: I reckon if you ran `weave_on $HOST1 prime` here you could get rid of the `sleep`.

ipam/allocator.go
brb: Are we OK with longer than 80 or 120 char lines?

ipam/http.go
brb: To make the code to look more consistent:
```
A = ...
B = ...
C = ...
foo(A, B, C).
```

ipam/http.go
brb: Addressed.

brb:
PTAL.

test/320_claim_3_test.sh
```diff
@@ -52,4 +54,22 @@ weave_on $HOST3 prime
...
+# not able to establish the ring due to host2 being offline.
+CMD="proxy_start_container $HOST1 --name c4 -e WEAVE_CIDR=10.32.0.4/12"
+assert_raises "timeout 5 cat <( $CMD )" 124
+# However, allocation for an external subnet should not block.
+assert_raises "proxy_start_container $HOST1 -e WEAVE_CIDR=10.48.0.1/12"
+
+# Launch host2, so that host1 can establish the ring.
+weave_on $HOST2 launch --ipalloc-range $UNIVERSE $HOST1
+weave_on $HOST1 prime
+assert_raises "[ $(container_ip $HOST1 c4) == 10.32.0.4 ]"
```
bboreham: Does this rely on the timed-out command carrying on running?
I found that surprising.

ipam/http.go
```diff
@@ -38,32 +38,54 @@ func writeAddresses(w http.ResponseWriter, cidrs []address.CIDR) {
...
 		}
-		badRequest(w, err)
 		return
 	}
-
 	fmt.Fprintf(w, "%s/%d", addr, subnet.PrefixLen)
 }
 
+func (alloc *Allocator) handleHTTPClaim(dockerCli *docker.Client, w http.ResponseWriter, ident string, cidr address.CIDR,
+	checkAlive, noErrorOnUnknown bool) {
```
bboreham: please make this bikeshed longer and less tall.

test/320_claim_3_test.sh
brb: Yes, because the `c4` container started via proxy is not killed when the timeout occurs.

bboreham:
LGTM apart from a couple of nits

test/320_claim_3_test.sh
bboreham: OK, but does the `docker run` inside `proxy_start_container` get killed?

ipam/http.go
brb: (flame) Personally, I wouldn't call making lines to be less than 80/120 chars long as bikeshedding, because longer lines breaks visualization of code on some tools (incl. GH) and makes code harder to read.

Anyway, I'm changing it to be compliant with the rest of `ipam/*`.

test/320_claim_3_test.sh
brb: The `docker run` request process gets killed, but the proxy is not notified about it, so it keeps the container running.

bboreham: I filed #2270.

test/320_claim_3_test.sh
```diff
@@ -2,6 +2,8 @@
 
 . ./config.sh
 
+UNIVERSE=10.32.0.0/12
```
rade: Unless there is a very good reason to do otherwise, the universe for testing should be `10.2.0.0/16` or `10.2.x.0/24`. That is what we use everywhere else.

brb: Addressed in https://github.com/weaveworks/weave/pull/2278

Issue #2275, opened 2016-05-10
Unable to claim address immediately after ring is formed


bboreham:
````
host1$ weave reset
host2$ weave reset
host1$ weave launch $HOST2
host2$ weave launch $HOST1
host2$ docker $(weave config) run --rm -e WEAVE_CIDR=10.32.1.1/12 -ti alpine true
docker: Error response from daemon: Unable to claim: address 10.32.1.1/12 is owned by other peer 0e:2a:9b:8e:23:fd (host1)
````

It appears that the unicast request for space arrives before the broadcast of the initial ring.

bboreham:
A unicast message is encoded before the call to send it, whereas a broadcast is encoded by a call back into the `Allocator`, which has to wait for the actor, so actually we will always send unicasts before broadcasts from the same actor.

By the same reasoning, the `Sleep` in `Shutdown` is pointless.

PR #2278, opened 2016-05-10
Fix 320_claim_3_test.sh

brb:
UPDATED.

The problem was that the claim at https://github.com/weaveworks/weave/pull/2278/files#diff-1b73b6468431320099c9cbd2340c9a01L65 was sometimes failing because of `10.32.0.0/13` being assigned to `host2`. As a consequence, the assertion at https://github.com/weaveworks/weave/pull/2278/files#diff-1b73b6468431320099c9cbd2340c9a01L73 was failing.

The fix ensures that we always try to claim at `host1` for IP addr from the range assigned to the host. It's done by reusing the IP addr of `c4`.

### DIFF
--- a/test/320_claim_3_test.sh
+++ b/test/320_claim_3_test.sh
@@ -2,14 +2,29 @@
 
 . ./config.sh
 
-UNIVERSE=10.32.0.0/12
-C1=10.40.0.1
+UNIVERSE=10.2.0.0/16
+C1=10.2.128.1
+C4=10.2.128.4
 
 delete_persistence() {
     for host in "$@" ; do
         docker_on $host rm -v weavedb >/dev/null
         docker_on $host rm weave >/dev/null
     done
+}
+
+wait_for_container_ip() {
+    host=$1
+    container=$2
+    for i in $(seq 1 120); do
+        echo "Waiting for $container ip at $host"
+        if container_ip $host $container > /dev/null 2>&1 ; then
+            return
+        fi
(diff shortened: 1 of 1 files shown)


bboreham:
Can you put a description of what changed here, please?

brb:
Reduced the test size.

test/320_claim_3_test.sh
```diff
@@ -54,22 +68,25 @@ weave_on $HOST3 prime
 stop_weave_on $HOST2
 weave_on $HOST1 launch-router --ipalloc-range $UNIVERSE $HOST3
 
+docker_on $HOST1 stop c1
 stop_weave_on $HOST1
-stop_weave_on $HOST2
 stop_weave_on $HOST3
-delete_persistence $HOST1 $HOST2 $HOST3
+## Do not remove the state of host2 to keep the previously established ring.
```
bboreham: Not stopping host2 seems to invalidate the statement about "host2 being offline" below.

bboreham: Oh, wait, it was stopped already.

bboreham: "to keep the previously established ring so the start of `c4` below invokes a claim request"

I.e. you are forcing the conditions for bug #2275.  If I understood correctly.

brb: Simplified a bit the code and changed the comment(s).

PR #2279, opened 2016-05-10
Ignore IPAM requests for space until we have a ring

### LINKED ISSUE #2275: Unable to claim address immediately after ring is formed

bboreham:
Fixes #2275 

In the case that we do receive a ring very soon, the consequence of this change is that the requestor doesn't hear back so retries within 5 seconds.

I also moved the debug message so we still see it on requests that are ignored.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -512,9 +512,11 @@ func (alloc *Allocator) OnGossipUnicast(sender mesh.PeerName, msg []byte) error 
 	alloc.actionChan <- func() {
 		switch msg[0] {
 		case msgSpaceRequest:
-			// some other peer asked us for space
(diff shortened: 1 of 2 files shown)


bboreham:
Wrong branch.

PR #2280, opened 2016-05-10
Move sleep on reset from IPAM Allocator to weave script

bboreham:
Since the implementation of `GossipBroadcast` requires a call back into the IPAM actor to encode the message, with the previous implementation it wouldn't actually get sent until after the sleep finishes.
Since we only use this function in one place, move the sleep out to the caller.


brb:
LGTM.

PR #2281, opened 2016-05-10
Ignore IPAM requests for space until we have a ring

### LINKED ISSUE #2275: Unable to claim address immediately after ring is formed

bboreham:
Fixes #2275

In the case that we do receive a ring very soon, the consequence of this change is that the requestor doesn't hear back so retries within 5 seconds.

I also moved the debug message so we still see it on requests that are ignored.

### DIFF
--- a/test/320_claim_3_test.sh
+++ b/test/320_claim_3_test.sh
@@ -3,6 +3,7 @@
 . ./config.sh
 
 UNIVERSE=10.32.0.0/12
+C1=10.40.0.1
 
 delete_persistence() {
     for host in "$@" ; do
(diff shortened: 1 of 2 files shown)


test/320_claim_3_test.sh
```diff
@@ -13,12 +14,11 @@ delete_persistence() {
 
 start_suite "Claiming addresses"
 
-weave_on $HOST1 launch-router --ipalloc-range $UNIVERSE
+weave_on $HOST1 launch-router --ipalloc-range $UNIVERSE $HOST2
```
brb: Why is it needed?

bboreham: without this change, host1 can form a ring without talking to host2, but I want to test what happens when it has to wait for consensus.

test/320_claim_3_test.sh
```diff
@@ -13,12 +14,11 @@ delete_persistence() {
 
 start_suite "Claiming addresses"
 
-weave_on $HOST1 launch-router
-weave_on $HOST2 launch-router $HOST1
+weave_on $HOST1 launch-router --ipalloc-range $UNIVERSE $HOST2
```
brb: Why passing `$HOST2` is needed?

brb:
Have you considered calling `alloc.tryPendingOps()` after establishing the ring? It could make re-try to happen sooner.

brb:
LGTM.

test/320_claim_3_test.sh
bboreham: Without this change, host1 can form a ring without talking to host2, but I want to test what happens when it has to wait for consensus.  I could equally have told it a peer count of 2.

PR #2287, opened 2016-05-11
Donate CIDR-aligned ranges instead of just slicing at the mid-point.

bboreham:
If, say, we wanted to use the ranges for routing rules, having them aligned so you can describe each one with a single CIDR is useful.
Also fix up space tests to match.

### DIFF
--- a/ipam/space/space.go
+++ b/ipam/space/space.go
@@ -122,8 +122,11 @@ func (s *Space) biggestFreeRange(r address.Range) (biggest address.Range) {
 	biggestSize := address.Count(0)
 	s.walkFree(r, func(chunk address.Range) bool {
 		if size := chunk.Size(); size >= biggestSize {
-			biggest = chunk
-			biggestSize = size
+			chunk = chunk.BiggestPow2AlignedRange()
+			if size = chunk.Size(); size >= biggestSize {
+				biggest = chunk
+				biggestSize = size
+			}
 		}
 		return false
 	})
@@ -137,9 +140,12 @@ func (s *Space) Donate(r address.Range) (address.Range, bool) {
 		return address.Range{}, false
 	}
 
-	// Donate half of that biggest free range. Note size/2 rounds down, so
-	// the resulting donation size rounds up, and in particular can't be empty.
-	biggest.Start = address.Add(biggest.Start, address.Offset(biggest.Size()/2))
+	// Donate no more than half what we have available
+	if biggest.Size() > s.NumFreeAddressesInRange(r)/2 {
+		// Note size/2 rounds down, so the resulting donation size
+		// rounds up, and in particular can't be empty.
+		biggest.Start = address.Add(biggest.Start, address.Offset(biggest.Size()/2))
+	}
 
 	s.ours = subtract(s.ours, biggest.Start, biggest.End)
 	s.free = subtract(s.free, biggest.Start, biggest.End)
--- a/ipam/space/space.go
+++ b/ipam/space/space.go
@@ -122,7 +122,7 @@ func (s *Space) biggestFreeRange(r address.Range) (biggest address.Range) {
 	biggestSize := address.Count(0)
 	s.walkFree(r, func(chunk address.Range) bool {
 		if size := chunk.Size(); size >= biggestSize {
-			chunk = chunk.BiggestPow2AlignedRange()
+			chunk = chunk.BiggestCIDRRange()
 			if size = chunk.Size(); size >= biggestSize {
 				biggest = chunk
 				biggestSize = size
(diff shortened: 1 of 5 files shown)


net/address/address_test.go
```diff
@@ -0,0 +1,26 @@
...
+	require.Equal(t, NewRange(2, 2), NewRange(1, 4).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(4, 4), NewRange(1, 8).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(4, 4), NewRange(2, 8).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(4, 4), NewRange(4, 8).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(8, 4), NewRange(8, 7).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(8, 8), NewRange(8, 8).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(0, 0x4000), NewRange(0, 0x7fff).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(0x2000, 2), NewRange(0x2000, 3).BiggestPow2AlignedRange())
+	require.Equal(t, NewRange(0xffff, 1), NewRange(0xffff, 1).BiggestPow2AlignedRange())
+}
```
rade: I would be tempted to write a helper that allows me to write

    check(t, 0x2000, 2, 0x2000, 3)

Also, I do rather struggle with working out why *exactly* these tests are done.

bboreham: > _exactly_ these tests

Something like "the first time each bit in the result changes or might change if there was a bug, plus some boundary conditions"

brb: Have you considered using https://golang.org/pkg/testing/quick/ ?

ipam/space/space.go
```diff
@@ -122,8 +122,11 @@ func (s *Space) biggestFreeRange(r address.Range) (biggest address.Range) {
 	biggestSize := address.Count(0)
 	s.walkFree(r, func(chunk address.Range) bool {
 		if size := chunk.Size(); size >= biggestSize {
```
brb: The biggest chunk does not guarantee that it contains the biggest CIDR-aligned range.

ipam/space/space.go
```diff
@@ -122,8 +122,11 @@ func (s *Space) biggestFreeRange(r address.Range) (biggest address.Range) {
...
-			biggest = chunk
-			biggestSize = size
+			chunk = chunk.BiggestPow2AlignedRange()
```
brb: The biggest chunk does not guarantee that it contains the biggest CIDR-aligned range.

If we don't care too much about a few CPU cycles being wasted, then I'd suggest to merge the if-statements.

bboreham: The first if-statement discards chunks that cannot possibly contain the biggest CIDR-aligned range.
The second if-statement accepts the biggest CIDR-aligned range.

Yes, it is coded this way to run faster than if we looked for the biggest CIDR-aligned chunk in each free chunk.

net/address/address.go
```diff
@@ -35,6 +35,35 @@ func (r Range) AsCIDRString() string {
 	return CIDR{Addr: r.Start, PrefixLen: prefixLen}.String()
 }
 
+// return the highest bit set in a
```
brb: "Return the highest bit set in v"?

net/address/address.go
```diff
@@ -35,6 +35,35 @@ func (r Range) AsCIDRString() string {
...
+	v |= v >> 1
+	v |= v >> 2
+	v |= v >> 4
+	v |= v >> 8
+	v |= v >> 16
+	// Cast to 64-bit to avoid overflow when original v >= 0x80000000
+	return Count((uint64(v) + 1) / 2)
+}
+
+func (r Range) BiggestPow2AlignedRange() Range {
```
brb: `BiggestCIDRRange`?

net/address/address_test.go
```diff
@@ -0,0 +1,41 @@
...
+	for ; x > 1; x /= 2 {
+		if x&1 != 0 {
+			return false
+		}
+	}
+	return true
+}
+
+func TestBiggestPow2AlignedRange(t *testing.T) {
+	require.Equal(t, NewRange(0, 1), NewRange(0, 1).BiggestPow2AlignedRange())
```
brb: The `cover` tool says that the following five assertions do not execute all four possible execution paths. I'd suggest to have a test case per each.

net/address/address.go
bboreham: I steered away from "CIDR" because we have a type called `CIDR` that we aren't using here, but `BiggestCIDRRange` might be ok.

bboreham: Should I paste in the Greek-letter explanation of what is going on here?

bboreham:
I addressed most comments; PTAL

net/address/address.go
brb: Ghm. I'm fine with understanding what it does from the code. What @rade @awh thinks?

PR #2291, opened 2016-05-12
Refactor wait_for_{proxy,attached}

brb:
This PR:

* Extracts and moves `wait_for_proxy` to _config.sh_.
* Introduces `wait_for_attached`.

### DIFF
--- a/test/640_proxy_restart_reattaches_test.sh
+++ b/test/640_proxy_restart_reattaches_test.sh
@@ -4,11 +4,6 @@
 
 NAME=seetwo.weave.local
 
-check_attached() {
-    assert_raises "exec_on $HOST1 c2 $CHECK_ETHWE_UP"
-    assert_dns_record $HOST1 c1 $NAME $C2
-}
-
 N=50
 # Create and remove a lot of containers in a small subnet; the failure
 # mode is that this takes a long time as it has to wait for the old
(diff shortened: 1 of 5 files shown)


test/640_proxy_restart_reattaches_test.sh
```diff
@@ -27,39 +22,41 @@ proxy_start_container_with_dns $HOST1 -di --name=c1 --restart=always
...
-check_attached
+wait_for_attached $HOST1 c2
+assert_dns_record $HOST1 c1 $NAME $C2
 
 # Restart weave router
 docker_on $HOST1 restart weave
+wait_for_attached $HOST1 c2
 sleep 1
-check_attached
+assert_dns_record $HOST1 c1 $NAME $C2
```
bboreham: we shouldn't need the sleeps any more, should we?

bboreham:
Nice reduction in code size!

test/640_proxy_restart_reattaches_test.sh
brb: Unfortunately, without the sleeps it fails on my machine from time to time.

awh: You're seeing this because `weave attach` brings the container's ethwe interface up before it adds the DNS records.

awh: We could eliminate the sleeps at a later date by implementing `wait_for_dns_record`.

PR #2295, opened 2016-05-13
Change the type of entry.Free to address.Count

brb:
This change is backward compatible, because the underlying types of `address.Offset` (the previous type) and `address.Count` are the same and the Gob encoder does not require types to exactly correspond.

In addition to this PR, I'd suggest to rename `func (r *Ring) distance(start, end address.Address)`, because the existing function name implies the `address.Offset` return type.

### DIFF
--- a/ipam/ring/entry.go
+++ b/ipam/ring/entry.go
@@ -14,8 +14,7 @@ type entry struct {
 	Token   address.Address // The start of this range
 	Peer    mesh.PeerName   // Who owns this range
 	Version uint32          // Version of this range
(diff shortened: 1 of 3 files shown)


bboreham:
Why would we do this on the 1.5 branch?

brb:
It does not introduce any changes to a behaviour?

bboreham:
Generally we only do bug-fixes on the branch

brb:
Re-opened: https://github.com/weaveworks/weave/pull/2299

PR #2296, opened 2016-05-13
Move kill-on-failure responsibility to caller of weave script

### LINKED ISSUE #2284: failure during start with `--rm` causes "You cannot remove a running container" error
From #2283 ...

```
$ weave launch
$ docker $(weave config) run --rm -e WEAVE_CIDR=invalid -ti alpine true
...(complaint about invalid CIDR)...
Error response from daemon: You cannot remove a running container d392639f97100e25facb14b59bf59439557ecacd2d44726a6259ae639c74dd3e. Stop the container before attempting removal or use -f
```

I've poked around this a little, and it looks like the removal comes from the Docker client:
```
INFO: 2016/05/10 22:01:50.258376 DELETE /v1.23/containers/ee3e6e074f60eb31d99ac0bfecacc85653146fbecddf880c9266ec46109f0b34?v=1
DEBU: 2016/05/10 22:01:50.258906 Response from target: 409 Conflict map[Content-Type:[text/plain; charset=utf-8] Server:[Docker/1.11.1 (linux)] X-Content-Type-Options:[nosniff] Date:[Tue, 10 May 2016 22:01:50 GMT] Content-Length:[159]]
```

I am guessing the client expects the container to be dead when 'start' returns an error.

bboreham:
Fixes #2284; also ends up rather simpler.  Undoes much of #2264.

The change also removes two flags from the interface of `attach()`, at the cost of a duplicate `Inspect` in case of error.  I prefer it that way.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -103,7 +103,7 @@ func (proxy *Proxy) attachWithRetry(id string) {
 
 	j := &attachJob{id: id, tryInterval: initialInterval}
 	j.timer = time.AfterFunc(time.Duration(0), func() {
-		if err := proxy.attach(id, false, false); err != nil {
+		if err := proxy.attach(id); err != nil {
 			// The delay at the nth retry is a random value in the range
 			// [i-i/2,i+i/2], where i = initialInterval * 1.5^(n-1).
 			j.timer.Reset(j.tryInterval)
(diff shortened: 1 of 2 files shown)


proxy/proxy.go
```diff
@@ -383,9 +383,20 @@ func (proxy *Proxy) listen(protoAndAddr string) (net.Listener, string, error) {
...
 func (proxy *Proxy) ContainerStarted(ident string) {
-	// In case of attach failure, if we have a request waiting on the start, kill the container,
-	// otherwise assume it is a Docker-initated restart and kill the process inside.
-	err := proxy.attach(ident, true, proxy.waitChan(ident) == nil)
+	err := proxy.attach(ident)
+	if err != nil {
+		// attach failed: if we have a request waiting on the start, kill the container,
+		// otherwise assume it is a Docker-initated restart and kill the process inside.
+		if proxy.waitChan(ident) != nil {
+			proxy.client.KillContainer(docker.KillContainerOptions{ID: ident})
```
brb: `KillContainer` might return an error. Perhaps in such case, it'd make sense to log it. Same applies to `Kill` below.

bboreham: I guess.  The previous code did `>/dev/null 2>&1`.  

BTW I was a little nonplussed that `FindProcess` can never error on Unix.

bboreham:
Added some error handling and squashed commits.

proxy/proxy.go
```diff
@@ -383,9 +383,26 @@ func (proxy *Proxy) listen(protoAndAddr string) (net.Listener, string, error) {
...
+			var c *docker.Container
+			if c, e = proxy.client.InspectContainer(ident); e == nil {
+				var process *os.Process
+				if process, e = os.FindProcess(c.State.Pid); e == nil {
+					e = process.Kill()
+				}
+			}
+		}
+		if e != nil {
+			Log.Warningf("Error killing %s: %s", ident, e)
```
brb: Maybe it makes sense to log which way of killing has failed? It would help to distinguish between two different execution paths.

bboreham: There are _four_ things which can fail, and the code looks horrible if you put different messages against each one.  I think we can worry about not knowing the code path if that ever becomes a problem.

Issue #2303, opened 2016-05-19
`weave upgrade` for easier upgrades


rade:
The idea has been mooted that perhaps we should give users a really simple/obvious way to upgrade weave, in the form of a `weave upgrade` command, especially now that `weave status` tells them when a new version is available. The command would do something like...

- download the latest script
- stop the existing containers
- replace the current script with the latest script
- start new containers, based on the new images, etc, but retaining as much info as possible (e.g. peer lists)

TBD:
- what to do when things go wrong
- dealing with non-backward-compatible upgrades

cons:
- we are somewhat re-inventing bits of a package manager here

awh:
Related: https://github.com/weaveworks/weave/issues/2186#issuecomment-211305536

PR #2305, opened 2016-05-20
Peer list persistence

### LINKED ISSUE #2186: Peer list persistence
Following up #2172, my thoughts on how peer list persistence should work from the user's perspective:

### First interactive `weave launch`

Use supplied peer list. Rationale: it's the only possibility.

### Router restarted by Docker

Use peer list supplied by Docker from immutable container state + any intervening connect/forget. Rationale: if Docker restarts for me it's because something unexpected happened, so it's reasonable for it to restore the summation of my interactive commands up to this point.

### Subsequent interactive `weave launch`

Use supplied peer list only - intervening connect/forget are ignored. Rationale - I must have stopped weave manually, and will have an expectation that what I specify now is obeyed.

### Router started by systemd (--no-restart)

Use supplied peer list only - intervening connect/forget are ignored. Rationale - I expect this to be used in automated provisionings where the 'resilience on reboot' requirement is fulfilled by distributing a systemd environment file with an updated peer list.

awh:
Fixes #2186.

Breaking change: `weave launch` is no longer idempotent (for inclusion in release notes).


rade:
A side effect of the reversion of #1967 is that `weave launch*` is no longer idempotent, which is undesirable.

rade:
> no longer idempotent

Though based on #2186 the intended semantics of `weave launch[-router]` is "Use supplied peer list only", which is non-monotonic, thus making idempotence awkward and perhaps of little value to users.

router/network_router.go
```diff
@@ -230,29 +231,21 @@ func (router *NetworkRouter) ForgetConnections(peers []string) {
...
-	/*if _, err := router.db.Load(peersIdent, &storedPeers); err != nil {
-		return nil, err
-	}*/
-
-	if storedPeers != nil && !equal(peers, storedPeers) {
-		log.Println("Overriding initial peer list with stored list:", storedPeers)
-		peers = storedPeers
-	} else {
-		log.Println("Initial set of peers:", peers)
+	if _, err := os.Stat("/restart.sentinel"); err == nil {
```
rade: If we ever wanted to run weave non-containerised, this would be an awkward place to put that file. Perhaps the current dir would do?

router/network_router.go
```diff
@@ -230,29 +231,21 @@ func (router *NetworkRouter) ForgetConnections(peers []string) {
...
+		var storedPeers []string
+		if _, err := router.db.Load(peersIdent, &storedPeers); err != nil {
+			return nil, err
+		}
+		log.Println("Docker restart detected - using persisted peer list:", storedPeers)
```
rade: s/Docker/

router/network_router.go
```diff
@@ -230,29 +231,21 @@ func (router *NetworkRouter) ForgetConnections(peers []string) {
...
 	}
-	for i, v := range a {
-		if v != b[i] {
-			return false
-		}
-	}
-	return true
+	sentinel.Close()
+
+	log.Println("Weave launch detected - using supplied peer list:", peers)
```
rade: s/Weave/

test/175_target_peer_persistence_test.sh
```diff
@@ -0,0 +1,31 @@
...
+assert_targets $HOST1 192.168.48.12 192.168.48.13
+
+# Ensure persisted peer changes are ignored after stop and subsequent restart
+weave_on $HOST1 stop
+weave_on $HOST1 launch 192.168.48.11 192.168.48.12
+assert_targets $HOST1 192.168.48.11 192.168.48.12
+check_restart $HOST1 weave
+assert_targets $HOST1 192.168.48.11 192.168.48.12
+
+end_suite
```
rade: I find the "Docker restart" terminology confusing, since it could mean "restarting Docker". I'd simply drop the "Docker" bit.

rade:
It seems rather inconvenient that re-launching the router clears the persisted peer list. It makes all changes to the configuration, e.g. changing the log level, lossy. Upgrades per #2303 are just a special case of that.

Perhaps we need another variant (or flag) for `weave launch`. After all, we do have a variant of `weave connect`, i.e. `weave connect --replace <peer-list>`.

rade:
strawman A: Add `weave launch --replace`.

- when the weave container is restarted - we detect this via the sentinel file - the persisted peer list is used
- when the weave container is launched with `--replace`, the command line peer list is used
- when the weave container is launched without `--replace`, the union of the persisted peer list and the command line peer list is used

Downside: This is a change in the current behaviour of `weave launch`.

strawman B: make `weave launch` with any peers behave as "with `--replace`" above, and "without `--replace`" otherwise

Downside: This is still a change in current behaviour for the nil command line peer list case. It is also inconsistent with the `weave connect --replace` semantics (though we could change that to match).

strawman C: `--add`, i.e. the inverse of the `--replace` flag

Downside: Having a `--replace` flag for `weave connect` and the inverse flag for `weave launch` is inconsistent.

test/175_target_peer_persistence_test.sh
awh: Done.

awh:
@rade I think I would prefer `weave launch --resume` which does not allow the specification of any peers.

awh:
CC @abuehrle doc change

site/using-weave/finding-adding-hosts-dynamically.md
```diff
@@ -41,6 +41,17 @@ will be forgotten, and the new hosts will be added:
 
     host# weave connect --replace $NEW_HOST1 $NEW_HOST2
 
+If Weave Net is restarted by Docker it will automatically remember any
+previous connect and forget operations, however if you stop it
```
rade: 'connect' and 'forget' need quoting.

site/using-weave/finding-adding-hosts-dynamically.md
```diff
@@ -41,6 +41,17 @@ will be forgotten, and the new hosts will be added:
...
+manually and launch it again it will not. If you wish to launch again
+and retain the results of those operations you can do so with
+`--resume`:
+
+    host# weave launch --resume
+
+Note that in this case you cannot specify a list of addresses as the
+previous peer list is used exclusively.
```
rade: "previous peer list" is ambiguous. "remembered peer list"?

weave
```diff
@@ -1886,14 +1858,19 @@ EOF
         ;;
     launch)
         deprecation_warnings "$@"
+        check_not_running router $CONTAINER_NAME        $BASE_IMAGE
+        check_not_running proxy  $PROXY_CONTAINER_NAME  $BASE_EXEC_IMAGE
+        # Don't check the plugin here, since we start it with `--restart=always`
```
bboreham: This comment seems inappropriate now we start everything with `--restart=always`.  Presumably we should check all three are not running.

bboreham:
LGTM apart from one comment.

I guess we should note in the description the breaking change that `weave launch` is no longer idempotent, for inclusion in release notes.

Issue #1726, opened 2015-11-26
Make veth settings closer to those produced by the weave script


awh:
The `weave` script does a few things like `ethtool tx off` that the plugin doesn't currently do.

Most of these are not relevant to FDP, only the "slow" path.

Replaces weaveworks/docker-plugin#53; see there for additional commentary.

bboreham:
Also the `weave` script adds a multicast route for each container

bboreham:
Multicast route covered by #1776; `ethtool tx off` less important now we're mostly not using pcap.

rade:
> `ethtool tx off` less important now we're mostly not using pcap.

See #1788 for why that is the case.

rade:
So what is left to do here?

bboreham:
The predecessor issue noted "The arp stuff is relevant".

The `ethtool` point also relates to #1255

bboreham:
The Docker plugin is not given direct access to the container's namespace, so adjusting its arp settings requires an indirect approach.  We could do it in the same place we add DNS names, i.e. after the container has started.

Note #2030 adds a utility function to do the arp settings.

bboreham:
`tx off` now implemented by #2307; only arp settings and gratuitous arp remain.

These would have to be done asynchronously by the plugin, after the container starts, because they need to be done in the container's namespace.  Docker does not provide a hook to do this.

Issue #639, opened 2015-05-07
arping on container start/attach limits scalability


rade:
As of #457, we send an unsolicited ARP response whenever a container is attached to the weave network. We do this in order to work around a bug in the Linux kernel. More details in #441.

This does limit scalability, since a) the ARP response is broadcast, and b) weave create an entry in their MAC cache when seeing that response. Both of these are especially problematic when there is lots of container churn, and on restart of a cluster, when there are potentially lots of containers being attached to the weave network in a short space of time.

rade:
The fix to the kernel bug is in 3.19. Which will take a while to get widely deployed. But perhaps we should make the unsolicited ARPing conditional on pre-3.19 kernels, so that on more recent kernels we can avoid the aforementioned downsides. The only problem is that we'd really need to know the kernel version on all the hosts *other* than ourselves. Hmm. I suspect most deployments are uniform though. And I guess we could add some switch/env-var to depart from whatever behaviour we make the default.

@dpw, what do you reckon?

dpw:
Have we or our users observed measurable performance consequences of the unsolicited ARP? It sounds marginal to me, even if you had a big weave cluster with all hosts churning away starting containers, but that's just a guess.

rade:
Oh, I'm quite sure it would take "a while" to run into this. It would likely manifest itself as packet loss.

It's a bit of a blemish. AFAIK it's the only thing in weave that would impose a *global* limit on the rate of container starts.

dpw:
Even if we're looking at this because it might bite someone in the future, rather than because it is a problem for anyone now, it would still be good to see a projection based on some kind of measurements (e.g. might it become an issue at 100 hosts? 1000? 10000?).

Also, if you have that much container churn, aren't they likely to contain client applications, and therefore be ARPing in order to locate servers? Ok, unsolicited ARPs will make matters worse, but if we are worried about ARP limiting weave scalability, surely it would be good to have some estimate of where that limit lies taking all causes of ARPs into account.

rade:
I am not opposed to running some measurements.

Good point re client containers ARPing for servers.

PR #2307, opened 2016-05-23
Move container attachment from sh to Go

bboreham:
Part of #1940 

Most of this is refactoring/enhancing the current plugin code to be closer to the script; see also #639 for `arping` and #1726 for `ethtool tx off`.

The plugin previously leaked a veth if something went wrong configuring it.

This PR does not mean the proxy can avoid shelling out when attaching a container.  Still to do: allocate IP addresses, rewrite hosts file, add to DNS.

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -2,12 +2,14 @@ package common
 
 import (
 	"fmt"
-	"github.com/vishvananda/netlink"
-	"github.com/vishvananda/netns"
 	"net"
 	"os"
-	"runtime"
 	"strings"
+
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+
+	weavenet "github.com/weaveworks/weave/net"
 )
 
 // Assert test is true, panic otherwise
@@ -23,25 +25,6 @@ func ErrorMessages(errors []error) string {
 		result = append(result, err.Error())
 	}
 	return strings.Join(result, "\n")
-}
-
-func WithNetNS(ns netns.NsHandle, work func() error) error {
-	runtime.LockOSThread()
-	defer runtime.UnlockOSThread()
-
-	oldNs, err := netns.Get()
-	if err == nil {
-		defer oldNs.Close()
-
-		err = netns.Set(ns)
-		if err == nil {
-			defer netns.Set(oldNs)
-
-			err = work()
-		}
-	}
-
-	return err
 }
 
 type NetDev struct {
@@ -63,7 +46,7 @@ func FindNetDevs(processID int, match func(link netlink.Link) bool) ([]NetDev, e
 	}
 	defer ns.Close()
 
-	err = WithNetNS(ns, func() error {
+	err = weavenet.WithNetNS(ns, func() error {
 		return forEachLink(func(link netlink.Link) error {
 			if match(link) {
 				netDev, err := linkToNetDev(link)
--- a/net/route.go
+++ b/net/route.go
@@ -3,6 +3,7 @@ package net
 import (
 	"fmt"
 	"net"
+	"os"
 
 	"github.com/vishvananda/netlink"
 )
@@ -54,4 +55,17 @@ func forEachRoute(ignoreIfaceNames map[string]struct{}, check func(r netlink.Rou
 		}
 	}
 	return nil
+}
+
+func AddRoute(link netlink.Link, scope netlink.Scope, dst *net.IPNet, gw net.IP) error {
+	err := netlink.RouteAdd(&netlink.Route{
+		LinkIndex: link.Attrs().Index,
+		Scope:     scope,
+		Dst:       dst,
+		Gw:        gw,
+	})
+	if os.IsExist(err) { // squash duplicate route errors
+		err = nil
+	}
+	return err
 }
--- a/net/veth.go
+++ b/net/veth.go
@@ -28,29 +28,34 @@ func CreateAndAttachVeth(localName, peerName, bridgeName string, mtu int) (*netl
 		return nil, fmt.Errorf(`could not create veth pair %s-%s: %s`, local.Name, local.PeerName, err)
 	}
 
+	cleanup := func(format string, a ...interface{}) (*netlink.Veth, error) {
+		netlink.LinkDel(local)
+		return nil, fmt.Errorf(format, a...)
+	}
+
 	switch maybeBridge.(type) {
 	case *netlink.Bridge:
 		if err := netlink.LinkSetMasterByIndex(local, maybeBridge.Attrs().Index); err != nil {
-			return nil, fmt.Errorf(`unable to set master of %s: %s`, local.Name, err)
+			return cleanup(`unable to set master of %s: %s`, local.Name, err)
 		}
 	case *netlink.GenericLink:
 		if maybeBridge.Type() != "openvswitch" {
-			return nil, fmt.Errorf(`device "%s" is of type "%s"`, bridgeName, maybeBridge.Type())
+			return cleanup(`device "%s" is of type "%s"`, bridgeName, maybeBridge.Type())
 		}
 		if err := odp.AddDatapathInterface(bridgeName, local.Name); err != nil {
-			return nil, fmt.Errorf(`failed to attach %s to device "%s": %s`, local.Name, bridgeName, err)
+			return cleanup(`failed to attach %s to device "%s": %s`, local.Name, bridgeName, err)
 		}
 	case *netlink.Device:
 		// Assume it's our openvswitch device, and the kernel has not been updated to report the kind.
 		if err := odp.AddDatapathInterface(bridgeName, local.Name); err != nil {
-			return nil, fmt.Errorf(`failed to attach %s to device "%s": %s`, local.Name, bridgeName, err)
+			return cleanup(`failed to attach %s to device "%s": %s`, local.Name, bridgeName, err)
 		}
 	default:
-		return nil, fmt.Errorf(`device "%s" is not a bridge`, bridgeName)
+		return cleanup(`device "%s" is not a bridge`, bridgeName)
 	}
 
 	if err := netlink.LinkSetUp(local); err != nil {
-		return nil, fmt.Errorf("unable to bring veth up: %s", err)
+		return cleanup("unable to bring veth up: %s", err)
 	}
 
 	return local, nil
--- a/net/veth.go
+++ b/net/veth.go
@@ -85,14 +85,30 @@ func SetupGuest(guest netlink.Link, name string) error {
 	return nil
 }
 
-func AddAddresses(guest netlink.Link, cidrs []*net.IPNet) error {
-	for _, ipnet := range cidrs {
-		if err := netlink.AddrAdd(guest, &netlink.Addr{IPNet: ipnet}); err != nil {
-			return fmt.Errorf("failed to add IP address to %q: %v", guest.Attrs().Name, err)
-		}
-		arping.GratuitousArpOverIfaceByName(ipnet.IP, guest.Attrs().Name)
+func AddAddresses(guest netlink.Link, cidrs []*net.IPNet) (newAddrs []*net.IPNet, err error) {
+	existingAddrs, err := netlink.AddrList(guest, netlink.FAMILY_V4)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get IP address for %q: %v", guest.Attrs().Name, err)
 	}
-	return nil
+	for _, ipnet := range cidrs {
+		if contains(existingAddrs, ipnet) {
+			continue
+		}
+		if err := netlink.AddrAdd(guest, &netlink.Addr{IPNet: ipnet}); err != nil {
+			return nil, fmt.Errorf("failed to add IP address to %q: %v", guest.Attrs().Name, err)
+		}
+		newAddrs = append(newAddrs, ipnet)
+	}
+	return newAddrs, nil
+}
+
+func contains(addrs []netlink.Addr, addr *net.IPNet) bool {
+	for _, x := range addrs {
+		if addr.IP.Equal(x.IPNet.IP) {
+			return true
+		}
+	}
+	return false
 }
 
 func interfaceExistsInNamespace(ns netns.NsHandle, ifName string) bool {
@@ -103,13 +119,14 @@ func interfaceExistsInNamespace(ns netns.NsHandle, ifName string) bool {
 	return err == nil
 }
 
-func AttachContainer(ns netns.NsHandle, id, ifName, bridgeName string, mtu int, cidrs []*net.IPNet) error {
+func AttachContainer(ns netns.NsHandle, id, ifName, bridgeName string, mtu int, withMulticastRoute bool, cidrs []*net.IPNet) error {
 	if !interfaceExistsInNamespace(ns, ifName) {
 		if len(id) > 5 {
 			id = id[:5]
 		}
 		name, peerName := "vethwepl"+id, "vethwg"+id
 		_, err := CreateAndAttachVeth(name, peerName, bridgeName, mtu, func(local, guest netlink.Link) error {
+			EthtoolTXOff(peerName) // TODO: do we want to do this under fastdp?
 			if err := netlink.LinkSetNsFd(guest, int(ns)); err != nil {
 				return fmt.Errorf("failed to move veth to container netns: %s", err)
 			}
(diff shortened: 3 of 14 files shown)


.gitmodules
```diff
@@ -70,3 +70,9 @@
 [submodule "vendor/github.com/pkg/profile"]
 	path = vendor/github.com/pkg/profile
 	url = https://github.com/pkg/profile
+[submodule "vendor/github.com/bboreham/arping"]
```
brb: Your [PR](https://github.com/j-keck/arping/pull/1) has been merged, so I assume that this submodule is not  needed.

bboreham: Fixed (by forced update)

common/ethtool.go
```diff
@@ -0,0 +1,51 @@
...
+	if len(name) > 15 {
+		return fmt.Errorf("name too long")
+	}
+
+	value := EthtoolValue{ETHTOOL_STXCSUM, 0}
+	request := IFReqData{Data: uintptr(unsafe.Pointer(&value))}
+
+	copy(request.Name[:], name)
+
+	socket, err := syscall.Socket(syscall.AF_INET, syscall.SOCK_DGRAM, 0)
```
brb: `ethtool` uses `PF_INET`, but according to the [internet](http://stackoverflow.com/questions/6729366/what-is-the-difference-between-af-inet-and-pf-inet-in-socket-programming) it should not matter.

common/utils.go
```diff
@@ -44,6 +44,16 @@ func WithNetNS(ns netns.NsHandle, work func() error) error {
 	return err
 }
 
+func WithNetNSLink(ns netns.NsHandle, ifName string, work func(link netlink.Link) error) error {
```
brb: I'd move net ns related functions from `common/utils.go` to `common/netns.go`

net/route.go
```diff
@@ -55,3 +56,16 @@ func forEachRoute(ignoreIfaceNames map[string]struct{}, check func(r netlink.Rou
...
 }
+
+func AddRoute(link netlink.Link, scope netlink.Scope, dst *net.IPNet, gw net.IP) error {
+	err := netlink.RouteAdd(&netlink.Route{
+		LinkIndex: link.Attrs().Index,
+		Scope:     scope,
+		Dst:       dst,
+		Gw:        gw,
+	})
+	if os.IsExist(err) { // squash duplicate route errors
```
brb: Are you sure that is the *right* way to check?

net/veth.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"github.com/j-keck/arping"
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+
+	"github.com/weaveworks/weave/common"
+	"github.com/weaveworks/weave/common/odp"
+)
+
+// create and attach local name to the Weave bridge
```
brb: Maybe be more specific by saying what `local name` is.

net/veth.go
```diff
@@ -0,0 +1,198 @@
...
+		LinkAttrs: netlink.LinkAttrs{
+			Name: localName,
+			MTU:  mtu},
+		PeerName: peerName,
+	}
+	if err := netlink.LinkAdd(local); err != nil {
+		return nil, fmt.Errorf(`could not create veth pair %s-%s: %s`, local.Name, local.PeerName, err)
+	}
+
+	cleanup := func(format string, a ...interface{}) (*netlink.Veth, error) {
```
brb: You could use `defer` instead.

net/veth.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	if err := netlink.LinkSetUp(local); err != nil {
+		return cleanup("unable to bring veth up: %s", err)
+	}
+
+	return local, nil
+}
+
+func SetupGuest(guest netlink.Link, name string) error {
```
brb: It'd be useful to provide a comment on what `Guest` is/means in the context of weave.

net/veth.go
```diff
@@ -0,0 +1,198 @@
...
+	if err = netlink.LinkSetName(guest, name); err != nil {
+		return err
+	}
+	if err = common.ConfigureARPCache(name); err != nil {
+		return err
+	}
+	return nil
+}
+
+func AddAddresses(guest netlink.Link, cidrs []*net.IPNet) (newAddresses []*net.IPNet, err error) {
```
brb: s/guest/link/ ?

net/veth.go
```diff
@@ -0,0 +1,198 @@
...
+	existingAddrs, err := netlink.AddrList(guest, netlink.FAMILY_V4)
```
brb: nitpicking on consistency: existing**Addrs** and new**Addresses**.

net/veth.go
```diff
@@ -0,0 +1,198 @@
...
+	return false
+}
+
+func AttachContainer(ns netns.NsHandle, id, ifName, bridgeName string, mtu int, withMulticastRoute bool, cidrs []*net.IPNet) error {
+	// Check if interface already exists
+	if err := common.WithNetNS(ns, func() error {
+		_, err := netlink.LinkByName(ifName)
+		return err
+	}); err != nil {
+		if len(id) > 5 {
```
brb: I'd introduce const for container iface name prefix and would use `len(thatPrefix)` instead of `5`.

net/veth.go
```diff
@@ -0,0 +1,198 @@
...
+			id = id[:5]
+		}
+		name, peerName := "vethwepl"+id, "vethwg"+id
```
brb: Where does `vethwepl` come from? Grepping for it on master branch does not find anything, only `vethwl`.

brb: Don't mind. @rade just showed me this https://github.com/weaveworks/weave/blob/master/weave#L814

Anyway, I'd create a const for these values.

net/route.go
bboreham: I believe it was constructed empirically, from observing the "File exists" message from `ip route`, then examining the construction of `os.IsExist()`.  It seems better to depend on `os` rather than `syscall` as it is higher-level.

But this code has just been moved out of the CNI plugin; if we think it's wrong then that should be raised as a separate issue.

common/utils.go
bboreham: Yes; or perhaps `weaveworks/weave/net/netns.go`

common/ethtool.go
bboreham: ping @awh who wrote this

net/veth.go
bboreham: I think that has other issues; you'd have to arrange that `err` always refers to the right scope.

net/veth.go
bboreham: The limit should be the maximum netlink device name length _minus_ "vethwepl".  Which is 7.
I think 5 was picked because originally we were using pids and the longest pid is 5 characters, and we want the names to line up.  But that doesn't seem like a good reason.

net/veth.go
bboreham: Fixed - "local name" was a typo, but I have removed "local" entirely.

net/veth.go
bboreham: Fixed by inlining this function that was only called once, but I have also removed "guest" entirely.

net/veth.go
bboreham: done

prog/weaveutil/attach.go
```diff
@@ -0,0 +1,105 @@
...
+	err = weavenet.AttachContainer(ns, fmt.Sprint(pid), weavenet.VethName, args[1], mtu, withMulticastRoute, cidrs)
+	// If we detected an error but the container has died, tell the user that instead.
+	if err != nil && !processExists(pid) {
+		err = fmt.Errorf("Container %s died", args[0])
+	}
+	return err
+}
+
+func findPid(containerID string) (int, error) {
+	c, err := docker.NewVersionedClientFromEnv("1.18")
```
brb: `"1.18"` as a const?

net/veth.go
bboreham: Fixed

prog/weaveutil/attach.go
```diff
@@ -0,0 +1,105 @@
...
+
+	pid, err := findPid(args[0])
+	if err != nil {
+		return err
+	}
+	ns, err := netns.GetFromPid(pid)
+	if err != nil {
+		return err
+	}
+	if nsToplevel, err := netns.GetFromPid(1); err != nil {
```
brb: Maybe:
`nsToplevel` -> `nsHost`
`ns` -> `nsContainer`
`"unable to open root namespace"` -> `"unable to open host namespace"`

prog/weaveutil/attach.go
```diff
@@ -0,0 +1,105 @@
...
+
+	withMulticastRoute := true
+	if args[0] == "--no-multicast-route" {
+		withMulticastRoute = false
+		args = args[1:]
+	}
+
+	pid, err := findPid(args[0])
+	if err != nil {
+		return err
```
brb: Have you thought about wrapping each returned err? E.g. `return fmt.Errorf("findPid failed: %s", err)`. It might be a bit redundant, but IMHO it would improve debugging.

prog/weaveutil/attach.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	cidrs, err := parseCIDRs(args[3:])
+	if err != nil {
+		return err
+	}
+	err = weavenet.AttachContainer(ns, fmt.Sprint(pid), weavenet.VethName, args[1], mtu, withMulticastRoute, cidrs)
+	// If we detected an error but the container has died, tell the user that instead.
+	if err != nil && !processExists(pid) {
+		err = fmt.Errorf("Container %s died", args[0])
```
brb: The original error is lost here.

bboreham: That is the desired behaviour, hence "tell the user that instead".
The container dying will cause the namespace to disappear, which will manifest as all sorts of network failures, that nobody cares about because the interesting thing is that the container died.

prog/weaveutil/attach.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return
+}
+
+func detach(args []string) error {
+	if len(args) < 2 {
+		cmdUsage("detach-container", "<container-id> <cidr>...")
+	}
+
+	pid, err := findPid(args[0])
```
brb: DRY: Perhaps, `findPid` and `GetFromPid` can be moved to the `containerPidAndNs` function.

weave
```diff
@@ -2184,7 +2059,7 @@ EOF
         [ $# -eq 1 ] || usage
         CONTAINER=$(container_id $1)
         ipam_cidrs lookup $CONTAINER $CIDR_ARGS
-        with_container_netns $CONTAINER detach $ALL_CIDRS >/dev/null
+        util_op detach-container $CONTAINER $ALL_CIDRS
```
brb: Suppress stdout?

bboreham: My reasoning was that I didn't code it to send anything to `stdout` so there was no reason to suppress it.

But equally there is no problem suppressing it.

.gitmodules
```diff
@@ -70,3 +70,6 @@
 [submodule "vendor/github.com/pkg/profile"]
 	path = vendor/github.com/pkg/profile
 	url = https://github.com/pkg/profile
+[submodule "vendor/github.com/j-keck/arping"]
```
brb: The `arping` project does not have any tests :-(

Will weave integration tests fail if the arping does not work as expected?

bboreham: yes, test 300

plugin/net/driver.go
```diff
@@ -100,22 +100,16 @@ func (driver *driver) EndpointInfo(req *api.EndpointInfoRequest) (*api.EndpointI
 func (driver *driver) JoinEndpoint(j *api.JoinRequest) (*api.JoinResponse, error) {
 	driver.logReq("JoinEndpoint", j, fmt.Sprintf("%s:%s to %s", j.NetworkID, j.EndpointID, j.SandboxKey))
 
-	local, err := createAndAttach(j.EndpointID, WeaveBridge, 0)
-	if err != nil {
+	name, peerName := vethPair(j.EndpointID)
+	if _, err := weavenet.CreateAndAttachVeth(name, peerName, WeaveBridge, 0, nil); err != nil {
```
brb: The previous function used here did not delete veth in the case of error. Is it safe to delete it?

bboreham: Yes, that's part of "refactoring the current plugin code to be closer to the script"; since we have just created the veth, and we don't tell anyone else about it until we get to `JoinResponse` below, we should delete it on error otherwise it leaks.

net/veth.go
bboreham: I rewrote this code.

prog/weaveutil/attach.go
bboreham: `findPid` seems a bad example, because every error there is already wrapped with a description.

prog/weaveutil/attach.go
bboreham: Done

net/veth.go
```diff
@@ -0,0 +1,200 @@
...
+	vethPrefix       = "v" + VethName // starts with "veth" to suppress UI notifications
+	maxNetDevNameLen = 15
+)
+
+func AttachContainer(ns netns.NsHandle, id, ifName, bridgeName string, mtu int, withMulticastRoute bool, cidrs []*net.IPNet) error {
+	// Check if interface already exists
+	if err := common.WithNetNS(ns, func() error {
+		_, err := netlink.LinkByName(ifName)
+		return err
+	}); err != nil {
```
brb: A bit difficult to read. Maybe create a var for the work func?

net/veth.go
```diff
@@ -0,0 +1,200 @@
...
+		maxIDLen := maxNetDevNameLen - len(vethPrefix+"pl")
+		if len(id) > maxIDLen {
+			id = id[:maxIDLen] // trim passed ID if too long
+		}
+		name, peerName := vethPrefix+"pl"+id, vethPrefix+"pg"+id
+		_, err = CreateAndAttachVeth(name, peerName, bridgeName, mtu, func(veth netlink.Link) error {
+			common.EthtoolTXOff(peerName) // TODO: do we want to do this under fastdp?
```
brb: For `fastdp` it used to be a noop.

common/ethtool.go
```diff
@@ -0,0 +1,51 @@
...
+
+// linux/ethtool.h 'struct ethtool_value'
+type EthtoolValue struct {
+	Cmd  uint32
+	Data uint32
+}
+
+// Disable TX checksum offload on specified interface
+func EthtoolTXOff(name string) error {
+	if len(name) > 15 {
```
brb: Is it related to `maxNetDevNameLen` ?

bboreham: Yes, but possibly more related to the [16] earlier.
Which comes from http://lxr.free-electrons.com/source/include/uapi/linux/if.h#L28
I will consider how to combine them.

net/veth.go
bboreham: Created `interfaceExistsInNamespace()`; also it turns out errors from `CreateAndAttachVeth` were being ignored.

net/veth.go
```diff
@@ -0,0 +1,200 @@
...
+	if err := common.WithNetNS(ns, func() error {
+		_, err := netlink.LinkByName(ifName)
+		return err
+	}); err != nil {
+		maxIDLen := maxNetDevNameLen - len(vethPrefix+"pl")
+		if len(id) > maxIDLen {
+			id = id[:maxIDLen] // trim passed ID if too long
+		}
+		name, peerName := vethPrefix+"pl"+id, vethPrefix+"pg"+id
+		_, err = CreateAndAttachVeth(name, peerName, bridgeName, mtu, func(veth netlink.Link) error {
```
brb: Just noting, that the order of ops is different from `connect_container_to_bridge`. Hope it does not create undesired behaviours.

net/veth.go
brb: Why is return value suppressed?

bboreham: Suppressing the return value appears to be an error.
The "TODO" is here because, at the moment we cannot tell the difference here between bridge and bridged_fastdp, so more info would need to be passed in.

But, I think it's important not to turn tx offload off for AWSVPC and similar.

common/utils.go
bboreham: Moved netns code to net/

net/veth.go
bboreham: Fixed

common/ethtool.go
bboreham: Combined `maxNetDevNameLen` and `16` as `IFNAMSIZ`

net/veth.go
bboreham: The old order in the `weave` script was:

create veth
ethtool tx off
move guest end inside netns
set name inside netns
set host end up
add host end to bridge
configure arp cache
add addresses
set guest end up
send gratuitous arp

The old order in the CNI plugin was:

create veth
add host end to bridge
move guest end inside netns
set host end up
set name inside netns
set guest end up
configure arp cache
add address

The new order is:

create veth
add host end to bridge
ethtool tx off
move guest end inside netns
set name inside netns
configure arp cache
set host end up
add addresses
set guest end up
send gratuitous arp

The previous differences between script and CNI indicate that the order doesn't matter too much, although CNI callers probably don't attempt to use the interface before the plugin returns after completing all tasks.

Most of the differences arise from sharing implementation with the Docker plugin, which cannot do the parts inside the container's netns.

Setting the host end up later seems like an improvement.

net/veth.go
```diff
@@ -0,0 +1,201 @@
...
+			if err := EthtoolTXOff(peerName); err != nil {
+				return cleanup(`unable to set tx off on %q: %s`, peerName, err)
+			}
+		}
+	case Fastdp:
+		if err := odp.AddDatapathInterface(bridgeName, name); err != nil {
+			return cleanup(`failed to attach %s to device "%s": %s`, name, bridgeName, err)
+		}
+	default:
+		return cleanup(`invalid bridge configuration`)
```
brb: Maybe we want to log whether bridge type is `None` or `Inconsistent`?

bboreham: It can't be `None` by this stage: we detect "bridge not present" at the top of the function.

PR #2313, opened 2016-05-24
Send out arp update from CNI plugin

bboreham:
To fix issues of packets disappearing when you re-use an IP address with a different MAC.

This brings the CNI plugin closer in behaviour to the weave script.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -70,3 +70,6 @@
 [submodule "vendor/github.com/coreos/go-iptables"]
 	path = vendor/github.com/coreos/go-iptables
 	url = https://github.com/coreos/go-iptables
+[submodule "vendor/github.com/j-keck/arping"]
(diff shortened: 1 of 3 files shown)


brb:
Would it be difficult to provide a test case for it?

bboreham:
I have added a smoke-test.

Doing this uncovered a bug in the previous test where it was allocating the same IP address to each container because it passed in "c1" as the ID for each of them.

However the smoke-test does not work without the `sleep` between killing the old container and creating the new container.  I haven't yet figured out why it is necessary.

errordeveloper:
Happy to test with KA, if there is an image on Docker Hub.

brb:
LGTM.

Re `sleep 1`, as discussed IRL, @bboreham is going to create a new issue.

PR #2317, opened 2016-05-26
Discovered peers persistence

### LINKED ISSUE #2187: Persist discovered peers
When you manually add a new peer to an existing network it is not desirable to go back to the existing peers and do `weave connect` (assuming that persists as per #2186) - it would be better if existing peers remembered the new peer automatically.

awh:
Extends weaveworks/weave#2305, which should be merged first. Only the last two commits in this PR are relevant.

Depends on weaveworks/mesh#46, which should also be merged first.

Fixes #2187.

There are some issues:
1. The supporting mesh PR is a bit grotty - see there for commentary.
2. Due to the way `ConnectionMaker` periodically synchronises the set of targets to `cm.directPeers` + the subset of `cm.peers` to which it does not have connections, it was necessary to introduce a third data structure (`cm.indirectPeers`) to include discovered peers from a previous run in the synchronisation. Whereas previously we would stop trying to connect to 'indirect' peers if they had disappeared from `cm.peers`, that is no longer the case after a resume; `cm.indirectPeers` can accumulate garbage which is only collected by `weave stop` (Note `weave forget` does not help, because that only removes things from `cm.directPeers`).

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -255,7 +255,8 @@ func main() {
 	router := weave.NewNetworkRouter(config, networkConfig, name, nickName, overlay, db)
 	Log.Println("Our name is", router.Ourself)
 
-	if peers, err = router.InitialPeers(resume, peers); err != nil {
+	var indirectPeers []string
+	if peers, indirectPeers, err = router.InitialPeers(resume, peers); err != nil {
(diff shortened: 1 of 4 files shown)


bboreham:
Needs rebase

bboreham:
Rebased.

bboreham:
Closing due to inactivity

Issue #1960, opened 2016-02-08
Can't connect a container to two weave plugin networks


bboreham:
````
$ docker network create --driver=weavemesh weave2
$ docker run --name=foo -tdi --net=weave alpine /bin/sh
c4a6967fd9e80a32e6c37df6c045c349db7a868fe7e45606de62eba7dbfe9a76
$ docker network connect weave2 foo
Error response from daemon: failed to add interface vethwg89e21 to sandbox: error setting interface "vethwg89e21" routes to ["224.0.0.0/4"]: file exists
````

This is because the plugin asks Docker to set up the multicast route, and Linux doesn't let you add the same route pointing in two directions.

The plugin isn't given enough context to know there is already a multicast route.

rade:
Is there a way to pass some config information ("don't create routes") to the plugin as part of 'create', and for that information then to be accessible at attach time?

rade:
> Is there a way to pass some config information ("don't create routes") to the plugin as part of 'create'?

Looks like that wouldn't be sufficient for at least one user:

> I have multiple apps/sites running in a docker swarm. Each app is distributed over multiple containers and has it's own weave network. The containers need multicast. I also want to have 1 ip-adress for the entire cluster, with a HAProxy that routes traffic to the proper app. This HAProxy is the only container connected to multiple weave networks. The HAProxy does not need multicast.

bboreham:
> Is there a way to pass some config information to the plugin as part of 'create'

Yes:

    docker network create --driver=weavemesh --opt=[foo=bar] mynet

> and for that information then to be accessible at attach time?

It's available via `docker network inspect <network_id>`, which would avoid us needing to persist it.

bboreham:
There is no similar interface for `docker network connect` or for connecting as part of `docker run`, to specify parameters per-container.

sheerun:
Launching plugin with `weave launch-plugin --no-multicast-route` is sufficient workaround for me now.

sheerun:
@bboreham @rade Could I get confirmation the docket network with `weavemesh` driver need to be created manually on each host? This is different behavior from native overlay network as it needs to be created only on single host and then is automatically propagated. If no then I'll open an issue.

awh:
> Could I get confirmation the docker network with weavemesh driver need to be created manually on each host?

It does need to be created on each host, but not manually as it's done for you by `weave launch`/`weave launch-plugin`

[...]

PR #2327, opened 2016-05-31
Add an option to disable multicast route per-network

bboreham:
This gives users a better way out of the problem in #1960

### DIFF
--- a/prog/plugin/main.go
+++ b/prog/plugin/main.go
@@ -42,7 +42,7 @@ func main() {
 	flag.StringVar(&logLevel, "log-level", "info", "logging level (debug, info, warning, error)")
 	flag.StringVar(&address, "socket", "/run/docker/plugins/weave.sock", "socket on which to listen")
 	flag.StringVar(&meshAddress, "meshsocket", "/run/docker/plugins/weavemesh.sock", "socket on which to listen in mesh mode")
-	flag.BoolVar(&noMulticastRoute, "no-multicast-route", false, "do not add a multicast route to network endpoints")
+	flag.BoolVar(&noMulticastRoute, "no-multicast-route", false, "deprecated (this is now the default)")
 
 	flag.Parse()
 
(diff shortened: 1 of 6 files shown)


rade:
Docs?

Does docker allow you to specify option *flags*, i.e. no value, e.g. `--opt works.weave.nomulticast`? That would be marginally preferable in the UI, though I gather that the option actually acts as an override for whatever the overall plugin setting ("--no-multicast" or not) is, and for that the explicit true/false does work better.

bboreham:
Yes, you can say `docker network create --driver=weavemesh --opt works.weave.nomulticast weave2`.  What arrives at the plugin is a map from `works.weave.nomulticast` to blank.

bboreham:
I amended the code to look for `nomulticast`, and added docs.  Ping @abuehrle for docs review.

Note that you can say `--opt works.weave.nomulticast=false` if you want to invert the default.  I didn't document this, as it seems a niche interest.

rade:
See [my comment](https://github.com/weaveworks/weave/issues/1960#issuecomment-222823591) how I reckon the various options should interplay. In summary

- get rid of the `--no-multicast-route` option on the plugin
- created networks have multicast disabled by default
- supplying the `works.weave.multicast` option on network creation enables multicast
- create the `weave` network with multicast enabled

bboreham:
> get rid of the `--no-multicast-route` option on the plugin

assume you mean "issue a deprecation warning and ignore"

bboreham:
Changed as suggested

prog/plugin/main.go
```diff
@@ -73,26 +73,29 @@ func main() {
 	}
 
 	Log.Println("Weave plugin", version, "Command line options:", os.Args[1:])
+	if noMulticastRoute {
+		Log.Warning("--no-multicast-route option has been removed; multicast is off by default")
+	}
```
rade: we may want to provide a clue as to how to enable it.

site/plugin/plugin-how-it-works.md
```diff
@@ -26,6 +26,13 @@ store, but the first part of
 
 >**Note:** In the case of multiple networks using the `weave` driver, all containers are on the same virtual network but Docker allocates their addresses on different subnets so they cannot talk to each other directly.
 
+The plugin accepts the following options via `docker network create ... --opt`:
+
+ * `works.weave.multicast` -- tells weave to add a static IP
+   route for multicast traffic onto its interface.
+
+>**Note:** If you connect a container to multiple Weave networks, at
+   most one of them can have the multicast route enabled.
```
rade: we ought to mention somewhere that the `weave` network has multicast enabled.

prog/plugin/main.go
bboreham: maybe, but presumably a person who specified `--no-multicast-route` is not the target audience.

rade: d'oh, right.

awh:
LGTM

bboreham:
Added mention of how `--net=weave` has multicast turned on but `docker network create --driver=weave` will default to no multicast.  Also removed obsolete flag from docs.

PR #2331, opened 2016-06-02
Augment warning on active plugin endpoints

### LINKED ISSUE #2320: augment dire warning when `weave stop`ing and there are containers attached via the plugin
```
$ weave stop
unable to remove network: API error (500): network weave has active endpoints

WARNING: docker may hang in this state; re-launch weave and remove attached containers before shutting down
```

Firstly, iirc docker no longer completely hangs in this state. Furthermore, all is well if you restart weave. That is kinda implied by the warning, but could be made clearer. In particular, it is perfectly ok to `weave stop`, perhaps upgrade, and then `weave launch` with some tweaked options. As it stands, the error and warning may put off users from doing exactly that.

bboreham:
Fixes #2320 

Example message:

````
$ weave stop-plugin
WARNING: the following containers are still attached to network "weave":
  9e396c66e0d2 
Docker operations involving those containers may pause or fail
while Weave is not running
````

### DIFF
--- a/prog/weaveutil/plugin_network.go
+++ b/prog/weaveutil/plugin_network.go
@@ -47,6 +47,18 @@ func removePluginNetwork(args []string) error {
 	}
 	err = d.RemoveNetwork(networkName)
 	if _, ok := err.(*docker.NoSuchNetwork); !ok && err != nil {
+		if info, err2 := d.NetworkInfo(networkName); err2 == nil {
+			if len(info.Containers) > 0 {
+				containers := ""
+				for container := range info.Containers {
+					containers += fmt.Sprintf("  %.12s ", container)
+				}
+				return fmt.Errorf(`WARNING: the following containers are still attached to network %q:
+%s
+Docker operations involving those containers may pause or fail
+while Weave is not running`, networkName, containers)
+			}
+		}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: prog/weaveutil/plugin_network.go [hunk 1]
```diff
@@ -47,6 +47,19 @@ func removePluginNetwork(args []string) error {
...
+		if info, err2 := d.NetworkInfo(networkName); err2 == nil {
+			if len(info.Containers) > 0 {
+				containers := ""
+				for container := range info.Containers {
+					containers += fmt.Sprintf("  %.12s ", container)
+				}
+				return fmt.Errorf(`WARNING: the following containers are still attached to network %q:
+%s
+Docker operations involving those containers may pause or fail in this state.
+If you do not intend to carry on using Weave, re-launch
```
rade: s/re-launch/re-launch Weave.

prog/weaveutil/plugin_network.go
```diff
@@ -47,6 +47,19 @@ func removePluginNetwork(args []string) error {
...
+and remove those containers before shutting down.`, networkName, containers)
```
rade: s/shutting down/stopping Weave.

bboreham:
I changed the text of the error message after comments.

PR #2332, opened 2016-06-02
Warn on 'weave reset' if DELETE fails

### LINKED ISSUE #2326: `weave reset` silently skips freeing IPAM if router is down
We should print a warning as per https://github.com/weaveworks/weave/pull/1978#discussion_r64896561

bboreham:
Fixes #2326 

I did not put anything in about re-launching and re-resetting, because during `reset` we will have deleted the persistence data so re-resetting would require that we can connect to another peer who can supply us with our old IPAM ring and then do the reset.

### DIFF
--- a/weave
+++ b/weave
@@ -90,7 +90,7 @@ weave stop
       stop-proxy
       stop-plugin
 
-weave reset
+weave reset         [--force]
       rmpeer        <peer_id> ...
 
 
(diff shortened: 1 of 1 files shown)


rade:
Hmm. If the user has just run `weave stop`, this is very much a "you shouldn't have done that; but I'll carry on anyway" behaviour. Or, it can be an annoyance, in the case where there was just one node.

Perhaps a) the warning should more directly highlight that problem, and b) instead of warning this should be an error, which the user can get around with a `--force`.

bboreham:
> highlight that problem

what is "that problem"?  They shouldn't have run `stop` ?

If the underlying issue is that the weave router has crashed (maybe is repeatedly crashing), this would be unhelpful.

I guess we can check with Docker that the `weave` container exists, is not running, `Restarting` is false and `ExitCode` is 0, then assume the user stopped it.

bboreham:
We could have a `weaveutil` command that examines the persistent IPAM data to check if there were any other peers in the ring.

rade:
> what is "that problem"? They shouldn't have run `stop` ?

Correct. And I know from my own workflow that I run `weave reset` habitually, often some time after a `weave stop`. Getting a warning in this scenario is just annoying, but I can easily retrain my fingers to type `weave reset --force` to avoid it.

bboreham:
I added the `--force` and specific handling for the running, not-running and not-present cases.

I did not handle the case where the container seems to be running but the DELETE fails, and I did not check to see if some other bogus argument was added after `weave reset`.

weave
```diff
@@ -2177,11 +2177,25 @@ EOF
         stop_plugin
         ;;
     reset)
-        [ $# -eq 0 ] || usage
+        [ $# -lt 2 ] || usage
```
awh: In other places we have checked the flag e.g.
```
[ $# -eq 0 ] || [ $# -eq 1 -a "$1" = "--force" ] || usage
```

awh:
One minor comment + needs rebase

bboreham:
Updated and rebased

Issue #2322, opened 2016-05-27
AWSVPC enhancements


bboreham:
Following on from #2091, some ideas for enhancement:

 - try to donate bigger ranges on `claim`, up to half the space available (to reduce tiny-range fragmentation)
 - the "don't donate more than half" calculation seems to be slightly off - if you have allocated 1 address then you should still give away the bigger half, not drop down to a quarter.
 - multi-transport AWSVPC/fastdp/sleeve
 - copy the MTU from the AWS network interface on the host (or default it to 9001 which seems pretty standard for AWS)
 - support isolation between groups of containers (perhaps via subnets)
 - allow users to have more than 50 hosts (limit from AWS route table)
 - make `weave launch` async (currently it is blocking on `expose`)
 - Make update ops of the route table asynchronous (i.e., move to a separate goroutine).
 - Improve semantics for choosing a peer for donations.

 - Add tests for plugin
 - What happens if $RCIDR is not found
 - Enable --awsvpc in other tests
 - Test re-attach of containers

Smoke-test enhancements:
* do not delete keys
* create a separate VPC for each run
* AWS=1 might be set for GCE machines, add the reset cmd

In progress:
 - docs

Done:
 - Test with >2 hosts
 - Get rid of ethtool tx off for AWSVPC
 - Rename useAWSVPC -> isAWSVPC

bboreham:
Figure out some way to have the last peer remove the last route table entry on `weave reset`

(complication: what if you're not really the last peer but you just think you are?)

brb:
Just to add more context to the "do not delete keys" item: each time we run tests on AWS, we replace the global weavenet_ci key with the newly generated one (due to missing GC functionality in CircleCI). Thus, concurrent builds might fail in obscure ways.

brb:
* CNI / Plugin: Add the default subnet check and keep the TX offloading on

bboreham:
I had an issue where the route table had a left-over entry, and containers wouldn't communicate until I deleted it manually:

Destination|Target|Status|Propagated
----|---|---|---|---
172.31.0.0/16|local|Active|No
0.0.0.0/0|igw-0563c660	|Active|No
10.32.0.0/13|eni-139cc059 / i-d9055d53|Active|No
10.40.0.0/13|eni-cc603c86 / i-26055dac|Active|No
10.40.0.0/14|eni-139cc059 / i-d9055d53|Active|No

```
$ weave report
...
        "Entries": [
            {
[...]

PR #2341, opened 2016-06-07
Add AWSVPC

brb:
Please be aware that all the necessary enhancements are listed in https://github.com/weaveworks/weave/issues/2322.

### DIFF
--- a/ipam/ring/ring_test.go
+++ b/ipam/ring/ring_test.go
@@ -36,8 +36,12 @@ func merge(r1, r2 *Ring) error {
 	return err
 }
 
+func NewRing(start, end address.Address, peer mesh.PeerName) *Ring {
+	return New(start, end, peer, nil)
+}
+
 func TestInvariants(t *testing.T) {
-	ring := New(start, end, peer1name)
+	ring := NewRing(start, end, peer1name)
 
 	// Check ring is sorted
 	ring.Entries = []*entry{{Token: dot245, Peer: peer1name}, {Token: dot10, Peer: peer2name}}
@@ -48,7 +52,7 @@ func TestInvariants(t *testing.T) {
 	require.True(t, ring.checkInvariants() == ErrTokenRepeated, "Expected error")
 
 	// Check tokens are in bounds
-	ring = New(dot10, dot245, peer1name)
+	ring = NewRing(dot10, dot245, peer1name)
 	ring.Entries = []*entry{{Token: start, Peer: peer1name}}
 	require.True(t, ring.checkInvariants() == ErrTokenOutOfRange, "Expected error")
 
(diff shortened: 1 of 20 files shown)


ipam/ring/ring_test.go
```diff
@@ -540,11 +544,11 @@ func TestClaimForPeers(t *testing.T) {
 	// Different end to usual so we get a number of addresses that a)
 	// is smaller than the max number of peers, and b) is divisible by
 	// some number of peers. This maximises coverage of edge cases.
-	end := dot10
+	end := dot8
```
rade: why the change?

bboreham: We only support rings with power of two size now.

prog/weaveutil/attach.go
```diff
@@ -20,25 +17,17 @@ func attach(args []string) error {
...
-		args = args[1:]
-	}
-	if isAWSVPC {
-		withMulticastRoute = false
+	for i := 0; i < len(args); i++ {
+		switch args[i] {
+		case "--no-multicast-route":
+			withMulticastRoute = false
+			args = append(args[:i], args[i+1:]...)
+		case "--keep-tx-on":
```
rade: needs to be added to usage.

weave
```diff
@@ -894,6 +893,8 @@ container_in_host_ns() {
 attach() {
     ATTACH_ARGS=""
     [ -n "$NO_MULTICAST_ROUTE" ] && ATTACH_ARGS="--no-multicast-route"
+    # Relying on AWSVPC being set in 'ipam_cidrs allocate', except for 'weave restart'
+    [ -n "$AWSVPC" ] && ATTACH_ARGS="--no-multicast-route --keep-tx=on"
```
brb: you probably meant `--keep-tx-on`

weave
```diff
@@ -1285,6 +1286,13 @@ ipam_cidrs() {
         allocate)
             METHOD=POST
             CHECK_ALIVE="?check-alive=true"
+            # If we get an error here, we skip applying necessary steps for AWSVPC, because
+            # "attach" should work without the weave router running.
+            [ "$(call_weave GET /ipinfo/tracker)" != "awsvpc" ] || AWSVPC=1
```
brb: maybe move to a separate function, because there exist two instances of the same check and set.

weave
bboreham: good spot! fixed.

weave
bboreham: done.

prog/weaveutil/attach.go
bboreham: `--keep-tx-on` added to usage

Issue #2351, opened 2016-06-09
weave run interprets container args


bboreham:
Arguments after the image name are meant for the container; they should not be interpreted by `weave run`.
To illustrate:

```
weave run -ti alpine /bin/sh --name=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Container name too long to be used as hostname
45f745308a532a1b0c3c351741c64a64318860a107dc0e049789fe34fcfe215d
```

Unfortunately it is hard to know where the image name appears; it depends on exactly which `docker run` flags take a following parameter.

rade:
I think we were aware of that.

PR #2352, opened 2016-06-09
Print usage on --help argument to any command

### LINKED ISSUE #2318: All `weave` commands should accept `--help`
Encountered by one of our users.  E.g.

    $ weave launch --help

bboreham:
Fixes #2318. 

~~Note that `weave run` with `--help` as part of the container command-line will produce odd results, because of #2351.~~ This no longer applies.

### DIFF
--- a/weave
+++ b/weave
@@ -108,6 +108,13 @@ usage() {
     exit 1
 }
 
+handle_help_arg() {
+    if [ "$1" = "--help" ] ; then
+        usage_no_exit
+        exit 0
+    fi
+}
+
 docker_sock_options() {
     # Pass through DOCKER_HOST if it is a Unix socket;
     # a TCP socket may be secured by TLS, in which case we can't use it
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -1863,6 +1871,8 @@ deprecation_warnings() {
 COMMAND=$1
 shift 1
 
+[ "$COMMAND" = "run" ] || handle_help_arg "$@"
```
rade: why not simply check $1? Looking for 'help' beyond that seems OTT.

bboreham: just a general principle that we allow arguments to come in any order.

(although this isn't true, for instance, for `--without-dns`)

rade: > just a general principle that we allow arguments to come in any order.

right, but nobody is going to `weave run --no-multicast-route --help ...`

> although this isn't true

That too :)

bboreham:
I rewrote it to just check the first argument

PR #2361, opened 2016-06-13
Validate CIDRs in proxy before attaching a container

### LINKED ISSUE #2283: invalid WEAVE_CIDR results in garbled error message
```
$ weave launch
$ docker $(weave config) run --rm -e WEAVE_CIDR=invalid -ti alpine true
docker: Error response from daemon: Usage:

                                          weave --help | help
                                                                   setup
                                                                              version

     weave launch        <same arguments as 'weave launch-router'>
                                                                        launch-router [--password <pass>] [--name <mac>] [--nickname <nickname>]
                                                                                      [--log-level=debug|info|warning|error]
                                                                  [--no-restart] [--ipalloc-init <mode>]
                                              [--ipalloc-range <cidr> [--ipalloc-default-subnet <cidr>]]
                                              [--no-discovery] [--no-dns]
                                                                                               [--trusted-subnets <cidr>,...] <peer> ...
                                                              launch-proxy  [-H <endpoint>] [--without-dns] [--no-multicast-route]
                                                                        [--log-level=debug|info|warning|error]
                                                    [--no-rewrite-hosts] [--no-default-ipalloc] [--no-restart]
                                                    [--hostname-from-label <labelkey>]
                            [--hostname-match <regexp>]
                                                                             [--hostname-replacement <replacement>]
                                                         [--rewrite-inspect]
                                                                                  launch-plugin [--no-restart] [--no-multicast-route]
                                                                           [--log-level=debug|info|warning|error]

                                 weave prime

                                            weave env           [--restore]
                                                                                 config
             dns-args

                     weave connect       [--replace] [<peer> ...]
                                                                       forget        <peer> ...

               weave run           [--without-dns] [--no-rewrite-hosts] [--no-multicast-route]
                                    [<addr> ...] <docker run args> ...
                                                                            start         [<addr> ...] <container_id>
                                           attach        [<addr> ...] <container_id>
          detach        [<addr> ...] <container_id>
                                                         restart       <container_id>

     weave expose        [<addr> ...] [-h <fqdn>]
                                                       hide          [<addr> ...]

 weave dns-add       [<ip_address> ...] <container_id> [-h <fqdn>] |
                                                                                        <ip_address> ... -h <fqdn>
                                        dns-remove    [<ip_address> ...] <container_id> [-h <fqdn>] |
                                         <ip_address> ... -h <fqdn>
                                                                         dns-lookup    <unqualified_name>

                         weave status        [targets | connections | peers | dns]
        report        [-f <format>]
                                         ps            [<container_id> ...]

                                                                           weave stop
           stop-router
                            stop-proxy
                                            stop-plugin

                                                       weave reset
                                                                        rmpeer        <peer_id> ...


                   where <peer>     = <ip_address_or_fqdn>[:<port>]
                                                                         <cidr>     = <ip_address>/<routing_prefix_length>
                                                <addr>     = [ip:]<cidr> | net:<cidr> | net:default
                         <endpoint> = [tcp://][<ip_address>]:<port> | [unix://]/path/to/socket
                    <peer_id>  = <nickname> | <weave internal peer ID>
                                                                            <mode>     = consensus[=<count>] | seed=<mac>,... | observer.
                                                         Error response from daemon: You cannot remove a running container d392639f97100e25facb14b59bf59439557ecacd2d44726a6259ae639c74dd3e. Stop the container before attempting removal or use -f
   $
```

brb:
Alternatively, we could do the validation inside `proxy.weaveCIDRs`. But in this case, the method has to be changed to return a third parameter which would indicate whether CIDRs are valid and the returned value would be ignored in some cases to keep a container alive.

Fixes #2283

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -20,6 +20,7 @@ import (
 	"github.com/weaveworks/weave/common"
 	weavedocker "github.com/weaveworks/weave/common/docker"
 	weavenet "github.com/weaveworks/weave/net"
+	"github.com/weaveworks/weave/net/address"
 )
 
 const (
@@ -505,6 +506,10 @@ func (proxy *Proxy) attach(containerID string) error {
 		return nil
 	}
 	Log.Infof("Attaching container %s with WEAVE_CIDR \"%s\" to weave network", container.ID, strings.Join(cidrs, " "))
+	if err := validateCIDRs(cidrs); err != nil {
+		return err
+	}
+
 	args := []string{"attach"}
 	args = append(args, cidrs...)
 	if !proxy.NoRewriteHosts {
@@ -529,6 +534,23 @@ func callWeaveAttach(container *docker.Container, args []string) error {
 		return errors.New(string(stderr))
 	} else if len(stderr) > 0 {
 		Log.Warningf("Attaching container %s to weave network: %s", container.ID, string(stderr))
+	}
+	return nil
+}
+
+func validateCIDRs(cidrs []string) error {
+	for _, cidr := range cidrs {
+		if cidr == "net:default" {
+			continue
+		}
+		for _, prefix := range []string{"ip:", "net:", ""} {
+			if strings.HasPrefix(cidr, prefix) {
+				if _, err := address.ParseCIDR(strings.TrimPrefix(cidr, prefix)); err == nil {
+					break
+				}
+				return fmt.Errorf("invalid CIDR: %s", cidr)
+			}
+		}
 	}
 	return nil
 }


proxy/proxy.go
```diff
@@ -533,6 +539,26 @@ func callWeaveAttach(container *docker.Container, args []string) error {
...
+				if _, err := address.ParseCIDR(strings.TrimPrefix(cidr, prefix)); err == nil {
+					return true
+				}
+			}
+		}
+		return false
+	}
+
+	for _, cidr := range cidrs {
+		if cidr != "net:default" && !check(cidr, []string{"ip:", "net:", ""}) {
```
rade: Why special-case "net:default"? What you are checking here has nothing to do with docker's `--net` option.

proxy/proxy.go
```diff
@@ -533,6 +539,26 @@ func callWeaveAttach(container *docker.Container, args []string) error {
...
+	check := func(cidr string, prefixes []string) bool {
+		for _, prefix := range prefixes {
+			if strings.HasPrefix(cidr, prefix) {
+				if _, err := address.ParseCIDR(strings.TrimPrefix(cidr, prefix)); err == nil {
+					return true
+				}
+			}
+		}
+		return false
+	}
```
rade: unnecessary closure. Depending on the resolution of my query below, either inline or lift it.

rade:
This should be on the 1.5 branch.

proxy/proxy.go
brb: According to https://github.com/weaveworks/weave/blob/bf8c8833e5f07ce5911f40e4e1d13f936dd9b508/weave#L275 , "net:default" is a valid argument.

brb:
I will reopen after comments get addressed.

proxy/proxy.go
rade: ah yes, I forget we had that.

proxy/proxy.go
brb: Why do you think that this closure has to be lifted up into a separate function? The closure is used only by this function and it does not make sense to make it accessible by other functions.

rade: It would be clearer when lifted. Less indentation and it would be obvious that it doesn't close over anything.

Anyway, you could easily inline it. Which would be better.

proxy/proxy.go
```diff
@@ -533,6 +539,29 @@ func callWeaveAttach(container *docker.Container, args []string) error {
 	return nil
 }
 
+func validateCIDRs(cidrs []string) error {
+	prefixes := []string{"ip:", "net:", ""}
```
rade: I'd inline that.

proxy/proxy.go
```diff
@@ -533,6 +539,29 @@ func callWeaveAttach(container *docker.Container, args []string) error {
...
+		for _, prefix := range prefixes {
+			if strings.HasPrefix(cidr, prefix) {
+				if _, err := address.ParseCIDR(strings.TrimPrefix(cidr, prefix)); err == nil {
+					ok = true
+				}
+				break
+			}
+		}
+		if !ok {
+			return fmt.Errorf("invalid CIDR: %s", cidr)
```
rade: Because of the "" prefix the 'if' inside the loop will succeed at least once. So why not simply return error/nil there? Then we don't need all that 'ok' logic.

PR #2363, opened 2016-06-13
Add a switch for running AWSVPC tests

brb:
Adding `CLOUD_ENV: aws` to `machine.environment` in circle.yml starts tests on AWS. In addition, it enables the AWSVPC tests.


bboreham:
How would someone make use of this?  Create a branch and add that `machine.environment` line?

brb:
Yep, or use [parametrized builds](https://circleci.com/docs/nightly-builds/). Unfortunately, the latter is supported only via REST API.

brb:
Unfortunately, the easy switch does not work, because for AWS, `parallel` has to be set to `false` (otherwise, a concurrent builder overrides `weavenetci-keys`) and it seems that in CircleCI it is not possible to control the variable thus it is hardcoded.

To address the issue, we could reuse the same instance of `weavenetci-keys` as it is done for GCE.

bboreham:
`aws.sh` is roughly parallel to `gce.sh`; should the common parts be hoisted out to another file?  E.g. the function `vm_names` appears to be the same in each.

(There is a TODO comment speaking about DRY, but the code in that area does seem to be different between aws and gce.  At minimum the TODO should be removed)

brb:
re moving the common parts: I'd suggest to keep it as it is for a moment while https://github.com/weaveworks/weave/pull/1895 is open.

brb:
This can be closed after merging https://github.com/weaveworks/weave/pull/2694 and cherry-picking `test/900_awsvpc_3_test.sh`.

brb:
Closing this PR, as the AWS provisioning part is no longer required. The 900 integration test can be cherry-picked in the future if someone decides to run the integration tests on AWS.

PR #2366, opened 2016-06-13
Publish weavedb container from release script

### LINKED ISSUE #2189: Update release script to publish weavedb container
It has been manually pushed and probably won't change, but it will trip up anyone who is testing the release script against a different dockerhub account in future.

bboreham:
Fixes #2189 

Note that the weavedb image will not be pushed by the CI build, to address the point at https://github.com/weaveworks/weave/issues/2189#issuecomment-215380936

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -165,14 +165,14 @@ publish() {
     setup
     cd $PWD/$RELEASE_DIR
 
-    UPDATE_LATEST=false
+    IS_MAINLINE=false
     if [ "$RELEASE_TYPE" = 'MAINLINE' ] ; then
-        UPDATE_LATEST=true
+        IS_MAINLINE=true
     fi
 
     if [ "$RELEASE_TYPE" = 'PRERELEASE' ] ; then
         echo "== Tagging and pushing images on docker hub as user $DOCKERHUB_USER"
-        make UPDATE_LATEST=$UPDATE_LATEST SUDO=$SUDO WEAVE_VERSION=$VERSION DOCKERHUB_USER=$DOCKERHUB_USER publish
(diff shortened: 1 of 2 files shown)


bin/release
```diff
@@ -172,7 +172,7 @@ publish() {
 
     if [ "$RELEASE_TYPE" = 'PRERELEASE' ] ; then
         echo "== Tagging and pushing images on docker hub as user $DOCKERHUB_USER"
-        make UPDATE_LATEST=$UPDATE_LATEST SUDO=$SUDO WEAVE_VERSION=$VERSION DOCKERHUB_USER=$DOCKERHUB_USER publish
+        make PUBLISH_WEAVEDB=$UPDATE_LATEST UPDATE_LATEST=$UPDATE_LATEST SUDO=$SUDO WEAVE_VERSION=$VERSION DOCKERHUB_USER=$DOCKERHUB_USER publish
```
awh: This doesn't do anything, because `UPDATE_LATEST` is only true for mainline releases. You need to pass this into the `make ... publish` in the else clause of the surrounding if statement as well.

bboreham: Only doing something on mainline releases was intentional.  I want to minimize the number of times `weave setup` sees a new version of this image to download.

awh: You are not passing a value for `PUBLISH_WEAVEDB` into the `make ... publish` that publishes mainline releases - the comparison to `true` in the makefile will fail

awh:
Also, not sure that `PUBLISH_WEAVEDB=$UPDATE_LATEST` makes a huge amount of sense. Perhaps we should rename the `UPDATE_LATEST` variable at the top of publish() to `IS_MAINLINE` now

bin/release
bboreham: right, I get it now.  have re-done and overwritten.

PR #2368, opened 2016-06-13
Small doc fixes

bboreham:
Fix a couple of things I noticed in passing:
- implying that users should read and type in the enormous hex ID of a container
- telling users to look "above" when we're at the top of a file

ping @abuehrle for doc review.


abuehrle:
LGTM

PR #2369, opened 2016-06-13
Describe creating multiple networks using the plugin

### LINKED ISSUE #2335: plugin docs should explain how to create multiple networks
https://www.weave.works/docs/net/latest/plugin/ makes no mention of how to create multiple networks. That info is hidden away in https://www.weave.works/docs/net/latest/plugin/plugin-how-it-works/, and not especially clear.

NB: This needs to point out that creating multiple networks *does* require a cluster store.

Related, https://www.weave.works/docs/net/latest/plugin/#cluster-store makes no sense. Bypassing the cluster store is something that happens for the default network; it's not something a user has to configure, or indeed *can* configure as outlined in that section (which seems to be aimed at network plugin *authors* not users).

bboreham:
Fixes #2335 

Ping @abuehrle

### DIFF
--- a/site/plugin.md
+++ b/site/plugin.md
@@ -56,6 +56,30 @@ then run your containers using the Docker command-line:
     64 bytes from 10.32.0.2: icmp_seq=2 ttl=64 time=0.052 ms
 
 
+###<a name="multi"></a>Creating multiple Docker Networks
+
+Docker lets you create multiple independent networks and attach
+different sets of containers to each network. However, to coordinate
+this between hosts Docker requires that you configure a
+["key-value store](https://docs.docker.com/engine/userguide/networking/get-started-overlay/#step-1-set-up-a-key-value-store).
+
+If your Docker installation has a key-value store, you can create a
+network based on Weave Net like this:
+
+    $ docker network create --driver=weave mynetwork
+
+then use it to connect a container:
+
+    $ docker run --net=mynetwork ...
+
+or
+
+    $ docker network connect mynetwork somecontainer
+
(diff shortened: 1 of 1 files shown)


site/plugin.md
```diff
@@ -56,6 +56,30 @@ then run your containers using the Docker command-line:
     64 bytes from 10.32.0.2: icmp_seq=2 ttl=64 time=0.052 ms
 
 
+###<a name="multi"></a>Creating multiple Docker Networks
+
+Docker lets you create multiple independent networks and attach
+different sets of containers to each network. However, to coordinate
```
abuehrle: --minor tweaks--
Docker enables you to create multiple independent networks and attach different sets of containers to each network.  However, coordinating this between hosts, Docker requires that configure a "key-value store"  <-- missing a quote on the end and also tweaked that sentence.

site/plugin.md
```diff
@@ -56,6 +56,30 @@ then run your containers using the Docker command-line:
...
+this between hosts Docker requires that you configure a
+["key-value store](https://docs.docker.com/engine/userguide/networking/get-started-overlay/#step-1-set-up-a-key-value-store).
+
+If your Docker installation has a key-value store, you can create a
+network based on Weave Net like this:
+
```
abuehrle: If your Docker installation has a key-value store, create a network based on Weave Net as follows:

site/plugin.md
bboreham: Assuming "Docker requires that configure" should be "Docker requires that you configure"

PR #2372, opened 2016-06-14
Add kernel version, Docker version and awsvpc setting to checkpoint info

### LINKED ISSUE #2310: include docker and kernel version in checkpoint
This would help us make decision about version compatibility.

bboreham:
Fixes #2310 

I couldn't find any library anywhere that did the necessary kernel wrangling, until after I wrote my own, when I found https://github.com/docker/docker/blob/master/pkg/platform/.  I don't like the way they report errors unconditionally via `logrus` so didn't use it.

(No errors are documented for `uname(2)` except "bad pointer" which can't happen here)

### DIFF
--- a/site/installing-weave.md
+++ b/site/installing-weave.md
@@ -31,11 +31,19 @@ With Weave Net downloaded onto your VMs or hosts, you are ready to launch a Weav
 Weave Net [periodically contacts Weaveworks servers for available
 versions](https://github.com/weaveworks/go-checkpoint).  New versions
 are announced in the log and in [the status
-summary](/site/troubleshooting.md#weave-status).  To disable this
-check, run the following before launching Weave Net:
+summary](/site/troubleshooting.md#weave-status).
+
+The information sent in this check is:
+
+ * Host UUID hash
+ * Kernel version
+ * Docker version
+ * Weave Net version
+ * Network mode, e.g. 'awsvpc'
+
+To disable this check, run the following before launching Weave Net:
 
     export CHECKPOINT_DISABLE=1
-
 
 ###Guides for Specific Platforms
 
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -84,6 +84,15 @@ func (c *Client) Info() string {
 	return fmt.Sprintf("Docker API on %s: %v", c.Endpoint(), env)
 }
 
+func (c *Client) DockerVersion() string {
+	if env, err := c.Version(); err == nil {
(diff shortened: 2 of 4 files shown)


awh:
I propose we add a new entry at the very top of `site/operational-guide/tasks.md` named 'Disable checkpoint' which describes the data we collect, and how to switch it off.

rade:
Ops guide seems a strange place. I would have though this info belongs in the "Checkpoint" section of https://www.weave.works/docs/net/latest/installing-weave/

awh:
It can go there too

rade:
It shouldn't go in two places. And "Installing Weave" is the better of the two, imo.

awh:
LGTM; just needs docs

bboreham:
Added a section to the docs re checkpoint info

site/installing-weave.md
```diff
@@ -36,6 +36,12 @@ check, run the following before launching Weave Net:
 
     export CHECKPOINT_DISABLE=1
 
+The information sent in this check is:
```
brb: @abuehrle This requires your attention.

site/installing-weave.md
```diff
@@ -36,6 +36,12 @@ check, run the following before launching Weave Net:
...
+
+ * Weave current version
```
rade: "Weave Net" (and I'd drop the 'current')

site/installing-weave.md
```diff
@@ -36,6 +36,12 @@ check, run the following before launching Weave Net:
...
+ * Kernel version
+ * Docker version
+ * AWSVPC mode
```
rade: "Network mode, e.g. 'awsvpc'"

bboreham: Note we don't send any other modes at present.

rade: I know.

site/installing-weave.md
rade: Surely we should also mention the host UUID.

abuehrle:
LGTM

site/installing-weave.md
bboreham: We don't send that. We send a hash, which I didn't unpick.

rade: "host UUID hash"? "anonymized host UUID"

Basically I think it's important to communicate that we can tie the information to a single, albeit anonymised, host.

rade:
I have addressed the remaining points.

PR #2373, opened 2016-06-14
Fallback to 1500 MTU if it is not possible to set it to >1500 on vxlan-vport

### LINKED ISSUE #1853: Fast datapath MTU limited to 1500 on CoreOS 877 (Linux 4.3)
Symptoms are that if you `weave launch-router` with WEAVE_MTU set to anything over 1500 (on a network which supports it), the connection falls back to sleeve.

tcpdump shows the first heartbeat packet (which is sized to the MTU) doesn't get as far as the network device.

Although CoreOS 877 is a beta release, this is the first version with Docker 1.9, needed for Docker Network plugin support.

brb:
This check started to look a bit as an overkill to me, mainly because we need to specify exact port for vxlan-vport.

Fixes #1853

### DIFF
--- a/prog/weaveutil/datapath.go
+++ b/prog/weaveutil/datapath.go
@@ -4,15 +4,24 @@ package main
 import (
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/weaveworks/weave/common/odp"
+	wnet "github.com/weaveworks/weave/net"
 )
 
 func createDatapath(args []string) error {
-	if len(args) != 1 {
-		cmdUsage("create-datapath", "<datapath>")
+	if len(args) != 2 {
+		cmdUsage("create-datapath", "<datapath> <mtu>")
 	}
-	odpSupported, err := odp.CreateDatapath(args[0])
+
+	dpName := args[0]
+	mtu, err := strconv.Atoi(args[1])
+	if err != nil {
+		return fmt.Errorf("unable to parse mtu %q: %s", args[1], err)
+	}
+
+	odpSupported, err := odp.CreateDatapath(dpName)
 	if !odpSupported {
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)
@@ -21,7 +30,15 @@ func createDatapath(args []string) error {
 		// status to distinguish it for the weave script.
 		os.Exit(17)
 	}
-	return err
+	if err != nil {
+		return err
+	}
+
+	if err := wnet.SetMTU(dpName, mtu); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func deleteDatapath(args []string) error {
--- a/prog/weaveutil/datapath.go
+++ b/prog/weaveutil/datapath.go
@@ -3,9 +3,12 @@ package main
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"strconv"
+	"syscall"
 
+	libodp "github.com/weaveworks/go-odp/odp"
 	"github.com/weaveworks/weave/common/odp"
 	wnet "github.com/weaveworks/weave/net"
 )
@@ -15,13 +18,13 @@ func createDatapath(args []string) error {
 		cmdUsage("create-datapath", "<datapath> <mtu>")
 	}
 
-	dpName := args[0]
+	dpname := args[0]
 	mtu, err := strconv.Atoi(args[1])
 	if err != nil {
 		return fmt.Errorf("unable to parse mtu %q: %s", args[1], err)
 	}
 
-	odpSupported, err := odp.CreateDatapath(dpName)
+	odpSupported, err := odp.CreateDatapath(dpname)
 	if !odpSupported {
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)
@@ -34,7 +37,16 @@ func createDatapath(args []string) error {
 		return err
 	}
 
-	if err := wnet.SetMTU(dpName, mtu); err != nil {
+	if mtu > 1500 {
+		// Check whether a host is exposed to https://github.com/weaveworks/weave/issues/1853
+		// If yes, fallback to 1500 MTU.
+		if !checkMTU(dpname, mtu) {
+			fmt.Fprintf(os.Stderr, "WARNING: Unable to set fastdp MTU to %d, possibly due to 4.3/4.4 kernel version. Setting MTU to 1500 instead.\n", mtu)
+			mtu = 1500
+		}
+	}
+
+	if err := wnet.SetMTU(dpname, mtu); err != nil {
 		return err
 	}
 
@@ -54,8 +66,14 @@ func checkDatapath(args []string) error {
 		cmdUsage("check-datapath", "<datapath>")
 	}
 
-	if err := createDatapath(args); err != nil {
+	dpname := args[0]
+
+	odpSupported, err := odp.CreateDatapath(dpname)
+	if err != nil {
 		return err
+	}
+	if !odpSupported {
+		return fmt.Errorf("kernel does not have ODP support")
 	}
 
 	return odp.DeleteDatapath(args[0])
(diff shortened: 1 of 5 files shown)


prog/weaveutil/datapath.go
```diff
@@ -21,7 +41,48 @@ func createDatapath(args []string) error {
...
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+// Really dummy and racy way to get an ephemeral port
+func getUDPPortNo() uint16 {
```
brb: This needs to be changed. Either by reusing the same UDP port used by weave or by choosing randomly and re-trying.

prog/weaveutil/datapath.go
```diff
@@ -3,16 +3,36 @@ package main
...
-	if len(args) != 1 {
-		cmdUsage("create-datapath", "<datapath>")
+	var (
+		mtu    int
+		err    error
+		setMTU bool
+	)
+
+	if len(args) != 1 && len(args) != 3 {
+		cmdUsage("create-datapath", "[--mtu <mtu>] <datapath>")
```
rade: no point making this optional. "create-datapath <datapath> <mtu>" suffices.

prog/weaveutil/datapath.go
```diff
@@ -21,7 +41,48 @@ func createDatapath(args []string) error {
...
+		return err
+	}
+
+	if setMTU {
+		if err := wnet.SetMTU(dpName, mtu); err != nil {
+			return fmt.Errorf("unable to set MTU to %d on %q interface: %s", mtu, dpName, err)
+		}
+		// Check whether we are able to set the given MTU on a vxlan vport in order
+		// to warn users exposed to https://github.com/weaveworks/weave/issues/1853
+		if mtu > 1500 {
```
rade: why bother with that check?

prog/weaveutil/datapath.go
brb: `create-datapath` is used in two places in the weave script. Here https://github.com/weaveworks/weave/blob/7c51c313579d11c92d1840c365a4dddf7aad6e15/weave#L628 you probably don't want to mess with setting mtu and doing the check.

prog/weaveutil/datapath.go
brb: The problem exists only with > 1500 MTU, so to avoid the unnecessary check.

rade: overkill

prog/weaveutil/datapath.go
rade: TRTTD there is to introduce a separate command (`check-datapath`?) for the 2nd case.

prog/weaveutil/datapath.go
```diff
@@ -21,7 +41,48 @@ func createDatapath(args []string) error {
...
+			if vpid, vpname, err := odp.CreateVxlanVport(dpName, "vxlan", getUDPPortNo()); err == nil {
+				defer func() {
+					if err := odp.DeleteVport(dpName, vpid); err != nil {
+						fmt.Fprintf(os.Stderr, "unable to remove unused %q vxlan vport: %s\n", vpname, err)
+					}
+				}()
+				if err := wnet.SetMTU(vpname, mtu); err != nil {
+					if errno, ok := err.(syscall.Errno); ok && errno == syscall.EINVAL {
+						fmt.Fprintf(os.Stderr, "WARNING: fastdp mode is not able to work with %d MTU potentially due to 4.3 or 4.4 kernel version\n", mtu)
```
rade: As I mentioned on slack, this really should tell the user what MTU *is* being used. Otherwise the warning might be misinterpreted as indicating that fastdp isn't working *at all*.

prog/weaveutil/datapath.go
rade: btw, I'd suggest doing that as a separate commit at the beginning, so we could potentially cherry-pick it.

prog/weaveutil/datapath.go
```diff
@@ -31,9 +54,81 @@ func deleteDatapath(args []string) error {
...
+	}
+	// Couldn't create the vport, skip the check
+	if vpname == "" {
+		return
+	}
+
+	// Setting >1500 MTU on affected host' vport should fail with EINVAL
+	if err := wnet.SetMTU(vpname, mtu); err != nil {
+		if errno, ok := err.(syscall.Errno); ok && errno == syscall.EINVAL {
+			fmt.Fprintf(os.Stderr, "WARNING: fastdp won't work with %d MTU potentially due to 4.3 or 4.4 kernel version\n", mtu)
```
brb: @bboreham is this warning clear enough?

brb:
I'd suggest to add a note about the problem to https://www.weave.works/docs/net/latest/features/#fast-data-path

prog/weaveutil/datapath.go
rade: No, it's not. See https://github.com/weaveworks/weave/pull/2373#discussion_r66999339

Also, I thought we said we'd make this an *error*?

rade: > No, it's not

Ah, that was based on my misreading that fdp was still working but using a lower MTU. The warning is not at all clear about the fact that FDP is in fact being disabled.

Why not fall back to using a lower MTU?

brb: > The warning is not at all clear about the fact that FDP is in fact being disabled.

We do not disable it. 

> Why not fall back to using a lower MTU?

Probably it's a good idea.

rade:
also, 150 lines of code just to print a warning for a condition that, to the best of our knowledge, has puzzled a single user, seems OTT.

brb:
I expect to see more puzzled users once 4.3 / 4.4 is widespread.

rade:
very, very few users set a WEAVE_MTU.

prog/weaveutil/datapath.go
```diff
@@ -31,9 +60,83 @@ func deleteDatapath(args []string) error {
...
+				if err := odp.DeleteVport(dpname, vpid); err != nil {
+					fmt.Fprintf(os.Stderr, "unable to remove unused %q vxlan vport: %s\n", vpname, err)
+				}
+			}()
+			break
+		} else {
+			if errno, ok := err.(syscall.Errno); ok && errno == syscall.EADDRINUSE {
+				continue
+			}
+		}
```
rade: The 'else' branch contributes nothing to the logic.

prog/weaveutil/datapath.go
```diff
@@ -21,7 +33,24 @@ func createDatapath(args []string) error {
...
-	return err
+	if err != nil {
+		return err
+	}
+
+	if mtu > 1500 {
+		// Check whether a host is exposed to https://github.com/weaveworks/weave/issues/1853
+		// If yes, fallback to 1500 MTU.
+		if !checkMTU(dpname, mtu) {
+			fmt.Fprintf(os.Stderr, "Unable to set MTU to %d potentially due to 4.3 or 4.4 kernel version; Setting MTU to 1500\n", mtu)
```
rade: "WARNING: Unable to set MTU to %d, possibly due to 4.3/4.4 kernel version. Setting MTU to 1500 instead.\n"

(this will still look odd if it pops up when the user is actually running a different kernel)

prog/weaveutil/datapath.go
```diff
@@ -21,7 +33,24 @@ func createDatapath(args []string) error {
...
-	return err
+	if err != nil {
+		return err
+	}
+
+	if mtu > 1500 {
+		// Check whether a host is exposed to https://github.com/weaveworks/weave/issues/1853
+		// If yes, fallback to 1500 MTU.
+		if !checkMTU(dpname, mtu) {
+			fmt.Fprintf(os.Stderr, "fastdp: Unable to set MTU to %d, potentially due to 4.3 or 4.4 kernel version; Setting MTU to 1500\n", mtu)
```
rade: "WARNING: Unable to set fastdp MTU to %d, possibly due to 4.3/4.4 kernel version. Setting MTU to 1500 instead.\n"

rade:
I have cherry-picked the first commit onto master. The rest is too dangerous to merge into 1.6.0 so close to a release. I suspect it's also too dangerous for a patch release, but for the moment let's assume it's fine -> moving to 1.6.1

brb:
OK, thanks.

I will reopen the PR once we have created the 1.6 branch.

brb:
Re-opening against 1.6

Issue #1914, opened 2016-01-20
Use Docker network events in the plugin


bboreham:
Currently, `weaveplugin` listens for container started events, to know when to add a container name to DNS.  We do some poking around in the API to see whether the container is actually connected to the weave network.

This approach is rather indirect, and misses `docker network connect` / `disconnect`.

In 1.10, Docker introduces [network events](https://github.com/docker/docker/pull/18888), which give a much more direct way to achieve the requirement.

Given #813, we may be justified in requiring minimum Docker 1.10 when using the plugin.

bboreham:
Also, as noted on IRC, users can use `--alias` to give additional names in `docker network connect`

bboreham:
Post #2347 we should also do the arp configuration on `docker network connect`

dlmiddlecote:
I'm going to work on this.

bboreham:
Thanks @dlmiddlecote - if you have any questions please ask here or in the [Slack](https://weaveworks.github.io/community-slack/)

PR #2374, opened 2016-06-14
Perform arp configuration for plugin-connected containers

### LINKED ISSUE #1726: Make veth settings closer to those produced by the weave script

bboreham:
Fixes #1726, the last remaining part.

Note that we are only reacting to container `start` events, so not doing the arp stuff when a user `docker network connect`s a container.  See #1914

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 
@@ -34,6 +35,7 @@ type NetDev struct {
 }
 
 // Search the network namespace of a process for interfaces matching a predicate
+// Note that the predicate is called while the goroutine is inside the process' netns
 func FindNetDevs(processID int, match func(link netlink.Link) bool) ([]NetDev, error) {
 	var netDevs []NetDev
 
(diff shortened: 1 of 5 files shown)


bboreham:
There is the usual race arising from Docker sending the `start` event before actually starting the process.  But we have to receive the event and complete the DNS-add http call before the interface is in place to lose the race.

weave
```diff
@@ -1747,6 +1747,7 @@ launch_plugin_if_not_running() {
     if ! PLUGIN_CONTAINER=$(docker run -d --name=$PLUGIN_CONTAINER_NAME \
         $(docker_run_options) \
         $RESTART_POLICY \
+        --pid=host \
```
brb: Short comment either here or in the commit message would be useful for why we start container in host PID namespace.

bboreham: If we don't do this we cannot navigate to the container's namespace given its PID.  At least I don't know how to do it.

common/utils.go
```diff
@@ -137,3 +139,20 @@ func GetBridgeNetDev(bridgeName string) ([]NetDev, error) {
 		return link.Attrs().Name == bridgeName
 	})
 }
+
+// Do post-attach configuration of all veths we have created
+func ConfigureARPforVeths(processID int, prefix string) error {
+	_, err := FindNetDevs(processID, func(link netlink.Link) bool {
+		ifName := link.Attrs().Name
+		if strings.HasPrefix(ifName, prefix) {
```
awh: We rewrote `GetWeaveNetDevs` to find the veths that were actually attached to the weave bridge, rather than using this comparison - should we not be reusing that approach here?

bboreham: I was thinking we only want plugin-generated interfaces rather than all weave interfaces.  In context that probably doesn't matter. 
The `GetWeaveNetDevs` code would need a bit of refactoring to be re-used here.

PR #2378, opened 2016-06-16
Observe resolv.conf changes

### LINKED ISSUE #2364: Weave DNS should observe changes to resolv.conf
The router currently reads `/etc/resolv.conf` once on startup, which causes problems with environments that manage the contents dynamically (e.g. see #2360). We should instead listen or poll for changes and update accordingly.

awh:
Fixes #2364.

### DIFF
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"os"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/miekg/dns"
@@ -17,7 +19,6 @@ import (
 const (
 	topDomain        = "."
 	reverseDNSdomain = "in-addr.arpa."
-	etcResolvConf    = "/etc/resolv.conf"
 	udpBuffSize      = uint16(4096)
 	minUDPSize       = 512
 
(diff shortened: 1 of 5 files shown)


weave
```diff
@@ -1625,13 +1628,16 @@ launch_router() {
            --label=weavevolumes $WEAVEDB_IMAGE >/dev/null
     fi
 
+    RESOLV_CONF_DIR=$(dirname $(chroot $HOST_ROOT readlink -f /etc/resolv.conf))
```
awh: Really I should take the `basename` of this too and pass it into the router - right now we assume that the ultimate file is still called `resolv.conf`. Thoughts?

nameserver/dns.go
```diff
@@ -51,22 +57,19 @@ func filter(ss []string, s string) []string {
...
+		ttl:              ttl,
+		address:          address,
+		effectiveAddress: effectiveAddress,
+		upstream:         &dns.ClientConfig{},
+		tcpClient:        &dns.Client{Net: "tcp", ReadTimeout: clientTimeout},
+		udpClient:        &dns.Client{Net: "udp", ReadTimeout: clientTimeout, UDPSize: udpBuffSize},
 	}
 
-	err = s.listen(address)
+	s.readUpstream()
```
awh: I make an attempt here so that `weave report` stands a chance of containing useful upstream infos before any recursive lookups have occurred.

nameserver/dns.go
```diff
@@ -27,15 +29,19 @@ const (
...
+	domain           string
+	ttl              uint32
+	address          string
+	effectiveAddress string
+
+	servers          []*dns.Server
+	upstream         *dns.ClientConfig
+	upstreamMutex    sync.Mutex
+	upstreamChecked  time.Time
+	upstreamModified time.Time
```
rade: I reckon encapsulating these four plus the effectiveAddress in a separate struct would improve the code.

nameserver/dns.go
```diff
@@ -96,6 +99,40 @@ func (d *DNSServer) listen(address string) error {
...
+				d.ns.debugf("Read error: %v; returning existing cached value %v", err, d.upstream)
+			}
+		} else {
+			d.ns.debugf("Unmodified; returning existing cached value %v", d.upstream)
+		}
+	} else {
+		d.ns.debugf("Stat error: %v; returning existing cached value %v", err, d.upstream)
+	}
+	return d.upstream
+}
```
rade: All these debug else branches makes this rather hard to read. Do we really need them?

Without them we would have
```
fi, err := os.Stat(etcResolvConf)
if err != nil || fi.ModTime() == d.upstreamModified {
    return d.upstream
}
if upstream, err := dns.ClientConfigFromFile(etcResolvConf); err == nil {
    upstream.Servers = filter(upstream.Servers, d.effectiveAddress)
    d.upstream = upstream
    d.upstreamModified = fi.ModTime()
}
return d.upstream
```

awh: Indeed - I wrote it that way initially, and then I had to figure out why it wasn't working :smile: Do you really want them removed? They'll be helpful if this breaks again in future...

rade: remove

weave
```diff
@@ -168,6 +168,8 @@ exec_remote() {
         $(docker_run_options) \
         --pid=host \
         $(cni_volume_options "$@") \
+        -v /:/host \
+        -e HOST_ROOT="/host" \
```
rade: perhaps we should do this only for launch commands. Similar to what `cni_volume_options` does.

nameserver/dns.go
awh: Good idea.

weave
awh: OK

nameserver/dns.go
```diff
@@ -96,6 +103,29 @@ func (d *DNSServer) listen(address string) error {
 	return nil
 }
 
+func (d *DNSServer) upstreamConfig() *dns.ClientConfig {
```
rade: this should be a method on 'upstream'. and renamed.

weave
```diff
@@ -424,6 +429,7 @@ PORT=${WEAVE_PORT:-6783}
 HTTP_ADDR=${WEAVE_HTTP_ADDR:-127.0.0.1:6784}
 PROXY_PORT=12375
 PROXY_CONTAINER_NAME=weaveproxy
+HOST_ROOT=${HOST_ROOT:-/}
```
rade: now that HOST_ROOT isn't passed all the time, this should be moved (and perhaps linlined?) to the router launch code.

nameserver/dns.go
awh: Done.

weave
awh: Done.

weave
awh: Done.

PR #2380, opened 2016-06-16
Docker 1.12: handle daemon named 'dockerd'

### LINKED ISSUE #2379: Docker 1.12 breaks Weave
(as of 1.12.0-rc1)
The daemon now runs as a process called `dockerd`, which breaks `docker-tls-args` which is looking for `docker -d`.

bboreham:
Fixes #2379

### DIFF
--- a/prog/weaveutil/docker_tls_args.go
+++ b/prog/weaveutil/docker_tls_args.go
@@ -26,7 +26,12 @@ func dockerTLSArgs(args []string) error {
 			continue
 		}
 
-		if comm, err := ioutil.ReadFile(filepath.Join(procRoot, dirName, "comm")); err != nil || string(comm) != "docker\n" {
+		isDaemon := false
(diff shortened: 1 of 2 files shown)


rade:
This should be on the 1.5 branch.

awh:
LGTM. Quick aside - this

https://github.com/weaveworks/weave/pull/2380/files#diff-5fb26b2385c2ea3455765e854c85c607R50

`break` statement is redundant isn't it?

Did you want to retarget this at the 1.5 branch?

bboreham:
I imagine that `break` statement was meant to end the loop, but it doesn't.  

However, if we did make it break the loop, I suspect we could mistake a `docker run -d` command for the daemon process.

bboreham:
Replaced by #2382 on 1.5 branch

bboreham:
Changed my mind: the `break` is just superfluous

PR #2383, opened 2016-06-17
Use buffered writer for http headers

bboreham:
As seen in wireshark, we are doing four sends for every entry in the http header, many of which contain just 2 bytes of data (and, if over TCP, 54 bytes of packet header).  This change uses the already-existing buffered writer to condense them all into one send.

I also condensed all the `if err != nil` lines into one.

### DIFF
--- a/proxy/proxy_intercept.go
+++ b/proxy/proxy_intercept.go
@@ -77,24 +77,22 @@ func doRawStream(w http.ResponseWriter, resp *http.Response, client *httputil.Cl
 	}
 	defer down.Close()
 	defer up.Close()
+	defer func() {
+		if err != nil {
+			Log.Warning(err)
+		}
+	}()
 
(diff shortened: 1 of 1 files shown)


proxy/proxy_intercept.go
```diff
@@ -78,17 +78,14 @@ func doRawStream(w http.ResponseWriter, resp *http.Response, client *httputil.Cl
...
-
-	if err := resp.Header.Write(down); err != nil {
-		Log.Warning(err)
-		return
+	if _, err = downBuf.Write([]byte("HTTP/1.1 " + resp.Status + "\n")); err == nil {
+		if err = resp.Header.Write(downBuf); err == nil {
+			if _, err = downBuf.Write([]byte("\n")); err == nil {
+				err = downBuf.Flush()
+			}
+		}
```
rade: This nesting for four steps is horrible.

I suggest flattening it and using a 'defer' to log warnings,

brb:
LGTM.

Issue #2144, opened 2016-04-08
`weave expose` is forgotten when the machine reboots


bboreham:
We can tell from IPAM persistence that the bridge has been given an IP address (under the name `weave:expose`), so we would re-apply the operation in `attach-router`

bboreham:
This is relevant when you need `weave expose`, e.g. to use the CNI plugin; without this feature you need to set up something like a `systemd` unit to run `weave expose` after Weave Net is running.

PR #2384, opened 2016-06-17
Fix re-attaching of weave router in awsvpc mode

### LINKED ISSUE #2381: Attaching the router after its restart does not work in the awsvpc mode
The following should be done in `attach-router`:

* Detect the awsvpc mode
* Call `expose_ip`
* Restore vpc related routes on a host

Also, make sure that TX offloading is not disabled in `create_bridge`.

brb:
In order to re-attach the router and make it work in awsvpc mode, the following has to be done:

* `expose_ip` on the router
* adjusting arp settings
* restoring host routes (routes are taken from VPC rt)

To detect that we run in awsvpc mode, we inspect cmd parameters of the weave router instead of GET'ing /ipinfo/tracker. This is so, because the HTTP server started by the router might not be running at the time of restoration.

Fixes #2381

### DIFF
--- a/weave
+++ b/weave
@@ -1655,13 +1655,12 @@ launch_router() {
         "$@")
     setup_router_iface_$BRIDGE_TYPE
     wait_for_status $CONTAINER_NAME http_call $HTTP_ADDR
-    populate_router
     setup_awsvpc
+    populate_router
 }
 
 setup_awsvpc() {
     if [ -n "$AWSVPC" ]; then
-        expose_ip
         # Set proxy_arp on the bridge, so that it could accept packets destined
         # to containers within the same subnet but running on remote hosts.
         # Without it, exact routes on each container are required.
@@ -1670,6 +1669,7 @@ setup_awsvpc() {
         # placing the request into a bounded queue as it can be seen:
         # https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/net/ipv4/arp.c?id=refs/tags/v4.6.1#n819
         echo 0 >/proc/sys/net/ipv4/neigh/$BRIDGE/proxy_delay
+        expose_ip
     fi
 }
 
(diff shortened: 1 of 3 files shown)


weave
```diff
@@ -1646,6 +1646,10 @@ launch_router() {
     setup_router_iface_$BRIDGE_TYPE
     wait_for_status $CONTAINER_NAME http_call $HTTP_ADDR
     populate_router
+    setup_awsvpc
```
rade: why does the order here differ from `attach-router`?

weave
```diff
@@ -1945,10 +1949,13 @@ EOF
     attach-router)
         check_running $CONTAINER_NAME
         enforce_docker_bridge_addr_assign_type
+        # We cannot use detect_awsvpc here, because HTTP server might not be started
```
rade: Why not change `detect_awsvpc` to this?

brb: I have it in mind, but will do it in a separate PR once you are happy with the check.

weave
brb: fixed.

weave
```diff
@@ -1945,10 +1949,13 @@ EOF
...
+        ! docker inspect -f '{{.Config.Cmd}}' $CONTAINER_NAME | grep -q -- "--awsvpc" || AWSVPC=1
         create_bridge
         fetch_router_args
         setup_router_iface_$BRIDGE_TYPE
         wait_for_status $CONTAINER_NAME http_call $HTTP_ADDR
+        setup_awsvpc
```
brb: IMO the order of requests in the case of re-attaching the router seems to be a bit illogical:

1. `expose_ip` (in `setup_awsvpc`) requests the allocator to allocate a IP for `weave:expose`. The allocator returns the previously owned IP to `weave:expose` which gets assigned to the "weave" bridge afterwards. 
2. `populate_router` tries to reclaim the address returned in the (1) step.

bboreham: See also #2144 - we should really drive the logic the other way round on a reboot.

I think I was holding off on that pending #1958

brb: I added a check before doing the reclaim for `weave:expose`. The check becomes redundant, once we have resolved #2144.

rade:
Can this go into 1.6.1 or is it too low value / high risk?

bboreham:
Why do we restore host routes from the VPC table rather than from IPAM?  Are we worried that they might be out of sync?

bboreham:
I think this could go into 1.6.1.

brb:
> Why do we restore host routes from the VPC table rather than from IPAM? 

The motivation for doing that is to avoid extending the `ipam/tracker` interface and the `ipam/ring` module.

Anyway, it is probably better to assume that IPAM is a single source of truth, so I will change it.

brb:
Done. Re-opening the PR against the 1.6 branch.

PR #2410, opened 2016-06-29
Fallback to 1500 MTU if it is not possible to set it to >1500 on vxlan-vport

### LINKED ISSUE #1853: Fast datapath MTU limited to 1500 on CoreOS 877 (Linux 4.3)

brb:
This check started to look a bit as an overkill to me, mainly because we need to specify exact port for vxlan-vport.

Fixes #1853


brb:
The original PR: https://github.com/weaveworks/weave/pull/2373

prog/weaveutil/datapath.go
```diff
@@ -50,3 +85,56 @@ func addDatapathInterface(args []string) error {
...
+// Helpers
+
+func checkMTU(dpname string, mtu int) bool {
+	var (
+		vpid   libodp.VportID
+		vpname string
+	)
+
+	// Create a dummy vxlan vport
+	for i := 0; i < 5; i++ {
```
bboreham: Why 5?

prog/weaveutil/datapath.go
```diff
@@ -21,7 +30,15 @@ func createDatapath(args []string) error {
 		// status to distinguish it for the weave script.
 		os.Exit(17)
 	}
-	return err
+	if err != nil {
+		return err
+	}
+
+	if err := wnet.SetMTU(dpName, mtu); err != nil {
```
bboreham: Should we remove the datapath on error here?

prog/weaveutil/datapath.go
```diff
@@ -67,3 +85,56 @@ func addDatapathInterface(args []string) error {
...
+		} else if errno, ok := err.(syscall.Errno); !(ok && errno == syscall.EADDRINUSE) {
+			return true
+		}
+	}
+	// Couldn't create the vport, skip the check
+	if vpname == "" {
+		return true
+	}
+
+	// Setting >1500 MTU on affected host' vport should fail with EINVAL
```
bboreham: when you say "should fail", do you mean "will fail if the user is affected by the kernel issue we are looking for"?

common/odp/odp.go
```diff
@@ -84,3 +84,35 @@ func AddDatapathInterface(dpname string, ifname string) error {
 	_, err = dp.CreateVport(odp.NewNetdevVportSpec(ifname))
 	return err
 }
+
+func CreateVxlanVport(dpname string, name string, portno uint16) (odp.VportID, string, error) {
```
bboreham: It looks like `CreateDatapath` could be refactored to call this code, to reduce duplication.  Even the `ListenUDP` call is duplicated, although `CreateDatapath` leaves it open while `checkMTU` closes it before use.

prog/weaveutil/datapath.go
brb: Inspired by https://github.com/weaveworks/weave/blob/master/router/fastdp.go#L402 which we call with `retries = 5`.

prog/weaveutil/datapath.go
brb: Yep, it "_will fail_". Does "_should fail_" read differently?

bboreham: My point is you need the condition.  With no condition it reads like you expect the failure every run.

prog/weaveutil/datapath.go
bboreham: At least leave a comment.  The explanation for that one seems to be at https://github.com/weaveworks/weave/blob/master/router/fastdp.go#L95

brb: The comment you linked describes a bit different problem (async deletion of vxlan vport) which is not relevant to my case.

common/odp/odp.go
brb: I've tried refactoring it in two ways:

* Reuse `CreateDummyVxlanVport` (https://github.com/weaveworks/weave/pull/2410/commits/19f52c9486831ae8b1ad42ad0cd75d111c520bb7)
* Move the MTU check to `common/odp.CreateDatapath` (https://github.com/weaveworks/weave/pull/2410/commits/4090b5f0a55095dbb086ee6663daf10fe8fa0656)

The latter does not create multiple netlink sockets for the checks. Also, it reuses the vxlan-vport created in the support check. However, it introduces the `common/net` artifact (due to circular deps).

prog/weaveutil/datapath.go
bboreham: ok; I'm asking you to add a comment in this code explaining this problem.

prog/weaveutil/datapath.go
brb: Good point. In the bash version we do not remove the datapath if setting MTU fails, so I'm leaving as it is.

brb:
PTAL. Once we are happy with changes, I'll squash / re-oder some commits.

net/odp.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil && !odp.IsDatapathNameAlreadyExistsError(err) {
+		return true, validMTU, err
+	}
+
+	var (
+		vpid      odp.VportID
+		vpname    string
+		errCreate error
+	)
+	if vpid, vpname, errCreate = createDummyVxlanVport(dp); err != nil {
```
brb: The previous code did not close an UDP socket before creating an vxlan-vport. I don't see why closing the socket might be an issue when doing the support check. So to make the vport-vxlan reusable for the latter check, I do close the socket.

rade: doesn't that introduce a race?

brb: I don't think so. If `vxlan_vport` kernel module is not present, the check will fail before trying to open any port.

rade: The race I was thinking of is somebody stealing the port.

awh:
Needs rebasing on `1.7` after the 1.7.0 release.

brb:
PTAL.

bboreham:
This change seems so complex compared to the level of pain users are experiencing (only one person reported it, and they were only trying to run a benchmark), I am going to close it.

PR #2412, opened 2016-06-29
Refactor plugin to create veths in CreateEndpoint

### LINKED ISSUE #1803: `docker network inspect weave` does not show MAC addresses
For example:

````
$ docker network inspect weave
[
    {
        "Name": "weave",
        "Id": "9da63bb3e525fb9c519c3fff2192f0670b85085bb74a35830db86257e7df6e43",
        "Scope": "local",
        "Driver": "weavemesh",
        "IPAM": {
            "Driver": "weavemesh",
            "Config": [
                {}
            ]
        },
        "Containers": {
            "7ccf913ac82de2140ca9ff7a6df71d4f89cce332b5dd8925d2b596f053b23f5e": {
                "EndpointID": "5cecb7e0a676975793d742692119e696e29b5f5a45ff9928551b1b31bafee29c",
                "MacAddress": "",
                "IPv4Address": "10.32.0.1/12",
                "IPv6Address": ""
            }
        },
        "Options": {}
    }
]
````

### PR DESCRIPTION
**Pull request by bboreham**
Re-doing #1808, which was pulled because we were uncertain about the semantics of create/delete vs join/leave.  Subsequent Docker changes indicate that this is not something to worry about.

Fixes #1803.

By moving endpoint creation into CreateEndpoint we can return the MAC address at the time Docker expects. We find the endpoint later simply by looking it up with the same name, derived from the endpoint ID.

The second commit was also tagged onto #1808; changes from using 5 chars from the endpoint-ID to using 8, to reduce the chance of collision from ~1/1000 to ~1/64K. This is not backwards-compatible, e.g. if you stop/upgrade/start the plugin it will log a warning and leak the veth when asked to remove an endpoint.

### DIFF
--- a/plugin/net/driver.go
+++ b/plugin/net/driver.go
@@ -8,6 +8,7 @@ import (
 	"github.com/docker/libnetwork/drivers/remote/api"
 	"github.com/docker/libnetwork/netlabel"
 	"github.com/docker/libnetwork/types"
+	"golang.org/x/sys/unix"
 
 	"github.com/vishvananda/netlink"
 	weaveapi "github.com/weaveworks/weave/api"
@@ -100,7 +101,7 @@ func (driver *driver) CreateEndpoint(create *api.CreateEndpointRequest) (*api.Cr
 		return nil, driver.error("CreateEndpoint", "Not supported: creating an interface from within CreateEndpoint")
 	}
 
-	// create veths. note we assume endpoint IDs are unique in the first 5 chars
+	// create veths. note we assume endpoint IDs are unique in the first 9 chars
 	name, peerName := vethPair(create.EndpointID)
 	if _, err := weavenet.CreateAndAttachVeth(name, peerName, weavenet.WeaveBridgeName, 0, false, nil); err != nil {
 		return nil, driver.error("JoinEndpoint", "%s", err)
@@ -119,7 +120,13 @@ func (driver *driver) DeleteEndpoint(deleteReq *api.DeleteEndpointRequest) error
 	name, _ := vethPair(deleteReq.EndpointID)
 	veth := &netlink.Veth{LinkAttrs: netlink.LinkAttrs{Name: name}}
 	if err := netlink.LinkDel(veth); err != nil {
-		driver.warn("LeaveEndpoint", "unable to delete veth: %s", err)
+		// Try again using the name construction from earlier plugin version,
+		// in case user has upgraded with endpoints still extant
+		veth.Name = "vethwl" + deleteReq.EndpointID[:5]
+		if err2 := netlink.LinkDel(veth); err2 != nil {
+			// Note we report the first error
+			driver.warn("LeaveEndpoint", "unable to delete veth %q: %s", name, err)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugin/net/driver.go [hunk 1]
```diff
@@ -206,7 +211,7 @@ func (driver *driver) DiscoverDelete(disco *api.DiscoveryNotification) error {
 }
 
 func vethPair(id string) (string, string) {
-	return "vethwl" + id[:5], "vethwg" + id[:5]
+	return "vethwl" + id[:8], "vethwg" + id[:8]
```
rade: Do we know for sure that the id will be at least 8 (or, previously, 5) long?

In any case, if we made this safe for shorted ids then I think this would then allow us to avoid the leak on upgrade by attempting to delete `vethPair(id[:5])` if the delete of `vethPair(id)` fails.

rade:
Would this be safe & worthwhile for 1.6.1?

bboreham:
> Would this be safe & worthwhile for 1.6.1?

No customer has asked for this feature; I just noticed it when trying to re-implement `weave ps`, which turned out to be unhelpful given what we had to do for CNI.  So I'd have to say it falls outside of our rules for a point release.

plugin/net/driver.go
bboreham: The IDs come from Docker; I don't know how hard the guarantee is but empirically they're always 64 characters.

bboreham: I added the re-try with 5 chars.

plugin/net/driver.go
```diff
@@ -94,18 +94,39 @@ func (driver *driver) DeleteNetwork(delreq *api.DeleteNetworkRequest) error {
...
 }
 
 func (driver *driver) DeleteEndpoint(deleteReq *api.DeleteEndpointRequest) error {
 	driver.logReq("DeleteEndpoint", deleteReq, deleteReq.EndpointID)
+	name, _ := vethPair(deleteReq.EndpointID)
+	veth := &netlink.Veth{LinkAttrs: netlink.LinkAttrs{Name: name}}
+	if err := netlink.LinkDel(veth); err != nil {
+		// Try again using the name construction from earlier plugin version,
+		// in case user has upgraded with endpoints still extant
+		veth.Name = "vethwl" + deleteReq.EndpointID[:5]
```
rade: `veth.Name = veth.Name[:11]`?

plugin/net/driver.go
rade: Where does the 14 character limit come from? A comment with a link explaining that would be really helpful here.

plugin/net/driver.go
bboreham: I think it's better to explicitly use the construction from the old version than to rely on the similarity which is hidden in another function.

plugin/net/driver.go
bboreham: I re-did the code to import the name limit from `golang.org/x/sys`

rade:
The switch from Join/Leave to Create/Delete makes me nervous. Does this introduce any failure modes / races / leaks?

bboreham:
> The switch from Join/Leave to Create/Delete makes me nervous. 

Well, that was the whole reason this change was pulled the last time.  Then the Docker Overlay plugin was changed from create in Join and delete in Leave to do the veth deletion in Delete, and I asked if it mattered and didn't get much of an answer.  The Docker calling code calls `CreateEndpoint` then `Join` in the same routine, with a deferred `Delete` if things go wrong in between.

As an alternative, we could choose the MAC using the bits from the endpoint ID.

PR #2413, opened 2016-06-29
document --host

### LINKED ISSUE #2347: document `--host`
`--host` is currently not mentioned in the usage, and not documented. See https://github.com/weaveworks/weave/issues/810#issuecomment-224620109 for why that is useful.

rade:
and add some broader points about access control early on, since some users will wonder about that right from the start.

Fixes #2347

### DIFF
--- a/site/using-weave.md
+++ b/site/using-weave.md
@@ -69,6 +69,17 @@ For example:
 
 Peers can also be dynamically added. See [Adding Hosts Dynamically](/site/using-weave/finding-adding-hosts-dynamically.md) for more information.
 
+####Restricting Access
+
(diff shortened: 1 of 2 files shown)


rade:
Probably needs some wordsmithing. cc @abuehrle

abuehrle:
I would add an example to this paragraph (specifying an internal network and an untrusted subnet as well as how to apply firewall rules).

rade:
> I would add an example to this paragraph (specifying an internal network and an untrusted subnet as well as how to apply firewall rules). 

I think that is too much detail for this very early stage documentation. And I don't think that is necessary - we know that Weave users have been configuring firewalls, and haven't received many questions on that subject. Furthermore, the firewall configuration depends on the firewall product being used.

PR #2415, opened 2016-06-30
improve --trusted-subnets docs

### LINKED ISSUE #2211: usage of --trusted-subnets is not clear from docs page
It's not quite clear how to use `--trusted-subnets` from `security-untrusted-networks.md`. It does seem like one can pass it multiple CIDRs, but would be nicer if we documented it right there and not just in the usage summary.

rade:
Fixes #2211


rade:
cc @abuehrle

rade:
@abuehrle I've made some more changes; please take another look.

abuehrle:
LGTM

PR #2416, opened 2016-06-30
Check that weave router is running before launching plugin

### LINKED ISSUE #2293: explain that weave launch-plugin cannot create network when weave router is down
Running the following on an Ubuntu host OS, firewall disabled:

```
ubuntu@ubuntu2950:~$ weave launch-plugin
e37d9a767e481859130e699e3e0cb32329e8eb65babfe94cbc78b18903c8e207
unable to create network: API error (500): IpamDriver.RequestPool: Get http://127.0.0.1:6784/ipinfo/defaultsubnet: dial tcp 127.0.0.1:6784: getsockopt: connection refused
```
 and when I try to launch a container I get: `docker: Error response from daemon: network weave not found.`

Weave versions:
```
weave version
weave script 1.5.1
weave router 1.5.1
weave proxy  1.5.1
weave plugin 1.5.1
```

Bug or fixable?

bboreham:
Fixes #2293.

I decided to error the entire `launch-plugin` operation if `weave` is not running; if I just errored on the network creation then there is no way to re-try that operation; next time round the code will see the plugin is running and return early.

Fundamentally, it doesn't seem to be very useful to run the plugin *and* output an error.

### DIFF
--- a/test/090_docker_restart_policy_2_test.sh
+++ b/test/090_docker_restart_policy_2_test.sh
@@ -23,8 +23,8 @@ assert_raises "! check_restart $HOST1 weave"
 assert_raises "! check_restart $HOST1 weaveproxy"
 assert_raises "! check_restart $HOST1 weaveplugin"
 
-# Relaunch the plugin to prevent the `weave stop` in `end_suite`
+# Relaunch to prevent the `weave stop` in `end_suite`
 # timing out trying to remove the plugin network
-weave_on $HOST1 launch-plugin
+weave_on $HOST1 launch --no-restart
 
 end_suite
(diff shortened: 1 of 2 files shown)


rade:
This has broken a test:
```
>>> Test ./090_docker_restart_policy_2_test.sh finished after 47.0 secs with error: exit status 1
Test docker restart policy
ERROR: weave container must be running before plugin can be launched
```

bboreham:
The broken test was starting just the plugin because it "knew this was OK"; the network had already been created.  We may expect that customers have the same assumption in special cases, so we need to highlight this in the release notes, but at the end of the day it isn't OK to run the plugin without the router.

rade:
argh. I've just realised that this was on the wrong branch.

rade:
> argh. I've just realised that this was on the wrong branch.

Fixed.

PR #2418, opened 2016-07-04
Fix "weave ps"

### LINKED ISSUE #2388: 'weave ps' fails with "Cannot find weave bridge: Link not found"
... even if the weave bridge exists:

```
weave     Link encap:Ethernet  HWaddr 06:c2:01:12:41:d5  
          inet addr:10.32.0.1  Bcast:0.0.0.0  Mask:255.240.0.0
          inet6 addr: fe80::4c2:1ff:fe12:41d5/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1410  Metric:1
          RX packets:732925 errors:0 dropped:0 overruns:0 frame:0
          TX packets:1016150 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:597366797 (597.3 MB)  TX bytes:2877166348 (2.8 GB)
```

This happened when running a local instance of the scope service (kubernetes anywhere). 

Running with `bash -x` I get:

```
+ docker run --rm --privileged --net=host -v /var/run/docker.sock:/var/run/docker.sock --pid=host -e DOCKERHUB_USER=weaveworks -e WEAVE_VERSION -e WEAVE_DEBUG -e WEAVE_DOCKER_ARGS -e WEAVEPROXY_DOCKER_ARGS -e WEAVEPLUGIN_DOCKER_ARGS -e WEAVE_PASSWORD -e WEAVE_PORT -e WEAVE_HTTP_ADDR -e WEAVE_CONTAINER_NAME -e WEAVE_MTU -e WEAVE_NO_FASTDP -e WEAVE_NO_BRIDGED_FASTDP -e WEAVE_NO_PLUGIN -e DOCKER_BRIDGE -e DOCKER_CLIENT_HOST= -e DOCKER_CLIENT_ARGS -e PROXY_HOST=127.0.0.1 -e COVERAGE -e CHECKPOINT_DISABLE weaveworks/weaveexec:1.5.2 --local ps
Cannot find weave bridge: Link not found
```

Output when running the docker command above with WEAVE_DEBUG=1: https://gist.github.com/2opremio/cfe5c7fa236854e9963dda0f8afc5754

Some details about my system:

```
$ uname -a
Linux vagrant-ubuntu-wily-64 4.2.0-36-generic #42-Ubuntu SMP Thu May 12 22:05:35 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

```
$ docker version
Client:
 Version:      1.10.2
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   c3959b1
 Built:        Mon Feb 22 21:40:35 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.2
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   c3959b1
 Built:        Mon Feb 22 21:40:35 2016
 OS/Arch:      linux/amd64
```

brb:
This PR fixes the problem described in #2388. In addition, the PR optimizes the `GetBridgeNetDevs` and `containerAddrs` functions.

Keep in mind, that the fix is specific for this case, because we move all the code depending on a non-host netns to the bottom of `containerAddrs` instead of making `WithNetNS` safe.

To address the safety problem, I've filed https://github.com/weaveworks/weave/issues/2419 .

Fixes #2388.

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -90,29 +90,24 @@ func linkToNetDev(link netlink.Link) (NetDev, error) {
 	return netDev, nil
 }
 
-// Lookup the weave interface of a container
-func GetWeaveNetDevs(processID int) ([]NetDev, error) {
-	// Bail out if this container is running in the root namespace
-	nsToplevel, err := netns.GetFromPid(1)
+// ConnectedToBridgePredicate returns a function which is used to query whether
+// a given link is a veth interface which one end is connected to a bridge.
+// The returned function should be called from a container network namespace which
+// the bridge does NOT belong to.
+func ConnectedToBridgePredicate(bridgeName string) (func(link netlink.Link) bool, error) {
+	var br netlink.Link
+
+	err := weavenet.WithNetNSLinkByPid(1, bridgeName, func(link netlink.Link) error {
+		br = link
+		return nil
+	})
 	if err != nil {
-		return nil, fmt.Errorf("unable to open root namespace: %s", err)
-	}
-	nsContainr, err := netns.GetFromPid(processID)
-	if err != nil {
-		return nil, fmt.Errorf("unable to open process %d namespace: %s", processID, err)
-	}
-	if nsToplevel.Equal(nsContainr) {
-		return nil, nil
+		return nil, err
 	}
 
-	weaveBridge, err := netlink.LinkByName("weave")
-	if err != nil {
-		return nil, fmt.Errorf("Cannot find weave bridge: %s", err)
-	}
-	// Scan devices in root namespace to find those attached to weave bridge
 	indexes := make(map[int]struct{})
 	err = forEachLink(func(link netlink.Link) error {
-		if link.Attrs().MasterIndex == weaveBridge.Attrs().Index {
+		if link.Attrs().MasterIndex == br.Attrs().Index {
 			peerIndex := link.Attrs().ParentIndex
 			if peerIndex == 0 {
 				// perhaps running on an older kernel where ParentIndex doesn't work.
@@ -126,11 +121,39 @@ func GetWeaveNetDevs(processID int) ([]NetDev, error) {
 	if err != nil {
 		return nil, err
 	}
-	return FindNetDevs(processID, func(link netlink.Link) bool {
+
+	return func(link netlink.Link) bool {
 		_, isveth := link.(*netlink.Veth)
 		_, found := indexes[link.Attrs().Index]
 		return isveth && found
-	})
+	}, nil
+
+}
+
+func GetNetDevsWithPredicate(processID int, predicate func(link netlink.Link) bool) ([]NetDev, error) {
+	// Bail out if this container is running in the root namespace
+	nsToplevel, err := netns.GetFromPid(1)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open root namespace: %s", err)
+	}
+	nsContainr, err := netns.GetFromPid(processID)
+	if err != nil {
+		return nil, fmt.Errorf("unable to open process %d namespace: %s", processID, err)
+	}
+	if nsToplevel.Equal(nsContainr) {
+		return nil, nil
+	}
+
+	return FindNetDevs(processID, predicate)
+}
+
+// Lookup the weave interface of a container
+func GetWeaveNetDevs(processID int) ([]NetDev, error) {
+	p, err := ConnectedToBridgePredicate("weave")
+	if err != nil {
+		return nil, err
+	}
+	return GetNetDevsWithPredicate(processID, p)
 }
 
 // Get the weave bridge interface.
--- a/common/utils.go
+++ b/common/utils.go
@@ -95,29 +95,24 @@ func linkToNetDev(link netlink.Link) (NetDev, error) {
 // The returned function should be called from a container network namespace which
 // the bridge does NOT belong to.
 func ConnectedToBridgePredicate(bridgeName string) (func(link netlink.Link) bool, error) {
-	var br netlink.Link
-
-	err := weavenet.WithNetNSLinkByPidUnsafe(1, bridgeName, func(link netlink.Link) error {
-		br = link
-		return nil
-	})
-	if err != nil {
-		return nil, err
-	}
-
 	indexes := make(map[int]struct{})
-	err = forEachLink(func(link netlink.Link) error {
-		if link.Attrs().MasterIndex == br.Attrs().Index {
-			peerIndex := link.Attrs().ParentIndex
-			if peerIndex == 0 {
-				// perhaps running on an older kernel where ParentIndex doesn't work.
-				// as fall-back, assume the indexes are consecutive
-				peerIndex = link.Attrs().Index - 1
-			}
-			indexes[peerIndex] = struct{}{}
-		}
-		return nil
-	})
+
+	// Scan devices in root namespace to find those attached to weave bridge
+	err := weavenet.WithNetNSLinkByPidUnsafe(1, bridgeName,
+		func(br netlink.Link) error {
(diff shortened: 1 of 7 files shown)


common/utils.go
```diff
@@ -133,11 +133,20 @@ func GetWeaveNetDevs(processID int) ([]NetDev, error) {
...
 
-// Get the weave bridge interface
-func GetBridgeNetDev(bridgeName string) ([]NetDev, error) {
-	return FindNetDevs(1, func(link netlink.Link) bool {
-		return link.Attrs().Name == bridgeName
+// Get the weave bridge interface.
+func GetBridgeNetDev(bridgeName string) (NetDev, error) {
+	var netdev NetDev
+
+	err := weavenet.WithNetNSLinkByPid(1, bridgeName, func(link netlink.Link) error {
```
bboreham: I think this code was only requesting the root namespace because it was necessary to pass something into `FindNetDevs`.  I believe all uses of `GetBridgeNetDev` are in a context where we may assume we are in the root namespace, so this isn't necessary.

prog/weaveutil/addrs.go
```diff
@@ -41,7 +41,11 @@ func containerAddrs(args []string) error {
 
 func getNetDevs(bridgeName string, c *docker.Client, containerID string) ([]common.NetDev, error) {
 	if containerID == "weave:expose" {
-		return common.GetBridgeNetDev(bridgeName)
+		if netDev, err := common.GetBridgeNetDev(bridgeName); err != nil {
```
bboreham: Maybe just 

    netDev, err := common.GetBridgeNetDev(bridgeName)
    return []common.NetDev{netDev}, err
?

prog/weaveutil/main.go
```diff
@@ -30,6 +30,7 @@ func main() {
 		usage()
 		os.Exit(1)
 	}
+
```
bboreham: hmm

common/utils.go
```diff
@@ -90,29 +90,24 @@ func linkToNetDev(link netlink.Link) (NetDev, error) {
...
-	}
+// ConnectedToBridgePredicate returns a function which is used to query whether
+// a given link is a veth interface which one end is connected to a bridge.
+// The returned function should be called from a container network namespace which
+// the bridge does NOT belong to.
+func ConnectedToBridgePredicate(bridgeName string) (func(link netlink.Link) bool, error) {
+	var br netlink.Link
 
-	weaveBridge, err := netlink.LinkByName("weave")
+	err := weavenet.WithNetNSLinkByPid(1, bridgeName, func(link netlink.Link) error {
```
bboreham: Why did this have to change?

common/utils.go
```diff
@@ -90,29 +90,24 @@ func linkToNetDev(link netlink.Link) (NetDev, error) {
...
+		br = link
+		return nil
+	})
 	if err != nil {
-		return nil, fmt.Errorf("Cannot find weave bridge: %s", err)
+		return nil, err
 	}
-	// Scan devices in root namespace to find those attached to weave bridge
+
```
bboreham: You didn't like this comment?

common/utils.go
bboreham: OK, it is necessary if `weave:expose` can come in any position in the list of containers in `weave ps`.

bboreham: You could pull the check for `weave:expose` up into another loop that executed before you started changing namespace, then this wouldn't be necessary.  Maybe that's too obscure.

prog/weaveutil/addrs.go
```diff
@@ -11,47 +13,65 @@ func containerAddrs(args []string) error {
...
-			if _, ok := err.(*docker.NoSuchContainer); ok {
-				continue
+	pred, err := common.ConnectedToBridgePredicate(bridgeName)
+	if err != nil {
+		return err
+	}
+
+	containers := make(map[string]*docker.Container)
+	for _, cid := range containerIDs {
+		if containers[cid], err = client.InspectContainer(cid); err != nil {
```
bboreham: We could hoist `weave:expose` handling up to here, then we wouldn't try to `Inspect` it and we wouldn't have to do special pid-1 handling for the weave bridge address.

common/utils.go
brb: Forgot to c&p. Thanks for noting.

prog/weaveutil/main.go
brb: ouch :(

prog/weaveutil/addrs.go
brb: I'd be in favour of returning empty list in the case of error.

common/utils.go
brb: Good spot. Anyway, I'm wrapping both with `WithNetNS*`, because the `GetWeaveNetDevs`, which calls the predicate function, is used by the proxy and atm `WithNetNS*` is not safe.

prog/weaveutil/addrs.go
brb: I've considered that, but as you mentioned it makes the first loop a bit obscure. Anyway, changing it for the sake of simplicity.

prog/weaveutil/addrs.go
```diff
@@ -12,46 +14,63 @@ func containerAddrs(args []string) error {
...
 
 	return nil
 }
 
-func getNetDevs(bridgeName string, c *docker.Client, containerID string) ([]common.NetDev, error) {
-	if containerID == "weave:expose" {
-		return common.GetBridgeNetDev(bridgeName)
-	}
-
-	container, err := c.InspectContainer(containerID)
```
brb: Ehrm, I've changed the behaviour of `container-addrs`: if a user passes a non-existing container id, we return an error instead of ignoring it. This becomes relevant in racy cases, e.g. container dies immediately after we have invoked `container-addrs`. I'll revert the changes to the behaviour.

bboreham: Just to clarify, this comment about "changed the behaviour of `container-addrs`" is now out of date because you reverted the change?

common/utils.go
```diff
@@ -77,21 +77,54 @@ func forEachLink(f func(netlink.Link) error) error {
...
-func GetWeaveNetDevs(processID int) ([]NetDev, error) {
+// ConnectedToBridgePredicate returns a function which is used to query whether
+// a given link is a veth interface which one end is connected to a bridge.
+// The returned function should be called from a container network namespace which
+// the bridge does NOT belong to.
+func ConnectedToBridgePredicate(bridgeName string) (func(link netlink.Link) bool, error) {
+	indexes := make(map[int]struct{})
+
+	// Scan devices in root namespace to find those attached to weave bridge
+	err := weavenet.WithNetNSLinkByPidUnsafe(1, bridgeName,
```
bboreham: If I understand the series of comments, this changed from assuming we are in the root namespace to assuming we might not be, because it is called from the Proxy which also changes to other namespaces.

But, I believe we are changing the proxy to exec another process to change namespace, and once we have that we can revert this change which will simplify the code here and remove the need for `WithNetNSLinkByPidUnsafe()`

bboreham:
LGTM assuming my questions for clarification come out the way I think.

prog/weaveutil/addrs.go
brb: Yep, I've reverted the change. However, this PR changes the order of inspecting: `foreach container { inspect && print}` -> `foreach container { inspect }; foreach container { print }`, so it extends the "window" for a race when a container dies immediately after the inspect.

common/utils.go
brb: Yep, your reasoning is correct - once we have made `WithNetNS*` safe (which happens in #2419), the wrapper becomes redundant.

PR #2423, opened 2016-07-05
Refactor CNI documentation.

philwinder:
- Add Mesos CNI instructions
- Rename page to something more generic.
- Fix old links.
- Minor editing.

Review: @abuehrle 
cc: @bboreham, @pidster

### DIFF
--- a/site/cni-plugin.md
+++ b/site/cni-plugin.md
@@ -1,5 +1,5 @@
 ---
-title: Weave CNI Plugin
+title: Integrating Kubernetes and Mesos via the CNI Plugin
 menu_order: 65
 ---
 
--- a/site/cni-plugin.md
+++ b/site/cni-plugin.md
@@ -1,13 +1,14 @@
 ---
-title: Integrating Kubernetes via the CNI Plugin
+title: Weave CNI Plugin
 menu_order: 65
 ---
 
-CNI, the [_Container Network Interface_](https://github.com/appc/cni#cni---the-container-network-interface),
+CNI, the [_Container Network Interface_](https://github.com/containernetworking/cni),
 is a proposed standard for configuring network interfaces for Linux
 application containers.  CNI is supported by
-[Kubernetes](http://kubernetes.io/), an open source container cluster
-manager built by Google.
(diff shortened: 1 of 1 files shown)


abuehrle:
LGTM although, I do have some other edits on that page that I can address on a different PR.  Also, we should have a guide that shows an example of this as well, which I'll mention to @lukemarsden

site/cni-plugin.md
```diff
@@ -1,13 +1,14 @@
 ---
-title: Integrating Kubernetes via the CNI Plugin
+title: Weave CNI Plugin
```
rade: @abuehrle I thought we wanted all headings to be "actions". Also, I am concerned that w/o k8s/mesos being mentioned in the heading, folks really won't know know what this is about - 'CNI' is not a well-known term.

So I'd opt for 'Integrating Kubernetes and Mesos via the CNI Plugin', but that's rather long.

abuehrle: How about:  "Configuring the CNI Plugin for Kubernetes and Mesos"

rade: IMO that's putting the cart before the horse. "Configuring the CNI Plugin" is not something a user wants to do. It's a means to and end. And the end is using Weave Net with K8s or Mesos.

Also, note the two TOC entries just above this one:
- Integrating Docker via the API Proxy
- Integrating Docker via the Network Plugin

The title I suggested follows the same pattern.

abuehrle: sure.

abuehrle: 'Integrating Kubernetes and Mesos via the CNI Plugin' is fine with me, if you'd like to keep the same structure.  Regarding the action rule.  It's not an all in rule, IMO.  Normally you'd want the headings to be action-orientated if the user is required to do something.  But if it's straight up description or an explanation, then I'd say a regular title is just fine.  Sometimes a document might have a combination of these titles.

rade: @philwinder would you mind making that small change?

philwinder:
@rade: Done.

Issue #2406, opened 2016-06-28
weave --no-dns not works when weave-plugin is used


aksentyev:
Weave router was started with --no-dns option.When I run a docker container using weave socket everything is OK. Host dns is used there. When docker container has been run with --net=weave (using plugin) host dns settings are not appeared and weave dns is used.

bboreham:
Can you give a few more details of how you came to this conclusion?  I checked the code, and we don't create the struct that manages DNS if you say `--no-dns`.

Docker's own DNS is pretty similar these days, and is forced on everybody using a network via `--net=`.

If you still think you are getting Weave DNS, please supply the output of `docker logs weave` and also the contents of `/etc/resolv.conf` inside the container.

rade:
Closing this since we cannot reproduce it; if new information comes to light, please re-open.

aksentyev:
Hello,

aksentyev:
Hello, I run weave-router with this command
`weave launch-router --no-dns $PEER_IPs`,
then I get:

rade:
Update from @aksentyev via #2429:

I run weave-router with this command
weave launch-router --no-dns $PEER_IPs,

then I get in the container:

`/home/weave/weaver --port 6783 --name 42:8d:c2:87:2a:37 --nickname some_name --datapath datapath --ipalloc-range 10.32.0.0/12 --dns-effective-listen-address 172.17.0.1 --no-dns --http-addr 127.0.0.1:6784 --resolv-conf /var/run/weave/etc/resolv.conf --no-dns some_peers_ips`.

According to the weave sources `--dns-effective-listen-address` means:

    if config.EffectiveListenAddress != "" {
        listenAddr = config.EffectiveListenAddress
    }
    Log.Println("Listening for DNS queries on", listenAddr)
    return ns, dnsserver
}
So, the problem is in the shell script that runs weave router.

Sorry for long time waiting my answer.

rade:
The fact that the `--dns-effective-listen-address` is specified does not mean DNS will be listening; `--no-dns` makes that option ineffective.

Do you actually see a log line saying "Listening for DNS queries on..." in your weave logs? If not then DNS isn't listening.

rade:
There is actually a tiny bug here in that `--no-dns` appears twice in the command line. But that is harmless.

aksentyev:
Ok, maybe.
With weave plugin my resolv.conf looks like this:
```
nameserver 127.0.0.11
options ndots:0
```

[...]

PR #2430, opened 2016-07-06
remove duplication of `--no-dns`

rade:
This was harmless, but annoying.

Discovered while investigating #2406.


rade:
I've put this in 1.6 since it's a tiny change with no user impact, and doing so reduces divergence with master.

PR #2439, opened 2016-07-11
refactor: restore grouping of functions in script

rade:
Recent(ish) changes placed functions in the wrong sections. Fix that and label the sections more clearly.


rade:
I've put this in 1.6 since it's the kind of change that would otherwise cause continued merge headaches.

Note that github does a rather mediocre job of showing what has actually happened here, which is simply moving some functions and adding some section heading comments.

PR #2443, opened 2016-07-14
Add a safer version of WithNetNS*

### LINKED ISSUE #2419: Fix the usage of WithNetNS* functions
As it's been mentioned in #2388 , the `WithNetNS*` family of functions and their usage are not safe. To address the issues, the work function has to be executed in a separate process which is not associated with the Go runtime process which created it.

Currently, `WithNetNS*` is used by the following functions:

- [x] `common/utils.go:FindNetDevs`
- [x] `common/utils.go:GetWeaveNetDevs`
- [x] `net/veth.go:interfaceExistsInNamespace`
- [x] `net/veth.go:AttachContainer`
- [x] `net/veth.go:DetachContainer`
- [x] `plugin/net/cni.go:CmdAdd`
- [x] `plugin/net/cni.go:CmdDel`

In some cases, shelling out is not necessary - it's enough to move a netns dependant code to the end of execution and to ensure that the code does not create any go-routine which require a particular netns.

brb:
This PR introduces a safer version of `WithNetNS*`. The new function executes `work()` by shelling out `weaveutil` which before executing it enters a given network namespace (passed via `--netns-fd`).

**EDIT** ~~Unfortunately, the function isn't totally safe - any go-routine spawned by`work()` might be run in a wrong netns. To fix that we either need to create a small in non-Go shim or to compile `weaveutil` with a custom constructor function which gets called before the runtime initialization (EDITED: I'll file an issue for it once we have discussed in this PR).~~ `nsenter` is used when calling `weaveutil`.

Also, the PR replaces the usage of `WithNetNS*Unsafe` with the newly introduced function where needed.

Fixes #2419


brb:
@awh PTAL. I have replaced `weaveutil --netns-fd=... <cmd>` with `nsenter <..> weaveutil <cmd>`.

awh:
@brb as discussed too invasive for a bugfix release - please reopen against master and update the issue milestone to 1.7.0. Thanks!

brb:
Done: https://github.com/weaveworks/weave/pull/2475

PR #2444, opened 2016-07-14
Fix "weave ps" output when weave is not running

brb:
The #2418 has changed a behaviour of `weave ps` cmd when weave is not running: instead of outputting an empty line, the cmd returns `Link not found` error.

To fix that, we check (in an ugly way, because the netlink package does not export sentinel errors) whether the "weave" bridge can be found, as it has been done before through bunch of indirections.

### DIFF
--- a/prog/weaveutil/addrs.go
+++ b/prog/weaveutil/addrs.go
@@ -1,6 +1,7 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


brb:
Will open against 1.6

PR #2445, opened 2016-07-14
Fix "weave ps" output when weave is not running

brb:
The #2418 has changed a behaviour of `weave ps` cmd when weave is not running: instead of outputting an empty line, the cmd returns `Link not found` error.

To fix that, we check (in an ugly way, because the netlink package does not export sentinel errors) whether the "weave" bridge can be found, as it has been done before through bunch of indirections.

### DIFF
--- a/prog/weaveutil/addrs.go
+++ b/prog/weaveutil/addrs.go
@@ -5,7 +5,9 @@ import (
 
 	"github.com/fsouza/go-dockerclient"
 	"github.com/vishvananda/netlink"
+
 	"github.com/weaveworks/weave/common"
+	weavenet "github.com/weaveworks/weave/net"
 )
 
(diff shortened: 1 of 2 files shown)


bboreham:
Maybe return your own sentinel error from `ConnectedToBridgePredicate` that could be tested?

prog/weaveutil/addrs.go
```diff
@@ -21,6 +22,9 @@ func containerAddrs(args []string) error {
 
 	pred, err := common.ConnectedToBridgePredicate(bridgeName)
 	if err != nil {
+		if err.Error() == errors.New("Link not found").Error() {
```
rade: you could make that rhs a package level var.

brb:
Thanks for the suggestion. PTAL.

PR #2446, opened 2016-07-15
Make use of `sudo -s` more explicit

rade:
since people keep tripping over that.

Also

- add new error message spat out by recent dockers correct the socket
- URLs - they had fallen victim to a 'docker'->'Docker' replacement
- shorten the note on using the plugin

### DIFF
--- a/site/using-weave.md
+++ b/site/using-weave.md
@@ -29,26 +29,26 @@ Where,
  * The second line configures the Weave Net environment, so that containers launched via the Docker command line are automatically attached to the Weave network, and, 
  * The third line runs the [application container](/site/faq.md#own-image) using [a Docker command](https://docs.docker.com/engine/reference/run/).
 
-If the first command results in an error like
-
-    Cannot connect to the Docker daemon. Is the docker daemon running on this host?
-
-or
-
-    http:///var/run/docker.sock/v1.19/containers/create: dial unix/var/run/docker.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?
-
-then you likely need to be 'root' in order to connect to the Docker
-daemon. If so, run the above and all subsequent commands in a
-*single* root shell:
-
-    host1$ sudo -s
-    host1# weave launch
-    host1# eval $(weave env)
-    host1# docker run --name a1 -ti weaveworks/ubuntu
-
-Do *not* prefix individual commands with `sudo`, since some commands
-modify environment entries and hence they all need to be executed from
-the same shell.
+> **Note** If the first command results in an error like
+> ```
+>    Cannot connect to the Docker daemon. Is the docker daemon running on this host?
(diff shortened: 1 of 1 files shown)


rade:
cc @abuehrle

rade:
Can't say I am happy with this. The extra detail really interrupts the flow. But equally, it's a lot to put in into a 'note', plus it's hard (impossible?) to do that with nice spacing. And, OTOH, there is too little content here to justify a new page. @abuehrle @awh thoughts?

rade:
Fixing #2447 would make this all go away.

site/using-weave.md
```diff
@@ -29,12 +29,30 @@ Where,
...
+    host1# weave launch
+    host1# eval $(weave env)
+    host1# docker run --name a1 -ti weaveworks/ubuntu
+
+Do *not* prefix individual commands with `sudo`, since some commands
+modify environment entries and hence they all need to be executed from
+the same shell.
+
+> **Important!** if you are intending to use the the
+> [Weave Docker Network Plugin](/site/plugin.md) do not run `eval
```
abuehrle: make the sentence active instead of passive by: 

If you intend to use the...

Also, there's an extra 'the'

site/using-weave.md
```diff
@@ -29,12 +29,30 @@ Where,
...
+
+    host1$ sudo -s
+    host1# weave launch
+    host1# eval $(weave env)
+    host1# docker run --name a1 -ti weaveworks/ubuntu
+
+Do *not* prefix individual commands with `sudo`, since some commands
+modify environment entries and hence they all need to be executed from
+the same shell.
+
```
abuehrle: I think we should include a short note on the proxy as well, and then link to those topics, since it trips up people as well. Something like, to attach containers to a Weave network using regular docker commands, you must run the Weave Docker API proxy, however if you intend to use the Docker plugin, then Weave Docker API proxy is not necessary.  See, X for a larger explanation.

rade: The above *is* running the proxy. There is no harm in running the proxy, even if you intend to use the plugin. It's not something we need to talk about. We *do* need to talk about not running the 'eval', which we do below.

rade:
> foo bar
> ```
>    baz boo
> ```
> moo

site/using-weave.md
rade: fixed

rade:
>  it's a lot to put in into a 'note', plus it's hard (impossible?) to do that with nice spacing.

I have put it all back into a block quote, with nice spacing. Least worst option, IMO.

site/using-weave.md
rade: Actually since #2330 (i.e. 1.6.0) it is perfectly fine to use the plugin when the 'eval' has been done. Will fix the docs.

Issue #2451, opened 2016-07-20
Weave 1.6.0 only one connection possible


gw0:
In Weave 1.4.5 one could prepare a VM image with Docker and Weave inside that could be deployed anywhere and join the weave network with encrypted sleeve data path to a master server. Although all the images would have the same hostname, each one would use a different IP range, and everything would work as expected (even hosts could be exposed on the weave network).

```
WEAVE_NO_FASTDP=true
WEAVE_PASSWORD="foobar"
PEERS="11.22.33.10"
```

With Weave 1.6.0 this does not work anymore and in case you start the same image multiple times, only the last node stays connected to the weave network, others get the connection dropped with EOF. Same happens even if after a fresh `weave reset --force` on all nodes and bringing them up one by one. Logs on the master server contain:

```
# docker logs weave
INFO: 2016/07/20 14:32:40.669340 ->[11.22.33.131:17628] connection accepted
INFO: 2016/07/20 14:32:40.672749 ->[11.22.33.131:17628|aa:bb:cc:69:74:84(docker-vm2)]: connection ready; using protocol version 2
INFO: 2016/07/20 14:32:40.672824 overlay_switch ->[aa:bb:cc:69:74:84(docker-vm2)] using sleeve
INFO: 2016/07/20 14:32:40.672850 ->[11.22.33.131:17628|aa:bb:cc:69:74:84(docker-vm2)]: connection shutting down due to error: Multiple connections to aa:bb:cc:69:74:84(docker-vm2) added to aa:bb:cc:ba:10:86(master)
INFO: 2016/07/20 14:32:40.680709 sleeve ->[11.22.33.131:6783|aa:bb:cc:69:74:84(docker-vm2)]: Failed to decode packet: UDP packet decryption failed
```

Is there any flag to enable this again? Or is there another recommended approach to setup a cluster from the same VM image?

awh:
Hi @gw0 

In 1.5 we started using product & hypervisor UUIDs as a method of generating peer names that would be stable across reboots:
* #1866 Derive peer name from product_uuid
* #2037 fall back to hypervisor uuid for mac derivation
* [Operational Guide - Concepts - Peer Name](https://www.weave.works/docs/net/latest/operational-guide/concepts/)

[...]

PR #2442, opened 2016-07-13
CNI setup fixes

rade:
Fixes #2434
Fixes #2394
Fixes #2435

PR #2460, opened 2016-08-01
Detect and report peer name collisions

### LINKED ISSUE #2427: name/mac derivation not unique on some hosted and physical machines
This is no real issue - just a hint.

Online.net has cheap dedicated hosts (Dedibox) which seem to share all the same product UUID: 
12345678-1234-5678-90AB-CDDEEFAABBCC.

This causes identical peer names, which results in a non-working weave installation.

To run weave on these hosts, the peer names has to be specified manually with e.g. `--name ::1`.

I suggest an entry in the FAQ section of the documentation regarding this problem. Maybe there are other hosters that do the same thing with their firmware...

brb:
~~**NB**: Please review https://github.com/weaveworks/mesh/pull/51 first. After it has been reviewed and merged I will update the necessary submodule.~~ **DONE**.

This PR changes the way we generate the bridge MAC. Instead of relying on a single source, we concatenate multiple sources. In addition, we include the machine id, which is generated either by [systemd](https://www.freedesktop.org/software/systemd/man/machine-id.html) or [D-Bus](https://dbus.freedesktop.org/doc/dbus-uuidgen.1.html). The machine id is unique and should be stable across restarts unless a user regenerates it.

CC'ing @abuehrle as there are changes to the docs.

Fixes #2427 and probably #2451.

### DIFF
--- a/site/operational-guide/concepts.md
+++ b/site/operational-guide/concepts.md
@@ -49,12 +49,16 @@ When the router is launched on a host, it derives its peer name in order of pref
 
 * From the command line, where the user is responsible for uniqueness and
   stability
-* From the BIOS product UUID, which is generally stable across
-  restarts and unique across different physical hardware and certain
-  cloned VMs
-* From the hypervisor UUID, which is generally stable across restarts
-  and unique across VMs which do not provide access to a BIOS product
-  UUID
+* From the concatenated result of any of the following:
+    * The unique machine ID, which is generated either by [systemd](https://www.freedesktop.org/software/systemd/man/machine-id.html)
+      or [D-Bus](https://dbus.freedesktop.org/doc/dbus-uuidgen.1.html) and is
+      practically stable across restarts unless a user re-generates it.
+    * The BIOS product UUID, which is generally stable across
+      restarts and unique across different physical hardware and certain
+      cloned VMs
+    * The hypervisor UUID, which is generally stable across restarts
+      and unique across VMs which do not provide access to a BIOS product
+      UUID
 * From a random value, practically unique across different physical
   hardware and cloned VMs but not stable across restarts
 
--- a/weave
+++ b/weave
@@ -143,12 +143,22 @@ launch_router_options() {
     [ "$1" = "launch" -o "$1" = "launch-router" ] && echo -v /:/host -e HOST_ROOT=/host
 }
 
+exec_options() {
+    case "$1" in
+        # Required by create_bridge which might need machine-id
+        launch|launch-router|create-bridge|attach-router|run|start|attach|restart|expose|hide)
+            echo -v /etc:/host/etc -v /var/lib/dbus:/host/var/lib/dbus -e HOST_ROOT=/host
+            ;;
+    esac
+}
+
 exec_remote() {
     docker $DOCKER_CLIENT_ARGS run --rm \
         $(docker_run_options) \
         --pid=host \
         $(cni_volume_options "$@") \
         $(launch_router_options "$@") \
+        $(exec_options "$@") \
         -e DOCKERHUB_USER="$DOCKERHUB_USER" \
         -e WEAVE_VERSION \
         -e WEAVE_DEBUG \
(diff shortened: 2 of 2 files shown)


weave
```diff
@@ -696,16 +703,16 @@ init_bridge() {
...
-    MAC=$(mac_from_file /sys/class/dmi/id/product_uuid) || \
-        MAC=$(mac_from_file /sys/hypervisor/uuid) || \
-        MAC=$(random_mac)
+    # Derive the bridge MAC from concatenated machine-id (either systemd or
+    # dbus), the system (aka bios) UUID and the hypervisor UUID. If we cannot
+    # determine all of them we just generate a random MAC.
+    # Elsewhere we in turn derive the peer name from the bridge MAC, which we
+    # want to be stable across reboots but otherwise unique.
+
+    local uid="$MACHINE_ID"
```
rade: we don't use 'local' *anywhere else*.

weave
```diff
@@ -319,6 +320,13 @@ elif [ "$1" = "run" -a -z "$IS_LOCAL" ] ; then
     exit 0
 fi
 
+machineid() {
+    [ -e "/etc/machine-id" ] && cat "/etc/machine-id" && return 0 || true
+    [ -e "/var/lib/dbus/machine-id" ] && cat "/var/lib/dbus/machine-id" && return 0 || true
+}
+
+[ -n "$IS_LOCAL" ] || MACHINE_ID=$(machineid)
```
rade: We are not interested in the docker client's machine id but the server's. So the machine id must be determined on the server.

weave
brb: I know. I was wondering whether there is any particular reason for not using `local` in the weave script besides brevity?

Also, how do we denote that the scope of a variable is local? By lower case? If yes, then there are some inconsistencies, e.g. https://github.com/weaveworks/weave/blob/master/weave#L1414 vs https://github.com/weaveworks/weave/blob/master/weave#L473.

weave
```diff
@@ -143,12 +143,22 @@ launch_router_options() {
     [ "$1" = "launch" -o "$1" = "launch-router" ] && echo -v /:/host -e HOST_ROOT=/host
 }
 
+exec_options() {
```
brb: Will rebase after https://github.com/weaveworks/weave/pull/2442 has been merged.

site/operational-guide/concepts.md
```diff
@@ -49,12 +49,16 @@ When the router is launched on a host, it derives its peer name in order of pref
...
 * From the command line, where the user is responsible for uniqueness and
   stability
-* From the BIOS product UUID, which is generally stable across
-  restarts and unique across different physical hardware and certain
-  cloned VMs
-* From the hypervisor UUID, which is generally stable across restarts
-  and unique across VMs which do not provide access to a BIOS product
-  UUID
+* From a product of concatenation of:
+    * The unique machine ID, which is generated either by [systemd](https://www.freedesktop.org/software/systemd/man/machine-id.html)
```
abuehrle: The first sentence, edit to (if this is what was meant? any of these concatenated?). 

From the concatenated result of any of the following:

weave
brb: Done.

weave
```diff
@@ -696,16 +710,16 @@ init_bridge() {
...
-    # reboots but otherwise unique. The system/hypervisor UUID fits
-    # that bill, unlike, say, /etc/machine-id, which is often
-    # identical on VMs created from cloned filesystems. If we cannot
-    # determine the system/hypervisor UUID we just generate a random MAC.
-    MAC=$(mac_from_file /sys/class/dmi/id/product_uuid) || \
-        MAC=$(mac_from_file /sys/hypervisor/uuid) || \
-        MAC=$(random_mac)
+    # Derive the bridge MAC from concatenated machine-id (either systemd or
+    # dbus), the system (aka bios) UUID and the hypervisor UUID. If we cannot
+    # determine all of them we just generate a random MAC.
```
bboreham: "cannot determine any of them"

PR #2462, opened 2016-08-02
added note about net flag

abuehrle:
Added note to proxy about using the net flag, and what it does. 

Maybe add a link to this from the plugin section as well.

### DIFF
--- a/site/weave-docker-api.md
+++ b/site/weave-docker-api.md
@@ -3,30 +3,40 @@ title: Integrating Docker via the API Proxy
 menu_order: 50
 ---
 
-
 The Docker API proxy automatically attaches containers to the Weave
 network when they are started using the ordinary Docker
 [command-line interface](https://docs.docker.com/reference/commandline/cli/)
 or the [remote API](https://docs.docker.com/reference/api/docker_remote_api/),
 instead of `weave run`.
 
+
 ###Attaching Containers to a Weave Network
 
-There are two ways to attach containers to a Weave network, 
-the first uses the Weave Net Docker API Proxy, and is described in this section.  
-The other method uses the Docker Network Plugin framework. It’s your choice which method to use. 
+There are three ways to attach containers to a Weave network (which method to use is 
+entirely up to you):
 
-To use the Docker Network Plugin, launch Weave with the --net flag.  Note that after 
-you’ve launched Weave Net with this flag, then the Weave Docker API Proxy is automatically disabled.  
+ * The Weave Net Docker API Proxy. See [Setting Up the Weave Net Docker API Proxy](#weave-api-proxy).  
 
-For more information about using the Network Plugin, see 
-[Integrating Docker via the Network Plugin](plugin.md).  
+ * The Docker Network Plugin framework. The Docker Network Plugin is used when 
+Docker containers are started with the --net flag, for example: 
 
-For more information about using the Weave Docker 
-API proxy then read on. 
+`docker run --net <docker-run-options>`
+
+**Where,** 
+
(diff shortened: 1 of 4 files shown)


site/weave-docker-api.md
```diff
@@ -10,6 +10,21 @@ network when they are started using the ordinary Docker
...
 instead of `weave run`.
 
+###Attaching Containers to a Weave Network
+
+There are two ways to attach containers to a Weave network, 
+the first uses the Weave Net Docker API Proxy, and is described in this section.  
+The other method uses the Docker Network Plugin framework. It’s your choice which method to use. 
+
+To use the Docker Network Plugin, launch Weave with the --net flag.  Note that after 
+you’ve launched Weave Net with this flag, then the Weave Docker API Proxy is automatically disabled.
```
rade: The `--net` flag gets specified in `docker run`, i.e. when starting containers. Not when launching weave.

Also, specifying `--net` does *not* "use the Docker Network Plugin". It disables automatic attachment to the Weave network by the proxy. Nothing more. What happens networking-wise to such a container is completely up to Docker and what exactly you specified in `--net`.

Finally, there is another option that has the same effect: `-e WEAVE_CIDR=none`. It's mentioned in a few places in the docs. I reckon that option and the explanation about `--net` should be consolidated into one place.

site/weave-docker-api.md
```diff
@@ -10,6 +10,21 @@ network when they are started using the ordinary Docker
 or the [remote API](https://docs.docker.com/reference/api/docker_remote_api/),
 instead of `weave run`.
 
+###Attaching Containers to a Weave Network
+
+There are two ways to attach containers to a Weave network,
```
rade: Three. The third is mentioned right above, i.e. `weave run`.

site/weave-docker-api.md
```diff
@@ -10,6 +10,21 @@ network when they are started using the ordinary Docker
...
+The other method uses the Docker Network Plugin framework. It’s your choice which method to use. 
+
+To use the Docker Network Plugin, launch Weave with the --net flag.  Note that after 
+you’ve launched Weave Net with this flag, then the Weave Docker API Proxy is automatically disabled.  
+
+For more information about using the Network Plugin, see 
+[Integrating Docker via the Network Plugin](plugin.md).  
+
+For more information about using the Weave Docker 
+API proxy then read on.
```
rade: "For more information ... then read on." doesn't read well.

abuehrle: amended

site/weave-docker-api.md
abuehrle: amended

site/weave-docker-api.md
abuehrle: > Finally, there is another option that has the same effect: -e WEAVE_CIDR=none. It's mentioned in a few places in the docs. I reckon that option and the explanation about --net should be consolidated into one place.

Should we put this together with the --net flag explanation in their own place or should I add the -e WEAVE_CIDR=none option here to the --net flag point as well.  

It seems to me that this whole section could be put higher up in the docs or at least a link to it, so that people can read about this right away.  thoughts?

rade: Not sure. The *intention* expressed by `--net blah` and `WEAVE_CIDR=none` is quite different. In the former case the proxy "gets out of the way" since a user specifying `--net blah` probably wants to use Docker networking (including, but not limited to, using weave via docker networking, through the plugin). In the latter, the user explicitly *instructs* the proxy to get out of the way.

rade:
This really should have gone on 1.6.

PR #2468, opened 2016-08-03
Remove unused functions from the weave script

brb:


brb:
Pushing directly to master.

PR #2471, opened 2016-08-08
CNI plugin assigns an IP to the bridge by itself

bboreham:
No need to call `weave expose` before use.  This simplifies set-up for Kubernetes, etc.

It only does if the config doesn't specify any routes or gateway, which may be too restrictive, but I guess we can wait for customer feedback.

CC @abuehrle for review of doc change

### DIFF
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -40,6 +40,19 @@ func loadNetConf(bytes []byte) (*NetConf, error) {
 	return n, nil
 }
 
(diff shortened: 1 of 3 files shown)


abuehrle:
LGTM

bboreham:
We do still need the fuller `weave expose` in at least one case:

When a pod contacts a service on another node, it will start the conversation sending to the service address, which will be DNAT-ted to a node's real IP address.  When the packet arrives at that host, it will arrive on the underlay network interface, and be dropped as a martian.

So, we need the masquerade rules on the source host, so when the packet leaves the Weave network it changes its return address to the underlay IP address of the source host.

PR #2474, opened 2016-08-10
Check if a vxlan vport netdev is UP after its creation

### LINKED ISSUE #2375: Creation of vxlan-vport in fastdp does not fail if udp:6784 is taken
To reproduce:

```
$ weave reset
$ nc -l -u -p 6784 &
$ weave launch
```

In the case of failure, we would expect to see a warning https://github.com/weaveworks/weave/blob/v1.5.2/router/fastdp.go#L405 , but it doesn't happen.

It might be a case that the creation of vxlan-vport is async in the kernel, because no error is returned. However, in the dmesg report we can spot it:
```
[175079.718300] vxlan: Cannot bind port 6784, err=-98
```

Kernel version: 4.5.4-1-ARCH

brb:
Tested on 4.6.

Fixes #2375

### DIFF
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -9,6 +9,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/vishvananda/netlink"
 	"github.com/weaveworks/go-odp/odp"
 	"github.com/weaveworks/mesh"
 )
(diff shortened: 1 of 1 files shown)


router/fastdp.go
```diff
@@ -418,12 +419,28 @@ func (fastdp *FastDatapath) getVxlanVportID(udpPort int) (odp.VportID, error) {
...
+		odp.NewVxlanVportSpec(name, uint16(udpPort)))
 	if err != nil {
 		return 0, err
 	}
 
+	// If a netdev for the vxlan vport exists, we need to do an extra check
+	// to bypass the kernel bug which makes the vxlan creation to complete
+	// successfully regardless whether there were any errors when binding
+	// to the given UDP port.
+	if link, err := netlink.LinkByName(name); err == nil {
```
awh: Are there legitimate circumstances under which the netdev would not exist at this point? I'm wondering if failure to find the device should not be a fatal error...

brb: Yep, e.g. on kernel < 4.3.

PR #2475, opened 2016-08-10
Add a safer version of WithNetNS*

### LINKED ISSUE #2419: Fix the usage of WithNetNS* functions

brb:
This PR introduces a safe version of `WithNetNS*`. The new function executes `work()` by shelling out `weaveutil` which before executing it enters a given network namespace via `nsenter`.

Also, the PR replaces the usage of `WithNetNS*Unsafe` with the newly introduced function where needed.

Fixes #2419


brb:
PTAL.

net/veth.go
```diff
@@ -105,21 +105,18 @@ const (
...
-		_, err := netlink.LinkByName(ifName)
-		return err
-	})
+func interfaceExistsInNamespace(nsPath string, ifName string) bool {
+	_, err := WithNetNS(nsPath, "check-iface", ifName)
 	return err == nil
 }
 
-func AttachContainer(ns netns.NsHandle, id, ifName, bridgeName string, mtu int, withMulticastRoute bool, cidrs []*net.IPNet, keepTXOn bool) error {
+func AttachContainer(ns netns.NsHandle, nsPath, id, ifName, bridgeName string, mtu int, withMulticastRoute bool, cidrs []*net.IPNet, keepTXOn bool) error {
```
rade: it's rather odd that this takes both a ns and nsPath.

rade: This also introduces an asymmetry with DetachContainer.

bboreham:
I squashed the commits in this PR, in branch issues/2419-withnetns
If #2541 lands I will rebase on top of that

brb:
PTAL. A few things:

* I'm in favor of getting rid of all occurrences of`WithNetNS*Unsafe` as unaware users might trip over it.
* `ParseNetDev` is ugly. Open to suggestions.
* If moving `NetDev` pieces from `common/` to `net/` clogs reviewing, I will remove the corresponding commit from this PR.

prog/weaveutil/netdev.go
```diff
@@ -33,3 +94,39 @@ func configureARP(args []string) error {
 
 	return nil
 }
+
+// listNetDevs outputs network ifaces identified by the given indexes
+// in the format of weavenet.Dev.
+func listNetDevs(args []string) error {
```
awh: I think we should:
* Pass the indices as separate args so you can ditch the `Split`
* Don't bother de-duplicating - the list always comes from `ConnectedToBridgeVethPeerIds` anyway, which can't generate dupes
* Add annotations to the NetDev structure, and print as JSON - that'll fix your ParseNetDev issue at the same time

net/netdev.go
```diff
@@ -0,0 +1,159 @@
...
+	Name  string
+	MAC   net.HardwareAddr
+	CIDRs []*net.IPNet
+}
+
+func (d Dev) String() string {
+	return fmt.Sprintf("%s %s %s", d.Name, d.MAC, d.CIDRs)
+}
+
+func ParseNetDev(netdev string) (Dev, error) {
```
awh: As mentioned elsewhere, this (and the associated regexp/`Dev.String()`) can go away if `list-netdevs` prints JSON

net/netdev.go
```diff
@@ -0,0 +1,159 @@
...
+
+	peersStr := make([]string, len(peerIDs))
+	for i, id := range peerIDs {
+		peersStr[i] = strconv.Itoa(id)
+	}
+	netdevsStr, err := WithNetNSByPid(processID, "list-netdevs", strings.Join(peersStr, ","))
+	if err != nil {
+		return nil, fmt.Errorf("list-netdevs failed: %s", err)
+	}
+	for _, netdevStr := range strings.Split(netdevsStr, "\n") {
```
awh: This can be simplified if `list-netdevs` prints JSON encoded `Dev` structures

bboreham:
We don't really need to change `addrs.go` or `proc-addrs.go`; they concentrate all their namespace-changing in one region.  And if you don't do that then a lot of the rest of the complexity goes away.

prog/weaveutil/netdev.go
brb: Yep, sounds good!

brb:
As mentioned above, I'm in favor of getting rid of `WithNetNS*Unsafe` for safety reasons. Therefore I made the changes to `addrs.go` and `proc-addrs.go`.

brb:
PTAL.

PR #2477, opened 2016-08-12
Refactor: move bridge veth creation into 'init_bridge'

bboreham:
This simplifies the code and makes `weave create-bridge` more useful

The key change is that the veths are now created before `weaver` is running instead of afterwards, but I can't see that this would be important.  This is already the case if you stop/start `weaver`.


bboreham:
Wrong branch

PR #2491, opened 2016-08-26
Switch CircleCI build VM to 14.04

brb:
The 14.04 VM includes go1.6 which uses `vendor/`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ machine:
   environment:
     GOPATH: /home/ubuntu:$GOPATH
     SRCDIR: /home/ubuntu/src/github.com/weaveworks/weave
-    PATH: $PATH:$HOME/.local/bin:$HOME/google-cloud-sdk/bin/
+    PATH: $PATH:$HOME/.local/bin
     CLOUDSDK_CORE_DISABLE_PROMPTS: 1
 
 dependencies:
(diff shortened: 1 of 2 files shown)


rade:
This is not the first time that joffrey/docker-py tests fail. In the past we fixed things w/o pinning. Given that the purpose of the tests is to check that the proxy is faithfully handling all docker traffic - regardless of version - I rather think pinning is a bad idea.

brb:
The pinning to 1.8.1 is required because the latest version of joffrey/docker-py tests requires 1.24 Docker API (=> 1.12.x Docker). Unfortunately, 1.12.x Docker is very unreliable and hangs from time to time during a run of the smoke tests.

I've created a separate issue for switching back to joffrey/docker-py:latest and 1.12.1 Docker: https://github.com/weaveworks/weave/issues/2499

rade:
>  the latest version of joffrey/docker-py tests requires 1.24 Docker

@martynas That is a bug in docker-py. The docker-py tests are annotated with version information, so [only tests for the right version should be run](https://github.com/docker/docker-py/blob/1.9.0/tests/base.py#L17).

site/operational-guide/tasks.md
```diff
@@ -52,7 +52,7 @@ following command on any peer:
     weave status ipam
 
 This command displays the peer names and nicknames, absolute quantity/percentage of allocation 
-range managed by peer and also identifies the names of unreachable peers. If you are satisifed
+range managed by peer and also identifies the names of unreachable peers. If you are satisfied
```
rade: This is completely unrelated to this PR

rade:
> We should *not* be pinning the joffrey/docker-py image.

Hmm. The docker-py tests really ought to work out the server version by themselves. Alas I cannot find any code in docker-py that actually does that. I was convinced we had run into this sort of problem before, but have not been able to find an issue/pr in weave or docker-py for it, so I must have been imagining it.

I am perplexed what the purpose of the version annotations in the tests is when there is no way to actually run the tests with a version other than DEFAULT_DOCKER_API_VERSION, which is set to 1.24 in the latest docker-py.

brb:
https://github.com/docker/docker-py/issues/1183

PR #2498, opened 2016-09-02
fixed broken link to command line args

abuehrle:


brb:
Thanks!

Just one question: what is a motivation to have the "see also" link to the plugin docs in the "Allocating IP Addresses" page?

bboreham:
Maybe there used to be a reference to picking IP subnets using the plugin, but there isn't now, so I would just remove that link.

bboreham:
I rebased the branch to get latest CI build fixes

PR #2501, opened 2016-09-05
Reject empty peers

### LINKED ISSUE #2495: Launching router with an empty peer name results in a failure
```
$ weave launch ""
$ docker logs weave 2>&1 | head -n 15
INFO: 2016/08/31 16:08:20.872894 Command line options: map[nickname:fitzroy port:6783 resolv-conf:/var/run/weave/etc/resolv.conf http-addr:127.0.0.1:6784 ipalloc-range:10.32.0.0/12 name:1a:b8:95:98:72:98 datapath:datapath dns-effective-listen-address:172.21.0.1 dns-listen-address:172.21.0.1:53]
INFO: 2016/08/31 16:08:20.874107 Communication between peers is unencrypted.
INFO: 2016/08/31 16:08:20.919338 Our name is 1a:b8:95:98:72:98(fitzroy)
INFO: 2016/08/31 16:08:20.919573 Launch detected - using supplied peer list: []
INFO: 2016/08/31 16:08:20.920425 Docker API on unix:///var/run/docker.sock: &[GoVersion=go1.7 Os=linux Arch=amd64 KernelVersion=4.7.1-1-ARCH BuildTime=2016-08-19T02:03:02.073976990+02:00 Version=1.12.1 ApiVersion=1.24 GitCommit=23cf638]
INFO: 2016/08/31 16:08:20.920722 [allocator 1a:b8:95:98:72:98] No valid persisted data
INFO: 2016/08/31 16:08:20.951758 [allocator 1a:b8:95:98:72:98] Initialising via deferred consensus
INFO: 2016/08/31 16:08:20.953497 Listening for DNS queries on 172.21.0.1
INFO: 2016/08/31 16:08:20.953836 Sniffing traffic on datapath (via ODP)
INFO: 2016/08/31 16:08:20.959840 ->[:6783] attempting connection
INFO: 2016/08/31 16:08:20.960059 ->[127.0.0.1:53465] connection accepted
INFO: 2016/08/31 16:08:20.964729 ->[127.0.0.1:6783|1a:b8:95:98:72:98(fitzroy)]: connection shutting down due to error: cannot connect to ourself
[signal 0xb code=0x1 addr=0x0 pc=0x550a93]
```

A user has experienced the issue while launching weave as a systemd unit in which a var was empty at the end of weave launch cmd: https://weaveworks.slack.com/archives/weave-users/p1472658927000099

brb:
Fix #2495


brb:
Just realized that the same check is needed for `weave connect`, so maybe adding it to the mesh library makes more sense?

prog/weaver/main.go
```diff
@@ -575,3 +575,15 @@ func checkFatal(e error) {
 		Log.Fatal(e)
 	}
 }
+
+func filterEmpty(list []string) []string {
+	filtered := list[:0]
```
rade: Initialising `filtered` like this will end up modifying the underlying `list` array. Which seems way dodgy to me. https://play.golang.org/p/-wmI6yGLdV

rade:
> so maybe adding it to the mesh library makes more sense?

This is a user input sanitisation issue s doesn't really belong into a library; the mesh library should be able to assume it operates on clean data.

rade:
I am not even convinced we should be doing any filtering here.

Where does the panic in #2495 originate?

rade:
Seems to me that the empty address should be rejected in `connection_maker.InitiateConnections`, just like it rejects other invalid addresses.

However, first I'd like to understand where/why the panic arises.

brb:
Because https://github.com/weaveworks/mesh/blob/master/connection_maker.go#L89 does not check whether a host has been supplied.

rade:
but according to the stack trace you posted, it fails in [connectionTerminated](https://github.com/weaveworks/mesh/blob/master/connection_maker.go#L195).

brb:
```
connectionMaker.InitiateConnections(...)
   -> addrs[""] = ":0"
   -> request actor loop
        -> cm.directPeers[""] = ":0"
        -> connectionMaker.checkStateAndAttempConnections(...)
             -> cm.targets[":6783"] = &target{...}
             -> try to establish a connection (to itself)
                  -> connection fails => LocalConnection.teardown(...)
                       -> connectionMaker.connectionTerminated(...)
                            -> cm.targets["$IP:$PORT"] returns nil which makes it to panic
```

rade:
Right. So still it ...
> Seems to me that the empty address should be rejected in `connection_maker.InitiateConnections`, just like it rejects other invalid addresses.

brb:
Yes.

brb:
https://github.com/weaveworks/mesh/pull/58

brb:
Need to update mesh submodule after https://github.com/weaveworks/mesh/pull/58 gets merged.

awh:
@brb I have merged the mesh PR - update at will.

PR #2509, opened 2016-09-13
Create restart.sentinel after storing the peer list

### LINKED ISSUE #2503: Peer list passed as a param might be lost if initialization of router fails
The initialization of the router (`prog/weaver/main.go:func main()`) does the following:

1. `router.InitialPeers(..)`: restores a peer list if "restart.sentinel" exists; otherwise, it creates the sentinel.
2. <..>
3. `router.InitiateConnections(..)`: stores the peer list.

In the case of any failure at step 2 (e.g. `log.Fatal` is called) and `--restart=always` being used when starting the router container, the peer list is lost, because after a restart, step 1 restores an empty list of peers.

EDITED: This is only relevant for a case when launching "weave" container fails and `wait_for_status` from the script gets executed after the container has been successfully restarted.

brb:
Fix #2503

### DIFF
--- a/router/network_router.go
+++ b/router/network_router.go
@@ -239,12 +239,16 @@ func (router *NetworkRouter) InitialPeers(resume bool, peers []string) ([]string
 		return storedPeers, nil
 	}
 
+	log.Println("Launch detected - using supplied peer list:", peers)
+	return peers, nil
+}
+
+func (router *NetworkRouter) CreateRestartSentinel() error {
 	sentinel, err := os.Create("restart.sentinel")
 	if err != nil {
-		return nil, fmt.Errorf("error creating sentinel: %v", err)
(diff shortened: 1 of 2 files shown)


router/network_router.go
```diff
@@ -239,12 +239,16 @@ func (router *NetworkRouter) InitialPeers(resume bool, peers []string) ([]string
 		return storedPeers, nil
 	}
 
+	log.Println("Launch detected - using supplied peer list:", peers)
+	return peers, nil
+}
+
+func (router *NetworkRouter) FinishInitiateConnections() error {
```
awh: Can we rename this `CreateRestartSentinel`? It doesn't really have anything to do with connection initiation

brb: Sure!

Issue #2506, opened 2016-09-07
Cannot use uts.Release[:] (type []uint8) as type []int8


larmog:
I'm trying to build weave `1.6.1` for `armhf` and get this compile error: 
<pre>
make: Entering directory '/go/src/github.com/weaveworks/weave'
go build -i -ldflags "-extldflags \"-static\" -X main.version=git-11f56031f150" -tags netgo -o prog/weaver/weaver ./prog/weaver
# github.com/weaveworks/weave/prog/weaver
prog/weaver/checkpoint.go:37: cannot use uts.Release[:] (type []uint8) as type []int8 in argument to charsToString
</pre>

See the attached build script how to reproduce.
[build.sh](https://github.com/weaveworks/weave/files/460342/build.sh.txt)

2opremio:
I don't have an arm machine around to reproduce with your script but it happens because [Utsname](https://golang.org/pkg/syscall/#Utsname) is architecture dependent.

The following quick patch should get you going for now (note that this will break x86 and we would need a more robust way if/when upstreaming it).

```patch
diff --git a/prog/weaver/checkpoint.go b/prog/weaver/checkpoint.go
index fac1b08..840af8d 100644
--- a/prog/weaver/checkpoint.go
+++ b/prog/weaver/checkpoint.go
@@ -50,7 +50,7 @@ func checkForUpdates(dockerVersion string, network string) {
 	checker = checkpoint.CheckInterval(&params, updateCheckPeriod, handleResponse)
 }
 
-func charsToString(ca []int8) string {
+func charsToString(ca []uint8) string {
 	s := make([]byte, len(ca))
 	i := 0
 	for ; i < len(ca); i++ {
```

As you have probably figured out by now, we don't officially support ARM but we thank you for looking into it and might consider upstreaming your changes once you get it to a working state.

errordeveloper:
> ...we would need a more robust way if/when upstreaming it

I think you'd simply need to use build tags and have separate definitions of the func.

rade:
why not simply inline the function? then no type signature is needed.

PR #2517, opened 2016-09-20
Supports both []int8 and []uint8 from uts.Release[:].

larmog:
Solves #2506


rade:
In https://github.com/weaveworks/weave/issues/2506#issuecomment-248435313 I suggested to simply inline the function, which would be simpler.

rade:
Replaced by #2538.

PR #2524, opened 2016-09-23
Implement `--expect-npc` flag to `create-bridge`

awh:
Add a private flag for weave-kube to steer overlay traffic via the network policy engine chain.

### DIFF
--- a/weave
+++ b/weave
@@ -597,9 +597,17 @@ create_bridge() {
         add_iptables_rule filter INPUT -i $DOCKER_BRIDGE -p udp --dport 53  -j ACCEPT
         add_iptables_rule filter INPUT -i $DOCKER_BRIDGE -p tcp --dport 53  -j ACCEPT
 
-        # Work around the situation where there are no rules allowing traffic
-        # across our bridge. E.g. ufw
-        add_iptables_rule filter FORWARD -i $BRIDGE -o $BRIDGE -j ACCEPT
+        if [ "$2" = "--expect-npc" ] ; then # matches usage in weave-kube launch.sh
+            # Steer traffic via the NPC
+            run_iptables -N WEAVE-NPC >/dev/null 2>&1 || true
+            add_iptables_rule filter FORWARD -o $BRIDGE -j WEAVE-NPC
+            add_iptables_rule filter FORWARD -o $BRIDGE -j LOG --log-prefix=WEAVE-NPC:
+            add_iptables_rule filter FORWARD -o $BRIDGE -j DROP
+        else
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -597,9 +597,17 @@ create_bridge() {
         add_iptables_rule filter INPUT -i $DOCKER_BRIDGE -p udp --dport 53  -j ACCEPT
         add_iptables_rule filter INPUT -i $DOCKER_BRIDGE -p tcp --dport 53  -j ACCEPT
 
-        # Work around the situation where there are no rules allowing traffic
-        # across our bridge. E.g. ufw
-        add_iptables_rule filter FORWARD -i $BRIDGE -o $BRIDGE -j ACCEPT
+        if [ "$2" = "--expect-npc" ] ; then
```
bboreham: Specifying exactly argument 2 seems fragile.  At least it deserves a comment that it matches the usage in `weave-kube`

awh: The root problem is trying to do all this complex flag parsing in bash. The function is already fragile - a bit later on:
```
[ "$1" = "--without-ethtool" -o -n "$AWSVPC" ] || ...
```

awh:
This needs a little more thought - reassigning to me.

awh:
I've added a comment and removed the physdev matching as discussed - PTAL @bboreham

weave
bboreham: I still think the comment is needed - `--without-ethtool` can be found within the same file so is easier to track down.

weave
```diff
@@ -597,9 +597,17 @@ create_bridge() {
...
         add_iptables_rule filter INPUT -i $DOCKER_BRIDGE -p tcp --dport 53  -j ACCEPT
 
-        # Work around the situation where there are no rules allowing traffic
-        # across our bridge. E.g. ufw
-        add_iptables_rule filter FORWARD -i $BRIDGE -o $BRIDGE -j ACCEPT
+        if [ "$2" = "--expect-npc" ] ; then # matches usage in weave-kube launch.sh
+            # Steer traffic via the NPC
+            run_iptables -N WEAVE-NPC >/dev/null 2>&1 || true
+            add_iptables_rule filter FORWARD -o $BRIDGE -j WEAVE-NPC
+            add_iptables_rule filter FORWARD -o $BRIDGE -j LOG --log-prefix=WEAVE-NPC:
```
bboreham: do we want to ship with a LOG line?

awh: It hasn't been left in accidentally - I added it recently to help end-users design their policies after trying to do same for the microservices demo

PR #2529, opened 2016-09-28
Build container FROM golang:1.7.3

### LINKED ISSUE #2522: Upgrade build container to use golang 1.7.1
Still on 1.5.2 atm!

awh:
Fixes #2522.

### DIFF
--- a/proxy/malformed_host_override.go
+++ b/proxy/malformed_host_override.go
@@ -1,0 +1,122 @@
+// (C) Docker, Inc.
+// Redistributed under the Apache License 2.0
+
+package proxy
+
+import "net"
+
+// MalformedHostHeaderOverride is a wrapper to be able
+// to overcome the 400 Bad request coming from old docker
+// clients that send an invalid Host header.
+type MalformedHostHeaderOverride struct {
+	net.Listener
+}
+
+// MalformedHostHeaderOverrideConn wraps the underlying unix
+// connection and keeps track of the first read from http.Server
+// which just reads the headers.
+type MalformedHostHeaderOverrideConn struct {
+	net.Conn
+	first bool
+}
+
+var closeConnHeader = []byte("\r\nConnection: close\r")
+
+// Read reads the first *read* request from http.Server to inspect
+// the Host header. If the Host starts with / then we're talking to
+// an old docker client which send an invalid Host header. To not
+// error out in http.Server we rewrite the first bytes of the request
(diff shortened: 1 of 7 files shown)


awh:
`golint` appears to have sprouted some additional checks which we need to address (or disable)

awh:
See https://github.com/weaveworks/weave/issues/2522#issuecomment-252272045 before merging

awh:
I have tested this successfully with Docker 1.6.0 -> 1.12.3 inclusive, to make sure the malformed host header fix works as intended:

```
for v in 1.9.0 1.9.1 1.10.0 1.10.1 1.10.2 1.10.3 1.11.0 1.11.1 1.11.2 1.12.0 1.12.1 1.12.2 1.12.3 1.8.0 1.8.1 1.8.2 1.8.3 1.7.1 1.6.2 1.6.1 1.6.0; do wget https://get.docker.com/builds/Linux/x86_64/docker-$v.tgz; done
```

```
for file in *.tgz; do tar -x -z --one-top-level -f $file; done
```

```
for d in $(find -type f -name docker); do sudo $d -H=unix:///var/run/weave/weave.sock version > /dev/null  || echo $d failed; done
```

proxy/malformed_host_override.go
```diff
@@ -0,0 +1,122 @@
...
+		if len(buf) == 0 {
+			return c, nil
+		}
+		return len(buf), nil
+	}
+	return l.Conn.Read(b)
+}
+
+// Accept makes the listener accepts connections and wraps the connection
+// in a MalformedHostHeaderOverrideConn initilizing first to true.
```
marccarre: Minor typo: `initilizing` -> `initializing`

awh: Apart from the header I added, I copied this verbatim from dockers repo - I'd rather not make modifications to it

proxy/malformed_host_override.go
```diff
@@ -0,0 +1,122 @@
+// (C) Docker, Inc.
+// Redistributed under the Apache License 2.0
+
+package proxy
+
+import "net"
+
+// MalformedHostHeaderOverride is a wrapper to be able
+// to overcome the 400 Bad request coming from old docker
+// clients that send an invalid Host header.
```
awh: 1. That sounds about right
2. Yes - this will happen as a by-product of running the tests with different docker versions

PR #2545, opened 2016-10-10
Embed newer Docker client and claim to be older version

### LINKED ISSUE #2395: bump embedded docker version
Now that we have some usage stats on docker versions, we should consider bumping the version of the embedded docker, as a way to ameliorate #582.

As I am writing this, the embedded docker is 1.6.2, and usage stats indicate that 1.9+ has a 98% share. Need to keep an eye on it though.

bboreham:
Claim it operates at older API version to allow use of older daemons.
Build also changes as the binary changed location inside the tarfile.

Fixes #2395


bboreham:
Note I fixed a [test failure](https://circleci.com/gh/weaveworks/weave/6838) by adding the `DOCKER_VERSION` hack to the cni runner scripts.

This was needed because the net-calls-ipam sequence invoked the baked-in docker binary; users will only see it if they make changes to the CNI config file

awh:
As discussed on Slack, this is not safe to merge without testing against all the historic versions of Docker that we claim to support. Other work takes priority => I am moving this out to overflow along with its companion issue.

luxas:
I think this can be closed due to that #2713 merged and includes a similar fix, a bump to v1.10.3

bboreham:
This PR also exported `DOCKER_API_VERSION`, effectively lying to the Docker daemon about the version in use, which avoids limiting users to newer versions.

Also the 1.10 client lacks some features which 1.11 supports; if we can get the `DOCKER_API_VERSION` trick to work then it's better for users.  Probably we can ship an even newer client.

bboreham:
Found a blocker to using this trick with Docker 1.13 - https://github.com/moby/moby/issues/29040.

bboreham:
Closing due to inactivity

PR #2547, opened 2016-10-12
Prometheus metrics

### LINKED ISSUE #2523: Weave Net should raise an event when a connection drops
"raise an event" meaning something that can be captured by, e.g., Prometheus

### LINKED ISSUE #2535: Expose metrics
Following discussions on Slack:
We'd like to be able to monitor our Weave Net, so that we know that all is well. Preferrably, we'd like this exposed as a prometheus metrics endpoint.

Some suggested metrics:
- Number of connections, active and total, per type (sleeve/fastdp)
- Number of locally active ips (possibly split between container allocations and `expose` allocations?)
- IPAM allocation "fullness"
- Active flows (Would be interesting to know how this could be broken down?)
- Bytes in/out

It would be interesting to maintain the source/dest information in the last two, but from a prometheus perspective it might lead to generating too many timeseries?

bboreham:
Fixes much of #2535
Also fixes #2523, when used with Prometheus with the appropriate alert configured

Note that the style of this implementation is to take the existing 'status' data and report on the metrics therein it via Prometheus data formats, rather than to re-code the collection of data using Prometheus objects.

Currently implemented Weave metrics:

````
# HELP weave_bytes Number of bytes transferred.
# TYPE weave_bytes counter
weave_bytes{flow="total"} 896
# HELP weave_connections Number of peer-to-peer connections.
# TYPE weave_connections gauge
weave_connections{state="established"} 3
weave_connections{state="total"} 3
# HELP weave_dns_entries Number of DNS entries.
# TYPE weave_dns_entries gauge
weave_dns_entries{state="local"} 1
weave_dns_entries{state="total"} 1
# HELP weave_flows Number of FastDP flows.
# TYPE weave_flows gauge
weave_flows 4
# HELP weave_ips Number of IP addresses.
# TYPE weave_ips gauge
weave_ips{state="local-used"} 4
weave_ips{state="total"} 1.048576e+06
# HELP weave_packets Number of packets transferred.
# TYPE weave_packets counter
weave_packets{flow="total"} 12
````

Notably missing is a packet or byte counter for Sleeve traffic

### CODE REVIEW

File: prog/weaver/metrics.go [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+					//uint64Counter(m.packetsDesc, flow.Packets, tag)
+					//uint64Counter(m.bytesDesc, flow.Bytes, tag)
+					totalPackets += flow.Packets
+					totalBytes += flow.Bytes
+				}
+			}
+		}
+	}
+	intMetric(m.flowsDesc, flows)
+	uint64Counter(m.bytesDesc, totalBytes, "total")
```
brb: Prometheus [doc](https://prometheus.io/docs/concepts/metric_types/#counter) says that the counter metric cannot be used to represent a value which can go down. At the same time, we periodically flush flows (which includes stats related to them) which means it is possible `totalBytes[now-1]` to be greater than `totalBytes[now]`.

prog/weaver/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	reg.MustRegister(prometheus.NewProcessCollector(os.Getpid(), ""))
+	reg.MustRegister(newMetrics(router, allocator, ns))
+	return promhttp.HandlerFor(reg, promhttp.HandlerOpts{})
+}
+
+type metrics struct {
+	router              *weave.NetworkRouter
+	allocator           *ipam.Allocator
+	ns                  *nameserver.Nameserver
+	connectionCountDesc *prometheus.Desc
```
brb: Have you considered storing the descriptors in a list of structs which one field would be responsible for returning associated metric values?

nameserver/status.go
```diff
@@ -17,11 +17,7 @@ type EntryStatus struct {
 	Tombstone   int64
 }
 
-func NewStatus(ns *Nameserver, dnsServer *DNSServer) *Status {
-	if dnsServer == nil {
-		return nil
-	}
-
+func NewNSStatus(ns *Nameserver) *Status {
```
brb: I find the naming a bit confusing here. IMHO, `NewNSStatus` implies a nameserver status which I expect to  include upstream dns servers, ttl, etc.

prog/weaver/main.go
```diff
@@ -336,10 +338,21 @@ func main() {
...
+		muxRouter.Methods("GET").Path("/metrics").Handler(metricsHandler(router, allocator, ns))
 		http.Handle("/", common.LoggingHTTPHandler(muxRouter))
 		Log.Println("Listening for HTTP control messages on", httpAddr)
-		go listenAndServeHTTP(httpAddr)
+		go listenAndServeHTTP(httpAddr, nil)
+	}
+
+	if statusAddr != "" {
+		muxRouter := mux.NewRouter()
+		HandleHTTPStatus(muxRouter, version, router, allocator, defaultSubnet, ns, dnsserver)
```
brb: Why do we want to serve the status on two sockets? Wouldn't it make sense to serve only metrics on the latter socket (that would require s/statusAddr/metricsAddr)?

vendor/github.com/weaveworks/mesh
```diff
@@ -1 +1 @@
-Subproject commit 3e1ae95e2359c0828318fc0a593f2d08e3091c96
+Subproject commit 80411e92f3acb085a75a9eba545550e8f79de34c
```
brb: I've merged the required mesh change.

prog/weaver/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/weaveworks/weave/ipam"
+	"github.com/weaveworks/weave/nameserver"
+	"github.com/weaveworks/weave/net/address"
+	weave "github.com/weaveworks/weave/router"
+)
+
+func metricsHandler(router *weave.NetworkRouter, allocator *ipam.Allocator, ns *nameserver.Nameserver) http.Handler {
+	reg := prometheus.NewRegistry()
```
rade: Why? AFAIK we are not doing that in any of our other prometheus integrations, which simply use `prometheus.MustRegister(...)`

bboreham: Because that results in a shedload of low-level Go metrics cluttering up the data.

tomwilkie: Those metrics have proven quite useful for other services.

prog/weaver/metrics.go
bboreham: Prometheus understands that counter values sometimes get flushed, e.g. by the process restarting.  So long as this does not happen too close to the sampling interval, it's fine.

prog/weaver/http.go
```diff
@@ -118,6 +111,26 @@ var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 	"trimSuffix": strings.TrimSuffix,
 })
 
+func countDNSEntries(entries []nameserver.EntryStatus) int {
```
bboreham: The design for `status` was to put just data-collecting spread around the various modules and to put all summarising and aggregation in `http.go`.  We could change that, but should probably discuss it with @awh

prog/weaver/metrics.go
bboreham: The purpose of this exercise is not to enable end-users to troubleshoot the coding of Weave Net but to enable them to see how their application is behaving on the network.

rade: What is the *harm* in exposing the stock metrics? Presumably there is enough structure to be able to clearly differentiate between generic metrics and weave net specific metrics, no?

bboreham: The harm is 3x extra data traversing the network, going into Prometheus storage, etc.

I wouldn't call them "generic"; they are things like `go_memstats_mcache_inuse_bytes`, `go_memstats_mcache_sys_bytes`, `go_memstats_mspan_inuse_bytes`, etc.

tomwilkie:
Prometheus is incredibly efficient at storing these: samples, on average,
cost 1.3 bytes each (for a 64-bit timestamp and a 64-bit float).  Metadata
(names, labels) are stored once.

On the wire, these things are encoded as protos, which also dedupes a bunch
of stuff like metric name.

Overall cost is very low.  OTOH, a custom registry makes the typical
prometheus instrumentation case (with global variables) harder, and a lot
more code.

bboreham:
> typical prometheus instrumentation case (with global variables)

That is impractical here: we have already collected the data and just want to report it differently.

> a lot more code

I count one line creating the registry and one registering the process metrics.

bboreham:
I think we should exclude connect-to-self from the "terminations" count

rade:
weave-npc only uses the standard registry.

rade:
Can we please stop arguing about what metrics to include here. *One* metric suffices. Get this merged, and *then* figure out what other metrics to add.

prog/weaver/metrics.go
bboreham: Have done something like this.

prog/weaver/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+			"weave_dns_entries",
+			"Number of DNS entries.",
+			[]string{"state"},
+			prometheus.Labels{},
+		),
+	}
+}
+
+func (m *metrics) Collect(ch chan<- prometheus.Metric) {
+	intMetric := func(desc *prometheus.Desc, val int, labels ...string) {
```
bboreham: Done

prog/weaver/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	intMetric(m.connectionCountDesc, len(routerStatus.Connections)-established, "non-established")
+	intMetric(m.connectionCountDesc, established, "established")
+	uint64Counter(m.terminationDesc, uint64(routerStatus.TerminationCount))
+
+	flows := 0
+	var totalPackets, totalBytes uint64
+	if diagMap, ok := routerStatus.OverlayDiagnostics.(map[string]interface{}); ok {
```
bboreham: Done

prog/weaver/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+		if fastDPEntry, ok := diagMap["fastdp"]; ok {
+			if fastDPStatus, ok := fastDPEntry.(weave.FastDPStatus); ok {
+				flows = len(fastDPStatus.Flows)
+				for _, flow := range fastDPStatus.Flows {
+					//tag := flowTag(flow)
```
bboreham: done

ipam/space/space.go
```diff
@@ -88,6 +88,14 @@ func (s *Space) Claim(addr address.Address) error {
 	return nil
 }
 
+func (s *Space) NumOwnedAddresses() address.Count {
```
bboreham: done

router/overlay_switch.go
```diff
@@ -50,6 +52,15 @@ func (osw *OverlaySwitch) Diagnostics() interface{} {
 	return diagnostics
 }
 
+func (d OverlaySwitchDiagnostics) Stats() *FastDPMetrics {
```
brb: Ghmm, I'd prefer to keep `OverlaySwitch` as generic as possible, i.e. to not include any details about a particular overlay implementation.

bboreham:
Squashed commits; note it now needs https://github.com/weaveworks/mesh/pull/63

bboreham:
Just to note that the flow stats we are adding up reset periodically.  So the sum will not really make sense.

PR #2549, opened 2016-10-14
Log dropped policy controller traffic via NFLOG

awh:
Send blocked NEW connection attempts to NFLOG group 86 for accounting by the weave-npc process; required by weaveworks/weave-npc#7.

### DIFF
--- a/weave
+++ b/weave
@@ -601,7 +601,7 @@ create_bridge() {
             # Steer traffic via the NPC
             run_iptables -N WEAVE-NPC >/dev/null 2>&1 || true
             add_iptables_rule filter FORWARD -o $BRIDGE -j WEAVE-NPC
-            add_iptables_rule filter FORWARD -o $BRIDGE -j LOG --log-prefix=WEAVE-NPC:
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -601,7 +601,7 @@ create_bridge() {
             # Steer traffic via the NPC
             run_iptables -N WEAVE-NPC >/dev/null 2>&1 || true
             add_iptables_rule filter FORWARD -o $BRIDGE -j WEAVE-NPC
-            add_iptables_rule filter FORWARD -o $BRIDGE -j LOG --log-prefix=WEAVE-NPC:
+            add_iptables_rule filter FORWARD -o $BRIDGE -m state --state NEW -j NFLOG --nflog-group 86
```
brb: What is group 86?

awh: An arbitrary group identifier, falling in the permitted range 0-65535, to keep our log messages separate from other users of NFLOG on the system. I chose 86 because of https://en.wikipedia.org/wiki/86_(term). Good point though - this needs documenting! I will add a comment on the documentation issue.

PR #2551, opened 2016-10-17
Merge weave-kube into the Weave Net repo

bboreham:
This PR brings in code and artefacts from the https://github.com/weaveworks/weave-kube repo.

It also re-arranges the Docker image dependencies. Previously we had:

```
scratch <- weave
scratch <- alpine <- weaveexec <- plugin
```

Now we have:

```
scratch <- alpine <- weaveexec <- weave <- plugin <- weave-kube
```

This ensures that `weave-kube` has the `weaver` and `plugin` files it needs, while avoiding any double-copying of binaries.  It does mean that the `weave` container looks much bigger than before.

Note that `kube-peers` is kept as a separate executable rather than merging into `weaveutils`, since it would make the `weaveutils` binary 4x larger and hence (probably) slower to start up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ NETGO_CHECK=@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
 	echo "    sudo go install -tags netgo std"; \
 	false; \
 }
-BUILD_FLAGS=-i -ldflags "-extldflags \"-static\" -X main.version=$(WEAVE_VERSION)" -tags netgo
+BUILD_FLAGS=-i -ldflags "-linkmode external -extldflags -static -X main.version=$(WEAVE_VERSION)" -tags netgo
 
 PACKAGE_BASE=$(shell go list -e ./)
 
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ WEAVE_VERSION=git-$(shell git rev-parse --short=12 HEAD)
 WEAVER_EXE=prog/weaver/weaver
 WEAVEPROXY_EXE=prog/weaveproxy/weaveproxy
 SIGPROXY_EXE=prog/sigproxy/sigproxy
+KUBEPEERS_EXE=prog/kube-peers/kube-peers
 WEAVEWAIT_EXE=prog/weavewait/weavewait
 WEAVEWAIT_NOOP_EXE=prog/weavewait/weavewait_noop
 WEAVEWAIT_NOMCAST_EXE=prog/weavewait/weavewait_nomcast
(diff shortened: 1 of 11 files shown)


bboreham:
The new source files in this PR are up to date with weave-kube commit 05cad0449020f04524cb02f1c61054afc2e48c5e

prog/weaver/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM scratch
+FROM weaveworks/weaveexec
```
awh: Needs the `sed` treatment

Makefile
```diff
@@ -55,7 +55,7 @@ NETGO_CHECK=@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
 	echo "    sudo go install -tags netgo std"; \
 	false; \
 }
-BUILD_FLAGS=-i -ldflags "-extldflags \"-static\" -X main.version=$(WEAVE_VERSION)" -tags netgo
+BUILD_FLAGS=-i -ldflags "-linkmode external -extldflags -static -X main.version=$(WEAVE_VERSION)" -tags netgo
```
awh: What does this change accomplish, for my edification?

prog/kube-peers/main.go
```diff
@@ -0,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"log"
+
+	"k8s.io/client-go/1.4/kubernetes"
```
awh: `k8s.io/client-go` needs vendoring

prog/weave-kube/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM weaveworks/plugin
```
awh: Needs the `sed` treatment

prog/kube-peers/main.go
bboreham: I believe https://github.com/weaveworks/weave/pull/2551/files#diff-8903239df476d7401cf9e76af0252622R106 should do this.

awh: So it does, sorry - I missed it due to the suppressed diff.

awh:
#2564 implements improved '`sed` treatment` - we should rebase on that once it's merged.

Makefile
awh: Has this been broken since day one then? Why have we only noticed now?

awh:
Rebased on #2564 and added some commits to template the other Dockerfiles (`git rebase master --autosquash` to fold them in)

Makefile
bboreham: There is more to it - without this flag Go uses the external linker for all our programs except `kube-peers`.  The logic to decide which linker to use is rather convoluted - I reckon it was accidentally working from day one.

awh:
LGTM. I will replace the contents of the weave-kube repo with a README redirecting here, and orchestrate migration of the outstanding issues/PRs.

PR #2553, opened 2016-10-19
Use 'bash' image instead of 'weaveworks/ubuntu'

bboreham:
`bash` is `alpine` with the bash shell added.

I'm not entirely happy about the change: the default bash prompt is different from the alpine `sh` prompt, in that it doesn't show the hostname, and this removes some information from some of the examples.

`weaveworks/ubuntu` has a different `nc`, but it looks like all the examples are specified to work the same on `alpine`


bboreham:
Closing to remove clutter because it's not finished; anyone is welcome to clean this up and finish the job.

PR #2554, opened 2016-10-20
ran shfmt -w -i 4 weave

paulbellamy:


mvdan:
Let me know if you guys run into any issues or have any questions!

awh:
If we're going to merge this, we should do it just before the 1.8.0 release to avoid merge conflicts with the 1.7 branch.

awh:
Actually I don't think we should merge this as a one-off - if we're going to do this at all, `shfmt` should be added to the linter so it's kept up to date. Thoughts?

bboreham:
There are a few lines around https://github.com/weaveworks/weave/pull/2554/files#diff-ed6579eaedd0e67b3ed51c7ee7071044R592 and https://github.com/weaveworks/weave/pull/2554/files#diff-ed6579eaedd0e67b3ed51c7ee7071044R790 where this change loses vertical alignment of sub-parts of lines.

Almost all the changes in this PR are essentially just a different opinion about spaces around certain operators.

The real value this tool would bring is if lines get indented wrongly wrt branching logic, leading to errors in understanding.  That seems worth having, and as part of the lint operation.

mvdan:
This tool is a formatter, not a linter, so I don't know if you would be successful at only applying some of its changes or taking them as lint warnings.

In any case, I agree that running this once accomplishes little. The advantage the tool provides is consistency, which should be maintained via CI or something similar.

awh:
> This tool is a formatter, not a linter, so I don't know if you would be successful at only applying some of its changes or taking them as lint warnings.

I had in mind to run `shfmt` in our CI and ensure that a diff against the original was empty.

mvdan:
> I had in mind to run shfmt in our CI and ensure that a diff against the original was empty.

Ah, sure. I understood that @bboreham wanted to just take in the indentation changes, leaving out the rest that break vertical alignment of arguments.

That could be possible if I added an option for it, but then with it the tool would become little more than an indenter.

awh:
Needs more work, not a priority for this sprint => moving to overflow for consideration in 1.9 planning.

bboreham:
To clarify, I was saying the value of correct indenting makes it worth accepting other changes that we either don't care about or lose slightly from.  And it fits in a particular place in our CI where we run lint.

PR #2564, opened 2016-10-27
Improve Dockerfile template substitution

awh:
Encode `$(DOCKERHUB_USER)` into the name of the generated Dockerfile so we only rebuild on change.


brb:
LGTM

PR #2565, opened 2016-10-27
Improve 'status' to handle cases when the checkpoint API is not available

### LINKED ISSUE #2537: weave status wrong when check is blocked
weave status shows "up to date" even when the version check failed because of firewall block rules.

marccarre:
`weave status` now prints:

```
failed to check latest version - see logs
```

when the checkpoint API is not available.
Periodicity of the checks is left unchanged.
Smoke tests based on `iptables` rules added to reproduce the failure.
Fixes #2537.


marccarre:
Fixes 2537.

PR #2572, opened 2016-11-01
WeaveCloud/Cortex integration

brb:
This PR adds a configuration of headless services used by Prometheus to query `weave-{net/npc}` metrics endpoints. The services make an assumption that `weave-{net,npc}` metrics are served on tcp:8685 and tcp:8686 ports, respectively.

Relevant Prometheus configurations (incl. a configmap for queries proposed in weaveworks/service#935) be found at https://gist.github.com/brb/435d940a15b4443f5353aa7424ff469e

Keep in mind, that the current `weave-net` pod of the `weave-net` daemonset does not expose metrics. For testing purpose, you can use `brb0/weave-{npc,kube}:latest` (`sed -i "s/weaveworks/brb0/" prog/weave-kube/weave-daemonset.yaml`). **<- UPDATED:** only applies for `weave-npc`.

Fixes https://github.com/weaveworks/weave-kube/issues/43

### DIFF
--- a/prog/weave-kube/weave-monitoring-svc.yaml
+++ b/prog/weave-kube/weave-monitoring-svc.yaml
@@ -1,0 +1,28 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+   name: weave-monitoring-net
+   namespace: kube-system
+   annotations:
+     prometheus.io.scrape: "true"
+spec:
+   selector:
+      name: weave-net
+   clusterIP: None
+   ports:
+   - port: 6782
+---
+apiVersion: v1
+kind: Service
+metadata:
+   name: weave-monitoring-npc
+   namespace: kube-system
+   annotations:
+     prometheus.io.scrape: "true"
+spec:
+   selector:
+      name: weave-net
+   clusterIP: None
+   ports:
+   - port: 8686


prog/weave-kube/weave-monitoring-svc.yaml
```diff
@@ -0,0 +1,28 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+   name: weave-net-monitoring
+   namespace: kube-system
+   annotations:
+     prometheus.io.scrape: "true"
```
jml: The standard config we have scrapes all services, so this probably won't be needed.

prog/weave-kube/weave-monitoring-svc.yaml
```diff
@@ -0,0 +1,28 @@
...
+spec:
+   selector:
+      name: weave-net
```
jml: FWIW, the [kubelint](https://github.com/weaveworks/service-conf/blob/master/k8s/kubelint) script will bork on this, insisting that the `name` selector match the service name.

This probably isn't an issue for your repo, but I'd recommend keeping consistent with `service-conf` unless you have a compelling reason to do otherwise.

prog/weave-kube/weave-monitoring-svc.yaml
brb: Just to note, this config is also for a case when `remote_write` is used, i.e. scrapping happens on a Prometheus instance running outside Service.

brb:
Thanks for the review!

prog/weave-kube/weave-monitoring-svc.yaml
brb: Changed the name of each service `weave-{net,npc}-monitoring` -> `weave-monitoring-{net,npc}`). Also, pinged the weave-net team about using the kubelint script.

errordeveloper:
Why is this being added here and not in weave-kube?

brb:
weave-kube got merged into weave: https://github.com/weaveworks/weave/pull/2551

bboreham:
> weave-net serves GET /metrics on tcp:8685.

Any particular reason for that port?  I was using 6782 because it's consecutive to our current ports and unassigned in IANA.

Either way, didn't you have to configure that somewhere?  

I'm changing the default in #2576 so you wouldn't have to.

brb:
> Any particular reason for that port? I was using 6782 because it's consecutive to our current ports and unassigned in IANA.

Just a random port. I expected this port to be used only for testing this PR. Feel free to use any.

> Either way, didn't you have to configure that somewhere?

I created the `brb0/weave-kube` image which `launch.sh` passes `--status-addr=:8685` to `weaver`. Also, the port is specified in the service definition.

brb:
This PR becomes obsolete for [Prometheus 1.3](https://groups.google.com/forum/?fromgroups#!topic/prometheus-developers/Z61jI0ZQhVI) as _"pod discovery now generates one target per container port in the pod"_.

jml:
> This PR becomes obsolete for Prometheus 1.3 as "pod discovery now generates one target per container port in the pod".

Can we close it then?

brb:
Closing. @errordeveloper please use the latest version of the manifest: https://gist.github.com/brb/435d940a15b4443f5353aa7424ff469e . Also, please keep in mind that it installs the latest version of Prometheus (v1.3.0; it used to be v1.2.1).

PR #2574, opened 2016-11-01
allow to not run NPC: do not create iptables rules for NPC

pondohva:
We've got ability to run weave-kube as a static pod https://github.com/weaveworks/weave-kube/pull/36. But there is no posibility to run weave-npc as a static pod (static pod can't get secrets from service account).

And weave-kube doesn't work without weave-npc, because of default iptables rules for NPC.

```
-A FORWARD -o weave -j WEAVE-NPC
-A FORWARD -o weave -m state --state NEW -j NFLOG --nflog-group 86
-A FORWARD -o weave -j DROP
```

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -6,6 +6,9 @@ set -e
 IPALLOC_RANGE=${IPALLOC_RANGE:-10.32.0.0/12}
 HTTP_ADDR=${WEAVE_HTTP_ADDR:-127.0.0.1:6784}
 
+# Default for network policy
+EXPECT_NPC=${EXPECT_NPC:-1}
+
 # kube-proxy requires that bridged traffic passes through netfilter
 if [ ! -f /proc/sys/net/bridge/bridge-nf-call-iptables ] ; then
     echo /proc/sys/net/bridge/bridge-nf-call-iptables not found >&2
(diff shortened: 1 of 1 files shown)


prog/weave-kube/launch.sh
```diff
@@ -41,7 +44,11 @@ fi
 
 # Need to create bridge before running weaver so we can use the peer address
 # (because of https://github.com/weaveworks/weave/issues/2480)
-/home/weave/weave --local create-bridge --force --expect-npc
+if [ "${EXPECT_NPC}" = "0" ]; then
+    /home/weave/weave --local create-bridge --force
+else
+    /home/weave/weave --local create-bridge --force --expect-npc
```
bboreham: Stylistically would prefer not to repeat the elements that are the same, but to create another variable and populate it either with `""` or `"--expect-npc"`.  See how `NICKNAME_ARG` is done, for instance.

pondohva:
@bboreham fixed. I don't want to break current behaviour for daemonset, so we have to set EXPECT_NPC=0 to disable NPC. I haven't found yet where I can put any documentation in this repository.

bboreham:
Thanks for the PR!

lukemarsden:
`else` should be `fi` in the diff @bboreham

awh:
^^^ fixed by #2583.

lukemarsden:
also `/home/weave` should be `/home/weave/weave` @awh @bboreham

awh:
^^^ fixed by ac44fa63c8559336b9c029fd85c3ee5535bdd626

pondohva:
@lukemarsden thank you.

PR #2578, opened 2016-11-01
Serve metrics by default, on port 6782

### LINKED ISSUE #2576: Metrics should be on by default
Currently they need to be turned on with something like `--status-addr=0.0.0.0:6782`.

@rade and @pidster have suggested they should be on by default.

bboreham:
Fixes #2576

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -86,7 +86,7 @@ if [ -z "$IPALLOC_INIT" ]; then
 fi
 
 /home/weave/weaver --port=6783 $BRIDGE_OPTIONS \
-     --http-addr=127.0.0.1:6784 --docker-api='' --no-dns \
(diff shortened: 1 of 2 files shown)


prog/weave-kube/launch.sh
```diff
@@ -5,6 +5,7 @@ set -e
 # Default if not supplied - same as weave net default
 IPALLOC_RANGE=${IPALLOC_RANGE:-10.32.0.0/12}
 HTTP_ADDR=${WEAVE_HTTP_ADDR:-127.0.0.1:6784}
+STATUS_ADDR=${WEAVE_STATUS_ADDR:-127.0.0.1:6782}
```
bboreham: Good catch - it was supposed to be 0.0.0.0.  Too much copypasta.

brb:
LGTM.

@awh would you mind choosing a default weave-npc metrics port closer to 6784?

Issue #2557, opened 2016-10-21
Improve and extend Prometheus metrics


brb:
#2547 has introduced Prometheus metrics for weave-net. Some metrics can be improved though. Suggestions include:

1. `weave_{packet,bytes}_total`: remove the `"flow"` label.
2. `weave_{packet,bytes}_total`: add the `"overlay"` label.
3. `weave_{packet,bytes}_total{overlay="fastdp"}`: change calculation to `flushedFlowsTotal + sum(activeFlows)` as flows get flushed periodically . `flushedFlowsTotal` should be updated before flushing flows.
4. `weave_connections_termination_count`: do not include terminations which occur when connecting to itself.
5. Include metrics mentioned in https://github.com/weaveworks/weave/issues/2535

brb:
6. Distribute the metrics code (`prog/weaver/metrics.go`) across parts in the code which metrics are instrumenting (same as logging). Keep only those parts from `prog/weave/metrics.go` which are necessary for actual serving of metrics to Prometheus. The change helps with debugging. Keep in mind, that it'd change the internal metrics collection model from pull to push.
7. Initialize labeled metrics if all label values are known in advance. That prevents from subtle bugs, e.g. dividing two metrics when one value is missing.
8. When exposing a metric with bunch of labels, create a static counter with all labels and collect the metric with only a reference label. When querying, you can join two metrics in order to access the labels.
9. Split up metrics containing a "total" label into two separate metrics, as it simplifies querying and prevents from matching mistakes. E.g.  `weave_dns_entries{state="local"} / ignoring(state) weave_dns_entries{state="total"}` vs `weave_local_dns_entries / weave_dns_entries`.

juliusv:
Additionally, `weave_connection_termination_count` should be `weave_connection_terminations_total`.

brb:
> Additionally, weave_connection_termination_count should be weave_connection_terminations_total.

This is addressed by https://github.com/weaveworks/weave/pull/2568

frittentheke:
@brb  @bboreham I am unsure whether I am supposed to open a new issue. This one is quite old and seems to be "resolved" mostly?

[...]

PR #2579, opened 2016-11-01
Improve metric naming

bboreham:
Addresses some of #2557

### DIFF
--- a/prog/weaver/http.go
+++ b/prog/weaver/http.go
@@ -19,6 +19,8 @@ import (
 	weave "github.com/weaveworks/weave/router"
 )
 
+var allConnectionStates = []string{"established", "pending", "retrying", "failed", "connecting"}
+
 var rootTemplate = template.New("root").Funcs(map[string]interface{}{
 	"countDNSEntries": countDNSEntries,
 	"printList": func(list []string) string {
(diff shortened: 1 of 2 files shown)


prog/weaver/http.go
```diff
@@ -19,6 +19,8 @@ import (
 	weave "github.com/weaveworks/weave/router"
 )
 
+var allConnectionStates = []string{"established", "pending", "retrying", "failed", "connecting"}
```
brb: Nitpicking, but I'd prefer `mesh` to provide the list of all possible connection states. Also, this list does not include the `waiting` state (although the if-branch which creates it is marked with the `"shouldn't happen"` comment).

brb: Probably ignore this, as I've just spotted that the list existed before this PR.

prog/weaver/metrics.go
```diff
@@ -63,15 +61,19 @@ var metrics []metric = []metric{
 	{desc("weave_ips", "Number of IP addresses.", "state"),
 		func(s WeaveStatus, desc *prometheus.Desc, ch chan<- prometheus.Metric) {
 			if s.IPAM != nil {
-				ch <- intGauge(desc, s.IPAM.RangeNumIPs, "total")
 				ch <- intGauge(desc, s.IPAM.ActiveIPs, "local-used")
 			}
 		}},
-	{desc("weave_dns_entries", "Number of DNS entries.", "state"),
+	{desc("weave_max_ips", "Number of IP addresses used by allocator."),
```
brb: I read this description as "# of allocated IPs". Maybe "the size of IP address space" is more accurate?

bboreham: changed

prog/weaver/http.go
bboreham: I agree with your thought; maybe we can do that as part of a more general re-org.

brb:
LGTM.

Re the comment about states, I'm going to file a separate issue for keeping a list of possible improvements.

https://github.com/weaveworks/service/pull/948 needs to be updated.

Issue #2581, opened 2016-11-01
Configuration of /proc/sys/net/bridge/bridge-nf-call-iptables should be optional


marun:
The launch script running weave on kube currently assumes it can manipulate ``/proc/sys/net/bridge/bridge-nf-call-iptables``: 

https://github.com/weaveworks/weave/blob/master/prog/weave-kube/launch.sh#L12-#L18

For my use-case I'm deploying kube on docker-in-docker.  Since docker is already ensuring that bridge-nf-call-iptables is set on the host, and I'd rather not have to avoid mounting /proc/sys/net rw  into the containers, it would be nice to be able to be able to disable this behavior in the launch script via an env variable (e.g. SET_BRIDGE_NF_CALL_IPTABLES, defaulted to true).

PR #2582, opened 2016-11-01
weave-kube: allow /proc/sys/net/bridge/bridge-nf-call-iptables config to be optional

marun:
Targets #2581

cc: @bboreham

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -11,12 +11,11 @@ STATUS_ADDR=${WEAVE_STATUS_ADDR:-0.0.0.0:6782}
 EXPECT_NPC=${EXPECT_NPC:-1}
 
 # kube-proxy requires that bridged traffic passes through netfilter
-if [ ! -f /proc/sys/net/bridge/bridge-nf-call-iptables ] ; then
-    echo /proc/sys/net/bridge/bridge-nf-call-iptables not found >&2
-    exit 1
+if [ -f /proc/sys/net/bridge/bridge-nf-call-iptables ] ; then
+    echo 1 > /proc/sys/net/bridge/bridge-nf-call-iptables
+else
+    echo "/proc/sys/net/bridge/bridge-nf-call-iptables not found" >&2
 fi
-
-echo 1 > /proc/sys/net/bridge/bridge-nf-call-iptables
 
 SOURCE_BINARY=/usr/bin/weaveutil
 VERSION=$(/home/weave/weaver --version | sed -E 's/weave router (.*?)/\1/')


marun:
This will enable the image to be used in an env that doesn't permit sysctl calls, but the yaml at https://git.io/weave-kube won't make use of it.  Is there an alternative to maintaining a fork of that yaml that sets the env var enabled by this pr?

bboreham:
@marun we have some ideas around templating and configuring via parameterised web-service but are also open to suggestions.

rade:
rather than introducing yet another knob that users have to know about, could we not turn the failure into a warning? Or perhaps try to detect the conditions when sysctl is unnecessary.

marun:
@rade I'd be happy with a warning, and that would remove the need for custom yaml.  My requirement is that the launch script be able to work without error on a host that will be compatible (i.e. ``/proc/sys/net/bridge/bridge-nf-call-iptables=1``) but that won't allow modification from the container.  Should I revise this PR accordingly?

rade:
I shall defer to @bboreham here.

marun:
I imagine the issue is that early detection of an incompatible host state may be more important than supporting my niche use-case.  A knob might make more sense after all, since it's not really something general users should have to care about.

That said, is there a way to functionally test for the desired correct bridge-nf-call-iptables state?  Maybe set/detect the configuration via the fs if possible, and fall back to doing the functional check?  It's more complicated, but it would ensure correct behavior under all scenarios.

bboreham:
There was a comparable case in #2504 where we would fail due to inability to disable TX offload, even when it was already off.

awh:
@marun 

> Should I revise this PR accordingly?

Please do so. The behaviour should be:

1) If `/proc/sys/net/bridge/bridge-nf-call-iptables` does not exist, error
2) If `$(cat /proc/sys/net/bridge/bridge-nf-call-iptables)` does not equal `1`, attempt to set it, erroring on failure

marun:
@awh ``/proc/sys/net/bridge/bridge-nf-call-iptables`` does not exist in the environment that I am trying to run the weave daemonset in, but the kernel configuration is nonetheless compatible.   Ideally for me, the absence of ``/proc/sys/net/bridge/bridge-nf-call-iptables`` would either result in a warning instead of an error or the check would be suppressed as per the current proposal.

awh:
@marun ah, I misunderstood - I thought you just didn't have write permissions. In that case:

1) If `/proc/sys/net/bridge/bridge-nf-call-iptables` does not exist, echo a warning message `
﻿⁠⁠⁠⁠Cannot detect bridge-nf support - network policy and iptables mode kubeproxy may not work reliably` to stderr
2) Else if $(cat /proc/sys/net/bridge/bridge-nf-call-iptables) does not equal 1, attempt to set it, erroring on failure

marun:
@awh I think I've done what you asked.  I don't think it's necessary to attempt to set it and err on failure explicitly since ``set -e`` should ensure an error results from any command failure.

awh:
Hi @marun - not quite - to save any more toing and froing I've opened another PR #2659 let me know if it suffices.

PR #2583, opened 2016-11-02
weaveworks/weave-kube:latest broken - fixup! allow to not run NPC: do not create iptables rules for NPC

damoon:


damoon:
this fixes the currently broken docker image weaveworks/weave-kube:latest

bboreham:
Thanks!

PR #2589, opened 2016-11-03
Fix CNI plugin calling weaveutil

### LINKED ISSUE #2587: Weave-kube CNI plugin fails with missing weaveutil
Symptom is `nsenter: failed to execute weaveutil: No such file or directory`

Because weave-kube copies the plugin binary and makes no provision for that other file to exist.

bboreham:
By the cunning tactic of embedding the CNI plugin inside `weaveutil`

Note I force all uses of `WithNetNS` in `weaveutil` to exec the current program (as defined by `/proc/self/exe`); this seemed better than having it sometimes exec `weaveutil` and sometimes self.

Fixes #2587

### DIFF
--- a/prog/weave-kube/Dockerfile.template
+++ b/prog/weave-kube/Dockerfile.template
@@ -1,3 +1,3 @@
-FROM DOCKERHUB_USER/plugin
+FROM DOCKERHUB_USER/weave
 ADD ./launch.sh ./kube-peers /home/weave/
 ENTRYPOINT ["/home/weave/launch.sh"]
--- a/prog/weaveutil/main.go
+++ b/prog/weaveutil/main.go
@@ -4,6 +4,10 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
+
+	weavenet "github.com/weaveworks/weave/net"
 )
 
 var commands map[string]func([]string) error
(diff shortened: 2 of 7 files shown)


weave
```diff
@@ -1031,7 +1031,7 @@ docker run --rm $(docker_run_options) --pid=host -i \\
  -e CNI_VERSION -e CNI_COMMAND -e CNI_CONTAINERID -e CNI_NETNS \\
  -e CNI_IFNAME -e CNI_ARGS -e CNI_PATH \\
  -v /etc/cni:/etc/cni -v /opt/cni:/opt/cni \$EXTRA_VOLUMES \\
- $PLUGIN_IMAGE $COVERAGE_ARGS $2
+--entrypoint=/usr/bin/weaveutil $EXEC_IMAGE $2
```
brb: Why did `$COVERAGE_ARGS` disappear?

bboreham: I thought `weaveutil` is not compiled with code coverage

prog/weaveutil/main.go
```diff
@@ -28,10 +31,24 @@ func init() {
 		"del-iface":              delIface,
 		"setup-iface":            setupIface,
 		"list-netdevs":           listNetDevs,
+		"cni-net":                cniNet,
+		"cni-ipam":               cniIPAM,
 	}
 }
 
 func main() {
+	weavenet.WeaveUtilCmd = "/proc/self/exe" // force re-exec of this binary
```
brb: This trick does not work because `WithNetNS` called inside `weaveutil` will exec `nsenter --net=<..> /proc/self/exe <..>`, where `/proc/self/exe` links to `nsenter`.

bboreham: `filepath.EvalSymlinks()` to the rescue!

brb:
LGTM. Waiting for approval from @awh

prog/weave-kube/Dockerfile.template
```diff
@@ -1,3 +1,3 @@
-FROM DOCKERHUB_USER/plugin
+FROM DOCKERHUB_USER/weave
```
awh: The Makefile needs to be updated to reflect this dependency change

bboreham: done

Issue #2632, opened 2016-11-08
`Element cannot be added` caused by delayed deletion of Succeeded/Failed pods


awh:
_From @awh on October 28, 2016 10:12_

Pods which enter a phase of `Succeeded` or `Failed` are [not immediately destroyed](http://kubernetes.io/docs/user-guide/pod-states/#pod-lifetime):

> In general, pods which are created do not disappear until someone destroys them. This might be a human or a ReplicationController, or another controller. The only exception to this rule is that pods with a PodPhase of Succeeded or Failed for more than some duration (determined by the master) will expire and be automatically reaped.

Since such pods also appear to retain their podIP attribute, they are not withdrawn from any matching label selector ipsets; consequently, if that podIP is reissued to a new container it will cause an error:

```
add pod: ipset [add weave-3a)<C}f{Y(gA%7g3kj(imRlSe 10.36.0.4] failed: ipset v6.29: Elem    ent cannot be added to the set: it's already added
```

The above was deduced from the following logs, supplied by @davidchua (thanks!):

[weave-net.log.txt](https://github.com/weaveworks/weave-npc/files/558064/weave-net.log.txt)
[weave-npc.log.txt](https://github.com/weaveworks/weave-npc/files/558063/weave-npc.log.txt)

_Copied from original issue: weaveworks/weave-npc#26_

awh:
GC of terminated pods is controlled by the `--terminated-pod-gc-threshold` argument to kube-controller-manager ([docs](http://kubernetes.io/docs/admin/kube-controller-manager/)) - these things can potentially hang around for a long time:

> Number of terminated pods that can exist before the terminated pod garbage collector starts deleting terminated pods. If <= 0, the terminated pod garbage collector is disabled. (default 12500)

awh:
I have reproduced the podIP attribute retention, playing around with the following commands: 

```
root@k8s-01:~# kubectl run true --image alpine --restart=Never --command -- true
pod "true" created
root@k8s-01:~# kubectl run false --image alpine --restart=Never --command -- false
pod "false" created
root@k8s-01:~# kubectl get pods -o wide -a
NAME      READY     STATUS      RESTARTS   AGE       IP          NODE
false     0/1       Error       0          12s       10.32.0.3   k8s-01
true      0/1       Completed   0          18s       10.32.0.3   k8s-01
```

Sometimes the IPs are removed after a while, sometimes they seem to persist. I have yet to figure out why.

bboreham:
[...]

PR #2639, opened 2016-11-13
Fix "ipset add" behavior, when the record already exists.

ikatson:
I caught some very nasty bug, where weave-npc containers where dying and restarting all the time on all hosts with this message:
```
time="2016-11-13T00:31:59Z" level=fatal msg="add pod: ipset [add weave-k?Z;25^M}|1s7P3|H9i;*;MhG 10.32.0.5] failed: ipset v6.29: Element cannot be added to the set: it's already added\n: exit status 1"
```

It might happen to be a workaround for another weave bug, which caused it in the first place. I.e. why would weave try to add an ip with "ipset add" if it had done it before.
Nevertheless, it seems safe enough to merge, even if the original bug, which sounds like some race condition, persists.

### DIFF
--- a/npc/ipset/ipset.go
+++ b/npc/ipset/ipset.go
@@ -2,6 +2,7 @@ package ipset
 
 import (
 	"os/exec"
+	"strings"
 
 	"github.com/pkg/errors"
 )
@@ -38,7 +39,11 @@ func (i *ipset) Create(ipsetName Name, ipsetType Type) error {
 }
 
 func (i *ipset) AddEntry(ipsetName Name, entry string) error {
-	return doExec("add", string(ipsetName), entry)
+	err := doExec("add", string(ipsetName), entry)
+	if err != nil && strings.Contains(err.Error(), "Element cannot be added to the set: it's already added") {
+		return nil
+	}
+	return err
 }
 
 func (i *ipset) DelEntry(ipsetName Name, entry string) error {


awh:
Hi @ikatson - thanks for the PR - unfortunately it is _not_ safe.

What you have seen can only happen if you have two pods with the same IP address - which means either that we have allocated the same IP to two different live pods (a very serious condition which merits the crash you have seen) or it is an occurrence of #2632.

Unfortunately the fix for #2632 is probably to introduce reference counting for pod IPs in ipsets - if we simply ignore the duplication on insertion, we will remove the IP prematurely when the exited pod is GC'ed. This will introduce a far more subtle bug where policy rules are not correctly applied, so I am afraid I cannot merge this.

Do you happen to have logs for the npc container that was crashing? If you can send me the full logs from startup, I can diagnose which of the two cases above was the cause in your instance.

ikatson:
Thanks @awh for a great explanation. Unfortunately, I don't have the logs of it any more. If that happens again, I'll know to look at terminated pods, and save the logs to help you debug it.

Any ideas on more graceful handling of this error and potentially similar errors? The whole SDN went down on all nodes because of this bug, not just one pod.

awh:
> Any ideas on more graceful handling of this error and potentially similar errors?

It is pretty likely that you have experienced #2632, which we are addressing as a matter of priority (e.g. in 1.8.1). In other cases where one of our assumptions is violated we will continue to crash as soon as possible as it is much easier to diagnose - obviously though we are aiming to get incidences of these down to 'basically never'!

ikatson:
I started using the official build again, and reproduced the issue. I can confirm this occured because multiple "Completed" pods where assigned the same IP address.

bboreham:
@ikatson we released 1.8.1 yesterday with our fix - if you could try that and report back this would be a useful check.

ikatson:
Hi, Brian, I can't check your build in the next few days, but I compiled
myself after the fix and it works like a charm. Thank you guys for taking
care of this quickly!

PR #2642, opened 2016-11-14
Fix weave-kube launch.sh version usage

bboreham:
The version string can be anything, including the default `"(unreleased version)"`, so must be quoted when used in a shell command-line.

Also we must supply extra args when the program is compiled with code-coverage.


awh:
#### Documentation changes:
modified [troubleshooting](https://dev-weavewww.pantheonsite.io/docs/net/fix-launch-version/troubleshooting/)

PR #2657, opened 2016-11-21
Remove weave bridge if something goes wrong when creating it

bboreham:
Otherwise it is left in a half-created state and the user has to run `weave reset` to continue.

This is particularly important when using weave-kube, where most users don't have the `weave` script available.

### DIFF
--- a/weave
+++ b/weave
@@ -571,7 +571,7 @@ detect_bridge_type() {
     MTU=${WEAVE_MTU:-$(cat /sys/class/net/$BRIDGE/mtu)}
 }
 
-create_bridge() {
+try_create_bridge() {
     if ! detect_bridge_type ; then
         BRIDGE_TYPE=bridge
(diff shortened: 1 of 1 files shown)


weave
```diff
@@ -656,6 +656,14 @@ EOF
     configure_arp_cache $BRIDGE
 }
 
+create_bridge() {
+    if ! try_create_bridge "$@" ; then
+        echo "Creating bridge '$BRIDGE' failed" >&2
+        destroy_bridge
```
awh: Did you read the comment at the top of `destroy_bridge`?

awh:
This is broken IMO, as it violates the constraint specified at the top of `destroy_bridge` (e.g. that `detect_bridge_type` must not have been called)

bboreham:
Restored the constraint and rebased

PR #2658, opened 2016-11-21
Make NPC only look at IP addresses on running pods

bboreham:
This is an improvement on the behaviour reported at #2632 - a complete fix will need to cover the case where the observed lifetimes of two pods overlap.

I am relying on the pod phases being as documented at http://kubernetes.io/docs/user-guide/pod-states/

Empirically this fixes the problem that any 'completed' or 'failed' pods can remain showing their old IP address, that is re-used by a later pod invocation.

### DIFF
--- a/npc/namespace.go
+++ b/npc/namespace.go
@@ -302,8 +302,9 @@ func (ns *ns) deleteNamespace(obj *coreapi.Namespace) error {
 }
 
 func hasIP(pod *coreapi.Pod) bool {
-	// Ensure pod has an IP address and isn't sharing the host network namespace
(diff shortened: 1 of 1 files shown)


bboreham:
Further empirical observations:

On a new pod we see the `PodIP` appear and the `PodStatus` go to `"Running"` in the same update event.  The preceeding status is `"Pending"`.

bboreham:
Latest feeling is that this will be a complete fix for #2632 because Kubernetes will not send us updates for two running pods with the same IP; the previous one will move into non-running phase before we see the new one get the IP.

PR #2659, opened 2016-11-22
Optional bridge-nf detection

### LINKED ISSUE #2581: Configuration of /proc/sys/net/bridge/bridge-nf-call-iptables should be optional

awh:
Fixes #2581 
Replacing #2582


marun:
lgtm

PR #2662, opened 2016-11-23
Unify weave and launch.sh CNI setup

### LINKED ISSUE #2594: Convert `weave setup-cni` to use the standalone weaveutil binary
`weave setup-cni` currently install scripts into `/opt/cni/bin` that invoke the CNI plugin via `docker run`; however the new Kubernetes addon uses a new method that install a standalone binary that doesn't depend on docker. The weave script should be converted to use the same method.

bboreham:
Fixes #2594 

Points to note:
* The symlinks in `/opt/cni/bin` are now relative to that directory (i.e. to `weave-plugin-1.9.0`), not absolute.
* Test 830 has to run the plugin under sudo now.
* As before, `weave` only installs the CNI plugin if the directories `/opt/cni/bin` and `/etc/cni/net.d` exist, whereas `launch.sh` creates the directories it needs.
* The plugin location is overridable via `$WEAVE_CNI_PLUGIN_DIR`

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -22,51 +22,6 @@ else
     fi
 fi
 
-SOURCE_BINARY=/usr/bin/weaveutil
-VERSION=$(/home/weave/weaver $EXTRA_ARGS --version | sed -e 's/weave router //')
-PLUGIN="weave-plugin-$VERSION"
-
-install_cni_plugin() {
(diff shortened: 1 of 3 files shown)


awh:
This LGTM @bboreham; reassigning to you in case you wanted to push the first commit separately or make the script changes discussed on Slack.

bboreham:
I pushed the `-v` fix straight onto master

PR #2663, opened 2016-11-23
Add MTU to connection status

### LINKED ISSUE #2389: Report MTU size in 'weave status connections'
With current implementation, the value would be the same for all fastdp or awsvpc connections, but potentially variable between sleeve connections and between invocations.

bboreham:
Fixes #2389 

I carry the data through Mesh in a map to make it generic; AWSVPC declines to supply it in this implementation.

Sample output:

````
$ weave status connections
<- 192.168.48.1:59366    established sleeve 6a:41:53:e1:d7:29(vagrant-ubuntu-wily-64) mtu=1438
<- 192.168.48.11:42128   established fastdp 5e:7e:09:ac:95:94(host1) mtu=1410
$ weave report
...
        "Connections": [
            {
                "Address": "192.168.48.1:59366",
                "Outbound": false,
                "State": "established",
                "Info": "sleeve 6a:41:53:e1:d7:29(vagrant-ubuntu-wily-64)",
                "Data": {
                    "mtu": 1438
                }
            },
            {
                "Address": "192.168.48.11:42128",
                "Outbound": false,
                "State": "established",
                "Info": "fastdp 5e:7e:09:ac:95:94(host1)",
                "Data": {
                    "mtu": 1410
                }
            }
        ],
...
````

### DIFF
--- a/test/700_status_and_report_2_test.sh
+++ b/test/700_status_and_report_2_test.sh
@@ -9,6 +9,7 @@ start_suite "weave status/report"
 
 
 weave_on $HOST1 launch --ipalloc-range $UNIVERSE --name 8a:3e:3e:3e:3e:3e --nickname nicknamington
+weave_on $HOST2 launch --ipalloc-range $UNIVERSE
 
 check() {
     assert "weave_on $HOST1 status | grep -oP '(?<= $1: ).*'" "$3"
@@ -27,6 +28,9 @@ assert_raises "weave_on $HOST1 report | grep nicknamington"
 weave_on $HOST1 connect 10.2.2.1
 assert "weave_on $HOST1 status targets" "10.2.2.1"
 assert "weave_on $HOST1 status connections | tr -s ' ' | cut -d ' ' -f 2" "10.2.2.1:6783"
+weave_on $HOST1 connect $HOST2
+sleep 1
+assert "weave_on $HOST1 status connections | grep -q 'mtu=1410'"
 
 assert "weave_on $HOST1 report -f '{{.VersionCheck.Enabled}}'" "false"
 assert_raises "weave_on $HOST1 status | grep 'version check update disabled'"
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by brb** [comment]
Left some comments at https://github.com/weaveworks/mesh/issues/64

**Comment by brb** [comment]
@bboreham would you mind adding a test case to `700_status_and_report_test.sh`?

**Comment by rade** [comment]
`GetDisplayNameAndAttrs()` is an odd function name and has an odd signature. Wouldn't it be better to either include the display name as an attribute (so we would only have `GetAttrs()`, or rather `Attrs()`), or have two functions?

**Comment by bboreham** [comment]
@rade two functions were rejected as a source of races.  Name as an attribute is plausible; wdyt @brb ?

**Comment by brb** [comment]
I fine with name as an attribute.

**Comment by bboreham** [comment]
I changed the method to `Attrs()` returning just a map, and adjusted all uses to assume the name is under `"name"`.

I also added a test case; note the test is renamed because it requires two hosts now.

File: test/700_status_and_report_2_test.sh [hunk 1]
```diff
@@ -27,6 +28,9 @@ assert_raises "weave_on $HOST1 report | grep nicknamington"
 weave_on $HOST1 connect 10.2.2.1
 assert "weave_on $HOST1 status targets" "10.2.2.1"
 assert "weave_on $HOST1 status connections | tr -s ' ' | cut -d ' ' -f 2" "10.2.2.1:6783"
+weave_on $HOST1 connect $HOST2
+sleep 1
+assert "weave_on $HOST1 status connections | grep established | tr -s ' ' | cut -d ' ' -f 6" "mtu=1410"
```
brb: Nitpicking: I'd replace `tr` + `cut` with `grep -q`.

brb:
* I've merged the PR in `mesh`. You can update the vendored dep.
* I think we should update docs. In particular: https://www.weave.works/docs/net/latest/troubleshooting/#weave-status-connections

bboreham:
I updated the test again; PTAL

awh:
#### Documentation changes:
modified [troubleshooting](https://dev-weavewww.pantheonsite.io/docs/net/2389-report-mtu/troubleshooting/)

PR #2670, opened 2016-11-24
Run GCE test VMs on a locked-down network

### LINKED ISSUE #2667: CI hosts have open Docker port
So could be subverted, albeit only for a few minutes until they are destroyed

bboreham:
and open up a rule from the specific IP running the tests

I created the `test` network by hand in the console; it has the following rules by default:

````
Name                  Source tag/IP    Allowed protocols/ports
test-allow-icmp       0.0.0.0/0        icmp
test-allow-internal   10.128.0.0/9     tcp:0-65535,udp:0-65535,icmp
test-allow-ssh        0.0.0.0/0        tcp:22
test-allow-weave      test             tcp:6783,udp:6783,udp:6784
````

and this PR adds specific rules for each test runner to open the Docker ports.

After this is merged we should be able to close down the `default` network rules.

Fixes #2667

### DIFF
--- a/test/gce.sh
+++ b/test/gce.sh
@@ -34,6 +34,9 @@ function vm_names {
 
 # Delete all vms in this account
 function destroy {
+	if [ $(gcloud compute firewall-rules list test-allow-docker$SUFFIX 2>/dev/null | wc -l) -gt 0 ] ; then
+		gcloud compute firewall-rules delete test-allow-docker$SUFFIX
+	fi
 	names="$(vm_names)"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by n054** [comment]
Count me in. @n054

**Comment by awh** [comment]
After discussion on Slack, we removed the
```
test-allow-weave      test             tcp:6783,udp:6783,udp:6784
```
rule because the connectivity is already allowed by `test-allow-internal` (as are some other things, such as k8s master/worker comms). Tests pass on rebuild.

PR #2674, opened 2016-11-28
Detect looping flows

### LINKED ISSUE #2650: Weave Net can create looping flow rules when hairpin is turned on on the bridge veth
Analysis from @awh:

Consider the following simple configuration with two containers A and B connected to the weave bridge:
````
  A                     B
  +                     +
  |                     |
  |                     |
+-+---------------------+-+
| 2                     3 |
|                weave    |
| 1                       |
+-+-----------------------+
  |
  |
  |
  |
+-+-----------------------+
| vport1                  |
|               datapath  |
| vxlan vport             |
+-------------------------+

````

Furthermore, imagine that there are no flow rules, and that neither the weave bridge nor the fastdp bridge implementation have learnt any MAC addresses yet. Then:

1. A sends a packet to B
2. The weave bridge learns A=2, and floods the packet out of 1 & 3
3. B receives the packet but does not respond
4. The router miss handler receives the packet on vport1, learns that A=vport1, and floods the packet out of the vxlan vport
5. The weave bridge A=2 entry expires
6. B sends a packet to A
7. The weave bridge learns that B=3, and floods the packet out of 1 & 2
8. The router miss handler receives the packet on vport1, learns that B=vport1, installs a flow 'B->A in vport1 send to vport1', and sends the packet out of vport1
9. The weave bridge receives the B->A packet on 1, and learns that B=1. Depending on whether A responded after 7, it may have learnt A=1 or not; it either floods 1 & 3 or sends to 1 (it doesn't actually matter)
10. A responds to B
11. The weave bridge learnt in 9 that B=1, so it sends to 1
12. The router miss handler receives the packet on vport1, installs a flow 'A->B in vport1 send to vport1', and sends the packet out of vport1
13. The weave bridge receives the A->B packet on 1, and learns that A=1. Forwards packet out of 1
14. The weave bridge now thinks A and B are associated with port 1, and the router thinks (both in terms of MAC table and installed flow rules) that A and B are associated with vport1. Packets circulate continuously.

Step 13 requires hairpin to be turned on; one possible cause of this is https://github.com/kubernetes/kubernetes/issues/19766

### PR DESCRIPTION
**Pull request by brb**
This PR includes the following changes:

* Detect loops in fastdp flows (see commit message for a meaning of loops in terms of fastdp).
* Detect enabling of hairpin mode.
* Logging of a case when own packet is received from a remote peer.

BLOCKERS: https://github.com/vishvananda/netlink/pull/179 (I've temporarily changed the vendored netlink dep).

Fixes #2650

### DIFF
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -1,6 +1,7 @@
 package net
 
 import (
+	"github.com/Sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 )
 
@@ -49,5 +50,33 @@ func isDatapath(link netlink.Link) bool {
 		return true
 	default:
 		return false
+	}
+}
+
+func DetectHairpin(portIfName string, log *logrus.Logger) {
+	link, err := netlink.LinkByName(portIfName)
+	if err != nil {
+		log.Errorf("Unable to find link %q: %s", portIfName, err)
+	}
+
+	ch := make(chan netlink.LinkUpdate)
+	// See EnsureInterface for why done channel is not passed
+	if err := netlink.LinkSubscribe(ch, nil); err != nil {
+		log.Errorf("Unable to subscribe to netlink updates: %s", err)
+	}
+
+	pi, err := netlink.LinkGetProtinfo(link)
+	if err != nil {
+		log.Errorf("Unable to get link protinfo %q: %s", portIfName, err)
+	}
+	if pi.Hairpin {
+		log.Errorf("Hairpin mode enabled on %q", portIfName)
+	}
+
+	for up := range ch {
+		if up.Attrs().Name == portIfName && up.Attrs().Protinfo != nil &&
+			up.Attrs().Protinfo.Hairpin {
+			log.Errorf("Hairpin mode enabled on %q", portIfName)
+		}
 	}
 }
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -50,7 +50,8 @@ type FastDatapath struct {
 	seenMACs map[MAC]struct{}
 
 	// vxlan vports associated with the given UDP ports
-	vxlanVportIDs    map[int]odp.VportID
+	vxlanUDPPorts    map[int]odp.VportID
+	vxlanVportIDs    map[odp.VportID]struct{}
 	mainVxlanVportID odp.VportID
 
 	// A singleton pool for the occasions when we need to decode
@@ -87,7 +88,8 @@ func NewFastDatapath(iface *net.Interface, port int) (*FastDatapath, error) {
 		sendToPort:    nil,
 		sendToMAC:     make(map[MAC]bridgeSender),
 		seenMACs:      make(map[MAC]struct{}),
-		vxlanVportIDs: make(map[int]odp.VportID),
+		vxlanUDPPorts: make(map[int]odp.VportID),
+		vxlanVportIDs: make(map[odp.VportID]struct{}),
 		forwarders:    make(map[mesh.PeerName]*fastDatapathForwarder),
 	}
 
@@ -433,7 +435,7 @@ func (fastdp *FastDatapath) getVxlanVportID(udpPort int) (odp.VportID, error) {
 	fastdp.lock.Lock()
 	defer fastdp.lock.Unlock()
 
-	if vxlanVportID, present := fastdp.vxlanVportIDs[udpPort]; present {
+	if vxlanVportID, present := fastdp.vxlanUDPPorts[udpPort]; present {
 		return vxlanVportID, nil
 	}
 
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: router/fastdp.go [hunk 1]
```diff
@@ -1171,6 +1179,45 @@ func (fastdp *FastDatapath) takeDecoder(lock *fastDatapathLock) *EthernetDecoder
...
+func (fastdp *FastDatapath) checkFlowLoop(flow *odp.FlowSpec) bool {
+	var vxlanKey odp.TunnelAttrs
+
+	inVports := make(map[odp.VportID]struct{})
+	inVxlan := false
+
+	for _, key := range flow.FlowKeys {
+		switch k := key.(type) {
+		case odp.InPortFlowKey:
+			inVports[k.VportID()] = struct{}{}
```
awh: Trying to figure out why this is a map - under what circumstances will there be more than one `InPortFlowKey`?

router/fastdp.go
```diff
@@ -1171,6 +1179,45 @@ func (fastdp *FastDatapath) takeDecoder(lock *fastDatapathLock) *EthernetDecoder
 	return dec
 }
 
+// A checkFlowLoop checks whether the flow is free from the following loops:
+// * in_port == out_port, where in_port is non-vxlan vport;
+// * a packet is sent back to a vxlan tunnel it has been received from.
+func (fastdp *FastDatapath) checkFlowLoop(flow *odp.FlowSpec) bool {
```
awh: I think we should rename this function:

* It's not obvious from the name under which conditions it returns true (or false)
* The flowspec on its own doesn't contain a loop

How about `flowReflectsPackets`?

router/fastdp.go
```diff
@@ -1134,6 +1137,11 @@ func (fastdp *FastDatapath) send(fops FlowOp, frame []byte, lock *fastDatapathLo
 		}
 	}
 
+	if !fastdp.checkFlowLoop(&flow) {
+		log.Error("ODP flow contains a loop ", flow)
```
awh: See comment on the function name. How about `Vetoed creation of hairpin flow`?

router/fastdp.go
```diff
@@ -1171,6 +1179,45 @@ func (fastdp *FastDatapath) takeDecoder(lock *fastDatapathLock) *EthernetDecoder
...
+	for _, action := range flow.Actions {
+		switch a := action.(type) {
+		case odp.OutputAction:
+			if _, ok := inVports[a.VportID()]; ok {
+				if _, ok := fastdp.vxlanVportIDs[a.VportID()]; !ok {
+					return false
+				}
+			}
+		case odp.SetTunnelAction:
+			if inVxlan && a.TunnelAttrs.TunnelId == vxlanKey.TunnelId &&
```
awh: Do you not need to swap the two halves of the VNI (e.g. the src/dst short peer IDs) before making this comparison?

router/fastdp.go
```diff
@@ -1171,6 +1179,45 @@ func (fastdp *FastDatapath) takeDecoder(lock *fastDatapathLock) *EthernetDecoder
...
+		case odp.TunnelFlowKey:
+			inVxlan = true
+			vxlanKey = k.Key()
+		}
+	}
+
+	for _, action := range flow.Actions {
+		switch a := action.(type) {
+		case odp.OutputAction:
+			if _, ok := inVports[a.VportID()]; ok {
```
awh: This is a bit subtle, so worth a comment IMO - the only time it's valid for a packet to go back out the same vport is when it's a vxlan vport (and only then if the tunnel is different too, which is dealt with by the next case)

router/fastdp.go
```diff
@@ -1171,6 +1179,45 @@ func (fastdp *FastDatapath) takeDecoder(lock *fastDatapathLock) *EthernetDecoder
...
+			inVports[k.VportID()] = struct{}{}
+		case odp.TunnelFlowKey:
+			inVxlan = true
+			vxlanKey = k.Key()
+		}
+	}
+
+	for _, action := range flow.Actions {
+		switch a := action.(type) {
+		case odp.OutputAction:
```
awh: I'd probably switch these two cases around, because the actions are ordered, and the set tunnel action always comes first

router/fastdp.go
brb: Good spot. IMO, it's not possible.

router/fastdp.go
brb: Agreed.

router/fastdp.go
brb: Sounds good.

router/fastdp.go
brb: I was referring to the case when `relay`ing a packet, i.e. the packet's final dst is known, so the VNI does not change.

However, there is a second case, which I did not consider, when we do `relayBroadcast`, i.e. the dst peer changes, so the VNI does as well.

I suggest to modify the condition by replacing it with `(a.TunnelAttrs.TunnelId == vxlanKey.TunnelId || a.TunnelAttrs.TunnelId == reverse(vxlanKey.TunnelId)`.

router/fastdp.go
brb: OK.

router/fastdp.go
brb: OK.

net/bridge.go
```diff
@@ -51,3 +52,31 @@ func isDatapath(link netlink.Link) bool {
...
+		log.Errorf("Unable to find link %q: %s", portIfName, err)
+	}
+
+	ch := make(chan netlink.LinkUpdate)
+	// See EnsureInterface for why done channel is not passed
+	if err := netlink.LinkSubscribe(ch, nil); err != nil {
+		log.Errorf("Unable to subscribe to netlink updates: %s", err)
+	}
+
+	pi, err := netlink.LinkGetProtinfo(link)
```
bboreham: it seems to me that this will crash if `link` is invalid; we should `return` if we hit the first error

brb:
PTAL.

PR #2679, opened 2016-12-01
Run iptables commands in a loop

bboreham:
in case they get  "temporarily unavailable" exit status

### DIFF
--- a/net/veth.go
+++ b/net/veth.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/coreos/go-iptables/iptables"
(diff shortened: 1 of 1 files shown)


net/veth.go
```diff
@@ -272,19 +273,37 @@ func subnets(addrs []netlink.Addr) map[string]struct{} {
...
 }
 
+func addNatRule(ipt *iptables.IPTables, rulespec ...string) error {
+	// Loop until we get an exit code other than "temporarily unavailable"
+	for {
+		if err := ipt.AppendUnique("nat", "WEAVE", rulespec...); err != nil {
+			if ierr, ok := err.(*iptables.Error); ok {
+				if status, ok := ierr.ExitError.Sys().(syscall.WaitStatus); ok {
+					// (magic exit code 4 found in iptables source code; undocumented)
+					if status == 4 {
```
awh: `status` is actually a bitfield which combines the exit code with a number of flags (see `main 2 wait`):
```
$ cat test.go 
package main

import "fmt"
import "os/exec"
import "syscall"

func main() {
	err := exec.Command("false").Run()
	fmt.Println(err.(*exec.ExitError).Sys().(syscall.WaitStatus))
	fmt.Println(err.(*exec.ExitError).Sys().(syscall.WaitStatus).ExitStatus())
}
$ go run test.go 
256
1
```
You need to do `status.ExitStatus() == 4` instead.

Issue #2677, opened 2016-11-29
Update documentation for Prometheus Endpoint /metrics


mmethner:
Hi,

thanks for releasing the prometheus endpoint in 1.8.0.
Today my team tried to configure prometheus to scrape weave for metrics, but at first we were only able to scrape from localhost. After searching this repo we found, that its the WEAVE_STATUS_ADDR environment variable, which defaults to 127.0.0.1:6782, what makes weave only serve on localhost.

We set it to serve on eth0 as well to publish our metrics to the underlaying machine and scrape it throughout our cluster. That works fine.

Did we miss the documentation for that?
I only found https://www.weave.works/docs/net/latest/metrics/, were its not mentioned.
If missing yet, wouldnt it be a good idea to update the docs? Can I help with that?

Mathias

bboreham:
Hi @mmethner, I agree 100% with your analysis.

Note that `weave-kube` has a different default, to listen on all interfaces.  We're a little more conservative with the `weave` script.

We'd love it if you added it into the docs - just submit a PR.

Also if you have any thoughts on improvements to the metrics please comment on #2557 or raise a new issue.

mmethner:
@bboreham Ok, thanks for your reply.

I already forked the Repo and found metrics.md. I will submit the PR in the next days.

brb:
Fixed by https://github.com/weaveworks/weave/pull/2680

PR #2680, opened 2016-12-05
#2677 docu, how to publish prometheus metric endpoint on all interfaces

mmethner:
Hi @abuehrle, Hi @bboreham  

I added a short explanation on publish metrics on all interfaces throughout the cluster.

Ticket is https://github.com/weaveworks/weave/issues/2677

It's the first time I submit a PR to an mostly unknown Repo. Do I need to add more information or something?

### DIFF
--- a/site/metrics.md
+++ b/site/metrics.md
@@ -20,6 +20,19 @@ exposed:
 * `weave_dns_entries` - Number of DNS entries.
 * `weave_flows` - Number of FastDP flows.
 
+#### Publish Router Metrics Endpoint
+
+By default, when started via `weave launch`, weave listens on its local 
+interface to serve metrics. To publish your metrics throughout your cluster, 
+e.g. if your prometheus server is installed on a different host machine,
+you need to set `WEAVE_STATUS_ADDR` to your corresponding IP and port.
+Default port is 6782.
+
+`WEAVE_STATUS_ADDR=X.X.X.X:PORT`
+
+You can set `WEAVE_STATUS_ADDR=0.0.0.0:6782` to listen on all interfaces, 
+but be aware, this may expose your metrics to the public internet.
+
 ### Kubernetes Network Policy Controller Metrics
 
 The endpoint address is `localhost:6781`; the following metric is


site/metrics.md
```diff
@@ -20,6 +20,15 @@ exposed:
 * `weave_dns_entries` - Number of DNS entries.
 * `weave_flows` - Number of FastDP flows.
 
+#### Publish Router Metrics Endpoint
+
+By default weave only listens on it's local interface to serve metrics.
```
bboreham: This is true when started from the `weave` script, but not true when started from Kubernetes via `weave-kube`.

Suggest "By default, when started via `weave launch`"

site/metrics.md
```diff
@@ -20,6 +20,15 @@ exposed:
 * `weave_dns_entries` - Number of DNS entries.
 * `weave_flows` - Number of FastDP flows.
 
+#### Publish Router Metrics Endpoint
+
+By default weave only listens on it's local interface to serve metrics. 
+To publish your metrics throughout your cluster, e.g. if
+your prometheus server is installed on a different host machine,
+you need to set `WEAVE_STATUS_ADDR` to listen on all interfaces:
```
bboreham: If a user has multiple interfaces they may not not want to listen on interfaces exposed to the public Internet.  I don't think we want to flatly state they should set it to listen on all interfaces.

site/metrics.md
bboreham: it's -> its

site/metrics.md
```diff
@@ -20,6 +20,15 @@ exposed:
...
+`WEAVE_STATUS_ADDR=0.0.0.0:6782`
```
bboreham: Also they may want to change the port number, if 6782 clashes with another port in use on their machine.

mmethner:
Thanks, that you took the time to help me to be more secure and detailed on this part of the documentation. I updated the text according to your hints.

site/metrics.md
```diff
@@ -20,6 +20,19 @@ exposed:
...
+By default, when started via `weave launch`, weave listens on its local 
+interface to serve metrics. To publish your metrics throughout your cluster, 
+e.g. if your prometheus server is installed on a different host machine,
+you need to set `WEAVE_STATUS_ADDR` to your corresponding IP and port.
+Default port is 6782.
+
+`WEAVE_STATUS_ADDR=X.X.X.X:<IP>`
+
+You can set `WEAVE_STATUS_ADDR=0.0.0.0:6782` to listen on all interfaces, 
+but be aware, this may exposes your metrics to the public internet.
```
bboreham: s/exposes/expose/

site/metrics.md
```diff
@@ -20,6 +20,19 @@ exposed:
...
 
+#### Publish Router Metrics Endpoint
+
+By default, when started via `weave launch`, weave listens on its local 
+interface to serve metrics. To publish your metrics throughout your cluster, 
+e.g. if your prometheus server is installed on a different host machine,
+you need to set `WEAVE_STATUS_ADDR` to your corresponding IP and port.
+Default port is 6782.
+
+`WEAVE_STATUS_ADDR=X.X.X.X:<IP>`
```
bboreham: where you have 'IP' I think we'd call it 'port'; also we use angle brackets to denote something a user has to add within a bigger command so just leave those out.

So: 

    ```
    WEAVE_STATUS_ADDR=X.X.X.X:PORT
    ```

mmethner:
Im glad you took the time to point me to the right directions. 
Im aware, that you could easily have changed the docu by yourself in two minutes. Thank you.

I even rebase squashed the four commits into one. Never did this before, but it looks like it worked.

bboreham:
Merged via 7ffd96f4254cdabacea656aefb01f20755ae8b09 - it needed to go on the 1.8 branch to update the docs website

bboreham:
Thanks!

PR #2683, opened 2016-12-06
Re-add expose to ensure it gets run on all nodes

### LINKED ISSUE #2673: Kubernetes API cannot reach Kubernetes Dashboard
In my setup I have a master and a node both with a public IP and connected to each other using a VPN. Kubernetes was deployed using Kubeadm and weave was deployed using https://git.io/weave-kube. Kubeadm used the -api-advertise-addresses parameter where only the VPN assigned address for the master was given.

This all works, master and node can see each other on both sides of the VPN. Weave starts and both sides can find each other using the public IPs (eth0). Those are not connected to the same switch.

This gives us the following networking situation:

Master       | Node
------------ | -------------
eth0 (IPv4)  | eth0 (IPv4)
tun0 (VPN)  | tun0 (VPN)
weave          | weave

Next I deployed the Kubernetes Dashboard which got assigned to the node. The dashboard starts and connects to the api deployed on the master.

To access the dashboard I use the following command on the master node:

```
curl -v -L http://127.0.0.1:8080/ui
```

After a while it fails with:

```
Error: 'dial tcp 10.32.0.1:9090: i/o timeout'
```

After some searching I noticed the weave interace on the master node had no address assigned. I had a chat with @bboreham and doing a manual local expose in the weave pod on the master solved my problem. He told me weave will only assign an address if a pod uses the pod network on the specific host. In my case only pods using the host network were deployed and therefore no address was assigned. 

This leads to the problem that the kubernetes api can't use the pod network where the kubernetes dashboard pod is deployed. 

The setup is composed of two Ubuntu 16.04 machines with nothing special. Uname -a lists for both:

```
Linux 4.4.0-47-generic #68-Ubuntu SMP Wed Oct 26 19:39:52 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

I hope this report helps.

bboreham:
Fixes #2673.

Previous change in a07e3301b1a8 meant that if there are no pods on this node then the 'expose' never gets run.

Also I documented the use of `WEAVE_EXPOSE_IP`, from https://github.com/weaveworks/weave-kube/pull/36

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -155,4 +155,7 @@ if [ ! -f /etc/cni/net.d/10-weave.conf ] ; then
 EOF
 fi
 
+# Expose the weave network so host processes can communicate with pods
+/home/weave/weave --local expose $WEAVE_EXPOSE_IP
+
 wait $WEAVE_PID
--- a/site/kube-addon.md
+++ b/site/kube-addon.md
@@ -126,3 +126,6 @@ The list of variables you can set is:
 * IPALLOC\_INIT - set the initialization mode of the [IP Address
   Manager](/site/operational-guide/concepts.md#ip-address-manager)
   (defaults to consensus amongst the KUBE\_PEERS)
+* WEAVE\_EXPOSE\_IP - set the IP address used as a gateway from the
+  Weave network to the host network - this is useful if you are
+  configuring the addon as a static pod.


awh:
#### Documentation changes:
modified [kube-addon](https://dev-weavewww.pantheonsite.io/docs/net/2673-expose-again/kube-addon/)

PR #2685, opened 2016-12-09
Take peerName from weave bridge if not specified

### LINKED ISSUE #2480: weaver takes its identity from wrong mac if run stand-alone
At https://github.com/weaveworks/weave/blob/7ebf3d7a8806abfd5fd4f38e4f1f987992f65f32/prog/weaver/main.go#L227 it takes the name from the `bridge`, but in fastdp mode this is the `datapath` device not `weave` where we have placed our carefully-constructed MAC address.

This behaviour is mitigated by the `weave` script passing in the right name via `--name`, but it makes the program harder to use on its own.

bboreham:
Fixes #2480

Cleans up `launch.sh` slightly, but not as much as when originally filed as weave-npc needs the bridge to be set up explicitly now.

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -22,8 +22,7 @@ else
     fi
 fi
 
-# Need to create bridge before running weaver so we can use the peer address
-# (because of https://github.com/weaveworks/weave/issues/2480)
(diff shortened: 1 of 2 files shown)


brb:
LGTM.

Issue #1737, opened 2015-11-30
[fastdp] can get stuck in 'sleeve' mode even though fastdp is possible


rade:
As things stand, fastdp can only be selected once, during the first `HeartbeatTimeout` seconds of a connection's life. If we fail to establish fastdp connectivity during that time, or if it later fails, then we are stuck in 'sleeve' mode.

notmaxx:
Any plans to implement return from `sleeve` to `fastdp`?

I opened ticket that might be related http://github.com/weaveworks/weave/issues/1737

brb:
Sorry, no plans yet. However, PR is more than welcome!

notmaxx:
@brb regarding, help, do you have any design documents or ideas how it should work correctly? any WIP?

brb:
Sorry, but AFAIK there is no WIP or a design doc. Also, I haven't thought about it much.

https://github.com/weaveworks/weave/blob/v2.2.0/router/overlay_switch.go is responsible for choosing which overlay (in most cases it is either `sleeve` or `fastdp`) to use between two peers.
Each overlay sends a periodic heartbeat messages to detect any failures in an overlay connection. E.g. when `fastdp` times out waiting for a heartbeat message from a peer (https://github.com/weaveworks/weave/blob/v2.2.0/router/fastdp.go#L713) it notifies `overlay_switch` which consequently chooses another best overlay for the connection (https://github.com/weaveworks/weave/blob/v2.2.0/router/overlay_switch.go#L318).

A possible fix is to re-try establishing a connection (with a backoff timer) after it failed due to a missing heartbeat. After it got established, the existing `overlay_switch` implementation will handle the rest.

Cryptophobia:
So basically add a timer to the overlay_switch and re-`chooseBest()` in order to prefer fastdp after a failure until fastdp takes over again as the prefered overlay connection. 

The above sounds like a way to do it. But the unknown question is why fastdp is being dropped on almost every single cluster with weave-net I have seen.

bboreham:
@Cryptophobia I'd recommend opening your own issue with the specifics

Cryptophobia:
@bboreham : There are just too many clusters that have this problem with weave-net. They are all on AWS.  Should I be opening an issue for each cluster. It's not sustainable and will quickly become annoying. 

Also the below issues are being closed as duplicates:

Example issue here: https://github.com/weaveworks/weave/issues/3252

notmaxx:
thx @brb

[...]

PR #2687, opened 2016-12-09
Fastdp crypto

### LINKED ISSUE #1644: Authenticated encryption for fast datapath
The current AE implementation operates in userspace and is therefore incompatible with the fast datapath overlay - an alternative (almost certainly IPsec) is required.

brb:
Missing for the preview release:
- [x] Tests.
- [x] Less "harmful" skb mark.
- [x] Do not store `{local,remote}SAKey` in Mesh connection state.
- [x] Rekeying.

Related branch in Mesh: https://github.com/weaveworks/mesh/tree/issues/1644-fastdp-crypto

Fixes #1644


net/ipsec/ipsec.go
```diff
@@ -0,0 +1,402 @@
...
+				Mode:  netlink.XFRM_MODE_TRANSPORT,
+				Spi:   int(spi),
+			},
+		},
+	}
+}
+
+// Helpers
+
+func newSPI(srcPeer, dstPeer mesh.PeerShortID) (SPI, error) {
```
awh: It's possible for a peer's shortID to change if there is a collision; does that matter?

net/ipsec/ipsec.go
```diff
@@ -0,0 +1,402 @@
...
+	ipsec := &IPSec{
+		ipt: ipt,
+		rc:  newConnRefCount(),
+	}
+
+	return ipsec, nil
+}
+
+// Protect establishes IPsec between given peers.
+func (ipsec *IPSec) Protect(srcPeer, dstPeer mesh.PeerShortID, srcIP, dstIP net.IP, dstPort int, localKey, remoteKey []byte) (SPI, error) {
```
awh: Does this enable replay protection?

net/ipsec/ipsec.go
brb: It should not matter as we are storing SPI in the `fastDatapathForwarder` state. So, e.g. when terminating the connection, `ipsec.Destroy` will receive the correct SPI.

net/ipsec/ipsec.go
brb: Partially, because we do not do key rotation, so Seq Nr (4 bytes) in ESP might overflow which creates a possibility for a replay attack.
Btw, I have updated the doc which includes the sec properties we provide.

rade: > Seq Nr (4 bytes) in ESP might overflow which creates a possibility for a replay attack

How annoying. In weave's user-space crypto we use 64-bit sequence numbers, so neatly avoid this.

net/ipsec/ipsec.go
```diff
@@ -0,0 +1,395 @@
...
+	ipsec := &IPSec{
+		ipt: ipt,
+		rc:  newConnRefCount(),
+	}
+
+	return ipsec, nil
+}
+
+// Protect establishes IPsec between given peers.
+func (ipsec *IPSec) Protect(srcPeer, dstPeer mesh.PeerShortID, srcIP, dstIP net.IP, dstPort int, masterKey *[32]byte) (SPI, error) {
```
awh: Apropos the replay issue with 32 bit sequence number wrapping - how do we feel about setting the `packet-hard` limit in the SA as a stopgap until we implement rekeying? My (untested) hope is that once this limit is reached, xfrm will start dropping dataplane traffic (including heartbeats) causing a control plane reset with new keys.

Time to wrap in minutes on gig ethernet given max (smallest packets)/min (biggest packets) frame rates:

```
2^32/(1488096*60)
48
2^32/(81274*60)
880
```

IOW between 48 minutes and ~36 days.

net/ipsec/ipsec.go
```diff
@@ -0,0 +1,395 @@
...
+		return 0,
+			errors.Wrap(err, fmt.Sprintf("derive SPI (%x, %x)", dstPeer, srcPeer))
+	}
+
+	localKey, remoteKey, err := deriveKeys(masterKey[:])
+	if err != nil {
+		return 0, errors.Wrap(err, "derive keys")
+	}
+	if srcPeer > dstPeer {
+		localKey, remoteKey = remoteKey, localKey
```
awh: Perhaps the HKDF `info` field https://tools.ietf.org/html/rfc5869#section-3.2 is more appropriate for the src/dest ID/IPs? We could also supply the `password` as the salt value.

awh: One final comment - it would be better to use the full peer ID here rather than the short peer ID as the former is immutable...

rade:
Given that this is the first release of this feature, we should probably reserve the right to change details of the protocols. A desirable behaviour here would be that communication between old and new weaves would fall back to user space crypto, just like it would if fastdp crypto wasn't available or not working.

I am hoping that *should* just work. Any gotchas anybody can think of?

awh:
> I am hoping that should just work

I think it will 'just work' (as in we get it for free) in the following cases:

* Old (non-encrypting) fastdp communicating with encrypting fastdp (fastdp overlay won't be available on the older peer)
* Older encrypting fastdp communicating with a newer encrypting fastdp that uses a different cipher scheme (fastdp overlay won't establish due to heartbeat decryption failure)

This case needs more thought:

* Older encrypting fastdp communicating with a newer encrypting fastdp that supports rekeying

as it will require protocol changes (maybe a version bump, but perhaps we can get away with just introducing additional ControlMessage tags)

awh:
User facing docs need updating to reflect that fastdp is no longer disabled with `--password`.

net/ipsec/ipsec.go
brb: After the `packet-hard` limit has been reached, the SA is removed. However, if we keep the policy, the tunneled traffic will be blocked.

There are the following problems with such approach:

1. After missing a heartbeat, the dataplane will be switched to `sleeve` and in the current implementation I don't see a mechanism for trying to switch back to `fastdp`.
2. There is a window between removing SA (i.e., blocking the tunneled traffic) and switching to `sleeve` (missing a heartbeat).

docs/fastdp-crypto.md
```diff
@@ -0,0 +1,67 @@
...
+* Data confidentiality.
+* Data origin authentication.
+* Integrity.
+* Partial anti-replay service (partial - because we do not rotate keys which
+  makes replay attacks possible when Sequence Number in ESP header overflows).
+* Limited traffic flow confidentiality.
+
+# Key derivation
+
+For each connection direction a different key (and salt) is used. The keys
```
brb: The sentence in docs is ambiguous, as it talks about AES-GCM-ESP keys.

net/ipsec/ipsec.go
brb: Hmm. The SPI collisions are harmful only on the recv path, as `xfrm_input`, which handles received ESP packets, does SA lookup with daddr + SPI as a key: https://github.com/torvalds/linux/blob/master/net/xfrm/xfrm_input.c#L247 .

Doing `ip xfrm state allocspi` for inbound-SPIs would work, but such approach requires to exchange some handshake message between peers.

Alternatively, we could use a 32-bit hash function to derive SPIs, but that does not prevent from collisions.

docs/fastdp-crypto.md
```diff
@@ -0,0 +1,67 @@
...
+The kernel VXLAN driver does not set a dst port of a tunnel in the ip flow
+descriptor, thus XFRM policy lookup cannot match a policy which includes
+the dst port. This makes impossible to encrypt only tunneled traffic between
+peers. To work around, we mark such outgoing packets with iptables and set
+the same mark in the policy selector. Funnily enough, iptables_mangle module
+eventually sets the missing dst port in the flow descriptor. The challenge
+here is to pick a mark that it would not interfere with other networking
+applications before OUTPUT'ing a packet. For example, Kubernetes by default
+uses 1<<14 and 1<<15 marks. Additionally, such workaround brings
+the requirement for at least the 4.2 kernel.
```
brb: Noup, as no XFRM policy can be selected. However, installing the iptables rule mentioned above in the TODO comment should make it failsafe.

net/ipsec/ipsec.go
awh: Argh yes, I forgot about that :disappointed: There's an issue for it #1737

docs/fastdp-crypto.md
awh: We should implement that before release then ideally...

net/ipsec/ipsec.go
awh: > The SPI collisions are harmful only on the recv path

Looks like this is enforced early - I just tried to manually create two SAs with the same dst + SPI and received `RTNETLINK answers: File exists` for the second (please confirm this!). This is actually not so bad - in the unlikely event that the circumstance I described arises, the second `Protect` call will return an error and that connection will fall back to sleeve?

> but such approach requires to exchange some handshake message between peers.

Ah yes of course, because you'd need to communicate the SPI you had allocated locally to the other end.

> Alternatively, we could use a 32-bit hash function to derive SPIs, but that does not prevent from collisions.

True. It looks like collisions are fail-safe though (assuming what I've said above is true) - in which case, perhaps we should pick the method that is least likely to produce collisions?

net/ipsec/ipsec.go
awh: @rade thoughts?

rade: Look into fixing #1737. Hopefully it's not that hard, and it's really quite an annoying bug anyway.

net/ipsec/ipsec.go
brb: > please confirm this!

Confirming.

As discussed IRL, we are not going to change the SPI derivation.

brb:
@awh PTAL.

Please keep in mind:

* There are some `TODO(mp)`. Most of them I will address after your review.
* `fastdp-crypto.md` is obsolete. I am working at the moment to fix this.
* `vishvananda/netlink` origin is my repo. I will change after my PRs get merged.
* Some changes in the related branch in `mesh` became obsolete. I will address it.

brb:
I have updated the implementation [docs](https://github.com/weaveworks/weave/blob/issues/1644-fastdp-crypto/docs/fastdp-crypto.md). It should help you to understand the iptables rules.

net/ipsec/ipsec.go
```diff
@@ -0,0 +1,732 @@
...
+
+// IPSec
+
+type IPSec struct {
+	sync.RWMutex
+	ipt *iptables.IPTables
+
+	rc      *connRefCount
+	spiInfo map[spiID]spiInfo
+	spis    map[SPI]*spiInfo
```
brb: This is a reference to `spiInfo` which might be expired, i.e. `spiInfo.spi` has been expired.

net/ipsec/ipsec.go
```diff
@@ -0,0 +1,732 @@
...
+type msgInitSARemote struct {
+	nonce []byte
+	spi   SPI
+}
+
+func deserializeMsgInitSARemote(b []byte) (*msgInitSARemote, error) {
+	if len(b) == 0 {
+		return nil, fmt.Errorf("empty msg")
+	}
+	if vsn := b[0]; vsn != protoVsn {
```
brb: Agreed.

brb:
PTAL @awh

docs/fastdp-crypto.md
```diff
@@ -0,0 +1,154 @@
...
+iptables -t filter -A OUTPUT ! -p esp -m policy --dir out --pol none \
+         -m mark --mark ${MARK} -j DROP
+```
+
+## ESN
+
+To prevent from cycling SeqNo which makes replay attacks possible, we use
+64-bit extended sequence numbers known as [ESN](esn).
+
+## MTU
```
awh: We need to revise the the default fastdp MTU downwards so that crypto works out of the box on GCP, and document the upgrade procedure to include the use of `weave reset` so that bridge devices are recreated with the correct MTU.

net/ipsec/ipsec.go
```diff
@@ -0,0 +1,563 @@
...
+// Destroy destroys any (inbound / outbound) ipsec establishment between the peers.
+func (ipsec *IPSec) Destroy(localPeer, remotePeer mesh.PeerName, connUID uint64, localIP, remoteIP net.IP, udpPort int) error {
+	outSPIID := getSPIId(localPeer, remotePeer, connUID)
+	inSPIID := getSPIId(remotePeer, localPeer, connUID)
+
+	ipsec.Lock()
+	defer ipsec.Unlock()
+
+	// Destroy inbound
+	ipsec.log.Infof(">>>>>>>>>>>>>>>>>> DESTROY <<<<<<<<<<<<<<<<<<<<<<")
```
brb: Oops, unintentionally left.

docs/fastdp-crypto.md
brb: I have reduced the default MTU to address the ESP overhead (34 bytes in our case). Alternatively, we could reduce it only if `weave` detects that the encryption is enabled. However, it would increase complexity of the script.

@bboreham / @rade what do you think?

brb:
@awh PTAL.

The documentation changes will land in a separate PR.

docs/fastdp-crypto.md
bboreham: I would leave it at 1376 for now; if we're going to complicate the script we should default higher and lower it for GCP (or, better, when we see that host adaptors have a low value).

PR #2688, opened 2016-12-09
Exec weaver from weave-kube launch rather than running in background

### LINKED ISSUE #2684: weave-kube weaver process does not always terminate cleanly
Mentioned here https://github.com/weaveworks/weave/issues/2654#issuecomment-263913191:

> The main process is not exiting. weaver is not being killed, and docker kill on the weave contianer is even failing.

One possibility is that the `launch.sh` script is not propagating the kill signal from docker to its children under certain circumstances, leaving the weaver process running but unreachable by `docker kill`. If that is the case, we may investigate the possibility of `exec /home/weave/weaver` in the launch.sh, somehow arranging the expose & reclaim steps to be done afterwards.

bboreham:
So we only have one process running inside the container after all set-up is finished

Fixes #2684

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -65,24 +65,6 @@ if [ -z "$IPALLOC_INIT" ]; then
     IPALLOC_INIT="consensus=$(peer_count $KUBE_PEERS)"
 fi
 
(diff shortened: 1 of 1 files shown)


awh:
Needs rebase. Should this be in the 1.9.0 milestone?

bboreham:
Rebased.  I don't know why this (and the issue) were put in the 1.8 milestone.

Issue #2647, opened 2016-11-15
Epic - Improve development & testing infrastructure


marccarre:
## Summary
Re-factor the current smoke tests in order to:

* better re-use provisioning & configuration "modules" / "APIs"
* run integration tests on top of these layers
* run cross-version tests on top of these layers
* run stress tests on top of these layers
* be able to easily spawn development & manual testing environments.

This should hopefully:
* lead to a nicer development & testing experience 
* boost Weave Net developer productivity
* enabler richer testing scenarios.

```
+------------------+ +-----------------+ +-----------------------+ +---------------------+
| make stress-test | | make cross-test | | make integration-test | | Weave Net developer |
+-------+----------+ +--------+--------+ +--------+--------------+ +--+--------------+---+
        |                     |                   |                   |              |
        | +-------------------+-----------------+ |                   |              |
        | | Cartesian-product of deps. versions | |                   |              |
        | +-------------------+-----------------+ |                   |              |
        |                      \                  |                   |              |
+-------+------+                \        +--------+----------+        |              |
| Stress Tests |                 `-------+ Integration Tests |        |              |
+-------+------+                         +--------+----------+        |              |
        |                                         |                   |              |
+-------+-----------------------------------------+-------------------+---+          |
|          Configuration (Go, Docker, K8S, etc. via Ansible)              |          |
+-------+-----------------------------------------+-------------------+---+          |
        |                                         |                   |              |
+-------+-----------------------------------------+-------------------+--------------+---+
|                Provisioning (VMs via Vagrant, AWS/GCP/DO via Terraform)                |
+----------------------------------------------------------------------------------------+
```

## Progress

- [x] #2696 Scripts now source dependencies using relative paths
- [x] Evaluate [Docker Infrakit](https://github.com/docker/infrakit): does not seem appropriate, more to deploy a `pod`/a set of containers.
- [x] Add new targets to `Makefile`
- [x] #2664 Script to list versions of Go, Docker, Kubernetes 
- [x] Script to orchestrate integration testing
- [x] Script to generate cross product of versions
- [x] #2682 Ansible script to configure test machine with Weave Net (standalone)
- [x] #2682 Ansible script to configure test machine with Weave Net & Kubernetes
- [x] #2676 Vagrant script to add provisioning on local VM
- [x] #2676 Terraform script to add provisioning on GCP, DO and AWS
- [x] Integrate with CircleCI

### Current state
Weave Net is currently integration-tested in two main ways:
[...]

PR #2698, opened 2016-12-14
Finding and running tests no longer depends on current directory.

marccarre:
Currently, depending on your current directory, `make`, tests, etc. may or may not run smoothly.
This is a pre-change refactoring for the #2647 work.


test/840_weave_kube_3_test.sh
```diff
@@ -2,6 +2,8 @@
 
 . "$(dirname "$0")/config.sh"
 
+cd
```
bboreham: what is this for?

marccarre: Definitely a mistake on my end -- removing it now.
Sorry about that and well spotted.

PR #2694, opened 2016-12-13
Testing MVP

marccarre:
Fixes weaveworks/weave#2664.
Fixes weaveworks/weave#2676.
Fixes weaveworks/weave#2682.

See also: 
* weaveworks/weave#2647.
* weaveworks/weave#2665.
* weaveworks/build-tools#52.
* **weaveworks/build-tools#53**.

PR #2702, opened 2016-12-16
Upgrade Vagrantfile to Ubuntu 16.04 'bento' box

bboreham:
We need to upgrade Ubuntu because 'wily' (15.10) is unsupported and so many packages (e.g. Docker RCs) are not released for it.

The 'ubuntu' boxes seem to be rather broken, e.g. create a user called 'ubuntu' instead of 'vagrant', so I moved to the 'bento' box.

As a bonus, we don't need to remove puppet and chef.

I needed to upgrade to Virtualbox 5.1 before it would work.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,7 +5,7 @@ require './vagrant-common.rb'
 vm_ip = "172.16.0.3" # arbitrary private IP
 
 pkgs = %w(
-  docker-engine=1.10.2-0~wily
+  docker-engine=1.10.2-0~xenial
   aufs-tools
   build-essential
   ethtool
@@ -19,7 +19,7 @@ pkgs = %w(
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
-  config.vm.box = "ubuntu/wily64"
+  config.vm.box = "bento/ubuntu-16.04"
 
   config.vm.network "private_network", ip: vm_ip
   config.vm.provider :virtualbox do |vb|
--- a/test/Vagrantfile
+++ b/test/Vagrantfile
@@ -14,7 +14,7 @@ require '../vagrant-common.rb'
 def configure_docker(host, hostname, ip)
   pkgs = %w(docker-engine ethtool)
 
-  host.vm.box = "ubuntu/wily64"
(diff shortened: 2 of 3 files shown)


brb:
Did all tests pass locally?

bboreham:
Yes, except 840 because we don't (yet) install Kubernetes packages and 602 for some Docker-version-related reason.

awh:
The toplevel Vagrantfile is attempting to install docker-engine=1.10.2-0~xenial but has configured the apt docker.list to point at the wily repo. Furthermore, there is no 1.10.2-0~xenial in the correct repo.

marccarre:
FYI: also needed to upgrade `vagrant` from `1.8.1` to `1.9.1` for it to support VirtualBox `5.1`.
It was otherwise failing with the below:

    $ vagrant up
    No usable default provider could be found for your system.
    
    Vagrant relies on interactions with 3rd party systems, known as
    "providers", to provide Vagrant with resources to run development
    environments. Examples are VirtualBox, VMware, Hyper-V.
    
    The easiest solution to this message is to install VirtualBox, which
    is available for free on all major platforms.
    
    If you believe you already have a provider available, make sure it
    is properly installed and configured. You can see more details about
    why a particular provider isn't working by forcing usage with
    `vagrant up --provider=PROVIDER`, which should give you a more specific
    error message for that particular provider.

and if you pass `--provider=virtualbox`:

    $ vagrant up --provider=virtualbox
    The provider 'virtualbox' that was requested to back the machine
    'default' is reporting that it isn't usable on this system. The
    reason is shown below:
    
    Vagrant has detected that you have a version of VirtualBox installed
    that is not supported by this version of Vagrant. Please install one of
    the supported versions listed below to use Vagrant:
    
    4.0, 4.1, 4.2, 4.3, 5.0
    
    A Vagrant update may also be available that adds support for the version
    you specified. Please check www.vagrantup.com/downloads.html to download
    the latest version.

marccarre:
This has been replicated & reworked in #2694, so I believe we should be able to close this PR.

brb:
https://github.com/weaveworks/weave/pull/2694 got merged, so closing this PR.

Issue #2648, opened 2016-11-15
Epic - "Stress test" Weave Net


marccarre:
## Scope

1. Starting very large number of containers wired together using Weave Net.
2. Network faults injections using:
  * [mininet](http://mininet.org/)
  * `tc`
  * `iptables`
  * etc.
3. Hosts/containers killing, a-la Chaos Monkey.
4. All the above with/without Kubernetes.
5. All the above with/without Docker Swarm.

## Dependencies

#2647

PR #2704, opened 2016-12-19
Parallelise setup.sh's individual tasks.

marccarre:
Time spent running the setup steps becomes a blocker when trying to ramp up the number of test servers, given we are currently mainly blocking on I/O -- i.e. many files are uploaded/downloaded, which could easily be done faster.
This is a pre-change refactoring for the #2647 and #2648 work.

### DIFF
--- a/test/setup.sh
+++ b/test/setup.sh
@@ -2,6 +2,7 @@
 
 set -e
 
+begin=$(date +%s)
 cd "$(dirname "${BASH_SOURCE[0]}")"
 
 . "$(dirname "$0")/config.sh"
@@ -59,4 +60,4 @@ for ppid in $ppids; do
     wait $ppid;
 done
 
-echo "Setup completed successfully."
+echo "Setup completed successfully in $(date -u -d @$(($(date +%s)-$begin)) +"%T")."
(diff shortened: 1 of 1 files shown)


test/setup.sh
```diff
@@ -11,27 +12,52 @@ cd "$(dirname "${BASH_SOURCE[0]}")"
...
 echo "  prefetch test images"
 
+exists_on() {
+    docker_on $1 inspect --format=" " $2 >/dev/null 2>&1
+}
+
+pull_if_missing() {
+    if ! exists_on "$1" "$2" ; then
+        echo "Pulling $2 on $1..."
+        docker_on $1 pull $2 >/dev/null 2>&1
```
bboreham: I don't think this should redirect stderr; if something goes wrong with the pull we wan't to know about it.

marccarre: Good point, I'll only redirect `stdout` then. Thanks!

marccarre:
As suggested, I removed the redirect of `stderr` to `/dev/null`. I initially thought of simply removing any noise, but it is true that errors could still happen at this stage, in which case we may want to know what happened.

PR #2707, opened 2016-12-21
Fixed typo in sanity check of weave's version

marccarre:
Previously checking local weave version.
Now checking test host's weave version.


marccarre:
LGTM'ed by Bryan -- see also: https://weaveworks.slack.com/archives/net/p1482314804001585

PR #2710, opened 2016-12-21
Swap HTTP and container checks in wait_for_status.

### LINKED ISSUE #2709: weave can transparently fail if weave-kube is running
#### Steps to reproduce:

* Install `weave`.
* Install K8S and `weave-kube`, **keep `kubelet` running**.
* Run test `130_expose_test.sh`.

#### Symptoms:

* `weave launch --ipalloc-range $UNIVERSE` appears to succeed (it returns `0`), but 
* the `weave` container actually fails to start with (see `docker logs weave`):

```
INFO: 2016/12/21 14:22:39.240455 Command line options: map[[...] ipalloc-range:10.2.3.0/24]
[...]
FATA: 2016/12/21 14:26:04.601824 listen udp4 :6783: bind: address already in use
[...]
```

* test continues to run as if nothing happened, but containers get allocated to the default IP range (`IPRANGE="10.32.0.0/12"`) instead of the desired one (`10.2.3.0/24`), hence the test eventually fails at later stages.

#### Root cause:

Weave currently checks that containers, including the `weave` container, are running using:

```
wait_for_status() {
    container="$1"
    shift
    while true ; do
        "$@" GET /status >/dev/null 2>&1 && return 0           # check #1
        if ! check_running $container >/dev/null 2>&1 ; then   # check #2
            kill_container $container # stop it restarting
            echo $(death_msg $container) >&2
            return 1
        fi
        fractional_sleep 0.1
    done
}
```

In this case, given `weave-kube` is running and listens on port `6784`, `check #1` passes, even though `check #2` would have failed, and `wait_for_status` therefore returns `0`, and the `weave` script therefore believes the container started successfully and continues happily.

#### Potential resolution:

1. Swap `check #1` and `check #2`, which would have allowed the detection of the issue.
2. Identify whether `weave-kube` is running and maybe print a warning?
  K8S somewhat obfuscates container `Names`, but `weave-net` still appears, e.g.:

```
$ docker ps --format "{{.ID}} {{.Image}} {{.Names}}" | grep weave
b39c42a0fe42 weaveworks/plugin:latest weaveplugin
37204e3958a1 weaveworks/weaveexec:latest weaveproxy
84ce83fc34e0 weaveworks/weave:latest weave
7dd83142ac2d weaveworks/weave-kube:1.8.2 k8s_weave.6e1bac7e_weave-net-xhhpw_kube-system_73b14172-c795-11e6-835d-42010a800002_70eb395b
0b3c68800d04 weaveworks/weave-npc:1.8.2 k8s_weave-npc.83aa94cd_weave-net-xhhpw_kube-system_73b14172-c795-11e6-835d-42010a800002_a831cef3
e15c0bb4381c gcr.io/google_containers/pause-amd64:3.0 k8s_POD.d8dbe16c_weave-net-xhhpw_kube-system_73b14172-c795-11e6-835d-42010a800002_d09c50c8
```

marccarre:
Another process may be listening on `6784` and reply incorrectly or with misleading responses, e.g. when both `weave` and `weave-kube` are running and `weave` crashed for some reason.
Swapping the two checks should make things more robust.

Fixes #2709.

### DIFF
--- a/weave
+++ b/weave
@@ -994,12 +994,12 @@ wait_for_status() {
     container="$1"
     shift
     while true ; do
-        "$@" GET /status >/dev/null 2>&1 && return 0
         if ! check_running $container >/dev/null 2>&1 ; then
             kill_container $container # stop it restarting
             echo $(death_msg $container) >&2
             return 1
         fi
+        "$@" GET /status >/dev/null 2>&1 && return 0
         fractional_sleep 0.1
     done
 }


bboreham:
I tried it this way: in one window, I ran, outside of any container:

    weaver --db-prefix /tmp/weave --no-dns --http-addr=127.0.0.1:6784

in another window I ran `weave launch-router`

This change did not fix the problem: it appears that the container had not failed before the first check to see if it was running.

marccarre:
Indeed, and manually testing as described in #2709 also shows simply swapping the two lines does not fix the issue: tests talk to the Weave proxy started by Kubernetes via `exec_remote`/`docker`.

Testing on the host with `weave launch --local --ipalloc-range 10.2.3.0/24` does actually reach `wait_for_status`, but runs the below and fails to detect any issue:

```
check_running weave
http_call 127.0.0.1:6784 GET /status
```

I will try to fix it properly later, once done with the testing MVP.

PR #2713, opened 2016-12-23
Make Weave Net multi-architecture

luxas:
_Will add more description of what this does later..._

TL;DR; This builds images for Weave Net for amd64, arm and arm64 and pushes manifest lists (which makes docker automatically pull the right layers for the specific arch).
How to test:
```
$ DOCKERHUB_USER=luxas_or_whatever WEAVE_VERSION=v1.9.0-alpha.0 make publish
```

Very straightforward IMO :smile:!
This is work in progress...
cc @bboreham


build/Dockerfile
```diff
@@ -1,7 +1,66 @@
...
+		flex \
+		bison
+
+RUN go get \
+	github.com/golang/lint/golint \
+	github.com/fzipp/gocyclo \
+	github.com/client9/misspell/cmd/misspell
+
+RUN go clean -i net \
+	&& go install -tags netgo std \
```
luxas: Is it really necessary to run this twice?
First without race and then with race...

luxas:
Added now a new command:
```
$ DOCKERHUB_USER=luxas_or_whatever make publish-all
```

That will push all images for all arches, and make weave-kube and weave-npc manifest lists so they can be used with multiarch k8s clusters out-of-the-box

build/Dockerfile
bboreham: We use `-race` during CI tests and non`-race` for the final build.  Go puts the library files in different subdirectories so it is valid to run both.  Good question, though!

weave
```diff
@@ -328,15 +300,82 @@ fi
 # main (remote and --local) - settings
 ######################################################################
 
+## Versions ##
+SCRIPT_VERSION="unreleased"
```
bboreham: You cannot move this after the first usage (which is at line 241 in this version)

weave
```diff
@@ -328,15 +300,82 @@ fi
...
+IMAGE_VERSION=latest
+[ "$SCRIPT_VERSION" = "unreleased" ] || IMAGE_VERSION=$SCRIPT_VERSION
+IMAGE_VERSION=${WEAVE_VERSION:-$IMAGE_VERSION}
+
+## Image Names ##
+
+DOCKERHUB_USER=${DOCKERHUB_USER:-weaveworks}
```
bboreham: You cannot move this after the first use (which is in the call to `exec_remote` on line 260)

weave
```diff
@@ -328,15 +300,82 @@ fi
...
+    # amd64 does not use the suffix
+    ARCH_IMAGE_EXT=""
+else
+    # arm and arm64 has -arch-suffixed images
+    ARCH_IMAGE_EXT="-${ARCH}"
+fi
+
+# These are needed for remote execs, hence we introduce them here
+BASE_EXEC_IMAGE=${DOCKERHUB_USER}/weaveexec${ARCH_IMAGE_EXT}
+EXEC_IMAGE=${BASE_EXEC_IMAGE}:${IMAGE_VERSION}
```
bboreham: You cannot move this after the first use (which is in the call to exec_remote on line 260)

I'm going to stop checking these now; please leave things in the same order or justify the movement.

weave
```diff
@@ -328,15 +300,82 @@ fi
...
+WEAVEDB_IMAGE=${DOCKERHUB_USER}/weavedb
+
+# - Docker versions prior to 1.6 cannot pull images from DockerHub
+# - The weavexec image embeds a Docker 1.6.2 client. Docker will give
+#   a "client is newer than server error" if the daemon has an older
+#   API version, which could be confusing if the user's Docker client
+#   is correctly matched with that older version.
+#
+# We therefore check that the user's Docker *client* is >= 1.6
+MIN_DOCKER_VERSION=1.6.0
```
bboreham: Does this need to be increased to allow manifests?

Makefile
```diff
@@ -10,9 +8,91 @@ RM=--rm
...
+	CC=gcc
+	
+	# Optional parameters that can be passed when linking C code into the Go binary
+	CGO_LDFLAGS=
+endif
+ifeq ($(ARCH),arm)
+	# Here we're using a much newer version of the Docker client
+	# because Docker on ARM hasn't been released in many older versions
+	WEAVEEXEC_DOCKER_ARCH?=armel
+	WEAVEEXEC_DOCKER_VERSION=1.12.3
```
bboreham: Do we really have to pick such a new version?  This will prevent Weave Net being used with 1.11, for instance.

Although #2545 presents an alternative solution.

Makefile
```diff
@@ -10,9 +8,91 @@ RM=--rm
...
+	# Using the (semi-)official alpine image
+	ALPINE_BASEIMAGE=aarch64/alpine:edge
+
+	# arm64 images have the -arm64 suffix, for instance weaveworks/weave-arm64:latest
+	ARCH_EXT?=-arm64
+
+	# The name of the gcc binary
+	CC=aarch64-linux-gnu-gcc
+
+	# The name of the gcc binary
```
bboreham: I don't think it is.

Makefile
```diff
@@ -61,6 +145,9 @@ NETGO_CHECK=@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
 	echo "    sudo go install -tags netgo std"; \
 	false; \
 }
+# The flags we are passing to go build. -extldflags -static for making a static binary, 
+# -linkmode external for linking external C libraries into the binary, -X main.version for telling the
```
bboreham: I think `-linkmode external` is to force the use of the `ld` linker, external to the Go toolchain, and the reason we need to do that is otherwise it doesn't make a sufficiently static executable that it will run on its own in a container.

Although we don't actually do that now.

Makefile
```diff
@@ -61,6 +145,9 @@ NETGO_CHECK=@strings $@ | grep cgo_stub\\\.go >/dev/null || { \
...
+# Go binary which version it is, -tags netgo for enforcing the native Go DNS resolver
```
bboreham: More specifically it's to enforce that the cgo resolver is not linked, because that breaks static linking.

Makefile
```diff
@@ -184,25 +291,63 @@ tools/.git:
...
+	$(MAKE) ARCH=$* DOCKER_HOST=$(DOCKER_HOST) DOCKERHUB_USER=$(DOCKERHUB_USER) WEAVE_VERSION=$(WEAVE_VERSION) UPDATE_LATEST=$(UPDATE_LATEST) publish-one
+
+./manifest-tool:
+	wget https://github.com/luxas/manifest-tool/releases/download/v0.3.0/manifest-tool
+	chmod +x manifest-tool
 
 clean-bin:
 	-$(SUDO) DOCKER_HOST=$(DOCKER_HOST) docker rmi $(IMAGES)
-	rm -rf $(EXES) $(IMAGES_UPTODATE) $(WEAVE_EXPORT) .pkg
+	find prog -type f -print | grep "Dockerfile." | grep -v template | xargs rm -f
```
bboreham: `find -print | grep pattern` can be elided as `find -name pattern -print`

(allowing that the patterns have different syntax, but the syntax here seems suspicious anyway)

luxas:
@bboreham Updated, but I haven't had a chance to run a full push/publish suite on it yet that makes sure it actually works.
More comments?

bboreham:
> `make publish-all`

I cannot see this command; perhaps that comment is out of date?

luxas:
I just updated it to simply publish, will update my outdated comment above

luxas:
BTW, I ran the build yesterday on an empty Scaleway server and pushed everything to luxas with version v1.9.0-alpha.5 and then ran the official weave script on an Scaleway ARM server and it worked just fine!

Makefile
```diff
@@ -184,25 +285,71 @@ tools/.git:
...
+publish-one-arch: $(PUBLISH)
+
+# This rule handles the pushing of the built images
 $(PUBLISH): publish_%: $(IMAGES_UPTODATE)
-	$(SUDO) DOCKER_HOST=$(DOCKER_HOST) docker tag  $(DOCKERHUB_USER)/$* $(DOCKERHUB_USER)/$*:$(WEAVE_VERSION)
+	# Tag :latest with the real version
+	$(SUDO) DOCKER_HOST=$(DOCKER_HOST) docker tag  $(DOCKERHUB_USER)/$*$(ARCH_EXT) $(DOCKERHUB_USER)/$*$(ARCH_EXT):$(WEAVE_VERSION)
+	# Push the image with the arch suffix for arm and arm64, and without suffix for amd64
+	$(MAKE) DOCKER_HOST=$(DOCKER_HOST) DOCKERHUB_USER=$(DOCKERHUB_USER) WEAVE_VERSION=$(WEAVE_VERSION) UPDATE_LATEST=$(UPDATE_LATEST) push_$*$(ARCH_EXT)
+
```
bboreham: This blank line is suspicious - doesn't it terminate the recipe ?

bboreham: ah, no, ["Blank lines and lines of just comments may appear among the recipe lines; they are ignored."](https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html)

weave
```diff
@@ -8,14 +8,13 @@ IMAGE_VERSION=latest
...
-# - The weavexec image embeds a Docker 1.6.2 client. Docker will give
+# - The weavexec image embeds a Docker 1.10.3 client. Docker will give
 #   a "client is newer than server error" if the daemon has an older
 #   API version, which could be confusing if the user's Docker client
 #   is correctly matched with that older version.
 #
-# We therefore check that the user's Docker *client* is >= 1.6
-MIN_DOCKER_VERSION=1.6.0
+# We therefore check that the user's Docker *client* is >= 1.10.3
+MIN_DOCKER_VERSION=1.10.3
```
luxas: ok, I'll set it to 1.10.0

prog/weave-npc/Dockerfile.template
```diff
@@ -0,0 +1,23 @@
+FROM ALPINE_BASEIMAGE
```
luxas: oops, missed that deletion

Makefile
```diff
@@ -10,9 +8,82 @@ RM=--rm
...
+# from amd64 machines.
+QEMU_VERSION=v2.7.0
+
+# A list of all supported architectures here. Should be named as Go is naming platforms
+# All supported architectures must have an "ifeq" block below that customizes the parameters
+ALL_ARCHITECTURES=amd64 arm arm64
+ML_PLATFORMS=linux/amd64,linux/arm,linux/arm64
+
+ifeq ($(ARCH),amd64)
+	# The architecture to use when downloading the docker binary
```
luxas: Should I do anything about it? I'm using Sublime Text, so I don't see it ;)

Makefile
```diff
@@ -132,27 +214,46 @@ lint:
...
+# Creates the Dockerfile.your-user-here file from the template
+# Also replaces all placeholders with real values
+# If the architecture is amd64, it deletes all CROSS_BUILD lines
+# but otherwise, it only removes the "CROSS_BUILD_" placeholder and downloads QEMU
 %/Dockerfile.$(DOCKERHUB_USER): %/Dockerfile.template
-	sed -e "s/DOCKERHUB_USER/$(DOCKERHUB_USER)/" $^ > $@
+	echo "DOCKERHUB_USER|$(DOCKERHUB_USER)|g;s|ARCH_EXT|$(ARCH_EXT)|g;s|ALPINE_BASEIMAGE|$(ALPINE_BASEIMAGE)|g;s|QEMUARCH|$(QEMUARCH)"
+	sed -e "s|DOCKERHUB_USER|$(DOCKERHUB_USER)|g;s|ARCH_EXT|$(ARCH_EXT)|g;s|ALPINE_BASEIMAGE|$(ALPINE_BASEIMAGE)|g;s|QEMUARCH|$(QEMUARCH)|g" $^ > $@
+ifeq ($(ARCH),amd64)
+	# When building "normally" for amd64, remove the whole line, it has no part in the amd64 image
```
luxas: Is that a bad thing?

Makefile
```diff
@@ -1,7 +1,5 @@
-PUBLISH=publish_weave publish_weaveexec publish_plugin publish_weave-kube publish_weave-npc
-
 .DEFAULT: all
-.PHONY: all exes testrunner update tests lint publish $(PUBLISH) clean clean-bin prerequisites build run-smoketests
+.PHONY: all exes testrunner update tests lint publish-one $(PUBLISH) clean clean-bin prerequisites build run-smoketests
```
luxas: fixed

Makefile
```diff
@@ -132,27 +214,45 @@ lint:
...
+	sed -i "s/CROSS_BUILD_//g" $@
+endif
 
+# The targets below builds the weave images
 $(WEAVER_UPTODATE): prog/weaver/Dockerfile.$(DOCKERHUB_USER) $(WEAVER_EXE) $(WEAVEEXEC_UPTODATE)
 	$(SUDO) DOCKER_HOST=$(DOCKER_HOST) docker build -f prog/weaver/Dockerfile.$(DOCKERHUB_USER) -t $(WEAVER_IMAGE) prog/weaver
 	touch $@
 
-$(WEAVEEXEC_UPTODATE): prog/weaveexec/Dockerfile prog/weaveexec/symlink $(DOCKER_DISTRIB) weave $(SIGPROXY_EXE) $(WEAVEPROXY_EXE) $(WEAVEWAIT_EXE) $(WEAVEWAIT_NOOP_EXE) $(WEAVEWAIT_NOMCAST_EXE) $(WEAVEUTIL_EXE)
+$(WEAVEEXEC_UPTODATE): prog/weaveexec/Dockerfile.$(DOCKERHUB_USER) prog/weaveexec/symlink $(DOCKER_DISTRIB) weave $(SIGPROXY_EXE) $(WEAVEPROXY_EXE) $(WEAVEWAIT_EXE) $(WEAVEWAIT_NOOP_EXE) $(WEAVEWAIT_NOMCAST_EXE) $(WEAVEUTIL_EXE)
```
luxas: That's a quite hard thing now; there are a lot of other changes and I'm not sure it's worth spending time on extracting specific lines into separate commits.

Makefile
```diff
@@ -184,25 +285,71 @@ tools/.git:
...
+	$(SUDO) DOCKER_HOST=$(DOCKER_HOST) docker push $(DOCKERHUB_USER)/weavedb:latest
+endif
+
+sub-publish-%:
+	$(MAKE) DOCKER_HOST=$(DOCKER_HOST) clean-bin
+	$(MAKE) ARCH=$* DOCKER_HOST=$(DOCKER_HOST) DOCKERHUB_USER=$(DOCKERHUB_USER) WEAVE_VERSION=$(WEAVE_VERSION) UPDATE_LATEST=$(UPDATE_LATEST) $(IMAGES_UPTODATE)
+	$(MAKE) ARCH=$* DOCKER_HOST=$(DOCKER_HOST) DOCKERHUB_USER=$(DOCKERHUB_USER) WEAVE_VERSION=$(WEAVE_VERSION) UPDATE_LATEST=$(UPDATE_LATEST) publish-one-arch
+
+./manifest-tool:
+	wget https://github.com/luxas/manifest-tool/releases/download/v0.3.0/manifest-tool
```
luxas: Works for me, should I send a PR against build-tools with a submodule to luxas/manifest-tool?

build/build.sh
```diff
@@ -25,4 +25,7 @@ echo "weave:x:$uid:$gid::$WEAVE_SRC:/bin/sh" >>/etc/passwd
 echo "weave:*:::::::" >>/etc/shadow
 echo "weave	ALL=(ALL)	NOPASSWD: ALL" >>/etc/sudoers
 
+# Allow full write access to the Go folders for the user Weave
+chmod -R 777 /usr/local/go
```
luxas: Yes, I know that it's full control for everyone, I guess that was just a quick fix early on to get it working.

Now the user that's running go (weave) has to be able to write to the `/usr/local/go/pkg/linux_${ARCH}_cgo` packages IIRC and by default only root can do that. amd64 has this precompiled

Makefile
```diff
@@ -38,21 +111,25 @@ WEAVEDB_UPTODATE=.weavedb.uptodate
 
 IMAGES_UPTODATE=$(WEAVER_UPTODATE) $(WEAVEEXEC_UPTODATE) $(PLUGIN_UPTODATE) $(WEAVEKUBE_UPTODATE) $(WEAVENPC_UPTODATE) $(WEAVEDB_UPTODATE)
 
-WEAVER_IMAGE=$(DOCKERHUB_USER)/weave
-WEAVEEXEC_IMAGE=$(DOCKERHUB_USER)/weaveexec
-PLUGIN_IMAGE=$(DOCKERHUB_USER)/plugin
-WEAVEKUBE_IMAGE=$(DOCKERHUB_USER)/weave-kube
-WEAVENPC_IMAGE=$(DOCKERHUB_USER)/weave-npc
+# The names of the images. Note that the images for other architectures than amd64 have a suffix in the image name.
+WEAVER_IMAGE=$(DOCKERHUB_USER)/weave$(ARCH_EXT)
```
luxas: fixed that now

luxas:
@bboreham Updated this now with the comments you had. PTAL

bboreham:
Please can you add `manifest-tool` as a submodule rather than directly including it.  There are some pointers how to do this at https://github.com/weaveworks/weave/tree/master/vendor

Makefile
bboreham: actually I'm wrong here; prog/weaveexec/Dockerfile was renamed

bboreham:
I pushed a copy of this branch (+- manifest-tool submodule) to the weaveworks repo so it would trigger a full CI build https://circleci.com/gh/weaveworks/weave/7707

luxas:
@bboreham Yes, it does indeed look strange now in the diff, but I added it as described here: https://github.com/weaveworks/build-tools#using-build-toolsgit (seems like that wasn't the right path though :smile:)

bboreham:
I followed this procedure (starting from your branch at 2392acd734a529f57c28b5c1f30bfc352a2f327f):

```
$ git reset --hard 906f9e8048938cfaedcd031bb540dec3c15b3b67
[edit .gitignore to remove manifest-tool]
$ git submodule add https://github.com/estesp/manifest-tool manifest-tool
$ git add manifest-tool .gitignore
$ git commit -m "Add manifest-tool as a submodule"
$ git cherry-pick 2392acd734a529f57c28b5c1f30bfc352a2f327f
```

However, subsequent [runs on CI](https://circleci.com/gh/weaveworks/weave/7715) suggest it would be best to bring it in as `vendor/github.com/estesp/manifest-tool` because that way it will be excluded from lint and unit test runs.

Makefile
bboreham: Yes, please move all the comments you have added so the `#` is in the first column

Makefile
```diff
@@ -164,45 +266,92 @@ $(WEAVEKUBE_UPTODATE): prog/weave-kube/Dockerfile.$(DOCKERHUB_USER) prog/weave-k
...
+# This target first runs "make publish" for each architecture
+# Then it pushes the manifest lists
+publish: $(MANIFEST_TOOL_EXE) $(addprefix sub-publish-,$(ALL_ARCHITECTURES)) $(WEAVEDB_UPTODATE)
+	$(MAKE) DOCKER_HOST=$(DOCKER_HOST) DOCKERHUB_USER=$(DOCKERHUB_USER) WEAVE_VERSION=$(WEAVE_VERSION) UPDATE_LATEST=$(UPDATE_LATEST) $(PUSH_ML)
+ifeq ($(PUBLISH_WEAVEDB),true)
+	$(SUDO) DOCKER_HOST=$(DOCKER_HOST) docker push $(DOCKERHUB_USER)/weavedb:latest
+endif
+
+sub-publish-%:
+	$(MAKE) ARCH=$* DOCKER_HOST=$(DOCKER_HOST) clean-bin
```
bboreham: I was expecting the platform-specific `.uptodate` files to remove the need for this line.  It makes debugging the `make publish` process fantastically tedious.

luxas: There are still other files, like the docker binary, the templates, etc. which will get cached and has to be cleaned between consequent builds

luxas:
Pushed the mechanical submodule changes now.
You might have noted it already, but my changes were merged into the upstream estesp repo, so we're good to go with that one now

Makefile
bboreham: Hmm.
The `docker` binary is overwritten as part of the recipe for `weaveexec` container.
However the Docker distribution tarfile is more of a problem - it could be renamed to be per-platform.
All of our own binaries (`weaver`, `weaveutil`, etc.) would also have to be moved or renamed.
What I hate about having `clean-bin` here is that if you ever enter into this process then you're several minutes away from doing anything else.

bboreham:
Please can you rebase on master or otherwise get 2298a40327c101034052bdd6ce6943f28d357854 included on your branch.

Also please add 1cc08f4d71503baffea0cb10b3d5cb8363cb3a52, otherwise the move to Docker 1.10 minimum kills our CI.

Makefile
```diff
@@ -164,45 +267,91 @@ $(WEAVEKUBE_UPTODATE): prog/weave-kube/Dockerfile.$(DOCKERHUB_USER) prog/weave-k
...
+	$(MAKE) ARCH=$* DOCKER_HOST=$(DOCKER_HOST) DOCKERHUB_USER=$(DOCKERHUB_USER) WEAVE_VERSION=$(WEAVE_VERSION) UPDATE_LATEST=$(UPDATE_LATEST) $(IMAGES_UPTODATE)
```
bboreham: I don't think this is right - `$(IMAGES_UPTODATE)` needs to be modified for the different `ARCH`.  However I don't think this line is needed at all since `publish-one-arch` depends on `$(PUBLISH)` which depends on `$(IMAGES_UPTODATE)`

luxas: When I didn't have this line, `$(IMAGES_UPTODATE)` executed just once, so it's needed to some extent at least.

luxas:
@bboreham Rebased and the commit you pointed out is now included

Makefile
```diff
@@ -132,27 +216,46 @@ lint:
...
-	sed -e "s/DOCKERHUB_USER/$(DOCKERHUB_USER)/" $^ > $@
+	echo "DOCKERHUB_USER|$(DOCKERHUB_USER)|g;s|ARCH_EXT|$(ARCH_EXT)|g;s|ALPINE_BASEIMAGE|$(ALPINE_BASEIMAGE)|g;s|QEMUARCH|$(QEMUARCH)"
+	sed -e "s|DOCKERHUB_USER|$(DOCKERHUB_USER)|g;s|ARCH_EXT|$(ARCH_EXT)|g;s|ALPINE_BASEIMAGE|$(ALPINE_BASEIMAGE)|g;s|QEMUARCH|$(QEMUARCH)|g" $^ > $@
+ifeq ($(ARCH),amd64)
+	# When building "normally" for amd64, remove the whole line, it has no part in the amd64 image
+	sed -i "/CROSS_BUILD_/d" $@
+else
+	# When cross-building, only the placeholder "CROSS_BUILD_" should be removed
+	# Register /usr/bin/qemu-ARCH-static as the handler for ARM binaries in the kernel
+	curl -sSL https://github.com/multiarch/qemu-user-static/releases/download/$(QEMU_VERSION)/x86_64_qemu-$(QEMUARCH)-static.tar.gz | tar -xz -C $(shell dirname $@)
```
bboreham: I investigated whether this download and copy could be replaced by saying `FROM multiarch/alpine:arch-version` in our Dockerfiles.  It seems those images aren't published on all the architectures we need.  Is that a possible alternative?

Makefile
bboreham: Could that have been before the `.uptodate` files were tagged with the architecture?  Removing this line works fine for me; leaving it in I get messages that it doesn't know how to build them on non-x84 arch.

Makefile
luxas: What about just checking in the binary that's needed?
Or checking it into the `tools/` submodule and bumping version?

BTW; With a kernel that's of version 4.8+, we don't have to include the binary in the image anymore, but for now since I suppose not everyone's got 4.8+ (me included), I'm copying it into the container

luxas: With 4.8+ that binary only have to exist on the host that's building the image

Makefile
```diff
@@ -164,45 +267,91 @@ $(WEAVEKUBE_UPTODATE): prog/weave-kube/Dockerfile.$(DOCKERHUB_USER) prog/weave-k
...
-tools/.git:
+tools/.git $(MANIFEST_TOOL_DIR)/.git:
 	git submodule update --init
 
 $(RUNNER_EXE): tools/.git
 	make -C tools/runner
 
+$(MANIFEST_TOOL_EXE): $(MANIFEST_TOOL_DIR)/.git
+	make -C $(MANIFEST_TOOL_DIR)
+	chown $(shell id -u):$(shell id -g) $(MANIFEST_TOOL_EXE)
```
luxas: The problem is that the binary that `make -C manifest-tool` is outputting has `root:root` permissions.
I could send a PR against the estesp repo to fix if you want

bboreham: I think it might be best to run the `go build` in our own build container.
This will need to be revisited if the `manifest-tool` build process gets more complicated.

PR #2716, opened 2016-12-28
fix a bug of updating network policy

liubog2008:
...


awh:
@liubog2008 great spot - thanks! How did you notice it?

bboreham:
Note [Kubernetes currently prohibits updating of network policies](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/extensions/validation/validation.go#L869)

Issue #2733, opened 2017-01-16
Error launching weave from NixOS: chroot: can't execute 'readlink': No such file or directory


roberth:
Steps to reproduce:

 - Run NixOS
 - Download the launch script as instructed by https://www.weave.works/docs/net/latest/installing-weave/
 - (Add set -x to top of script)
 - `$ weave launch --log-level=debug`

Actually I did not install to /usr/local/bin but to a user directory, but that does not seem to be the cause.

Error:

    + docker run --rm --privileged --net=host -v /var/run/docker.sock:/var/run/docker.sock --pid=host -v /:/host -e HOST_ROOT=/host -e DOCKERHUB_USER=weaveworks -e WEAVE_VERSION -e WEAVE_DEBUG -e WEAVE_DOCKER_ARGS -e WEAVEPROXY_DOCKER_ARGS -e WEAVEPLUGIN_DOCKER_ARGS -e WEAVE_PASSWORD -e WEAVE_PORT -e WEAVE_HTTP_ADDR -e WEAVE_STATUS_ADDR -e WEAVE_CONTAINER_NAME -e WEAVE_MTU -e WEAVE_NO_FASTDP -e WEAVE_NO_BRIDGED_FASTDP -e WEAVE_NO_PLUGIN -e DOCKER_BRIDGE -e DOCKER_CLIENT_HOST= -e DOCKER_CLIENT_ARGS -e PROXY_HOST=127.0.0.1 -e COVERAGE -e CHECKPOINT_DISABLE -e AWSVPC weaveworks/weaveexec:1.8.2 --local launch --log-level=debug
    chroot: can't execute 'readlink': No such file or directory

roberth:
Working on it. Can't currently assign to myself.

bboreham:
Why is this affected by the host?  It ought to be running inside a container at that point, which does contain the `readlink` executable.

> Add set -x to top of script

You can achieve this by setting `WEAVE_DEBUG` to nonblank before running.

roberth:
The script chroots into the host to determine the right path as a one-liner. So it is actually inside a chroot inside a container.
It works quite well when it does, but it does not work on my machine, hence the fallback.
Alternatively, we can avoid the duplicate logic and always determine the link using some string processing, but I did not want to take that risk for a first contribution. It would make the container less dependent on the host.
Thanks for the hint!

bboreham:
Ah, thanks, I had failed to spot that subtlety.

PR #2734, opened 2017-01-16
Add partial support for non-FHS host systems

roberth:
Fixes weaveworks/weave#2733 partially, where <host>/etc/resolv.conf is not a
symlink

### DIFF
--- a/weave
+++ b/weave
@@ -1667,7 +1667,16 @@ launch_router() {
            --label=weavevolumes $WEAVEDB_IMAGE >/dev/null
     fi
 
-    RESOLV_CONF=$(chroot ${HOST_ROOT:-/} readlink -f /etc/resolv.conf)
+    # Figure out the location of the actual resolv.conf file because
+    # we want to bind mount its directory into the container.
+    if [ -L ${HOST_ROOT:-/}/etc/resolv.conf ]; then # symlink
+        # This assumes a host with readlink in FHS directories...
+        # Ideally, this would resolve the symlink manually, without
+        # using host commands.
+        RESOLV_CONF=$(chroot ${HOST_ROOT:-/} readlink -f /etc/resolv.conf)
+    else
+        RESOLV_CONF=/etc/resolv.conf
+    fi
     RESOLV_CONF_DIR=$(dirname "$RESOLV_CONF")
     RESOLV_CONF_BASE=$(basename "$RESOLV_CONF")
 


weave
```diff
@@ -1667,7 +1667,31 @@ launch_router() {
            --label=weavevolumes $WEAVEDB_IMAGE >/dev/null
     fi
 
-    RESOLV_CONF=$(chroot ${HOST_ROOT:-/} readlink -f /etc/resolv.conf)
+    echo "Container PATH is $PATH"
+    # Avoid readlink if not present in FHS directories of host
+    if test -x ${HOST_ROOT:-/}/bin/readlink \
```
bboreham: could this be `command_exists readlink` ?

roberth: `command_exists` can only check the current environment, not the chroot. I don't know of any standard command that would do it for another root. Something like `chroot ... sh -c "command -v ..."` might work in some cases, even NixOS, but that's executing a binary ("sh") from the host and I'd rather not depend on the host like that.

rade: so perhaps simply
1. run `test -L ...` to determine whether we are looking at a symlink
2. if false then no-op
3. if true then attempt the chroot'ed readlink

This doesn't give us a nice error when readlink isn't there but IMO the stock error is quite ok.

Or perhaps even just
```
RESOLV_CONF=/etc/resolv.conf
[ -f ${HOST_ROOT:-/}$RESOLV_CONF ] || RESOLV_CONF=$(chroot ${HOST_ROOT:-/} readlink -f $RESOLV_CONF)
```

roberth:
I've put the symlink condition first

weave
```diff
@@ -1667,7 +1667,14 @@ launch_router() {
            --label=weavevolumes $WEAVEDB_IMAGE >/dev/null
     fi
 
-    RESOLV_CONF=$(chroot ${HOST_ROOT:-/} readlink -f /etc/resolv.conf)
+    if [ -f ${HOST_ROOT:-/}/etc/resolv.conf ]; then # regular file
```
bboreham: Sadly `-f` which tests for "regular file" in this context does not exclude symbolic links to files; we need a test such as `-L`

roberth: Thanks. I copied that from someone's comment on the previous version of this PR. I will switch back to `-L`

weave
```diff
@@ -1667,7 +1667,16 @@ launch_router() {
...
 
-    RESOLV_CONF=$(chroot ${HOST_ROOT:-/} readlink -f /etc/resolv.conf)
+    # Figure out the location of the actual resolv.conf file because
+    # we want to bind mount its directory into the container.
+    if [ -L ${HOST_ROOT:-/}/etc/resolv.conf ]; then # symlink
+        # This assumes a host with readlink in FHS directories...
+        # Ideally, this would resolve the symlink manually, without
+        # using host commands.
+        RESOLV_CONF=$(chroot ${HOST_ROOT:-/} readlink -f /etc/resolv.conf)
+    else # /etc/resolv.conf is not a symlink
```
rade: The comment is misleading since we will also end up in this branch when the file does not exist.

bboreham:
Thanks for the contribution, @roberth !

PR #2337, opened 2016-06-06
Speed up CircleCI builds

bboreham:
The main idea was to cache the Go code from vendored libraries.
CircleCI saves the cache between "dependencies" and "tests", so the build needs to move to "dependencies".

A much bigger time-saving comes from running the GCE commands in the background.  

Other changes:
- Don't spell-check compressed files, now `weave.tar.gz` is created before this test is run
- Build test runner inside the build container so it can share the cached files (and this is better for consistency too)

PR #2742, opened 2017-01-19
Build improvements

bboreham:
Subset of #2337 which should be of general usefulness; also something similar to the `testrunner` changes seems to be needed for the ARM port


marccarre:
LGTM.

PR #2711, opened 2016-12-22
Better machine unique IDs

bboreham:
Fixes #2427, taking additional input from `/etc/machine-id` or `/var/lib/dbus/machine-id` if available.

In order to avoid changing the peer ID on an upgrade we read the peer ID persisted by IPAM, and use that if it matches the previous style.

This code uses the approach from #2460, plus code to generate MACs from #1958.

Note that we cannot open the BoltDB file if it is currently locked for writing (e.g. by a weave process).  Opening it read-only is therefore a small gain - it allows multiple instances of `unique-id` to run concurrently - but it seems better to be able to declare intent.

PR #2743, opened 2017-01-19
Don't fail if only half the system IDs are blank

bboreham:
Fix a bug in #2711 
The symptom, on Scaleway, is:

```+ weaveutil unique-id /host//weave /host
All system IDs are blank
```


marccarre:
LGTM:
We need at least one system ID (i.e. either `uuid` or `machineid`, or both) in order to be able to generate a unique hash in `MACfromUUID`, so `getSystemUUID` should only fail if *both* `uuid` and `machineid` couldn't be read from the system.

Issue #2689, opened 2016-12-11
WeaveNet started via weave-kube breaks multicast


bergtwvd:
When I use Docker Compose on a cluster of docker hosts and weave overlay networking / multicast then I use the following declaration in the network section of the Docker compose file:

````
networks:
 training:
  driver: weave
  driver_opts:
   works.weave.multicast: "1"
````

I have set up Kubernetes with Weave Net with a couple of containers talking multicast.
How can I achieve on Kubernetes what I do on a Docker overlay network ?

(they  run fine, when deployed as a pod, but I want them to run on different nodes)

bboreham:
I just tried this with the `weave-kube` deployment, and I saw that the Network Policy Controller is blocking multicast traffic.  You can check if you have the same problem by looking at the log files of the `weave-npc` container on each host.

We're just pondering how best to fix.  You can work around by disabling the firewall completely (delete the daemonset; edit the template to set env var `EXPECT_NPC` to `0` for weave-kube container, reboot and recreate the daemonset) or by adding a specific iptables rule on each host to allow multicast traffic.

bergtwvd:
Just to confirm, I see the following message in the weave npc log on the node where the (multicast) pod runs:

``time="2016-12-12T23:00:00Z" level=warning msg="UDP connection from 10.42.0.4:20913 to 239.20.9.13:20913 blocked by Weave NPC."``

bergtwvd:
Not being a kube expert yet, where / how can I edit the template to set EXPECT_NPC  ?

bboreham:
@bergtwvd we give an example at https://www.weave.works/docs/net/latest/kube-addon/#configuration-options

bboreham:
Some notes on the underlying problem:

`weave-npc` works via `iptables` rules saying (via ipset) which IP addresses can send to which other IP addresses.  At present no multicast addresses are in the sets, so all such traffic is blocked.

If we think of adding multicast to the set of allowed addresses, we have the problem of distinguishing which containers should receive it - the same multicast address may be used by multiple pods, but there is only one set of rules.

Trying to enumerate the rules that `weave-npc` should enforce, it should allow multicast traffic:
 1. from anywhere to pods in namespaces which are `DefaultAllow` (the default).
 1. to pods in namespaces which are `DefaultDeny`, if permitted by a rule which allows the port number
 1. to pods in namespaces which are `DefaultDeny`, from pod IP addresses specifically selected by a rule

[...]

PR #2747, opened 2017-01-23
Add rule to allow multicast at top level

bboreham:
Simple kludge for #2689 

I have made some steps towards a fuller solution in branch issues/2689-npc-mcast, but I realised it needs quite a bit more work - it should tie in with the way `net/veth.go` adds per-subnet multicast rules.

### DIFF
--- a/prog/weave-npc/main.go
+++ b/prog/weave-npc/main.go
@@ -27,6 +27,7 @@ var (
 	version     = "unreleased"
 	metricsAddr string
 	logLevel    string
+	allowMcast  bool
 )
 
 func handleError(err error) { common.CheckFatal(err) }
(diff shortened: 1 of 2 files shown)


bboreham:
I guess we should document this, and the flag to turn it off

awh:
#### Documentation changes:
modified [kube-addon](https://dev-weavewww.pantheonsite.io/docs/net/2689-npc-mcast-simple/kube-addon/)

kgelinas:
You can turn this behaviour off (block
 +  all multicast traffic) by adding `--allow-mcast` as an argument to
 +  `weave-npc` in the YAML configuration.

I'm confused
do you allow multicast with allow-mcast or do you block all multicast traffic?

rade:
All multicast traffic is enabled by default, and can be blocked by specifying `--allow-mcast=false`.

rade:
I have updated the docs in a1fb08216e70cd619d5b101a52c95281535e39e8.

kgelinas:
thanks :)

Issue #53, opened 2014-09-13
eliminate tx offload disabling in application container


rade:
As of https://github.com/zettio/weave/commit/6eb7e1aa437df512cf8d891d0044d1acfd7266b0 we disable tx offload on the 'weave' bridge. It is possible that as a result we no longer need to disable it on the container's ethwe interfaces, since these are connected to that bridge. All depends on when the kernel decides to do offloading.

PR #2749, opened 2017-01-24
Upgrade Weave Net to latest version of build-tools, skipping bash scripts when lint-ing.

marccarre:
Lint-ing all of Weave Net is currently not desirable because: 

* it would create conflicts on many branches,
* it would pollute the git history, 
* etc. 

and because `lint` currently lints all files except `./vendor`, this breaks Weave Net's build.
As a result, Weave Net is not using the latest commit in `build-tools`'s `master` branch, which is also source of problems:

* it limits standardisation of projects across Weaveworks, 
* it limits the ability to use the latest features of `build-tools`, 
* it complexifies merging changes in `build-tools` (e.g. #53 and weaveworks/weave#2694 for weaveworks/weave#2647), 
* etc.

This change:

* brings Weave Net to the latest version of `build-tools` (which now supports `.lintignore` files),
* adds such a `.lintignore` file to skip all bash scripts (in order to avoid breaking the build).

`GLOB`s in `.lintignore` can then, over time and as scripts are `shfmt`-ed, be relaxed and maybe even removed.

### DIFF
--- a/.lintignore
+++ b/.lintignore
@@ -1,0 +1,8 @@
+# IMPORTANT: Please only use GLOBs to ignore files and directories.
+
+# Do not lint any bash script in Weave Net, in order to 
+# avoid breaking the build on make lint:
+*.sh
+bin/*
+prog/weaveexec/symlink
+weave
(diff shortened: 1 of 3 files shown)


brb:
LGTM.

PR #2752, opened 2017-01-25
Makefile cleanup after multiarch merge

bboreham:
Make manifest-tool using weave build container, and remove whitespace in blank lines and comment lines.

The latter principally because emacs complains about it, but also to remove the overhead of firing up a shell process to ignore a comment in a recipe.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ ifeq ($(ARCH),amd64)
 
 # The name of the gcc we're using for compiling C code
 	CC=gcc
-	
+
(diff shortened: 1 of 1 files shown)


bboreham:
I have tagged on an extra commit: a fix to a pre-existing bug where `make clean-bin` would remove the `weavedb` image but leave the file `.weavedb.uptodate`

marccarre:
> I have tagged on an extra commit
Yes, saw that, and guessed the reason correctly.
Thanks for making it explicit though!
Still LGTM.

PR #2757, opened 2017-01-30
Upgrade to latest 'common' and 'build-tools'.

marccarre:
If building against the latest `build-tools`, you typically get:

```
tools/runner/runner.go:18:2: cannot find package github.com/weaveworks/common/mflag
```

This change fixes this issue by upgrading both: 

* `weaveworks/common`, and
* `weaveworks/build-tools`

to their latest versions.


marccarre:
Thanks for the suggestion Bryan, just did it in 8f9b2a3.
I had a look at `docker/docker` too, and it seems we only use it for/in:
```
/proxy/tls.go:	"github.com/docker/docker/pkg/homedir"
```

tomwilkie:
> I suggest moving Weave's use of mflag from weaveworks/docker to weaveworks/common, to match the change in runner.
> Then we can remove weaveworks/docker entirely

was going to do this today, you beat me to it!

marccarre:
I just saw we also replaced:
`"github.com/weaveworks/weave/common/mflagext"`
with
`"github.com/weaveworks/common/mflagext"`
in `build-tools`, so I'll replace this in Weave Net as well, unless anyone has any objection.

PR #2760, opened 2017-02-01
Minor build improvements

bboreham:
`touch` exes after running build command - the executables have dependencies that are not apparent to `go build`; notably the build container. If Go doesn't see anything is out of date then it will do nothing, so the next time you run `make` the process will repeat, wasting time.
Hence, touch the file so next time we run `make` it is up to date.

~Remove~Move down `WORKDIR` lines from `Dockerfile`s - they seem to stop Docker using its build cache, and ~are unnecessary since we don't do anything that needs to run in a specific directory~ are mostly unnecessary.


Makefile
```diff
@@ -179,6 +179,7 @@ exes $(EXES) tests lint: $(BUILD_UPTODATE)
 		-v $(shell pwd)/.pkg:/go/pkg \
 		-e GOARCH=$(ARCH) -e CGO_ENABLED=1 -e GOOS=linux -e CIRCLECI -e CIRCLE_BUILD_NUM -e CIRCLE_NODE_TOTAL -e CIRCLE_NODE_INDEX -e COVERDIR -e SLOW -e DEBUG \
 		$(BUILD_IMAGE) COVERAGE=$(COVERAGE) WEAVE_VERSION=$(WEAVE_VERSION) CC=$(CC) QEMUARCH=$(QEMUARCH) CGO_LDFLAGS=$(CGO_LDFLAGS) $@
+	touch $@
```
brb: Shouldn't we remove `exes` from `.PHONY`?

brb:
Could you change the base branch to `1.9`?

brb:
The smoke tests are timing out.

Makefile
bboreham: "A phony target is one that is not really the name of a file; rather it is just a name for a recipe to be executed when you make an explicit request."

Seems appropriate - there is no file named `exes`

bboreham:
I fixed the issue that was breaking the tests.

marccarre:
LGTM.

PR #2762, opened 2017-02-01
Allow traffic to forward off the weave bridge

### LINKED ISSUE #2758: Docker 1.13 defaults IP forwarding to DROP
This change is in https://github.com/docker/docker/pull/28257.  Effects include "pod-to-pod traffic works but pod to Internet does not".  It will probably also break most uses of `weave expose`.

Docker add these rules:

    -A FORWARD -i docker0 ! -o docker0 -j ACCEPT
    -A FORWARD -i docker0 -o docker0 -j ACCEPT

so we could do something similar for the weave bridge.  It needs to take into account the `weave-npc` rules which already cover traffic onto the weave bridge (if enabled).

bboreham:
Fixes #2758

### DIFF
--- a/weave
+++ b/weave
@@ -620,6 +620,10 @@ try_create_bridge() {
             # across our bridge. E.g. ufw
             add_iptables_rule filter FORWARD -i $BRIDGE -o $BRIDGE -j ACCEPT
         fi
+        # Forward from weave to the rest of the world
+        add_iptables_rule filter FORWARD -i $BRIDGE ! -o $BRIDGE -j ACCEPT
+        # and allow replies back
+        add_iptables_rule filter FORWARD -o $BRIDGE -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 
         # create a chain for masquerading
         run_iptables -t nat -N WEAVE >/dev/null 2>&1 || true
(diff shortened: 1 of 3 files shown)


weave
```diff
@@ -620,6 +620,10 @@ try_create_bridge() {
             # across our bridge. E.g. ufw
             add_iptables_rule filter FORWARD -i $BRIDGE -o $BRIDGE -j ACCEPT
         fi
+        # Forward from weave to the rest of the world
+        add_iptables_rule filter FORWARD -i $BRIDGE ! -o $BRIDGE -j ACCEPT
+        # and allow replies back
+        add_iptables_rule filter FORWARD -o weave -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
```
marccarre: I am wondering why this rule has `-o weave` and not `-o $BRIDGE`? (given we set: `BRIDGE=weave` earlier)
Is it guaranteed that there will always be a bridge named `weave`? (it seems that `create_bridge` in the `weave` script always uses `$BRIDGE`, so I'm wondering if this could break if someone sets `BRIDGE` to a different value than the default)

weave
```diff
@@ -803,6 +807,8 @@ destroy_bridge() {
     run_iptables -t filter -D INPUT -i $DOCKER_BRIDGE -p udp --dst $DOCKER_BRIDGE_IP --dport $PORT          -j DROP >/dev/null 2>&1 || true
     run_iptables -t filter -D INPUT -i $DOCKER_BRIDGE -p udp --dst $DOCKER_BRIDGE_IP --dport $(($PORT + 1)) -j DROP >/dev/null 2>&1 || true
 
+    run_iptables -t filter -D FORWARD -i $BRIDGE ! -o $BRIDGE -j ACCEPT 2>/dev/null || true
+    run_iptables -t filter -D FORWARD -o weave -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT 2>/dev/null || true
```
marccarre: I am wondering why this rule has `-o weave` and not `-o $BRIDGE`? (given we set: `BRIDGE=weave` earlier)
Is it guaranteed that there will always be a bridge named `weave`? (it seems that `create_bridge` in the `weave` script always uses `$BRIDGE`, so I'm wondering if this could break if someone sets `BRIDGE` to a different value than the default)

weave
bboreham: good catch - will fix

marccarre:
LGTM! Thank you!

PR #2775, opened 2017-02-08
Remove hard-coded NUM_HOSTS env. var.

marccarre:
Instead of hardcoding this in the code, this should be passed by CircleCI from: 
```
Settings 
  > weaveworks 
    > weave 
      > Environment Variables.
```


brb:
Does it default to any value if no NUM_HOSTS is set in CircleCI?

marccarre:
Yes, it does default to `3` -- see also [run-integration-tests.sh#L27](https://github.com/weaveworks/weave/blob/master/test/run-integration-tests.sh#L27)

PR #2787, opened 2017-02-09
Claim existing IP addresses at startup with no race condition

### LINKED ISSUE #2784: Race in reclaiming existing addresses
When weave launches, the first thing it does is to iterate over existing addresses to "claim" them so they cannot be given out as new allocations.

Unfortunately nothing stops a new allocation getting there faster than the reclaim process.

This is believed to be the cause of https://github.com/weaveworks/weave/issues/2768#issuecomment-277887403

I can recreate as follows:
1. Fired up test 840 so I have 3 nodes, 3 pods running weave hello-world.  Don't let the test clear down.
2. ssh onto one of the minions which has 2 of the hello-world pods
3. find the location of the Weave persistence file - ` docker inspect <weavecontainer> | grep weavedb`
4. kill weave and hello-world pods, and wipe out the persisted data: `docker kill $(docker ps | grep pause | cut -d ' ' -f 1 | head -3) ; rm <that path>/weavedata.db`
5. `weave ps` to see if it duplicated
repeat about 6 times

bboreham:
Fixes #2784 

`reclaim.go` is a re-implementation of the `launch.sh` code to claim addresses in Go.
We already had a `pendingClaims` list in IPAM, so we simply pre-populate this with entries for each existing address.  The claims will be processed as soon as the ring is known.

As it stands, this code is only good enough for `weave-kube`, not for `weave launch`
The code to find existing IPs works through the process table, and thus we do not have the Docker container ID.  To resolve that I propose to add another loop to go through Docker container IDs if we have a Docker API connection.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -77,8 +77,9 @@ type Allocator struct {
 }
 
 type PreClaim struct {
-	Ident string
-	Cidr  address.CIDR
+	Ident       string
+	IsContainer bool
+	Cidr        address.CIDR
 }
 
 type Config struct {
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -76,9 +76,10 @@ type Allocator struct {
 	now               func() time.Time
 }
 
+// PreClaims are IP addresses discovered before we could initialize IPAM
 type PreClaim struct {
-	Ident       string
-	IsContainer bool
+	Ident       string // a container ID, something like "weave:expose", or api.NoContainerID
+	IsContainer bool   // true if Ident is a container ID
 	Cidr        address.CIDR
 }
 
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -76,6 +76,11 @@ type Allocator struct {
 	now               func() time.Time
 }
 
+type PreClaim struct {
+	Ident string
+	Cidr  address.CIDR
+}
+
 type Config struct {
 	OurName     mesh.PeerName
 	OurUID      mesh.PeerUID
@@ -83,6 +88,7 @@ type Config struct {
 	Seed        []mesh.PeerName
 	Universe    address.CIDR
 	IsObserver  bool
+	PreClaims   []PreClaim
 	Quorum      func() uint
 	Db          db.DB
 	IsKnownPeer func(name mesh.PeerName) bool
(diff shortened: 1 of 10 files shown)


prog/weaver/reclaim.go
```diff
@@ -0,0 +1,54 @@
+package main
```
chrislovecnm: Where would it be appropriate to add some logging?

bboreham: What would you like it to log?

bboreham: I added logging of the addresses found.
Probably this should be at debug level, but I left it at INFO for now.

prog/weaver/main.go
```diff
@@ -306,13 +306,16 @@ func main() {
...
-		allocator, defaultSubnet = createAllocator(router, ipamConfig, db, t, isKnownPeer)
-		observeContainers(allocator)
+		var allContainerIDs []string
 		if dockerCli != nil {
-			ids, err := dockerCli.AllContainerIDs()
+			allContainerIDs, err = dockerCli.AllContainerIDs()
 			checkFatal(err)
-			allocator.PruneOwned(ids)
 		}
+		preClaims, err := findExistingAddresses(dockerCli, allContainerIDs, weavenet.WeaveBridgeName)
```
brb: Will it be called in the case of `attach-router`?

prog/weaver/reclaim.go
```diff
@@ -0,0 +1,84 @@
...
+	// Then find all veths connected to the bridge
+	peerIDs, err := weavenet.ConnectedToBridgeVethPeerIds(bridgeName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Now iterate over all containers to see if they have a network
+	// namespace with an attached interface
+	if dockerCli != nil {
+		for _, cid := range containerIDs {
```
brb: Maybe it's worth factoring out the code inside if-else blocks into two separate functions ?

prog/weaver/main.go
bboreham: It is called every time the process starts; `weave attach-router` is only called in our software when the proxy sees `weaver` start, so it comes to the same thing.

prog/weaver/reclaim.go
bboreham: I made an attempt at this but it was less readable.

ipam/allocator.go
```diff
@@ -76,13 +76,20 @@ type Allocator struct {
 	now               func() time.Time
 }
 
+type PreClaim struct {
+	Ident       string
+	IsContainer bool
```
bboreham: Have added some comments documenting this and related fields

PR #2795, opened 2017-02-13
Reference-count ipset entries

### LINKED ISSUE #2792: weave-npc can receive duplicate address after a disconnect
Trying to recreate #2772, I broke the connection between `weave-npc` and `api-server` and deleted a pod while the connection was down.  When weave-npc reconnected, it received the event to add the replacement pod before it had received anything about the old pod being deleted.

Logs:

```
INFO: 2017/02/10 11:47:34.118373 adding entry 10.46.0.0 to weave-k?Z;25^M}|1s7P3|H9i;*;MhG
```
then I broke the connection
```
E0210 12:16:46.076827    3679 reflector.go:311] github.com/weaveworks/weave/vendor/k8s.io/client-go/tools/cache/reflector.go:109: Failed to watch *v1.Pod: Get https://10.96.0.1:443/api/v1/watch/pods?resourceVersion=85093&timeoutSeconds=598: dial tcp 10.96.0.1:443: i/o timeout
```
after reconnect
```
INFO: 2017/02/10 12:16:54.111290 EVENT UpdateNamespace {"metadata":{"name":"default","selfLink":"/api/v1/namespace ...
...
DEBU: 2017/02/10 12:16:54.114055 EVENT AddPod {"metadata":{"name":"hello-2533203682-r9xcf","generateName":"hello-2
INFO: 2017/02/10 12:16:54.114090 adding entry 10.46.0.0 to weave-k?Z;25^M}|1s7P3|H9i;*;MhG
FATA: 2017/02/10 12:16:54.115594 add pod: ipset [add weave-k?Z;25^M}|1s7P3|H9i;*;MhG 10.46.0.0] failed: ipset v6.29: Element cannot be added to the set: it's already added
```

bboreham:
To avoid crashing on error from `ipset` when we receive an Add before a Delete; fixes #2792 

Note that a real, long-lasting, IP duplication is a real problem, and this change makes it much less likely that we would spot that.

### DIFF
--- a/npc/ipset/ipset.go
+++ b/npc/ipset/ipset.go
@@ -53,18 +53,22 @@ func (i *ipset) DelEntry(ipsetName Name, entry string) error {
 }
 
 func (i *ipset) Flush(ipsetName Name) error {
+	i.removeSet(ipsetName)
 	return doExec("flush", string(ipsetName))
 }
 
 func (i *ipset) FlushAll() error {
+	i.refCount = newRefCount()
 	return doExec("flush")
 }
 
 func (i *ipset) Destroy(ipsetName Name) error {
+	i.removeSet(ipsetName)
 	return doExec("destroy", string(ipsetName))
 }
 
 func (i *ipset) DestroyAll() error {
+	i.refCount = newRefCount()
 	return doExec("destroy")
 }
 
@@ -100,4 +104,12 @@ func (rc *refCount) dec(ipsetName Name, entry string) int {
 	k := key{ipsetName, entry}
 	rc.ref[k]--
 	return rc.ref[k]
+}
+
+func (rc *refCount) removeSet(ipsetName Name) {
+	for k := range rc.ref {
+		if k.ipsetName == ipsetName {
+			delete(rc.ref, k)
+		}
+	}
 }
--- a/npc/ipset/ipset.go
+++ b/npc/ipset/ipset.go
@@ -27,10 +27,11 @@ type Interface interface {
 }
 
 type ipset struct {
+	refCount
 }
 
 func New() Interface {
-	return &ipset{}
+	return &ipset{refCount: newRefCount()}
 }
 
 func (i *ipset) Create(ipsetName Name, ipsetType Type) error {
@@ -38,10 +39,16 @@ func (i *ipset) Create(ipsetName Name, ipsetType Type) error {
 }
 
 func (i *ipset) AddEntry(ipsetName Name, entry string) error {
+	if i.inc(ipsetName, entry) > 1 { // already in the set
+		return nil
+	}
 	return doExec("add", string(ipsetName), entry)
 }
 
 func (i *ipset) DelEntry(ipsetName Name, entry string) error {
+	if i.dec(ipsetName, entry) > 0 { // still needed
+		return nil
+	}
 	return doExec("del", string(ipsetName), entry)
 }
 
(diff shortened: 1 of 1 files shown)


bboreham:
I have manually tested this as described at https://github.com/weaveworks/weave/issues/2792#issuecomment-278941846

npc/ipset/ipset.go
```diff
@@ -67,3 +78,38 @@ func doExec(args ...string) error {
...
+	return rc.ref[k]
+}
+
+func (rc *refCount) dec(ipsetName Name, entry string) int {
+	k := key{ipsetName, entry}
+	rc.ref[k]--
+	return rc.ref[k]
+}
+
+func (rc *refCount) removeSet(ipsetName Name) {
```
marccarre: *Minor*: we could probably make `removeSet`:

- `O(1)` with a `map[Name]map[string]int` instead of 
- `O(n)` with a `map[key]int`.

marccarre:
LGTM assuming:
```
// note no locking is required as all operations are serialised in the controller
``` 
indeed holds, but you may want to have someone else with more knowledge of Weave Net review it.

npc/ipset/ipset.go
bboreham: This would come at the cost of making `inc` and `dec` much more complicated and slightly slower.
In this program `removeSet` is only called when a network policy is updated or removed, which are expected to be infrequent operations.

npc/ipset/ipset.go
```diff
@@ -67,3 +78,38 @@ func doExec(args ...string) error {
...
 	return nil
 }
+
+// Reference-counting
+type key struct {
+	ipsetName Name
+	entry     string
+}
+
+// note no locking is required as all operations are serialised in the controller
```
marccarre: Not sure if this matters or not -- still looking at the code and the Go memory model -- but:
```
func (npc *controller) onNewNsSelector(selector *selector) error
``` 
does not lock like other method calls which do:
```
	npc.Lock()
	defer npc.Unlock()
```

npc/ipset/ipset.go
marccarre: Fair enough, let's not do it then.

npc/ipset/ipset.go
bboreham: What is the code path to get there, that has not already been through a lock?

marccarre: *Minor*: the fact this was relatively long & painful to track is maybe a hint that we should do it differently.

npc/ipset/ipset.go
```diff
@@ -27,37 +27,48 @@ type Interface interface {
...
+	i.removeSet(ipsetName)
 	return doExec("flush", string(ipsetName))
 }
 
 func (i *ipset) FlushAll() error {
+	i.refCount = newRefCount()
 	return doExec("flush")
 }
 
 func (i *ipset) Destroy(ipsetName Name) error {
```
brb: Is it possible for `Destroy` to be called twice with the same `ipsetName`?

bboreham: I don't think so, from the logic in `provision` and `deprovision`, but also I don't see why it matters.

brb: > I don't see why it matters.

Destroying the same `ipsetName` results in `EEXIST`.

bboreham: Oh, ok, the `ss.users` lookup effectively plays the same part as a reference-count for creation and deletion of ipset names.

PR #2796, opened 2017-02-13
Add versions manifest file for dependencies.

marccarre:
**Changelog**:

- One unique file to list all dependencies and dependencies' versions (except Go, currently defined in `build/Dockerfile`): `DEPENDENCIES`.
- Custom versions can now be injected into Vagrant VMs.
- Vagrant dev VMs, Vagrant test VMs, and `run-integration-tests.sh` all source from `DEPENDENCIES`, to keep things 'DRY'.
- Vagrant versions can be overriden using environment variables, like it was possible in `run-integration-tests.sh`.
- Fix bug in `get_go_version_from_build_dockerfile`.

**Manual tests**:

```
weave (versions-manifest)$ cat DEPENDENCIES 
DOCKER_VERSION=1.11.2
KUBERNETES_VERSION=1.5.2
KUBERNETES_CNI_VERSION=0.3.0.1
```

1. Default (dev VM):
```
weave (versions-manifest)$ vagrant up
weave (versions-manifest)$ vagrant ssh
[...]
vagrant@vagrant:~$ docker version
Client:
 Version:      1.11.2
[...]
```

2. Override (dev VM, without `export`):
```
weave (versions-manifest)$ DOCKER_VERSION="1.13.1" vagrant up
weave (versions-manifest)$ vagrant ssh
[...]
vagrant@vagrant:~$ docker version
Client:
 Version:      1.13.1
[...]
```

3. Override (test VMs, using `export`):
```
weave/test (versions-manifest)$ export DOCKER_VERSION="1.13.1"
weave/test (versions-manifest)$ vagrant up
weave/test (versions-manifest)$ vagrant ssh host1
[...]
vagrant@host1:~$ docker version
Client:
 Version:      1.13.1
[...]
```


marccarre:
Bryan's comments:
> It looks pretty complicated to me.  Doesn't Ansible have some sort of way to set variables in a file?
> You appear to have picked a file format that is not convenient for either Bash or Ansible to read

Here is why I took this approach:
1) Ansible needs YAML, and needs it in specific folders: there are conventions on its "inventory", related variables and folder structure,
2) Ansible currently has these defaults in `build-tools`: https://github.com/weaveworks/build-tools/blob/51785b5fc923a6db5d2ee5ae7bffca97d43d1c66/config_management/group_vars/all
3) however, given Ansible's defaults are shared cross-projects, and since we should also be able to override these easily from higher-level scripts which means, at least to me, either env. var. or CLI arguments, I then initially added these in `run-integration-tests.sh` https://github.com/weaveworks/weave/blob/master/test/run-integration-tests.sh#L32-L34 but didn't have anything, at the time, for Vagrant.
4) **if** we do not want this `DEPENDENCIES` file and the logic (`grep` and some Ruby code) to load it, but still want to define defaults in Weave Net in one place, **then** we would need a YAML file, and either:

    1) set a "hook" like in: http://docs.ansible.com/ansible/playbooks_variables.html#variable-file-separation but that means the file is mandatory
    2) have a set of tasks checking for the existence of the file, and if present, loading it:
http://docs.ansible.com/ansible/include_vars_module.html
    3) fiddle with Ansible's current variables file, e.g. replace it dynamically or have a symlink to Weave Net's override instead of using `tools/config_management/all`,

    none of which I find very elegant (maybe ii. isn't too bad), and none of which allow for an easy override via environment variables, which means this would have to be extra logic in Ansible.
  
    IMO, the only benefit of having ii. and dealing with environment variables in Ansible, is that it may avoid other projects having to replicate:
    - https://github.com/weaveworks/weave/blob/master/test/run-integration-tests.sh#L32-L34 and 
    - `get_dependencies_version_from_file_and_env` and `ansibleize` in their `Vagrantfile`s.
    but this would be at the cost of being explicit, and instead relying on a convention deeply defined in Ansible.

Happy to change how this is done, but here are the trade-offs I can think of.

bboreham:
If Ansible can read the file format at https://github.com/weaveworks/build-tools/blob/51785b5fc923a6db5d2ee5ae7bffca97d43d1c66/config_management/group_vars/all, then why not stick with that format?  It would be all the same to Bash.

marccarre:
The fact Ansible can read YAML files is not enough to be able to override values, 

- to override "shared" values from `tools/config_management/group_vars/all` with "Weave Net" values from another YAML file, you'd have to do one of 4.i./ii./iii. above;
- to override "Weave Net" values from environment variables, you'd have to port the logic I currently have in this PR to Ansible (see end of 4. above).

Let me know if you'd want to see how that is done to compare it with this PR.

PR #2812, opened 2017-02-23
Use a keyserver port that's better for firewalls

mikebryant:
It doesn't really matter if this makes it into master, but it can help improve the contributor experience in some organisations. - so feel free to close it if you're not interested :)

(Our corporate firewall doesn't allow the hkp port out, but does allow 80, so this change lets me run the weave build)


bboreham:
Thanks for the contribution, @mikebryant !

We need to research the effects this change will have before approving.

marccarre:
LGTM.
Many thanks for the PR @mikebryant, this may indeed be useful to other people!

For the record, I've checked a couple things about HKP, `pcap`-ed HKP requests/reponses before/after the change, and this doesn't seem to introduce any security vulnerability, hence the above "LGTM".
See also discussion at: https://weaveworks.slack.com/archives/net/p1488466417009653

Issue #2751, opened 2017-01-25
ipset: Set cannot be destroyed when running alongside keepalived-vip on the same host


stephan2012:
When [keepalived-vip](https://github.com/kubernetes/contrib/tree/master/keepalived-vip) and Weave run on the same host, the weave-npc container constantly crashes:

```
# docker logs -f a95eb6b0429ea321c5494f8321413b082c5f3593dbbcea4b1dda1e5a7f140c0b 
time="2017-01-25T10:17:01Z" level=info msg="Starting Weaveworks NPC 1.8.2" 
time="2017-01-25T10:17:01Z" level=info msg="Serving /metrics on :6781" 
Wed Jan 25 10:17:01 2017 <5> ulogd.c:843 building new pluginstance stack: 'log1:NFLOG,base1:BASE,pcap1:PCAP'
time="2017-01-25T10:17:02Z" level=fatal msg="ipset [destroy] failed: ipset v6.29: Set cannot be destroyed: it is in use by a kernel component\n: exit status 1" 
```

Apparently Weave does not expect ipsets created by others. Only solution to bring Weave back working at the moment is to remove keepalived-vip and reboot the host.

Any ideas on how to make Weave co-exist with keepalived-vip on the same host?

Environment: Kubernetes 1.5, CentOS 7.3

awh:
> Any ideas on how to make Weave co-exist with keepalived-vip on the same host?

Yes, fix this: https://github.com/weaveworks/weave/blob/master/prog/weave-npc/main.go#L75

marccarre:
Closed by #2846 (for some reason it didn't get automatically closed by the PR).

PR #2813, opened 2017-02-23
Destroy ipsets prefixed `weave-` only (fixes issue #2751)

stephan2012:
Changes to destroy weave's ipsets only so that it can co-exist with other tools that use ipsets (e.g., keepalived-vip).

### DIFF
--- a/npc/ipset/ipset.go
+++ b/npc/ipset/ipset.go
@@ -1,7 +1,9 @@
 package ipset
 
 import (
+	"bytes"
 	"os/exec"
+	"strings"
 
 	"github.com/pkg/errors"
 )
@@ -22,6 +24,8 @@ type Interface interface {
 	Flush(ipsetName Name) error
 	Destroy(ipsetName Name) error
 
+	List(prefix string) ([]Name, error)
+
 	FlushAll() error
 	DestroyAll() error
 }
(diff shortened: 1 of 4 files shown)


npc/ipset/ipset.go
```diff
@@ -35,48 +39,77 @@ func New() Interface {
...
+	err, _ := doExec("destroy")
+	return err
+}
+
+// Fetch a list of all existing sets
+func (i *ipset) List(prefix string) ([]Name, error) {
+	err, output := doExec("list","-name","-output","plain")
+
+	var selected []Name
+	if err == nil && len(output) > 0 {
```
bboreham: The Go idiom is to test for error and return, thus leaving the happy path un-indented.

npc/ipset/ipset.go
```diff
@@ -35,48 +39,77 @@ func New() Interface {
...
+		output = bytes.TrimRight(output, "\n")
+		sets := strings.Split(string(output[:]), "\n")
```
bboreham: why `[:]` ?

npc/ipset/ipset.go
```diff
@@ -35,48 +39,77 @@ func New() Interface {
...
+
+		plen := len(prefix)
+		for _, v := range sets {
+			if (plen <= len(v)) && (prefix == v[:len(prefix)]) {
```
bboreham: `strings.HasPrefix(v, prefix)`

npc/ipset/ipset.go
```diff
@@ -35,48 +39,77 @@ func New() Interface {
...
+
+	}
+
+	return selected, err
 }
 
-func doExec(args ...string) error {
-	if output, err := exec.Command("ipset", args...).CombinedOutput(); err != nil {
-		return errors.Wrapf(err, "ipset %v failed: %s", args, output)
+func doExec(args ...string) (error, []byte) {
```
bboreham: Also it gave a `gofmt` error:

npc/ipset/ipset.go:107:1: error should be the last type when returning multiple items

bboreham:
Thanks for the contribution, @stephan2012 ! 

I've made some comments in a review.

npc/ipset/ipset.go
```diff
@@ -35,48 +39,77 @@ func New() Interface {
...
+	return err
+}
+
+// Fetch a list of all existing sets
+func (i *ipset) List(prefix string) ([]Name, error) {
+	err, output := doExec("list","-name","-output","plain")
+
+	var selected []Name
+	if err == nil && len(output) > 0 {
+		output = bytes.TrimRight(output, "\n")
```
stephan2012: Without stripping the final newline we have an empty element in the list of prefixes.

bboreham: which will then fail the `HasPrefix` test

prog/weave-npc/main.go
```diff
@@ -80,14 +80,22 @@ func resetIPTables(ipt *iptables.IPTables) error {
...
 		return err
 	}
 
-	if err := ips.DestroyAll(); err != nil {
-		return err
+	 common.Log.Debugf("Got list of ipsets: %v", sets)
+
+	for _, s := range sets {
+		common.Log.Debugf("Destroying ipsets '%s'", string(s))
+		if err := ips.Destroy(s); err != nil {
```
stephan2012: I've tested it and ipset does not require flushing first:

```
# ipset --version
ipset v6.19, protocol version: 6

# ipset create foo bitmap:ip range 192.168.1.0/24  
# ipset add foo 192.168.1.11
# ipset add foo 192.168.1.12
# ipset add foo 192.168.1.13
# ipset destroy foo
```

npc/ipset/ipset.go
stephan2012: Ok, clearly looks better.

npc/ipset/ipset.go
stephan2012: That way you have execs in different places.

I've haven't seen any error when compiling it. The patched npc container is actually working in my Kubernetes setup. Maybe a Go version thing?

stephan2012:
@bboreham: Thanks for your feedback. The idiomatic stuff is related to me being a newbie in Go…

What are the expectations regarding this PR?

npc/ipset/ipset.go
bboreham: the error comes when you run `gofmt -s -w npc/ipset/ipset.go`, which we do as part of the CI

bboreham:
If you fix up the points we can merge it.  Or someone else will fix them up.

Thanks!

One more thought: it would be good to have a test for the extended functionality, e.g. create an `ipset` before running, then check it's still there afterwards, in `test/840_weave_kube_3_test.sh`.  But to execute that test you'd need to set up 3 VMs with Kubernetes installed, which is quite a lot of faff.  (We have scripts to do it, but they have dependencies, etc.)

bboreham:
I made the changes that I suggested, and added a test: https://github.com/weaveworks/weave/compare/stephan2012-master

CI build passed at https://circleci.com/gh/weaveworks/weave/8207
I also checked that the test fails on existing 1.9: https://circleci.com/gh/weaveworks/weave/8208

brb:
@bboreham Could you open PR with your changes?

marccarre:
Closing this PR as #2846 has been merged instead, but thanks again for the contribution @stephan2012!

PR #2814, opened 2017-02-23
Fetch list of container IDs inside weaveutil

bboreham:
This makes `weave ps` slightly cheaper, and will help Weave Scope to stop embedding the docker binary.

### DIFF
--- a/prog/weaveutil/addrs.go
+++ b/prog/weaveutil/addrs.go
@@ -27,10 +27,25 @@ func containerAddrs(args []string) error {
 		return err
 	}
 
+	containerArgs := []string{}
+	for _, cid := range args[1:] {
+		if cid == "weave:allids" { // expand to all container IDs
+			all, err := client.ListContainers(docker.ListContainersOptions{All: true})
+			if err != nil {
+				return err
+			}
+			for _, c := range all {
+				containerArgs = append(containerArgs, c.ID)
+			}
+		} else {
+			containerArgs = append(containerArgs, cid)
+		}
+	}
+
 	var containerIDs []string
 	containers := make(map[string]*docker.Container)
 
-	for _, cid := range args[1:] {
+	for _, cid := range containerArgs {
 		if cid == "weave:expose" {
 			netDev, err := weavenet.GetBridgeNetDev(bridgeName)
 			if err != nil {
--- a/weave
+++ b/weave
@@ -999,12 +999,12 @@ wait_for_status() {
 
 # Call $1 for all containers, passing container ID, all MACs and all IPs
 with_container_addresses() {
-    COMMAND=$1
+    COMMAND_WCA=$1
     shift 1
 
     CONTAINER_ADDRS=$(util_op container-addrs $BRIDGE "$@") || return 1
     echo "$CONTAINER_ADDRS" |  while read CONTAINER_ID CONTAINER_IFACE CONTAINER_MAC CONTAINER_IPS; do
-        $COMMAND "$CONTAINER_ID" "$CONTAINER_IFACE" "$CONTAINER_MAC" "$CONTAINER_IPS"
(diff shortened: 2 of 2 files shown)


weave
```diff
@@ -1009,7 +1009,7 @@ with_container_addresses() {
 }
 
 echo_addresses() {
-    echo $1 $3 $4
+    echo $(printf "%.12s" $1) $3 $4
```
brb: Nitpick: `printf "%.12s %s %s\n" $1 $3 $4`?

prog/weaveutil/addrs.go
```diff
@@ -27,10 +27,25 @@ func containerAddrs(args []string) error {
 		return err
 	}
 
+	containerArgs := []string{}
+	for _, cid := range args[1:] {
+		if cid == "weave:allids" { // expand to all container IDs
```
brb: Nitpick, but I'd probably implement `weave:allids` as a flag, i.e. `weaveutil container-addrs --all`. It could also cover `weave:expose`.

weave
bboreham: In the case that `$3` and/or `$4` are blank, this will output extra spaces compared to the original.

And the smoke-test checks for exact string equality.

prog/weaveutil/addrs.go
bboreham: `allids` is used in two places: one requires `expose` and the other does not, so I don't think having it do `expose` implicitly is a win.  Maybe you meant change to two flags?

brb: OK, I see.

I thought about introducing `--all` flag instead of special container id `weave:allids`. However, given that one place does not require `weave:expose`, in the other place we would need to call `weaveutil container-addrs --all weave:expose` which IMO looks a bit confusing.

bboreham:
@brb is there anything still outstanding on this PR ?

weave
```diff
@@ -999,17 +999,17 @@ wait_for_status() {
 
 # Call $1 for all containers, passing container ID, all MACs and all IPs
 with_container_addresses() {
-    COMMAND=$1
+    COMMAND_WCA=$1
```
brb: Naming nitpick: s/COMMAND_WCA/CMD?

weave
```diff
@@ -1218,12 +1218,12 @@ when_weave_running() {
 # and $3.. as additional arguments
 with_container_fqdn() {
     CONT="$1"
-    COMMAND="$2"
+    COMMAND_WCF="$2"
```
brb: Naming nitpick: s/COMMAND_WCF/CMD?

weave
```diff
@@ -1713,14 +1713,14 @@ fetch_router_args() {
...
+
 populate_router() {
     if [ -z "$NO_DNS_OPT" ] ; then
         # Tell the newly-started weaveDNS about existing weave IPs
-        for CONTAINER in $(docker ps -q --no-trunc) ; do
-            if CONTAINER_IPS=$(with_container_addresses echo_ips $CONTAINER) && [ -n "$CONTAINER_IPS" ] ; then
-                with_container_fqdn $CONTAINER put_dns_fqdn $CONTAINER_IPS
-            fi
-        done
+        with_container_addresses populate_one weave:allids
```
brb: Nitpick, but we are populating multiple entries, thus `populate_one` is a bit misleading.

brb:
Besides a few nitpicks, LGTM.

weave
bboreham: The point of this commit is to avoid having the same variable name in three places, where it can overwrite if called re-entrantly.  So changing two of them to `CMD` leaves us with the same danger.

The name is perhaps ugly but was deliberately chosen to avoid clashes.

weave
bboreham: one container; multiple addresses

bboreham: I renamed it to `populate_dns_for_one_container`

bboreham:
Test 840 was failing; rebased to pick up changes to that

marccarre:
LGTM.

Issue #2817, opened 2017-02-27
reverse DNS broken when query uses upper case


tobowers:
Running weave 1.8.2, Docker 1.13.1, centos 7.2, aws, elastic container service

I noticed a problem while trying to cluster two rabbitmqs using an autocluster plugin and DNS. That plugin requires reverse-dns and I was able to isolate down to weave/erlang combination.

Forward DNS works fine in both docker/weave erlang consoles (only reverse DNS has a problem).

A docker-only (no weave) container will correctly resolve reverse dns from the erlang console:

```
Eshell V8.2.2  (abort with ^G)
1>  inet_res:gethostbyaddr({216,58,217,174}).
{ok,{hostent,"iad23s44-in-f14.1e100.net",[],inet,4,
             [{216,58,217,174}]}}
```

A container started through the weave proxy will *not* resolve reverse DNS.

```
Eshell V8.2.2  (abort with ^G)
1> inet_res:gethostbyaddr({216,58,217,174}).
{error,nxdomain}
```

However, bash on the container started through weave proxy *can* do the reverse DNS.

```
[root@ip-myip ~]# echo $DOCKER_HOST
unix:///var/run/weave/weave.sock
[root@ip-myip ~]# docker run -it --rm erlang /bin/bash -l
root@8974630db404:/# getent hosts 216.58.217.174
216.58.217.174  iad23s44-in-f174.1e100.net
```

I'm not familiar enough with weave/erlang internals to give this a proper diagnosis.

bboreham:
@tobowers can you give the command-line used to start the container in the second example?

Since that doesn't look like a Weave address, the issue is in the recursive call to the upstream DNS.  `weave status` should say what the upstream DNS is set to; if that looks right then run Weave in debug mode, re-do the reverse DNS test and upload the logs (`docker logs weave`) please.

I.e. `weave stop`; `weave launch --log-level=debug`

tobowers:
@bboreham thanks for looking. I discovered the problem using weave addresses but I thought the above google ip was going to be easier to diagnose. Here's an example using just weave addresses. It'll take me a sec to launch in debug mode... it's not as trivial as that command in my ECS-backed-by-systemd setup.

```
[root@ip-172-16-2-244 ~]# eval "$(weave env)"
[root@ip-172-16-2-244 ~]# docker run -it --rm erlang
[...]

PR #2818, opened 2017-02-27
lowercase IP string before trimming in reverse DNS lookup (see #2817)

tobowers:
Some DNS clients send reverse lookups with uppercase suffixes. This lowercases the request before trimming the lowercase suffix. Problem discovered and discussed here: https://github.com/weaveworks/weave/issues/2817

I could not find any specific reverse dns tests to create/update. The current DNS tests all pass.

### DIFF
--- a/nameserver/dns.go
+++ b/nameserver/dns.go
@@ -222,7 +222,7 @@ func (h *handler) handleReverse(w dns.ResponseWriter, req *dns.Msg) {
 		return
 	}
 
-	ipStr := strings.TrimSuffix(req.Question[0].Name, "."+reverseDNSdomain)
+	ipStr := strings.TrimSuffix(strings.ToLower(req.Question[0].Name), "."+reverseDNSdomain)
 	ip, err := address.ParseIP(ipStr)
 	if err != nil {
 		h.nameError(w, req)
--- a/test/245_dns_case_insensitive_test.sh
+++ b/test/245_dns_case_insensitive_test.sh
@@ -6,6 +6,9 @@ C1=10.2.0.78
 C2=10.2.0.79
 C3=10.2.0.80
 
+REVERSE_C1_LOWER=78.0.2.10.in-addr.arpa
+REVERSE_C1_UPPER=78.0.2.10.IN-ADDR.ARPA
+
 start_suite "DNS lookup case (in)sensitivity"
 
 weave_on $HOST1 launch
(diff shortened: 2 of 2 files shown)


bboreham:
Thanks for the PR!

There are many reverse DNS tests, but all of them work by calling `getent hosts <IP>`.  Either `test/265_dns_reverse_test.sh` or `test/245_dns_case_insensitive_test.sh` would seem suitable homes for a test for this particular bug.

Still pondering your change.  Maybe we should just cut the right number of characters off the end?  We ought to be guaranteed that the domain is correct at the time we enter the callback, by the DNS library.

tobowers:
hmm... The right number of characters feels like it would be slightly confusing in code to me... easiest to read would probably be a regex (but that probably has performance bottle necks). Idiomatic golang for "trim after 4th dot" sounds like it would be longer.

maybe a small function that trims it which could use number of characters (but still provide readability). The compiler will inline the function anyway so shouldn't be a performance penalty.

Are there any docs for running those bash tests?

bboreham:
The instructions to run the bash tests are in https://github.com/weaveworks/weave/blob/master/test/README.md

The most important bit, if you don't want to run every single script the same way we do is:

> set the environment variable HOSTS to the space-separated list of IP addresses of the docker hosts, 
> and set the environment variable SSH to a command that will log into either (which may just be ssh).

Note that both of those test scripts only need one host.

Also there is a `-v` flag that is useful when things go wrong, i.e. `./245_dns_case_insensitive_test.sh -v`

tobowers:
hmmm. Thanks for your help... I'm getting:

```
$ HOSTS=192.168.99.100 SSH="docker-machine ssh default" ./245_dns_case_insensitive_test.sh -v
Cleaning up on 192.168.99.100: removing all containers and resetting weave
13:30 ~/code/go/src/github.com/weaveworks/weave/test (fix_reverse_dns) $ echo $?
1
```

Locally I'm on docker 1.11.2 using docker toolbox (can't upgrade to docker for mac just yet)

tobowers:
and
```
HOSTS=192.168.99.100 SSH="docker-machine ssh default" ./setup.sh 
./setup.sh: line 10: ./tls: cannot execute binary file
```

I'm probably going to have to leave this for a bit and go back to the real job :)

bboreham:
Sorry; if you want to push the non-tested test we can fix it up.

You should be able to `make test/tls/tls`, or it will build as one of the first steps of `make run-smoketests`

tobowers:
I added a take on a reverse test in 245_dns_case_insensitive_test.sh . Implementation is currently the same as it was before (using lower). If you think the number of character approach is better I can get to that later today / tomorrow morning.

bboreham:
I fixed up the test, [ran the CI](https://circleci.com/gh/weaveworks/weave/8139) and merged in 242b8907bfbc60cf358fe386e89801e8af4662fc.  Thanks!

tobowers:
Thank you! I'm trying to find the commits where the fixed test are merged, do you have a link?

bboreham:
Sorry, the merge is in 9db22c3cd6d3e1abc023386b6fc3a9b727d07fd0 and the specific change to the test is in e57d79bd709fe00717b74453da0bde7f0c8e269e

I updated the comment above where I'd put the wrong commit

PR #2819, opened 2017-02-27
Ensure we have the right modules loaded for weave-npc

### LINKED ISSUE #2726: weave-kube does not work with `dockerd --iptables=false`
This configuration of docker is recommended in https://github.com/kelseyhightower/kubernetes-the-hard-way/blob/master/docs/05-kubernetes-worker.md; the problem is that we rely on docker to load the `br_netfilter` module, and it doesn't do so when `--iptables=false`.

On further investigation it turns out that loading the module from inside the weave-kube container is pretty straightforward; Alpine's busybox already implements `modprobe`, and other than that we need `-v /lib/modules:/lib/modules` and one of `--privileged` or `--cap-add SYS_MODULE`:

```
$ sudo docker run --cap-add SYS_MODULE -v /lib/modules:/lib/modules -ti alpine rmmod br_netfilter
$ sudo docker run --cap-add SYS_MODULE -v /lib/modules:/lib/modules -ti alpine lsmod|grep br_netfilter
$ sudo docker run --cap-add SYS_MODULE -v /lib/modules:/lib/modules -ti alpine modprobe br_netfilter
$ sudo docker run --cap-add SYS_MODULE -v /lib/modules:/lib/modules -ti alpine lsmod|grep br_netfilter
br_netfilter           24576  0 
bridge                126976  1 br_netfilter
```

bboreham:
Fixes #2726 

At least when `weave-npc` is executed via the `weave-kube` daemonset.  It would be nicer to separate those two concerns, but we already need to know that `weave-npc` is in use when creating the bridge.

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -15,6 +15,12 @@ CONN_LIMIT=${CONN_LIMIT:-30}
 # Default for network policy
 EXPECT_NPC=${EXPECT_NPC:-1}
 
+# Ensure we have the required modules for NPC
+if [ "${EXPECT_NPC}" != "0" ]; then
+    modprobe br_netfilter
+    modprobe xt_set
+fi
+
 # kube-proxy requires that bridged traffic passes through netfilter
 if ! BRIDGE_NF_ENABLED=$(cat /proc/sys/net/bridge/bridge-nf-call-iptables); then
     echo "Cannot detect bridge-nf support - network policy and iptables mode kubeproxy may not work reliably" >&2
--- a/docs/weavenpc-design.md
+++ b/docs/weavenpc-design.md
@@ -19,6 +19,9 @@ NetworkPolicy object updates from the k8s API server:
   network policy, containing the IP addresses of all pods in the
   namespace whose labels match that selector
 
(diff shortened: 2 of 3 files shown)


bboreham:
I have tested this by hand, by modifying the Docker daemon parameters to include `--iptables=false` and starting up `weave-kube`.

prog/weave-kube/launch.sh
```diff
@@ -15,6 +15,12 @@ CONN_LIMIT=${CONN_LIMIT:-30}
 # Default for network policy
 EXPECT_NPC=${EXPECT_NPC:-1}
 
+# Ensure we have the required modules for NPC
+if [ "${EXPECT_NPC}" != "0" ]; then
+    modprobe br_netfilter
```
weitzj: I am seeing errors like:

`modprobe: module br_netfilter not found in modules.dep`

I guess this feature works only, if the module is **not compiled into** the kernel itself.
See a related discussion https://github.com/kubernetes/kubernetes/issues/12459

bboreham: Sorry; we are fixing this in #2821

PR #2821, opened 2017-02-28
Ignore modprobe errors in weave-kube/launch.sh

### LINKED ISSUE #2820: modprobe: module br_netfilter not found in modules.dep on CentOS7
Hi,

Just pulled in weave 1.9.1 via https://git.io/weave-kube on CentOS7

```
image: weaveworks/weave-kube:1.9.1
```

The container `weave` does not come up because it complains about the missing `br_netfilter` module:

```
modprobe: module br_netfilter not found in modules.dep
```

The container is right that the module is not there, because it is built in in the CentOS default kernel:

```bash
ls -l /proc/sys/net/bridge
total 0
-rw-r--r--. 1 root root 0 Feb 28 09:31 bridge-nf-call-arptables
-rw-r--r--. 1 root root 0 Feb 28 09:31 bridge-nf-call-ip6tables
-rw-r--r--. 1 root root 0 Feb 28 09:31 bridge-nf-call-iptables
-rw-r--r--. 1 root root 0 Feb 28 10:05 bridge-nf-filter-pppoe-tagged
-rw-r--r--. 1 root root 0 Feb 28 10:05 bridge-nf-filter-vlan-tagged
-rw-r--r--. 1 root root 0 Feb 28 10:05 bridge-nf-pass-vlan-input-dev
```

It worked with older versions of `weave`, so maybe there was a recent change which causes this?

brb:
`weave-kube` is run in `alpine` image which uses busybox. `modprobe` of busybox will return the `"module not found"` error, if a module is built-in, which is an opposite behavior to `modprobe` distributed in mainstream distros (e.g. Ubuntu).

https://bugs.busybox.net/show_bug.cgi?id=5270

Fix #2820

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -3,7 +3,20 @@
 set -e
 
 modprobe_safe() {
-    modprobe $1 || echo "Ignore the error if \"$1\" is built-in in the kernel"
+    modprobe $1 || echo "Ignore the error if \"$1\" is built-in in the kernel" >&2
+}
+
+# Check whether xt_set actually exists
+xt_set_exists() {
+    ipset create weave-kube-test hash:ip
+    iptables -t filter -N WEAVE-KUBE-TEST
+    if ! iptables -A WEAVE-KUBE-TEST -m set --match-set weave-kube-test src -j DROP; then
+        NOT_EXIST=1
+    fi
+    iptables -F WEAVE-KUBE-TEST
+    iptables -X WEAVE-KUBE-TEST
+    ipset destroy weave-kube-test
+    [ -z "$NOT_EXIST" ] || (echo "\"xt_set\" does not exist" >&2 && return 1)
 }
 
 # Default if not supplied - same as weave net default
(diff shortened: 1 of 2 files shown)


prog/weave-kube/launch.sh
```diff
@@ -17,8 +21,8 @@ EXPECT_NPC=${EXPECT_NPC:-1}
 
 # Ensure we have the required modules for NPC
 if [ "${EXPECT_NPC}" != "0" ]; then
-    modprobe br_netfilter
-    modprobe xt_set
+    modprobe_safe br_netfilter
+    modprobe_safe xt_set
```
rmohr: Note that this one is actually a module on CentOS.

As far as I see there is an explicit check for br_netfilter further down. Does it also make sense to check for xt_set somehow? Just to not be too forgiving ...

brb: Thanks for comment. I've added an indirect check for `xt_set`.

brb:
Tested manually on CentOS 7. PTAL @bboreham

prog/weave-kube/launch.sh
```diff
@@ -17,9 +34,10 @@ EXPECT_NPC=${EXPECT_NPC:-1}
...
 fi
+xt_set_exists
```
bboreham: this should be inside the `if`

prog/weave-kube/launch.sh
```diff
@@ -2,6 +2,23 @@
 
 set -e
 
+modprobe_safe() {
+    modprobe $1 || echo "Ignore the error if \"$1\" is built-in in the kernel"
```
bboreham: echo should go to stderr

PR #2825, opened 2017-03-02
Do not reset fastdp heartbeat timer for non-confirmed connections

### LINKED ISSUE #2824: Segfault in fastdp.go
2/5 of the weave nodes in my cluster have started crashing.

This was on 1.9.0, upgrading to 1.9.2 hasn't helped.

We're using weave-kube, with additional options `CHECKPOINT_DISABLE`, `IPALLOC_RANGE`, and `WEAVE_PASSWORD`
We're also mounting `weavedb` as a `hostPath` instead of `emptyDir`, to provide consistency on whole-cluster reboots

```
core@kubernetes-kubernetes-az1-0-1487773414 ~ $ uname -a
Linux kubernetes-kubernetes-az1-0-1487773414 4.7.3-coreos-r2 #1 SMP Thu Feb 2 02:26:10 UTC 2017 x86_64 Intel Core Processor (Haswell) GenuineIntel GNU/Linux
```

[weave-crash-logs-1.9.2.txt](https://github.com/weaveworks/weave/files/811306/weave-crash-logs-1.9.2.txt)
[weave-crash-logs-1.9.0.txt](https://github.com/weaveworks/weave/files/811305/weave-crash-logs-1.9.0.txt)

brb:
It is possible to receive `FastDatapathCryptoInitSARemote` control plane message before `Confirm`ing a connection. In such case, the handler for `FastDatapathCryptoInitSARemote` tries to reset `heartbeatTimer` which is nil, thus it makes `weaver` to crash. The timer is initialized in `Confirm`. 

The fix is to skip the reset for such situations, as the timer will be reset in `Confirm` (both handling `<..>InitSARemote` and `Confirm` are mutually exclusive).

Fix #2824

### DIFF
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -577,8 +577,8 @@ type fastDatapathForwarder struct {
 	confirmed         bool
 	remoteAddr        *net.UDPAddr
 	heartbeatInterval time.Duration
-	heartbeatTimer    *time.Timer
-	heartbeatTimeout  *time.Timer
+	heartbeatTimer    *time.Timer // for sending
+	heartbeatTimeout  *time.Timer // for receiving
 	ackedHeartbeat    bool
 	stopChan          chan struct{}
 	stopped           bool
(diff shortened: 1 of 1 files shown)


router/fastdp.go
```diff
@@ -856,7 +856,9 @@ func (fwd *fastDatapathForwarder) handleCryptoInitSARemote(msg []byte) {
 		return
 	}
 
-	if !fwd.isOutboundIPSecEstablished {
+	// FastDatapathCryptoInitSARemote can be received before Confirm'ing
+	// connection, thus before InitSALocal.
+	if fwd.confirmed && !fwd.isOutboundIPSecEstablished {
```
bboreham: interested to know why you do not (also) test `heartbeatTimer != nil`

brb: `fwd.confirmed` denotes that `Confirm` has been successfully called, and `Confirm` ensures the initialization of `heatbeatTimer`.

PR #2827, opened 2017-03-06
Peer Discovery via cloud service

### LINKED ISSUE #2799: Peer discovery via Weave Cloud
As a user I want to give each peer a token to simplify cluster configuration, for instance `weave launch --token ABCDEABCDEABCDE`.

A peer started in such way would connect to Weave Cloud to retrieve list of peers to which it has to connect.

Before starting implementing this feature we need to:

1. Refine requirements.
2. Create a design doc.

bboreham:
Fixes #2799, covering the 'Day One' and "day 1" user stories:

- [x] Obtaining a token for user’s first cluster - this is identical to obtaining a token for a Scope instance today
- [x] Obtaining a token for user’s nth cluster - ditto
- [x] Configuring Weave Net given a token
- [x] Peers that can see other peers directly form a Weave Net cluster
- [x] Removing a dead peer from a cluster, via the CLI
- [x] documentation


prog/weaver/main.go
```diff
@@ -184,6 +187,9 @@ func main() {
 	mflag.StringVar(&trustedSubnetStr, []string{"-trusted-subnets"}, "", "comma-separated list of trusted subnets in CIDR notation")
 	mflag.StringVar(&dbPrefix, []string{"-db-prefix"}, "/weavedb/weave", "pathname/prefix of filename to store data")
 	mflag.BoolVar(&isAWSVPC, []string{"-awsvpc"}, false, "use AWS VPC for routing")
+	mflag.StringVar(&discoveryEndpoint, []string{"-peer-discovery-url"}, "https://cloud.weave.works/api/weavenet", "url for peer discovery")
```
marccarre: *Nitpick*: why not `cloud.weave.works/api/net`?

prog/weaver/main.go
```diff
@@ -257,7 +263,21 @@ func main() {
...
+			localAddrs, err := weavenet.LocalAddresses()
+			checkFatal(err)
+			for _, addr := range localAddrs {
+				addresses = append(addresses, addr.IP.String())
+			}
+		} else {
+			addresses = strings.Split(advertiseAddress, ",")
+		}
+		discoveredPeers, err := peerDiscoveryUpdate(discoveryEndpoint, token, name.String(), nickName, addresses)
+		checkFatal(err) // TODO: what if it is a transient error, and we are restarting?
```
marccarre: *Minor*: ICYMI, you may want to deal with this `TODO` out before we merge the PR?

prog/weaver/discovery.go
```diff
@@ -79,6 +89,12 @@ func HandleHTTPPeer(router *mux.Router, alloc *ipam.Allocator) {
 
 	router.Methods("DELETE").Path("/peer/{id}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ident := mux.Vars(r)["id"]
+		if discoveryEndpoint != "" {
+			// TODO: deal with this being either a peername or a nickname
```
marccarre: *Minor*: ICYMI, you may want to deal with this `TODO` out before we merge the PR?

prog/weaver/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+)
+
+// TODO: move these definitions somewhere more shareable
```
marccarre: *Minor*: ICYMI, you may want to deal with this `TODO` out before we merge the PR?

prog/weaver/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+	body := new(bytes.Buffer)
+	err := json.NewEncoder(body).Encode(request)
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequest(verb, discoveryEndpoint+"/peer", body)
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Scope-Probe token=%s", token))
```
marccarre: Just for my understanding, why "`Scope-`"?

prog/weaver/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+	req.Header.Set("X-Weave-Net-Version", version)
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
```
marccarre: Just for my understanding, why "`application/x-www-form-urlencoded`" and not "`application/json`" when we actually send JSON?

prog/weaver/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+	return err
+}
+
+func peerDiscoveryUpdate(discoveryEndpoint, token, peername, nickname string, addresses []string) ([]string, error) {
+	request := PeerUpdateRequest{
+		Name:      peername,
+		Nickname:  nickname,
+		Addresses: addresses,
+	}
+	var updateResponse PeerUpdateResponse
```
marccarre: Out of curiosity, what is the advantage of creating the response outside of `do` and passing it as an input vs. having `do` return a tuple `(PeerUpdateResponse, error)` like `([]string, error)` in this method?

prog/weaver/discovery.go
```diff
@@ -0,0 +1,66 @@
...
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		rbody, _ := ioutil.ReadAll(resp.Body)
+		return errors.New(resp.Status + ": " + string(rbody))
+	}
+	if response == nil {
```
marccarre: Given `response` is passed in to `do`, this could be moved at the very beginning of the method, as it is "input validation" and so that we avoid any work.

prog/weaver/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+		if alloc != nil {
+			alloc.Shutdown()
+		}
+		w.WriteHeader(204)
+	})
+
+	router.Methods("DELETE").Path("/peer/{id}").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ident := mux.Vars(r)["id"]
+		if discoveryEndpoint != "" {
+			// TODO: deal with this being either a peername or a nickname
```
marccarre: *Minor*: ICYMI, you may want to deal with this `TODO` out before we merge the PR?

prog/weaver/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/gorilla/mux"
+
+	"github.com/weaveworks/weave/ipam"
+)
+
+// TODO: move these definitions somewhere more shareable
```
marccarre: *Minor*: ICYMI, you may want to deal with this `TODO` out before we merge the PR?

marccarre:
LGTM overall. 
Just a few minor comments, and some questions regarding naming, style/structure/Go idioms and remaining `TODO`s.

site/using-weave/weave-cloud.md
```diff
@@ -0,0 +1,124 @@
...
+
+This section contains the following topics: 
+
+ * [Launching Weave Net](#launching)
+ * [Creating Peer Connections Between Hosts](#peer-connections)
+ * [Testing Container Communications](#testing)
+ * [Starting the Netcat Service](#start-netcat)
+
+
+###<a name="launching"></a>Signing in to Weave Cloud
```
marccarre: A whitespace should be added for titles.
Jekyll, for the new website, apparently does not like them w/o whitespaces (see: #2889).

prog/weaver/discovery.go
bboreham: It's hard-wired into the Weave Cloud authorization flow - https://github.com/weaveworks/service/issues/1127 suggests that it be changed to something more neutral

Also I think this line should be hidden behind a helper - see https://github.com/weaveworks/common/issues/29

prog/weaver/discovery.go
bboreham: bug, thanks

prog/weaver/discovery.go
bboreham: `do()` is generic; it doesn't know what the type of the response is.
Though, with only `PeerUpdateResponse` and `nil` in the current code this is a minimal benefit.

prog/weaver/discovery.go
bboreham: We do want to make the call when `response` is nil; we just don't want to decode any json coming back.  We expect a `StatusOK` success code or an error.

prog/weaver/discovery.go
marccarre: +1 to both, makes a lot of sense to:
- rename it,
- hide details,
- factorise / avoid duplication.

Thanks for clarifying and pointing to these issues.

prog/weaver/discovery.go
marccarre: I see, thanks for the explanation. 
Such genericity may come too early, but at least it avoids refactoring later on, so LGTM.

prog/weaver/discovery.go
marccarre: I didn't expect `do` to be used this way, but thanks for clarifying it.

bboreham:
Another issue: I am equating one address with one peer, but actually each peer can register more than one address.  So the calculation for paxos quorum will be wrong.

prog/weaver/main.go
bboreham: I changed `weavenet` to `net` in URLs, thanks

site/using-weave/weave-cloud.md
```diff
@@ -0,0 +1,124 @@
...
+ * [Launching Weave Net](#launching)
+ * [Launching on Additional Hosts](#peer-connections)
+ * [How Peer Discovery Works](#how-it-works)
+
+### <a name="signing-in"></a>Signing in to Weave Cloud
+
+Weave Cloud is a fast and simple way to visualize, manage and monitor
+containers and microservices.  In order to use Weave Cloud to set up
+Weave Net, you must sign up to Weave Cloud and create an "Instance".
+
```
abuehrle: Are we using Weave Cloud to set up Net?  Or can we visualize the network once it's in Weave Cloud?

site/using-weave/weave-cloud.md
```diff
@@ -0,0 +1,124 @@
...
+The instance will have a Service Token which is a 32-character string
+used to uniquely identify it.  Get hold of the Service Token for your
+Instance.
+
```
abuehrle: I've been calling this a cloud token, but I see it is service token in the UI. 

Could use my screen capture here (https://github.com/weaveworks/guides/blob/master/images/weave-cloud-token.png)

bboreham:
> the calculation for paxos quorum will be wrong

I fixed this by including the peer count in the return from the service.

site/using-weave/weave-cloud.md
bboreham: The point of this PR is the former - instead of typing in other hosts to connect to, you use Weave Cloud.
There already is some visualisation, which we can improve over time.

bboreham:
> We need to reconcile what we call them probes [...] The Docker Store is calling them 'Cloud Agents' 

This feature has nothing to do with probes.  You install the Weave Net "router", plugin and/or proxy as before.  I would not want to call any of those "Cloud Agents".

site/using-weave/weave-cloud.md
bboreham: Thanks; added the screen cap

prog/weaver/discovery.go
bboreham: moved to #2891

prog/weaver/discovery.go
bboreham: moved to #2891

abuehrle:
>This feature has nothing to do with probes. You install the Weave Net "router", plugin and/or proxy as before. I would not want to call any of those "Cloud Agents".

That's not what I'm talking about.  The UI where you get your Service Token, says launch the probes with this token.  Yet in the Docker Store they are called Agents, and as you point out here, none of these things has anything to do with Weave Net.  Maybe you could also file a bug to that effect, since my doing so has no effect.

bboreham:
The Weave Cloud UI for this feature is at https://github.com/weaveworks/service-ui/pull/396; let's discuss UI points there.

marccarre:
_Usability_: this can be improved later, but note that if an user passes in an invalid token by mistake, the container just dies impolitely:
```
$ weave launch-router --peer-discovery-url=https://frontend.dev.weave.works/api/net --token=<token>
$ docker logs weave
[...]
FATA: 2017/04/11 13:44:11.646674 401 Unauthorized:
```

bboreham:
See #2891 "better error handling"

marccarre:
LGTM.

PR #2829, opened 2017-03-06
Only check running containers for existing addresses

### LINKED ISSUE #2815: findExistingAddresses is checking dead containers
Because I re-used the `allContainerIDs` list we already had, but this contains live and dead containers (for a good reason).  

This doesn't stop it working; just that it would be more efficient to concentrate on the live ones when looking for IP addresses.

Here: https://github.com/weaveworks/weave/blob/f7ce16cba88a31d190727b9e5219de721554e18e/prog/weaver/main.go#L314

dlmiddlecote:
Fixes #2815

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -306,20 +306,18 @@ func main() {
 			}
 			trackerName = "awsvpc"
 		}
-		var allContainerIDs []string
-		var runningContainerIDs []string
-		if dockerCli != nil {
-			allContainerIDs, err = dockerCli.AllContainerIDs()
-			checkFatal(err)
 
-			runningContainerIDs, err = dockerCli.RunningContainerIDs()
-			checkFatal(err)
-		}
-		preClaims, err := findExistingAddresses(dockerCli, runningContainerIDs, weavenet.WeaveBridgeName)
+		preClaims, err := findExistingAddresses(dockerCli, weavenet.WeaveBridgeName)
 		checkFatal(err)
+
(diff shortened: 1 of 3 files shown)


dlmiddlecote:
Hey - I'm not sure why the checks failed.

```
Cannot run smoke tests: no secret key
```

```
can't load package: package _/home/ubuntu/weave/api
```

I'm not sure if these are my issues.

marccarre:
Many thanks for the PR @dlmiddlecote! -- someone will review it soon.

No worries about `Cannot run smoke tests: no secret key` and the subsequent error, it is just that, for security reasons, people outside the Weaveworks organisation cannot automatically run integration tests on Weaveworks' CIT servers.
I've just pushed your changes and you should be able to check the build status here: [weave/8192](https://circleci.com/gh/weaveworks/weave/8192).

prog/weaver/main.go
```diff
@@ -307,11 +307,15 @@ func main() {
 			trackerName = "awsvpc"
 		}
 		var allContainerIDs []string
+		var runningContainerIDs []string
```
bboreham: Better to have these lists inside the scope they are used in.
If you look at 7075b5d9fd6872c647633e9ec9580e25753dc901 where `allContainerIDs` was introduced, the variable was hoisted so it could be used in two places.  But not you're removing that need so the code should go back more like it was before.
In the same vein, `runningContainerIDs` should be inside `findExistingAddresses`.

bboreham:
Thanks!

PR #2850, opened 2017-03-15
Upgrade cni library to version 0.5.0

bboreham:
Prompted by the impending release of Kubernetes 1.6, which seems to rely on the behaviour that a blank netns is allowed on DEL.

Lots of lines changed, because the CNI return format changed.  The library should take care of returning backwards-compatible results to older runtimes.

### DIFF
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -7,9 +7,10 @@ import (
 	"net"
 	"syscall"
 
-	"github.com/appc/cni/pkg/ipam"
-	"github.com/appc/cni/pkg/skel"
-	"github.com/appc/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/ipam"
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 	weaveapi "github.com/weaveworks/weave/api"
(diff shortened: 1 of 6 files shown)


jbeda:
I think one problem is that there is no cniVersion member in the weave configuration:
```
cat /etc/cni/net.d/10-weave.conf
{
    "name": "weave",
    "type": "weave-net"
}
```

I suspect (no confirmation) that the runtime then assumes the latest version.  It might be that setting `"cniVersion": "0.2.0"` there fixes things.

bboreham:
That's an interesting idea, but the CNI spec says that an absence of version shall be interpreted as version 0.1, and the currently-released Weave Net plugin is at CNI spec level 0.1. 

As discussed on sig-network Slack, kubelet needs to be prepared for the DEL to error, and move on past that operation at some point.

plugin/net/cni.go
```diff
@@ -70,28 +77,31 @@ func (c *CNIPlugin) CmdAdd(args *skel.CmdArgs) error {
 	if err != nil {
 		return fmt.Errorf("unable to allocate IP address: %s", err)
 	}
+	// Only expecting one address
+	ip := result.IPs[0]
```
marccarre: Is "only expecting one address" guaranteed at this point?
Should we log a warning or return an error if we get more than one?

bboreham: This is a valid question.  However I am not sure we can do much better:
* `getIP` errors on zero addresses.
* there is no mechanism to log a warning in CNI.

luxas:
I can confirm that this fixes the weave-kube problem with v1.6.
However, we should bump the version of weave-plugin so the new plugin binary will be used; when I tested this I had to manually remove the current weave CNI plugin binary in order to get the new one deployed

bboreham:
Thanks @luxas; we will make a new release soon which will have a higher version number.

luxas:
Was this merged into master as well?
Is there any reason to not fix this on master as well?
I'd like to be able to use the CI-built (`:latest`) version to actually test it out and make sure it's working from official builds

bboreham:
@luxas merged to master now; sorry for oversight

However the CI build flaked first time

PR #2851, opened 2017-03-16
Upgrade to latest build-tools to be able to test against 1.6+ and/or "unstable" versions of K8S

marccarre:
Integrate latest changes required for testing against Kubernetes 1.6.0 and/or "unstable" versions (`beta`, `alpha`).
See also: weaveworks/build-tools/pull/81


bboreham:
Wouldn't we want this on the 1.9 branch?

marccarre:
I based the change on `master` as I wasn't sure when/where we would want it. Changed to `1.9` now.

PR #2853, opened 2017-03-17
Update smoke tests for kubernetes 1.6

bboreham:
With this change test 840 works with the latest kubeadm but fails against 1.5; ideally it would be updated to work on both.


brb:
Needs rebasing.

bboreham:
Rebased, and tests passed

PR #2855, opened 2017-03-19
WIP: DNS entries added/removed on network connect/disconnect

### LINKED ISSUE #1914: Use Docker network events in the plugin

dlmiddlecote:
This is an initial PR, to get feedback.

@bboreham I'm not sure what should happen in the Proxy ContainerDisconnect method if there is an error attaching the container.

Also, in the allocator, I added a new map for disconnected containers, instead of treating them as dead. I guess one issue is (`disconnect -> died -> start with --net`) will leave the container in the disconnected map, even though it is connected.

TODO:
- [x] Proxy ContainerDisconnect Implementation
- [x] Improve tests
- [ ] Support for `docker network connect --alias`

Fixes #1914

### DIFF
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -119,12 +119,10 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
 						ob.ContainerDestroyed(event.ID)
 					case "network:connect":
 						if containerID, ok := event.Actor.Attributes["container"]; ok {
-							pending.finish(containerID)
 							ob.ContainerConnected(containerID)
 						}
 					case "network:disconnect":
 						if containerID, ok := event.Actor.Attributes["container"]; ok {
-							pending.finish(containerID)
 							ob.ContainerDisconnected(containerID)
 						}
 					}
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -21,6 +21,8 @@ type ContainerObserver interface {
 	ContainerStarted(ident string)
 	ContainerDied(ident string)
 	ContainerDestroyed(ident string)
+	ContainerConnected(ident string)
+	ContainerDisconnected(ident string)
 }
 
 type Client struct {
@@ -115,6 +117,12 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
 					case "destroy":
 						pending.finish(event.ID)
 						ob.ContainerDestroyed(event.ID)
+					case "network:connect":
+						pending.finish(event.ID)
+						ob.ContainerConnected(event.ID)
+					case "network:disconnect":
+						pending.finish(event.ID)
+						ob.ContainerDisconnected(event.ID)
 					}
 				}
 				if time.Since(start) > retryInterval {
--- a/common/docker/client.go
+++ b/common/docker/client.go
@@ -118,11 +118,15 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
 						pending.finish(event.ID)
 						ob.ContainerDestroyed(event.ID)
 					case "network:connect":
-						pending.finish(event.ID)
-						ob.ContainerConnected(event.ID)
+						if containerID, ok := event.Actor.Attributes["container"]; ok {
+							pending.finish(containerID)
+							ob.ContainerConnected(containerID)
+						}
 					case "network:disconnect":
-						pending.finish(event.ID)
-						ob.ContainerDisconnected(event.ID)
+						if containerID, ok := event.Actor.Attributes["container"]; ok {
+							pending.finish(containerID)
+							ob.ContainerDisconnected(containerID)
+						}
 					}
 				}
 				if time.Since(start) > retryInterval {
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -24,10 +24,9 @@ const (
 	msgRingUpdate
 	msgSpaceRequestDenied
 
-	tickInterval                 = time.Second * 5
-	MinSubnetSize                = 4 // first and last addresses are excluded, so 2 would be too small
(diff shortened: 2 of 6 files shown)


common/docker/client.go
```diff
@@ -118,11 +118,15 @@ func (c *Client) AddObserver(ob ContainerObserver) error {
 						pending.finish(event.ID)
 						ob.ContainerDestroyed(event.ID)
 					case "network:connect":
-						pending.finish(event.ID)
-						ob.ContainerConnected(event.ID)
+						if containerID, ok := event.Actor.Attributes["container"]; ok {
+							pending.finish(containerID)
```
bboreham: Don't assume that `connect` has anything to do with starting a container.
We should still get the `start` event, which should still do what it did before.

bboreham:
Issue #1914 is about the plugin.  Look here: https://github.com/weaveworks/weave/blob/master/plugin/net/watcher.go#L43

ipam/allocator.go
```diff
@@ -378,6 +381,36 @@ func (alloc *Allocator) ContainerStarted(ident string) {
...
+func (alloc *Allocator) ContainerConnected(ident string) {
+	alloc.actionChan <- func() {
+		delete(alloc.disconnected, ident) // delete is no-op if key not in map
+	}
+}
+
+func (alloc *Allocator) ContainerDisconnected(ident string) {
+	alloc.actionChan <- func() {
+		if alloc.hasOwnedByContainer(ident) {
+			alloc.debugln("Container", ident, "disconnected; noting to remove later")
```
bboreham: We don't really want this because Docker tells us directly via the plugin when it is connecting or disconnecting a container.

nameserver/nameserver.go
```diff
@@ -137,6 +138,19 @@ func (n *Nameserver) ContainerDied(ident string) {
 	n.broadcastEntries(entries...)
 }
 
+func (n *Nameserver) ContainerDisconnected(ident string) {
+	n.Lock()
+	entries := n.entries.tombstone(n.ourName, func(e *Entry) bool {
+		if e.ContainerID == ident {
+			n.infof("container %s disconnected; tombstoning entry %s", ident, e.String())
```
bboreham: Need to think about the case that a container is connected twice.

bboreham:
Sorry, hadn't read down far enough.

proxy/proxy.go
```diff
@@ -404,6 +404,14 @@ func (proxy *Proxy) ContainerStarted(ident string) {
 	proxy.notifyWaiters(ident, err)
 }
 
+func (proxy *Proxy) ContainerConnected(ident string) {
+	err := proxy.attach(ident)
+	// if err != nil {
+	// TODO: Not sure what is needed here.
```
bboreham: Nothing.  We only expect container-connected events when going via the plugin, not the proxy.

bboreham:
Thanks for the contribution; I think you're on the right track in the plugin, but stay clear of the proxy.

common/docker/client.go
dlmiddlecote: I will remove the calls to `pending.finish(...)`, as this will not stop `ob.ContainerStarted(...)` being called.

ipam/allocator.go
dlmiddlecote: What don't we want? The delaying of removing the disconnected containers, or any of the `Allocator` changes?

proxy/proxy.go
dlmiddlecote: Ok

ipam/allocator.go
bboreham: Any of them.

nameserver/nameserver.go
dlmiddlecote: Is it possible for this to happen? It seems that docker stops this from happening.

plugin/net/watcher.go
```diff
@@ -58,3 +58,31 @@ func (w *watcher) ContainerDied(id string) {
 }
 
 func (w *watcher) ContainerDestroyed(id string) {}
+
+func (w *watcher) ContainerConnected(id string) {
+	w.driver.debug("ContainerConnected", "%s", id)
+	info, err := w.client.InspectContainer(id)
```
bboreham: This code should replace the `ContainerStarted` version, not add to it.

Also it should use the [information inside the event](https://github.com/docker/docker/blob/d8406fd7a029f38ece55421294568045d1c9de92/daemon/events.go#L91) about what network was connected, which avoids having to loop round hunting for it.

nameserver/nameserver.go
bboreham: It can be connected to multiple networks, of which several can map to the same plugin.

Without having thought about it deeply, I suspect the right thing to do is remove the specific address in `ContainerDisconnected` in the plugin, and not handle `ContainerDisconnected` here.

bboreham:
@dlmiddlecote are you still working on this?  Do you need more help?

dlmiddlecote:
Hi @bboreham - sorry that this has been hanging around for so long... I definitely want to get this finished for you. Can you give me some pointers on where to continue? Thanks, Dan.

bboreham:
I'm going to close due to inactivity; feel free to reopen when ready.

PR #2856, opened 2017-03-20
Ensure we don't send containerID to IPAM plugin when requesting bridge IP

### LINKED ISSUE #2839: CNI ADD: weave bridge and the first connected container get the same IP addr
When doing CNI ADD, we allocate and assign an IP addr not only to a container, but also to the weave bridge, if it does not have any IP addr (https://github.com/weaveworks/weave/blob/1.9/plugin/net/cni.go#L77).

A function which calls IPAM allocate might not use ContainerID from `args` passed to the function (https://github.com/weaveworks/weave/blob/1.9/plugin/net/cni.go#L51), and instead it will use the one from `CNI_CONTAINERID` env var which contains the container id.

As a result, we will end up calling IPAM allocate with the same container id twice (instead of the container id and `weave:expose`),  which will allocate the same IP addr to the weave bridge.

This might explain the weave-kube bug reports from the past, where the weave bridge and some container ended up having the same IP addr. However, to make this happen, the first CNI allocation should have happened before executing https://github.com/weaveworks/weave/blob/1.9/prog/weave-kube/launch.sh#L115 .

bboreham:
Fixes #2839 

Plus a couple of improvements encountered when testing.

### DIFF
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -54,6 +54,9 @@ func (c *CNIPlugin) getIP(ipamType string, args *skel.CmdArgs) (newResult *curre
 	if err != nil {
 		return nil, err
 	}
+	if result == nil {
+		return nil, fmt.Errorf("Received no usable result from IPAM plugin")
+	}
 	newResult, err = current.NewResultFromResult(result)
 	// Check if ipam returned no results without error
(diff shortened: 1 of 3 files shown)


marccarre:
LGTM.

Issue #2843, opened 2017-03-13
Code deficiencies of various kinds and severities


dominikh:
Hi,

I ran staticcheck on weave at commit bec2470d, filtered out benign issues and false positives and assembled the following report, grouped by category and sorted by severity in descending order. I hope this report is useful to you.

# Subtle bugs/potential for wrong behaviour

| Location | Issue |
|----------|-------|
|[prog/plugin/main.go:97:39](https://github.com/weaveworks/weave/blob/bec2470ddfdbb6db5aef12e690f5aff54617e2b7/prog/plugin/main.go#L97) | os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (SA1016)|
|[prog/weaveutil/docker_tls_args.go:48:5](https://github.com/weaveworks/weave/blob/bec2470ddfdbb6db5aef12e690f5aff54617e2b7/prog/weaveutil/docker_tls_args.go#L48) | ineffective break statement. Did you mean to break out of the outer loop? (SA4011)|
|[prog/weaveutil/unique_id.go:19:13](https://github.com/weaveworks/weave/blob/bec2470ddfdbb6db5aef12e690f5aff54617e2b7/prog/weaveutil/unique_id.go#L19) | printf-style function with dynamic first argument and no further arguments should use print-style function instead ([SA1006](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck#sa1006--printf-with-dynamic-first-argument-and-no-further-arguments))|
|[ipam/allocator.go:215:6](https://github.com/weaveworks/weave/blob/bec2470ddfdbb6db5aef12e690f5aff54617e2b7/ipam/allocator.go#L215) | identical expressions on the left and right side of the '==' operator (SA4000)|
|[router/sleeve.go:1041:5](https://github.com/weaveworks/weave/blob/bec2470ddfdbb6db5aef12e690f5aff54617e2b7/router/sleeve.go#L1041) | this value of err is never used (SA4006)|
|[router/sleeve.go:1062:41](https://github.com/weaveworks/weave/blob/bec2470ddfdbb6db5aef12e690f5aff54617e2b7/router/sleeve.go#L1062) | use net.IP.Equal to compare net.IPs, not bytes.Equal ([SA1021](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck#sa1021--using-bytesequal-to-compare-two-netip))|
[...]

PR #2857, opened 2017-03-20
Fix various small issues revealed by staticcheck

bboreham:
This covers most of #2843, with the exception of the obsolete http client used in the Docker proxy.

I sense that is either done that way on purpose or needs extra thought to update.


marccarre:
LGTM: good improvements to have merged in IMO.

marccarre:
This PR aside, we already do `go vet` in [`lint`](https://github.com/weaveworks/build-tools/blob/master/lint#L83), but would it be worth adding [`staticcheck`](https://github.com/dominikh/go-tools) too/instead?

bboreham:
Moved discussion to https://github.com/weaveworks/build-tools/issues/82

PR #2859, opened 2017-03-21
README: Fix formatting of "Further Information" header

mgalgs:


errordeveloper:
@marccarre is Circle failing because this effectively run on @mgalgs' fork?

bboreham:
@errordeveloper yes, we do not include the secrets required to spin up a full test on PRs from forks. It's a defence against people bitcoin-mining on our test VMs.

PR #2860, opened 2017-03-21
README: Drop Slack Status badge

mgalgs:
The Slack Status badge doesn't work, at least for
non-signed-in users.  This gives a bad impression
as to the quality of the project.

Here's what it looks like for me, for example:

image


errordeveloper:
@mgalgs thanks for fixing it! This LGTM, but I wonder if anyone else has any comments.

bboreham:
Thanks!  It broke because we moved the inviter site to a different implementation.

PR #2861, opened 2017-03-21
Update weave-kube docs for k8s  1.6

bboreham:

PR #2876, opened 2017-03-31
Disable temporary the 840 weave kube smoke test

brb:
This should trigger the docs update (#2861).

#2853 will re-enable the smoke test.


bboreham:
This was reverted for the 1.9 branch by #2886

PR #2885, opened 2017-04-05
Remove 'weave run', 'start' and 'restart'

### LINKED ISSUE #2351: weave run interprets container args

### LINKED ISSUE #2353: ditch 'weave run|start|restart'
I doubt very many users are using `weave run|start|restart`, and instead are using the proxy or plugin. The lack of "waiting for the interface" in `weave run` is an achilles heel (which is one reason we introduced the proxy), and it suffers from other issues such as #2351.

Is there anything `weave run|start|restart` can do that the proxy cannot?

bboreham:
Fixes #2353 
Fixes #2351 

We should announce this ahead of the release.


bboreham:
I should put a 'this command has been removed' message in, to save confusion.

rade:
`ipam_cidrs_or_die` is no longer used.

bboreham:
Also docs need amending

rade:
@abuehrle pls check there aren't any guides that reference these commands.

weave
```diff
@@ -2009,8 +2009,12 @@ EOF
         NO_MULTICAST_ROUTE=
         collect_cidr_args "$@"
         shift $CIDR_ARG_COUNT
+        WITHOUT_DNS=
         while [ $# -gt 0 ]; do
             case "$1" in
+                --without-dns)
+                    WITHOUT_DNS=1
+                    ;;
```
rade: Is this intentionally undocumented / left out from the usage?

bboreham: fixed

test/280_with_or_without_dns_test.sh
```diff
@@ -12,7 +12,9 @@ check_dns() {
     chk=$1
     shift
 
-    container=$(weave_on $HOST1 run $@ $IP/24 -t --add-host=$STATIC:$STATIC_IP $DNS_IMAGE /bin/sh)
+    container=$(docker_on $HOST1 run $(dns_args $HOST1 "$@") -dt $DNS_IMAGE /bin/sh)
+    weave_on $HOST1 attach $IP/24 "$@" --rewrite-hosts --add-host=$STATIC:$STATIC_IP $container
+    echo $chk $HOST1 $container $TARGET
```
marccarre: Given there is `$chk $HOST1 $container $TARGET` on line 18, is this:
a) just a temporary add which was accidentally checked in?
b) something we want to conditionally print, e.g. `[ -z "$DEBUG" ] || echo ...`? or
c) something we want to always print from now onwards?

bboreham: an accident

marccarre: Thanks for the confirmation. I'll remove it and then merge this PR.

PR #2893, opened 2017-04-11
IPAM: Make sure we have a ring before calling tryOps()

### LINKED ISSUE #2892: IPAM tryOps() can get called recursively
This call chain is possible:
```
Allocator.tryPendingOps()
Allocator.tryOps()
Claim.Try()
Allocator.establishRing()
Allocator.createRing()
Allocator.ringUpdated()
Allocator.tryPendingOps()
Allocator.tryOps()
```

then when we return all the way through that the slice has been modified by the inner `tryOps` call.
If the slice is now shorter than where `i` is pointing, we get a bounds check panic.  If it's still longer we discard whatever operation `i` is pointing at now.

bboreham:
because otherwise the pending ops can be modified from inside Try()

Fixes #2892

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -265,6 +265,10 @@ func (alloc *Allocator) tryPendingOps() {
 	alloc.tryOps(&alloc.pendingAllocates)
 }
 
+func (alloc *Allocator) havePendingOps() bool {
+	return len(alloc.pendingPrimes)+len(alloc.pendingClaims)+len(alloc.pendingAllocates) > 0
+}
+
 func (alloc *Allocator) spaceRequestDenied(sender mesh.PeerName, r address.Range) {
 	for i := 0; i < len(alloc.pendingClaims); {
 		claim := alloc.pendingClaims[i].(*claim)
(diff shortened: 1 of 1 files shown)


marccarre:
LGTM.

ipam/allocator.go
```diff
@@ -696,8 +700,12 @@ func (alloc *Allocator) actorLoop(actionChan <-chan func(), stopChan <-chan stru
...
 			if alloc.awaitingConsensus {
 				alloc.propose()
-			} else {
-				alloc.tryPendingOps()
+			} else if alloc.havePendingOps() {
+				if alloc.ring.Empty() {
+					alloc.establishRing()
+				} else {
+					alloc.tryPendingOps()
+				}
```
rade: why is that whole else branch not just
```
} else {
   alloc.establishRing()
   alloc.tryPendingOps()
}
```
?

bboreham: `establishRing()` may not actually establish a ring; it may just send out some messages.  And not calling `tryPendingOps()` until we have a ring is the whole point of this change.  We could probably assume that combination is ok, because of what we know about the implementation of the consensus protocol, but I'd rather not.

`if alloc.havePendingOps()` is to defer establishing a ring until there is some need for it; this is primarily useful in the case that we don't know the right number of peers, and are still forming connections.

rade: I updated my suggestion earlier, after realising the need for the `alloc.havePendingOps()` so we don't establish a ring unnecessarily.

How about
```
} else if alloc.havePendingOps() {
    alloc.establishRing()
    if !alloc.ring.Empty() {
        alloc.tryPendingOps()
    }
}
```
?

bboreham: > surely we'd want to call tryPendingOps() immediately after establishing the ring

`ringUpdated()` calls `tryPendingOps()`, so that isn't necessary.  Nonetheless your version is one line shorter, and perhaps clearer, so I changed it.

rade: ah, good point. Now I like your version better :(

bboreham: ok, turns out my `git push` failed, so I haven't changed it :-)

marccarre:
LGTM.

PR #2727, opened 2017-01-10
Create a V2 Docker plugin

bboreham:
Fixes #2397 #2816 #2651 #2805

Build the plugin via `make v2plugin`; requires Docker 1.13.1 (1.13.0 does not work)

This branch is based on top of the docker-swarm branch used in #2721 to get the extra entry-points, although it doesn't seem to make any difference.

Items outstanding:
- [x] [Connecting up peers](https://github.com/weaveworks/weave/issues/2805) - currently you have to `weave connect` to test
- [x] Docker don't support `--pid=host` on v2 plugins; need another way to do the arp settings on the bridge
- [x] No access to the Docker socket inside the plugin, so the hack to do arp settings on the veth doesn't work
- [x] No access to the Docker socket inside the plugin, so need to persist anything we are told on network creation (e.g. the multicast option)
- [x] `weave status` does work for the router, but doesn't see the plugin
- [x] `weave launch` should detect if this plugin is installed and exit with error
- [x] Better UX for troubleshooting - can't do `docker logs` to see what went wrong

PR #2897, opened 2017-04-13
More help for upgraders who use plugin commands

bboreham:
Print helpful message; also stop and reset old weaveplugin container in case user has upgraded from <v2.0 and left it running.

per https://github.com/weaveworks/weave/pull/2727/files#r109466258, https://github.com/weaveworks/weave/pull/2727/files#r109468485 and https://github.com/weaveworks/weave/pull/2727/files#r109468670


bboreham:
Wasn't planning to add anything.

marccarre:
Related to #2727.

PR #2899, opened 2017-04-13
Add website preview via Netlify

errordeveloper:
With the new website in the works, we are gonna have full documentation and website builds for feature branches and pull-requests.

During the website build, docs are pull in from this and Scope repos, that's done for a given revision and some of the mechanics are still up for a debate.

The easiest way to enable preview sites on Netlify for this and Scope repo was to create a Netlify site for each of these repos and pull-in website repo during the build. Here you can see how this is done. The website repo is currently private, hence a token is required to fetch the code, which is configured in Netlify.


.ruby-version
```diff
@@ -0,0 +1 @@
+2.3.0
```
errordeveloper: This is required for Netlify build to work, hope folks don't mind adding this little file. It will need to be kept in sync with website repo, but I don't suppose this will change too often. I've managed to work around having to add any of the other config files.

Makefile
```diff
@@ -404,3 +404,11 @@ integration-tests: all testrunner
 #     integration-tests
 #
 	RUNNER_ARGS="-parallel" ./test/run-integration-tests.sh
+
+netlify-site-preview:
+	@mkdir -p site-build
+	@curl --user $(WEBSITE_GITHUB_USER) --silent 'https://codeload.github.com/weaveworks/website-next/tar.gz/$(WEBSITE_BRANCH)' \
```
errordeveloper: Both, `WEBSITE_GITHUB_USER` and `WEBSITE_BRANCH` are in Netlify configuration. I'm using tarball API, as far as I know it's faster then `git clone` (even when you pass `--depth 10 --single-branch`), but happy to change if we have a good reason.

errordeveloper:
@awh you should have an invite to Netlify also, so you can see how things are configured there.

errordeveloper:
As you can see down below, there is now a Netlify link to the newly built site.

You can see one of the documentation pages here: https://deploy-preview-2899--weaveworks-weave-docs-preview.netlify.com/docs/net/latest/introducing-weave/ (more work needed on layouts and other stuff, but as far as I'm concerned, this is already really great)

Makefile
```diff
@@ -404,3 +404,11 @@ integration-tests: all testrunner
 #     integration-tests
 #
 	RUNNER_ARGS="-parallel" ./test/run-integration-tests.sh
+
+netlify-site-preview:
+	@mkdir -p site-build
```
marccarre: 1. Do we want to delete this folder at the end of the `netlify-site-preview` target?
2. Do we want to `git`-ignore it?

errordeveloper: We can delete it, but this is only intended to run in Netlify CI, and it won't work anywhere else without a bunch of dependencies. If you need to build locally, the best way to do it is via website repo.

errordeveloper: Perhaps I should the above as a comment in the Makefile, WDYT?

marccarre: I think a comment regarding dependencies/where it is intended to run/etc. would indeed be useful as it doesn't seem obvious 👍 
I would still add an entry for `site-build` to `.gitignore` as we don't want to have this checked in by mistake.

errordeveloper:
@marccarre I've added `site-build` to `.gitignore` and added a comment in `Makefile`.

.gitignore
```diff
@@ -71,3 +71,6 @@ test/coverage.*
 terraform.tfstate
 terraform.tfstate.backup
 prog/net-plugin/rootfs/
+
+# Website
+site-build
```
marccarre: 👍

errordeveloper:
> Coverage decreased (-2.1%) to 68.278%

Why?

errordeveloper:
I've checked with @brb and it sounds like the coveralls is somewhat flaky, clearly these changes have nothing to do with it.

PR #2901, opened 2017-04-18
Remove deprecation warnings for things removed a long time ago

bboreham:
Since we are making breaking changes in 2.0, it seems a good opportunity to clean out some old code.

I left in `--init-peer-count` since it's not as old as the others.


marccarre:
LGTM.
Quick `grep`-ing of the site/docs show they are up-to-date and do not mention any deprecated CLI option, so nothing else to change as far as I know, and good to be merged IMO.

PR #2903, opened 2017-04-18
Update weave-kube docs and YAML files to leverage Launch Generator

### LINKED ISSUE #2754: Launch generator for weave-kube
Currently, if a user wants to configure some parameter of weave-kube, such as the IP allocation range, we advise that they download a copy of the standard config, edit it and run from that edited copy. This is complex, error-prone and risks falling out of sync with changes in the standard config. 

Instead, we could generate the desired config on demand from a cloud service.

marccarre:
### Context:

The current Kubernetes YAML for `weave-kube`

1. is static, which forces people to manually edit the file in order to customise it, to add environment variables, etc. and
2. exists in two versions, one for Kubernetes `v1.5`, and one for `v1.6` and above, which forces users to make choices about how they install `weave-kube`, hence increasing mental burden.

The latest version of the Launch Generator supports, among others: 
- passing additional options, arguments, and environment variables (weaveworks/launch-generator#66), which addresses 1. above;
- passing the output of `kubectl version` to automatically extract Kubernetes' "server version" and return appropriate YAML configuration (weaveworks/launch-generator#79) without having users to care about it, which addresses 2. above.

### Changelog:

1. Update documentation to cover point 1. from the above "Context" section.
2. Update documentation to cover point 2. from the above "Context" section.
3. Update Weave-Kube daemonset YAML files to standardise configuration based on Launch Generator's output.

### Related work:
- weaveworks/launch-generator#66
- weaveworks/launch-generator#79
- weaveworks/launch-generator#88
- weaveworks/launch-generator#91

This PR and the above ones:
- fix weaveworks/launch-generator#43
- fix #2754

cc: @errordeveloper

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
@@ -1,121 +1,128 @@
-kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
-metadata:
-  name: weave-net
-rules:
-- apiGroups:
-  - ""
-  resources:
-  - pods
-  - namespaces
-  - nodes
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups:
-  - extensions
-  resources:
-  - networkpolicies
-  verbs:
-  - get
-  - list
-  - watch
----
 apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: weave-net
-  namespace: kube-system
----
-kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
-metadata:
-  name: weave-net
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: weave-net
-subjects:
-- kind: ServiceAccount
-  name: weave-net
-  namespace: kube-system
----
-apiVersion: extensions/v1beta1
-kind: DaemonSet
-metadata:
-  name: weave-net
-  namespace: kube-system
-spec:
-  template:
+kind: List
+items:
+  - apiVersion: v1
+    kind: ServiceAccount
     metadata:
+      name: weave-net
+      labels:
+        name: weave-net
+  - apiVersion: rbac.authorization.k8s.io/v1beta1
+    kind: ClusterRole
+    metadata:
+      name: weave-net
+      labels:
+        name: weave-net
+    rules:
+      - apiGroups:
+          - ''
+        resources:
+          - pods
+          - namespaces
+          - nodes
+        verbs:
+          - get
+          - list
+          - watch
+      - apiGroups:
+          - extensions
+        resources:
+          - networkpolicies
+        verbs:
+          - get
+          - list
+          - watch
+  - apiVersion: rbac.authorization.k8s.io/v1beta1
+    kind: ClusterRoleBinding
+    metadata:
+      name: weave-net
+      labels:
+        name: weave-net
+    roleRef:
+      kind: ClusterRole
+      name: weave-net
+      apiGroup: rbac.authorization.k8s.io
+    subjects:
+      - kind: ServiceAccount
+        name: weave-net
+        namespace: kube-system
+  - apiVersion: extensions/v1beta1
+    kind: DaemonSet
+    metadata:
+      name: weave-net
       labels:
         name: weave-net
     spec:
-      hostNetwork: true
-      hostPID: true
-      containers:
-        - name: weave
-          image: weaveworks/weave-kube:latest
-          imagePullPolicy: Always
-          command:
-            - /home/weave/launch.sh
-          livenessProbe:
-            initialDelaySeconds: 30
-            httpGet:
-              host: 127.0.0.1
-              path: /status
-              port: 6784
+      template:
+        metadata:
+          labels:
+            name: weave-net
+        spec:
+          containers:
+            - name: weave
+              command:
+                - /home/weave/launch.sh
+              env: []
+              image: 'weaveworks/weave-kube:latest'
+              imagePullPolicy: Always
+              livenessProbe:
+                httpGet:
+                  host: 127.0.0.1
+                  path: /status
+                  port: 6784
+                initialDelaySeconds: 30
+              resources:
+                requests:
+                  cpu: 10m
+              securityContext:
+                privileged: true
+              volumeMounts:
+                - name: weavedb
+                  mountPath: /weavedb
+                - name: cni-bin
+                  mountPath: /host/opt
+                - name: cni-bin2
+                  mountPath: /host/home
+                - name: cni-conf
+                  mountPath: /host/etc
+                - name: dbus
+                  mountPath: /host/var/lib/dbus
+                - name: lib-modules
+                  mountPath: /lib/modules
+            - name: weave-npc
+              image: 'weaveworks/weave-npc:latest'
+              imagePullPolicy: Always
+              resources:
+                requests:
+                  cpu: 10m
+              securityContext:
+                privileged: true
+          hostNetwork: true
+          hostPID: true
+          restartPolicy: Always
           securityContext:
-            privileged: true
-          volumeMounts:
+            seLinuxOptions:
+              type: spc_t
+          serviceAccountName: weave-net
+          tolerations:
+            - effect: NoSchedule
+              key: node-role.kubernetes.io/master
+          volumes:
             - name: weavedb
-              mountPath: /weavedb
+              hostPath:
+                path: /var/lib/weave
             - name: cni-bin
-              mountPath: /host/opt
+              hostPath:
+                path: /opt
             - name: cni-bin2
-              mountPath: /host/home
+              hostPath:
+                path: /home
             - name: cni-conf
-              mountPath: /host/etc
+              hostPath:
+                path: /etc
             - name: dbus
-              mountPath: /host/var/lib/dbus
+              hostPath:
+                path: /var/lib/dbus
             - name: lib-modules
-              mountPath: /lib/modules
-          resources:
-            requests:
-              cpu: 10m
-        - name: weave-npc
-          image: weaveworks/weave-npc:latest
-          imagePullPolicy: Always
-          resources:
-            requests:
-              cpu: 10m
-          securityContext:
-            privileged: true
-      restartPolicy: Always
-      tolerations:
-      - key: node-role.kubernetes.io/master
-        effect: NoSchedule
-      serviceAccountName: weave-net
-      securityContext:
-        seLinuxOptions:
-          type: spc_t
-      volumes:
-        - name: weavedb
-          hostPath:
-            path: /var/lib/weave
-        - name: cni-bin
-          hostPath:
-            path: /opt
-        - name: cni-bin2
-          hostPath:
-            path: /home
-        - name: cni-conf
-          hostPath:
-            path: /etc
-        - name: dbus
-          hostPath:
-            path: /var/lib/dbus
-        - name: lib-modules
-          hostPath:
-            path: /lib/modules
+              hostPath:
+                path: /lib/modules
--- a/prog/weave-kube/weave-daemonset.yaml
+++ b/prog/weave-kube/weave-daemonset.yaml
@@ -1,81 +1,89 @@
-apiVersion: extensions/v1beta1
-kind: DaemonSet
-metadata:
-  name: weave-net
-  namespace: kube-system
-spec:
-  template:
+apiVersion: v1
+kind: List
+items:
+  - apiVersion: v1
+    kind: ServiceAccount
     metadata:
+      name: weave-net
       labels:
         name: weave-net
-      annotations:
-        scheduler.alpha.kubernetes.io/tolerations: |
-          [
-            {
-              "key": "dedicated",
-              "operator": "Equal",
-              "value": "master",
-              "effect": "NoSchedule"
-            }
-          ]
+  - apiVersion: extensions/v1beta1
+    kind: DaemonSet
+    metadata:
+      name: weave-net
+      labels:
+        name: weave-net
     spec:
-      hostNetwork: true
-      hostPID: true
-      containers:
-        - name: weave
-          image: weaveworks/weave-kube:latest
-          imagePullPolicy: Always
-          command:
-            - /home/weave/launch.sh
-          livenessProbe:
-            initialDelaySeconds: 30
-            httpGet:
-              host: 127.0.0.1
-              path: /status
-              port: 6784
+      template:
+        metadata:
+          annotations:
+            scheduler.alpha.kubernetes.io/tolerations: >-
+              [{"key":"dedicated","operator":"Equal","value":"master","effect":"NoSchedule"}]
+          labels:
+            name: weave-net
+        spec:
+          containers:
+            - name: weave
+              command:
+                - /home/weave/launch.sh
+              env: []
+              image: 'weaveworks/weave-kube:latest'
+              imagePullPolicy: Always
+              livenessProbe:
+                httpGet:
+                  host: 127.0.0.1
+                  path: /status
+                  port: 6784
+                initialDelaySeconds: 30
+              resources:
+                requests:
+                  cpu: 10m
+              securityContext:
+                privileged: true
+              volumeMounts:
+                - name: weavedb
+                  mountPath: /weavedb
+                - name: cni-bin
+                  mountPath: /host/opt
+                - name: cni-bin2
+                  mountPath: /host/home
+                - name: cni-conf
+                  mountPath: /host/etc
+                - name: dbus
+                  mountPath: /host/var/lib/dbus
+                - name: lib-modules
+                  mountPath: /lib/modules
+            - name: weave-npc
+              image: 'weaveworks/weave-npc:latest'
+              imagePullPolicy: Always
+              resources:
+                requests:
+                  cpu: 10m
+              securityContext:
+                privileged: true
+          hostNetwork: true
+          hostPID: true
+          restartPolicy: Always
           securityContext:
-            privileged: true
-          volumeMounts:
+            seLinuxOptions:
+              type: spc_t
+          serviceAccountName: weave-net
+          volumes:
             - name: weavedb
-              mountPath: /weavedb
+              hostPath:
+                path: /var/lib/weave
             - name: cni-bin
-              mountPath: /host/opt
+              hostPath:
+                path: /opt
             - name: cni-bin2
-              mountPath: /host/home
+              hostPath:
+                path: /home
             - name: cni-conf
-              mountPath: /host/etc
+              hostPath:
+                path: /etc
             - name: dbus
-              mountPath: /host/var/lib/dbus
+              hostPath:
+                path: /var/lib/dbus
             - name: lib-modules
-              mountPath: /lib/modules
-          resources:
-            requests:
-              cpu: 10m
-        - name: weave-npc
-          image: weaveworks/weave-npc:latest
-          imagePullPolicy: Always
-          resources:
-            requests:
-              cpu: 10m
-          securityContext:
-            privileged: true
-      restartPolicy: Always
-      volumes:
-        - name: weavedb
-          hostPath:
-            path: /var/lib/weave
-        - name: cni-bin
-          hostPath:
-            path: /opt
-        - name: cni-bin2
-          hostPath:
-            path: /home
-        - name: cni-conf
-          hostPath:
-            path: /etc
-        - name: dbus
-          hostPath:
-            path: /var/lib/dbus
-        - name: lib-modules
-          hostPath:
-            path: /lib/modules
+              hostPath:
+                path: /lib/modules
--- a/site/kube-addon.md
+++ b/site/kube-addon.md
@@ -21,16 +21,8 @@ The following topics are discussed:
 Weave Net can be installed onto your CNI-enabled Kubernetes cluster
 with a single command:
 
-* Kubernetes versions `1.6` and above:
-
 ```
-$ kubectl apply -f https://git.io/weave-kube-1.6
-```
-
-* Kubernetes versions up to `1.5`:
-
-```
-$ kubectl apply -f https://git.io/weave-kube
+$ kubectl apply -n kube-system -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"
 ```
 
 After a few seconds, a Weave Net pod should be running on each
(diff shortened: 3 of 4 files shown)


prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -1,120 +1,167 @@
...
-  template:
-    metadata:
+kind: List
+items:
+  - metadata:
       labels:
         name: weave-net
+      name: weave-net
+      annotations:
+        cloud.weave.works/launcher-info: |-
```
errordeveloper: Should we strip support info annotation out or replace the content with something that indicates the origin is the repo and not the launch generator itself?

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -1,120 +1,167 @@
...
+              imagePullPolicy: Always
+              resources:
+                requests:
+                  cpu: 10m
+              securityContext:
+                privileged: true
+          volumes:
             - name: weavedb
-              mountPath: /weavedb
+              emptyDir: {}
```
errordeveloper: The diff is quite confusing in this section :(
Do we wanna manually sort the keys for the purpose of the diff?

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
marccarre: The idea with these files was to keep them as close as what we'd get using the Launch Generator in order to spot issues. These annotations shouldn't change anything in terms of behaviour though, so we could indeed remove/change them. 
What is your concern? Do you think these could be misleading and/or create confusion?

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
marccarre: My opinion is that the diff doesn't matter too much for this change.
I have just copied what I got out of the Launch Generator, so we could also sort the YAML on the Launch Generator.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
bboreham: I would remove mention of cloud.weave.works from this file.  Looks like `server-version`, `url` and `date` relate to where and when you fetched it; maybe put that in the commit message?

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -1,120 +1,167 @@
...
----
-apiVersion: extensions/v1beta1
-kind: DaemonSet
-metadata:
-  name: weave-net
-  namespace: kube-system
-spec:
-  template:
-    metadata:
+kind: List
```
bboreham: I didn't even know there was a `List` kind!

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -1,120 +1,167 @@
...
+          {
+            "server-version": "master-c1f3803",
+            "original-request": {
+              "url": "/k8s/v1.6/net?v=latest",
+              "date": "Thu Apr 20 2017 09:43:20 GMT+0000 (UTC)"
+            },
+            "email-address": "support@weave.works"
+          }
+    apiVersion: v1
+    kind: ServiceAccount
```
bboreham: This is the `kind` of the first item in the list?  Seems weird it comes at the end.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -1,120 +1,167 @@
...
+    roleRef:
+      apiGroup: rbac.authorization.k8s.io
+      kind: ClusterRole
+      name: weave-net
+    subjects:
+      - kind: ServiceAccount
+        name: weave-net
+        namespace: kube-system
+    apiVersion: rbac.authorization.k8s.io/v1beta1
+    kind: ClusterRoleBinding
```
bboreham: Slightly less racy to put the ClusterRoleBinding before the DaemonSet.

site/kube-addon.md
```diff
@@ -18,22 +18,21 @@ The following topics are discussed:
...
 ```
-$ kubectl apply -f https://git.io/weave-kube
+$ kubectl apply -n kube-system -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"
 ```
 
 After a few seconds, a Weave Net pod should be running on each
 Node and any further pods you create will be automatically attached to the Weave
 network.
 
+**Note:** In Kubernetes versions `1.6` and above, depending on how your cluster is configured, you may need to:
```
bboreham: Kubeadm instructions have to leave you with a workable system.  I don't want this cluttering up our instructions.

site/kube-addon.md
```diff
@@ -18,22 +18,21 @@ The following topics are discussed:
...
+
+- add `sudo`,
+- add `--kubeconfig /etc/kubernetes/admin.conf` (or equivalent),
+
+to the above `kubectl version` command.
```
bboreham: Presumably to both the `kubectl version` and the `kubectl apply`, were it necessary.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
marccarre: Removed from file and added to commit message, as suggested.

site/kube-addon.md
marccarre: Indeed. Updated now.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
marccarre: We should change this in the Launch Generator directly then. Cc: @errordeveloper.
Happy to do it, but after I'm done with the proxy merge.

site/kube-addon.md
marccarre: I don't think they do at the moment, so maybe we should submit a PR to improve `kubeadm`'s docs. Cc: @errordeveloper 
Happy to do it, but after I'm done with the proxy merge.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
marccarre: Now fixed by re-generating YAML files using weaveworks/launch-generator/pull/106's Launch Generator.

site/kube-addon.md
marccarre: As mentioned [here](https://weaveworks.slack.com/archives/C0DNGPKLY/p1494948825467088), I'm tempted to drop this "Note" altogether. @bboreham, WDYT?

bboreham: `kubeadm` itself has not been updated to print out:

```
To start using your cluster, you need to run (as a regular user):

  sudo cp /etc/kubernetes/admin.conf $HOME/
  sudo chown $(id -u):$(id -g) $HOME/admin.conf
  export KUBECONFIG=$HOME/admin.conf
```

so we can remove the detail from our instructions.

bboreham:
One thought: on release, we need to update the git.io short URLs, but there is no way to synchronise that with the release.

I guess it's ok if we update the static files first, then update the short URLs, and the docs are wrong about the mapping in the meantime.

site/kube-addon.md
```diff
@@ -62,6 +62,32 @@ Shut down Kubernetes, and _on all nodes_ perform the following:
 Then relaunch Kubernetes and install the addon as described
 above.
 
+**Note:** You can customise the generated YAML file by passing Weave Net's options, arguments and environment variables as query parameters:
```
bboreham: I think this should go in the "changing configuration options" section down on line 241

site/kube-addon.md
marccarre: Fine by me :+1: -- I've now removed the below:
```
**Note:** In Kubernetes versions `1.6` and above, depending on how your cluster is configured, you may need to:

- add `sudo`,
- add `--kubeconfig /etc/kubernetes/admin.conf` (or equivalent),

to the above `kubectl version` and `kubectl apply` commands.
```

marccarre:
If we proceed in the order described below, though, it should be OK?
There will indeed be a window of time during which things will be inconsistent with "_**Note:** URLs_", but I suspect very few people will read this and most will stop at the main instructions?

1. Release LG
  No one uses `/k8s/net` and `/k8s.yaml` (cloud) doesn't include Net, so serving 2.0's YAML before 2.0 is actually released should be fine.
2. Release Net, static files and docs (normal Net release process)
  a. people following the new docs will then start to use the LG, which serves up-to-date YAML. Again, I suspect most people will ignore "_**Note:** URLs_"
  b. people having scripted something based on the git.io short URLs will point to the static files, themselves just a copy of what the LG generates, i.e. 2.0's YAML
3. Update git.io short URLs to point to the LG -- when this is _eventually_ done by GitHub's support, then "_**Note:** URLs_" will finally be consistent.

Alternatively we could just keep the git.io short URLs pointing to the static files and remove "_**Note:** URLs_", but I did it this way because we apparently wanted to have all users leverage the LG.

site/kube-addon.md
marccarre: Had missed this after my previous change, as GitHub had hidden it. Will change then :+1:

site/kube-addon.md
```diff
@@ -55,6 +55,32 @@ Shut down Kubernetes, and _on all nodes_ perform the following:
...
+
+  - `version`: Weave Net's version. Default: latest release.
```
marccarre: I've just noticed this could break for previous versions of Net, given the YAML has evolved a bit.
1. Do we want to maintain backward compatible logic for versions of Net too?
2. I doubt we do so maybe we should just not document this query parameter?

site/kube-addon.md
```diff
@@ -55,6 +55,32 @@ Shut down Kubernetes, and _on all nodes_ perform the following:
...
+  - `trusted-subnets`: comma-separated list of CIDRs. Default: empty.
+  - `disable-npc`: boolean (`true|false`). Default: `false`.
+  - `enable-encryption`: boolean (`true|false`). Default: `false`.
+  - `env.NAME=VALUE`: add environment variable `NAME` and set it to `VALUE`.
+
+Example:
+```
+$ kubectl apply -n kube-system -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')&env.MTU=1337"
+```
+This command -- notice `&env.MTU=1337` at the end of the URL -- generates a YAML file containing, among others:
```
marccarre: This ought to be `WEAVE_MTU` instead of `MTU`, since `MTU` is our "internal" environment variable.

site/kube-addon.md
marccarre: Re-written this section & pushed the result now.

site/kube-addon.md
marccarre: Fixed now.

site/kube-addon.md
```diff
@@ -205,21 +207,16 @@ UDP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
...
+If you are getting your YAML file from `cloud.weave.works` as described above, then you can customise it by passing some of Weave Net's options, arguments and environment variables as query parameters:
 
-```
-      containers:
-        - name: weave
-          env:
-            - name: IPALLOC_RANGE
-              value: 10.0.0.0/16
-```
+  - `version`: Weave Net's version. Default: latest release.
```
marccarre: I've just noticed this could break for previous versions of Net, given the YAML has evolved a bit.
1. Do we want to maintain backward compatible logic for versions of Net too?
2. I doubt we do so maybe we should just not document this query parameter?

What do you think @bboreham?

bboreham: Maybe add a note that if they use this feature they should run from a saved copy of the yaml and not download it fresh every time?

bboreham:
> If we proceed in the order described below, though, it should be OK?

Yes; I was just musing about the lack of synchronisation and the docs being wrong in the interim.

site/kube-addon.md
marccarre: SGTM, attempted it in 889dfb5, feedback welcome!

bboreham:
cc @abuehrle if you want to review the docs changes

marccarre:
N.B.: I'll fix the conflicts once weaveworks/launch-generator/issues/106 and weaveworks/launch-generator/issues/107 have been approved, and will simply re-generate the YAML files using the Launch Generator.

site/kube-addon.md
```diff
@@ -63,10 +55,20 @@ Shut down Kubernetes, and _on all nodes_ perform the following:
 Then relaunch Kubernetes and install the addon as described
 above.
 
-The URLs [https://git.io/weave-kube](https://git.io/weave-kube) and [https://git.io/weave-kube-1.6](https://git.io/weave-kube-1.6) point
-to the YAML file for the [latest release](https://github.com/weaveworks/weave/releases/tag/latest_release) of the Weave Net addon.
-Historic versions are archived on our [GitHub release
-page](https://github.com/weaveworks/weave/releases).
+**Note:** URLs:
+
```
abuehrle: The URLs:

site/kube-addon.md
```diff
@@ -63,10 +55,20 @@ Shut down Kubernetes, and _on all nodes_ perform the following:
...
+- [https://git.io/weave-kube](https://git.io/weave-kube), and
+- [https://git.io/weave-kube-1.6](https://git.io/weave-kube-1.6)
+
+point respectively to:
+
```
abuehrle: point to:

site/kube-addon.md
```diff
@@ -63,10 +55,20 @@ Shut down Kubernetes, and _on all nodes_ perform the following:
...
+- [https://cloud.weave.works/k8s/v1.5/net](https://cloud.weave.works/k8s/v1.5/net), and
+- [https://cloud.weave.works/k8s/v1.6/net](https://cloud.weave.works/k8s/v1.6/net).
+
+In the past, these URLs were pointing to static YAML files for the [latest release](https://github.com/weaveworks/weave/releases/tag/latest_release) of the Weave Net addon, respectively:
+
```
abuehrle: In the past, these URLs pointed to the static YAML files for ...

site/kube-addon.md
```diff
@@ -205,21 +207,16 @@ UDP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
...
 
-The default configuration settings can be changed by saving and editing the
-addon YAML before running `kubectl apply`. Additional arguments may be
-supplied to the Weave router process by adding them to the `command:`
-array in the YAML file.
+#### Using `cloud.weave.works`
 
-Some parameters are changed by environment variables; these can be
-inserted into the YAML file like this:
+If you are getting your YAML file from `cloud.weave.works` as described above, then you can customise it by passing some of Weave Net's options, arguments and environment variables as query parameters:
```
abuehrle: If the YAML file is from '`cloud.weave.works` as described above, then you can....

site/kube-addon.md
```diff
@@ -205,21 +207,16 @@ UDP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
...
 
-```
-      containers:
-        - name: weave
-          env:
-            - name: IPALLOC_RANGE
-              value: 10.0.0.0/16
-```
+  - `version`: Weave Net's version. Default: `latest`, i.e. latest release. *WARNING*: Backward compatibility of the YAML is not guaranteed. If you lock Weave Net's version using something else than `latest` for this query parameter, please make sure you do so with the latest release, save a copy of the YAML and use this copy instead of downloading it every time from `cloud.weave.works`.
+  - `known-peers`: comma-separated list of hosts. Default: empty.
```
abuehrle: Not entirely clear what you mean here: 

If you lock Weave Net's version using something else than `latest` for this query parameter, please make sure you do so with the latest release, save a copy of the YAML and use this copy instead of downloading it every time from `cloud.weave.works`.

Do you mean: 

If you set the version of Weave Net to something other than 'latest', ensure that the YAML file matches the release that you have specified.   Save a copy of the 'YAML' file from the release repository and use that copy instead of downloading it each time from 'cloud.weave.works'.

site/kube-addon.md
```diff
@@ -241,3 +238,41 @@ The list of variables you can set is:
...
+
+**Note**: The YAML file can also be saved for later use or manual editing by using, for example:
+```
+$ curl -fsSLo weave-daemonset.yaml "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"
+```
+
+#### Manually editing the YAML file
+
+Whether you saved the YAML file served at `cloud.weave.works` or downloaded a static YAML file from our [releases page](https://github.com/weaveworks/weave/releases), you can manually edit it to suit your needs.
+
```
abuehrle: Whether you saved the YAML file served from `cloud.weave.works`

site/kube-addon.md
marccarre: What I meant is:
1. if, for example, the `latest` version is `2.0.0`, but one sets `version=1.9.5`, one will end up with a daemonset configured for `2.0.0` but using Weave Net `1.9.5`'s containers, which is likely to break;
2. therefore one should make sure:
  a. that if `version` is set to something else than `latest` (the default), it is set to the latest version at the time, e.g. in this example, `2.0.0`,
  b. the produced YAML is saved to a local file (e.g. `my.yaml`) and kept (at least until the next upgrade from that user, as for example, tomorrow we might release a change in the LG so that it supports `3.0.0` which in turn, will break `2.0.0`)
  c. install using the file previously saved, e.g. `kubectl -f apply my.yaml`
  d. that way one would have `2.0.0` containers with `2.0.0` YAML / daemonset configuration, and things should be consistent.

This was my attempt to address [this comment](https://github.com/weaveworks/weave/pull/2903#discussion_r117765377).

marccarre: My first attempt was aiming for conciseness, but probably failed to be clear.
Also, if we left it undocumented, and users store a YAML with `latest` in it, this could also cause troubles for them over time.
Would you rather have me detail it more as per my previous comment?
WDYT @abuehrle, @bboreham?

abuehrle: More detail is better, IMHO

bboreham: How about "This only changes the specified version inside the generated 'YAML' file; it does not ensure that the rest of the 'YAML' is compatible with that version.  To freeze the YAML version save a copy of the 'YAML' file from the release repository and use that copy instead of downloading it each time from 'cloud.weave.works'

prog/weave-kube/weave-daemonset.yaml
```diff
@@ -1,81 +1,89 @@
...
+                privileged: true
+          hostNetwork: true
+          hostPID: true
+          restartPolicy: Always
           securityContext:
-            privileged: true
-          volumeMounts:
+            seLinuxOptions:
+              type: spc_t
+          serviceAccountName: weave-net
```
bboreham: I don't think we want `serviceAccountName` in the 1.5 version

prog/weave-kube/weave-daemonset.yaml
```diff
@@ -1,81 +1,89 @@
...
+                  cpu: 10m
+              securityContext:
+                privileged: true
+          hostNetwork: true
+          hostPID: true
+          restartPolicy: Always
           securityContext:
-            privileged: true
-          volumeMounts:
+            seLinuxOptions:
```
bboreham: Do we want the SELinux options in the 1.5 version?

prog/weave-kube/weave-daemonset.yaml
bboreham: Ilya says we do; withdrawn

prog/weave-kube/weave-daemonset.yaml
marccarre: I do not know at the moment @bboreham, but I can add a query parameter to the LG to enable/disable this if it can help

bboreham: I commented initially because it was a change from the previous version; seeing #2990 lends weight to the idea of an option.  But that can go in a separate change.

marccarre: Raised an [issue for it](weaveworks/launch-generator/issues/114). Will do later :+1:

site/kube-addon.md
marccarre: I've changed it to your version @bboreham. I much prefer it as it doesn't phrase it as a problem like it was in my case, but more as a recommendation

marccarre:
YAML for Kubernetes 1.5 seems to work fine:

```
master$ sudo kubeadm reset && sudo rm -r -f /opt/cni/bin/*weave*
master$ k8s_version="$(kubelet --version | grep -oP "(?<=Kubernetes )v[\d\.\-beta]+")"
master$ k8s_version_option="$([[ "$k8s_version" > "v1.6" ]] && echo "kubernetes-version" || echo "use-kubernetes-version")"
master$ sudo systemctl start kubelet && sudo kubeadm init --$k8s_version_option=$k8s_version
master$ git clone https://github.com/weaveworks/weave.git && cd weave && git checkout issues/2754-launch-generator-for-net

worker$ sudo kubeadm reset && sudo rm -r -f /opt/cni/bin/*weave*
worker$ sudo systemctl start kubelet && sudo kubeadm join --token=260b6a.c036d577c3919d8f 10.128.0.5

master$ kubectl apply -f prog/weave-kube/weave-daemonset.yaml
serviceaccount "weave-net" created
daemonset "weave-net" created

master$ kubectl get po --all-namespaces
NAMESPACE     NAME                              READY     STATUS    RESTARTS   AGE
default       weave-net-9jsr1                   2/2       Running   0          1m
default       weave-net-bh8fk                   2/2       Running   0          1m
default       weave-net-ch6hr                   2/2       Running   0          1m
kube-system   dummy-2088944543-m3rnp            1/1       Running   0          3m
kube-system   etcd-k15-0                        1/1       Running   0          2m
kube-system   kube-apiserver-k15-0              1/1       Running   2          2m
kube-system   kube-controller-manager-k15-0     1/1       Running   0          2m
kube-system   kube-discovery-1769846148-l8918   1/1       Running   0          3m
kube-system   kube-dns-2924299975-xwbzc         4/4       Running   0          3m
kube-system   kube-proxy-flgk8                  1/1       Running   0          2m
kube-system   kube-proxy-l6v5n                  1/1       Running   0          2m
kube-system   kube-proxy-n32vx                  1/1       Running   0          3m
kube-system   kube-scheduler-k15-0              1/1       Running   0          2m

hosts$ curl -sS http://127.0.0.1:6784/status
        Version: git-4f20f545dc44 (up to date; next check at 2017/05/26 18:49:28)

        Service: router
       Protocol: weave 1..2
           Name: XX:XX:XX:XX:XX:XX(k15-0|1|2)
     Encryption: disabled
  PeerDiscovery: enabled
        Targets: 3
    Connections: 3 (2 established, 1 failed)
          Peers: 3 (with 6 established connections)
 TrustedSubnets: none

        Service: ipam
         Status: ready
          Range: 10.32.0.0/12
  DefaultSubnet: 10.32.0.0/12

master$ kubectl run nettest --image=weaveworks/network-tester:latest --replicas=3 -- -peers=3 -dns-name=nettest.default.svc.cluster.local.
deployment "nettest" created

master$ kubectl get po
NAME                       READY     STATUS    RESTARTS   AGE
nettest-4020221811-rr8nm   1/1       Running   0          33s
nettest-4020221811-tj4k3   1/1       Running   0          33s
nettest-4020221811-vq7r0   1/1       Running   0          33s
weave-net-9jsr1            2/2       Running   0          6m
weave-net-bh8fk            2/2       Running   0          6m
weave-net-ch6hr            2/2       Running   0          6m

master$ kubectl create -f - <<EOF
> apiVersion: v1
> kind: Service
> metadata:
>   name: nettest
> spec:
>   clusterIP: None
>   ports:
>   - port: 80
>     targetPort: 8080
>     protocol: TCP
>   selector:
>     run: nettest
> EOF
service "nettest" created

master$ podIP=$(kubectl get pods -l run=nettest -o go-template='{{(index .items 0).status.podIP}}' 2>/dev/null)
master$ echo $podIP
10.32.0.3
master$ curl -s -S http://$podIP:8080/status
pass
```

PR #2905, opened 2017-04-20
Remove some things we don't want in the v2 plugin image

bboreham:
The presence of the (old, v1.10) `docker` binary triggered a vulnerability check, because of something it embeds.  And these things are unnecessary bulk anyway, so get rid of them.


marccarre:
Makes a lot of sense to remove these.
LGTM.

PR #2907, opened 2017-04-21
Add checkpoint flag per connection type

### LINKED ISSUE #2771: Include overlay and encryption modes in checkpoint reports
Currently, we do not have any information on what percentage of users are using fastdp and whether they use it in encrypted mode.

Such information could be useful to prioritize improvements and features.

bboreham:
Fixes #2771 
Requires https://github.com/weaveworks/go-checkpoint/pull/16 and https://github.com/weaveworks/mesh/pull/75

I took out the previous "network" flag which was awsvpc-specific.

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -275,11 +275,7 @@ func main() {
 		dockerVersion = dockerCli.DockerVersion()
 	}
 
-	network := ""
-	if isAWSVPC {
-		network = "awsvpc"
-	}
-	checkForUpdates(dockerVersion, network)
+	checkForUpdates(dockerVersion, router)
 
 	observeContainers := func(o docker.ContainerObserver) {
 		if dockerCli != nil {
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -254,7 +254,8 @@ func main() {
 	checkFatal(err)
 	defer db.Close()
 
-	router := weave.NewNetworkRouter(config, networkConfig, name, nickName, overlay, db)
+	router, err := weave.NewNetworkRouter(config, networkConfig, name, nickName, overlay, db)
+	checkFatal(err)
 	Log.Println("Our name is", router.Ourself)
 
 	if peers, err = router.InitialPeers(resume, peers); err != nil {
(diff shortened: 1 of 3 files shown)


prog/weaver/main.go
```diff
@@ -330,11 +331,7 @@ func main() {
 		dockerVersion = dockerCli.DockerVersion()
 	}
 
-	network := ""
-	if bridgeConfig.AWSVPC {
-		network = "awsvpc"
```
marccarre: Do we still track this information (AWSVPC) in another way?

prog/weaver/main.go
```diff
@@ -473,6 +470,20 @@ func canonicalName(f *mflag.Flag) string {
...
 
+func checkpointFlags(router *weave.NetworkRouter) map[string]string {
+	flags := map[string]string{}
+	status := weave.NewNetworkRouterStatus(router)
+	for _, conn := range status.Connections {
+		if connectionName, ok := conn.Attrs["name"].(string); ok {
+			if _, encrypted := conn.Attrs["encrypted"]; encrypted {
+				connectionName = connectionName + " encrypted"
+			}
+			flags[connectionName] = "network"
```
marccarre: Is this expected to have `connectionName` as the key instead of as the value?
I may be missing something, but it looks like key and value have been swapped compared to: 
- `flags["network"] = network` which we had before, and 
- related logic [here](https://github.com/weaveworks/go-checkpoint/blob/a2ed17a2dd36296191268351ec9e16acbf2229f9/checkpoint.go#L148-L158).

prog/weaver/main.go
bboreham: It will come back as the name of a connection

prog/weaver/main.go
bboreham: That is correct.  It seemed to me to be more common to have multiple distinct keys than multiple distinct values for a single key.

marccarre: Alright, thanks for clarifying it 👍

prog/weaver/main.go
marccarre: Thanks for the confirmation 👍

bboreham:
I think, on reflection, I might re-do `ExtraFlags` so we can have multiple instances of the same key.

bboreham:
Updated and rebased to 1.9 branch.
weaveworks/go-checkpoint#16 has been changed too.

PR #2909, opened 2017-04-24
Remove deprecation warnings for `launch-dns` and `stop-dns` commands

marccarre:
## Context

The `launch-dns` and `stop-dns` commands have been deprecated ~2 years ago.
Given we are planning a 2.0 release for Weave Net, now is a good time to fully remove these.

## Changelog

1. Removed `launch-dns` command.
2. Removed `stop-dns` command.

N.B.: I couldn't find any mention of these commands in `docs` and `site` by `grep -R`-ing these, so that should be it.

### DIFF
--- a/weave
+++ b/weave
@@ -2103,14 +2103,6 @@ EOF
         done
         [ $res -eq 0 ]
         ;;
-    launch-dns)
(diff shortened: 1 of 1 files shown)


marccarre:
Thanks @bboreham!

PR #2910, opened 2017-04-24
Rename router.Bridge to router.InjectorConsumer

bboreham:
Since what it does it injects and consumes packets, and something else entirely does bridging.

The name seems to come from https://github.com/weaveworks/weave/pull/991#issuecomment-120427085; previously it was `IntraHost`.  But that wasn't right either, for what it does.

### DIFF
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -209,25 +209,25 @@ func (lock *fastDatapathLock) relock() {
 	}
 }
 
-// Bridge bits
+// InjectorConsumer bits
 
-type fastDatapathBridge struct {
+type fastDatapathInjectorConsumer struct {
 	*FastDatapath
 }
 
-func (fastdp *FastDatapath) Bridge() Bridge {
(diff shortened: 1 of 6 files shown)


bboreham:
I can fix up either, but probably easiest to merge #1958 first.

router/fastdp.go
```diff
@@ -209,25 +209,25 @@ func (lock *fastDatapathLock) relock() {
 	}
 }
 
-// Bridge bits
+// InjectorConsumer bits
 
-type fastDatapathBridge struct {
+type fastDatapathInjectorConsumer struct {
```
brb: Not sure about this name, as besides injecting / consuming, it does bridging (see `bridge` method of the struct) as well.

bboreham: Are you talking about `func (fastdp *FastDatapath) bridge(...)` ?

That is a member of another struct, although exposed by Go's anonymous struct rules.

PR #2916, opened 2017-04-27
fix(kube-peers): Add support for ExternalIP addresses

mikebryant:
Why:
* In some environments Nodes have an ExternalIP but no InternalIP

This change addresses the need by:
* Falling back to ExternalIP addresses if there isn't an InternalIP for a node

### DIFF
--- a/prog/kube-peers/main.go
+++ b/prog/kube-peers/main.go
@@ -35,10 +35,20 @@ func getKubePeers() ([]string, error) {
 	}
 	addresses := make([]string, 0, len(nodeList.Items))
 	for _, peer := range nodeList.Items {
+		var internalIP, externalIP string
 		for _, addr := range peer.Status.Addresses {
 			if addr.Type == "InternalIP" {
-				addresses = append(addresses, addr.Address)
+				internalIP = addr.Address
+			} else if addr.Type == "ExternalIP" {
+				externalIP = addr.Address
 			}
+		}
+
+		// Fallback for cases where a Node has an ExternalIP but no InternalIP
+		if internalIP != "" {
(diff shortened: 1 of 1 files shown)


bboreham:
Do we expect a node to have more than one InternalIP?  Previously we would add all of them; now we add just the last one.

Although this question raises additional complications because the initial IPAM consensus assumes the number of peers is the same as the number of addresses.

mikebryant:
Would that not already severely break things - doesn't `addresses := make([]string, 0, len(nodeList.Items))` mean that if you had more addresses than the number of nodes it would cause a runtime panic when it exceeds the slice capacity?

bboreham:
If you can, please rebase this on the 1.9 branch and change the base branch of the PR.  We may be doing a 1.9.5 release shortly, which this could go in.

mikebryant:
I would naively expect nodes to have at most one `InternalIP`

mikebryant:
Sure, will rebase now

bboreham:
The list of addresses is first created with zero size and `len(nodeList.Items)` capacity; if we `append` more than that then it will be reallocated in memory but will not panic.

The consensus assumption is a real problem, albeit one that wouldn't bite unless the number of internal addresses was quite large compared to the number of active peers.

mikebryant:
Ah, I see. I've just been reading about slices and capacity today, thanks for the info :)

bboreham:
Tests passed at https://circleci.com/gh/weaveworks/weave/8704

Thanks for the contribution, @mikebryant !

bboreham:
FYI Weave Net 1.9.5 was released today with this PR included.

PR #2918, opened 2017-04-27
Add a network test that isn't ping

bboreham:
This is intended to grow into a test we can use to get confidence the network works under various stress situations. `ping` is not good because it can be answered by a netdev we didn't expect.

The network-tester container was originally from Kubernetes, but modified a bit here.  It obtains a list of peer containers (currently via command-line args or via dns) and then attempts to contact all of them.

We use the new network-tester to upgrade the Kubernetes smoke-test; previously it only tested whether containers would run at all on the network and now it tests whether they can talk to each other.

### DIFF
--- a/test/101_cross_hosts_2_test.sh
+++ b/test/101_cross_hosts_2_test.sh
@@ -1,0 +1,28 @@
+#! /bin/bash
+
+. "$(dirname "$0")/config.sh"
+
+C1=10.32.1.4
+C2=10.32.1.7
+IMAGE=weaveworks/network-tester
+
+network_tester_status() {
+    # we only need to contact one container as it checks contact in both directions
+    status=$($SSH $HOST1 curl -sS http://127.0.0.1:8080/status)
+    [ -n "$status" -a "$status" != "running" ]
+}
+
+start_suite "Network test over cross-host weave network"
+
+weave_on $HOST1 launch
+weave_on $HOST2 launch $HOST1
+
+docker_on $HOST1 run --name=c1 -dt -p 8080:8080 $IMAGE -peers=2 --iface=ethwe $C1 $C2
+weave_on  $HOST1 attach $C1/24 c1
+docker_on $HOST2 run --name=c2 -dt -p 8080:8080 $IMAGE -peers=2 --iface=ethwe $C1 $C2
+weave_on  $HOST2 attach $C2/24 c2
+
+wait_for_x network_tester_status "network tester status"
+assert "echo $status" "pass"
+
+end_suite
--- a/test/840_weave_kube_3_test.sh
+++ b/test/840_weave_kube_3_test.sh
@@ -18,6 +18,7 @@ NUM_HOSTS=$(howmany $HOSTS)
 SUCCESS="$(( $NUM_HOSTS * ($NUM_HOSTS-1) )) established"
 KUBECTL="sudo kubectl --kubeconfig /etc/kubernetes/admin.conf"
 KUBE_PORT=6443
+IMAGE=weaveworks/network-tester:latest
 
 tear_down_kubeadm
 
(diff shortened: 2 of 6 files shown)


bboreham:
I figured out it was failing to pick the right route in some cases because it started before the weave `ethwe` interface was available.  So added the `EnsureInterface` code to wait for the i/f.

Could also have changed it to work with the proxy, but I want this tester to work in every mode.

test/101_cross_hosts_2_test.sh
```diff
@@ -0,0 +1,50 @@
+#! /bin/bash
+
+. "$(dirname "$0")/config.sh"
+
+C1=10.2.1.4
+C2=10.2.1.7
+UNIVERSE=10.2.0.0/16
+SUBNET_1=10.2.2.0/24
+SUBNET_2=10.2.3.0/24
```
marccarre: `SUBNET_2 ` is unused and could be removed.

test/101_cross_hosts_2_test.sh
```diff
@@ -0,0 +1,50 @@
+#! /bin/bash
+
+. "$(dirname "$0")/config.sh"
+
+C1=10.2.1.4
+C2=10.2.1.7
+UNIVERSE=10.2.0.0/16
+SUBNET_1=10.2.2.0/24
```
marccarre: Just for my understanding, what do `UNIVERSE` and `SUBNET_1` give us in this test compared to using the default?

test/101_cross_hosts_2_test.sh
```diff
@@ -0,0 +1,50 @@
...
+SUBNET_2=10.2.3.0/24
+IMAGE=weaveworks/network-tester
+
+wait_for_connections() {
```
marccarre: `wait_for_connections` seems to be unused, and could therefore be removed.

test/101_cross_hosts_2_test.sh
```diff
@@ -0,0 +1,50 @@
...
+        echo "Waiting for connections"
+        sleep 1
+    done
+    echo "Timed out waiting for connections to establish" >&2
+    exit 1
+}
+
+wait_for_network_tester_status() {
+    for i in $(seq 1 45); do
+        status=$($SSH $HOST1 curl -sS http://127.0.0.1:8080/status)
```
marccarre: Would it be worth adding a comment explaining that even though we start `network-tester` on two hosts, we only need to wait for one to be ready because they "sync" with other instances internally already?

test/840_weave_kube_3_test.sh
```diff
@@ -72,17 +73,38 @@ assert "run_on $HOST2 ps aux | grep -c '[d]efunct'" "0"
...
+run_on $HOST1 "$KUBECTL create -f -" <<EOF
+apiVersion: v1
+kind: Service
+metadata:
+  name: nettest
+spec:
+  clusterIP: None
+  ports:
+  - port: 80
+    targetPort: 80
```
marccarre: I am not too familiar with `Service`s, but shouldn't that be `8080` to match the port exposed by the pod and used for `curl` commands below?

bboreham: Yes; actually it doesn't matter because it's a headless service, but I may as well make them match.

It doesn't work at all if you leave the port out ☹️

test/101_cross_hosts_2_test.sh
bboreham: Nothing; it's all a hold-over from the very old days pre-ipam

marccarre: I'd be in favour of removing these to reduce the amount of "noise" then, as one may think they are required for this test case.

test/840_weave_kube_3_test.sh
marccarre: Oh, it is actually [named this way in Kubernetes](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services). TIL.

Okay, so that's just for the labels/selectors and DNS as mentioned in the comment.
I think it could be valuable to add a comment saying the ports aren't used/do not need to match, but cannot be removed because Kubernetes complains otherwise.

With such a comment, I think it may be even better to keep a different port to show it doesn't matter / it not used, than to have them to the same value, but this is less important than the comment IMO.

PR #2920, opened 2017-04-27
Move plugin network creation from shell script to Go

### LINKED ISSUE #2919: v1 plugin doesn't use --ipalloc-default-subnet
```
$ weave launch --ipalloc-default-subnet=10.32.1.0/24
vagrant@host2:~$ weave status
...
        Service: ipam
         Status: idle
          Range: 10.32.0.0/12
  DefaultSubnet: 10.32.1.0/24
...
$ docker run -ti --net=weave bash
bash-4.4# hostname -i
10.32.0.1 172.18.0.2
```

This is because the `create-plugin-network` line in the weave script just passes in the IP range, not the subnet.

bboreham:
In the process, use the default subnet instead of the whole IP allocation range.

Fixes #2919

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -431,6 +431,12 @@ func main() {
 	if enablePlugin || enablePluginV2 {
 		go plugin.Start(httpAddr, dockerCli, pluginSocket, pluginMeshSocket, !noDNS, enablePluginV2)
 	}
+	if enablePlugin {
+		Log.Println("Creating default 'weave' network")
+		options := map[string]interface{}{plugin.MulticastOption: "true"}
+		// TODO: the driver name should be extracted from pluginMeshSocket
+		dockerCli.EnsureNetwork("weave", "weavemesh", defaultSubnet.String(), options)
+	}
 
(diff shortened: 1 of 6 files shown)


prog/weaver/main.go
```diff
@@ -431,6 +431,12 @@ func main() {
 	if enablePlugin || enablePluginV2 {
 		go plugin.Start(httpAddr, dockerCli, pluginSocket, pluginMeshSocket, !noDNS, enablePluginV2)
 	}
+	if enablePlugin {
+		Log.Println("Creating default 'weave' network")
+		options := map[string]interface{}{plugin.MulticastOption: "true"}
+		// TODO: the driver name should be extracted from pluginMeshSocket
```
marccarre: Is this `TODO` an omission or just something to do in another PR?

bboreham: Either.  It was hard-coded in the `weave` script before; it's hard-coded in the `.go` file now.

marccarre:
This change makes sense, looks good overall, and we don't seem to be using `plugin-status.sock` in any other place. LGTM.

PR #2925, opened 2017-05-03
added search tag type

abuehrle:
Added `search_type: Documentation` as a first step for search in the new site.  This variable groups all documentation type content together. 

@bboreham please check that I haven't gaffe'ed anything.


bboreham:
This change has broken wordepress; even on this branch it didn't work.

PR #2928, opened 2017-05-04
Ignore errors on network interface delete during CNI Delete

### LINKED ISSUE #2921: k8s 1.6.2 delete service error
k8s:1.6.2
weave: 1.9.4

```bash
root@kb-node1:~# lsb_release -a
LSB Version:	core-9.20160110ubuntu0.2-amd64:core-9.20160110ubuntu0.2-noarch:security-9.20160110ubuntu0.2-amd64:security-9.20160110ubuntu0.2-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.2 LTS
Release:	16.04
Codename:	xenial
root@kb-node1:~# uname -a
Linux kb-node1 4.4.0-75-generic #96-Ubuntu SMP Thu Apr 20 09:56:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

Created it by `kubectl apply -f https://git.io/weave-kube-1.6`

When i delete a service with `kubectl delete -f abc.yaml`,  the pod status become `Terminating`, and the master docker container was Exited, but the minon docker container is still UP, look like it is hold by something

The weave container log on minon:
```bash
network: error removing interface "eth0": nsenter: cannot open /proc/14582/ns/net: No such file or directory
```

marccarre:
## Changelog:

- Ignore errors when deleting a container's interface. Fixes #2921.
- Comply to CNI 0.5 spec by always returning `nil` on CNI `DEL`. Improvement.

### DIFF
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -195,8 +195,8 @@ func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
 	// As of CNI 0.3 spec, runtimes can send blank if they just want the address deallocated
 	if args.Netns != "" {
 		if _, err = weavenet.WithNetNS(args.Netns, "del-iface", args.IfName); err != nil {
+			// We log the error and carry on instead of returning nil, as there may still be resources to free in IPAM.
 			logOnStderr(fmt.Errorf("error removing interface %q: %s", args.IfName, err))
-			return nil
 		}
 	}
 
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"strings"
 	"syscall"
 
 	"github.com/containernetworking/cni/pkg/ipam"
@@ -183,16 +182,21 @@ func assignBridgeIP(bridgeName string, ipnet net.IPNet) error {
 	return nil
 }
 
+// As of CNI 0.5 spec:
+//   "Plugins should generally complete a DEL action without error even if some resources are missing"
+// this method should therefore return nil in most, if not all, cases.
 func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
 	conf, err := loadNetConf(args.StdinData)
 	if err != nil {
-		return err
+		logOnStderr(err)
+		return nil
 	}
 
 	// As of CNI 0.3 spec, runtimes can send blank if they just want the address deallocated
 	if args.Netns != "" {
 		if _, err = weavenet.WithNetNS(args.Netns, "del-iface", args.IfName); err != nil {
-			return fmt.Errorf("error removing interface %q: %s", args.IfName, err)
+			logOnStderr(fmt.Errorf("error removing interface %q: %s", args.IfName, err))
+			return nil
 		}
 	}
 
@@ -202,14 +206,14 @@ func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
 	} else {
 		err = ipam.ExecDel(conf.IPAM.Type, args.StdinData)
 	}
-	// Hack - don't know how we should detect this situation properly
-	if args.Netns == "" && strings.Contains(err.Error(), "no addresses") {
-		err = nil
-	}
-	if err != nil {
-		return fmt.Errorf("unable to release IP address: %s", err)
-	}
+	logOnStderr(fmt.Errorf("unable to release IP address: %s", err))
 	return nil
+}
+
+// Standard error is inherited from the caller (e.g. Kubelet, when running Weave Net's CNI plugin
+// under Kubernetes) hence provided errors should eventually appear in the caller's log files.
+func logOnStderr(err error) {
+	fmt.Fprintln(os.Stderr, "weave-cni:", err)
 }
 
 type NetConf struct {


plugin/net/cni.go
```diff
@@ -183,16 +183,21 @@ func assignBridgeIP(bridgeName string, ipnet net.IPNet) error {
...
-		return err
+		logToKubelet(err)
+		return nil
 	}
 
 	// As of CNI 0.3 spec, runtimes can send blank if they just want the address deallocated
 	if args.Netns != "" {
-		if _, err = weavenet.WithNetNS(args.Netns, "del-iface", args.IfName); err != nil {
-			return fmt.Errorf("error removing interface %q: %s", args.IfName, err)
+		if _, err = weavenet.WithNetNS(args.Netns, "del-iface", args.IfName); err != nil && !os.IsNotExist(err) {
```
bboreham: I don't think weaveutil will return something that responds correctly to `IsNotExist`

plugin/net/cni.go
```diff
@@ -183,16 +183,21 @@ func assignBridgeIP(bridgeName string, ipnet net.IPNet) error {
...
+			logToKubelet(fmt.Errorf("error removing interface %q: %s", args.IfName, err))
+			return nil
```
bboreham: we should carry on and try to deallocate the address even if the interface removal fails.

plugin/net/cni.go
```diff
@@ -203,15 +208,18 @@ func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
 		err = ipam.ExecDel(conf.IPAM.Type, args.StdinData)
 	}
 	// Hack - don't know how we should detect this situation properly
-	if args.Netns == "" && strings.Contains(err.Error(), "no addresses") {
-		err = nil
-	}
-	if err != nil {
-		return fmt.Errorf("unable to release IP address: %s", err)
+	if args.Netns != "" || !strings.Contains(err.Error(), "no addresses") {
```
bboreham: it's ok to log "no addresses"; the point of the code before was to not return an error back to kubelet, but you're never going to do that

plugin/net/cni.go
```diff
@@ -203,15 +208,18 @@ func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
...
+		logToKubelet(fmt.Errorf("unable to release IP address: %s", err))
 	}
 	return nil
 }
 
+// Standard error is inherited from the caller (typically Kubelet) hence
+// provided errors should eventually appear in the caller's log files.
+func logToKubelet(err error) {
```
bboreham: "Kubelet" in the name is unnecessarily Kubernetes-specific; CNI is agnostic.  Suggest `logOnStderr`

plugin/net/cni.go
```diff
@@ -203,15 +208,18 @@ func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
...
+	fmt.Fprintln(os.Stderr, err)
```
bboreham: It would be good if this inserted some text that was easy to search for in the logs of the caller, mentioning "weave-cni" or something like that.

plugin/net/cni.go
marccarre: I thought it would work given:
- `WithNetNS` wraps around the error returned by `exec.Command` and 
- `IsNotExist` extracts the underlying error:
``` 
func isNotExist(err error) bool {
    err = underlyingError(err)
    return err == syscall.ENOENT || err == ErrNotExist
}
```

marccarre: Should I just check if the string contains `No such file or directory` instead?

plugin/net/cni.go
marccarre: OK, will completely remove it then. 👍

plugin/net/cni.go
marccarre: Changing now, thanks! 👍

plugin/net/cni.go
marccarre: Sure, will go for `weave-cni:` as a prefix then.
See: https://play.golang.org/p/JcRUOCC5Fq

plugin/net/cni.go
marccarre: Actually, given [this comment](https://github.com/weaveworks/weave/pull/2928/files#r114815795), we probably do not care about `ENOENT` at all then.

plugin/net/cni.go
marccarre: Updated in latest version of this PR.

plugin/net/cni.go
bboreham: `WithNetNS` returns using `fmt.Errorf()`; there's no way anyone is going to extract anything from that except by string parsing.

I did change my mind on Slack about `ENOENT`; I said "we should just allow anything and everything to go wrong."

plugin/net/cni.go
```diff
@@ -183,16 +182,21 @@ func assignBridgeIP(bridgeName string, ipnet net.IPNet) error {
...
 }
 
+// As of CNI 0.5 spec:
+//   "Plugins should generally complete a DEL action without error even if some resources are missing"
+// this method should therefore return nil in most, if not all, cases.
 func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
 	conf, err := loadNetConf(args.StdinData)
 	if err != nil {
-		return err
+		logOnStderr(err)
```
bboreham: I'm not 100% convinced about this change; if we can't even load the config that's pretty bad.  And returning immediately without trying to do anything is not really in the spirit of "complete a DEL action"

However I think we should go ahead because Kubernetes is being super-dumb about sending us rubbish and then getting stuck when we complain, so let's try never erroring and see how it goes.

marccarre: Not even being able to deserialise Kubernetes' input sounded like a good reason to exit early, but as you wish, I'll remove that `return nil` and let it go ahead.

marccarre: Also note that the behaviour isn't consistent between `CmdAdd` and `CmdDel` now. I was thinking we should harmonise it, but I'll leave this choice to you.

PR #2929, opened 2017-05-05
Work around missing docker.sock in plugin-v2

### LINKED ISSUE #2906: Docker plugin-v2 can be started before /var/run/docker.sock is ready
If `/var/run/docker.sock` is not ready, plugin-v2 will fail silently to start:

```
16:07:48+01:00" level=info msg="Starting launch.sh" plugin=d4429df88c47efe094b7dd3c471aa7f4f61612e378814ad706c7b7d0
...
16:08:26.948417772+01:00" level=info msg="API listen on /var/run/docker.sock"
```

Worse than this is Docker reporting that the plugin is enabled:

```
$ docker plugin ls
ID                  NAME                           DESCRIPTION                   ENABLED
d4429df88c47        weaveworks/net-plugin:latest   Weave Net plugin for Docker   true
```

Reproduced with Docker 17.04.0-ce after restarting `docker.service`.

brb:
Current limitations:

* No `configure-arp`.
* Multicast can be enabled only for all networks (via `docker plugin set weaveworks/net-plugin WEAVE_MULTICAST=1` on each host).

Partly fixes #2906

### DIFF
--- a/plugin/net/driver.go
+++ b/plugin/net/driver.go
@@ -29,28 +29,39 @@ type network struct {
 
 type driver struct {
 	sync.RWMutex
-	scope        string
-	docker       *docker.Client
-	dns          bool
-	isPluginV2   bool
-	networks     map[string]network
-	isNetworkOur func(driverName string) bool
+	name  string
+	scope string
+	// Docker API is not available for plugin-v2
+	docker     *docker.Client
+	dns        bool
+	isPluginV2 bool
+	// Enable multicast regardless whether multicast opt is passed to Docker;
+	// used only by plugin-v2
+	forceMulticast bool
+	networks       map[string]network
 }
 
-func New(client *docker.Client, weave *weaveapi.Client, scope string, dns, isPluginV2 bool, isNetworkOur func(string) bool) (skel.Driver, error) {
+func New(client *docker.Client, weave *weaveapi.Client, name, scope string, dns, isPluginV2, forceMulticast bool) (skel.Driver, error) {
 	driver := &driver{
-		scope:        scope,
-		docker:       client,
-		dns:          dns,
-		isPluginV2:   isPluginV2,
-		networks:     make(map[string]network),
-		isNetworkOur: isNetworkOur,
+		name:       name,
+		scope:      scope,
+		docker:     client,
+		dns:        dns,
+		isPluginV2: isPluginV2,
+		// make sure that it's used only by plugin-v2
+		forceMulticast: isPluginV2 && forceMulticast,
+		networks:       make(map[string]network),
 	}
 
-	_, err := NewWatcher(client, weave, driver)
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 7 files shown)


prog/net-plugin/launch.sh
```diff
@@ -16,42 +16,38 @@ echo "Starting launch.sh"
...
-/usr/bin/weaveutil is-swarm-manager 2>/dev/null || STATUS=$?
-if [ $STATUS -eq 0 ]; then
-    IS_SWARM_MANAGER=1
-elif [ $STATUS -eq 20 ]; then
-    echo "Host swarm is not \"active\"; exiting." >&2
-    exit 1
+# We need to get a list of Swarm nodes which might run the net-plugin:
+# - In the case of missing restart.sentinel, we assume that net-plugin has started
+#   for the first time via the docker-plugin cmd. So it's safe to request docker.sock.
+# - If restart.sentinel present, let weaver restore from it as docker.sock is not
```
bboreham: Does this mean we will hang on to a peer in the list which has actually been removed from the Docker Swarm?
(this could be filed as a separate issue; doesn't have to be fixed here.)

plugin/net/driver.go
```diff
@@ -189,11 +200,30 @@ func (driver *driver) findNetworkInfo(id string) (network, error) {
...
 	}
+
+	// plugin-v2 does not have access to docker.sock, so we cannot call Docker
+	// API for the network info.
+	if driver.isPluginV2 {
+		// safe to set, as isOurs used only by the watcher which is disabled for plugin-v2
+		network.isOurs = true
+		network.hasMulticastRoute = driver.forceMulticast
+
+		driver.Lock()
```
bboreham: I think I might have omitted adding the data to the map, since we would just end up back here next time in that case.
Locking/adding in two places makes me nervous.

prog/weaveutil/swarm.go
```diff
@@ -2,48 +2,32 @@ package main
...
 	for _, managerNode := range info.Swarm.RemoteManagers {
 		ip, err := ipFromAddr(managerNode.Addr)
 		if err != nil {
 			return errors.Wrap(err, "ipFromAddr")
 		}
-		fmt.Println(ip)
+		ips = append(ips, ip)
 	}
 
+	fmt.Println(strings.Join(ips, " "))
```
bboreham: guessing this change was for easier testing?

brb: Yep

plugin/net/driver.go
brb: Same though, I'm going to remove it.

prog/net-plugin/launch.sh
brb: Yep. I've created a separate issue for plugin-v2 improvements: https://github.com/weaveworks/weave/issues/2938

brb:
PTAL.

PR #2932, opened 2017-05-05
Fix protect_against_docker_hang

brb:
Check not only for plugin-v2 but also for plugin-v1, as both create the same plugin sockets (`weave.sock` and `weavemesh.sock`).

Should fix flakiness in 80* smoke tests.

### DIFF
--- a/weave
+++ b/weave
@@ -1442,7 +1442,7 @@ stop_proxy() {
 
 protect_against_docker_hang() {
     # If the plugin is not running, remove its socket so Docker doesn't try to talk to it
-    if ! is_plugin_v2_enabled 2>/dev/null ; then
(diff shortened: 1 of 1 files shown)


marccarre:
`[` (`test`) actually evaluates the expression as a string of non-zero length:
```
#!/bin/sh
foo() {
	false
}

if [ ! foo ] && [ ! foo ] ; then
	echo "NOT foo"
else
	echo "foo"
fi
```
outputs: "`foo`" which is unexpected as the intent really is:
```
#!/bin/sh
foo() {
	false
}

if ! foo && ! foo ; then
	echo "NOT foo"
else
	echo "foo"
fi
```
which outputs "`NOT foo`" as expected.

marccarre:
Fixed in `master` in b46ccc0

PR #2935, opened 2017-05-08
Clean up xt_set_exists components in advance

### LINKED ISSUE #2934: weave-net crashes on startup: ipset v6.29: Set cannot be created: set with the same name already exists
```
m@hotcpc17686:~/git/kubernetes-deployments/andover-cit$ k logs -f weave-net-96lwr -c weave
ipset v6.29: Set cannot be created: set with the same name already exists
```

weave-kube: 1.9.5

mikebryant:
If this function dies in the middle, the leftover weave-kube-test set
causes this function and weave-npc to break. In particular
this means we can never restart weave with the launch script.

Here we clean up in advance, in case there's leftovers from a previous run

Fixes #2934

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -8,6 +8,11 @@ modprobe_safe() {
 
 # Check whether xt_set actually exists
 xt_set_exists() {
+    # Clean everything up in advance, in case there's leftovers
+    iptables -F WEAVE-KUBE-TEST 2>/dev/null || true
+    iptables -X WEAVE-KUBE-TEST 2>/dev/null || true
+    ipset destroy weave-kube-test 2>/dev/null || true
+
     ipset create weave-kube-test hash:ip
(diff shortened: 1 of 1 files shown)


mikebryant:
@bboreham Does this seem like a reasonable fix?

mikebryant:
Hmm, and can this go into the 1.9 series?

bboreham:
Yes, it should be branched off 1.9.

I guess this is fine, though it seems slightly nutty to have a 14-line function dedicated to detecting whether one kernel feature is available.  I cannot think of a simpler/easier way however - we can't just look for the module since it may be compiled in.

mikebryant:
rebased :)

bboreham:
LGTM; I'm running the full tests at https://circleci.com/gh/weaveworks/weave/8850 so after that passes I will merge.

Thanks!

bboreham:
FYI we have a CI failure that seems to stem from an external change to openssl; the smoke-tests passed so I'm going to merge.

PR #2936, opened 2017-05-08
Merge proxy and router into one single process.

### LINKED ISSUE #1642: RFC: Package router, proxy, and plugin in same container
Original reasons for running the proxy in a separate container:
- Avoid the router having a hard dependency on Docker
- Proxy needed to run in host netns to access Docker TCP port

The situation has now changed - the router already has a dependency on Docker due to the weaveDNS merge, and it also now runs in the host netns due to fast datapath. I therefore propose that we distribute the two in the same container image (exact implementation TBD) to obtain the following benefits:
- Simplified `weave` script - in particular `weave launch` and `weave stop` won't have to deal with edge cases around making start/stop of multiple containers appear to be an atomic step
- `weave status` and `weave report` can include proxy related information (inclusion of proxy at bottom of `weave status` is currently implemented via bash script hackery)
- Elimination of races e.g. #1556 

I suspect this may a) be contentious and b) have downsides I haven't thought of - the purpose of this issue is to act as a vehicle for discovery and documentation. Comments please.

marccarre:
Fixes #1642.


prog/weaver/main.go
```diff
@@ -452,18 +452,21 @@ func main() {
...
+		proxy, err := weaveproxy.NewProxy(*proxyConfig)
+		if err != nil {
+			Log.Fatalf("Could not start proxy: %s", err)
+		}
+		defer proxy.Stop()
+		listeners := proxy.Listen()
+		proxy.AttachExistingContainers()
+		go proxy.Serve(listeners)
+		go proxy.ListenAndServeStatus("/home/weave/status.sock")
+		signals.SignalHandlerLoop(common.Log, proxy, router)
```
bboreham: why doesn't the plugin figure here?

weave
```diff
@@ -1393,51 +1425,17 @@ launch_router() {
         --http-addr $HTTP_ADDR \
         --status-addr $STATUS_ADDR \
         --resolv-conf "/var/run/weave/etc/$RESOLV_CONF_BASE" \
-        "$@")
+        $PROXY_ARGS)
     wait_for_status $CONTAINER_NAME http_call $HTTP_ADDR
+    # Wait for Weave Proxy:
+    [ -n "$NO_PROXY" ] || wait_for_status $CONTAINER_NAME http_call_unix $CONTAINER_NAME status.sock
```
bboreham: We shouldn't need this.

prog/weaver/main.go
marccarre: Purely because it isn't there in [`master`](https://github.com/weaveworks/weave/blob/master/prog/weaver/main.go#L448) at the moment and I didn't want to diverge too much from it given I'm still discovering the codebase, and I had enough troubles to get it to a stable state already.
I can add it though, now that all smoke tests are passing.

weave
```diff
@@ -1553,25 +1550,10 @@ EOF
...
-        launch_router "$@"
-        echo $ROUTER_CONTAINER
-        ;;
-    launch-proxy)
-        deprecation_warnings "$@"
-        check_not_running proxy $PROXY_CONTAINER_NAME $BASE_EXEC_IMAGE
-        check_not_running_plugin_v2
-        launch_proxy "$@"
-        echo $PROXY_CONTAINER
+        launch --plugin "$COMMON_ARGS" "$@"
```
marccarre: Sure, will change this now :+1:

weave
```diff
@@ -1553,25 +1550,10 @@ EOF
     launch)
         deprecation_warnings "$@"
         check_not_running router $CONTAINER_NAME        $BASE_IMAGE
-        check_not_running proxy  $PROXY_CONTAINER_NAME  $BASE_EXEC_IMAGE
         check_not_running_plugin_v2
         COMMON_ARGS=$(common_launch_args "$@")
```
marccarre: Sure, will simplify now :+1:

weave
marccarre: I will remove this line entirely then. :+1: 

For the record, we have such a `wait_for_status` call in [`master`](https://github.com/weaveworks/weave/blob/master/weave#L1435) at the moment, hence I kept it and moved it here as I merged the router and the proxy. I added a condition given it makes no sense to wait if the proxy is disabled.

prog/weaver/main.go
bboreham: Having thought about it a little more, you already have `defer proxy.Stop()` so it's pointless passing proxy in here.
And there is no `plugin.Stop` so that's a separate issue.

marccarre: Good point, `Stop`ping twice is unnecessary. I'll remove this line entirely, and move `SignalHandlerLoop(common.Log, router)` at the very end of the function.

weave
bboreham: Changed my mind again: users can add `--plugin=false` and Go will take the last one as the final version.

weave
bboreham: I mean "it shouldn't be necessary because `weaver` should ensure that everything is up and running before answering the main status call".  But I see that isn't actually the case - it calls `ListenAndServeHTTP` before starting the plugin and the proxy.  In the plugin case this is perhaps to make the `EnsureNetwork` call work.  

I can't see a reason to start the proxy after the main http listener.

weave
marccarre: As discussed offline, there really are two "APIs" here:
- the "API" exposed by the `weave` script;
- the "API" exposed by the `weaver` container.

It has been agreed that:
1. Go code should have plugin and proxy disabled by default. 
    - This is leaner by default. 
    - This makes calling `weaver` slightly nicer in the case of the `weave-kube`'s `launch.sh` script or for third parties like Cloud66.
2. `--proxy`/`--plugin` should be passed in by `weave launch` to continue to start all three components by default and have a consistent behaviour with what was there previously.
3. It should be possible to pass `--x=false` (`x = { proxy, plugin }`), which `mflag.BoolVar(&x, []string{"-x"}, false, "...")` would understand, and given the last value takes precedence, `--x --x=false` would lead to `x` being disabled.

prog/net-plugin/launch.sh
```diff
@@ -46,6 +46,7 @@ exec /home/weave/weaver $EXTRA_ARGS --port=6783 $(router_bridge_opts) \
     --host-root=/host \
     --proc-path=/host/proc \
     --http-addr=$HTTP_ADDR --status-addr=$STATUS_ADDR \
+    --without-proxy \
```
marccarre: We have `--without-dns`, hence I went for `--without-proxy`. 
However, as discussed offline and summarised [here](https://github.com/weaveworks/weave/pull/2936#discussion_r115501309), we should really flip this and have `--proxy`. This raises the question of `--without-dns` vs. `--dns`, but prob. for another PR.

weave
marccarre: > I can't see a reason to start the proxy after the main http listener.

OK, I can move it to the beginning of the `main` and see how that goes (which is what I initially had, but wasn't convenient for `signals.SignalHandlerLoop(common.Log, proxy, router)`).

prog/weaver/main.go
```diff
@@ -497,6 +517,31 @@ func (nopPacketLogging) LogPacket(string, weave.PacketKey) {
...
 }
 
+func configureProxy(version string, defaultDockerHost string) *weaveproxy.Config {
+	proxyConfig := weaveproxy.Config{
+		Version:      version,
+		Image:        getenvOrDefault("EXEC_IMAGE", "weaveworks/weaveexec"),
+		DockerBridge: getenvOrDefault("DOCKER_BRIDGE", "docker0"),
+		DockerHost:   defaultDockerHost,
+	}
+	mflag.BoolVar(&proxyConfig.Disabled, []string{"-no-proxy"}, false, "instruct Weave not to start the Docker proxy")
```
brb: Personal taste, but I would prefer these flags to be moved to the same place as all other flags. It would make it easier for a reader to see what flags are possible.

weave
marccarre: Done in `fa9b1a5c4b61 TO SQUASH: start proxy before router, remove SignalHandlerLoop on proxy, do not sync on proxy status.`

weave
marccarre: Done in 2d665608b950 `TO SQUASH: Removed/simplified common_launch_args. Removed local as only supported by bash.`

prog/weaver/main.go
marccarre: Done in fa9b1a5c4b61 `TO SQUASH: start proxy before router, remove SignalHandlerLoop on proxy, do not sync on proxy status.`

prog/weaver/main.go
```diff
@@ -246,6 +253,19 @@ func main() {
...
 	config.ProtocolMinVersion = byte(protocolMinVersion)
 
+	if !proxyConfig.Disabled {
+		// Start Weave Proxy:
+		proxy, err := weaveproxy.NewProxy(*proxyConfig)
+		if err != nil {
+			Log.Fatalf("Could not start proxy: %s", err)
+		}
+		defer proxy.Stop()
+		listeners := proxy.Listen()
```
brb: Nitpick, but maybe it makes sense to create `proxy.Start()` and to  move the rest of the lines to this method?

marccarre: That's what I wanted to do as a subsequent refactoring. Now may be the right time given it seems to be in a functional state.

prog/weaver/main.go
marccarre: Sure, will do :+1: 
I did it this way cause I thought grouping may make sense.
Maybe we should just have all `mflag` instructions next to each others, but use separate `struct`s for proxy, router, etc. to separate concerns?

prog/net-plugin/launch.sh
marccarre: Done in 974b867dd672e024ad9527dbcaea4c82d3608736 `Add --proxy option to enable/disable proxy.`

weave
marccarre: Done in 974b867dd672e024ad9527dbcaea4c82d3608736 `Add --proxy option to enable/disable proxy.`

prog/weaver/main.go
marccarre: Fixed in latest version.

prog/weaver/main.go
marccarre: Will cover in #2943. Let's continue the discussion there.

weave
```diff
@@ -48,7 +48,7 @@ weave launch        [--password <pass>] [--trusted-subnets <cidr>,...]
                     [--no-restart] [--resume] [--no-discovery] [--no-dns]
                     [--ipalloc-init <mode>]
                     [--ipalloc-range <cidr> [--ipalloc-default-subnet <cidr>]]
-                    [--plugin]
+                    [--plugin[=false] --proxy[=false]]
```
bboreham: I think this makes more sense as `[--plugin=false] [--proxy=false]`
Currently it says you can have both or neither, and it doesn't make sense to add without `=false`.

prog/weaver/main.go
marccarre: Will cover in #2943. Let's continue the discussion there.

bboreham:
Looking good apart from a couple more nits.

weave
marccarre: Fixed in latest version.

prog/weaver/main.go
```diff
@@ -223,19 +229,20 @@ func main() {
...
 
 	common.SetLogLevel(logLevel)
-
-	if justVersion {
-		fmt.Printf("weave router %s\n", version)
-		os.Exit(0)
-	}
-
 	Log.Println("Command line options:", options())
+	Log.Infoln("Command line arguments:", strings.Join(os.Args[1:], " "))
```
bboreham: We can't do this - it could print out the password.  I think the line before should cover the requirement.

weave
```diff
@@ -48,7 +48,7 @@ weave launch        [--password <pass>] [--trusted-subnets <cidr>,...]
                     [--no-restart] [--resume] [--no-discovery] [--no-dns]
                     [--ipalloc-init <mode>]
                     [--ipalloc-range <cidr> [--ipalloc-default-subnet <cidr>]]
-                    [--plugin]
+                    [--plugin=false --proxy=false]
```
bboreham: This is again saying we can have both or neither, which is wrong.  They are independent.
`[--plugin=false] [--proxy=false]`

marccarre: OIC, I previously thought you were referring to the values and overlooked the brackets grouping both options. Fixed in latest version.

prog/weaver/main.go
marccarre: Removed now.

bboreham:
Docs still pending

PR #2937, opened 2017-05-08
WIP: In the build container, request install/upgrade of openssl

bboreham:
It seems that the install of libssl causes builds to break, because Python can no longer load the SSL code.
Example: https://circleci.com/gh/weaveworks/weave/8855


bboreham:
Replaced by #2940

PR #2940, opened 2017-05-09
Fix build libssl issue: Use cross-compilers from debian rather than ubuntu

leth:
Golang containers seem to built on debian jessie, not ubuntu.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -20,13 +20,12 @@ ENV GCC_CROSSCOMPILERS \
 
 # Temporarily add the Ubuntu repositories, because we're going to install gcc cross-compilers from there
 # Install the build-essential and crossbuild-essential-ARCH packages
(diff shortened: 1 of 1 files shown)


leth:
build seems to be broken because the branch name contains underscores

bboreham:
I'm going to check the cross-compiled builds still run

leth:
Yeah good call!

bboreham:
Cross-platform builds completed without error.
cc @luxas for info - you originally did this part.

luxas:
FYI emdebian doesn't provide s390x packages if that's of interest in the future.
I chose ubuntu because that's what we're using in Kubernetes crossbuilds.

bboreham:
Thanks for info.  Likewise you might want to watch out for mismatched openssl libs in the Kubernetes build container.

luxas:
@bboreham What were the versions respectively?

bboreham:
I never figured that out; you can see the full logs in that CircleCI transcript, but one of them is expecting `OPENSSL_1.0.0` to be defined and the other isn't exporting it.

PR #2945, opened 2017-05-11
Update docs following merge of proxy & router

marccarre:
Related to #1642, and following #2936.

Note that I've also deleted trailing whitespaces for the files I have modified.
This is done as a separate commit for ease of review, but should be squashed before merging.


bboreham:
@abuehrle do you want to review?

PR #2946, opened 2017-05-11
Wait until ready; make test 810 more reliable 

bboreham:
Various parts of the system initialize asynchronously at startup.
Callers should not attempt operations until all parts are ready; the `weave` script already delays until `curl /status` returns 200, but this could happen too early.
So, we introduce a `ready` function to be called by each sub-part: plugin, proxy, expose; this function comes from a wrapped `WaitGroup`.
We don't actually wait on the `WaitGroup`; that could perhaps be simplified, but I thought it was a good idea to follow the pattern.

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -23,19 +23,19 @@ const (
 
 var Log = common.Log
 
-func Start(weaveAPIAddr string, dockerClient *docker.Client, address string, meshAddress string, dns bool, isPluginV2, forceMulticast bool) {
+func Start(weaveAPIAddr string, dockerClient *docker.Client, address string, meshAddress string, dns bool, isPluginV2, forceMulticast bool, defaultSubnet string) {
 	weave := weaveapi.NewClient(weaveAPIAddr, Log)
 
 	Log.Info("Waiting for Weave API Server...")
 	weave.WaitAPIServer(30)
 	Log.Info("Finished waiting for Weave API Server")
 
-	if err := run(dockerClient, weave, address, meshAddress, dns, isPluginV2, forceMulticast); err != nil {
+	if err := run(dockerClient, weave, address, meshAddress, dns, isPluginV2, forceMulticast, defaultSubnet); err != nil {
 		Log.Fatal(err)
 	}
 }
 
-func run(dockerClient *docker.Client, weave *weaveapi.Client, address, meshAddress string, dns, isPluginV2, forceMulticast bool) error {
+func run(dockerClient *docker.Client, weave *weaveapi.Client, address, meshAddress string, dns, isPluginV2, forceMulticast bool, defaultSubnet string) error {
 	endChan := make(chan error, 1)
 
 	if address != "" {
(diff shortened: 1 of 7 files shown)


proxy/proxy.go
```diff
@@ -280,13 +280,17 @@ func (proxy *Proxy) Listen() []net.Listener {
...
 			errs <- (&http.Server{Handler: proxy}).Serve(listener)
 		}(listener)
 	}
+	// It would be better if we could delay calling Done() until all
+	// the listeners are ready, but it doesn't seem to be possible to
+	// hook the right point in http.Server
+	ready()
 	for range listeners {
 		err := <-errs
 		if err != nil {
```
marccarre: Not caused by this change, but I noticed that, at least [according to the docs](https://golang.org/pkg/net/http/#Server.Serve) this condition should always be met, leading to potentially confusing log messages:
> Serve always returns a non-nil error. 
> After Shutdown or Close, the returned error is ErrServerClosed.

It should probably check if the error returned is `ErrServerClosed` instead.

proxy/proxy.go
```diff
@@ -280,13 +280,17 @@ func (proxy *Proxy) Listen() []net.Listener {
...
+func (proxy *Proxy) Serve(listeners []net.Listener, ready func()) {
 	errs := make(chan error)
 	for _, listener := range listeners {
 		go func(listener net.Listener) {
 			errs <- (&http.Server{Handler: proxy}).Serve(listener)
 		}(listener)
 	}
+	// It would be better if we could delay calling Done() until all
+	// the listeners are ready, but it doesn't seem to be possible to
+	// hook the right point in http.Server
```
marccarre: I looked at Go's sources and also couldn't find anything. This is indeed a bit annoying.
Could something like the below help to reduce the window for a race, even if still not strictly correct?

```
func (proxy *Proxy) Serve(listeners []net.Listener) {
	created := make(chan bool)
	errs := make(chan error)
	for _, listener := range listeners {
		go func(listener net.Listener) {
			server := http.Server{Handler: proxy}
			created <- true
			errs <- server.Serve(listener)
		}(listener)
	}
	for range listeners {
		<-created
	}
	ready()

	for range listeners {
		err := <-errs
		if err != nil {
			Log.Fatalf("Serve failed: %s", err)
		}
	}
}
```

bboreham: Reducing the window for a race just makes it harder to find, while leaving the race.  I don't think it is worth the extra complexity.

proxy/proxy.go
bboreham: I think Close and Shutdown were added in Go 1.8; we aren't calling them.

plugin/plugin.go
```diff
@@ -54,6 +54,13 @@ func run(dockerClient *docker.Client, weave *weaveapi.Client, address, meshAddre
 		defer os.Remove(meshAddress)
 		defer meshListener.Close()
 	}
+	if !isPluginV2 {
```
brb: Ouch, too late, but this could go to `if meshAddress != "" {` branch, as the network should be created only if the local scope driver (`weavemesh`) is started.

PR #2947, opened 2017-05-11
Attach containers from Proxy using Go code

### LINKED ISSUE #1940: Remove proxy dependency on weave script

bboreham:
Rather than calling out to the shell script.
This is the last part to fix #1940

### DIFF
--- a/proxy/common.go
+++ b/proxy/common.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/fsouza/go-dockerclient"
 	"github.com/weaveworks/weave/common"
+	weavedocker "github.com/weaveworks/weave/common/docker"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


proxy/common.go
```diff
@@ -25,42 +23,6 @@ var (
 	Log = common.Log
 )
 
-func callWeave(args ...string) ([]byte, []byte, error) {
```
marccarre: Are there parts of the `shell` script we can now get the rid of thanks to this refactoring?

bboreham: We were calling `weave attach`, and that remains a valid command that users can call separately, so I don't think so.  It's plausible that we could trim down the argument-parsing it used to do - e.g. `--add-host` was not a documented option - but that's pretty low-benefit.

Issue #2939, opened 2017-05-09
ipset: v6.29: Set cannot be destroyed: it is in use by a kernel component


david-rooke:
Environment:
* kubernetes: 1.6.0
* Alpine Linux: 3.4.6
* Weave: 1.9.5

I am seeing regular weave-npc crashes relating to ipset:

    INFO: 2017/05/09 07:58:21.961956 destroying ipset: &npc.selectorSpec{key:"", selector:labels.internalSelector{}, ipsetType:"hash:ip", ipsetName:"weave-94D4P%N1ac^8#m}i;KwFqXYQu"}
    FATA: 2017/05/09 07:58:21.977964 ipset [destroy weave-94D4P%N1ac^8#m}i;KwFqXYQu] failed: ipset v6.29: Set cannot be destroyed: it is in use by a kernel component
    : exit status 1

stuart-warren:
related to #2915 ?

marccarre:
It would appear as a duplicate, yes. 

Closing for now, feel free to comment on #2915 directly, or to re-open this issue if you think they are distinct.

@david-rooke, 

1. did you try the [workaround from #2915](https://github.com/weaveworks/weave/issues/2915#issuecomment-297608065)?
2. you may also want to look for IPTables rules (`iptables-save`) which are using such an ipset, and delete these.

stuart-warren:
I actually now think this is a separate issue, can it be reopened?

#2915 occurs on startup and is resolved by #2949

#2939 occurs during runtime and is likely due to this Destroy method: https://github.com/weaveworks/weave/blob/1.9/npc/selector.go#L100 failing.

mikebryant:
Thought - this is probably because we have an aggressive runtime end-to-end test going, which among other things, is creating/deleting namespaces a lot, so I think we're exercising a code path that's probably not that well tested, or used as frequently elsewhere

mikebryant:
We're also seeing things like:
```
FATA: 2017/05/12 15:38:51.072861 add namespace: exit status 4: iptables: Resource temporarily unavailable.
```

Both of these issues appear to be somewhat racy, the same namespace can be brought up or taken down on some nodes and not others

bboreham:
We [have code](https://github.com/weaveworks/weave/blob/master/net/veth.go#L270-L284) that retries an iptables operation until that doesn't happen; it will be easy to call that from weave-npc and/or move the code up to the coreos iptables lib.

[...]

PR #2949, opened 2017-05-12
Flush weave ipsets before destroying

### LINKED ISSUE #2915: Can't restart weave-npc because of ipset error "Set cannot be destroyed"
Hi,

We currently encountered an issue on our Kubernetes cluster v1.4 where we weren't able to restart the weave-net pod because the weave-npc container were always failing with the following error message:
```
ipset [destroy weave-ct?g{Ln1Djp[:;d;4OhuoU$cJ] failed: ipset v6.29: Set cannot be destroyed: it is in use by a kernel component
```

Complete logs are:
```
INFO: 2017/04/27 03:53:22.845217 Starting Weaveworks NPC 1.9.4
INFO: 2017/04/27 03:53:22.846795 Serving /metrics on :6781
Thu Apr 27 03:53:22 2017 <5> ulogd.c:843 building new pluginstance stack: 'log1:NFLOG,base1:BASE,pcap1:PCAP'
DEBU: 2017/04/27 03:53:22.892522 Got list of ipsets: [weave-ct?g{Ln1Djp[:;d;4OhuoU$cJ weave-viMd_3{ytI{fsqkGEq>hQZj0E weave-uryNoZWn|qTJOXF5L7Rq1:CqC weave-Lb5cDl39gCzYJ4CZ9N;UFY[/o weave-]j}tH1hV{.pj>EQSy7|]of9uH weave-E0_C&$DB[j.JV_Y[u[S2(Elug weave-k?Z;25^M}|1s7P3|H9i;*;MhG weave-k^ECI/NRo6zq5ZWEdpMk?{lDi weave-XwzO78vI{0XK#vHrdH;M*@%M_ weave-fk/~8ml5jU/KNi7yrO<ux9Jys weave-}VOW.2vMv;dOx<11<i*8Z$9yi weave-hVDldOfE<%;@Y&^7d52k1Px|K weave-ep3/r<H%.JRG<#/a)kg66<3qf weave-]2^zFy(_q?M2Q}vavlH0ij{|6 weave-~8f>?yo$hQ^}|veb/Jy;9|Lt) weave-JJai@G2zdWbZjQO(uvd]qtI7v weave-nn;DA}.Sqh*tCD6yB*P97u#TY weave-OXze2|dZ.m@4lTf8p>#&1/kc<]
DEBU: 2017/04/27 03:53:22.892550 Destroying ipsets 'weave-ct?g{Ln1Djp[:;d;4OhuoU$cJ'
ERRO: 2017/04/27 03:53:22.893761 Failed to destroy ipset 'weave-ct?g{Ln1Djp[:;d;4OhuoU$cJ'
FATA: 2017/04/27 03:53:22.893793 ipset [destroy weave-ct?g{Ln1Djp[:;d;4OhuoU$cJ] failed: ipset v6.29: Set cannot be destroyed: it is in use by a kernel component
: exit status 1
```

The listed ipset were not used at all in all iptables tables (default, filter, mangle, nat, raw, security)

After some investigation, it appears that weave was not able to destroy the ipset because the first ipset that weave tried to destroy was referenced in another weave ipset:
```
Name: weave-ct?g{Ln1Djp[:;d;4OhuoU$cJ
Type: hash:ip
Revision: 1
Header: family inet hashsize 1024 maxelem 65536
Size in memory: 17632
References: 1
Members:
172.100.96.12
[...]

Name: weave-nn;DA}.Sqh*tCD6yB*P97u#TY
Type: list:set
Revision: 1
Header: size 8
Size in memory: 128
References: 0
Members:
weave-ct?g{Ln1Djp[:;d;4OhuoU$cJ
```

Looking at that issue it seems weave resetIPSets function should be updated to destroy ipsets by taking into account references.

I can propose a patch for that but I would like to know if the situation we encountered is not caused by some other things as I am surprised no one else had the issue.

mikebryant:
Fixes #2915

### DIFF
--- a/prog/weave-npc/main.go
+++ b/prog/weave-npc/main.go
@@ -91,7 +91,7 @@ func resetIPSets(ips ipset.Interface) error {
 	common.Log.Debugf("Got list of ipsets: %v", sets)
 
 	for _, s := range sets {
-		common.Log.Debugf("Destroying ipsets '%s'", string(s))
+		common.Log.Debugf("Destroying ipset '%s'", string(s))
 		if err := ips.Destroy(s); err != nil {
 			common.Log.Errorf("Failed to destroy ipset '%s'", string(s))
 			return err
--- a/prog/weave-npc/main.go
+++ b/prog/weave-npc/main.go
@@ -90,6 +90,15 @@ func resetIPSets(ips ipset.Interface) error {
 
 	common.Log.Debugf("Got list of ipsets: %v", sets)
 
+	// Must remove references to ipsets by other ipsets before they're destroyed
+	for _, s := range sets {
+		common.Log.Debugf("Flushing ipset '%s'", string(s))
+		if err := ips.Flush(s); err != nil {
+			common.Log.Errorf("Failed to flush ipset '%s'", string(s))
+			return err
+		}
+	}
+
 	for _, s := range sets {
 		common.Log.Debugf("Destroying ipset '%s'", string(s))
 		if err := ips.Destroy(s); err != nil {


mikebryant:
I'm building and testing this locally at the moment, but would be nice to see what the integration tests have to say

bboreham:
I'll push the branch.
Can you think of an additional unit test or smoke test for this change?

mikebryant:
Hmm. My hypothesis on why it's intermittent is that it destroys ipsets in an arbitrary order, so sometimes the name hashing makes it work, sometimes it doesn't.
But it's not alphabetical, so I'm not sure how to force it into a situation where they're in the wrong order and the test would fail.

I guess you could write a unit test that:
- creates a hash:ip ipset A
- creates a list:set ipset that references the hash ip B
- mocks out the ipset.List call, to return [B, A]
- runs the resetIPSets method

It's more of a regression than unit test I suppose, since it's quite contrived

mikebryant:
It's also worth noting that even with this patch, we're occasionally hitting the same situation after weave-npc has been up for several minutes (which currently means it crashes and gets restarted, and now the flush logic fixes the situation) - but this definitely isn't ideal

mikebryant:
I think we should probably track the crashing-at-runtime bug in #2939 though, it's a bit different to the issue this fixes, and can't be fixed as easily

PR #2952, opened 2017-05-12
WIP: Don't mount root as part of `weave launch`

bboreham:
There's a comment in the script: "We don't mount '/' to avoid recursive mounts of '/var'", which relates to issues such as #1455.

So I changed the script to avoid mounting `/`.  This now means we have to compute the path to `/etc/resolv.conf` before the remote-exec, because it could be anywhere on the drive, but that move means we don't need a `chroot`

It turns out the attempt to find `weavedb` files wasn't working anyway for `attach|expose|hide`, and wasn't a good idea, so I took that out.

WIP because two tests are failing and I have no idea why.

### DIFF
--- a/weave
+++ b/weave
@@ -120,7 +120,7 @@ docker_run_options() {
 exec_options() {
     case "$1" in
         setup|setup-cni|launch|reset)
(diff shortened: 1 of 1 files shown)


bboreham:
Closing to tidy up PR list

PR #2954, opened 2017-05-12
Issues/2953 reorg images

bboreham:
Fixes #2953 

Results:
```
On master:
weaveworks/weave            latest      e3263035a80c        52 seconds ago      108 MB
weaveworks/weave-kube       latest      a34820b12cc7        22 seconds ago      150 MB
weaveworks/weave-npc        latest      aa9ee3fef304        13 seconds ago      54.7 MB
weaveworks/weaveexec        latest      d894ce240403        55 seconds ago      80.8 MB
-rw-rw-r-- 1 vagrant vagrant 56214606 May 12 16:42 weave.tar.gz

on 2953 branch
weaveworks/weave            latest      29e6d7d0d911        About a minute ago   58.2 MB
weaveworks/weave-kube       latest      e159fd3ec977        19 seconds ago       101 MB
weaveworks/weave-npc        latest      a9ace58c1338        10 seconds ago       54.7 MB
weaveworks/weaveexec        latest      a34044eb2e63        50 seconds ago       108 MB
-rw-rw-r-- 1 vagrant vagrant 56209157 May 12 16:45 weave.tar.gz
```

50MB less for weave-kube seems like a win; weaveexec is nominally bigger but the layers are shared so that's an ilusion

PR #2955, opened 2017-05-13
Don't check docker version when run --local

bboreham:
The standard case for `--local` is when the weave script in the `weaveexec` container is run by the `weave` script, which has just checked the version. Also in this case we check the version of the docker binary embedded in `weaveexec`, which is entirely predictable.

`--local` is also used from `weave-kube` where we don't mount the docker socket and therefore cannot check.

Since `--local` is undocumented we will not worry about any other cases; the worst that can happen is a confusing error message. We are not using any recently-added features of the `docker` command so will most likely work with whatever docker version the user has.

This came up while working on #2954; it seems like a worthwhile (if small) performance improvement for other cases, e.g. Weave Scope running `weave ps --local` every three seconds.

### DIFF
--- a/weave
+++ b/weave
@@ -218,8 +218,6 @@ kill_container() {
 # main
 ######################################################################
 
-check_docker_version
-
 [ "$1" = "--local" ] && shift 1 && IS_LOCAL=1
 
 # "--help|help" are special because we always want to process them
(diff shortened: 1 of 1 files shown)


marccarre:
Closing has already in #2954's 1e2dc43df6e3cf336b0d785864dd5a1825f4f829.

PR #2959, opened 2017-05-15
fix(npc): create ipsets with only valid xml characters

### LINKED ISSUE #2958: npc: shortName() includes chars that are invalid in xml
As part of #2939 I was trying to parse the `ipset list` output to find references from one set to another and unlink them before destroying. I thought that parsing the raw stdout was likely a bit unmaintainable but there is an xml output option. Unfortunately the output is not valid xml due to the occasional bad char in the fields.

The ShortNameSymbols in `npc/shortname.go` include chars that are invalid in xml [<>&].

Can we replace them with other chars? Maybe  [¬`+]

stuart-warren:
Why:
* xml output of ipset is invalid because ipset itself doesn't appear to
 escape invalid chars

* fixes #2958

### DIFF
--- a/npc/shortname.go
+++ b/npc/shortname.go
@@ -7,12 +7,13 @@ import (
 
 const (
 	// This array:
-	// * Must include only printable UTF8 characters that are represented with a single byte
+	// * Must include only printable UTF8 characters that are represented with a single byte (http://www.utf8-chartable.de/)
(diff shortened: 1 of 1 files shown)


marccarre:
For the record, the only reason CircleCI showed up as "failed" is because of a lint error which has been fixed in #2962.

Issue #2036, opened 2016-03-10
Control weave via unix domain socket


awh:
This would obviate any iptables issues (e.g. #2035) and allow access control via filesystem permissions.

PR #2960, opened 2017-05-15
Remove proxy status.sock

bboreham:
This is a hold-over from the multi-container implementation, so we can slim things down a bit by removing it.

Although #2036 recommends going in the opposite direction :thinking:

### DIFF
--- a/prog/weaver/main.go
+++ b/prog/weaver/main.go
@@ -258,9 +258,11 @@ func main() {
 
 	var waitReady common.WaitGroup
 
+	var proxy *weaveproxy.Proxy
+	var err error
 	if proxyConfig.Enabled {
 		// Start Weave Proxy:
-		proxy, err := weaveproxy.NewProxy(*proxyConfig)
+		proxy, err = weaveproxy.NewProxy(*proxyConfig)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


prog/weaver/main.go
```diff
@@ -441,6 +442,7 @@ func main() {
 		HandleHTTP(muxRouter, version, router, allocator, defaultSubnet, ns, dnsserver, &waitReady)
 		HandleHTTPPeer(muxRouter, allocator, discoveryEndpoint, token, name.String())
 		muxRouter.Methods("GET").Path("/metrics").Handler(metricsHandler(router, allocator, ns, dnsserver))
+		muxRouter.Methods("GET").Path("/proxyaddrs").HandlerFunc(proxy.StatusHTTP)
```
marccarre: If `!proxyConfig.Enabled`, `proxy` will be `nil`.
Will not this panic with `panic: runtime error: invalid memory address or nil pointer dereference`?

bboreham: Yes, I think it will, if called.  We should have a smoke-test for that.

bboreham: Interestingly `weave status` doesn't indicate the problem, since `weaver` can just crash and restart.

PR #2962, opened 2017-05-15
Fix shfmt lint error by escaping $ in sed expression

marccarre:
This fixes new error raised by latest version of shfmt, leading to build failures.


bboreham:
Do we want this on the 1.9 branch?

marccarre:
Yes, we do, I had based it on `1.9` but forgot to set that in the PR.

PR #2965, opened 2017-05-15
Improve is_plugin_v1_enabled to cover all cases.

### LINKED ISSUE #2956: is_plugin_v1_enabled is bugged
It is used in two places: before removing the default network `weave` in `stop` and `reset` and before removing left-over plugin sockets in `launch` and `reset`.  In both cases we want to know whether the plugin is _running_.

It works by checking the arguments for `--plugin`.  This misses the case `--plugin=false` which would disable it. And it does not check if the container with that argument is running.

marccarre:
Fixes #2956.

### DIFF
--- a/weave
+++ b/weave
@@ -583,7 +583,11 @@ attach() {
 
 # Docker Plugin V1 runs in a Docker container.
 is_plugin_v1_enabled() {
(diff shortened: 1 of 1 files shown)


marccarre:
Closing as better resolved by #2966

PR #2966, opened 2017-05-15
Simplify checks for plugin running

### LINKED ISSUE #2709: weave can transparently fail if weave-kube is running

### LINKED ISSUE #2956: is_plugin_v1_enabled is bugged

bboreham:
Fixes #2956 
Fixes #2709

1. `weave status` now does plugin-v1 and plugin-v2 in Go code.
2. all other uses of `is_plugin_vX_enabled`collapse to "is weaver running?"
2a. in `stop` before calling `remove-plugin-network`, if plugin was disabled then `RemoveNetwork` will return `NoSuchNetwork`
2b. in `protect_against_docker_hang` if weaver is not running it's ok to remove the plugin sockets.
2c. in `launch` we now error if weaver is already running for any reason

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -57,8 +57,7 @@ func run(dockerClient *docker.Client, weave *weaveapi.Client, address, meshAddre
 	if !isPluginV2 {
 		Log.Println("Creating default 'weave' network")
 		options := map[string]interface{}{MulticastOption: "true"}
-		// TODO: the driver name should be extracted from pluginMeshSocket
-		dockerClient.EnsureNetwork("weave", "weavemesh", defaultSubnet, options)
+		dockerClient.EnsureNetwork("weave", pluginNameFromAddress(meshAddress), defaultSubnet, options)
 	}
 	ready()
 
@@ -70,7 +69,7 @@ func listenAndServe(dockerClient *docker.Client, weave *weaveapi.Client, address
 	if isPluginV2 {
 		name = pluginV2Name
 	} else {
-		name = strings.TrimSuffix(path.Base(address), ".sock")
+		name = pluginNameFromAddress(address)
 	}
 
 	d, err := netplugin.New(dockerClient, weave, name, scope, dns, isPluginV2, forceMulticast)
@@ -94,4 +93,9 @@ func listenAndServe(dockerClient *docker.Client, weave *weaveapi.Client, address
 	}()
 
 	return listener, nil
+}
+
+// Take a socket address like "/run/docker/plugins/weavemesh.sock" and extract the plugin name "weavemesh"
+func pluginNameFromAddress(address string) string {
+	return strings.TrimSuffix(path.Base(address), ".sock")
 }
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -18,6 +18,7 @@ import (
 
 const (
 	pluginV2Name    = "net-plugin"
+	defaultNetwork  = "weave"
 	MulticastOption = netplugin.MulticastOption
 )
 
@@ -53,11 +54,11 @@ func run(dockerClient *docker.Client, weave *weaveapi.Client, address, meshAddre
 		}
 		defer os.Remove(meshAddress)
 		defer meshListener.Close()
-	}
-	if !isPluginV2 {
-		Log.Println("Creating default 'weave' network")
-		options := map[string]interface{}{MulticastOption: "true"}
-		dockerClient.EnsureNetwork("weave", pluginNameFromAddress(meshAddress), defaultSubnet, options)
+		if !isPluginV2 {
+			Log.Printf("Creating default %q network", defaultNetwork)
+			options := map[string]interface{}{MulticastOption: "true"}
+			dockerClient.EnsureNetwork(defaultNetwork, pluginNameFromAddress(meshAddress), defaultSubnet, options)
+		}
 	}
 	ready()
 
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -24,40 +24,62 @@ const (
 
 var Log = common.Log
 
-func Start(weaveAPIAddr string, dockerClient *docker.Client, address string, meshAddress string, dns bool, isPluginV2, forceMulticast bool, defaultSubnet string, ready func()) {
+type Config struct {
+	Socket            string
+	MeshSocket        string
+	Enable            bool
+	EnableV2          bool
+	EnableV2Multicast bool
+	DNS               bool
+	DefaultSubnet     string
+}
+
+type Plugin struct {
+	Config
+}
+
+func NewPlugin(config Config) *Plugin {
+	if !config.Enable && !config.EnableV2 {
+		return nil
+	}
+	plugin := &Plugin{Config: config}
+	return plugin
+}
+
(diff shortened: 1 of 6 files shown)


plugin/plugin.go
```diff
@@ -57,8 +57,7 @@ func run(dockerClient *docker.Client, weave *weaveapi.Client, address, meshAddre
 	if !isPluginV2 {
 		Log.Println("Creating default 'weave' network")
 		options := map[string]interface{}{MulticastOption: "true"}
-		// TODO: the driver name should be extracted from pluginMeshSocket
-		dockerClient.EnsureNetwork("weave", "weavemesh", defaultSubnet, options)
+		dockerClient.EnsureNetwork("weave", pluginNameFromAddress(meshAddress), defaultSubnet, options)
```
marccarre: Minor/Nitpick: replace `"weave"` with a constant.

plugin/status.go
```diff
@@ -0,0 +1,19 @@
...
+	DriverName     string
+	MeshDriverName string `json:"MeshDriverName,omitempty"`
+	Version        int
+}
+
+func NewStatus(address, meshAddress string, isPluginV2 bool) *Status {
+	status := &Status{
+		DriverName:     pluginNameFromAddress(address),
+		MeshDriverName: pluginNameFromAddress(meshAddress),
+		Version:        1,
```
marccarre: Minor/Nitpick: replace `1` with a constant, esp. as used in `prog/weaver/http.go` in `{{if eq .Plugin.Version 1}}`

plugin/status.go
```diff
@@ -0,0 +1,19 @@
...
+	}
+	if isPluginV2 {
+		status.Version = 2
```
marccarre: Minor/Nitpick: replace `2` with a constant.

prog/weaver/http.go
```diff
@@ -203,6 +205,18 @@ var statusTemplate = defTemplate("status", `\
...
 {{end}}\
+{{if .Proxy}}\
+
+        Service: proxy
+        Address: {{printList .Proxy.Addresses}}
+{{end}}\
+{{if .Plugin}}\
+
+        Service: plugin (v{{.Plugin.Version}})
+{{if eq .Plugin.Version 1}}\
```
marccarre: Minor/Nitpick: replace `1` with a constant, esp. as used in `plugin/status.go` in `Version:        1,`

plugin/plugin.go
brb: Not related to your PR, but this should be moved to the `if meshAddress != '' {` branch.

prog/weaver/main.go
```diff
@@ -427,6 +427,12 @@ func main() {
 	}
 	checkFatal(router.CreateRestartSentinel())
 
+	// Slightly ugly that plugin status is separated from the plugin.Start call
```
brb: I'd suggest to slightly refactor the plugin package:

* Introduce the `func NewPlugin(...) *Plugin` function.
* Create the `func (p *Plugin) Start()` method.
* Create the `func (p *Plugin) Status() Status` method.

The change would allow us to pass the `*Plugin` instance instead of  `pluginStatus`.

plugin/status.go
bboreham: What were you thinking this constant should be called?

plugin/plugin.go
bboreham: fixed both; good spot @brb

plugin/status.go
marccarre: No strong opinion but maybe:
- `const PluginV1 = 1`?
- `const PluginV2 = 2`?

given we have `isPluginV2`.

bboreham: Fixed a different way

prog/weaver/main.go
bboreham: I did this refactoring.

prog/weaver/http.go
```diff
@@ -203,6 +205,20 @@ var statusTemplate = defTemplate("status", `\
...
+
+        Service: proxy
+        Address: {{printList .Proxy.Addresses}}
+{{end}}\
+{{if .Plugin}}\
+
+{{if .Plugin.Config.EnableV2}}\
+        Service: plugin (v2)
+{{else}}\
+        Service: plugin (v1)
```
brb: Shall we rename it to `plugin (legacy)`?

brb:
LGTM besides one nit and you probably don't want to commit the `prog/weaver/weaveutil` executable.

bboreham:
nits fixed

PR #2967, opened 2017-05-16
Change weavedb from an emptyDir to a hostPath volume

### LINKED ISSUE #2610: Improve persistence of IPAM data
_From @bboreham on October 6, 2016 9:30_

Currently weave-daemonset.yaml specifies an `emptyDir` volume which is deleted when the pod is deleted. Since upgrading `weave-kube` require deleting the pods, this is unsafe.

Reading http://kubernetes.io/docs/user-guide/volumes/, I don't see an obviously better choice.

_Copied from original issue: weaveworks/weave-kube#27_

bboreham:
`emptyDir` has the same lifetime as the pod, whereas we want our persistence to extend past that, e.g. when Weave Net is upgraded.

We assume that `/var/lib` is writable.

Fixes #2610

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
@@ -102,7 +102,8 @@ spec:
           type: spc_t
       volumes:
         - name: weavedb
-          emptyDir: {}
+          hostPath:
+            path: /var/lib/weave
         - name: cni-bin
           hostPath:
             path: /opt
--- a/weave
+++ b/weave
@@ -126,7 +126,7 @@ docker_run_options() {
 
 exec_options() {
     case "$1" in
-        setup|setup-cni|launch)
+        setup|setup-cni|launch|reset)
             echo -v /:/host -e HOST_ROOT=/host
             ;;
         # All the other commands that may create the bridge and need machine id files.
@@ -1634,6 +1634,7 @@ case "$COMMAND" in
         protect_against_docker_hang
         VOLUME_CONTAINERS=$(docker ps -qa --filter label=weavevolumes)
         [ -n "$VOLUME_CONTAINERS" ] && docker rm -v $VOLUME_CONTAINERS  >/dev/null 2>&1 || true
+        rm -f $HOST_ROOT/var/lib/weave/weave-netdata.db >/dev/null 2>&1 || true
         destroy_bridge
         for LOCAL_IFNAME in $(ip link show | grep v${CONTAINER_IFNAME}pl | cut -d ' ' -f 2 | tr -d ':') ; do
             ip link del ${LOCAL_IFNAME%@*} >/dev/null 2>&1 || true
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -117,6 +117,7 @@ post_start_actions &
 /home/weave/weaver $EXTRA_ARGS --port=6783 $(router_bridge_opts) \
      --host-root=$HOST_ROOT \
      --http-addr=$HTTP_ADDR --status-addr=$STATUS_ADDR --docker-api='' --no-dns \
+     --db-prefix="/weavedb/weave-net" \
      --ipalloc-range=$IPALLOC_RANGE $NICKNAME_ARG \
      --ipalloc-init $IPALLOC_INIT \
      --conn-limit=$CONN_LIMIT \
--- a/prog/weave-kube/weave-daemonset.yaml
+++ b/prog/weave-kube/weave-daemonset.yaml
@@ -62,7 +62,8 @@ spec:
       restartPolicy: Always
       volumes:
         - name: weavedb
-          emptyDir: {}
+          hostPath:
+            path: /var/lib/weave
         - name: cni-bin
           hostPath:
             path: /opt


marccarre:
As discussed [on Slack](https://weaveworks.slack.com/archives/C0DNGPKLY/p1494943165658148), this requires a similar change in the Launch Generator.

marccarre:
Should we add a step to delete this directory in the tests?

bboreham:
> Should we add a step to delete this directory in the tests?

I guess so; from the point of view of ensuring that we test the same thing each time.  
Ideally we would extend the tests to do some sort of 'delete pods and restart' to check persistence worked.

marccarre:
> Ideally we would extend the tests to do some sort of 'delete pods and restart' to check persistence worked.

Agreed.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -102,7 +102,8 @@ spec:
           type: spc_t
       volumes:
         - name: weavedb
-          emptyDir: {}
+          hostPath:
+            path: /var/lib/weave-net
```
brb: `net-plugin` uses `/var/lib/weave`. I think we should unify. Thoughts?

marccarre: I'd say we should unify to `/var/lib/weave/net`
**if** we foresee other open source projects will need store things here too,
**else** just `/var/lib/weave`

bboreham: Thanks @brb; I didn't realise that.

bboreham:
> Should we add a step to delete this directory in the tests?

I now think `weave reset` should clear out this persistence directory; this will help, e.g. if someone needs to move a peer from one cluster to another.

(it isn't easy to run `weave reset` in a Kubernetes context, but that's a different problem)

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
brb: Just let me know the final name (I'm fine with `/var/lib/weave/net`), and I will change it in the plugin-v2.

bboreham: I think the directory `/var/lib/weave` is fine, but we could use the `--db-prefix` setting to put "net" into the filename.

I guess there are few enough plugin-v2 users that we don't mind disrupting them slightly

marccarre:
@brb, what do you think of these latest changes? (I'll wait for your final review before I merge)

weave
```diff
@@ -1634,6 +1634,7 @@ case "$COMMAND" in
         protect_against_docker_hang
         VOLUME_CONTAINERS=$(docker ps -qa --filter label=weavevolumes)
         [ -n "$VOLUME_CONTAINERS" ] && docker rm -v $VOLUME_CONTAINERS  >/dev/null 2>&1 || true
+        rm -f $HOST_ROOT/var/lib/weave/weave-netdata.db >/dev/null 2>&1 || true
```
brb: Where does `weave-netdata.db` come from? Grepping for `netdata` does not reveal anything.

bboreham: prefix "/var/lib/weave/weave-net" + https://github.com/weaveworks/weave/blob/master/db/boltdb.go#L33

brb: Ah, I see. Shouldn't it be `/var/lib/weave/` + `weavedb/weave-net` + `data.db` in the case of weave kube?

bboreham: I couldn't see any point in having an extra directory there.  

If we expected multiple files it would make sense, although I would name the directory `/var/lib/weave/net` or similar to make the grouping plain.

marccarre: On my end I have, like Martynas: 
[1] `/var/lib/weave` + 
[2] `/weavedb/weave-net` + 
[3] `data.db` = 
`/var/lib/weave/weavedb/weave-netdata.db` != 
`/var/lib/weave/weave-netdata.db`

so it would seem the path is missing `weavedb` (unless I've missed something myself!)

[1] `hostPath: path: /var/lib/weave` from https://github.com/weaveworks/weave/pull/2967/files#diff-63ba20698025213b0dce5f67c19f76dbR106

[2] `--db-prefix="/weavedb/weave-net"` from https://github.com/weaveworks/weave/pull/2967/files#diff-eb27072d080b69e928812b51e99c8f9dR120
and then from `&dbPrefix, []string{"-db-prefix"}, "/weavedb/weave"`:
- https://github.com/weaveworks/weave/blob/master/prog/weaver/main.go#L202
- https://github.com/weaveworks/weave/blob/master/prog/weaver/main.go#L320
- https://github.com/weaveworks/weave/blob/master/db/boltdb.go#L37
- https://github.com/weaveworks/weave/blob/master/db/boltdb.go#L33

[3] `fileName  = "data.db"` from: 
- https://github.com/weaveworks/weave/blob/master/db/boltdb.go#L33
- https://github.com/weaveworks/weave/blob/master/db/boltdb.go#L29

bboreham: What you are missing is that the directory shows up inside the container at `mountPath: /weavedb`. The directory `/var/lib/weave` is mapped to that, not concatenated.

marccarre: Thanks for clarifying it :+1:

PR #2971, opened 2017-05-16
[WIP] First cut of adding hairpin functionality to weave

deitch:
As discussed with @bboreham , there is an issue wherein kubernetes in CRI seems to have forgotten to handle hairpin. In pre-CRI, when kubelet received the veth back from the CNI plugin, it enabled hairpin (or not, depending on the kubelet command-line setting).

In CRI, it passes the flag on, and then just loses it. 

kubenet relies on the bridge plugin to enable hairpin. It appears to use an undocumented option in `/etc/cni/net.d/*.conf` to pass the hairpin mode. 

This first draft does the following:

1. Capture the env var `HAIRPIN_MODE` when launching the kube daemonset and sets it in `/etc/cni/net.d/10-weave.conf`
2. Reads the `hairpinMode` setting from the conf when running `weave-net` in CNI mode
3. Passes that setting on to `AttachContainer()`
4. If set to `true`, enable it in the veth on the bridge.
5. Add the `--hairpin-mode` option when running directly `weave attach-container .... `

Note that proxy mode just sets it to false because, well, I didn't know what to do that and better a default that is unchanged from current setting.

Note: This is a first draft, no tests or proper build runs. This is only a draft to see that the direction is correct before digging into tests.

@bboreham thoughts?


deitch:
FYI, CNI hairpin issue with kubernetes at https://github.com/kubernetes/kubernetes/issues/45790

bboreham:
Thanks for this, @deitch.  I will need to digest for a bit.

You've plumbed the setting into the CNI config file but are not actually using that setting for `weave-kube` ?  (using the command-line arg instead)

net/veth.go
```diff
@@ -186,6 +186,15 @@ func AttachContainer(netNSPath, id, ifName, bridgeName string, mtu int, withMult
 		return err
 	}
 
+	// set hairpin mode
+	veth, err := netlink.LinkByName(name)
```
bboreham: This should probably be in the closure passed to `CreateAndAttachVeth`, where we already have the veth, and if you return an error there then cleanup will get called

deitch:
> You've plumbed the setting into the CNI config file but are not actually using that setting for weave-kube ? (using the command-line arg instead)

I thought I did. Over [here](https://github.com/weaveworks/weave/pull/2971/files#diff-d8d91a307b3fdfc30c8904386b7687efR225) it loads it into the `NetConf struct`, and passes it to `AttachContainer()` over [here](https://github.com/weaveworks/weave/pull/2971/files#diff-d8d91a307b3fdfc30c8904386b7687efR131) 

The command-line [here](https://github.com/weaveworks/weave/pull/2971/files#diff-0f5d6635f82aae77377b8e67d2b45447R20) is because `weaveutil` also can be called in non-CNI mode, and thus needs to be able to have `attach-container <options>` as well, since that also calls `AttachContainer()` over [here](https://github.com/weaveworks/weave/pull/2971/files#diff-0f5d6635f82aae77377b8e67d2b45447R65) .



Did I misunderstand how it is used?

net/veth.go
deitch: Did you mean over [here](https://github.com/deitch/weave/blob/b75543d0e5ab8733fb9af4494742445ab8943c50/net/veth.go#L127)?

I thought about it, but there were 2 reasons I didn't.

1. That only gets called f the interface does not yet exist in the namespace [here](https://github.com/deitch/weave/blob/b75543d0e5ab8733fb9af4494742445ab8943c50/net/veth.go#L121). In theory that should never happen, but...
2. Isn't the argument to the callback the side of the veth *in* the container? Don't we want to set hairpin on the side attached to the bridge?

For the second, I guess we could either (a) pass both back in the callback or (b) determine it from the veth.

Out of curiosity, why is it in a callback and not in the next step? Is it just to wrap cleanup?

deitch:
@bboreham based on your comment [here](https://github.com/kubernetes/kubernetes/issues/45790#issuecomment-302203283), does this mean this should be remove as options, and should just always be set to true? Certainly would simplify the code, just let me know.

deitch:
So [now](https://github.com/kubernetes/kubernetes/issues/45790#issuecomment-302316571) it looks like it should not always be set, but should be an option. Is that correct?

If so, then this PR with its options remains relevant. Do let me know if we are going down the right path with the options here and your comments above, and I will clean up the code here.

deitch:
@bboreham, here is a question. 

1. Once a cluster manager sets `HAIRPIN_MODE=true` in the yaml, that will be true for every container ever attached to the weave network.
2. We had originally thought this was something kubelet was going to pass to CNI plugin, probably via `/etc/cni/net.d/*.conf`, perhaps env var, but somehow. Now we understand that it is something the admin will set when first configuring weave.

In other words, env var `HAIRPIN_MODE` (or its command-line equivalent `weave --hairpin-mode`), will be true or false but consistent across all containers on the weave network.

Given that, it seems silly to save the config into `/etc/cni/net.d/10-weave.conf` to be read later when kubelet invokes `weaveutil` by calling `/opt/cni/bin/weave-net`.

Is there any way to bypass `/etc/cni/net.d` entirely? Instead, when `launch.sh` in the weave-kube container calls `weaver` to have `weaver` set it as standard for all `veth`s attached to the bridge? I know of no way to do that at the bridge level - the `veth` is just an interface that happens to be attached to the bridge, AFAIK the hairpin setting as an _interface_ thing, not a _bridge_ thing - and `weaver`  and `weaveutil` are 2 independently called binaries; `weaver` does not call `weaveutil`; `kubelet` does.

Any way to do this and thus simplify it?

net/veth.go
bboreham: ok, my mistake; the host end of the veth is returned from `CreateAndAttachVeth`

It's a callback so we can do initialization on the peer side before bringing the host side up.

bboreham:
The CNI plugin runs quite independently from the `weaver` "router" daemon, so it's easier to put the option in a file.

The two could communicate via REST (they already do for IP allocation), but that seems overkill for this.

bboreham:
>> You've plumbed the setting into the CNI config file but are not actually using that setting for weave-kube ? 
> I thought I did. 

You are setting `HAIRPIN_MODE` and calling `weave --local setup-cni`, but I think you need to `export HAIRPIN_MODE` before it will go through.

bboreham:
I have pushed a couple of commits to the hairpin-option branch in the weaveworks/weave repo - building now at https://circleci.com/gh/weaveworks/weave/9074

deitch:
> The CNI plugin runs quite independently from the weaver "router" daemon, so it's easier to put the option in a file.
> The two could communicate via REST (they already do for IP allocation), but that seems overkill for this.

So just leave it in `/etc/cni/net.d/10-weave.conf`? OK, works for me.

deitch:
> You are setting HAIRPIN_MODE and calling weave --local setup-cni, but I think you need to export HAIRPIN_MODE before it will go through.

Oh that! It sort of works, I missed the unset case. If you set `HAIRPIN_MODE` in your `weave.yml`, then it is set as an env var in the container, so `weave` will inherit it, not just `launch.sh`. The case I missed is if it is not set, then the setting in `launch.sh` will be lost. Oops.

Looks like I made a few mistakes here:

1. Not letting `launch.sh` set it correctly (i.e. `export HAIRPIN_MODE`) in case it was not set
2. The default now should be set to `true`, not `false`
3. I passed it on to `weaver` in the end of `launch.sh` (no idea how I ended up doing that...., glad we have reviews!)

I will clean up 3 & 2. As for 1, I think it is redundant to set the env var to the default and then export the exact same env var so that `weave` script can set it in `10-weave.conf`. I think I either will set it entirely as env var in `weave` (and nothing in `launch.sh`) or have `launch.sh` pass it as `weave --local setup-cni --hairpin-mode <mode>`.

deitch:
> I have pushed a couple of commits to the hairpin-option branch in the weaveworks/weave repo 

OK, looking now.

deitch:
Follow up:

1. Instead of `export HAIRPIN_MODE`, since it is set in the `weave.yml`, which makes it exported, I just read it in `weave`. Wasn't much point in reading it in `launch.sh` and blindly passing it on to `weave`.
2. I see you cleaned up the `veth` stuff. Will merge in...

deitch:
2 is done: took your changes and merged them in.

BTW, I made it all a single commit, overriding previous. I usually prefer multiple commits on a single PR, but the team reviewing PRs I have done for linuxkit wanted them all squashed. Happy to do it any way you want. For now, single commit.

Take a look?

deitch:
Some questions about tests.


#### Proper test
Trying to figure out the sane way to test this. We already have [840_weave_kube_3_test.sh](https://github.com/weaveworks/weave/blob/master/test/840_weave_kube_3_test.sh) which sets up and tears down kubernetes with weave. 

It would seem easiest either to extend that file or copy it three times: `HAIRPIN_MODE=true`, `HAIRPIN_MODE=false`, `HAIRPIN_MODE` unset. 

My concern is how long it will take.

Alternatively, we could just remove the weave daemonset and then relaunch it on the same cluster, each time changing the `HAIRPIN_MODE` env var. 

Either way, I think we would want to test by checking that values set to `0` for `/sys/devices/virtual/net/weave/brif/veth*/hairpin_mode`

Some pointers?


#### Builds all fail
All of the builds seem to fail on go dependencies? here is the output (from your [link above](https://circleci.com/gh/weaveworks/weave/9074), same for the next one):

```
Step 8 : RUN go get 	github.com/golang/lint/golint 	github.com/fzipp/gocyclo 	gopkg.in/mvdan/sh.v1/cmd/shfmt 	github.com/fatih/hclfmt 	github.com/client9/misspell/cmd/misspell
 ---> Running in 3ce2ab1f866c
# gopkg.in/mvdan/sh.v1/cmd/shfmt
src/gopkg.in/mvdan/sh.v1/cmd/shfmt/main.go:26: undefined: "github.com/mvdan/sh/syntax".ParseMode
src/gopkg.in/mvdan/sh.v1/cmd/shfmt/main.go:27: undefined: "github.com/mvdan/sh/syntax".PrintConfig
src/gopkg.in/mvdan/sh.v1/cmd/shfmt/main.go:46: undefined: "github.com/mvdan/sh/syntax".ParseComments
src/gopkg.in/mvdan/sh.v1/cmd/shfmt/main.go:48: undefined: "github.com/mvdan/sh/syntax".PosixConformant
src/gopkg.in/mvdan/sh.v1/cmd/shfmt/main.go:73: undefined: "github.com/mvdan/sh/syntax".Parse
src/gopkg.in/mvdan/sh.v1/cmd/shfmt/main.go:139: undefined: "github.com/mvdan/sh/syntax".Parse
The command '/bin/sh -c go get 	github.com/golang/lint/golint 	github.com/fzipp/gocyclo 	gopkg.in/mvdan/sh.v1/cmd/shfmt 	github.com/fatih/hclfmt 	github.com/client9/misspell/cmd/misspell' returned a non-zero code: 2

bin/circle-dependencies-post returned exit code 1
```

bboreham:
The `mvdan/sh` thing is worked around in 72532324ed54d9a2c3ac53e74688b947e8866dba; if you rebase you will pick it up.

bboreham:
Re testing I think the most important thing is to test hairpining actually works, which requires

deitch:
> if you rebase you will pick it up 

OK, just synced the fork and rebased the branch and pushed out. Running now.

deitch:
> Re testing I think the most important thing is to test hairpining actually works, which we could do by launching the network-tester with a Kubernetes service address. I think this will work with the current image if just a single peer is specified. Or, feel free to think up something better.

That would test the _manifestation_ of the issue, but it could just as easily fail due to something else. To test that hairpinning option actually works, would it not make sense to test just the hairpin is set/not-set?

Come to think of it, why not do both?

1. We should test that a pod can reach itself by service address. It took us a while to discover that this existed and the source of it; why not have an actual test for that _functionality_?
2. Test that hairpin mode setting actually sets hairpin mode

Of course, the first only makes sense if hairpin mode is set.

So:

* when hairpin mode not set, test that value of `/sys/devices/....` is `0`
* when hairpin mode set to `false`, test that value of `/sys/devices/....` is `0`
* when hairpin mode set to `true`, test that value of `/sys/devices/....` is `0` **and** test that a pod can reach itself on service policy

OK, fine I will leverage the existing tests. Give me some pointers on the quick setup scripts that make it easy? And, yeah, a single kube node would be more than enough to test it.

deitch:
Ah. Build failing because `net/veth.go:141: undefined: cleanup`. Fixing it...

deitch:
That one is fixed, but builds still failing. I don't get this one:

```
$ $SRCDIR/bin/circle-teardown-pre
exit status 1: can't load package: package _/home/ubuntu/weave/api: cannot find package "_/home/ubuntu/weave/api" in any of:
	/usr/local/go/src/_/home/ubuntu/weave/api (from $GOROOT)
	/home/ubuntu/src/_/home/ubuntu/weave/api (from $GOPATH)
	/home/ubuntu/.go_workspace/src/_/home/ubuntu/weave/api
	/usr/local/go_workspace/src/_/home/ubuntu/weave/api


$SRCDIR/bin/circle-teardown-pre returned exit code 1

Action failed: $SRCDIR/bin/circle-teardown-pre
```

bboreham:
I think that is a quirk of the coverage-gathering code that we sometimes see; haven't tracked it down.

I have re-pushed your branch in the weaveworks repo so it will run the full smoke-test suite - https://circleci.com/gh/weaveworks/weave/9079

deitch:
I just updated the documentation. Assuming that the current run passes, just need to add the tests.

deitch:
Strange. So your run 9079 passed but mine 9080 (just changed the docs) failed with that same error.

bboreham:
Note that we run 3rd-party branches in a limited environment, to avoid having our testing machines abused.  That isn't why you're getting that second error, but it is why I'm not focused on it; unless we give you write access to the repo you have to wait for someone else to trigger the full build.

deitch:
> someone else to trigger the full build

@bboreham = someone else! :-)

The build you launched did go through cleanly though, so that looks good. Are you good with the design and everything to date, just some good tests and we will be good?

net/veth.go
```diff
@@ -136,6 +136,11 @@ func AttachContainer(netNSPath, id, ifName, bridgeName string, mtu int, withMult
 		if err != nil {
 			return err
 		}
+		// set hairpin mode
```
bboreham: comment seems superfluous

site/kube-addon.md
```diff
@@ -192,9 +192,9 @@ $ kubectl logs <weave-pod-name-as-above> -n kube-system weave-npc
 
 When the Weave Network Policy Controller blocks a connection, it logs the following details about it:
 
-* protocol used, 
-* source IP and port, 
-* destination IP and port, 
+* protocol used,
```
bboreham: please don't include unrelated changes (removal of space at end of lines)

site/using-weave/dynamically-attach-containers.md
```diff
@@ -5,7 +5,7 @@ search_type: Documentation
 ---
 
 
-When containers may not know the network to which they will be attached, Weave Net enables you to dynamically attach and detach containers to and from a given network, even when a container is already running. 
+When containers may not know the network to which they will be attached, Weave Net enables you to dynamically attach and detach containers to and from a given network, even when a container is already running.
```
bboreham: unrelated change

weave
```diff
@@ -1197,6 +1202,9 @@ launch() {
                 IPRANGE="${1#*=}"
                 IPRANGE_SPECIFIED=1
                 ;;
+            --hairpin-mode=*)
+                HAIRPIN_MODE="${1#*=}"
```
bboreham: I don't think we should bother with this; users can set the environment variable if they really want to.
(This comment is in the context that this setting only gets used in the CNI code path)

site/using-weave/dynamically-attach-containers.md
```diff
@@ -37,6 +37,13 @@ specify `--rewrite-hosts` when running `weave attach`:
 
     host1$ weave attach --rewrite-hosts c1
 
+Weave Net will set the bridge side of the `veth` pair to enable hairpin mode when attaching a container.
```
bboreham: This is not true as the PR currently stands.

bboreham:
Yes; happy with the broad approach; I have some nits I will post in a minute.

site/kube-addon.md
deitch: Probably the editor doing that automatically.

site/using-weave/dynamically-attach-containers.md
deitch: Same.

site/using-weave/dynamically-attach-containers.md
deitch: Yeah, good point. We need to change that. We want `hairpin=true` by default, do we not?

weave
deitch: I think we would want it to be a general option. Hairpin mode is set by default whenever you use weave. It can be disabled via env var (when in CNI) or command-line option. This fits as well with the other options that are controlled via options to `weave` script.

deitch:
Fixed most of the nits you raised. Just need to make the default `hairpin=true`

site/using-weave/dynamically-attach-containers.md
deitch: Default now set to true. I didn't love the `switch` statement in `attach.go`, but since the default is `true`, then we don't really need a regex, just a straight compare of `--hairpin-mode=false`

deitch:
@bboreham : extended the [830_cni_plugin_test.sh](https://github.com/weaveworks/weave/pull/2971/files#diff-e355707fcc8089d42aa04444e35341d8) to create three more containers, connect them with CNI, and use 3 different configs. It then goes into the containers, gets the veth peer ID, and checks hairpin mode on it.

Have a look-see please.

test/830_cni_plugin_test.sh
```diff
@@ -85,6 +113,24 @@ assert_raises "[ $C2IP != $C3IP ]"
 assert_raises "[ $BRIP != $C3IP ]"
 assert_raises "exec_on $HOST1 c1 $PING $C3IP"
 
+# get the value of hairpin on the veth for each of ch0, ch1, ch2
+CH0_PEER_ID=$(exec_on $HOST1 ch0 cat /sys/class/net/eth0/iflink)
+CH1_PEER_ID=$(exec_on $HOST1 ch1 cat /sys/class/net/eth0/iflink)
+CH2_PEER_ID=$(exec_on $HOST1 ch2 cat /sys/class/net/eth0/iflink)
+
+CH0_PEER=$(run_on $HOST1 ip link show | awk "/^${CH0_PEER_ID}:/"' {print $2}' | sed 's/@.*$//g' )
```
bboreham: Unfortunately `run_on` adds decoration to the output; you need to drop down to `$SSH` to get this to work.

But, we can take advantage of the special naming of veths created by the CNI plugin and boil this down to:

```
assert "$SSH $HOST1 cat /sys/devices/virtual/net/weave/brif/vethwepl${CH0:0:7}/hairpin_mode" "1"
```

Yes this is fragile to changes in the implementation, but should break fairly obviously and about the same as being fragile to changes in `ip link show` format.

weave
bboreham: I could understand this argument if `--hairpin-mode` did anything in non-CNI modes, or if you installed via `weave setup`, or maybe some other cases.  

But the only thing it does is affects the CNI config file generated the first time you run `weave launch`, which I find too specific to be worth doing.

test/830_cni_plugin_test.sh
```diff
@@ -85,6 +113,24 @@ assert_raises "[ $C2IP != $C3IP ]"
...
+CH2_PEER=$(run_on $HOST1 ip link show | awk "/^${CH2_PEER_ID}:/"' {print $2}' | sed 's/@.*$//g' )
+
+CH0_HAIRPIN=$(run_on $HOST1 cat /sys/devices/virtual/net/weave/brif/$CH0_PEER/hairpin_mode)
+CH1_HAIRPIN=$(run_on $HOST1 cat /sys/devices/virtual/net/weave/brif/$CH1_PEER/hairpin_mode)
+CH2_HAIRPIN=$(run_on $HOST1 cat /sys/devices/virtual/net/weave/brif/$CH2_PEER/hairpin_mode)
+
+# now we can compare what the modes should be
+assert_raises "[ $CH0_HAIRPIN == 1 ]"
+assert_raises "[ $CH1_HAIRPIN == 0 ]"
+assert_raises "[ $CH2_HAIRPIN == 1 ]"
```
bboreham: This test fails because the default for CNI if no value is specified is `false`

deitch:
Just came back from a few days, checking comments now...

test/830_cni_plugin_test.sh
deitch: I was wondering how the naming worked. I can live with that fragility if you can. I like going into the container and getting the peer interface number better, but, yeah, can live with this. Maybe we should at least leave a comment in case future changes break it, save them the headache of digging.

test/830_cni_plugin_test.sh
deitch: Shouldn't the default be `true`? Isn't that what we want? Based on the conversation on the k8s issue - where you raised the desirability of making hairpin always set to `true` and ignoring kernels that have an issue with it - I had understood (and it made sense to me) that hairpin always would be `true`, and you would need an explicit option to turn if off.

deitch:
Tests cleaned up to use your "underhanded knowledge". :-)

I just added the comments about how you would go about it the long (and less fragile) way should you ever need to. If it ever breaks, save someone the work.

test/830_cni_plugin_test.sh
deitch: Tests cleaned up to use your "underhanded knowledge". :-)

I just added the comments about how you would go about it the long (and less fragile) way should you ever need to. If it ever breaks, save someone the work.

weave
deitch: OK, let us raise it then. Should the default not be set to `true`? Should it not be uniform behaviour across the board, wherein `weave launch` also will run in hairpin mode unless disabled? That would mean putting it in `weaver`, which is the entrypoint to the `weaveworks/weave` container, right?

Or maybe we should leave it as 2 separate PRs...

bboreham: Yes I think the default should be `true`.
Yes we could add `--hairpin-mode` as an option to `weaver` which would apply to the proxy (Docker does its own thing for the plugin).
Yes I think that would be best as a separate PR since I would like the key bug-fix for CNI to go on the 1.9 branch and the `main.go` for `weaver` has changed a lot between 1.9 and 2.0.
So I would suggest just removing the four lines from the `launch)` case in this PR.

bboreham:
Tests passed at https://circleci.com/gh/weaveworks/weave/9142.
I'm going to try cherry-picking onto the 1.9 branch now.

site/using-weave/dynamically-attach-containers.md
bboreham: For extra clarity, here is the line where `weave attach` will call down to `weaveutil`: https://github.com/deitch/weave/blob/7a83fd829b9b65cc3c358cb8fd801fcf024f06bb/weave#L1507

it does not pass any arguments straight from the command-line.

weave
deitch: > Yes I think the default should be true
> Yes I think that would be best as a separate PR

That makes sense. OK.

site/using-weave/dynamically-attach-containers.md
```diff
@@ -37,6 +37,13 @@ specify `--rewrite-hosts` when running `weave attach`:
...
+If you do *not* want hairpin mode enabled, e.g. your kernel is one of those that panics for unknown reasons
+when hairpin is enabled on a `veth`, you can set it to false:
+
+host1$ weave attach --hairpin-mode=false c1
```
bboreham: this doesn't work because `weave attach` will not pass the option on.

deitch:
OK, I removed those lines from `weave` and the documentation. Now it is CNI only, rest left for future PR.

deitch:
> OK, I removed those lines from weave and the documentation. Now it is CNI only, rest left for future PR

Ugh, git hash conflicts. Will push in a minute.

deitch:
OK, I think I got it now. I will check that git conflict raised above

deitch:
And that is done too.

Hey, @bboreham, did I see that the dynamic creation of the kube yml is done?

bboreham:
I created another branch hairpin-mode-1.9 for merging into the 1.9 branch - tests running at https://circleci.com/gh/weaveworks/weave/9147

bboreham:
@deitch yes, we now have a yaml generator described at https://github.com/weaveworks/weave/blob/master/site/kube-addon.md#using-cloudweaveworks, though it is not announced/released as yet.

site/using-weave/dynamically-attach-containers.md
deitch: Right. So I removed it. As you said, separate PR, let's get the CNI working.

deitch:
> we now have a yaml generator described at https://github.com/weaveworks/weave/blob/master/site/kube-addon.md#using-cloudweaveworks, though it is not announced/released as yet.

Oh lovely. You would not believe the awk-fu I had to do to add it in. I probably should wait until released to remove that, but even so, real nice. 

ETA?

deitch:
And I see that the git.io address just redirects to github.com, which obviously doesn't process the query string params. Is there intent to change the git.io redirect to point to cloud.weave.works?

bboreham:
> Is there intent to change the git.io redirect to point to cloud.weave.works?

Yes; see https://github.com/weaveworks/weave/pull/2903#issuecomment-303103666

deitch:
I am confused. Based on that comment:

> No one uses /k8s/net and /k8s.yaml (cloud) doesn't include Net, so serving 2.0's YAML before 2.0 is actually released should be fine.

... no one uses it, so I should not? Or are we already at step 2, where I can?

bboreham:
Yes we are at step 2.  The service is in production for other Weaveworks components so it's unlikely we would randomly take it offline or break it.

deitch:
> Yes we are at step 2. The service is in production for other Weaveworks components so it's unlikely we would randomly take it offline or break it.

Ah, got it. The implication appeared to be that the generator put out yml that would only work with the not-yet-released weavenet 2.0. So I can go ahead and use it?

bboreham:
I don't think the changes in yaml are incompatible with the 1.9 release; however they do bundle up some changes like moving the persistence file from a pod volume to a `hostPath` volume.

bboreham:
@deitch are you happy with this subset of your changes which I propose to merge into the 1.9 branch:
https://github.com/weaveworks/weave/compare/1.9...hairpin-option-1.9

bboreham:
@deitch you now have a merge commit on this branch, which I suspect will go away if you rebase

bboreham:
Tests passed on the 1.9 branch at https://circleci.com/gh/weaveworks/weave/9154; I'm going to create another PR to merge that.

deitch:
Looks like that one is good. Can I close this PR in favour of that?

bboreham:
Sure

deitch:
When does the image go out, such that I can remove the `--enable-cri=false` from my kubelet launch?

bboreham:
@deitch v1.9.6 out now https://github.com/weaveworks/weave/releases/v1.9.6

PR #2972, opened 2017-05-17
Fixing cross cluster communication

neith00:
When using weave with kubernetes between 2 clusters, **all** ingress traffic is dropped by default.
This commit enables communication between 2 clusters by dropping all the traffic **but** from weave interface/subnet.


neith00:
Unfortunately it seems the CI is missing a token to run (See the logs above)

bboreham:
Sorry, but this change would break the behaviour of Kubernetes network policies.
The intended behaviour is that all incoming connections, from anywhere, are blocked unless specifically allowed by policy.
Let's diagnose your actual issue at #2973.

Issue #2973, opened 2017-05-17
Enabling cross cluster communication in k8s


neith00:
When using weave with k8s to connect two k8s clusters, the default iptables configuration prevent traffic going from pods on cluster one to pods of cluster two.
This PR https://github.com/weaveworks/weave/pull/2972 fix the iptables rules to fix this issue.

bboreham:
A connection should be allowed if the namespace is in `DefaultAllow` mode (the default), or by a more specific [network policy](https://kubernetes.io/docs/concepts/services-networking/networkpolicies/).

Since you mention two clusters, you are perhaps hitting the limitation that Kubernetes Network Policies can only refer specifically to one cluster.
You can write non-specific rules, opening all connections to a port.

Or we could consider some extension to Kubernetes Network Policies to support your use-case.

Please can you explain more about what you want to do.

neith00:
As you ask on slack here is the iptables output
```
# Generated by iptables-save v1.6.0 on Wed May 17 11:07:04 2017
*mangle
:PREROUTING ACCEPT [7553478:1939413210]
:INPUT ACCEPT [7553474:1939412922]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [7587220:1982189840]
:POSTROUTING ACCEPT [7587217:1982189660]
:WEAVE-IPSEC-IN - [0:0]
:WEAVE-IPSEC-IN-MARK - [0:0]
:WEAVE-IPSEC-OUT - [0:0]
:WEAVE-IPSEC-OUT-MARK - [0:0]
-A INPUT -j WEAVE-IPSEC-IN
-A OUTPUT -j WEAVE-IPSEC-OUT
-A WEAVE-IPSEC-IN -s 192.168.204.30/32 -d 192.168.204.27/32 -p esp -m esp --espspi 3646789848 -j WEAVE-IPSEC-IN-MARK
-A WEAVE-IPSEC-IN-MARK -j MARK --set-xmark 0x20000/0x20000
-A WEAVE-IPSEC-OUT -s 192.168.204.27/32 -d 192.168.204.30/32 -p udp -m udp --dport 6784 -j WEAVE-IPSEC-OUT-MARK
-A WEAVE-IPSEC-OUT-MARK -j MARK --set-xmark 0x20000/0x20000
COMMIT
# Completed on Wed May 17 11:07:04 2017
# Generated by iptables-save v1.6.0 on Wed May 17 11:07:04 2017
*nat
[...]

PR #2979, opened 2017-05-18
weave-npc to check local addresses only

### LINKED ISSUE #2622: Filter ipset entries based on node
_From @awh on September 20, 2016 13:18_

For the sake of simplicity, the current implementation builds the same rules and ipsets on all nodes. This limits scalability - instead, a pod's IP should only be added to a node's matching ipsets if that pod is scheduled on that node (it may be possible to implement this by putting a filter on the k8s notification API subscription).

_Copied from original issue: weaveworks/weave-npc#6_

bboreham:
This branch is based off the branch for #2978, because I need the rule to contact kube-dns.
Only the last four commits are relevant to this PR.

This PR does two things: it fixes #2622 by only adding IP addresses from pods on the current node; and it then corrects a bug exposed by that change: weave-npc used to filter on the destination address at both ends of a connection.

I suspect this is the cause of #2973.

To address that, I create a new IPset which has all local pod addresses, and if a new connection comes from a local pod but does not target a local pod we let it through.  In this way all the real filtering is done at the destination end of the connection.

### DIFF
--- a/npc/controller.go
+++ b/npc/controller.go
@@ -29,6 +29,8 @@ type NetworkPolicyController interface {
 type controller struct {
 	sync.Mutex
 
+	nodeName string // my node name
+
 	ipt *iptables.IPTables
 	ips ipset.Interface
 
@@ -36,11 +38,12 @@ type controller struct {
 	nsSelectors *selectorSet   // selector string -> nsSelector
 }
 
-func New(ipt *iptables.IPTables, ips ipset.Interface) NetworkPolicyController {
+func New(nodeName string, ipt *iptables.IPTables, ips ipset.Interface) NetworkPolicyController {
 	c := &controller{
-		ipt: ipt,
-		ips: ips,
-		nss: make(map[string]*ns)}
+		nodeName: nodeName,
+		ipt:      ipt,
+		ips:      ips,
+		nss:      make(map[string]*ns)}
 
 	c.nsSelectors = newSelectorSet(ips, c.onNewNsSelector)
 
@@ -82,10 +85,21 @@ func (npc *controller) withNS(name string, f func(ns *ns) error) error {
 	return nil
 }
 
+func (npc *controller) checkLocalPod(obj *coreapi.Pod, event string) bool {
+	if obj.Spec.NodeName != npc.nodeName {
+		common.Log.Debugf("%s ignored for pod %s/%s on node %s", event, obj.ObjectMeta.Namespace, obj.ObjectMeta.Name, obj.Spec.NodeName)
+		return false
+	}
+	return true
+}
+
 func (npc *controller) AddPod(obj *coreapi.Pod) error {
 	npc.Lock()
 	defer npc.Unlock()
 
+	if !npc.checkLocalPod(obj, "AddPod") {
+		return nil
+	}
 	common.Log.Debugf("EVENT AddPod %s", js(obj))
 	return npc.withNS(obj.ObjectMeta.Namespace, func(ns *ns) error {
 		return errors.Wrap(ns.addPod(obj), "add pod")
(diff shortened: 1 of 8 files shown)


npc/controller.go
```diff
@@ -86,6 +89,10 @@ func (npc *controller) AddPod(obj *coreapi.Pod) error {
 	npc.Lock()
 	defer npc.Unlock()
 
+	if npc.nodeName != "" && obj.Spec.NodeName != npc.nodeName {
```
marccarre: I'd suggest to factorise this condition and similar ones in this commit in a function to better show the intent & hide implementation details.
Maybe something like:
```
func (npc *controller) isRemotePod(pod *coreapi.Pod) bool {
    return npc.nodeName != "" && pod.Spec.NodeName != npc.nodeName
}
```

or maybe the opposite `isLocalPod` which we would then negate.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -92,6 +92,11 @@ spec:
               cpu: 10m
           securityContext:
             privileged: true
+          env:
```
marccarre: I guess this implies the LG should be updated accordingly

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -92,6 +92,11 @@ spec:
...
+          - name: MY_NODE_NAME
```
marccarre: I am thinking `MY_NODE_NAME`'s name may lead people to think it isn't something actually used, or something which should be replaced with another name.
Maybe something like `WEAVE_NPC_NODE_NAME` would be more explicit and show it actually has a purpose as it stands?

bboreham: From that point of view, I think `NODE_NAME` is better.
The value is actually available in `$HOSTNAME`, for a pod in the host namespace, but I was afraid that was too obscure.  But today it doesn't seem so bad.

marccarre: `NODE_NAME` sounds good to me 👍 
`HOSTNAME` could indeed be a bit more obscure, e.g. what would happen if not in the host namespace? (even though this should always be the case for Weave Net)

marccarre:
Note that our linter seems to complain about:
`prog/weave-npc/main.go: run gofmt -s -w prog/weave-npc/main.go`

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
bboreham: `HOSTNAME` is also set by `bash`.

For other use-cases there is a command-line option `--node-name`.

prog/weave-npc/main.go
```diff
@@ -116,7 +117,14 @@ func createBaseRules(ipt *iptables.IPTables, ips ipset.Interface) error {
 
 func root(cmd *cobra.Command, args []string) {
 	common.SetLogLevel(logLevel)
-	common.Log.Infof("Starting Weaveworks NPC %s", version)
+	if nodeName == "" {
+		// HOSTNAME is set by Kubernetes for pods in the host network namespace
+		nodeName = os.Getenv("HOSTNAME")
+	}
+	if nodeName == "" {
+		common.Log.Fatalf("Must set node name via --node-name or $HOSTNAME")
```
marccarre: 👍

marccarre: Do you think we should even `exit(1)` or that is a bit too radical?

bboreham: `Fatalf` does call `os.Exit(1)`

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
marccarre: No longer applies, as the PR is no longer changing the YAML files and relies on `HOSTNAME` instead.

prog/weave-npc/main.go
marccarre: [TIL](https://github.com/sirupsen/logrus/blob/2c677e6a5e5ae8e1b967f1efcf2c3dbaaf625b2d/logger.go#L167); thanks for pointing this out!

bboreham:
Cherry-picked to 1.9 branch at 2d3e070288b67859060eeff8eb78a4b55519dbb4

PR #2980, opened 2017-05-19
wip: Retry iptables

### LINKED ISSUE #2939: ipset: v6.29: Set cannot be destroyed: it is in use by a kernel component

stuart-warren:
fixes #2939

In Kubernetes, multiple pods interact with iptables/ipsets. Unfortunately they are often using different binary versions or don't have access to a shared lock file.

For this reason we often run into problems like iptables exiting with status code 4 "temporarily unavailable" or ipset exiting with an "in use by a kernel component" error.

This change introduces: 
* retrying for iptables until the change is successful (expected to be temporary)
* multiple attempts for some ipset methods (to avoid getting stuck)

### DIFF
--- a/npc/ipset/ipset.go
+++ b/npc/ipset/ipset.go
@@ -67,7 +67,12 @@ func (i *ipset) FlushAll() error {
 
 func (i *ipset) Destroy(ipsetName Name) error {
 	i.removeSet(ipsetName)
-	return doExec("destroy", string(ipsetName))
+	// Loop until we get an exit code other than inUseErrStr
+	for {
+		if err := doExec("destroy", string(ipsetName)); !isErrInUse(err) {
+			return err
+		}
+	}
 }
 
 func (i *ipset) DestroyAll() error {
@@ -91,6 +96,18 @@ func (i *ipset) List(prefix string) ([]Name, error) {
 	}
 
 	return selected, err
+}
+
+var inUseErrStr = "Set cannot be destroyed: it is in use by a kernel component"
+
+// to catch inUseErrStr error
+func isErrInUse(err error) bool {
+	if ierr, ok := err.(*exec.ExitError); ok {
+		if strings.HasSuffix(ierr.Error(), inUseErrStr) {
+			return true
+		}
+	}
+	return false
 }
 
 func doExec(args ...string) error {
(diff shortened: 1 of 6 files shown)


npc/ipset/ipset.go
```diff
@@ -93,6 +106,18 @@ func (i *ipset) List(prefix string) ([]Name, error) {
 	return selected, err
 }
 
+var inUseErrStr = "Set cannot be destroyed: it is in use by a kernel component"
+
+// to catch inUseErrStr error
+func isErrInUse(err error) bool {
+	if ierr, ok := err.(*exec.ExitError); ok {
+		if strings.HasSuffix(ierr.Error(), inUseErrStr) {
```
stuart-warren: Any ideas on a better way to do this than looking for a string in the error?

bboreham:
Having seen https://github.com/kubernetes/kubernetes/pull/44895 I think that might be a better approach. We'd need to ensure we can mount the lock file by creating it before the container, but that should be feasible in the systemd unit or whatever is used to start the system.

dcbw:
Yeah, I'd like to not keep piling hacks on hacks here, the locking is the way to go to ensure things cooperate.  Relying on the undocumented error codes that iptables happens to return for EAGAIN/EBUSY or whatever it is seems fragile as well.

bboreham:
TIL that kubelet touches the lock file so have created #3134

bboreham:
I'm going to close this on the basis that #3134 is a better fix; if I'm mistaken please re-open.

Thanks for the contribution @stuart-warren!

PR #2981, opened 2017-05-19
First revision of how weave utils get launches

deitch:
As requested by @bboreham. This is _known_ to be incomplete (and probably partially wrong too). But at least if it is documented, someone won't have to reverse engineer all the starting points to figure out how to add a new option like I did. :-)

### DIFF
--- a/docs/entrypoints.md
+++ b/docs/entrypoints.md
@@ -1,0 +1,70 @@
+# Entrypoints
+
+This document describes the various ways that weave as a binary gets launched, and where and how it processes options.
+
+The goal is to enable maintainers to modify functionality without having to reverse engineer the code to find all of the places you can start a weave binary and what it depends upon.
+
+## Components
+### weaver
+
+`weaver` is the primary binary: it sets up the weave bridge to which containers are attached, manages IP address allocation, serves DNS requests and does many more things.
+
+Since version 2.0, `weaver` also bundles the Docker Plugin (legacy version) and Docker API Proxy which used to be shipped as separate containers.
+
+### weaveutil
+weaveutil is a binary which provides a number of functions for managing an existing weave network. It gets information about the network, attaches and detaches containers, etc.
+
+For the majority of interactions with an existing weave network, you will launch `weaveutil` in some manner or another.
+
+Almost all options can be passed as `--option` to `weaveutil`. These options are created in `prog/weaveutil/main.go` with a table mapping each command to a dedicated golang function.
+
+In addition to operating in normal mode, `weaveutil` has several additional operating modes. Before processing any commands, `weaveutil` checks the filename with which it was called.
+
+* If it was `weave-ipam` it delegates responsibility to the `cniIPAM()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's IPAM implementation from `plugin/ipam`.
+* If it was `weave-net` it delegates responsibility to the `cniNet()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's net plugin implementation from `plugin/net`.
+
+Finally, `weaveutil` is called by `weaver` when it needs to operate in a different network namespace.
+Go programs cannot safely switch namespaces; see this [issue](http://github.com/vishvananda/netns/issues/17).
+The call chain is `weaver`->`nsenter`->`weaveutil`.
+
+### weave
+Wrapping `weaver` and `weaveutil` is `weave`, a `sh` script that provides help information and calls `weaveutil` as relevant.
+
+### weave-kube
+weave-kube is an image with the weave binaries and a wrapper script installed. It is responsible for:
+
+1. Setting up the weave network
+2. Connecting to peers
+3. Copy the weave CNI-compatible binary plugin `weaveutil` to `/opt/cni/bin/weave-net` and weave config in `/etc/cni/net.d/10-weave.conf`
+4. Running the weave network policy controller (weave-npc) to implement kubernetes' `NetworkPolicy`
+
+Once installation is complete, each network-relevant change in a container leads `kubelet` to:
+
+1. Set certain environment variables to configure the CNI plugin, mostly related to the container ID
+2. Launch `/opt/cni/bin/weave-net`
+3. Pass the CNI config - the contents of `/etc/cni/net.d/10-weave.conf` to `weave-net`
+
+Thus, when each container is changed, and `kubelet` calls weave as a CNI plugin, it really just is launching `weaveutil` as `weave-net`.
+
+The installation and setup of all of the above - and therefore the entrypoint to the weave-kube image - is the script `prog/weave-kube/launch.sh`. `launch.sh` does the following:
+
+1. Read configuration variables from the environment. When the documentation for `weave-kube` describes configuring the weave network by changing the environment variables in the daemonset in the `.yml` file, `launch.sh` reads these environment variables.
+2. Set up the config file at `/etc/cni/net.d/10-weave.conf`
+3. Run the `weave` initialization script
+4. Run `weaver` with the correct configuration passed as command-line options
+
+
+## Adding Options
+To add new options to how weave should run with each invocation, you would do the following:
+
+1. determine to which `weave` command you want to add the option(s). `weave` normally is launched as `weave <command> <options> ...`.
+2. Add the option to `weave` script help for each `weave` command you wish to make it available. As of this writing, all of the commands and their options are listed in the function `usage_no_exist()`.
+3. Add the option to `weave` script option processing for the `weave` command, under the `case $COMMAND in` in the main `weave` script.
+4. In the function for the `weave` command, determine if the command should be passed on to `weaveutil` or `weaver`. Pass the option on to `weaveutil` or `weaver`, as appropriate, in the format `--option`, in the function for the `weave` command.
+5. Add a command-line option `--option` to `weaveutil` or `weaver` as appropriate.
+6. If the option can be configured for CNI:
+    * have `prog/weave-kube/launch.sh` read it as an environment variable and set inside
+    * set a default for the environment variable in `prog/weave-kube/weave-daemonset-k8s-1.6.yaml` and `weave-daemonset.yaml`
+    * if it should be set via `weaver` globally on its one-time initialization invocation, pass it on in `launch.sh`
+    * if it needs to be set via `weaveutil` on each invocation of `weave-net`, have `launch.sh` save it as an option in the CNI config file `/etc/cni/net.d/10-weave.conf` and then have the CNI code in `plugin/net/cni.go` in `weaveutil` read it and use where appropriate
+7. Document it!


docs/entrypoints.md
```diff
@@ -0,0 +1,56 @@
+# Entrypoints
+
+This document describes the various ways that weave as a binary gets launched, and where and how it processes options.
+
+The goal is to enable maintainers to modify functionality without having to reverse engineer the code to find all of the places you can start a weave binary and what it depends upon.
+
+## Components
+### weaveutil
```
bboreham: I consider `weaver` to be more primary :slightly_smiling_face: 

So I would insert:

### weaver
`weaver` is the primary binary: it sets up the `weave` bridge to which containers are attached; manages IP address allocation, serves DNS requests and does many more things.

Since version 2.0, `weaver` also bundles the Docker Plugin (legacy version) and Docker API Proxy which used to be shipped as separate containers.

docs/entrypoints.md
```diff
@@ -0,0 +1,56 @@
...
+
+Almost all options can be passed as `--option` to `weaveutil`. These options are created in `prog/weaveutil/main.go` with a table mapping each command to a dedicated golang function.
+
+In addition to operating in normal mode, `weaveutil` has two additional operating modes. Before processing any commands, `weaveutil` checks the filename with which it was called.
+
+* If it was `weave-ipam` it delegates responsibility to the `cniIPAM()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's IPAM implementation from `plugin/ipam`.
+* If it was `weave-net` it delegates responsibility to the `cniNet()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's net plugin implementation from `plugin/net`.
+
+### weave
+Wrapping `weaveutil` is `weave`, a `sh` script that provides help information and calls `weaveutil` as relevant.
```
bboreham: Wrapping `weaver` and `weaveutil`

docs/entrypoints.md
```diff
@@ -0,0 +1,56 @@
...
+
+For the majority of interactions with an existing weave network, you will launch `weaveutil` in some manner or another.
+
+Almost all options can be passed as `--option` to `weaveutil`. These options are created in `prog/weaveutil/main.go` with a table mapping each command to a dedicated golang function.
+
+In addition to operating in normal mode, `weaveutil` has two additional operating modes. Before processing any commands, `weaveutil` checks the filename with which it was called.
+
+* If it was `weave-ipam` it delegates responsibility to the `cniIPAM()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's IPAM implementation from `plugin/ipam`.
+* If it was `weave-net` it delegates responsibility to the `cniNet()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's net plugin implementation from `plugin/net`.
+
```
bboreham: `weaveutil` is also called by `weaver`, when it needs to operate in a different network namespace.
Go programs cannot safely switch namespace - see http://github.com/vishvananda/netns/issues/17
The call chain is `weaver`->`nsenter`->`weaveutil`.

docs/entrypoints.md
```diff
@@ -0,0 +1,56 @@
...
+2. Set up the config file at `/etc/cni/net.d/10-weave.conf`
+3. Run the `weave` initialization script
+4. Run `weaver` with the correct configuration passed as command-line options
+
+
+## Adding Options
+To add new options to how weave should run with each invocation, you would do the following:
+
+1. Add a command-line option `--option` to `weaveutil`
+2. Add the option to `weave` script help. You do not need to add it to how it calls `weaveutil` since `weave` just passes all options on.
```
bboreham: `weave launch` passes all options to `weaver`; beyond that it generally never passes options directly to `weaveutil`

docs/entrypoints.md
```diff
@@ -0,0 +1,56 @@
...
+3. If it can be configured for CNI:
+    * have `prog/weave-kube/launch.sh` read it as an environment variable and set inside `/etc/cni/net.d/10-weave.conf`
```
bboreham: I'm not sure where you were going with this part; maybe it relates to the previous misunderstanding.

docs/entrypoints.md
```diff
@@ -0,0 +1,56 @@
+# Entrypoints
+
+This document describes the various ways that weave as a binary gets launched, and where and how it processes options.
+
+The goal is to enable maintainers to modify functionality without having to reverse engineer the code to find all of the places you can start a weave binary and what it depends upon.
+
+## Components
+### weaveutil
+weaveutil is the primary binary for managing a existing weave network. It gets information about the network, attaches and detaches containers, etc.
```
bboreham: Suggest "`weaveutil` provides a number of functions for managing a existing weave network"

deitch:
Updated per almost all of your comments (and some grammatical corrections I found on the way).

docs/entrypoints.md
deitch: I don't understand the issue. Aren't CNI settings controlled by env vars in the `.yml` file, which the `launch.sh` sets and then either calls to `weaver` or sets for each invocation of `weave-net` by `10-weave.conf`?

docs/entrypoints.md
```diff
@@ -0,0 +1,67 @@
...
+The installation and setup of all of the above - and therefore the entrypoint to the weave-kube image - is the script `prog/weave-kube/launch.sh`. `launch.sh` does the following:
+
+1. Read configuration variables from the environment. When the documentation for `weave-kube` describes configuring the weave network by changing the environment variables in the daemonset in the `.yml` file, `launch.sh` reads these environment variables.
+2. Set up the config file at `/etc/cni/net.d/10-weave.conf`
+3. Run the `weave` initialization script
+4. Run `weaver` with the correct configuration passed as command-line options
+
+
+## Adding Options
+To add new options to how weave should run with each invocation, you would do the following:
```
bboreham: Since there are at least a dozen ways to invoke "weave", I'm unclear what is meant here.

docs/entrypoints.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. Add a command-line option `--option` to `weaveutil`
```
bboreham: `weaveutil` implements 20+ individual functions that all process arguments differently.  I have to suspect this is based on a misunderstanding.

docs/entrypoints.md
```diff
@@ -0,0 +1,67 @@
...
+2. Add the option to `weave` script help. You do not need to add it to how it calls `weaveutil` or `weaver`, since `weave` just passes all options on.
```
bboreham: `weave` passes all non-understood options to `weaver`; it only ever passes highly-crafted options to `weaveutil`.

docs/entrypoints.md
```diff
@@ -0,0 +1,67 @@
...
+3. If it can be configured for CNI:
+    * have `prog/weave-kube/launch.sh` read it as an environment variable and set inside
+    * set a default for the environment variable in `prog/weave-kube/weave-daemonset-k8s-1.6.yaml` and `weave-daemonset.yaml`
+    * if it should be set via `weaver` globally, pass it on in `launch.sh`
+    * if it should be set via `weaveutil` on each invocation of `weave-net`, have `launch.sh` save it as an option in `/etc/cni/net.d/10-weave.conf` and then have the CNI code read it from the config in `plugin/net/cni.go` and use where appropriate
```
bboreham: Something more like "if the option makes sense as a network-wide configuration, then it should go in the CNI config file and `weave-net` should read it from there".

deitch: Is there an alternative? in CNI mode, `weaveutil` for each container attach is invoked only by kubelet (or mesos). The only option we have to pass config is inside `/etc/cni/net.d/10-weave.conf`.

deitch: I made a small change to it to hopefully capture what you are saying.

docs/entrypoints.md
deitch: OK, we can make that a bit clearer. Commit coming.

deitch:
OK, I think this captures it better.

docs/entrypoints.md
```diff
@@ -0,0 +1,70 @@
...
+For the majority of interactions with an existing weave network, you will launch `weaveutil` in some manner or another.
+
+Almost all options can be passed as `--option` to `weaveutil`. These options are created in `prog/weaveutil/main.go` with a table mapping each command to a dedicated golang function.
+
+In addition to operating in normal mode, `weaveutil` has several additional operating modes. Before processing any commands, `weaveutil` checks the filename with which it was called.
+
+* If it was `weave-ipam` it delegates responsibility to the `cniIPAM()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's IPAM implementation from `plugin/ipam`.
+* If it was `weave-net` it delegates responsibility to the `cniNet()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's net plugin implementation from `plugin/net`.
+
+Finally, `weaveutil` is called by `weaver` when it needs to operatein a different network namespace.
```
marccarre: Typo: `operatein` -> `operate in`

docs/entrypoints.md
```diff
@@ -0,0 +1,70 @@
+# Entrypoints
+
+This document describes the various ways that weave as a binary gets launched, and where and how it processes options.
+
+The goal is to enable maintainers to modify functionality without having to reverse engineer the code to find all of the places you can start a weave binary and what it depends upon.
+
+## Components
+### weaver
+
+`weaver` is the primary binary: it sets up the weave bridge to which containers are attached; manages IP address allocation, serves DNS requests and does many more things.
```
marccarre: Minor: `are attached; manages` -> `are attached, manages`

deitch: Fixed.

docs/entrypoints.md
deitch: Fixed.

docs/entrypoints.md
```diff
@@ -0,0 +1,70 @@
...
+* If it was `weave-ipam` it delegates responsibility to the `cniIPAM()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's IPAM implementation from `plugin/ipam`.
+* If it was `weave-net` it delegates responsibility to the `cniNet()` function in `prog/weaveutil/cni.go`, which, in turn, calls the standard CNI plugin function `cni.PluginMain()`, passing it weave's net plugin implementation from `plugin/net`.
+
+Finally, `weaveutil` is called by `weaver` when it needs to operatein a different network namespace.
+Go programs cannot safely switch namespaces; see this [issue](http://github.com/vishvananda/netns/issues/17).
+The call chain is `weaver`->`nsenter`->`weaveutil`.
+
+### weave
+Wrapping `weaver` and `weaveutil` is `weave`, a `sh` script that provides help information and calls `weaveutil` as relevant.
+
```
marccarre: Suggestion: Add a comment about the history behind `weave` and `weaveutil`. Maybe something like the below?

> In versions of Weave Net prior to 2.0, `weave` used to have many utility functions implemented as simple shell functions, sometimes delegating to other Linux command line tools. 
> Current direction is to move some of this logic from `sh` to Go, for some or all of the below reasons: 
> - so it can re-use 3rd party libraries available in Go (e.g. to interact with Docker, Kubernetes, etc.),
> - so it can be re-used more conveniently in the various entry points, e.g.: `weave` and `launch.sh` (see below),
> - for better robustness and increased control over the flow of the program,
> - for better performance,
> - for better synchronisation and concurrency,
> - so it can be refactored and called directly from other components implemented in Go at a later point.
> 
> As a result, some of the utility functions originally implemented as `sh` functions in `weave` were moved to `weaveutil`.

Do you think this could be useful to new Weave Net developers @bboreham, @deitch?

docs/entrypoints.md
```diff
@@ -0,0 +1,70 @@
...
+The installation and setup of all of the above - and therefore the entrypoint to the weave-kube image - is the script `prog/weave-kube/launch.sh`. `launch.sh` does the following:
+
+1. Read configuration variables from the environment. When the documentation for `weave-kube` describes configuring the weave network by changing the environment variables in the daemonset in the `.yml` file, `launch.sh` reads these environment variables.
+2. Set up the config file at `/etc/cni/net.d/10-weave.conf`
+3. Run the `weave` initialization script
+4. Run `weaver` with the correct configuration passed as command-line options
+
+
+## Adding Options
```
marccarre: While this section may help someone right now, it is:
- paraphrasing the code a bit too much IMHO, 
- and therefore likely to go out of date quickly.

I'd suggest staying at a higher level or maybe to just move some of the information in point 6. to the `weave-kube` section.

marccarre:
Many thanks @deitch, such a technical documentation is very useful IMO, I wish I had something like this when I started on Weave Net!
Last section is a bit too detailed to stay relevant over time IMHO, so I'd suggest focusing on "what"/"where"/"why" in the documentation rather than "how", which would be explained by the code itself, but LGTM overall  :+1:

marccarre:
Could a [diagram](http://asciiflow.com/) ("_a picture is worth a thousand words_") be valuable in this documentation in order to highlight entry points and dependencies between components?

docs/entrypoints.md
deitch: Ha! Any and all of this would have been useful to me. :-) I wrote this up because I had to dig my way into the code and scripts to figure out how it all worked to get the hairpin enabled to deal with the CNI issue. I said to Bryan, "having a doc that describes how this works and the entrypoints would have made this much easier." Apparently, I volunteered myself. :-)

deitch: I am happy to add all of that into the doc. @bboreham ?

docs/entrypoints.md
deitch: While I am worried about it getting out of date, I still prefer leaving it if possible. The very reason I wanted to understand how "it all worked" was because I wanted to add options.

deitch:
> Could a diagram ("a picture is worth a thousand words") be valuable in this documentation in order to highlight entry points and dependencies between components?

Definitely. I am not convinced I get it enough to add it correctly. Still wrapping my head.

deitch:
OK, so what is next?

PR #2987, opened 2017-05-22
Use shfmt v1.3.0 instead of latest from master

marccarre:
Before this change, `go get`-ting` would pick the latest version of `shfmt` which keeps changing, hence frequently breaking our builds.

After this change, we download the `shfmt` binary from its GitHub releases page which is fast and locks the version we use. Note that `RUN` does not like herestrings, hence `echo "..." | sha256sum -c` for checksum.

Alternatives considered:
- `go get gopkg.in/mvdan/sh.v1/cmd/shfmt` -- but `gopkg.in` currently doesn't behave well for this tool
- `go build` from a `git` tag -- but this is much more involved.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -40,11 +40,15 @@ RUN apt-get update \
 		flex \
 		bison
 
+RUN curl -fsSLo shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt_v1.3.0_linux_amd64 && \
+	echo "b1925c2c405458811f0c227266402cf1868b4de529f114722c2e3a5af4ac7bb2  shfmt" | sha256sum -c && \
+	chmod +x shfmt && \
+	mv shfmt /usr/bin
+
 # Install common Go tools
 RUN go get \
 	github.com/golang/lint/golint \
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -40,11 +40,14 @@ RUN apt-get update \
 		flex \
 		bison
 
+RUN curl -fsSL -o shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt_v1.3.0_linux_amd64 && \
```
bboreham: Suggest checking this binary file is the one we expect similar to https://github.com/weaveworks/weave/blob/master/Makefile#L252

marccarre: Good point, done in latest version.

marccarre:
Same change has been LGTM'ed in `build-tools` and `cortex` already, so merging it in `weave` too in order to fix the builds ASAP.

marccarre:
Also merged in `master`:
```
commit 535bf9b06618fb7337ad8b1bd8ebe1084793413f
Merge: ceb7206 d4a98a5

    Merge remote-tracking branch 'origin/use-fixed-version-of-shfmt'
```

PR #2992, opened 2017-05-27
Remove unsafe netns change in proxy

bboreham:
This became unsafe when we started calling `AttachContainer` directly from the proxy.

We need two sub-functions which enter the container netns: one is called before we set the host end of the veth up, and one after.  Linux networking semantics require this split.

This leaves two instances of `WithNetNSUnsafe`: one from `weaveutil detach` and one when the CNI plugin configures routes.  In both cases the process exits straight away, so they are low risk.

### DIFF
--- a/net/veth.go
+++ b/net/veth.go
@@ -104,19 +104,12 @@ func interfaceExistsInNamespace(netNSPath string, ifName string) bool {
 	return err == nil
 }
 
-// NB: This function can be used only by a process that terminates immediately
-//     after calling the function as it changes netns via WithNetNSLinkUnsafe.
 func AttachContainer(netNSPath, id, ifName, bridgeName string, mtu int, withMulticastRoute bool, cidrs []*net.IPNet, keepTXOn bool, hairpinMode bool) error {
 	ns, err := netns.GetFromPath(netNSPath)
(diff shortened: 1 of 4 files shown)


brb:
The changes LGTM.

As discussed on Slack, this introduces a dependency on `nsenter` for CNI ADD, because the cmd calls `AttachContainer`. I see the following solutions:

* Replace `nsenter` with https://github.com/opencontainers/runc/tree/master/libcontainer/nsenter
* Make the CNI plugin to follow the client-server arch in which CNI ADD would call `weaver` to execute required steps.
* Just make `nsenter` as a requirement to run Weave Net CNI.

bboreham:
I tried the Runc libcontainer approach, but it is not very suited to this task.  It is focused on creating a new container which could have all different namespaces, paths, etc., from the parent. My attempt is preserved on the [runc-nsexec branch](https://github.com/weaveworks/weave/compare/runc-nsexec?expand=1).

I fell back to documenting the requirement.  Kubelet already requires it, so most of our audience is fine.

marccarre:
See also: https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix

Issue #2995, opened 2017-05-31
weave-npc blocks everything if kubelet --hostname_override flag used


bboreham:
As reported [on Slack](https://weave-community.slack.com/archives/C2ND76PAA/p1496237659127424)

we see in the `weave-npc` logs:

```
INFO: 2017/05/31 12:50:19.956363 Starting Weaveworks NPC 1.9.6; node name "kube01"
...
DEBU: 2017/05/31 12:50:20.318135 AddPod ignored for pod kube-system/weave-net-qbvhp on node 10.130.20.89
DEBU: 2017/05/31 12:50:20.318257 AddPod ignored for pod default/nginx-701339712-v3dkt on node 10.130.20.89
```

but these pods are actually local - `kubelet`'s args include `--hostname_override=10.130.20.89`

If we take the node name via the Kubernetes downward API it should match.

bboreham:
Fixed by #2996

PR #2996, opened 2017-05-31
Weave-NPC should get hostname from K8s Downward API.

Bregor:
It is necessary in case when kubelet's args includes `--hostname_override=`.
Without this change, pods on hostname overriding kubelets will never be local,
and all pod's traffic will be blocked.


Bregor:
Fixes https://github.com/weaveworks/weave/issues/2995

bboreham:
Thanks!

PR #3017, opened 2017-06-14
Update references to Weave Cloud

pidster:
Adds a message to `weave` script and updates the README to make it a bit more sane.

### DIFF
--- a/weave
+++ b/weave
@@ -96,7 +96,7 @@ where <peer>     = <ip_address_or_fqdn>[:<port>]
       <peer_id>  = <nickname> | <weave internal peer ID>
       <mode>     = consensus[=<count>] | seed=<mac>,... | observer
 
-To troubleshoot and visualize Weave networks, try Weave Cloud: http://www.weave.works/product/cloud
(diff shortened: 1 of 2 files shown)


weave
```diff
@@ -95,6 +95,8 @@ where <peer>     = <ip_address_or_fqdn>[:<port>]
       <endpoint> = [tcp://][<ip_address>]:<port> | [unix://]/path/to/socket
       <peer_id>  = <nickname> | <weave internal peer ID>
       <mode>     = consensus[=<count>] | seed=<mac>,... | observer
+
+To troubleshoot and visualize Weave networks, try Weave Cloud: http://www.weave.works/product/cloud/
```
bboreham: How about "To troubleshoot and visualize your cluster" - it does much more than just the Weave network part.

pidster: I was trying to make it relevant to Net

pidster: NB My laptop is b0rked so I won't be able to make any edits in the foreseeable.

bboreham: I think as "a word from our sponsor" it's fine to blab about vaguely-related things.
So would we explictly say something like "Weave Net is brought to you by the makers of ..." ?

PR #3018, opened 2017-06-15
modify tolerations to match any node

Shimi:
Currently weave doesn't run on nodes with tolerations except the master
The suggested code change it to run on all nodes

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
@@ -110,8 +110,8 @@ items:
               type: spc_t
           serviceAccountName: weave-net
           tolerations:
-            - effect: NoSchedule
-              key: node-role.kubernetes.io/master
+            - operator: "Exists"
+              effect: NoSchedule
           volumes:
             - name: weavedb
               hostPath:


bboreham:
@Shimi the current setting matches the taint used by kops and kubeadm; may I ask which installation you are using that has a different taint?

Shimi:
@bboreham I actually use kops. You can see kops taint support [here](https://github.com/kubernetes/kops/blob/master/docs/instance_groups.md#adding-taints-to-an-instance-group).

I have my own taints in addition to the master taints.  Without this setting, weave doesn't install on nodes with taints. In this case there is no networking support for these nodes and no other pods can run. 
taints can be used for many different use cases.
* dedicated nodes
* special hardware nodes
* spot  nodes
and more

bboreham:
@shimi thanks for explaining.  

This sounds like the right way to go.  Currently someone wanting to use their own taints for any other purpose has to have a modified copy of the Weave Net daemonset spec; if we make this change then only those people who want to control Weave Net itself with taints need to modify the daemonset.

Shimi:
@bboreham Is there a use case were weave shouldn't be install on all nodes?
As I see, it is a crucial component that should run on all nodes.

bboreham:
@Shimi it wouldn't be a common thing, but I could imagine some. For instance if you had some pods that ran in the host namespace so didn't need a pod network, and you wanted to dedicate them to nodes that ran nothing else.  

Or maybe in future if Kubernetes supported more than one pod network then you could have some nodes with Weave Net and some with a different product.

Anyway I'm agreeing with your change.

PR #3019, opened 2017-06-17
Ability to set IPALLOC_RANGE via plugin-v2 configuration parameters

unitymind:
Unfortunately, we can not set `--ipalloc-range` using `EXTRA_ARGS` plugin configuration parameter. It's just not take effect in `prog/net-plugin/launch.sh` and there is no way to change `--ipalloc-range` for plugin.

This PR simply add `IPALLOC_RANGE` as plugin configuration parameter and update documentation accordingly.

### DIFF
--- a/prog/net-plugin/config.json
+++ b/prog/net-plugin/config.json
@@ -74,6 +74,14 @@
           "value"
       ],
       "value": ""
+    },
+    {
+      "description": "The range of IP addresses used by Weave Net",
+      "name": "IPALLOC_RANGE",
+      "settable": [
+          "value"
+      ],
+      "value": ""
     }
   ],
   "network": {
--- a/site/plugin-v2.md
+++ b/site/plugin-v2.md
@@ -43,6 +43,9 @@ The parameters include:
   a larger size for better performance if your network supports jumbo
   frames - see [here](/site/using-weave/fastdp.md#mtu) for more
   details.
+* `IPALLOC_RANGE` - the range of IP addresses used by Weave Net and the subnet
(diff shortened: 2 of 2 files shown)


bboreham:
Thanks for the contribution, @unitymind

Can you say a bit more about how you use the weave net plugin?  We had guessed anyone using the v2 plugin would use Docker's IP allocator.

unitymind:
@bboreham I have, may be, strange case: still use Docker overlay network with swarm scope, but use Weave for containers which outside of swarm scope (system/control services) - for simple encrypted private multi-host networking... And would be great have ability mix swarm scoped containers with "system" containers via shared Weave Network. Weavemesh driver do not allowed in service definition for `docker stack deploy`, but we can use `docker network connect` after start - probably unwanted semi-automated steps, but how handle updates when new containers is come? In common words: swarm overlay network - data plane between services, but Weave - as control plane.

And yes, I am trap into subnet overlap because 10.42.0.x already in use between hosts (don't ask me why - brittle Vagrant/Virtualbox environment). But anyway, more configurable parameters for plugin - it's good :) Kubernetes integration is already more customizable.

bboreham:
Thanks for the contribution @unitymind !

unitymind:
@bboreham Thanks for accepting PR. Congrats with new version 2.0 release! But when updated plugin will be available? :)

bboreham:
It has to go through Docker's review and update process - somewhere around a week maybe.

unitymind:
@bboreham Got it, thanks for clarification!

PR #3022, opened 2017-06-19
WIP: remove peers that have disappeared from kubernetes

### LINKED ISSUE #2797: Remove deleted k8s nodes from Weave Net
`weave-kube` adds all current nodes as peers at startup, but never checks back to see if some nodes have been deleted.

In a situation such as a regularly expanding and contracting auto-scale group, the IPAM ring will eventually become clogged with peers that have gone away.

We need to do `weave rmpeer` on deleted nodes, and less importantly `weave forget`.  We will need some interlock to ensure the `weave rmpeer` is only done once.

How do we even detect that a Weave Net peer originated as a Kubernetes node?  This logic should be resilient to users adding non-Kubernetes peers to the network, even on a host that was previously a Kubernetes peer.

bboreham:
Fixes #2797 

Using Kubernetes annotations as a way to ensure only one peer at a time runs `rmpeer`

### DIFF
--- a/prog/kube-peers/main.go
+++ b/prog/kube-peers/main.go
@@ -56,23 +56,25 @@ const (
 // update the list of all peers that have gone through this code path
 func addMyselfToPeerList(c *kubernetes.Clientset, peerName, name string) (*peerList, error) {
 	cml := newConfigMapAnnotations(configMapNamespace, configMapName, c)
-	if err := cml.Init(); err != nil {
-		return nil, err
-	}
-	list, err := cml.GetPeerList()
-	if err != nil {
-		return nil, err
-	}
-	log.Println("Fetched existing peer list", list)
-	if !list.contains(peerName) {
-		list.add(peerName, name)
-		log.Println("Storing new peer list", list)
-		err = cml.UpdatePeerList(*list)
+	var list *peerList
+	err := cml.LoopUpdate(func() error {
+		var err error
+		list, err = cml.GetPeerList()
 		if err != nil {
-			return nil, err
+			return err
 		}
-	}
-	return list, nil
+		log.Println("Fetched existing peer list", list)
+		if !list.contains(peerName) {
+			list.add(peerName, name)
+			log.Println("Storing new peer list", list)
+			err = cml.UpdatePeerList(*list)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+	return list, err
 }
 
 // For each of those peers that is no longer listed as a node by
--- a/prog/kube-peers/main.go
+++ b/prog/kube-peers/main.go
@@ -10,15 +10,7 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-func getKubePeers() ([]string, error) {
-	config, err := rest.InClusterConfig()
-	if err != nil {
-		return nil, err
-	}
-	c, err := kubernetes.NewForConfig(config)
-	if err != nil {
-		return nil, err
-	}
+func getKubePeers(c *kubernetes.Clientset) ([]string, error) {
 	nodeList, err := c.Nodes().List(api.ListOptions{})
 	if err != nil {
 		return nil, err
@@ -50,7 +42,15 @@ func getKubePeers() ([]string, error) {
 }
 
 func main() {
-	peers, err := getKubePeers()
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		log.Fatalf("Could not get cluster config: %v", err)
+	}
+	c, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatalf("Could not make Kubernetes connection: %v", err)
+	}
+	peers, err := getKubePeers(c)
 	if err != nil {
 		log.Fatalf("Could not get peers: %v", err)
 	}
--- a/prog/kube-peers/main.go
+++ b/prog/kube-peers/main.go
@@ -21,17 +21,6 @@ func getKubePeers() ([]string, error) {
 	}
 	nodeList, err := c.Nodes().List(api.ListOptions{})
 	if err != nil {
-		// Fallback for cases (e.g. from kube-up.sh) where kube-proxy is not running on master
-		config.Host = "http://localhost:8080"
-		log.Print("error contacting APIServer: ", err, "; trying with fallback: ", config.Host)
-		c, err = kubernetes.NewForConfig(config)
-		if err != nil {
-			return nil, err
-		}
-		nodeList, err = c.Nodes().List(api.ListOptions{})
-	}
-
-	if err != nil {
 		return nil, err
 	}
 	addresses := make([]string, 0, len(nodeList.Items))
--- a/prog/kube-peers/main.go
+++ b/prog/kube-peers/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"log"
 	"net"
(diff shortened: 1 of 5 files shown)


prog/kube-peers/main.go
```diff
@@ -52,20 +40,109 @@ func getKubePeers() ([]string, error) {
...
+	// 12.      Then return to step 3 to claim X again
+
+	// Step 3-5 is to protect against two simultaneous rmpeers of X
+	// Step 4 is to pick up again after a restart between step 5 and step 7b
+	// Step 10 is to clean up if the peer doing the reclaim disappears between steps 5 and 7a
+	// If peer doing the reclaim disappears forever between 7a and 7b then we get a dangling annotation
+	// This should be sufficiently rare that we don't care.
+
+	// Question: Should we check against Weave Net IPAM?
+	// i.e. If peer X owns any address space and is marked unreachable, we want to rmpeer X
```
mikebryant: I think no:
On a temporary network-based node failure (e.g. bare metal, localised power failure for the relevant swith for example), we can get into a situation where:
-  peer X owns address space
- peer X is unreachable

We don't want to rmpeer it, as it can come back and still be expecting to connect back to the network. It may even still have pods running on those IPs

prog/kube-peers/main.go
```diff
@@ -52,20 +40,109 @@ func getKubePeers() ([]string, error) {
...
+	})
+	return list, err
+}
+
+// For each of those peers that is no longer listed as a node by
+// Kubernetes, remove it from Weave IPAM
+func reclaimRemovedPeers(apl *peerList, nodes []nodeInfo) error {
+	// TODO
+	// Outline of function:
+	// 1. Compare peers stored in the peerList against all peers reported by k8s now.
```
mikebryant: Be wary of kubernetes/kubernetes#46442

In several environments removed nodes may come back.
If weavedb persistence is enabled, this will mean those nodes can't join the cluster properly, as they will believe they own address space and the cluster will disagree.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -46,6 +46,44 @@ items:
       - kind: ServiceAccount
         name: weave-net
         namespace: kube-system
+  - apiVersion: rbac.authorization.k8s.io/v1beta1
+    kind: Role
+    metadata:
+      name: weave-net2
```
mikebryant: nit: Is there a reason the permissions aren't just added to the existing role?
If they should be separate, a more descriptive name would be nice (`weave-net-peers`?)

prog/kube-peers/main.go
```diff
@@ -52,20 +40,109 @@ func getKubePeers() ([]string, error) {
...
+	// 3. Check if there is an existing annotation with key X
+	// 4.   If annotation already contains my identity, ok;
+	// 5.   If non-existent, write an annotation with key X and contents "my identity"
+	// 6.   If step 4 or 5 succeeded, rmpeer X
+	// 7a.    Remove X from peerList
+	// 7b.    Remove annotation with key X
+	// 8.   If step 5 failed due to optimistic lock conflict, stop: someone else is handling X
+	// 9.   Else there is an existing annotation at step 3; call its owner peer Y.
+	// 10.    If Y is not known to k8s and marked unreachable, recurse to remove Y at 3
+	// 11.      If we succeed to claim the annotation for Y, remove its annotation for X too
```
mikebryant: On ordering, should it then be:
```
	// 7a:    Remove any annotations Z* that have contents X
	// 7b.    Remove X from peerList
	// 7c.    Remove annotation with key X
```

Instead of recursing, could change the algorithm such that for the loop around step 2, instead of going through each peer once, instead keep looping until we get the whole way through the list without doing any work
e.g. all peers that we want to remove are in step 8

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
bboreham: The existing one is a ClusterRole; this is a Role to narrow down the write permissions.

I agree with your point about naming; it was late...

prog/kube-peers/main.go
bboreham: Gah, that kills the whole idea.

We need to distinguish between "stopped but coming back" and "stopped and never coming back".

Possibly we could extend the use of the `peerList` annotation to say "If I'm not currently listed as a peer then delete any old IPAM persistence data"?

prog/kube-peers/main.go
bboreham: That question was meant as a narrowing of step 6, i.e. after Kubernetes has decided it's gone, should we additionally see if Weave Net thinks it's still there.  Not a widening.

prog/kube-peers/main.go
mikebryant: That would work for nodes that have been restarted. (Which is probably the most common case)

One that was just off the network for a while though.. I guess it depends on the exact failure mode in the cloud provider. Not sure what would happen if there were still running pods

prog/kube-peers/main.go
mikebryant: Oh, I see. I don't see that would harm anything then. 
It might help alleviate some of the issues in my comment about 46442, if kubernetes thinks it's gone but weave can still talk to it

prog/kube-peers/main.go
mikebryant: For info:
Our hacky rmpeers solution was hitting this quite a bit. So this is our current fix:
```
      initContainers:
      - name: reset-weavedb
        image: hub.docker.tech.lastmile.com/ocadotechnology/weave-net-reset-weavedb:0acd87fd3f7bde7d4cfe8b5009e0fafa2d0e9c18@sha256:05722937400c0e0449a80e9af8285323ac9ca7036c7f009e6c3c1e46d4bae855
        env:
        - name: HOSTNAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        volumeMounts:
        - name: weavedb
          mountPath: /weavedb
```

The source of which is:
```
#!/bin/sh

set -eux

CURRENT_NODE_UUID=$(kubectl get node -o custom-columns=ID:.metadata.uid --no-headers ${HOSTNAME})

if [ ! -f /weavedb/NODE_UUID ]; then
  # Does not exist, so create it with the Node's current UUID
  echo -n ${CURRENT_NODE_UUID} > /weavedb/NODE_UUID
  exit 0;
fi;

if [ "${CURRENT_NODE_UUID}" == "$(cat /weavedb/NODE_UUID)" ]; then
  # It's still the same, so we're not likely to be out of sync
  exit 0;
fi;

# If we're here, it didn't match, so reset the persistent data on the assumption
# that we've been rmpeer'd
rm -f /weavedb/weavedata.db
rm -f /weavedb/weave-netdata.db

# And also set the NODE_UUID
echo -n ${CURRENT_NODE_UUID} > /weavedb/NODE_UUID

exit 0
```

We're basing this of the assumption that if our Node object has gone away and come back with a new UUID, our IPs may have been hoovered up by our rmpeers script, so we'll start over.

bboreham: > One that was just off the network for a while though.. I guess it depends on the exact failure mode in the cloud provider. Not sure what would happen if there were still running pods

Suppose `rmpeer X` is done while X is alive but disconnected, then X will be unable to reconnect to any peer which has received that update.  The pods on X will be attached to a bridge but the bridge will be isolated from any other node.

However, I would expect the scheduler to evict pods on a node which is deleted from etcd, hence after rejoin kubelet would delete all those existing pods and the weave-kube pod would go through its restart cycle and everything should be OK after that.

prog/kube-peers/main.go
bboreham: I updated in line with this suggestion.

bboreham:
I have squashed some of the commits and written an implementation of the pseudo-code previously discussed.

Next steps: try it out, figure out how to test it properly.

bboreham:
Also still to do, as noted [earlier](https://github.com/weaveworks/weave/pull/3022#discussion_r122706899):

We need to distinguish between "stopped but coming back" and "stopped and never coming back".

Possibly we could extend the use of the peerList annotation to say "If I'm not currently listed as a peer then delete any old IPAM persistence data"?

caarlos0:
> We need to distinguish between "stopped but coming back" and "stopped and never coming back".

I'm not very familiar with weave, but, why? If a node disappears and we release its IPs and etc, and if it come back later, can't we just "re-setup" it (with new IPs and all)?

bboreham:
Weave Net has no central point of control: its native data structure is a CRDT. The "release its IPs" breaks the rules of the CRDT so we need some extra mechanism to avoid going mad.

caarlos0:
hmm, interesting. Thanks for explaining @bboreham  :)

bboreham:
Replaced by #3149

PR #3027, opened 2017-06-22
NPC: Track non-local pods for source rules; ignore for dst rules

### LINKED ISSUE #3025: Weave verison 1.9.8 does not allow for ingress pod selection in network policy application

Network policies would work (like in version 1.9.5), but in version 1.9.8, when we use an ingress filter on network policies to select a pod (the "from" pod), traffic is blocked.

## What happened?
Traffic is blocked, the pod filter is not applied

## How to reproduce it?
Setup a namespace with ingress turned off.   Set NS to whatever namespace you are interested in

kubectl annotate ns "${NS}" net.beta.kubernetes.io/network-policy='{"ingress":{"isolation":"DefaultDeny"}}'

This will block all inbound traffic.  Now, apply a network policy to allow traffic to a pod in this namespace.  This network policy will use a pod filter to only allow inbound traffic from pods that have certain labels.

Example policy from our example, just change the labels to what you want:

```
apiVersion: extensions/v1beta1
kind: NetworkPolicy
metadata:
  name: job-exporter
  namespace: monitoring
spec:
  ingress:
  - from:
    - podSelector:
        matchExpressions:
        - key: app
          operator: In
          values:
          - from-pod
    ports:
    - port: 8080
      protocol: TCP
  podSelector:
    matchLabels:
      app: my-pod

```

Apply the policy.  You should see traffic being allowed from "from-pod" to "my-pod" on port 8080, TCP.  This does not work though.

Kubernetes is running in AWS.

## Versions:

```kubectl version
Client Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.1", GitCommit:"b0b7a323cc5a4a2019b2e9520c21c7830b7f708e", GitTreeState:"clean", BuildDate:"2017-04-03T23:37:53Z", GoVersion:"go1.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"6", GitVersion:"v1.6.5", GitCommit:"490c6f13df1cb6612e0993c4c14f2ff90f8cdbf3", GitTreeState:"clean", BuildDate:"2017-06-14T20:03:38Z", GoVersion:"go1.7.6", Compiler:"gc", Platform:"linux/amd64"}
```

bboreham:
Fixes #3025 

Previously each namespace had only one set of pod-selectors, into which the destination selector per-policy and all source selectors were placed.  This PR extracts those into two sets, so we can do different things for `src` and `dst` when pod updates arrive.

I also extended test 840 to add a network policy which would catch the problem; see it fail at https://circleci.com/gh/weaveworks/weave/9262

### DIFF
--- a/npc/analyser.go
+++ b/npc/analyser.go
@@ -12,18 +12,18 @@ import (
 
 func (ns *ns) analysePolicy(policy *extnapi.NetworkPolicy) (
 	rules map[string]*ruleSpec,
-	nsSelectors, podSelectors map[string]*selectorSpec,
+	nsSelectors, srcPodSelectors map[string]*selectorSpec,
+	dstSelector *selectorSpec,
 	err error) {
 
 	nsSelectors = make(map[string]*selectorSpec)
-	podSelectors = make(map[string]*selectorSpec)
+	srcPodSelectors = make(map[string]*selectorSpec)
 	rules = make(map[string]*ruleSpec)
 
-	dstSelector, err := newSelectorSpec(&policy.Spec.PodSelector, ns.name, ipset.HashIP)
+	dstSelector, err = newSelectorSpec(&policy.Spec.PodSelector, ns.name, ipset.HashIP)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, nil, err
 	}
-	podSelectors[dstSelector.key] = dstSelector
 
 	for _, ingressRule := range policy.Spec.Ingress {
 		if ingressRule.Ports != nil && len(ingressRule.Ports) == 0 {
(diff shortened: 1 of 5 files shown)


npc/analyser.go
```diff
@@ -12,18 +12,18 @@ import (
 
 func (ns *ns) analysePolicy(policy *extnapi.NetworkPolicy) (
 	rules map[string]*ruleSpec,
-	nsSelectors, podSelectors map[string]*selectorSpec,
+	nsSelectors, srcPodSelectors map[string]*selectorSpec,
+	dstSelector *selectorSpec,
```
rade: hmm. time for a struct perhaps, containing these four?

test/images/network-tester/Dockerfile
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 FROM bash
+RUN apk add --update \
```
marccarre: Minor: `RUN apk add --no-cache curl` instead.
> As of Alpine Linux 3.3 there exists a new `--no-cache` option for `apk`. It allows users to install packages with an index that is updated and used on-the-fly and not cached locally: [...]
> This avoids the need to use `--update` and remove `/var/cache/apk/*` when done installing packages.

This should work given the `bash` image is on Alpine 3.6:
https://github.com/tianon/docker-bash/blob/75358344e900aa56e64722b2241a6bcd39ca67ee/4.4/Dockerfile#L1

bboreham:
There are at least three flaws with this PR as it stands:
* We need to have different names for the same selector appearing as a source or a destination set; currently the selector rule is hashed to an ipset name but the set needs to have different contents depending on whether we are ignoring local pods or not.
* The `allPods` selector for each namespace also needs to be divided into a source and destination form.  Maintaining the source side of these then requires that we look at every pod, which negates much of the point of filtering updates to local pods.  It would be better if the `allPods` set was created lazily on demand.
* The `LocalIpSet` used to avoid blocking connections on the originating side is now collecting all pod addresses because the code path used to return before it got there.

PR #3028, opened 2017-06-22
weave-npc: stop ignoring local pods

### LINKED ISSUE #3025: Weave verison 1.9.8 does not allow for ingress pod selection in network policy application

bboreham:
This is another take on #3027 
Fixes #3025 

Here we wind back to how weave-npc worked as at 1.9.5, but still track an ipset of local pod addresses so we can stop checking when we know the destination is not a local pod (because we only want to check ingress rules)

I also extended test 840 to add a network policy which would catch the problem; see it fail at https://circleci.com/gh/weaveworks/weave/9262

### DIFF
--- a/test/images/network-tester/Dockerfile
+++ b/test/images/network-tester/Dockerfile
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 FROM bash
+RUN apk add --no-cache curl
 ADD webserver webserver
 EXPOSE 8080
 ENTRYPOINT ["/webserver"]
--- a/npc/controller.go
+++ b/npc/controller.go
@@ -66,7 +66,7 @@ func (npc *controller) onNewNsSelector(selector *selector) error {
 func (npc *controller) withNS(name string, f func(ns *ns) error) error {
 	ns, found := npc.nss[name]
 	if !found {
-		newNs, err := newNS(name, npc.ipt, npc.ips, npc.nsSelectors)
+		newNs, err := newNS(name, npc.nodeName, npc.ipt, npc.ips, npc.nsSelectors)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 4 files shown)


test/images/network-tester/Dockerfile
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 FROM bash
+RUN apk add --update \
```
marccarre: Minor: `RUN apk add --no-cache curl` instead.
> As of Alpine Linux 3.3 there exists a new `--no-cache` option for `apk`. It allows users to install packages with an index that is updated and used on-the-fly and not cached locally: [...]
> This avoids the need to use `--update` and remove `/var/cache/apk/*` when done installing packages.

This should work given the `bash` image is on Alpine 3.6:
https://github.com/tianon/docker-bash/blob/75358344e900aa56e64722b2241a6bcd39ca67ee/4.4/Dockerfile#L1

bboreham: Thanks; fixed

PR #3031, opened 2017-06-23
Set the default namespace to weave-daemonset-k8s

arno01:
Since the ``namespace: kube-system`` was removed in https://github.com/weaveworks/weave/commit/067ce6de9f3800e06369140a832c12f74dcefb90#diff-63ba20698025213b0dce5f67c19f76db  the current master fresh deployment will fail (Tried k8s 1.6.5 with RBAC):

```
kubectl logs pods/weave-net-x6gj4 -c weave
2017/06/23 08:12:14 error contacting APIServer: the server does not allow access to the requested resource (get nodes); trying with fallback: http://localhost:8080
2017/06/23 08:12:14 Could not get peers: Get http://localhost:8080/api/v1/nodes: dial tcp 127.0.0.1:8080: getsockopt: connection refused
Failed to get peers
```

The following resources should be under the same namespace for the RBAC to properly function:
- ServiceAccount
- ClusterRoleBinding
- DaemonSet

I am wondering whether the ``weave-net`` common namespace would be more suitable than the ``default``?

**Upd:** I have noticed that the same is valid for the ``https://cloud.weave.works/k8s/v1.6/scope.yaml?k8s-service-type=NodePort`` but could not find this file in https://github.com/weaveworks/scope repo...

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
@@ -45,7 +45,7 @@ items:
     subjects:
       - kind: ServiceAccount
         name: weave-net
-        namespace: kube-system
+        namespace: default
   - apiVersion: extensions/v1beta1
     kind: DaemonSet
     metadata:


bboreham:
Thanks for this, @arno01, but we changed [the instructions](https://github.com/weaveworks/weave/blob/master/site/kube-addon.md) to add `-n kube-system` to the command-line.  This was done in the name of consistency.

I'm not entirely sure it was a good step because there's more to go wrong, but we do not intend for weave-kube to run in the default namespace so will not be merging this.

arno01:
@bboreham and what do you think about running the weave-net in the weave-net namespace and the weave-scope for the weave-scope? I think that would make sense and be nice, IMO. :-)

bboreham:
Yes, I would be sympathetic to that idea, except we should really run in ['critical' mode](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/) which requires being in the `kube-system` namespace.

The pod network is an essential component in Kubernetes.

PR #3043, opened 2017-06-27
Fix fallback to pcap based bridge when odp is not supported

brb:
The fix ensures that if odp is not supported (usually due to missing `openvswitch`, `vxlan` or `vport-vxlan` kernel modules), weaver will create the pcap based bridge instead of crashing.

### DIFF
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -62,7 +62,8 @@ type Bridge interface {
 	String() string                  // human-readable type string
 }
 
-var errBridgeFallback = errors.New("special error value used to trigger a fallback to bridge type")
+// Used to indicate a fallback to the Bridge type
+var errBridgeNotSupported = errors.New("bridge not supported")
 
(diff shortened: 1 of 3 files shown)


brb:
Tested manually with:

* `mv /lib/modules/$(uname -r)/build/kernel/net/openvswitch/{,backup}openvswitch.ko.gz`
* `mv /lib/modules/$(uname -r)/build/kernel/net/openvswitch/{,backup}vport-vxlan.ko.gz`

before starting Weave Net.

The warnings are for each test case are the following:

* `WARN: 2017/06/27 16:32:09.594113 Skipping bridge creation of "bridged_fastdp" due to: : bridge not supported`.
* `WARN: 2017/06/27 16:29:17.283111 Skipping bridge creation of "bridged_fastdp" due to: kernel does not have Open vSwitch VXLAN support: bridge not supported`.

Issue #3052, opened 2017-07-08
Q: in weave as k8s add-on, what is the equivalent of weave cli commands?


aalba6675:
When using weave as an add-on to kubernetes, how do you inspect the state of weave, or what is the equivalent of the weave CLI commands like `weave ipam status`?

I see that every node has a weave-net-* pod running, obviously, there are no node (systemd)services or binaries installed.

In this situation, what is the recommend best practice to manage weave when most of it is hidden in the pods?

If I do something like running weave from inside the pod, it doesn't work.
```
kubectl exec weave-net-0lhn5 -n kube-system -- /home/weave/weave status ipam
Defaulting container name to weave.
Use 'kubectl describe pod/weave-net-0lhn5' to see all of the containers in this pod.
/home/weave/weave: line 250: docker: not found
ERROR: Unable to parse docker version
```

…though kube-peers works
```
kubectl exec weave-net-0lhn5 -n kube-system -- /home/weave/kube-peers
Defaulting container name to weave.
Use 'kubectl describe pod/weave-net-0lhn5' to see all of the containers in this pod.
192.168.1.10
192.168.1.11
192.168.1.12
192.168.1.13
```

bboreham:
@aalba6675 you need to add `--local` to tell it not to go via Docker - see #2603 

Could you expand a bit on what kinds of things you want to do "to manage weave"?

#2603 covers the `status` command specifically.
#2612 is about troubleshooting.
#2797 is about removing deleted nodes.

When I try to think of a simple way to present this, a lot of complexities come to mind - e.g. if you have 10 nodes, do you want to see status from all of them?  Is it useful to pick one at random?

So, I'd be really interested if you could walk through some scenarios, help to generate some ideas about how this could work.

aalba6675:
@bboreham  I'm groping here, but generally for SDN add-ons (not just weave mind you) there is a disconnect between using it as add-ons and externally configured SDN with regards to documentation. Once it used as an add-on, it seems much more mysterious and opaque — don't know if that makes much sense to you — but coming from the angle of operator training and understanding viz a viz an external SDN.

Weave recommends using the add-on for k8s, but what if I want to have a "standard" weave for external services or just non-k8s workloads: in one use case, we want to have the k8s etcd cluster as a set container "pets" but also trying not to introduce another SDN.  Do you recommend separate weave enclaves and figure out routing some how, rather than "one-SDN-to-rule-them-all"?

[...]

PR #3055, opened 2017-07-10
Add operations section to weave-kube's docs.

marccarre:
Document examples of `weave` commands against/within a Kubernetes cluster.

This hopefully addresses part of #3052.
Documentation of best practices, examples of hybrid clusters, etc. are not included in this PR.

### DIFF
--- a/site/kube-addon.md
+++ b/site/kube-addon.md
@@ -292,3 +292,72 @@ For example,
             - name: IPALLOC_RANGE
               value: 10.0.0.0/16
 ```
+
+## <a name="operations"></a> Operations
+
+When using the Kubernetes addon, Weave Net's peers are managed via Kubernetes and very little effort should be required from your end to operate your network.
+
+However, should you need it, you can still run `weave` CLI commands to, for example, confirm Weave Net' status. This can be done in various ways:
+
+1. You could [install the `weave` script](/site/installing-weave.md) and run the same commands you would run without Kubernetes:
+
+```
+$ weave status
+        Version: 2.0.1 (up to date; next check at 2017/07/10 13:49:29)
+
+        Service: router
+       Protocol: weave 1..2
+           Name: 42:8e:e8:c4:52:1b(host-0)
+     Encryption: disabled
+  PeerDiscovery: enabled
+        Targets: 3
+    Connections: 3 (2 established, 1 failed)
+          Peers: 3 (with 6 established connections)
+ TrustedSubnets: none
+
+        Service: ipam
+         Status: ready
+          Range: 10.32.0.0/12
+  DefaultSubnet: 10.32.0.0/12
+```
+
+2. If you do not want to install anything else on your hosts, you can use the following `kubectl` commands, which will produce the exact same outcome as the previous example:
+
+```
+### Identify the Weave Net pods:
+
+$ kubectl get pods --all-namespaces -o wide | grep weave
+kube-system  weave-net-1jkl6  2/2  Running  0  1d  10.128.0.4  host-0
+kube-system  weave-net-bskbv  2/2  Running  0  1d  10.128.0.5  host-1
+kube-system  weave-net-m4x1b  2/2  Running  0  1d  10.128.0.6  host-2
+
+### The above shows all Weave Net pods available in your cluster.
+### You can see Kubernetes has deployed one Weave Net pod per host, in order to interconnect all hosts.
+
+### You then need to:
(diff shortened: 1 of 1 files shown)


site/kube-addon.md
```diff
@@ -292,3 +292,65 @@ For example,
             - name: IPALLOC_RANGE
               value: 10.0.0.0/16
 ```
+
+## <a name="operations"></a> Operations
+
+Weave Net can be used without Kubernetes, and if you have done this before, you may be familiar with the `weave` CLI.
```
bboreham: This sentence doesn't quite read right to me. 
Maybe just say what's possible without speculating on the users path to get here?

site/kube-addon.md
```diff
@@ -292,3 +292,65 @@ For example,
...
+
+Within Kubernetes, even though the network peers are managed via Kubernetes and very few commands from the `weave` CLI should be useful, you may still want to run commands to, for example, confirm Weave Net' status. This can be done in various ways:
+
+1. You could [install the `weave` script](/site/installing-weave.md) and run the same commands you would run without Kubernetes:
```
bboreham: Maybe stress not to run `weave launch`. 
Or does it detect "weave is already running"?

site/kube-addon.md
```diff
@@ -292,3 +292,65 @@ For example,
...
+2. If you do not want to install anything else on your hosts, you can use the following `kubectl` commands, which will produce the same outcome as the previous example:
+
+```
+### Identify the Weave Net pods:
+$ kubectl get pods -n kube-system | grep weave-net
+weave-net-1jkl6                  2/2       Running   0          1d
+weave-net-bskbv                  2/2       Running   0          1d
+weave-net-m4x1b                  2/2       Running   0          1d
+
+### Run the commands you want from within one of the pods:
```
bboreham: Say something about picking a specific pod?

site/kube-addon.md
```diff
@@ -292,3 +292,65 @@ For example,
...
+          Peers: 3 (with 6 established connections)
+ TrustedSubnets: none
+
+        Service: ipam
+         Status: ready
+          Range: 10.32.0.0/12
+  DefaultSubnet: 10.32.0.0/12
+```
+
+3. Finally you could also use [Weave Cloud](https://cloud.weave.works/) and monitor all your pods, including Weave Net's ones, from there.
```
bboreham: Would be interesting to see whether Explore shows the Weave Net topology in this configuration. I think it will.

site/kube-addon.md
marccarre: Yes, it does:
```
$ weave launch
Weave Net is already running
```

site/kube-addon.md
marccarre: I thought _one of the pods_ was clear enough but I'll add something about having one pod per host, choosing the right one, etc. then; thanks!

site/kube-addon.md
marccarre: Sure, will change :+1: Thanks!

site/kube-addon.md
marccarre: 1. The "_Pods_" view currently groups all Net pods as one `DaemonSet`.
Clicking on one of the pods, then `weave` container, then "_Exec shell_", `/home/weave/weave --local status` definitely works as expected. 
Maybe I should add some details about this to the doc?
Alternatively, do we have some Weave Cloud doc I could point to?

2. The "_Hosts > Weave Net_" view shows all hosts and their topology (i.e. in my case, the fact they are all interconnected), which is more valuable IMO.
Clicking on one of the hosts shows details typically available via `weave status`:

        INFO:
        VERSION: 2.0.1
        PROTOCOL: 1..2
        NAME: da:76:99:a4:e0:91
        ENCRYPTION: disabled
        PEER DISCOVERY: enabled
        TARGETS: 3
        PEERS: 3
        CONNECTIONS: 3
        TRUSTED SUBNETS:
        
        CONNECTIONS | STATE | INFO
        -> 10.128.0.6:6783 | failed | cannot connect to ourself, retry: never
        -> 10.128.0.5:6783 | established | fastdp aa:14:2a:5b:e2:f3(host-2)
        -> 10.128.0.4:6783 | established | fastdp 42:8e:e8:c4:52:1b(host-0)
        
        IPAM
        DEFAULT SUBNET: 10.32.0.0/12
        RANGE: 10.32.0.0/12
        STATUS: ready
        
        PLUGIN
        STATUS: not running
        
        PROXY
        STATUS: not running

    Is it something we want to document here?
    Alternatively, is there some Weave Cloud documentation I could point to? (which I think would be more appropriate in this case)

PR #3057, opened 2017-07-11
Fix weave-npc race condition

mikebryant:
It looks like the order in which `NetworkPolicy` and `Namespace` events are received can break weave-npc's ipset creation.

Not sure why, but here's a test that demonstrates it

See https://weave-community.slack.com/archives/C2ND76PAA/p1499776696249627

Also #3059

### DIFF
--- a/npc/controller_test.go
+++ b/npc/controller_test.go
@@ -1,0 +1,182 @@
+package npc
+
+import (
+	"log"
+	"testing"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
+	"github.com/weaveworks/weave/npc/ipset"
+	unversionedapi "k8s.io/client-go/pkg/api/unversioned"
+	coreapi "k8s.io/client-go/pkg/api/v1"
+	extnapi "k8s.io/client-go/pkg/apis/extensions/v1beta1"
+	"k8s.io/client-go/pkg/util/intstr"
+)
+
+type mockSet struct {
+	name    ipset.Name
+	setType ipset.Type
+	subSets map[string]bool
+}
+
+type mockIPSet struct {
+	sets map[string]mockSet
+}
+
+func newMockIPSet() mockIPSet {
+	i := mockIPSet{
+		sets: make(map[string]mockSet),
+	}
+
+	return i
+}
+
+func (i *mockIPSet) Create(ipsetName ipset.Name, ipsetType ipset.Type) error {
+	if _, ok := i.sets[string(ipsetName)]; ok {
+		return errors.Errorf("ipset %s already exists", ipsetName)
+	}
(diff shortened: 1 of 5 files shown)


mikebryant:
Huh. circleci isn't running my new `npc/controller_test.go` file

bboreham:
@mikebryant we [run](https://github.com/weaveworks/build-tools/blob/11400eaa2cea2f954fb3e5ecdf5e5d63649c25c8/test#L89-L94) [a Python script](https://github.com/weaveworks/build-tools/blob/master/sched) to divide work between the two shards

mikebryant:
Not sure why the smoke tests are failing - other branches/PRs seem to be affected too?

bboreham:
It appears to be because Ubuntu does some update on first boot that is leaving a defunct process, which triggers our check to see if _we_ left a defunct process.

And I think I know how to fix that but Docker have changed all their install processes so I have to update our Docker versioned install before I can apply the Ubuntu fix.

common/iptables.go
```diff
@@ -0,0 +1,7 @@
+package common
```
bboreham: Generally `common` is something we'd rather have less of - could this go in `net` instead?  Or maybe just local to `npc` since it doesn't appear to be used outside.

npc/controller_test.go
```diff
@@ -0,0 +1,177 @@
...
+	// Namespaces first
+	m := newMockIPSet()
+	controller := New("foo", &mockIPTables{}, &m)
+
+	controller.AddNamespace(sourceNamespace)
+	controller.AddNamespace(destinationNamespace)
+
+	controller.AddNetworkPolicy(networkPolicy)
+
+	require.Equal(t, true, m.sets["weave-I239Zp%sCvoVt*D6u=A!2]YEk"].subSets["weave-HboJG1fGgG]/SR%k9H#hv5e96"])
```
bboreham: could we have these strings as `const` or something that is calculated?

bboreham:
@mikebryant if you rebase on latest 2.0 it should clean up most of the CI errors.

mikebryant:
@bboreham I did that about a minute after the CI change merged, and it's now passed :) (Though netlify failed instead? weird)

will fix the comments now

common/iptables.go
mikebryant: Sure

mikebryant:
@bboreham How's that?

mikebryant:
Hmm. I'm pretty sure I'm rebased on the CI changes, but still seeing defunct processes

bboreham:
It's down from ~30 to 2 this time - I'll kick it off again and see if I can spot the reason

mikebryant:
@bboreham Tests are passing, can I get another review/merge please? :)

bboreham:
For the record, my change to avoid defunct processes caused by Ubuntu daily updates was ineffective.  But recreating the image "fixes" it for >24hrs, so we can mitigate until I fix it properly.

PR #3066, opened 2017-07-14
label weavedb image

rade:
Fixes weaveworks/scope#2610


bboreham:
I think I would have added the owner/source/description stuff at the same time, but this is OK for now.

PR #3071, opened 2017-07-19
Remove localhost:8080 fallback for Kubernetes api-server

bboreham:
It was a hack for a badly-configured Kubernetes install: it shouldn't be necessary, it will only work on an unsecured install, and it generates confusing error messages when something is really wrong.


brb:
LGTM.

I've noticed that the `2.0` branch has diverged from `master`: https://github.com/weaveworks/weave/compare/master...2.0?expand=1. Is it by accident, or have we stopped merging `2.0` to `master`?

PR #3078, opened 2017-08-02
clarify rmpeer caution in docs

rade:
- "You cannot call `weave rmpeer` on more than one host" - yes you can, but you must not call it on more than one host *for the same peer*
- `weave forget` has nothing to do with this
- "results in duplicate IPs" - no, it causes duplicate address range assignments
- "the address is reassigned" - it's the address *range* that is reassigned
- "claim" is a new and confusing term -> drop
- "address space" -> "address range" (that's what it's called elsewhere)

### DIFF
--- a/site/tasks/ipam/stop-remove-peers-ipam.md
+++ b/site/tasks/ipam/stop-remove-peers-ipam.md
@@ -37,22 +37,22 @@ name. Alternatively, you can supply a peer name as shown in `weave status`.
 
 ### <a name="caution-rmpeer"></a>Caution###
 
-You cannot call `weave rmpeer` on more than one host. The address
-space, which was owned by the stale peer cannot be left dangling, and
-as a result it gets reassigned. In this instance, the address is
-reassigned to the peer on which `weave rmpeer` was run. Therefore, if
-you run `weave forget` and then `weave rmpeer` on more than one host
-at a time, it results in duplicate IPs on more than one host.
+Do not invoke `weave rmpeer` for the same peer on more than one
+host.
 
-Once the peers detect the inconsistency, they log the error and drop
+The removed peer's address range cannot be left dangling and is
+therefore reassigned to the peer on which `weave rmpeer` was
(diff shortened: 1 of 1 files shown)


site/tasks/ipam/stop-remove-peers-ipam.md
```diff
@@ -37,22 +37,20 @@ name. Alternatively, you can supply a peer name as shown in `weave status`.
...
-You cannot call `weave rmpeer` on more than one host. The address
-space, which was owned by the stale peer cannot be left dangling, and
-as a result it gets reassigned. In this instance, the address is
-reassigned to the peer on which `weave rmpeer` was run. Therefore, if
-you run `weave forget` and then `weave rmpeer` on more than one host
-at a time, it results in duplicate IPs on more than one host.
-
-Once the peers detect the inconsistency, they log the error and drop
+You must not invoke `weave rmpeer` for the same peer on more than one
+host. The removed peer's address range cannot be left dangling and is
```
abuehrle: You could make the first sentence stronger by saying:

Do not invoke `weave rmpeer` for the same peer on more than one host.

site/tasks/ipam/stop-remove-peers-ipam.md
```diff
@@ -37,22 +37,20 @@ name. Alternatively, you can supply a peer name as shown in `weave status`.
...
+therefore reassigned to the peer on which `weave rmpeer` was
+run. Consequently, if you run `weave rmpeer` for the same peer on more
+than one host, the removed peer's address range will be owned by
+multiple peers.
+
```
abuehrle: LGTM sounds pretty clear to me

site/tasks/ipam/stop-remove-peers-ipam.md
rade: I thought "You must not" was stronger than "Do not". But the latter reads better, so I've made that change and also put the whole sentence in its own paragraph, so it doesn't get lost in the explanation that follows.

Issue #3084, opened 2017-08-09
DNS: auto-add custom container names/aliases


berlic:
In our weave network we launch several distributed apps in containers on different hosts.  
We name containers for `myapp` as `myapp-1`, `myapp-2`, `myapp-N` to be able to address them separately (for example from load balancers).  
Also for some application we give them custom host names as `myapp-master` (for `myapp-1`) and `myapp-slave` (for `myapp-2` .. `myapp-N`) to make use of DNS balancing.  

Current version of weave automatically manages only `{{.Config.Hostname}}.{{.Config.Domainname}}` value. See [watcher.go](https://github.com/weaveworks/weave/blob/v2.0.1/plugin/net/watcher.go#L45) code.

We've made a quick poc to automatically add not only hostname, but also container name if it differs:

```
diff --git a/plugin/net/watcher.go b/plugin/net/watcher.go
index 9a79561..c17f48b 100644
--- a/plugin/net/watcher.go
+++ b/plugin/net/watcher.go
@@ -46,6 +46,13 @@ func (w *watcher) ContainerStarted(id string) {
                                if err := w.weave.RegisterWithDNS(id, fqdn, net.IPAddress); err != nil {
                                        w.driver.warn("ContainerStarted", "unable to register %s with weaveDNS: %s", id, err)
                                }
+                                alias := fmt.Sprintf("%s.weave.local.", info.Name[1:]) // Remove prefix slash from name
+                                // w.driver.warn("ContainerStarted", "Container alias: %s", alias)
+                                if ( alias != fqdn ) {
+                                        if err := w.weave.RegisterWithDNS(id, alias, net.IPAddress); err != nil {
+                                               w.driver.warn("ContainerStarted", "unable to register %s with weaveDNS: %s", id, err)
+                                        }
+                                }
                        }
                        rootDir := "/"
                        if w.driver.isPluginV2 {
```
This works fine and do the job nicely.

Is it possible to add such a functionality into main branch? I can submit PR if there is a chance for it to be accepted.

If possible, please advice how can I refer custom domain suffix (supplied with `--dns-domain` option) inside `ContainerStarted` method of watcher? For sake of simplicity I've hardcoded it into my POC snippet.

This functionality can also rely on some container labels, such as `weave.works.dns-alias` or something.

bboreham:
Hi @berlic and thanks for the suggestion!

[...]

PR #3090, opened 2017-08-15
Register container name and its network aliases with weaveDNS

berlic:
Implements #3084 

### Before:

When container in "our" network is started, we try to register only  
`<Config.Hostname>.<Config.Domainname>` DNS record.

### After:

When container in "our" network is started, we try to register:

- `<Config.Hostname>.<Config.Domainname>` (as before)
- `<Name>.<Configured_weave_domain>`, default weave domain is `weave.local.`
- all aliases from `<NetworkSettings.Networks.our_network.Aliases>`

### Additional information

Was forced to update `fsouza/go-dockerclient` submodule, because version (rather old: Feb 20) pinned in current master was unaware of network aliases. It was introduced on May 22 with [this](https://github.com/fsouza/go-dockerclient/commit/435a498ac62c62d69de89778bf367331ccd55dae) commit. Still decided to update to current latest version – not sure if it is OK.

Also I've noted that for containers in weave network, an alias with container ID is automatically created. As we try to add every alias from the list it is being ignored because not in weave domain. Sample weave network object:

```
                "weave": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": [
                        "foo.weave.local",
                        "bar.weave.local",
                        "hello.world",
                        "394fa2c30138"
                    ],
                    "NetworkID": "049aa916a5c1...9880d1a1d",
                    "EndpointID": "e82435370cf7...99eea3f9d",
                    "Gateway": "",
                    "IPAddress": "172.16.0.2",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:a5:b6:55:9a:e5"
                }
```

### DIFF
--- a/plugin/net/watcher.go
+++ b/plugin/net/watcher.go
@@ -39,8 +39,16 @@ func (w *watcher) ContainerStarted(id string) {
 		if network.isOurs {
 			if w.driver.dns {
 				fqdn := fmt.Sprintf("%s.%s", info.Config.Hostname, info.Config.Domainname)
-				if err := w.weave.RegisterWithDNS(id, fqdn, net.IPAddress); err != nil {
-					w.driver.warn("ContainerStarted", "unable to register %s with weaveDNS: %s", id, err)
+				weaveDomain, _ := w.weave.DNSDomain()
+				// Container name is prefixed with '/' so we cut first character
+				name := fmt.Sprintf("%s.%s", info.Name[1:], weaveDomain)
+				aliases := append([]string(nil), fqdn, name)
+				aliases = append(aliases, net.Aliases...)
+				for _, alias := range aliases {
+					w.driver.debug("ContainerStarted", "going to register %s with weaveDNS", alias)
+					if err := w.weave.RegisterWithDNS(id, alias, net.IPAddress); err != nil {
+						w.driver.warn("ContainerStarted", "unable to register %s with weaveDNS: %s", id, err)
+					}
 				}
 			}
 			netNSPath := weavenet.NSPathByPidWithProc(w.driver.procPath, info.State.Pid)


Santinell:
@bboreham What about this feature?
It should be very useful.

Santinell:
Or maybe you can fix another problem with aliases?

My problem:
I am start container in weave network
Then I attach alias to container (with `weave dns-add` command)
And everything works fine until fall of container. 
When it happens Docker start container again(restart = always) but alias added on start don't recovering back

garkin:
Any love for this? 
This is life saving for primitive configurations with docker-compose.

bboreham:
I'm looking through old PRs, and I can't quite figure out what happened to this one.  There was a little bit of discussion on the issue.
Is this feature still wanted?

berlic:
Yes, we'd be happy to get this merged.
Use case for non-kube environment: replicated service `my-service` with `replica-1`, `replica-2`, `replica-3` containers.
Current WeaveDNS implementation allows to get either load balanced `my-service.weave.local` (if set as hostname in every container), or `replica-<N>.weave.local`.
With this modification, all names (load-balanced service and individual container) will be registered within DNS service, plus `my-service-master.weave.local`, etc. could be added via network aliases.

Santinell:
@bboreham Please, add this feature. It's really important

bboreham:
The mechanism for updating vendor has changed; @berlic if you would like to rebase and update vendor via `dep` that should resolve the merge conflict.

Also it would be good to update the [documentation](https://github.com/weaveworks/weave/tree/master/site/tasks/weavedns)
(maybe clarifying where this is useful - I think you have to be using the Weave Net Docker Plugin, but not Docker DNS).

Issue #3095, opened 2017-08-22
weave-npc crashing on k8s 1.7.4


bismarck:
## What you expected to happen?
weave 2.0.2 pods should start without an issue

## What happened?
weave-npc container crashes

## How to reproduce it?

On a k8s 1.7.4 cluster that has some network policies defined run
kubectl apply -f https://cloud.weave.works/k8s/v1.6/net

## Anything else we need to know?

We are running k8s on AWS, built with KOPS.

## Versions:

```
$ weave version
weave script 2.0.2
$ docker version
Client:
 Version:      1.12.6
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   78d1802
 Built:        Tue Jan 10 20:17:57 2017
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.6
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   78d1802
 Built:        Tue Jan 10 20:17:57 2017
 OS/Arch:      linux/amd64
$ uname -a
Linux ip-10-203-20-75 4.4.78-k8s #1 SMP Fri Jul 28 01:28:39 UTC 2017 x86_64 GNU/Linux
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.4", GitCommit:"793658f2d7ca7f064d2bdf606519f9fe1229c381", GitTreeState:"clean", BuildDate:"2017-08-17T17:03:51Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.4", GitCommit:"793658f2d7ca7f064d2bdf606519f9fe1229c381", GitTreeState:"clean", BuildDate:"2017-08-17T08:30:51Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
```

## Logs:
```
$ kubectl logs -n kube-system <weave-net-pod> weave -c weave-npc
INFO: 2017/08/22 18:09:46.888634 Starting Weaveworks NPC 2.0.2; node name "ip-10-203-19-219.ec2.internal"
INFO: 2017/08/22 18:09:46.905166 Serving /metrics on :6781
Tue Aug 22 18:09:46 2017 <5> ulogd.c:843 building new pluginstance stack: 'log1:NFLOG,base1:BASE,pcap1:PCAP'
[...]

PR #3096, opened 2017-08-23
fix: Missing nil checks

mikebryant:
Fixes #3095

PR #3097, opened 2017-08-23
Extend weave-kube smoke test: add a policy with blank 'from'

bboreham:
Per issue #3095 this crashes the commenting code.
Also added check to see if the pods are still running.


bboreham:
Single extra test failed as expected; I'm going to merge this first and then #3096

Issue #3102, opened 2017-08-26
weave report -f returns misleading error message


dtshepherd:
## What you expected to happen?

`weave report` should return helpful error message when invalid format string is passed to `-f` option.

## What happened?

`weave report -f '{{ .Router.Target }}'` returned an error message stating "weave container is not present. Have you launched it?".  The `weave` container was running, so the reason for the error was wrong.  The error was caused by an invalid format string `.Router.Target` does not exist, it needed to be `.Router.Targets`.  

## How to reproduce it?

1. Launch weave
2. Run `weave report -f '{{ .Router.Target }}'`

## Anything else we need to know?

Nope

## Versions:

```
$ weave version
weave script 2.0.1
weave 2.0.1
$ docker version
Client:
 Version:      17.06.0-ce
 API version:  1.30
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:23:31 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.06.0-ce
 API version:  1.30 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:19:04 2017
 OS/Arch:      linux/amd64
 Experimental: false
$ uname -a
Linux NUC 4.4.0-87-generic #110-Ubuntu SMP Tue Jul 18 12:55:35 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```
## Logs:

NA

## Network:

NA

dtshepherd:
I just updated the ticket above.  This only affects the `net-plugin`.  Normally launched `weave` reports a helpful message "error during template execution".

brb:
Thanks for the report.

The problem is that `call_weave` (called by `weave report`) calls `check_running` (https://github.com/weaveworks/weave/blob/v2.0.4/weave#L653) in the case of `/report` returning an error. The latter function expects the weave container to be running which is not the case for `net-plugin`.

dtshepherd:
Yeah, I figured that out. I commented in the community slack, but I'm working on a PR to make `weave` script more aware of the `net-plugin`. I know this may overlap with #2938, but maybe I can help iterate the improvements.

PR #3110, opened 2017-09-04
Remove docker binary from weaveexec container

### LINKED ISSUE #2957: Remove docker binary from weaveexec container
I looked at what it would take to do this.  The benefits would be:
 - avoid triggering virus scans (the docker binary includes some vulnerable code, although we don't execute that code)
 - smaller weaveexec image
 - avoid unnecessary version coupling caused by bundling docker 1.10.3
 - avoid icky feeling from downloading a tarfile and extracting a binary in the Makefile
 - weave net would be available on arm64 (where we cannot find a usable docker binary)

The subcommands used now are kill, stop, inspect, run, rm, exec, create, pull, ps.
I give arguments below why exec and pull would not be needed; the rest could be re-implemented in weaveutil.

Only a subset of `docker run` arguments are needed: `-d`, `-v`, `-e`, `--name`, `--privileged`, `--net=host`, `--pid=host`, `--volumes-from`; although we have suggested that users set `WEAVE_DOCKER_ARGS` to add arbitrary more.

Things that would need converting:
- kill_container, called from wait_for_status (stop a restarting weave) - docker kill
- stop, called from weave stop - docker stop
- ask_version - printing the version of the weave container - inspect and run
- is_plugin_v1_enabled - inspect, looking for --plugin; this is bugged anyway
- check_running - inspect
- check_not_running - inspect and rm
- container_id - inspect
- http_call_unix - exec - could move this feature to TCP socket on 6784
- rewrite_etc_hosts - can move the docker inspect to weaveutil rewrite-etc
- with_container_fqdn - inspect
- launch - creating weavedb and weavewait containers - inspect, create
- weave reset - docker stop, rm, ps
- setup - docker pull - this could move to the non-remoted part of the script

Things that don't need converting:
- try_create_bridge - inspect - this is pointless because `/` isn't mounted
- util_op - this code only used when not running inside weaveexec

dtshepherd:
Fixes #2957 

Creating `weaveutil` commands to remove `docker` binary dependency in `weaveexec`.

This is still a WIP... Need to port the `docker run`, `docker create`, `docker ps`, parts of `docker inspect` (although probably want high-level `weaveutil` commands for `check_running`, etc).

@bboreham


bboreham:
Thanks for this, @dtshepherd.  It does seem a little more complicated than we need - e.g. we never call `kill` with a signal so we don't need a table to decode them.  Or the timeout parameter.

dtshepherd:
Ok I'll remove the extra options.  I'm working on some of the `docker inspect` replacements now, which aren't using `docker inspect` directly.

Any idea how to get the CircleCI tests to pass?  It complains about a missing secret key.

bboreham:
The secret key is only accessible to PRs in the repo, to avoid people submitting PRs which steal the key and/or steal the resources it controls. 

I can push your branch to the repo and the tests will run.

dtshepherd:
Finished replacing `docker inspect` in `weave` script (`check_runnning`, `check_not_running`, `ask_version`, `rewrite_etc_hosts`, `with_container_fqdn`).

Still need to implement commands to create `weavedb` and `weavevolumes` containers.  Also, need to implement `docker ps`, which really just loads volume containers to delete, so maybe create a higher level command for that action.  Probably should generalize some of the work done for `weaveutil ask-version` to launch `weave` container too.

dtshepherd:
@bboreham Just committed the removal of the remaining `docker ...` commands except for the single complicated `WEAVE_CONTAINER=$(docker run ...)`.  The `launch`'s `docker run` takes many options (some from user-defined `WEAVE_DOCKER_ARGS`) and we'll need to at least implement `-d`, `-v`, `-e`, `--name`,  `--restart`, `--privileged`, `--net=host`, `--pid=host`, `--volumes-from`.

Thoughts on the implementation?  Probably need to implement a near direct pass-through of the `docker run` with the limited options...

dtshepherd:
OK, I tried a first cut at the `run-container` command.  I'm not super happy with it because it doesn't use a true argument parser, but the smoke tests seem to work (although, I've noticed some fragility with some of the tests (e.g. `840_weave_kube_3_test.sh`) even on `master` with none of my changes.

@bboreham Can you please kick off a CI build and test?

weave
```diff
@@ -1231,10 +1200,10 @@ launch() {
     fi
 
     # Create a data-only container for persistence data
-    if ! docker inspect -f ' ' $DB_CONTAINER_NAME > /dev/null 2>&1 ; then
-       protect_against_docker_hang
-       docker create -v /weavedb --name=$DB_CONTAINER_NAME \
-           --label=weavevolumes $WEAVEDB_IMAGE >/dev/null
+    if ! util_op container-state $DB_CONTAINER_NAME > /dev/null 2>&1 ; then
```
bboreham: Why `container-state` rather than `container-id`, say ?  We don't expect this one to be restarting, etc.

prog/weaveutil/container.go
```diff
@@ -0,0 +1,273 @@
...
+	container, err := c.InspectContainer(containerID)
+	if err != nil {
+		return fmt.Errorf("unable to inspect container %s: %s", containerID, err)
+	}
+
+	if len(args) == 1 {
+		fmt.Print(container.State.StateString())
+	} else {
+		image := args[1]
+		match1, _ := regexp.MatchString(image, container.Image)
```
bboreham: Why is regexp match the right thing to do here?

dtshepherd: `Image` is the `sha256:...` hash and `Config.Image` is the <name>:<tag>.  I was trying to support friendly name or hash for running the image comparison.    I guess for the `Image` field, I could just parse out the hash, but what if short hash was passed in.  As for `Config.Image`, I could parse out just the name without the tag (split on :) for the comparison.  The previous code did _not_ compare the tags, so I didn't want to change that.

What's your thoughts?  I was just trying to make the container-state check fairly generic.

prog/weaveutil/container.go
```diff
@@ -0,0 +1,273 @@
...
+	err = c.StartContainer(container.ID, &hostConfig)
+	if err != nil {
+		return fmt.Errorf("unable to start container: %s", err)
+	}
+
+	fmt.Print(container.ID)
+	return nil
+}
+
+func listContainers(args []string) error {
```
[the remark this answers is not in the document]
dtshepherd: Yep, with the added `label` arg.  I was avoiding using `weave/common/client` as we previously discussed, but I could reuse some of that code.  What do you think?

Maybe that's where the shared `InspectContainer` logic belongs too? (used in attach.go and this file). 
 Thoughts?

prog/weaveutil/version.go
```diff
@@ -0,0 +1,63 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+func askVersion(args []string) error {
```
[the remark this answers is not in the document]
dtshepherd: Yeah, I like that idea.  I didn't like this code anyway (hence my `TODO` still in the file :-)

Just so we're on the same page, I'm guessing this logic would belong in the `weave` script, right?

weave
dtshepherd: Yeah, I just liked the idea of using a command named `container-state` rather than `container-id` to fetch the state of a container.  It read like self-documenting code to me.  Maybe there should be command `is-container-running` or something like that, which is an alias for `container-id`?

prog/weaveutil/container.go
```diff
@@ -0,0 +1,273 @@
...
+		cmdUsage("container-id", "<container-name-or-short-id>")
+	}
+	containerID := args[0]
+
+	c, err := docker.NewVersionedClientFromEnv("1.18")
+	if err != nil {
+		return fmt.Errorf("unable to connect to docker: %s", err)
+	}
+
+	container, err := c.InspectContainer(containerID)
```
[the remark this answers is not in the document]
dtshepherd: Yeah, probably.  I didn't do a 2nd pass after I noticed the duplication.   I'll fix this in the other refactors.

prog/weaveutil/image.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+func pullImage(args []string) error {
```
dtshepherd: @bboreham What's your thoughts on me removing this command (and file) and following your suggestion here: https://github.com/weaveworks/weave/issues/2957#issuecomment-327787115

prog/weaveutil/version.go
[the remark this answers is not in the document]
dtshepherd: Ok I reworked, checkout https://github.com/weaveworks/weave/pull/3110/commits/2f64c9d3770211a21c38e0ab69887678916940d3

prog/weaveutil/image.go
dtshepherd: I went ahead and did it, since it was a simple change and I can back out if you decide you want the separate `weaveutil` command.

dtshepherd:
@bboreham I think this is ready for a 2nd pass review.  I just rebased with upstream master as well, since there was some conflicts with `weave` script.

weave
```diff
@@ -222,28 +216,40 @@ kill_container() {
...
-    # "env --restore" is special because we always want to process it
-    # at the client end.
-    if [ "${ORIG_DOCKER_HOST-unset}" = "unset" ] ; then
-        echo "Nothing to restore. This is most likely because there was no preceding invocation of 'eval \$(weave env)' in this shell." >&2
-        exit 1
-    else
-        echo "DOCKER_HOST=$ORIG_DOCKER_HOST"
+# Handle special case $1 commands that run locally at the client end
+case "$1" in
+    help|--help)
```
bboreham: Looks like `handle_help_arg` can be deleted as no longer called.

weave
```diff
@@ -265,8 +271,9 @@ PLUGIN_NAME="$DOCKERHUB_USER/net-plugin"
 OLD_PLUGIN_CONTAINER_NAME=weaveplugin
 CNI_PLUGIN_NAME="weave-plugin-$IMAGE_VERSION"
 CNI_PLUGIN_DIR=${WEAVE_CNI_PLUGIN_DIR:-$HOST_ROOT/opt/cni/bin}
+VOLUMES_LABEL=weavevolumes
 # Note VOLUMES_CONTAINER which is for weavewait should change when you upgrade Weave
-VOLUMES_CONTAINER_NAME=weavevolumes-$IMAGE_VERSION
+VOLUMES_CONTAINER_NAME=$VOLUMES_LABEL-$IMAGE_VERSION
```
bboreham: `VOLUMES_LABEL` and this container name are not related except by coincidence, so this is potentially confusing.

weave
bboreham: ok, `container-state` is fine.

weave
```diff
@@ -1231,10 +1211,10 @@ launch() {
     fi
 
     # Create a data-only container for persistence data
-    if ! docker inspect -f ' ' $DB_CONTAINER_NAME > /dev/null 2>&1 ; then
-       protect_against_docker_hang
-       docker create -v /weavedb --name=$DB_CONTAINER_NAME \
-           --label=weavevolumes $WEAVEDB_IMAGE >/dev/null
+    if ! util_op container-state $DB_CONTAINER_NAME > /dev/null 2>&1 ; then
+        protect_against_docker_hang
```
bboreham: spurious extra space

bboreham:
This is great; I can see some potential further refactorings which would bring the code weight down, but happy to merge what you have and leave them for the future.

Regarding squashing commits, one would be fine, but if you see a few logically separate ones then that is fine too.

weave
dtshepherd: An extra space because it wasn't aligned.  I can revert if you want that in a different commit.  I noticed it when adding the line below it...

weave
dtshepherd: Removed the reuse of the variable

weave
dtshepherd: Yep, not sure how I missed that...

prog/weaveutil/container.go
```diff
@@ -0,0 +1,282 @@
...
+		return fmt.Errorf("unable to start container: %s", err)
+	}
+
+	fmt.Print(container.ID)
+	return nil
+}
+
+func listContainers(args []string) error {
+	if len(args) < 1 {
+		cmdUsage("list-containers", "[<label>]")
```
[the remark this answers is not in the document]
dtshepherd: Made it support being optional, only added 3-4 lines of code.

prog/weaveutil/container.go
dtshepherd: Changed to `strings.HasPrefix`

dtshepherd:
@bboreham I think I responded to your review comments (some of the comments weren't on lines that needed to change so github didn't dismiss).  I squashed to two commits (all of my changes + commit stolen from #2952 for `try_bridge_create`).

I think some of the code weight could be reduced if we merged some of `prog/weaveutil/container.go` with `common/docker/client.go`.  Right now, the two are completely independent and I didn't want to change that in this branch.  That refactor seemed like an effort that may cover the entire repository, not just `weaveutil`.

Hopefully this next review pass looks good for merging.  I know this is a risky change, so I understand if its delayed for release to master.  I'd like to build upon this PR to fix #3102.

Makefile
```diff
@@ -406,7 +397,6 @@ integration-tests: all testrunner
 #     PLAYBOOK="<filename>" \
 #     RUNNER_ARGS="<...>" \
 #     TESTS="<...>" \  # Can be set to only run one or a few tests instead of the full test suite.
-#     DOCKER_VERSION=<...> \
```
bboreham: this shouldn't be removed - it refers to the docker version used to build VMs for integration testing

Makefile
```diff
@@ -307,10 +302,6 @@ $(NETWORKTESTER_UPTODATE): test/images/network-tester/Dockerfile $(NETWORKTESTER
 $(WEAVE_EXPORT): $(IMAGES_UPTODATE) $(WEAVEDB_UPTODATE)
 	$(SUDO) DOCKER_HOST=$(DOCKER_HOST) docker save $(addsuffix :latest,$(IMAGES)) | gzip > $@
 
-$(DOCKER_DISTRIB):
-	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
-	cd $(shell dirname $@) && sha256sum -c $(shell pwd)/build/shasums/docker-tgz-$(WEAVEEXEC_DOCKER_ARCH).sha256sum
```
bboreham: you could remove `build/shasums/docker-tgz-*`

weave
bboreham: doesn't matter

Makefile
dtshepherd: done

dtshepherd:
Fixed the Makefile and removed those shasum files, good catch.  We'll see what the CI looks like after these changes...

bboreham:
Nice! Image size comes down by ~34MB

dtshepherd:
I didn't even check that! Yeah that's a good win.

PR #3113, opened 2017-09-08
Add instructions on troubleshooting the V2 plugin

bboreham:
cc @abuehrle

### DIFF
--- a/site/troubleshooting.md
+++ b/site/troubleshooting.md
@@ -293,22 +293,22 @@ restart.
 
 ## <a name="v2plugin"></a>Troubleshooting the V2 plugin
 
-If you have installed Weave Net via `docker plugin install`, then you
-will need to download the `weave` script to run `weave status`, `weave ps`
-or `weave report` as above.  Do this as described in the
-[install instructions](/site/install/installing-weave.md).
+If Weave Net is installed via `docker plugin install`, download the
+`weave` script to run `weave status`, `weave ps` or `weave report` as
+above.  Install Weave Net by following the [install
+instructions](/site/install/installing-weave.md).
 
-Docker "v2" plugins run as containers, but at a lower level within the
-Docker environment, so you cannot view them with `docker ps`, `docker
-inspect`, etc.
+Docker "v2" plugins do run as containers, but at a lower level within
+the Docker environment.  Because of this, you cannot view them with
+`docker ps`, `docker inspect`, etc.
 
 Do not run `weave launch`, `weave stop` or similar commands when using
 this plugin; use the `docker plugin` commands instead.  You can run
 `weave reset`, but only after disabling the plugin via `docker plugin disable`.
 
-Diagnostic logs from the plugin will go to the same place as the
-Docker daemon, which will depend on your Linux install. For example,
-if it uses `systemd`, then do this to view the Docker and plugin logs:
+Diagnostic logs from the plugin go to the same place as the Docker
+daemon, which will depend on your Linux install. For example, if it
+uses `systemd`, then do this to view the Docker and plugin logs:
 
     sudo journalctl -u docker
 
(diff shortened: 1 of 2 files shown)


site/troubleshooting.md
```diff
@@ -290,6 +291,27 @@ Proxy](/site/tasks/weave-docker-api/weave-docker-api.md) will re-attach any appl
 containers that it originally attached to the Weave network when they
 restart.
 
+## <a name="v2plugin"></a>Troubleshooting the V2 plugin
+
+If you have installed Weave Net via `docker plugin install`, then you
+will need to download the `weave` script to run `weave status`, `weave ps`
+or `weave report` as above.  Do this as described in the
+[install instructions](/site/install/installing-weave.md).
+
```
abuehrle: (Removing some passive voice.)

If Weave Net is installed via `docker plugin install` download the `weave` script and run `weave status`, `weave ps` or `weave report`.   Install Weave Net by following the [install instructions](/site/install/installing-weave.md).

site/troubleshooting.md
```diff
@@ -290,6 +291,27 @@ Proxy](/site/tasks/weave-docker-api/weave-docker-api.md) will re-attach any appl
 containers that it originally attached to the Weave network when they
 restart.
```
abuehrle: (Remove wills. also passive voice as a general rule.)

Note that the Weave Net router creates the `weave` network bridge if necessary when it restarts. The [Weave Net Docker API Proxy](/site/tasks/weave-docker-api/weave-docker-api.md) then re-attaches any application
containers that it originally attached to the Weave network when they
restart.

site/troubleshooting.md
```diff
@@ -290,6 +291,27 @@ Proxy](/site/tasks/weave-docker-api/weave-docker-api.md) will re-attach any appl
...
+
+If you have installed Weave Net via `docker plugin install`, then you
+will need to download the `weave` script to run `weave status`, `weave ps`
+or `weave report` as above.  Do this as described in the
+[install instructions](/site/install/installing-weave.md).
+
+Docker "v2" plugins run as containers, but at a lower level within the
+Docker environment, so you cannot view them with `docker ps`, `docker
+inspect`, etc.
+
```
abuehrle: Docker "v2" plugins do run as containers, but at a much lower level within the
Docker environment. Because of this, you cannot view them with Docker commands like `docker ps`, `docker
inspect`, etc.

site/troubleshooting.md
```diff
@@ -290,6 +291,27 @@ Proxy](/site/tasks/weave-docker-api/weave-docker-api.md) will re-attach any appl
...
+Do not run `weave launch`, `weave stop` or similar commands when using
+this plugin; use the `docker plugin` commands instead.  You can run
+`weave reset`, but only after disabling the plugin via `docker plugin disable`.
+
+Diagnostic logs from the plugin will go to the same place as the
+Docker daemon, which will depend on your Linux install. For example,
+if it uses `systemd`, then do this to view the Docker and plugin logs:
+
```
abuehrle: (Remove will)

Diagnostic logs from the plugin go to the same place as the Docker daemon, which depends on your Linux install.

site/troubleshooting.md
```diff
@@ -290,6 +291,27 @@ Proxy](/site/tasks/weave-docker-api/weave-docker-api.md) will re-attach any appl
...
+    sudo journalctl -u docker
+
```
abuehrle: Put this in command format.

rade:
Can we ditch the "V2" everywhere?

bboreham:
@rade we need to distinguish the two plugins that we ship; "v2" is [Docker's term](https://docs.docker.com/engine/extend/config/); what do you suggest?

rade:
I guess I was wondering why we are still shipping the non-v2 (v1?) plugin.

bboreham:
So as not to impose a massive upheaval on users of the "legacy" plugin.
Since the code is mostly the same, the work on our side would be mostly in updating the docs and explaining the upgrade process.
Given how much more difficult it is to manage and troubleshoot V2 plugins, I am not especially eager to complete that task.

site/troubleshooting.md
bboreham: I don't follow: this is in the same format as the four commands above.

PR #3119, opened 2017-09-12
prometheus: export count of unreachable peers

caarlos0:
it should help monitoring for https://github.com/weaveworks/weave/issues/2797

### DIFF
--- a/prog/weaver/metrics.go
+++ b/prog/weaver/metrics.go
@@ -82,6 +82,18 @@ var metrics = []metric{
 				ch <- intGauge(desc, metrics.Flows)
 			}
 		}},
+	{desc("weave_ipam_unreachable_count", "Number of unreachable peers."),
+		func(s WeaveStatus, desc *prometheus.Desc, ch chan<- prometheus.Metric) {
+			if s.IPAM != nil {
+				var count int64
+				for _, entry := range s.IPAM {
+					if !entry.IsKnownPeer {
+						count++
+					}
+				}
+				ch <- intGauge(desc, count)
+			}
+		}},
 	{desc("weave_ipam_pending_allocates", "Number of pending allocates."),
 		func(s WeaveStatus, desc *prometheus.Desc, ch chan<- prometheus.Metric) {
 			if s.IPAM != nil {


bboreham:
Thanks for the contribution!

It isn't quite counting what the description says - each peer can have many "tokens" in the ring data structure, and you are counting the number of tokens owned by unreachable peers.

This metric would still be somewhat useful in monitoring #2797, but you might consider factoring out the computation in `printIPAMRanges` to generate a more tightly-scoped one?

caarlos0:
@bboreham thanks for the review!

I'm not quite sure I understand what you want...

as far as I understood, `printIPAMRanges` ranges over `ipam.Status.Entries` and prints `unreachable` when `IsKnownPeer` is false. The only relevant difference (unless I'm missing something, which is possible since this is my first pr to weave), is that it first checks if the peer is not the router - and I'm not really sure what this means...

Mind giving some more directions?

bboreham:
First it [builds a map](https://github.com/weaveworks/weave/blob/master/prog/weaver/http.go#L50) with the size for each peer, then it ranges [over that map](https://github.com/weaveworks/weave/blob/master/prog/weaver/http.go#L70) (i.e. one iteration per peer) to print them out.

caarlos0:
@bboreham hmm, right, so I can extract this somewhere, and use in both places, right?

bboreham:
Right.  Putting the summation in package `ipam` is slightly against the grain - the various `Status` methods return structs that are data-only and further analysis is meant to be done on the calling side.

caarlos0:
@bboreham any suggestions?

bboreham:
@caarlos0 I pushed f2d2326e5016b7c243336e985ef9253487cb0de9 to show what I meant

caarlos0:
@bboreham hmm, got it, cherry-picked your commit..

caarlos0:
@bboreham anything else I should do here?

bboreham:
Code looks fine now, thanks @caarlos0 !
I'm just re-running the full suite of CI tests.
If you want to squash any of the commits please do, otherwise I will probably squash them all on merge.

caarlos0:
@bboreham nice, feel free to squash them as you wish 🍻

caarlos0:
@bboreham I just added the updated to the docs I've missed earlier...

caarlos0:
thanks @bboreham, any plans for the release?

PR #3122, opened 2017-09-17
Enable FORWARDing to exposed network

### LINKED ISSUE #2758: Docker 1.13 defaults IP forwarding to DROP

brb:
Please review each commit separately. Each commit contains a self descriptive message helping to understand the introduced changes.

_Non-backward compatible change_: `weave expose` requires a running instance of `weaver` which makes not possible to run `weave expose` when weave has not been launched yet.

Fixes #2758


bboreham:
I fixed up the conflict (the other change was lint-driven) so it can be merged if tests still pass

Issue #3127, opened 2017-09-22
Service on StatefulSet pods in Default-deny namespace doesn't connect via Service's IP


RRAlex:
## What you expected to happen?

A namespace with a default-deny policy, with a statefulSet of pods using the following NetworkPolicy:
```
apiVersion: extensions/v1beta1
kind: NetworkPolicy
metadata:
  name: elasticsearch-logging-fluentd
  namespace: kube-system-logging
spec:
  podSelector:
    matchLabels:
      name: elasticsearch-logging
  ingress:
    - from:
        - podSelector:
            matchLabels:
              name: fluentd-es
      ports:
        - protocol: TCP
          port: 9200
```
... should let the ingress pods connect to the service's IP, resolved from its DNS.
ie: `elasticsearch-logging` (or `elasticsearch-logging.kube-system-logging.svc.cluster.local` or resolved IP) should be reachable from fluentd-es pods.

## What happened?

I'm unable to get a NetworkPolicy to work when using a deployment, daemonSet or StatefulSet using a service. 
As soon as a service is scaled and not just a single pod anymore, NetworkPolicy seems to stop working with the service IP. 
It does resolve, but doesn't connect to anything and timeouts.
```
root@fluentd-es-mgdr2:/# curl -v elasticsearch-logging:9200
* Rebuilt URL to: elasticsearch-logging:9200/
*   Trying 10.0.0.153...
* TCP_NODELAY set
^C
```

The elasticsearch pods' real IPs do go through, from any node...
```
root@fluentd-es-mgdr2:/# curl -v 10.1.0.26:9200
* Rebuilt URL to: 10.1.0.26:9200/
*   Trying 10.1.0.26...
* TCP_NODELAY set
* Connected to 10.1.0.26 (10.1.0.26) port 9200 (#0)
> GET / HTTP/1.1
> Host: 10.1.0.26:9200
> User-Agent: curl/7.52.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< content-type: application/json; charset=UTF-8
< content-length: 348
< 
{
  "name" : "elasticsearch-logging-0",
  "cluster_name" : "kubernetes-logging",
  "cluster_uuid" : "lgMREbo0QS-4Oo_Er2l8Xw",
  "version" : {
    "number" : "5.5.1",
    "build_hash" : "19c13d0",
    "build_date" : "2017-07-18T20:44:24.823Z",
    "build_snapshot" : false,
    "lucene_version" : "6.6.0"
[...]

PR #3128, opened 2017-09-25
Added 'things to watch out for' section in Kubernetes instructions

bboreham:
Triggered by #3127 

I typed it in a bit of a hurry; am sure the English can be improved

### DIFF
--- a/site/kubernetes/kube-addon.md
+++ b/site/kubernetes/kube-addon.md
@@ -14,6 +14,7 @@ The following topics are discussed:
 * [Network Policy Controller](#npc)
 * [Troubleshooting](#troubleshooting)
  * [Troubleshooting Blocked Connections](#blocked-connections)
+ * [Things to watch out for](#key-points)
 * [Changing Configuration Options](#configuration-options)
 
 
@@ -269,6 +270,15 @@ as per the below example:
 TCP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
 UDP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
 ```
+
+### <a name="key-points"></a> Things to watch out for
+
+- Don't turn on `--masquerade-all` on kube-proxy: this will change the
+  source address of every pod-to-pod conversation which will make it
+  impossible to correctly enforce network policies that restrict which
+  pods can talk.
+- If you do set the `--cluster-cidr` option on kube-proxy, make sure
+  it matches the `IPALLOC_RANGE` given to Weave Net (see below)
 
 ## <a name="configuration-options"></a> Changing Configuration Options
 


site/kubernetes/kube-addon.md
```diff
@@ -270,6 +271,14 @@ TCP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
 UDP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
 ```
 
+### <a name="key-points"></a><a name="key-points"></a> Things to watch out for
+
+- Don't turn on `--masquerade-all` on kube-proxy: this will change the
+  source address of every pod-to-pod conversation which will make it
+  impossible to write network policies that restrict which pods can talk.
```
rade: "write" -> "correctly enforce"

site/kubernetes/kube-addon.md
```diff
@@ -270,6 +271,14 @@ TCP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
 UDP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
 ```
 
+### <a name="key-points"></a><a name="key-points"></a> Things to watch out for
```
mikebryant: nit: Why is the `<a name="key-points"></a>` repeated?

bboreham: it was pasted from `troubleshooting` without going through the conscious mind...

site/kubernetes/kube-addon.md
```diff
@@ -14,6 +14,7 @@ The following topics are discussed:
 * [Network Policy Controller](#npc)
 * [Troubleshooting](#troubleshooting)
  * [Troubleshooting Blocked Connections](#blocked-connections)
+ * [Things to watch out for](#key-points)
 * [Changing Configuration Options](#configuration-options)
```
abuehrle: change to:  
   * Connection Options to be Aware of

(also list item should be indented as a child of Troubleshooting Blocked Connections)

site/kubernetes/kube-addon.md
```diff
@@ -270,6 +271,15 @@ TCP connection from 10.32.0.7:56648 to 10.32.0.11:80 blocked by Weave NPC.
...
+### <a name="key-points"></a> Things to watch out for
+
+- Don't turn on `--masquerade-all` on kube-proxy: this will change the
+  source address of every pod-to-pod conversation which will make it
+  impossible to correctly enforce network policies that restrict which
+  pods can talk.
+- If you do set the `--cluster-cidr` option on kube-proxy, make sure
+  it matches the `IPALLOC_RANGE` given to Weave Net (see below)
+
 ## <a name="configuration-options"></a> Changing Configuration Options
```
abuehrle: Don't use `--masquerade-all` on kube-proxy: this setting changes the
source address of every pod-to-pod conversation which makes it
impossible to correctly enforce the network policies that restrict which
pods can communicate.

If you set the `--cluster-cidr` option on kube-proxy, ensure
it matches the `IPALLOC_RANGE` given to Weave Net (see below)

site/kubernetes/kube-addon.md
bboreham: The list is not about connection options; it's all the things we want to tell users about Weave Net on Kubernetes.

site/kubernetes/kube-addon.md
bboreham: I don't think "the network policies" gives the right meaning - there are no policies by default; only ones you write, and you have to write ones which seek to restrict pods before you hit the problem (you can still write ones that restrict ports, for instance)

PR #3134, opened 2017-10-02
Mount iptables lock file in weave-kube

### LINKED ISSUE #2998: Improper iptables configuration in case of concurrent iptables access
We currently use weave on our Kubernetes cluster to provide the networking layer and we encounter from time to time networking issues with weave at container startup time.

Our non-production clusters are automatically stopped at night and re-started in the morning.
On several occasion, we noticed the Kubernetes network stack didn't work correctly in the morning.

The symptoms were that containers were not able to access resources outside of the cluster.
It generally impacted also internal access as the kube-dns was not properly working because the container was not able to reach external DNS servers.

After investigation, we noticed the following things:
  - the problem appear at the node level rather than cluster level: some of the nodes were not impacted and pods hosted on these nodes could access external resources,
  - restarting weave pods on a network-unhealthy node didn't solve the issue,
  - the packets sent outside of the cluster weren't properly masqueraded by iptables,
  - and indeed the WEAVE rule set was not present in the iptables nat table,
  - the weave container failed its healtcheck at startup at first and was restarted at least once on failed nodes

bboreham:
Fixes #2998 

`iptables` uses the file `/run/xtables.lock` so that if changes are under way, another instance of the `iptables` command can wait till it is finished.

We need to mount the lock file from the host to the container in order for this mechanism to work at all.

Note we rely on `kubelet` touching the file before running any pods - see https://github.com/kubernetes/kubernetes/pull/47212

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -91,6 +91,7 @@ build() {
     sed -i "/SCRIPT_VERSION=/ c\SCRIPT_VERSION=\"$VERSION\"" ./weave
     sed -i -e "s/:latest/:$VERSION/" -e "/imagePullPolicy: Always/d" ./prog/weave-kube/weave-daemonset.yaml
     sed -i -e "s/:latest/:$VERSION/" -e "/imagePullPolicy: Always/d" ./prog/weave-kube/weave-daemonset-k8s-1.6.yaml
+    sed -i -e "s/:latest/:$VERSION/" -e "/imagePullPolicy: Always/d" ./prog/weave-kube/weave-daemonset-k8s-1.7.yaml
     make SUDO=$SUDO WEAVE_VERSION=$VERSION DOCKERHUB_USER=$DOCKERHUB_USER
 
     if make tests; then
@@ -172,6 +173,13 @@ draft() {
         --tag $LATEST_TAG \
         --name "weave-daemonset-k8s-1.6.yaml" \
         --file "./prog/weave-kube/weave-daemonset-k8s-1.6.yaml"
+
+    github-release upload \
+        --user $GITHUB_USER \
+        --repo weave \
+        --tag $LATEST_TAG \
+        --name "weave-daemonset-k8s-1.7.yaml" \
+        --file "./prog/weave-kube/weave-daemonset-k8s-1.7.yaml"
 
     echo "** Draft $TYPE $RELEASE_NAME $VERSION created at"
     echo -e "\thttps://github.com/$GITHUB_USER/weave/releases/$LATEST_TAG"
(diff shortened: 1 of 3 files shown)


bboreham:
To make clear what is really changing between the 1.6 and 1.7 files:

```
$ diff -c prog/weave-kube/weave-daemonset-k8s-1.6.yaml prog/weave-kube/weave-daemonset-k8s-1.7.yaml
*** prog/weave-kube/weave-daemonset-k8s-1.6.yaml	2017-09-04 16:31:44.793192722 +0000
--- prog/weave-kube/weave-daemonset-k8s-1.7.yaml	2017-10-03 13:03:19.851290386 +0000
***************
*** 96,101 ****
--- 96,104 ----
                    mountPath: /host/var/lib/dbus
                  - name: lib-modules
                    mountPath: /lib/modules
+                 - name: xtables-lock
+                   mountPath: /run/xtables.lock
+                   readOnly: false
              - name: weave-npc
                env:
                  - name: HOSTNAME
***************
*** 110,115 ****
--- 113,122 ----
                    cpu: 10m
                securityContext:
                  privileged: true
+               volumeMounts:
+                 - name: xtables-lock
+                   mountPath: /run/xtables.lock
+                   readOnly: false
            hostNetwork: true
            hostPID: true
            restartPolicy: Always
***************
*** 138,142 ****
--- 145,152 ----
              - name: lib-modules
                hostPath:
                  path: /lib/modules
+             - name: xtables-lock
+               hostPath:
+                 path: /run/xtables.lock
        updateStrategy:
          type: RollingUpdate
```

bboreham:
cc @mikebryant for interest

mikebryant:
Looks good.
We've been running with this for several months and it's not caused any issues :)

Bregor:
@bboreham you made a change in `1.6.yaml` as well, but https://github.com/kubernetes/kubernetes/commit/d5c9d27ed763dfc312e3c2f3f234f08253fb64e4 exists only from `v1.7.0-beta.2`

bboreham:
@Bregor sorry, I'm not seeing the change to `1.6.yaml` - last commit is 3rd July
https://github.com/weaveworks/weave/blob/master/prog/weave-kube/weave-daemonset-k8s-1.6.yaml

Bregor:
@bboreham sorry, my bad. I looked at [this](https://github.com/weaveworks/weave/pull/3134#issuecomment-333835236) comment and saw 

    *** prog/weave-kube/weave-daemonset-k8s-1.6.yaml	2017-09-04 16:31:44.793192722 +0000

Sorry once again.

PR #3138, opened 2017-10-05
Add back EXEC_IMAGE so proxy can docker-run itself

### LINKED ISSUE #3136: Rewriting /etc/hosts is broken
Container's `/etc/hosts` to begin "# created by Weave"

## What happened?
(after I added more logging)
```
DEBU: 2017/10/04 13:04:00.292282 Running image "weaveworks/weaveexec"; entrypoint=["sh"]; cmd="[-c echo '# created by Weave - BEGIN\n# contai
ner hostname\n10.42.0.1       09...]"; binds=["/var/lib/docker/containers/09578e613661980fadb02f856d7d9909a5539a369bab6f087d6f89293187a03d:/container"]
ERRO: 2017/10/04 13:04:00.293423 Error from rewriting container's /etc/hosts: no such image
```

So the problem is that it is trying to run `weaveworks/weaveexec`, and it should actually be running something like `weaveworks/weaveexec:2.0.4`

bboreham:
Fixes #3136 

Also add error reporting so it's easier to figure out next time something breaks.

Was broken in #2947


brb:
LGTM, just waiting for tests to finish.

PR #3141, opened 2017-10-08
Upgrading client go and vendoring it's extra dependencies.

bjhaid:
I ran:

`go build -o weave-npc -a github.com/weaveworks/weave/prog/weave-npc`

and

`make tests`

to confirm this change doesn't break things, please advice if there other things I need to do. I also chose HEAD of origin/master on all the dependencies as at the time I did pull them.


bboreham:
Thanks for the contribution!

PR #3149, opened 2017-10-23
Fix 2797 on Kubernetes - Cluster goes down because all IPs become unreachable

### LINKED ISSUE #2797: Remove deleted k8s nodes from Weave Net

bricef:
PR to fix #2797. Includes changes from #3022.

### DIFF
--- a/prog/kube-peers/annotations.go
+++ b/prog/kube-peers/annotations.go
@@ -23,17 +23,17 @@ import (
 )
 
 type configMapAnnotations struct {
-	Name      string
-	Namespace string
-	Client    corev1client.ConfigMapsGetter
-	cm        *v1.ConfigMap
+	ConfigMapName string
+	Namespace     string
+	Client        corev1client.ConfigMapsGetter
+	cm            *v1.ConfigMap
 }
 
-func newConfigMapAnnotations(ns string, name string, client *kubernetes.Clientset) *configMapAnnotations {
+func newConfigMapAnnotations(ns string, configMapName string, clientset *kubernetes.Clientset) *configMapAnnotations {
 	return &configMapAnnotations{
-		Namespace: ns,
-		Name:      name,
-		Client:    client.CoreV1(),
+		Namespace:     ns,
+		ConfigMapName: configMapName,
+		Client:        clientset.CoreV1(),
 	}
 }
 
@@ -83,14 +83,14 @@ func (cml *configMapAnnotations) Init() error {
 		// so that if the configmap is created after our GET but before or CREATE, we'll gracefully
 		// re-try to get the configmap.
 		var err error
-		cml.cm, err = cml.Client.ConfigMaps(cml.Namespace).Get(cml.Name, api.GetOptions{})
+		cml.cm, err = cml.Client.ConfigMaps(cml.Namespace).Get(cml.ConfigMapName, api.GetOptions{})
 		if err != nil {
 			if !kubeErrors.IsNotFound(err) {
-				return errors.Wrapf(err, "Unable to fetch ConfigMap %s/%s", cml.Namespace, cml.Name)
+				return errors.Wrapf(err, "Unable to fetch ConfigMap %s/%s", cml.Namespace, cml.ConfigMapName)
 			}
 			cml.cm, err = cml.Client.ConfigMaps(cml.Namespace).Create(&v1.ConfigMap{
 				ObjectMeta: api.ObjectMeta{
-					Name:      cml.Name,
+					Name:      cml.ConfigMapName,
 					Namespace: cml.Namespace,
 				},
 			})
@@ -98,7 +98,7 @@ func (cml *configMapAnnotations) Init() error {
 				if kubeErrors.IsAlreadyExists(err) {
 					continue
 				}
-				return errors.Wrapf(err, "Unable to create ConfigMap %s/%s", cml.Namespace, cml.Name)
+				return errors.Wrapf(err, "Unable to create ConfigMap %s/%s", cml.Namespace, cml.ConfigMapName)
 			}
 		}
 		break
--- a/prog/kube-peers/annotations.go
+++ b/prog/kube-peers/annotations.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+
+	v1 "k8s.io/api/core/v1"
+	kubeErrors "k8s.io/apimachinery/pkg/api/errors"
+	api "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubernetes "k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type configMapAnnotations struct {
+	Name      string
+	Namespace string
+	Client    corev1client.ConfigMapsGetter
+	cm        *v1.ConfigMap
+}
+
+func newConfigMapAnnotations(ns string, name string, client *kubernetes.Clientset) *configMapAnnotations {
+	return &configMapAnnotations{
+		Namespace: ns,
+		Name:      name,
+		Client:    client.CoreV1(),
+	}
+}
+
+type peerList struct {
+	Peers []peerInfo
+}
+
+type peerInfo struct {
+	PeerName string // Weave internal unique ID
+	Name     string // Kubernetes node name
+}
+
+func (pl peerList) contains(peerName string) bool {
+	for _, peer := range pl.Peers {
+		if peer.PeerName == peerName {
+			return true
+		}
+	}
+	return false
+}
+
+func (pl *peerList) add(peerName string, name string) {
+	pl.Peers = append(pl.Peers, peerInfo{PeerName: peerName, Name: name})
+}
+
+const (
+	// KubePeersAnnotationKey is the default annotation key
+	KubePeersAnnotationKey = "kube-peers.weave.works/peers"
+)
+
+func (cml *configMapAnnotations) Init() error {
+	for {
+		// Since it's potentially racy to GET, then CREATE if not found, we wrap in a check loop
+		// so that if the configmap is created after our GET but before or CREATE, we'll gracefully
+		// re-try to get the configmap.
+		var err error
+		cml.cm, err = cml.Client.ConfigMaps(cml.Namespace).Get(cml.Name, api.GetOptions{})
+		if err != nil {
+			if !kubeErrors.IsNotFound(err) {
+				return errors.Wrapf(err, "Unable to fetch ConfigMap %s/%s", cml.Namespace, cml.Name)
+			}
+			cml.cm, err = cml.Client.ConfigMaps(cml.Namespace).Create(&v1.ConfigMap{
+				ObjectMeta: api.ObjectMeta{
+					Name:      cml.Name,
+					Namespace: cml.Namespace,
+				},
+			})
+			if err != nil {
+				if kubeErrors.IsAlreadyExists(err) {
+					continue
+				}
+				return errors.Wrapf(err, "Unable to create ConfigMap %s/%s", cml.Namespace, cml.Name)
+			}
+		}
+		break
+	}
+	if cml.cm.Annotations == nil {
+		cml.cm.Annotations = make(map[string]string)
+	}
+	return nil
+}
+
+func (cml *configMapAnnotations) GetPeerList() (*peerList, error) {
+	var record peerList
+	if cml.cm == nil {
+		return nil, errors.New("endpoint not initialized, call Init first")
+	}
+	if recordBytes, found := cml.cm.Annotations[KubePeersAnnotationKey]; found {
+		if err := json.Unmarshal([]byte(recordBytes), &record); err != nil {
+			return nil, err
+		}
+	}
+	return &record, nil
+}
+
+// Update will update and existing annotation on a given resource.
+func (cml *configMapAnnotations) UpdatePeerList(list peerList) error {
+	if cml.cm == nil {
+		return errors.New("endpoint not initialized, call Init first")
+	}
+	recordBytes, err := json.Marshal(list)
+	if err != nil {
+		return err
+	}
+	cml.cm.Annotations[KubePeersAnnotationKey] = string(recordBytes)
+	cml.cm, err = cml.Client.ConfigMaps(cml.Namespace).Update(cml.cm)
+	return err
+}
--- a/prog/kube-peers/annotations.go
+++ b/prog/kube-peers/annotations.go
@@ -1,3 +1,10 @@
+/*
+In order to keep track of active weave peers, we use annotations on the Kubernetes cluster.
+
+Kubernetes uses etcd to distribute and synchronise these annotations so we don't have to.
+
+This module deals with operations on the peerlist backed by Kubernetes' annotation mechanism.
+*/
 package main
 
 import (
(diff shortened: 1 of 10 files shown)


bricef:
Not just a location change for the API, but also a functional change. Will need to refactor code to use new API methods.

prog/kube-peers/main.go
```diff
@@ -41,20 +44,157 @@ func getKubePeers() ([]string, error) {
...
+	if err != nil {
+		log.Fatalf("Could not make Kubernetes connection: %v", err)
+	}
+	peers, err := getKubePeers(c)
 	if err != nil {
 		log.Fatalf("Could not get peers: %v", err)
 	}
-	for _, addr := range peers {
-		fmt.Println(addr)
+	if justReclaim {
```
brb: Perhaps we could move this branch to a separate function to make the `main` function smaller.

prog/kube-peers/annotations.go
```diff
@@ -0,0 +1,191 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	kubeErrors "k8s.io/apimachinery/pkg/api/errors"
+	api "k8s.io/apimachinery/pkg/apis/meta/v1"
+	wait "k8s.io/apimachinery/pkg/util/wait"
+	kubernetes "k8s.io/client-go/kubernetes"
+)
+
+type configMapAnnotations struct {
+	Name      string
```
brb: Perhaps a short clarification would be good here: is it a name of the annotation or a name of the ConfigMap?

prog/kube-peers/annotations.go
```diff
@@ -0,0 +1,191 @@
...
+	DefaultLeaseDuration = 5 * time.Second
+	retryPeriod          = time.Second * 2
+	jitterFactor         = 1.0
+
+	// KubePeersAnnotationKey is the default annotation key
+	KubePeersAnnotationKey = "kube-peers.weave.works/peers"
+)
+
+func (cml *configMapAnnotations) Init() error {
+	for { // Loop only if we call Create() and it's already there
```
brb: Why?

prog/kube-peers/annotations.go
```diff
@@ -0,0 +1,191 @@
...
+func (cml *configMapAnnotations) LoopUpdate(f func() error) error {
+	stop := make(chan struct{})
+	var err error
+	wait.JitterUntil(func() {
+		if err = cml.Init(); err != nil {
+			close(stop)
+			return
+		}
+		err = f()
+		if err != nil && kubeErrors.IsConflict(err) {
```
brb: Out of curiosity, how does it detect conflicts?

prog/kube-peers/main.go
```diff
@@ -41,20 +44,157 @@ func getKubePeers() ([]string, error) {
...
+					return cml.RemoveAnnotation(peer.PeerName)
+
+				})
+				common.Log.Debugln("Finished removal of ", peer.PeerName)
+			}
+			// 8.   If step 5 failed due to optimistic lock conflict, stop: someone else is handling X
+
+			// Step 3-5 is to protect against two simultaneous rmpeers of X
+			// Step 4 is to pick up again after a restart between step 5 and step 7b
+			// If the peer doing the reclaim disappears between steps 5 and 7a, then someone will clean it up in step 7aa
```
brb: Could we add some log statements (level=info) to be able in the future to detect such situations from user reports?

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -47,6 +47,44 @@ items:
       - kind: ServiceAccount
         name: weave-net
         namespace: kube-system
+  - apiVersion: rbac.authorization.k8s.io/v1beta1
```
brb: You probably want to update `prog/weave-kube/weave-daemonset-k8s-1.7.yaml` as well. Also, LG (Launch Generator).

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -47,6 +47,44 @@ items:
...
+    kind: Role
+    metadata:
+      name: weave-net2
```
mikebryant: Can this be renamed to `weave-net`?

mikebryant: (Also needs changes to the labels, and the rolebinding should be renamed, and reference the new name)

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
mikebryant: Should `prog/weave-kube/weave-daemonset-k8s-1.7.yaml` and `prog/weave-kube/weave-daemonset.yaml` also be updated?

prog/kube-peers/annotations.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (cml *configMapAnnotations) UpdatePeerList(list peerList) error {
+	recordBytes, err := json.Marshal(list)
+	if err != nil {
+		return err
+	}
+	return cml.UpdateAnnotation(KubePeersAnnotationKey, string(recordBytes))
+}
+
+func (cml *configMapAnnotations) GetAnnotation(key string) (string, bool) {
```
brb: `if cml.cm == nil { ... }`?

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
bricef: @mikebryant I'm not sure this can be changed, as it would clash with the `weave-net`  `ClusterRole` which grants different permissions at a different scope.

mikebryant: Why would it clash? They're two different types of resource, and names are only unique within a resource.

e.g.
```
(devtools/kube-system) ~ $ k get clusterrole mike
NAME      AGE
mike      9s
(devtools/kube-system) ~ $ k get role mike
NAME      AGE
mike      8s
```

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
bricef: @mikebryant `prog/weave-kube/weave-daemonset-k8s-1.7.yaml` seems like a good bet, and I've added that to the PR. 

The other one, however, seems to be a config file that covers v1.5, which didn't support the RBAC permission system (it used ABAC) and was permissive by default.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
bricef: 👍 I stand corrected. Changing now.

prog/kube-peers/annotations.go
bricef: 👍 Race condition between GET/CREATE, so retry on failure if another peer creates the configmap before we do. Added comment to clarify.

prog/kube-peers/annotations.go
bricef: 👍 Good catch. Turns out entire method is dead code as well, so that solves that.

prog/kube-peers/annotations.go
bricef: @brb The error includes detailed info to indicates it's a conflict. The update function will throw back the error from the k8s API. The `IsConflict()` method is just sugar to look inside the error.

prog/kube-peers/main.go
bricef: 👍 Added log statements as info so we keep track and can see in the log when this edge case occurs.

test/860_weave_removes_dead_nodes_from_ipam_3_test.sh
```diff
@@ -0,0 +1,122 @@
+#! /bin/bash
```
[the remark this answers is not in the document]
bricef: Agreed. Want to merge separately though and I'll refactor into a common code base later.

skny5:
Any thoughts on when this will be merged into a release? This is an extremely critical piece of functionality when running K8S on a dynamic infrastructure, weave becomes pretty much unusable when an error (#2797) occurs.

bboreham:
@skny5 right now we're trying to clear up all the points and give it another thorough review; hopefully days rather than weeks.
As you may understand it is a rather destructive operation if we get it wrong.

bricef:
@skny5 Working on it as we speak. Hoping this will pass review and hit mainline this week.

prog/kube-peers/annotations.go
brb: I meant how k8s internally detects conflicts AKA do we know that our approach is correct ;)

bboreham: The pattern is taken from [Kubernetes leader-election code](https://github.com/kubernetes/contrib/blob/master/election/lib/election.go#L71)

I'm pretty sure the conflict response is passed through from etcd.

bboreham:
Some notes from my testing:

I fired up a 3-node Kubernetes cluster using the test 840 script.
Checked logs from all three weave containers.
Checked `weave status ipam`
Deleted a node with `kubectl delete node brya-2`
Initially the weave network still had three peers, but then the eviction manager kicked in on brya-2 and deleted all pods, killing all containers including Weave.

Next I restarted one of the remaining weave-net pods, so it would notice the node had gone; this all appeared to go to plan:
```
INFO: 2017/11/08 12:14:09.239399 Added myself to peer list &{[{76:a1:dd:6c:12:51 brya-0} {0e:e4:4e:dd:92:ca brya-2} {ba:23:5a:0c:f3:10 brya-1}]}
DEBU: 2017/11/08 12:14:09.239596 Nodes that have disappeared: map[brya-2:{0e:e4:4e:dd:92:ca brya-2}]
DEBU: 2017/11/08 12:14:09.239643 Preparing to remove disappeared peer {0e:e4:4e:dd:92:ca brya-2}
DEBU: 2017/11/08 12:14:09.239665 Noting I plan to remove  0e:e4:4e:dd:92:ca
DEBU: 2017/11/08 12:14:09.245074 Nodes that have disappeared: map[brya-2:{0e:e4:4e:dd:92:ca brya-2}]
DEBU: 2017/11/08 12:14:09.245149 Preparing to remove disappeared peer {0e:e4:4e:dd:92:ca brya-2}
DEBU: 2017/11/08 12:14:09.245171 Existing annotation 76:a1:dd:6c:12:51
DEBU: 2017/11/08 12:14:09.245191 weave DELETE to http://127.0.0.1:6784/peer/0e:e4:4e:dd:92:ca with map[]
INFO: 2017/11/08 12:14:09.252523 rmpeer of 0e:e4:4e:dd:92:ca : 393216 IPs taken over from 0e:e4:4e:dd:92:ca
```

I attempted to rejoin brya-2 using `kubeadm reset` then `kubeadm join [...]` 
This resulted in the peer picking up the previous `/var/lib/weave/weave-netdata.db` so now it cannot join the cluster because it has an incompatible update.  This was [previously noted](https://github.com/weaveworks/weave/pull/3022#issuecomment-316859271)

I manually deleted the persistence file then went through `kubectl delete node` `kubeadm reset` `kubeadm join` again.  Weave Net fired up ok.

slightly puzzling ipam state, presumably because it reclaimed the bridge address and nothing further:
```
# weave status ipam
76:a1:dd:6c:12:51(brya-0)               917503 IPs (87.5% of total) (1 active)
0e:e4:4e:dd:92:ca(brya-2)                    1 IPs (00.0% of total) 
ba:23:5a:0c:f3:10(brya-1)               131072 IPs (12.5% of total) 
```
After moving one of the `nettest` pods onto `brya-2` it is more reasonable:
```
# weave status ipam
76:a1:dd:6c:12:51(brya-0)               655359 IPs (62.5% of total) (1 active)
0e:e4:4e:dd:92:ca(brya-2)               262145 IPs (25.0% of total) 
ba:23:5a:0c:f3:10(brya-1)               131072 IPs (12.5% of total) 
```

Next I am going to abruptly shut down `brya-1` and see what happens.

bricef:
> I attempted to rejoin brya-2 using `kubeadm reset` then `kubeadm join [...]`
This resulted in the peer picking up the previous `/var/lib/weave/weave-netdata.db` so now it cannot join the cluster because it has an incompatible update. This was previously noted
>
> I manually deleted the persistence file then went through kubectl delete node kubeadm reset kubeadm join again. Weave Net fired up ok.

@bboreham I haven't encountered this. Running `kubeadm reset` then `kubeadm join ...` leaves me with a (eventually) clean IPAM table across the cluster. I'm not doing `kubectl delete` though. 

I am getting the following from somewhere when listening to the `k8s_weave_weave-net-...` container:
```
INFO: 2017/11/08 16:13:55.261671 Removed unreachable peer e6:78:31:8a:e6:e1(vagr-1)
```
Although apparently not from the weave-net codebase? (ran a search for "Removed" and "unreachable", and git couldn't find it!)
And I feel that at this point (or after some retries) that we should be removing the IP allocations for that peer?

This is the log I get when removing a peer with `kubeadm reset`

```
INFO: 2017/11/08 16:26:14.364090 ->[10.128.0.19:56777|e6:78:31:8a:e6:e1(vagr-1)]: connection shutting down due to error: read tcp4 10.128.0.20:6783->10.128.0.19:56777: read: connection reset by peer
INFO: 2017/11/08 16:26:14.364386 ->[10.128.0.19:56777|e6:78:31:8a:e6:e1(vagr-1)]: connection deleted
INFO: 2017/11/08 16:26:14.365670 Removed unreachable peer e6:78:31:8a:e6:e1(vagr-1)
```

But `sudo weave status ipam` still shows up unreachable addresses.

```
9a:33:6c:88:6b:5f(vagr-0)               393216 IPs (37.5% of total) (1 active)
2e:eb:78:2c:d6:d8(vagr-2)               131072 IPs (12.5% of total) 
e6:78:31:8a:e6:e1(vagr-1)               524288 IPs (50.0% of total) - unreachable!
```

Pretty sure I'm running your latest version. (I get the `[kube-peers]` tagged logs), but I never get the `Preparing to remove disappeared peer` message, and the 860 test is (correctly) failing.

bboreham:
`kubeadm reset` removes the k8s installation and stopps running processes, but doesn't delete the node from the api-server. So this PR doesn't reclaim those IPs.

bricef:
Ok, I'll amend to fix.

test/860_weave_removes_dead_nodes_from_ipam_3_test.sh
```diff
@@ -0,0 +1,118 @@
...
+#
+# Test functions
+#
+
+function check_no_lost_ip_addresses {
+    host=$1
+    ipam_status=$(run_on $host "sudo weave status ipam")
+    echo $ipam_status
+    unreachable_count=$(echo $ipam_status | grep "unreachable" | wc -l)
+    if [ "$unreachable_count" != "0" ]; then
```
bboreham: why not use `assert` here?

bboreham:
I think this is in fairly good shape; I am minded to merge it and raise new issues for follow-up.

One area we need to cover off is what happens when you upgrade from a previous release - the current code will delete persisted data which is ok if you do it one node at a time but not idea.

Also we need to update the cloud.weave.works config generator.

PR #3151, opened 2017-10-23
Implement Kubernetes 1.7 NetworkPolicy semantics

### LINKED ISSUE #3105: weave-npc does not support kubernetes 1.7 NetworkPolicy semantics
Opening per request for @bboreham in https://github.com/weaveworks/weave/issues/3083

In Kubernetes pre-1.7, the network policy semantics were as follows:

* In a namespace, by default, allow all connections into all pods, whether or not a policy is defined
* If a namespace has the `DefaultDeny` annotation, block all traffic to all pods, _except_ for those that have a `NetworkPolicy` allowing access to the specific pods (by label, etc.)

In Kubernetes 1.7+, the semantics are as follows:

* In a namespace, by default, allow all connections into all pods, if no policy is defined
* If a pod has a NetworkPolicy attached to it, then deny all connections into _that_ pod, except for those explicitly allowed by the policy
* If a namespace has the `DefaultDeny` annotation, block all traffic to all pods, _except_ for those that have a `NetworkPolicy` allowing access to the specific pods (by label, etc.) (unchanged)

The changes are in the middle one: If no annotation is set but a network policy applies to a pod, then block all connections to that pod _except_ for those explicitly allowed by the policy.

As pointed out by @bboreham, weave-npc currently implements the pre-1.7 semantics, rather than the 1.7+. weave-npc should implement the 1.7+ semantics (or possibly check the kubernetes version and implement the appropriate semantics).

brb:
Please see commit messages for explanation.

The implementation is based on newly introduced "default-allow" ipset (one per each Namespace) which contains IP addrs of Pods which are not selected (as dst) by any NetworkPolicy.

I considered [an alternative and a bit less complex approach (blacklisting)](https://github.com/weaveworks/weave/commit/abdaf805fa2c734dcdc3a2cac872b9adfe7ab7fd), but that requires quite a few changes to iptables chains and also requires changing default policy of `-o weave` FORWARD chain from `DROP` to `ACCEPT`.

For pre 1.7 k8s users, we need to pass the `--use-legacy-netpol` param in LG.
 
Fixes #3105


npc/namespace.go
```diff
@@ -355,10 +494,61 @@ func isDefaultDeny(namespace *coreapi.Namespace) bool {
...
+		uid = p.ObjectMeta.UID
+	default:
+		err = errInvalidNetworkPolicyObjType
+		return
+	}
+	ns.policies[uid] = obj
+
+	// Analyse policy, determine which rules and ipsets are required
+	if ns.legacy {
+		rules, nsSelectors, podSelectors, err = ns.analysePolicyLegacy(obj.(*extnapi.NetworkPolicy))
```
bboreham: Looks like this will crash if a v1 policy comes in in legacy mode, or vice-versa.  But maybe this cannot happen because of the way we subscribe?
Maybe this if-statement should be merged with the above switch, so it's clear that things can only happen one way?

bboreham:
Should update unit test `TestRegressionPolicyNamespaceOrdering3059` to use v1 instead of beta objects

npc/namespace.go
brb: This cannot happen due to the way we subscribe, as you said. 

Maybe it makes sense to cast netpol objects directly w/o the type switch (inside `if ns.legacy { ... } else { ...}`). That would make the runtime panic if invalid object is given, but it's fine as we are no longer able to progress in such situation. WDYT?

npc/analyser.go
```diff
@@ -89,26 +90,114 @@ func (ns *ns) analysePolicy(policy *extnapi.NetworkPolicy) (
...
+	if p != nil {
+		proto = string(*p)
+	}
+
+	return proto
+}
+
+func port(p *intstr.IntOrString) string {
+	// If no port is specified, match any port. Let iptables executable handle
+	// service name resolution
```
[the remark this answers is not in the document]
brb: It is actually a copy pasta from the previous code: https://github.com/weaveworks/weave/blob/master/npc/analyser.go#L100

brb:
PTAL.

PR #3156, opened 2017-10-30
document using v2 plugin from dockerhub rather than store

### LINKED ISSUE #3153: use v2 plugin from dockerhub rather than store
Apparently it's quite hard to update the v2 plugin in the docker store. Hence #3150.

We should probably change our docs to refer to the plugin on dockerhub instead, i.e.
```
docker plugin install weaveworks/net-plugin:latest_release
```
etc.

We really ought to fix #3152 first though.

rade:
Since the latter is hard to update and hence frequently out of date.

Fixes #3153.


abuehrle:
LGTM

PR #3159, opened 2017-11-01
Fix: ingress rule lost podsrchost when neither pod and namespace selector is nil

mainred:
when we have ingress rule like this, this means traffict from pods of either condition
will be accepted:
1. pods in the default namespace swith label "role=frontend"
2. pods in the namespace with label "project=myproject"
```
metadata:
  name: test-network-policy
  namespace: default
spec:
  podSelector:
    matchLabels:
      role: db
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          project: myproject
    - podSelector:
        matchLabels:
          role: frontend
```
But, weave-npc now only accepts NO.2 , and misses NO.1.
ref: https://kubernetes.io/docs/concepts/services-networking/network-policies/

### DIFF
--- a/npc/analyser.go
+++ b/npc/analyser.go
@@ -54,32 +54,43 @@ func (ns *ns) analysePolicy(policy *extnapi.NetworkPolicy) (
 			// From is present and contains at least on item, this rule allows traffic only if the
 			// traffic matches at least one item in the from list.
 			for _, peer := range ingressRule.From {
-				var srcSelector *selectorSpec
+				var srcSelectors []*selectorSpec
 				if peer.PodSelector != nil {
-					srcSelector, err = newSelectorSpec(peer.PodSelector, ns.name, ipset.HashIP)
+					srcPodSelector, err := newSelectorSpec(peer.PodSelector, ns.name, ipset.HashIP)
 					if err != nil {
 						return nil, nil, nil, err
 					}
-					podSelectors[srcSelector.key] = srcSelector
+					podSelectors[srcPodSelector.key] = srcPodSelector
+					srcSelectors = append(srcSelectors, srcPodSelector)
 				}
 				if peer.NamespaceSelector != nil {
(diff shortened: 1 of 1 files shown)


bboreham:
Thanks for the contribution!

Let me see if I understand: the code will currently work for _either_ a pod selector or a namespace selector, and you've fixed it to work in the case the policy has _both_.  

I pushed this branch to the main repo so the full test suite will run.

Would probably put this on the 2.0 branch since it's a small change and a real bug.

mainred:
Yes, you got the point, bboreham.

brb:
@mainred Thanks for the contribution, but I don't see why the change is necessary. The current code [does iterate over peers]( https://github.com/weaveworks/weave/blob/master/npc/analyser.go#L56) (in your given policy there are two peers), and for the policy creates two rules:

```
-A WEAVE-NPC-INGRESS -m set --match-set weave-%:FTKw;6Z~kKUkeo5F)k8K+kI src -m set --match-set weave-2ZiDW)3]+I}!oR#ru=*KAc{x1 dst -m comment --comment "namespaces: selector: project=myproject -> pods: namespace: default, selector: role=db" -j ACCEPT
-A WEAVE-NPC-INGRESS -m set --match-set weave-UbymzRD2/QjeNF/Kib4(cV*{P src -m set --match-set weave-2ZiDW)3]+I}!oR#ru=*KAc{x1 dst -m comment --comment "pods: namespace: default, selector: role=frontend -> pods: namespace: default, selector: role=db" -j ACCEPT
```

Just noting, that if you omit the dash before the second `podSelector`, it would make the policy invalid as _"NetworkPolicyPeer describes a peer to allow traffic from. Exactly one of its fields must be specified."_ (src: https://kubernetes.io/docs/api-reference/v1.8/#networkpolicypeer-v1-networking)

mainred:
@brb ，you are right, and thank you very much for your correcting my mistake. Maybe more testing should be made before I open another pr :)    I'll close this pr.

PR #3160, opened 2017-11-02
Update gopacket to latest master to benefit from memory optimization

dlespiau:
Upstream has done some work saving 133k allocated objects on startup (init()),
which translates into a saving of 30-50MB RSS memory.

Details: https://github.com/google/gopacket/pull/377#issuecomment-338984309


bboreham:
Do you think this makes a difference even when `gopacket` isn't in use? 
We typically use [go-odp](https://github.com/weaveworks/go-odp) rather than `gopacket`.

dlespiau:
If the `github.com/google/gopacket/layers` package is imported and not used otherwise, the problematic  `init()` function is still executed and those objects allocated. On a memory profile it's about 18MB of objects hanging from an `init()` function, but haven't checked myself on the net daemon(s), just noticed today gopacket was used here as well.

bboreham:
Just running `weave launch`
Before:
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     17390  0.9  1.0 445908 62204 ?        Ssl  10:31   0:00 /home/weave/weaver [...]
```

After:
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     19503  0.5  0.7 364488 46736 ?        Ssl  10:35   0:00 /home/weave/weaver [...]
```

A 15MB drop, not bad at all.  Thanks!

PR #3172, opened 2017-11-15
weave-kube reclaim: check for upgrade case

### LINKED ISSUE #3170: Improve behaviour of kube peer deletion on upgrade

bboreham:
Set a flag in the persistence file, so we know whether we should expect to find our name in the peer list.

Fixes #3170

### DIFF
--- a/prog/weaveutil/db_flag.go
+++ b/prog/weaveutil/db_flag.go
@@ -27,7 +27,7 @@ func getDBFlag(args []string) error {
 		return err
 	}
 	if !nameFound {
-		return fmt.Errorf("Flag %q not found", flagName)
+		return fmt.Errorf("")
 	}
 	fmt.Print(value)
 	return nil
--- a/prog/weaveutil/db_flag.go
+++ b/prog/weaveutil/db_flag.go
@@ -1,0 +1,54 @@
+// Get/set arbitrary flags in the Weave Net persistence DB
+package main
+
+import (
+	"fmt"
+
+	"github.com/weaveworks/weave/db"
+)
+
+const dbFlagPrefix = "flag:"
+
+func getDBFlag(args []string) error {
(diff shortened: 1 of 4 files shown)


prog/weaveutil/db_flag.go
```diff
@@ -0,0 +1,54 @@
...
+		return err
+	}
+	defer d.Close()
+	var value string
+	nameFound, err := d.Load(dbFlagPrefix+flagName, &value)
+	if err != nil {
+		return err
+	}
+	if !nameFound {
+		return fmt.Errorf("")
```
bricef: Returning an empty error to signal we haven't found something seems unintuitive. Would be much clearer to have `getDBFlag` return the value instead of printing it out.

prog/weaveutil/db_flag.go
```diff
@@ -0,0 +1,54 @@
...
+	value := args[2]
+
+	d, err := db.NewBoltDB(dbPrefix)
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+	err = d.Save(dbFlagPrefix+flagName, value)
+	if err != nil {
+		return err
```
bricef: Surprised go linter is allowing this, usually complains about tail nil error checking.

prog/weaveutil/db_flag.go
bboreham: Printing the value is the pattern of all functions in `weaveutil`; I doubt it's worth breaking that pattern for this small point.

The empty error is along the lines of `if value, ok := foo[bar]; ok`

PR #3173, opened 2017-11-16
Rearrange peer reclaim checks in weave-kube launch script

bboreham:
If the persistence file doesn't exist (and it doesn't on first install), there's no point doing all the other checks to see if we should delete it.

And I was worried about what would happen if a user installs the new image with the old manifest, without adding the additional annotation permissions;  added `|| true` to make sure we continue past the `kube-peers reclaim` on error.  

Also, in that error case, it would report `Added myself to peer list <nil>`, so I fixed that too.

### DIFF
--- a/prog/kube-peers/main.go
+++ b/prog/kube-peers/main.go
@@ -79,7 +79,6 @@ func addMyselfToPeerList(cml *configMapAnnotations, c *kubernetes.Clientset, pee
 		}
 		return nil
 	})
-	common.Log.Infoln("[kube-peers] Added myself to peer list", list)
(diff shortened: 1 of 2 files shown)


bboreham:
> Could we not end up with some peers that are up and running, have IPs allocated, but aren't in the peerlist?

Yes, this is exactly the situation you will be in if you are running Weave Net 2.0.x

So if you upgrade and something goes wrong, no reclaiming will be done, but it won't break networking.  If you read the log file you will find out why no reclaiming is going on.  I think this is preferable to breaking networking.

Issue #3177, opened 2017-11-17
New NetworkPolicy implementation fails if pod updates come out of order


bboreham:
Code added in #3151 bypasses reference-counting on ipset membership.
Thus if we get the following sequence:

```
DEBU: 2017/11/17 16:00:25.485583 EVENT UpdatePod {"metadata":{"name":"users-697cd95b9b-7nqhf","generateName":"users-697cd95b9b-","namespace":"default",...
INFO: 2017/11/17 16:00:25.485636 adding entry 10.32.0.7 to weave-local-pods
INFO: 2017/11/17 16:00:25.486942 adding entry 10.32.0.7 to weave-k?Z;25^M}|1s7P3|H9i;*;MhG
INFO: 2017/11/17 16:00:25.488099 adding entry 10.32.0.7 to weave-E.1.0W^NGSp]0_t5WwH/]gX@L

DEBU: 2017/11/17 16:36:23.486668 EVENT UpdatePod {"metadata":{"name":"users-697cd95b9b-ngww4","generateName":"users-697cd95b9b-","namespace":"default",...
INFO: 2017/11/17 16:36:23.486714 adding entry 10.32.0.7 to weave-local-pods
INFO: 2017/11/17 16:36:23.486729 adding entry 10.32.0.7 to weave-k?Z;25^M}|1s7P3|H9i;*;MhG

DEBU: 2017/11/17 16:36:23.501206 EVENT UpdatePod {"metadata":{"name":"users-697cd95b9b-7nqhf","generateName":"users-697cd95b9b-","namespace":"default",..."state":{"terminated":{"exitCode":0,"reason...
INFO: 2017/11/17 16:36:23.501285 deleting entry 10.32.0.7 from weave-local-pods
INFO: 2017/11/17 16:36:23.501299 deleting entry 10.32.0.7 from weave-E.1.0W^NGSp]0_t5WwH/]gX@L
INFO: 2017/11/17 16:36:23.502343 deleting entry 10.32.0.7 from weave-k?Z;25^M}|1s7P3|H9i;*;MhG
```

traffic is now blocked:

```
WARN: 2017/11/17 16:40:41.430346 TCP connection from 10.32.0.10:50902 to 10.32.0.7:80 blocked by Weave NPC.
WARN: 2017/11/17 16:40:42.430273 TCP connection from 10.32.0.10:50902 to 10.32.0.7:80 blocked by Weave NPC.
[...]

PR #3178, opened 2017-11-17
Reference-count all ipset entries

bboreham:
Speculative change in view of #3177 
I'll try this out and see if I can figure out why it was done the other way.


bboreham:
Replaced by #3181

PR #3179, opened 2017-11-19
Allow the log level of the v2 plugin to be configured

lalinsky:


bboreham:
Thanks for the contribution!

I'm just running the full CI suite.

brb:
You probably want to add `WEAVE_LOG_LEVEL` to https://github.com/weaveworks/weave/blob/master/prog/net-plugin/config.json#L52 to make the param configurable via `docker plugin set <..>`.

lalinsky:
Hm, now I'm not sure if it's a good idea to do this. I will add it, but perhaps the Docker plugin configuration file shouldn't be overloaded with variables like this?

Btw, was it intentionally set to debug level?

brb:
> Hm, now I'm not sure if it's a good idea to do this. I will add it, but perhaps the Docker plugin configuration file shouldn't be overloaded with variables like this?

Maybe I'm missing something, but how would you configure the log level param if it hasn't been exposed as a Docker plugin param?

> Btw, was it intentionally set to debug level?

The plugin is quite new, so just for easier debugging of user problems.

bboreham:
Another version has been proposed: #3197 - did you have a specific reason for closing this one @lalinsky ?

PR #3183, opened 2017-11-22
Trim the JSON logged for Kubernetes objects by weave-npc

bboreham:
Log size of weave-npc was getting out of hand.  We currently run at debug level by default, which includes a full JSON dump of every Kubernetes objects received.

Remove things like "args" and "env" which bulk the logs, aren't relevant to the function of weave-npc, and may include sensitive information.

Use a black-list rather than a white-list because we can't be sure what information will be necessary in future.

### DIFF
--- a/npc/json.go
+++ b/npc/json.go
@@ -4,7 +4,56 @@ import (
 	"encoding/json"
 )
 
+// Return JSON suitable for logging an API object.
 func js(v interface{}) string {
+	// Get the raw JSON
 	a, _ := json.Marshal(v)
+	// Convert this back into a tree of key-value maps
+	var m map[string]interface{}
+	if err := json.Unmarshal(a, &m); err != nil {
+		// If that didn't work, just return the raw version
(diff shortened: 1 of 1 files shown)


npc/json.go
```diff
@@ -4,7 +4,56 @@ import (
 	"encoding/json"
 )
 
+// Return JSON suitable for logging an API object.
 func js(v interface{}) string {
+	// Get the raw JSON
 	a, _ := json.Marshal(v)
+	// Convert this back into a tree of key-value maps
+	var m map[string]interface{}
+	if err := json.Unmarshal(a, &m); err != nil {
```
brb: Is it possible to avoid `json.Marshal` and `json.Unmarshal`, and just directly cast `v`?

bboreham: No; `v` is expected to be a strongly-typed Go object, like a `Pod`

brb:
LGTM

PR #3185, opened 2017-11-22
Note that WeaveDNS is disabled in Kubernetes addon

bboreham:
We didn't make this clear.


bboreham:
I added it to the add-on page and also updated a few other items there.  Add-on page probably needs an overhaul but not today.

bboreham:
In that second doc the example policies are lacking a `apiVersion: networking.k8s.io/v1` line.  Maybe they were shortened to fit on a slide originally.

PR #3188, opened 2017-11-27
Put a more sane log-level as default for weave-npc

roffe:
The debug value prints out ALOT in the log file and i'm sure most users will get a chock the first time they open it not expecting debug to be on by default


bboreham:
This is really "do you want anyone to be able to debug what went wrong"?
We need the `debug` info to do that.
#3183 will cut it down by 50-60%.

Still, why `warning` and not `info` ?

roffe:
@bboreham you are right. it should be "info", mental fart on my side. 

I welcome #3183 but still believe that DEBUG should be something toggled by the user when needed. 

My main concern was that it was printing env variables and args etc in the logs but as i can see from #3183 that a lot of this will be removed to trim the size.

PR #3189, opened 2017-11-27
Fix release script for plugin latest_release

### LINKED ISSUE #3150: v2 docker plugin in docker store is ancient
Following the [weave net v2 docker plugin instructions](https://www.weave.works/docs/net/latest/install/plugin/plugin-v2/), installs a version that is from May 2017 - running `weave version` on it reports [`git-1da1943df9d8`](https://github.com/weaveworks/weave/commit/1da194).

bboreham:
We should never update the `latest_release` plugin on a `PRERELEASE` (meaning something not tagged like 2.1.4 or flagged as prerelease on GitHub)

We should always update it on a `MAINLINE` or `BRANCH` release (meaning the version name matches a pattern like 2.3.0 or 2.1.9)

Fixes #3150

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -196,7 +196,7 @@ publish() {
 
     if [ "$RELEASE_TYPE" = 'PRERELEASE' ] ; then
         echo "== Building and pushing plugin"
-        make UPDATE_LATEST=$IS_MAINLINE SUDO=$SUDO WEAVE_VERSION=$VERSION DOCKERHUB_USER=$DOCKERHUB_USER plugin_publish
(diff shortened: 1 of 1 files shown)


bboreham:
@rade there are two uses of `IS_MAINLINE` on that line.
One is for `weavedb`: it isn't ever necessary to update `weavedb` so this is some sort of kludge to ensure it got pushed once.

The other is important: `:latest` images should match the `master`git branch, so we don't want to update it on a change to a patch branch.

bboreham:
Although we could just not update `:latest` images at all from the release script, since CI should update them.

PR #3204, opened 2017-12-18
Do not skip bridge creation if bridge exists

### LINKED ISSUE #3133: Do something more useful when the weave bridge is DOWN
Encountered during an instance of #2998 - nothing was working on one node because its bridge was DOWN.

Not sure if the best thing to do is to make this more clear to the administrator, or try to mend the bridge so it can be UP.

Is it always a bug? Are there reasons for the machine owner to deliberately set it DOWN?

brb:
I recommend to review each commit separately.

This PR does a few things:

- Removes `create_bridge` from the weave script. As a consequence, `weave attach $CIDR $CONTAINER` is no longer possible without running weaver.
- Removes unnecessary `create_bridge` from `weave expose` and `weave hide`.
- Makes `net/bridge.go:EnsureBridge` do not skip creation of the bridge if such already exists, and also makes the function idempotent.

Part of the #3133 fix; going to add netlink subscribe for changes to the bridge in a separate PR.

### DIFF
--- a/common/odp/odp.go
+++ b/common/odp/odp.go
@@ -75,6 +75,14 @@ func DeleteDatapath(dpname string) error {
 }
 
 func AddDatapathInterface(dpname string, ifname string) error {
+	return addDatapathInterface(dpname, ifname, false)
+}
+
+func AddDatapathInterfaceIfNotExist(dpname string, ifname string) error {
+	return addDatapathInterface(dpname, ifname, true)
+}
+
+func addDatapathInterface(dpname, ifname string, ignoreIfExist bool) error {
 	dpif, err := odp.NewDpif()
 	if err != nil {
 		return err
@@ -87,5 +95,9 @@ func AddDatapathInterface(dpname string, ifname string) error {
 	}
 
 	_, err = dp.CreateVport(odp.NewNetdevVportSpec(ifname))
+	if ignoreIfExist && err == odp.NetlinkError(syscall.EEXIST) {
+		return nil
+	}
+
 	return err
 }
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -411,7 +411,7 @@ func configureIPTables(config *BridgeConfig) error {
 	}
 	if config.DockerBridgeName != "" {
 		if config.WeaveBridgeName != config.DockerBridgeName {
-			// This check is not ideal, as the rule should be at the very top
+			// This is not ideal, as it does not check whether the rule is at the top
 			// of the chain.
 			found, err := ipt.Exists("filter", "FORWARD", "-i", config.DockerBridgeName, "-o", config.WeaveBridgeName, "-j", "DROP")
 			if err != nil {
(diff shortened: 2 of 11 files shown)


deitch:
As commented on the other issue, thank you very much for this. The “reboot bug” bit us more than once.

deitch:
On my phone, so hard to read, but does this also ensure iptables correctness?

brb:
> On my phone, so hard to read, but does this also ensure iptables correctness?

Yep, it should.

If you want to do an early testing, please let me know - I could build images for you.

deitch:
I’d be happy to test. We have an infra sandbox where we setup and tear down clusters all the time (automated through terraform) so more than happy to run tests there if you can provide an image.

brb:
@deitch Cool, thanks! You can start using the fixed version by replacing `weaveworks/weave-{kube,npc}:latest` with `weaveworks/weave-{kube,npc}:git-a7ea71bd` in https://github.com/weaveworks/weave/blob/master/prog/weave-kube/weave-daemonset-k8s-1.7.yaml

brb:
PTAL

deitch:
Will do @brb, thank you.

deitch:
1. upgraded `weave` to the specific image, 
2. First test: manually recreated the problem: `weave` bridge down and `WEAVE` rules removed from iptables. Killed pod, let `daemonset` restore it. Everything came back perfectly.
3. Second test: Same as first, but reboot. Everything came back perfectly.

Looks good to me! Please do tag version here when it is merged in and released.

brb:
@deitch Thanks! I'm waiting for someone to review the PR.

bboreham:
I have read through the code, and that all seems fine.
I would like to know more about what it will do under different failure conditions, e.g. if some of the iptables rules are present and some are absent.

deitch:
> I have read through the code, and that all seems fine

I certainly am excited for this. It makes weave far more resilient. Instead of "works from nothing" and "works when it already works", it becomes, "always ensure good state on startup".

brb:
@bboreham It will restore iptables rules. The filter/FORWARD rules are restored in a correct order. Also, I've extended the 191 integration test to test restoration of iptables rules.

PTAL - two new commits since your review.

bboreham:
I like the table of rules - in future we should be able to use the same table to remove them on `weave reset`, and take that code out of the shell script.

deitch:
Fabulous! What version will it (did it?) go into?

bboreham:
2.2; I'd like to get #3210 in too

deitch:
Thanks @bboreham. Is there an ETA?

deitch:
Nice #3210 is merged too. So 2.2?

brb:
We are experiencing some integration test failures, so we would like to fix those before doing 2.2.

deitch:
> We are experiencing some integration test failures, so we would like to fix those before doing 2.2.

Thanks @brb. Not that I have _any_ cycles, but we have gotten bit by this one several times now. If there is anything we can do to help move it along, let us know?

bboreham:
Weave Net release 2.2 just out https://github.com/weaveworks/weave/releases/tag/v2.2.0

deitch:
Really great! Thanks to all!

Does it automatically get pushed to docker hub?

bboreham:
Yes, it's there on Docker Hub now

deitch:
Ah, got it. By default, it only shows scanned tags, not unscanned.

PR #3208, opened 2017-12-28
weave-kube: add WEAVE_SECRET environment variable

aconchillo:
Allows creating a password (e.g. through kubernetes secrets) and pass that to weaver through WEAVE_SECRET.

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -67,6 +67,13 @@ if [ -n "$HOSTNAME" ] ; then
     NICKNAME_ARG="--nickname=$HOSTNAME"
 fi
 
+# Router password. You can create a Kubernetes secret and specify that as the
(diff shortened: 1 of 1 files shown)


bboreham:
Hi @aconchillo, and thanks for the contribution!

We always kept the password out of command-line options because they show up in `ps` and other places.  And there is already a `WEAVE_PASSWORD` environment variable which `weaver` will read at startup.

We [document](https://www.weave.works/docs/net/latest/kubernetes/kube-addon/#configuration-options) how a Kubernetes secret can be used to set this environment variable, which seems like a safer option.

bboreham:
I'm going to close this unless there is more to consider.

aconchillo:
Thank you @bboreham. Keeping the password out of the arguments makes total sense. I knew about WEAVE_PASSWORD for the weave script, not for weaver, great! And thanks for the link as well. I'll close the issue.

PR #3210, opened 2018-01-02
Move weave-npc checks to top of FORWARD chain

### LINKED ISSUE #3209: weave allows through all traffic to a virtual service IP when NetworkPolicy is set to deny all ingress

Block all the traffic if the network policy is set to do so

## What happened?

Allows through all traffic to a virtual service IP

## How to reproduce it?

Create a new namespace and NetworkPolicy to deny all ingress:

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: default-deny
spec:
  podSelector: {}
  policyTypes:
  - Ingress
```
Deploy two nginx pods and create a service for it: NodePort, LoadBalancer, ClusterIP. If you run curl on that service IP you would still be able to hit the nodes even if the `NetworkPolicy` is set to deny all ingress. If you hit the pods from a different pod in the same namespace or different namespace it behaves as it should and the traffic is blocked as expected.

This specific situation was tested on AWS using Kops.

## Versions:

```
$ weave version
weave 2.1.3

$ docker version

Client:
 Version:      1.13.1
 API version:  1.26
 Go version:   go1.7.5
 Git commit:   092cba3
 Built:        Wed Feb  8 06:36:34 2017
 OS/Arch:      linux/amd64

Server:
 Version:      1.13.1
 API version:  1.26 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   092cba3
 Built:        Wed Feb  8 06:36:34 2017
 OS/Arch:      linux/amd64
 Experimental: false

$ uname -a
Linux ip-10-88-32-23 4.4.102-k8s #1 SMP Sun Nov 26 23:32:43 UTC 2017 x86_64 GNU/Linux

$ kubectl version
Server Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.5", GitCommit:"cce11c6a185279d037023e02ac5249e14daa22bf", GitTreeState:"clean", BuildDate:"2017-12-07T16:05:18Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
```

This seems to be happening because of https://github.com/kubernetes/kubernetes/pull/56524 
This rule `[24:1440] -A KUBE-FORWARD -m comment --comment "kubernetes forwarding rules" -m mark --mark 0x4000/0x4000 -j ACCEPT` which is above the WEAVE-NET rules might cause the issue

bboreham:
We want our DROP rule to be executed ahead of any ACCEPT rules which might have been added by another program on the host.

Since #3204 gave us the framework to ensure we are recreating every rule, just inserting them all at the top, in order, works nicely.

~I also moved the NFLOG and DROP rules from FORWARD to WEAVE-NPC to simplify the logic about which order they come in.~

Extend kubernetes test to include the circumstances that trigger the issue.

Fixes #3209

### DIFF
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -504,7 +504,7 @@ func linkSetUpByName(linkName string) error {
 // ensureRules ensures the presence of given iptables rules.
 //
 // If any rule from the list is missing, the function deletes all given
-// rules and re-appends them to ensure the order of the rules.
+// rules and re-inserts them to ensure the order of the rules.
 func ensureRules(table, chain string, rulespecs [][]string, ipt *iptables.IPTables) error {
 	allFound := true
 
@@ -524,13 +524,13 @@ func ensureRules(table, chain string, rulespecs [][]string, ipt *iptables.IPTabl
 		return nil
 	}
 
-	for _, rs := range rulespecs {
+	for pos, rs := range rulespecs {
 		// If any is missing, then delete all, as we need to preserve the order of
 		// given rules. Ignore errors, as rule might not exist.
 		if !allFound {
 			ipt.Delete(table, chain, rs...)
 		}
-		if err := ipt.Append(table, chain, rs...); err != nil {
+		if err := ipt.Insert(table, chain, pos+1, rs...); err != nil {
 			return errors.Wrapf(err, "ipt.Append(%s, %s, %s)", table, chain, rs)
 		}
 	}
--- a/test/840_weave_kube_3_test.sh
+++ b/test/840_weave_kube_3_test.sh
@@ -134,6 +134,22 @@ spec:
     run: nettest
 EOF
 
+# And a NodePort service so we can test virtual IP access
+run_on $HOST1 "$KUBECTL create -f -" <<EOF
+apiVersion: v1
+kind: Service
+metadata:
+  name: netvirt
+spec:
+  type: NodePort
+  ports:
+  - port: 80
+    targetPort: 8080
+    nodePort: 31138
+  selector:
+    run: nettest
+EOF
+
 podName=$($SSH $HOST1 "$KUBECTL get pods -l run=nettest -o go-template='{{(index .items 0).metadata.name}}'")
 
 check_all_pods_communicate() {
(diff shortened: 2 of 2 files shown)


bboreham:
https://circleci.com/gh/weaveworks/weave/9969 is a demonstration of the problem
https://circleci.com/gh/weaveworks/weave/9968 is the same test but without the two factors that trigger it - Kubernetes 1.8.5 and `--cluster-cidr` argument to kube-proxy

brb:
I like the idea of the chain simplification, but as discussed on Slack, moving `-j DROP` to the `WEAVE-NPC` chain creates a time window for allowing a traffic which should not be allowed by NetworkPolicy, because the weave-npc process clears the chain each time it starts.

However, we have another time window due to `-A WEAVE-NPC -m set ! --match-set weave-local-pods dst -j ACCEPT`. But this can be fixed by replacing the rule with `-A WEAVE-NPC -m set --match-set weave-local-pods src ! --match-set weave-local-pods dst -j ACCEPT`.

I suggest to create another level of indirection - a new chain WEAVE-NPC-FOOBAR which would contain `-j WEAVE-NPC`, "nflog" and `-j DROP` rules, and insert `-j WEAVE-NPC-FOOBAR` at the very top of "filter/FORWARD -o weave".

bboreham:
> But this can be fixed by replacing the rule with `-A WEAVE-NPC -m set --match-set weave-local-pods src ! --match-set weave-local-pods dst -j ACCEPT`.

The idea of this rule was to allow non-pod traffic; that change would disallow non-pod traffic originating on this machine.  Maybe that's a nice case.

net/bridge.go
```diff
@@ -439,13 +439,8 @@ func configureIPTables(config *BridgeConfig) error {
...
 		// If WEAVE-NPC chain doesn't exist then next line will fail
-		if err = ipt.AppendUnique("filter", "FORWARD", "-o", config.WeaveBridgeName, "-j", "WEAVE-NPC"); err != nil {
-			return err
-		}
-		if err = ipt.AppendUnique("filter", "FORWARD", "-o", config.WeaveBridgeName, "-m", "state", "--state", "NEW", "-j", "NFLOG", "--nflog-group", "86"); err != nil {
-			return err
-		}
-		if err = ipt.AppendUnique("filter", "FORWARD", "-o", config.WeaveBridgeName, "-j", "DROP"); err != nil {
+		// Use insert rather than append to try to get ahead of any ACCEPT rules which would defeat our policies
+		if err = insertUnique(ipt, "filter", "FORWARD", 1, "-o", config.WeaveBridgeName, "-j", "WEAVE-NPC"); err != nil {
```
brb: Maybe we could add a comment in the rule saying "NOTE: this must be before -j KUBE-FORWARD rule"

brb:
This needs rebasing after #3204 has been merged. Happy to do if you are busy.

bboreham:
Rebased and simplified.

PR #3224, opened 2018-01-20
Fix subnet isolation and bump build-tools to latest

brb:
* The "-i docker0 -o weave -j DROP" filter/FORWARD rule used to be inserted after the "-j WEAVE-EXPOSE" rule, so the subnet isolation has been broken for exposed subnets. This was the reason for the 130 smoke test failure.
* Add comment to note the importance of the order of the "-j WEAVE-NPC" rule.
* Improve naming of `ensureRules` function.
* ~Update k8s to 1.9.2 in CI.~
* Update build tools fix flaky tests due to the defunct process.


brb:
PTAL.

If you mind bumping k8s and build-tools to be part of this PR, I can create a separate one.

brb:
Updated.

Issue #3168, opened 2017-11-13
Support ipBlock for ingress rules of NetworkPolicy in weave-npc


brb:
k8s 1.8 extended `NetworkPolicyPeer` object by introducing `ipBlock`: https://kubernetes.io/docs/api-reference/v1.8/#networkpolicypeer-v1-networking

As it is stated in the linked doc, _"IPBlock describes a particular CIDR (Ex. "192.168.1.1/24") that is allowed to the pods matched by a NetworkPolicySpec's podSelector."_

kitt1987:
Any progress on this feature?

brb:
Sorry, no progress yet. PRs are very welcome!

kitt1987:
Could you please help me review this PR?

PR #3233, opened 2018-01-31
Add ipBlock support for ingress rules in weave-npc

kitt1987:
Implement https://github.com/weaveworks/weave/issues/3168

~~A new iptables rule will be generated for each NetworkPolicy with IPBlock peer just like others.  All rules will share the same ipset if NetworkPolicies except same CIDRs/IPs from IPBlocks.~~
Since external traffic is masqueraded by rules written by kube-proxy, ipblock based network policies can't be implemented by adding single rules in WEAVE-NPC-INGRESS chain like other policies.
IPblock based policies have to be applied before forwarding and destination nodes need to trust packets which src IP is one of weave bridges'. Rules below are added to Filter table.
```bash
-N WEAVE-NPC-INGRESS-IPBLOCK
-A WEAVE-NPC-INGRESS -m set --match-set weave-bridges src -m comment --comment "packets forwarded from weave bridges are safe" -j ACCEPT
-A WEAVE-NPC-INGRESS -j WEAVE-NPC-INGRESS-IPBLOCK
```
And, a new rule will be added to chain `WEAVE-NPC-INGRESS-IPBLOCK` when a new policy is created. Such as,

```bash
-A WEAVE-NPC-INGRESS-IPBLOCK -s 10.22.224.4/32 -m set --match-set weave-LxTLCYBhd(0L*JS+fQ1Yt/IV7 dst -m comment --comment "cidr: 10.22.224.4/32 -> pods: namespace: ffca51bb048a34684bfb905120f663c1f, selector: " -j ACCEPT
```

All IPs exposed on weave bridges are saved in ipset `weave-bridges` . It is important to keep `weave-bridges` up to date when nodes added or deleted, even when IP changed. I think the best way is to watch router changes, and get peer bridge IPs from ipam or router, then modify the ipset. But, this is too difficult to me because it causes too many changes in weave core modules. I choose a simpler way to achieve this, that is to save IPs in configmap `weave-net` after they are exposed by `kube-peers`. **weave-npc** will watch all pods created by DaemonSet `weave-net` and update the ipset using bridge IPs saved in configmap `weave-net`.

However,  packets send from nodes to pods will bypass all policies. The following rules have to be added to Filter table to prevent these packets(`10.20.0.0/17` is the pod CIDR).

```bash
-N WEAVE-NPC-LOCAL-INGRESS
-A WEAVE-NPC-LOCAL-INGRESS -j WEAVE-NPC-DEFAULT
-A WEAVE-NPC-LOCAL-INGRESS -j WEAVE-NPC-INGRESS-IPBLOCK
-A WEAVE-NPC-LOCAL-INGRESS -j NFLOG --nflog-group 86
-A WEAVE-NPC-LOCAL-INGRESS -j DROP
-A OUTPUT -d 10.20.0.0/17 -m state --state NEW -j WEAVE-NPC-LOCAL-INGRESS
```

If packets explicitly permitted by NetworkPolicy are only accepted, LivenessProbes and ReadinessProbes with type HTTPGet or TCPSocket will not work. We can filter and accept packets sent from kubelet by matching `--pid-owner`. It also needs to watch pid of kubelet to keep the rule up to date when kubelet restarted. I'm not very sure whether it is a right way since it is complicated and not very safe. Instead of adding a pid based rule and watching kubelet, an option `trust-weave-bridges` is provided for `weave-npc` . Chain `WEAVE-NPC-LOCAL-INGRESS` will keep empty if `trust-weave-bridges` is set to true. To keep all probes working, this option should be true.

### DIFF
--- a/npc/ipblock.go
+++ b/npc/ipblock.go
@@ -1,0 +1,121 @@
+package npc
+
+import (
+	"fmt"
+	"github.com/weaveworks/weave/common"
+	"github.com/weaveworks/weave/npc/ipset"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"net"
+	"sort"
+	"strings"
+)
+
+type ipBlock struct {
+	ipsetName ipset.Name // generated ipset name
+	spec      *networkingv1.IPBlock
+}
+
+func newIPBlock(ipb *networkingv1.IPBlock, ns string) (b *ipBlock, err error) {
+	if _, _, err = net.ParseCIDR(ipb.CIDR); err != nil {
+		return
+	}
+
+	for _, ex := range ipb.Except {
+		if _, _, err = net.ParseCIDR(ex); err != nil {
+			return
+		}
+	}
+
+	b = &ipBlock{spec: ipb}
+
+	if len(ipb.Except) > 0 {
+		sortedExcept := ipb.Except
+		sort.Strings(sortedExcept)
+		b.ipsetName = ipset.Name(IpsetNamePrefix + shortName("except:"+strings.Join(sortedExcept, "")))
+	}
+
+	return
+}
+
+func (ipb ipBlock) GetRuleSpec() (spec *ruleSpec) {
+	spec = &ruleSpec{
+		chain: IngressIPBlockChain,
+	}
+
+	if len(ipb.spec.Except) == 0 {
+		spec.args = append(spec.args, "-s", ipb.spec.CIDR)
+		spec.comment = fmt.Sprintf("cidr: %s", ipb.spec.CIDR)
+	} else {
+		spec.args = append(spec.args, "-s", ipb.spec.CIDR, "-m", "set", "!", "--match-set",
+			string(ipb.ipsetName), "src")
+		spec.comment = fmt.Sprintf("cidr: %s except [%s]", ipb.spec.CIDR,
+			strings.Join(ipb.spec.Except, `,`))
+	}
+	return
+}
+
+type ipBlockSet struct {
+	ips   ipset.Interface
+	users map[string]int
+}
+
+func newIPBlockSet(ips ipset.Interface) *ipBlockSet {
+	return &ipBlockSet{
+		ips:   ips,
+		users: make(map[string]int),
+	}
+}
+
+func (s *ipBlockSet) deprovision(uid types.UID, current *ipBlock) (err error) {
+	if current == nil || len(current.ipsetName) == 0 {
+		return
+	}
+
+	key := string(current.ipsetName)
+	if _, found := s.users[key]; !found {
+		return
+	}
+
+	s.users[key]--
+	if s.users[key] == 0 {
+		if err = s.ips.Destroy(current.ipsetName); err != nil {
+			return
+		}
+
+		delete(s.users, key)
+	}
+
+	return
+}
+
+func (s *ipBlockSet) provision(uid types.UID, desired *ipBlock) (err error) {
+	if desired == nil || len(desired.ipsetName) == 0 {
+		return
+	}
+
+	key := string(desired.ipsetName)
+	if _, found := s.users[key]; found {
+		common.Log.Debug()
+		s.users[key]++
+		return
+	}
+
+	err = s.ips.Create(desired.ipsetName, ipset.HashNet)
+	if err != nil {
+		return
+	}
+
+	comment := `excepted from ` + desired.spec.CIDR
+	for _, excepted := range desired.spec.Except {
+		if err = s.ips.AddEntry(uid, desired.ipsetName, excepted, comment); err != nil {
+			if details := s.ips.Destroy(desired.ipsetName); details != nil {
+				common.Log.Warn("can't destroy ipset ", desired.ipsetName, ":", details)
+			}
+			return
+		}
+	}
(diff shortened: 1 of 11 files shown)


bboreham:
Thanks for the contribution!
I find it odd that so many new package dependencies are required: do you know why that is ?

kitt1987:
Seem they are required by the upgraded kubernetes dependencies.

bboreham:
OK, that's rather an overhead for this change, but I guess we'll have to live with it.

Linter is complaining: `run gofmt -s -w npc/selector.go`

bboreham:
Can you give us some idea how you tested it?  Ideally write a test (e.g. extend [this one](https://github.com/weaveworks/weave/blob/master/test/840_weave_kube_3_test.sh)) but I understand those tests need a lot of infrastructure set-up so we can probably do that.

kitt1987:
OK. I'll fix the lint failure and write more tests. Thanks.

npc/ipblock.go
```diff
@@ -0,0 +1,129 @@
...
+			return
+		}
+	}
+
+	b = &ipBlock{spec: ipb}
+
+	if len(ipb.Except) > 0 {
+		sortedExcept := ipb.Except
+		sort.Strings(sortedExcept)
+		b.ipsetName = ipset.Name(IpsetNamePrefix + shortName("except:"+strings.Join(sortedExcept, "")))
```
brb: What is a value of `ipsetName` when `ipb.Except` is empty?

kitt1987: It is empty because no ipset will be created if `ipb.Except` is empty. I've refactored some code and will update this PR along with more test.

kitt1987:
I tested it between 2 k8s clusters. All traffic is not masqueraded because I exposed IPs in  10.20.0.0/16 for each weave bridge and assign IPs in 10.20.0.0/17 and 10.20.128.0/17 for pods in different clusters.  It is my scenario. I'm so sorry that I didn't test in the scenario you mentioned.

kitt1987:
I'm trying to fix my code.

kitt1987:
@brb I found that is difficult to apply IP block policies to external traffic. The way what I found is to apply all policies with IP block before forwarding for both external traffic and traffic from hosts in the cluster. So, packets received with any weave bridge IP as src is safe and we can trust weave bridges in FORWARD chain. What do you think about this solution?

brb:
> packets received with any weave bridge IP as src is safe and we can trust weave bridges in FORWARD chain

Unfortunately, that would allow us to access protected services from any node.

However, we are currently considering to remove [the MASQ rule for external traffic](https://github.com/weaveworks/weave/blob/ce9c5162bd94018ad5caca2c914cb5780727147c/net/expose.go#L93), so the original src IP would be kept. Would it help for your implementation?

kitt1987:
```
-A OUTPUT -d IP_RANGE -m state --state NEW -j WEAVE-NPC-LOCAL-INGRESS
-A WEAVE-NPC-LOCAL-INGRESS -j WEAVE-NPC-DEFAULT
-A WEAVE-NPC-LOCAL-INGRESS -j WEAVE-NPC-INGRESS-IPBLOCK
-A WEAVE-NPC-LOCAL-INGRESS -j NFLOG --nflog-group 86
-A WEAVE-NPC-LOCAL-INGRESS -j DROP
```
Thanks for the reply.
I think above rules can keep traffic from nodes safe.

> However, we are currently considering to remove the MASQ rule for external traffic, so the original src IP would be kept. Would it help for your implementation?

When accessing a service IP, external traffic is masqueraded by rules owned by kube-proxy. It is not enough to just change weave.

brb:
Sorry for taking some time to review. I'm currently trying to better understand all implications of the ipBlock implementation to weave-npc. Shortly, I will review your PR.

Meanwhile, have you tried to implement some tests (ideally - integration ones) for the feature?

kitt1987:
Sorry for the late reply since Chinese new year holiday. I will write a brief to explain how I implement this feature as well as tests in scenarios below.

- kube-proxy masquerade all traffic
- within/without excepted CIDRs
- internal/external traffic
- within/without protocol
- Pod on same/different hosts
- NodePort
- port proxy
- for legacy k8s
- ipalloc-range is not defined
- for Probes

mrdima:
- You can prevent the kube-proxy snat if you add spec":{"externalTrafficPolicy":"Local"} to a nodePort service definition
- You can prevent (as a test for now) the weave snat using my script, see https://gist.github.com/mrdima/e0098308f2351e90132c7cf0e4fa460c

brb:
@kitt1987 Could you squash your commits, so that `npc/ipblock.go` would be in a separate commit? I'd like to cherry-pick it for egress policies. Thanks.

kitt1987:
@brb I've squashed them.

brb:
@kitt1987 Thanks again for your contribution. I've applied some changes from your patch - https://github.com/weaveworks/weave/commit/9769c8d5b905808e0363f42506c4af19c275708e.

However, the tracking of bridge IP addrs and the proposed chains introduce a big complexity, which for the majority of use cases is unnecessary. Also, the _"packets forwarded from weave bridges are safe"_ constraint is not compliant with the k8s netpol spec.

Therefore, I'm closing this PR in favor of https://github.com/weaveworks/weave/pull/3367 which treats ipBlock as any other selector.

PR #3235, opened 2018-01-31
Wait 5 seconds to let pod connect before rolling next pod

bboreham:
During a rolling update we don't want Kubernetes to go as fast as it can across nodes, because each node will have a short interruption as iptables rules are recreated.

I think this will also have the effect of slowing down a first install, but hopefully people can wait a few extra seconds.

I picked 5 seconds somewhat arbitrarily; the setting may need more experimentation.

Note we don't specify a readiness check at the moment, so Kubernetes equates "running" with "ready".  If the pod crashes immediately it will not be ready.

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.7.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.7.yaml
@@ -101,6 +101,8 @@ items:
         name: weave-net
       namespace: kube-system
     spec:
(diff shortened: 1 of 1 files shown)


mikebryant:
I don't think it'll slow down an initial install, when we deploy weave for the first time, all pods are created immediately - it's only when it's doing a rolling update does it start caring that they're up for long enough before it destroys the next pod

PR #3251, opened 2018-03-01
Switch from emdebian to ubuntu package repository for cross-compiling…

### LINKED ISSUE #3245: Fail to build weave on OSX and using vagrant
Weave can't be built on OSX or with vagrant. There is some cross-compilation dependences broken. One of your engineer said it could come from the Dockerfile.

You can reproduce it by building on OSX with the following step:
- git clone project
- git submodule update --init --recursive
- make

OR using vagrant:
- git clone project
- git submodule update --init -- recursive
- vagrant up
- vagrant ssh
- cd src/github.com/weaveworks/weave
- make

img_4667

It seems that you rarely use OSX and vagrant. But it could be nice to build with in a dev environment without cross-compilation dependencies.
I propose to manage that by allowing developer to build without cross-compilation.
I can do it.

bboreham:
Fixes #3245 

Evidently emdebian is not something we should be using.

Note the cross-compile tools are installed after other apt installation so only the cross-compile is affected.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -21,16 +21,6 @@ ENV GCC_CROSSCOMPILERS \
 	aarch64-linux-gnu \
 	powerpc64le-linux-gnu
 
-# Temporarily add the Debian repositories, because we're going to install gcc cross-compilers from there
-# Install the build-essential and crossbuild-essential-ARCH packages
-RUN echo "deb http://emdebian.org/tools/debian/ jessie main" > /etc/apt/sources.list.d/cgocrosscompiling.list \
-  && curl http://emdebian.org/tools/debian/emdebian-toolchain-archive.key | apt-key add - \
-  && for platform in ${DEB_CROSSPLATFORMS}; do dpkg --add-architecture $platform; done \
-  && apt-get update \
-  && apt-get install -y build-essential \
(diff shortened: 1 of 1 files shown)


bboreham:
CI failed:

```
cffi.ffiplatform.VerificationError: importing '/usr/lib/python2.7/dist-packages/cryptography/_Cryptography_cffi_813c10e0x7adb75f8.x86_64-linux-gnu.so': /usr/lib/python2.7/dist-packages/cryptography/_Cryptography_cffi_813c10e0x7adb75f8.x86_64-linux-gnu.so: symbol SSLv2_client_method, version OPENSSL_1.0.0 not defined in file libssl.so.1.0.0 with link time reference
```

marccarre:
This error seems to be why we started to use emdebian -- see: #2940

bboreham:
cc @leth

leth:
One of the crossbuild install commands upgrades to libssl from ubuntu xenial, where the rest of the golang base image comes from debian jessie.

I can't find any context as to why these compilers need to come from ubuntu xenial, does anyone know?

bboreham:
If there's a debian source that still works, that would be fine.
Or we could use an ububtu-based golang image

bboreham:
Closed in favour of #3253

PR #3266, opened 2018-03-28
Update tests to run against Kubernetes 1.10

bboreham:
Had to fix up one place where it compares `1.6` against `1.10` and gets the wrong answer.


brb:
Shall we change the base to 2.2?

bboreham:
I was thinking we can’t run the build on the 2.2 branch any more because of the package issues. But maybe it would use a cached build container?

brb:
Building on the 2.2 seems to be working: https://circleci.com/gh/weaveworks/weave/10164

bboreham:
I changed the base to 2.2

bboreham:
and the build failed

PR #3268, opened 2018-03-29
Convert vendoring from git submodules to copy, using 'dep'

bboreham:
Initially because I didn't want to figure out the compatibility matrix by hand, although I had to do quite a lot of figuring anyway.

See individual commits for details; one notable change is that we now run `manifest-tool` from an image.


brb:
Nice work!

Mind adding a small README.md (similar to the legacy `vendor/README.md`) with info how to add, update, remove deps with dep?

bboreham:
I added the instructions to CONTRIBUTING.md since `dep` will delete anything you add in `vendor` and I couldn't think of a better place to put it.  CONTRIBUTING.md also points to the wiki, but we can't update the wiki with a PR.

brb:
Thanks. However, I don't see the instructions, have you pushed them?

brb:
If you rebase against origin/master, the last commit ("Update to Go 1.10.1") won't be needed.

Besides this nitpick - LGTM!

bboreham:
I guess we should take out the 404 work-around too, since Google fixed that.

PR #3269, opened 2018-03-30
Remove .ruby-version which we don't need

bboreham:
Just checking what happens when we remove this.  I think it was there for some reason relating to the website, but also I think the website build just copies the doc files out of this repo now.


bboreham:
OK, I don't know how to do this.

PR #3282, opened 2018-04-23
Add src and dst labels to weavenpc_blocked_connections_total metric

ApsOps:
- Useful for getting the source of blocked requests

Note: Not adding sport since it is random for every request so
it gets hard to group the metrics

EDIT: Now that I think about it, sport doesn't really make it hard to group metrics. Just that there would be a lot more metrics with different labels.

### DIFF
--- a/npc/metrics/metrics.go
+++ b/npc/metrics/metrics.go
@@ -20,7 +20,7 @@ var (
 			Name: "weavenpc_blocked_connections_total",
 			Help: "Connection attempts blocked by policy controller.",
 		},
-		[]string{"protocol", "dport"},
+		[]string{"protocol", "src", "dst", "dport"},
 	)
 )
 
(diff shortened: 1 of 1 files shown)


bboreham:
Thanks!  We didn't do this originally because of the [recommendation](https://prometheus.io/docs/practices/naming/#labels) not to use labels with high cardinality.

I guess under typical circumstances you wouldn't get _that_ many different address pairs being blocked.  Also I'm not very sure how many counts as too many - in our [hosted Prometheus](https://github.com/weaveworks/cortex) we cap at 50,000, but I believe in regular Prometheus the relative cost is higher.

ApsOps:
Yeah I thought so too, so removed `sport`. I realized it gets hard to see what's making the connections being blocked, especially since multiple apps listen on the same ports.

We can of course look this info in weave-npc logs, but who doesn't love to see this in grafana (maybe even alert on it)?

Do we need to make it optional for users who might have cases where a lot of bad agents try to hit disallowed IPs?

bboreham:
Making it optional is one possible way forward.  I dislike extra switches, but it does seem this change would be hidden until the day someone mounts a DDOS attack and melts your Prometheus store at the same time.

ApsOps:
@bboreham What's your final take on this? Should I make any changes or is this good to be merged?

bboreham:
I think it can’t be merged because it has the potential to create millions of timeseries and hence break Prometheus.

ApsOps:
Makes sense, closing this.

For the record, when I asked around, I found out that it should be okay if the number is not very large. So it might work for some controlled environments where external agents can't try to hit random IPs.
> This is ok in Prometheus 2.x, because we time-partition our indexing.  So if you have a few (single digit) thousand new label value per hour, it's ok.

bboreham:
Making it optional and/or putting a cap on the number reported per `weaver` would both be plausible.

ApsOps:
Cool, I'll try that 👍

PR #3298, opened 2018-05-13
2924: Preserve the client source IP

### LINKED ISSUE #2924: Weave-Net on Kubernetes should support externalTrafficPolicy: Local
Hello,

I opened the issue #44963 on kubernetes/kubernetes, because the OnlyLocal annotation to preserve the external client IP seems not to work on (at least) Kubernetes 1.6.1 and 1.6.2. 

https://github.com/kubernetes/kubernetes/issues/44963

As we are using the weave-net addon @MrHohn from Kubernetes asked me to also file the issue here to maybe get some input. Is it sufficient to point to the kubernetes issue or should I copy the debug information and output here? Thank you.

brb:
This PR implements a mechanism for preserving the client source IP addr when accessing a Service annotated with `service.spec.externalTrafficPolicy=Local`.

The idea for preserving the source IP addr has been presented by https://github.com/weaveworks/weave/issues/2924#issuecomment-358915192.

If the `--no-masq-local` flag is passed to weaver, it will dynamically maintain the `weave-no-masq-local` ipset containing CIDRs allocated by IPAM for the particular weaver. The ipset is used by the iptables rule: `iptables -t nat -I WEAVE 1 -m set --match-set weave-no-masq-local dst -j RETURN`. The rule prevents external traffic to be MASQUERADEd before being sent to a Pod.

I suggest to read each commit separately when reviewing this PR.

Fixes #2924.


net/bridge.go
```diff
@@ -489,8 +493,25 @@ func configureIPTables(config *BridgeConfig) error {
 	if err := ipt.ClearChain("nat", "WEAVE"); err != nil {
 		return errors.Wrap(err, "failed to clear/create nat/WEAVE chain")
 	}
+	if err := ipt.AppendUnique("nat", "POSTROUTING", "-j", "WEAVE"); err != nil {
+		return err
+	}
 
-	return ipt.AppendUnique("nat", "POSTROUTING", "-j", "WEAVE")
+	// k8s-only: Create the ipset to store CIDRs allocated by IPAM for local pods.
```
bboreham: "k8s-only" doesn't sound right: this feature would be useful in any similar environment where we expect to route weave-addressed packets from (some) hosts

prog/weave-npc/main.go
```diff
@@ -56,7 +57,7 @@ func resetIPTables(ipt *iptables.IPTables) error {
 }
 
 func resetIPSets(ips ipset.Interface) error {
-	// Remove ipsets prefixed `weave-` only
+	// Remove ipsets prefixed `weave-` excluding weave-no-masq-local (used by weaver).
```
bboreham: should we just prefix it something different?  `weaver-no-masq-local` ?  Too subtle?

net/bridge.go
```diff
@@ -489,8 +493,25 @@ func configureIPTables(config *BridgeConfig) error {
...
+	// External traffic sent to these CIDRs avoids SNAT'ing so that NodePort
+	// with `"externalTrafficPolicy":"Local"` would have the correct src IP addr.
+	if config.NoMasqLocal {
+		ips := ipset.New(common.LogLogger())
+		_ = ips.Destroy(NoMasqLocalIpset)
+		if err := ips.Create(NoMasqLocalIpset, ipset.HashNet); err != nil {
+			return err
+		}
+		if err := ipt.Insert("nat", "WEAVE", 1, "-m", "set", "--match-set", string(NoMasqLocalIpset), "dst", "-j", "RETURN"); err != nil {
```
bboreham: In weave-npc we often add comments to rules to aid comprehension

ipam/http.go
```diff
@@ -199,6 +199,6 @@ func (alloc *Allocator) HandleHTTP(router *mux.Router, defaultSubnet address.CID
 	})
 
 	router.Methods("GET").Path("/ipinfo/tracker").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Fprintf(w, tracker)
+		fmt.Fprintf(w, alloc.tracker.String())
```
bboreham: What is this used for?  I can't find anything hitting this endpoint in the repo.

bboreham:
Just trying to understand this, without having been immersed in the details lately, I think the following explanation is useful:

Consider a packet arriving on node A from source S destined for pod P. 
If P is on node A, then the reply can be sent straight to S. (however it got here, Linux and/or the network must know how to get it back)
If P is on node B, then a reply from B to S will be unexpected and/or not work.
Masquerading the packet as coming from node A means the reply from P on B will come via A, which is good.
So, this PR attempts to non-masquerade all cases where P is on A while still masquerading the cases where P is on B.

@brb can you confirm?

net/bridge.go
```diff
@@ -514,6 +518,42 @@ func configureIPTables(config *BridgeConfig) error {
...
+		ips:   ipset.New(common.LogLogger()),
+		owner: types.UID(0), // dummy ipset owner
+	}
+}
+
+func (t *NoMasqLocalTracker) String() string {
+	return "no-masq-local"
+}
+
+func (t *NoMasqLocalTracker) HandleUpdate(prevRanges, currRanges []address.Range, local bool) error {
```
bboreham: I think this should check that `local` is true

net/bridge.go
```diff
@@ -514,6 +518,42 @@ func configureIPTables(config *BridgeConfig) error {
...
 }
 
+type NoMasqLocalTracker struct {
+	ips   ipset.Interface
+	owner types.UID
+}
+
+func NewNoMasqLocalTracker() *NoMasqLocalTracker {
+	return &NoMasqLocalTracker{
+		// TODO(mp) Reuse ipset object as each time calling New creates a test set.
```
bboreham: I don't quite follow: this is only called once from `main()` so why do we care?

test/images/network-tester/webserver.go
```diff
@@ -186,6 +202,7 @@ func main() {
 	http.HandleFunc("/read", state.serveRead)
 	http.HandleFunc("/write", state.serveWrite)
 	http.HandleFunc("/status", state.serveStatus)
+	http.HandleFunc("/client_ip", state.serveClientIP)
```
bboreham: I like this.

test/840_weave_kube_3_test.sh
```diff
@@ -244,6 +244,26 @@ assert_raises "$SSH $HOST1 $KUBECTL exec $denyPodName -- curl -s -S -f -m 2 http
 assert_raises "$SSH $HOST1 curl -s -S -f -m 2 http://$VIRTUAL_IP/status >/dev/null"
 assert_raises "$SSH $HOST1 curl -s -S -f -m 2 http://$HOST2:31138/status >/dev/null"
 
+# Passing --no-masq-local and setting externalTrafficPolicy to Local must preserve
+# the client source IP addr
+
+CLIENT_IP_MASQ="$($SSH $HOST1 curl -sS http://$HOST2:31138/client_ip)"
+
+WEAVE_ENV_VARS="${WEAVE_ENV_VARS}\\n                - name: NO_MASQ_LOCAL\\n                  value: \"1\""
```
bboreham: I would expect trouble here from dropping the `test.coverprofile` arg

brb: It contains the `test.coverprofile` arg. I just renamed `COVERAGE_ARGS` to `WEAVE_ENV_VARS`, as it might include not only coverage related args.

net/bridge.go
brb: Good catch!

ipam/http.go
brb: By [detect_awsvpc](https://github.com/weaveworks/weave/blob/master/weave#L900).

prog/weave-npc/main.go
brb: I didn't want to introduce a new prefix, but having the new prefix is probably better than the if clauses.

brb:
> @brb can you confirm?

ACK.

Additional useful info: as mentioned in the code, this works only on k8s, as in the vanilla Docker case, the Docker bridge acts as a default gw for a container, while in the k8s case - the weave bridge. So in the Docker case, if S sends a packet to P, a reply to it might be sent via different interface which can result in dropped reply due to `rp_filter`.

bboreham:
It might be worth commenting that "Kubernetes only" is a shorthand for "containers where the Weave Net interface is the default gateway", since the latter can be achieved using vanilla Docker and `weave attach`, or via some other orchestrator using the Weave Net CNI plugin.

bboreham:
Is this PR ready for final review now?

brb:
Thanks for the review. Yep, it's ready for final review. Two last commits are relevant ones.

bboreham:
Some discussion on Twitter about whether we should default this on. Thoughts?

rade:
> Some discussion on Twitter about whether we should default this on.

See #3389

brb:
> Some discussion on Twitter about whether we should default this on. Thoughts?

Could you link to the discussion?

We can make it default only for cases where the `weave` bridge is a default gw for containers (do we know any k8s deployment where it is not the case?).

Issue #3318, opened 2018-06-11
Weavenet pod crashes with minikube 


pgandhipfpt:
.
** Is this a FEATURE REQUEST? **
yes as its easy to create kubernetes cluster with minikube , so would be really helpful if we can set n/w policies on that cluster with weavenet

## What you expected to happen?
Weavnet pod comes up correctly on minikube after performing 
kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')" on kubenetes cluster
## What happened?
weavenet pod doesnot run . Its always in CrashLoopBackOff state

kubectl get pods --all-namespaces
NAMESPACE NAME READY STATUS RESTARTS AGE
kube-system etcd-minikube 1/1 Running 0 21m
kube-system kube-addon-manager-minikube 1/1 Running 0 21m
kube-system kube-apiserver-minikube 1/1 Running 0 21m
kube-system kube-controller-manager-minikube 1/1 Running 0 21m
kube-system kube-dns-86f4d74b45-md2m6 3/3 Running 0 22m
kube-system kube-proxy-9h8t5 1/1 Running 0 22m
kube-system kube-scheduler-minikube 1/1 Running 0 21m
kube-system kubernetes-dashboard-5498ccf677-kgflm 1/1 Running 0 22m
kube-system storage-provisioner 1/1 Running 0 22m
kube-system weave-net-nm7f2 1/2 CrashLoopBackOff 8 16m

## How to reproduce it?
a. I tried on macOS Sierra. 
b. checked out latest changes from https://github.com/kubernetes/minikube.git so that it has latest code from https://github.com/kubernetes/minikube/pull/2876
b. built the source code and got minikube.iso
c. ./out/minikube start
d.kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')" 

output:
serviceaccount "weave-net" created
clusterrole.rbac.authorization.k8s.io "weave-net" created
clusterrolebinding.rbac.authorization.k8s.io "weave-net" created
role.rbac.authorization.k8s.io "weave-net" created
rolebinding.rbac.authorization.k8s.io "weave-net" created
daemonset.extensions "weave-net" created

e.kubectl get pods --all-namespaces
NAMESPACE NAME READY STATUS RESTARTS AGE
kube-system etcd-minikube 1/1 Running 0 21m
[...]

PR #3319, opened 2018-06-12
Make error "Weave is not running" less confusing when running on Kubernetes

brb:
We suggest to run `weave reset` when weaver detects an existing bridge of different type than the requested one.

In case of Kubernetes, if a user runs `weave reset`, it returns `Weave is not running` which might make the user think that `weave reset` was not successful (happened in https://github.com/weaveworks/weave/issues/3318).

### DIFF
--- a/weave
+++ b/weave
@@ -1248,7 +1248,7 @@ launch() {
 stop() {
     util_op remove-plugin-network weave || true
     warn_if_stopping_proxy_in_env
-    util_op stop-container $CONTAINER_NAME >/dev/null 2>&1 || echo "Weave is not running." >&2
+    util_op stop-container $CONTAINER_NAME >/dev/null 2>&1 || echo "Weave is not running (ignore on Kubernetes)." >&2
     # In case user has upgraded from <v2.0 and left old containers running
     util_op stop-container $OLD_PLUGIN_CONTAINER_NAME >/dev/null 2>&1 || true
     util_op stop-container $OLD_PROXY_CONTAINER_NAME >/dev/null 2>&1 || true


bboreham:
Whilst this change may help, it isn't that hard to detect that it _is_ running.
Also the current instructions to run `weave status` in Kubernetes are very complicated. Maybe we should provide a container which "does the right thing" and can be run by `kubectl run` ?

brb:
Totally agree. However, creating such container will take us longer than changing this error message. So, as a temporary solution I suggested the change to the message.

brb:
Created an issue for your suggestion: https://github.com/weaveworks/weave/issues/3321

brb:
IMO, in case of k8s, we should improve the error message `Existing bridge type "bridge" is different than requested "bridged_fastdp". Please do 'weave reset' and try again` too, as it does not say whether a user needs to delete the weave-kube deployment and only afterwards to run the cmd.

hswong3i:
Could we make it commit so a 2.4.0 release?

bboreham:
@hswong3i I can't parse your comment.  Are you asking if you can submit a PR to solve this issue?  That would be most welcome.

hswong3i:
@bboreham sorry for my poor english, I am looking forward for a v2.4.0 release, so I could include it into kubespray, e.g. https://github.com/kubernetes-incubator/kubespray/pull/2991. Hopefully we could tidy up the remaining issue for v2.4.0?

bboreham:
@hswong3i ok, we plan to get #3313 into 2.4 which is a big change so taking a while to review.  Hopefully not too much longer.

Issue #3249, opened 2018-02-28
weave-net doesn't update the `NetworkUnavailable` node status on GCE


mikebryant:
## What you expected to happen?

For weave-net networking to work on GCE

## What happened?

weave-net runs on a node, but the node has a `NetworkUnavailable` status so no pods are scheduled

## Anything else we need to know?

GCE, custom

See also:
- https://github.com/kubernetes/kubernetes/issues/44254
- https://github.com/kubernetes/kubernetes/pull/33573
- https://github.com/kubernetes/kops/issues/2087

## Versions:

weave is 2.0.5

## Logs:
```
kubectl describe node includes:
  NetworkUnavailable   True    Mon, 01 Jan 0001 00:00:00 +0000   Wed, 28 Feb 2018 21:49:54 +0000   NoRouteCreated               Node created without a route
```

mikebryant:
Presumably weave-net or weave-kube should update the node status once weave comes up and is working?

mikebryant:
(Also related #2896)

bboreham:
Was that the resolved "correct" way to do it?  Is it documented anywhere?
There was a lot of discussion starting from https://github.com/kubernetes/kubernetes/pull/33438 but none of those PRs got merged.

I would think Weave should signal to kubelet which updates the node status. Pretty sure there isn't a way to do that.

mikebryant:
I'm looking in particular at comments like https://github.com/kubernetes/kubernetes/pull/33573#issuecomment-251987336

It looks to me like the current behaviour is that the gce controller sets this up and expects any network plugin in use to set it, which is why it doesn't work. 

And currently either the network plugin needs to set it, or you disable the gce cloud provider (which then means volumes etc don't work, so not really an option)

I'm not confident what the right overall solution should be, but it does seem clear that the current behaviour doesn't work.

mikebryant:
current code has a special case for gce in `providerRequiresNetworkingConfiguration`

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_network.go#L150-L160

This then gets used here to unconditionally set this to false

https://github.com/kubernetes/kubernetes/blob/b1ef631e5971f649effd51905444c2eeee58e8b1/pkg/kubelet/kubelet_node_status.go#L257

So not sure if it's in the docs anywhere, but it does seem clear that networking plugins setting this status is expected behaviour in the current tree

mikebryant:
[...]

PR #3334, opened 2018-06-21
Add permissions to set node status

### LINKED ISSUE #3332: weave-kube: Cannot patch nodes/status at the cluster scope
Just spotted on a local cluster:

```
FATA: 2018/06/17 08:04:55.395186 [kube-peers] could not set node status: nodes "host1" is forbidden: User "system:serviceaccount:kube-system:weave-net" cannot patch nodes/status at the cluster scope
```

Seems that the following is not set in the latest Weave Net DS and launch-generator:

```
- apiGroups:
  - ""
  resources:
  - nodes/status
  verbs:
  - patch
  - update
```

(Taken from https://github.com/weaveworks/weave/issues/3249#issuecomment-369653750)

bboreham:
Fixes #3332

This permission is needed to update the node NetworkUnavailable condition see https://github.com/weaveworks/weave/issues/3249

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.7.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.7.yaml
@@ -41,6 +41,13 @@ items:
           - get
           - list
           - watch
+      - apiGroups:
+        - ''
+        resources:
+        - nodes/status
+        verbs:
+        - patch
+        - update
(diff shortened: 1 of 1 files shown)


brb:
Thanks! I guess we need to fix launch-generator as well.

Going to merge after the build has passed.

bboreham:
I've done the LG change.

PR #3343, opened 2018-07-08
Add interface name to veth pair name

Mierdin:
## TL;DR

Adding a shortened network interface name to the naming scheme for veth pairs, to allow for multiple Weave networks off of the same container ID

## Explanation of Changes

I have been working with the [`multus`](https://github.com/intel/multus-cni/) plugin to get multiple weave networks connected to my Kubernetes pods. Other plugins seem to work, but for some reason [I was getting a strange error](https://github.com/intel/multus-cni/issues/91) when using multiple Weave networks:

```
could not create veth pair vethweplf979bd1-vethwepgf979bd1: file exists
```

After poking around the Weave code, I noticed that the creation of veth pair names is mostly static, with the exception of the container id (the `f979bd1` shown above in this case). This is great, unless the same container has multiple interfaces for which to create veth pairs. In this case, the first one will succeed, and any further ones will fail because they'll try to create another one with the same name, since both network interfaces are connected to the same container ID.

I started playing with ways to inject the interface name into the pair name as well, to allow for more uniqueness.

I like that the current naming scheme preserves 7 characters worth of container ID, and my first attempt was to preserve this by replacing some of the text in the prefix with the interface name:

```
   (old) - vethwepl0abcdef
   (new) - venet0pl0abcdef
```

While this preserves the number of container IDs, that's only if the interface name is reasonably short, and based on some of the references I looked up, there are a few other places in the code that rely on `VethName`. Doing this didn't seem to break anything for me, but I didn't do that much testing, so I assume this isn't quite the way to go.

The only other way I can think to go is to append the network interface before the "pg" or "pl", which will end up eating into the available characters for the container ID.

To help mitigate this, I added a new function that lets us predictably use the last N digits of the network interface name in this use case, so we can preserve as many digits for the container ID as possible. Keeping it to 3 or 4 should be reasonable, since if there is ever a conflict, Kubernetes for instance will keep trying to schedule it, and ideally the next container won't conflict - and this is already the case. However I am open to suggestions to alternatives, especially considering Kubernetes isn't the only use case.

### DIFF
--- a/net/veth.go
+++ b/net/veth.go
@@ -10,6 +10,7 @@ import (
 	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
+	"github.com/weaveworks/weave/common"
 )
 
 // create and attach a veth to the Weave bridge
@@ -119,11 +120,15 @@ func AttachContainer(netNSPath, id, ifName, bridgeName string, mtu int, withMult
 	defer ns.Close()
 
 	if !interfaceExistsInNamespace(netNSPath, ifName) {
-		maxIDLen := IFNAMSIZ - 1 - len(vethPrefix+"pl")
+		shortIfLength := 4
+
+		maxIDLen := IFNAMSIZ - 1 - len(vethPrefix+getShortIfName(ifName, shortIfLength)+"pl")
 		if len(id) > maxIDLen {
 			id = id[:maxIDLen] // trim passed ID if too long
 		}
-		name, peerName := vethPrefix+"pl"+id, vethPrefix+"pg"+id
+		name, peerName := vethPrefix+getShortIfName(ifName, shortIfLength)+"pl"+id, vethPrefix+getShortIfName(ifName, shortIfLength)+"pg"+id
+
+		common.Log.Infof(`Creating new veth pair: %s - %s`, name, peerName)
 		veth, err := CreateAndAttachVeth(name, peerName, bridgeName, mtu, keepTXOn, true, func(veth netlink.Link) error {
 			if err := netlink.LinkSetNsFd(veth, int(ns)); err != nil {
 				return fmt.Errorf("failed to move veth to container netns: %s", err)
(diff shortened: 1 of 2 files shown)


Mierdin:
Looks like CI failed, but I don't **think** it's a test failure, looks like more of a pipeline issue. Ping me here if not and I'll fix what I need to.

bboreham:
> multiple Weave networks

Are you actually achieving this? It looks like you are adding multiple interfaces to the same network.  It would be more efficient to add multiple IP addresses to the one interface.

Could you say a bit more about where this idea is headed?  Kubernetes only supports one network at present, and one address per pod.  Maybe open an issue describing the problem you are trying to solve.

Mierdin:
Sorry for the delayed response here. Let me elaborate on the use case.

We're building an application that deploys virtual network functions like firewalls and routers into Kubernetes as a testing and learning platform. We decided to start deploying these in containers because a) we already have some platforms available in this format and b) we're planning to spin up a lot of these and would prefer to keep things as small and efficient as possible. Combine that with the fact that we're already familiar with Kubernetes, and it makes a lot of sense to go this direction.

However, the kind of labs and topologies we want to spin up requires that each containerized network function is able to use multiple network interfaces, just like they would in a real network in the real world (trying to emulate this as much as possible so they don't realize they're logging into a kubernetes-hosted device. So, after looking at multus briefly, we got this working, with the exception of the aforementioned name collision problem as above. So we can't just use multiple addresses on the same interface, the pod has to have multiple interfaces. And, this works well with the version of weave we have deployed that's based on my branch here.

Hopefully that's more clear on our use case - I'm open to alternatives if the approach in this PR isn't acceptable. I'm sure there's something network namespace-related that I've overlooked in my haste to get through my other deliverables. I *believe* I also tried assigning a unique bridge for each network and still ran into this, but I could be wrong.

murali-reddy:
@Mierdin thanks for the elaboration.

While this fix helps to get multiple interfaces inside the pod/container, like @bboreham asked earlier, veth end of pairs still connect to same `weave` bridge? Were you able to actually connect interfaces to different networks?

bboreham:
@Mierdin still looking for clarification - are you saying this is for test scenarios where nobody cares that there is really just one network underneath?  That would be ok, but I don't want to merge this on a misunderstanding.

Mierdin:
Apologies all - this fell off my radar in the past few months for both personal and professional reasons, but I am still motivated to follow-up with some more information. I will be slammed for the next week or so but will revisit this as soon as practical with more info on what I'm doing. Thanks for the patience!

bboreham:
Closing this to tidy up - if you do come back to it I'm happy to re-open.

PR #3346, opened 2018-07-11
3206: Protect against CNI plugin being called with the host namespace

### LINKED ISSUE #3206: Protect against CNI plugin being called with the host namespace
A user had 4 Kubernetes nodes break down with this symptom:

```
$ ip route
default via 172.21.128.1 dev eth0
100.96.0.0/11 dev weave  proto kernel  scope link  src 100.101.4.0
100.96.0.0/11 dev eth0  proto kernel  scope link  src 100.101.4.194
172.17.0.0/16 dev docker0  proto kernel  scope link  src 172.17.0.1
172.21.128.0/19 dev eth0  proto kernel  scope link  src 172.21.137.4
```

The `weave` and `eth0` routes are overlapping, so nothing works properly and Weave Net refuses to start.

It seems plausible that kubelet called the CNI plugin with a namespace that was actually the root namespace, so we assigned an address to `eth0` in the root.

We could check this and refuse to do it.

murali-reddy:
adding a check to verify if network namespace is not host network namespace

Fixes #3206

### DIFF
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -117,6 +117,16 @@ func (c *CNIPlugin) CmdAdd(args *skel.CmdArgs) error {
 	}
 	defer ns.Close()
 
+	hostNs, err := netns.Get()
+	if err != nil {
+		return fmt.Errorf("error accessing host network namespace: %s", err)
+	}
+	defer hostNs.Close()
+
+	if ns.Equal(hostNs) {
+		return fmt.Errorf("can not specify host network namespace as network namespace to which container should be added")
+	}
+
 	id := args.ContainerID
 	if len(id) < 5 {
 		data := make([]byte, 5)


bboreham:
CI failed on two counts: one lint:
> plugin/net/cni.go:127:21: error strings should not be capitalized or end with punctuation or a newline

"missing secret" is because non-weaveworks repos don't have the credentials to create VMs to run the smoke tests.

plugin/net/cni.go
```diff
@@ -117,6 +117,16 @@ func (c *CNIPlugin) CmdAdd(args *skel.CmdArgs) error {
 	}
 	defer ns.Close()
 
+	hostNs, err := netns.Get()
+	if err != nil {
+		return fmt.Errorf("error accessing host network namespace")
```
bboreham: Best to include the netns error in the error returned.

rade:
What is happening with this?

bboreham:
I've pushed the branch to the `weaveworks/weave` repo so it will run fill smoke tests.

bboreham:
Thanks!

PR #3356, opened 2018-07-25
support for portmap as additional plugin with wave-net

### LINKED ISSUE #3016: ability to enable cni portmap plugin
Hi, 

Can you add the possibility of enable portmap plugin on all host  on kubernetes 

This will solve the hostport problems 

Regards :)

murali-reddy:
Fix to use `10-weave.conflist` instead of `10-weave.conf` as CNI config file.

Also adds support for `portmap` as additional plugin along with weave-net

Fixes #3016

### DIFF
--- a/site/kubernetes.md
+++ b/site/kubernetes.md
@@ -30,6 +30,8 @@ Then run:
 
     weave setup
 
+Weave net depends on the *portmap* [standard](https://github.com/containernetworking/plugins) CNI plugin to support *hostport* functionality. Please ensure that *portmap* CNI plugin is installed in `/opt/cni/bin` directory.
+
 #### Launching Weave Net
 
 To create a network that spans multiple hosts, the Weave peers must be connected to each other.  
@@ -45,7 +47,7 @@ for a discussion on peer connections.
 
 All CNI plugins are configured by a JSON file in the directory
 `/etc/cni/net.d/`.  `weave setup` installs a minimal configuration
-file named `10-weave.conf`, which you can alter to suit your needs.
+file named `10-weave.conflist`, which you can alter to suit your needs.
 
 See the [CNI Spec](https://github.com/appc/cni/blob/master/SPEC.md#network-configuration)
 for details on the format and contents of this file.
--- a/weave
+++ b/weave
@@ -729,9 +729,20 @@ upgrade_cni_plugin() {
 create_cni_config() {
     cat >"$1" <<EOF
 {
+    "cniVersion": "0.3.0",
     "name": "weave",
-    "type": "weave-net",
-    "hairpinMode": $2
+    "plugins": [
+        {
+            "name": "weave",
+            "type": "weave-net",
+            "hairpinMode": $2
+        },
+        {
+            "type": "portmap",
+            "capabilities": {"portMappings": true},
+            "snat": true
+        }
+    ]
 }
 EOF
 }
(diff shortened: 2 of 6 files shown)


weave
```diff
@@ -741,8 +752,11 @@ setup_cni() {
     if install_cni_plugin $CNI_PLUGIN_DIR ; then
         upgrade_cni_plugin $CNI_PLUGIN_DIR
     fi
-    if [ -d $HOST_ROOT/etc/cni/net.d -a ! -f $HOST_ROOT/etc/cni/net.d/10-weave.conf ] ; then
-        create_cni_config $HOST_ROOT/etc/cni/net.d/10-weave.conf $HAIRPIN_MODE
+    if [ -d $HOST_ROOT/etc/cni/net.d -a -f $HOST_ROOT/etc/cni/net.d/10-weave.conf ] ; then
```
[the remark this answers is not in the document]
murali-reddy: First one is intended to delete if there is `10-weave.conf`. Like in case of upgrades. So both fresh installs and existing clusters after upgrade will only have `10-weave.conflist`. Please let me know if you see any issue with this approach.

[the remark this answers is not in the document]
bboreham: Yes, I get what you are trying to do, that part is fine.
My point is that `-f $HOST_ROOT/etc/cni/net.d/10-weave.conf` implies `-d $HOST_ROOT/etc/cni/net.d`, so you don't need both parts of the `-a`.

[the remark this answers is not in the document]
murali-reddy: Make sense. I removed redundant check.

weave
```diff
@@ -728,9 +728,20 @@ upgrade_cni_plugin() {
...
-    "type": "weave-net",
-    "hairpinMode": $2
+    "plugins": [
+        {
+            "name": "weave",
+            "type": "weave-net",
+            "hairpinMode": $2
+        },
+        {
+            "type": "portmap",
```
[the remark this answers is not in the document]
murali-reddy: I updated documentation.

site/kubernetes.md
```diff
@@ -30,6 +30,8 @@ Then run:
 
     weave setup
 
+Weave net depends on the *portmap* [standard](https://github.com/containernetworking/plugins) CNI plugin to support *hostport* functionality. Please ensure that *portmap* CNI plugin is installed in `/opt/cni/bin` directory.
```
bboreham: Please put a similar note in `site/kubernetes/kube-addon.md` - that is where we encourage people to look first.

murali-reddy: Sure will do, I am also planning to write a test case.

weave
```diff
@@ -741,8 +752,13 @@ setup_cni() {
     if install_cni_plugin $CNI_PLUGIN_DIR ; then
         upgrade_cni_plugin $CNI_PLUGIN_DIR
     fi
-    if [ -d $HOST_ROOT/etc/cni/net.d -a ! -f $HOST_ROOT/etc/cni/net.d/10-weave.conf ] ; then
-        create_cni_config $HOST_ROOT/etc/cni/net.d/10-weave.conf $HAIRPIN_MODE
+
+    # Weave will use .conflist for versions > 2.4.0. So in case of upgrade of old
+    # clusters with Weave version <= 2.4.0 remove existing 10-weave.conf
+    rm -f $HOST_ROOT/etc/cni/net.d/10-weave.conf
```
annismckenzie: I really don't like an `rm -f` on $HOST_ROOT in a script running in a container. I want the feature to be implemented so that we can stop putting a `00-weave.conflist` on the hosts via SaltStack but there's got to be an opt-in way for old clusters. Leave it alone and add a `09-weave.conflist` so as to not break other people's overrides?

murali-reddy: It would be ideal to avoid supporting both `.conf` and `.conflist` so that both new and existing deployments works in consistent manner. Also note that other CNI's (calico, flannel for e.g.) already supports `.conflist` only. Since the change explicitly deleting a single file should that be really of that concern? What could be possible issue doing so?

site/kubernetes.md
murali-reddy: @bboreham I have a smoke test and addressed your comments. Please see if this PR can be merged.

PR #3359, opened 2018-07-26
check for case where empty list of routes is returned in the result from IPAM plugin

### LINKED ISSUE #3331: Weave IPAM CNI plugin cannot distinguish between empty and nil routes
Consider the following CNI config:

```
{
    "name": "weave",
    "type": "weave-net",
    "hairpinMode": true,
    "ipam": {
       "type": "weave-ipam",
       "subnet": "10.44.3.0/24"
    }
}
```

When using the config, expected behavior is that the Weave CNI plugin will add a default route as can be seen [here](https://github.com/weaveworks/weave/blob/v2.3.0/plugin/net/cni.go#L145). However, the IPAM plugin [populates](https://github.com/weaveworks/weave/blob/v2.3.0/plugin/ipam/cni.go#L60) the `Routes` field with empty list, so the default route is not added.

murali-reddy:
Add default route entry in CNI plugin in case where IPAM plugin return a  0 length array of routes

Fixes #3331

### DIFF
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -152,7 +152,7 @@ func (c *CNIPlugin) CmdAdd(args *skel.CmdArgs) error {
 
 func setupRoutes(link netlink.Link, name string, ipnet net.IPNet, gw net.IP, routes []*types.Route) error {
 	var err error
-	if routes == nil { // If config says nothing about routes, add a default one
+	if routes == nil || len(routes) == 0 { // If config says nothing about routes, add a default one
 		if !ipnet.Contains(gw) {
 			// The bridge IP is not on the same subnet; add a specific route to it
 			gw32 := &net.IPNet{IP: gw, Mask: mask32}


plugin/net/cni.go
```diff
@@ -152,7 +152,7 @@ func (c *CNIPlugin) CmdAdd(args *skel.CmdArgs) error {
 
 func setupRoutes(link netlink.Link, name string, ipnet net.IPNet, gw net.IP, routes []*types.Route) error {
 	var err error
-	if routes == nil { // If config says nothing about routes, add a default one
+	if routes == nil || len(routes) == 0 { // If config says nothing about routes, add a default one
```
bboreham: This change makes the comment wrong - if it says _nothing_ then the value will be nil.

This is a change to the intended behaviour - can you check where the empty list is coming from, so we can understand why we need to change?

It would also be good to have some way to request the default Weave Net route to be added even when an IPAM plugin has added some routes.

murali-reddy: >  can you check where the empty list is coming from, so we can understand why we need to change?
Weave IPAM plug-in is just copying the routes from configuration to response.
https://github.com/weaveworks/weave/blob/master/plugin/ipam/cni.go#L60

Same is case with `host-local` plugin https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/main.go#L114

> This is a change to the intended behaviour 

In which case I will move the check in to Weave's IPAM so _it does not return empty list_ when _config says nothing about routes_

bboreham: But where was the zero-length slice created?  It's not in the config in #3331

murali-reddy: Since the CNI go [definition](https://github.com/containernetworking/cni/blob/master/pkg/types/current/types.go#L140) for IPAM response has Routes as array, json un-marshalling is creating empty slice

```go
// Result is what gets returned from the plugin (via stdout) to the caller
type Result struct {
	CNIVersion string         `json:"cniVersion,omitempty"`
	Interfaces []*Interface   `json:"interfaces,omitempty"`
	IPs        []*IPConfig    `json:"ips,omitempty"`
	Routes     []*types.Route `json:"routes,omitempty"`
	DNS        types.DNS      `json:"dns,omitempty"`
}
```

murali-reddy: Zero-length slice if getting created in libcni, Please see the comment 

https://github.com/weaveworks/weave/issues/3331#issuecomment-408388766

I will close this PR as behaviour in weave CNI and IPAM plug-in's are intended.

PR #3366, opened 2018-08-03
Use k8s/client-go v8.0.0

### LINKED ISSUE #3358: Please update vendored copy of k8s.io/client-go
Our current copy of k8s.io/client-go uses juju/ratelimit which is LGPL3. Upstream moved away from this in kubernetes/client-go@acc5249 - it'd be good if we followed this change as well.

brb:
Avoids including juju/ratelimit licensed under LGPL3.

Updated with dep v0.5.0.

Fix #3358


murali-reddy:
LGTM. Don't see any breaking changes with v8.0.0 that can impact so should be fully compatible.

Issue #3347, opened 2018-07-12
Network policy: ingress traffic is allowed when no namespaceSelector/podSelector is specified


murali-reddy:
## What you expected to happen?

An user reported this issues in the slack
https://weave-community.slack.com/archives/C2ND76PAA/p1531355773000036

When an ingress network policy applies to a pod, ingress traffic to the pod is blocked by default. Traffic that should be white-listed in specified through `namespaceSelector`, `podSelector` or `ipBlock`. Weave currently does not support `ipBlock` yet, however `namespaceSelector` and `podSelector` specified as network policy peer are honoured.

When a network policy is specified with no `namespaceSelector` or `podSelector` logically we should expect all the ingress traffic to be dropped as no source is specified. However it is observed that traffic is allowed to the pods.

## What happened?

Iptable rule that accepts the ingress traffic should be added only if there are ip's selected by pod/namespace selectors. iptable rule look like below.

```
Chain WEAVE-NPC-INGRESS (1 references)
pkts bytes target     prot opt in     out     source               destination
   1    84 ACCEPT     all  --  *             *0.0.0.0/0            0.0.0.0/0            match-set weave-Q29A#@8a]c!^3}j+e2]M~moXY dst /* anywhere -> pods: namespace: default, selector: app=guestbook */
```
there is no matching ipset for source IP's, thus resulting in allowing all ingress traffic to the pods.

## How to reproduce it?

Apply any network policy with no `namespaceSelector`/`podSelector`. For e.g.

```
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: pod-allow-to-web
  namespace: my-policy
spec:
  podSelector:
    matchLabels:
      app: web
  policyTypes:
  - Ingress
  ingress:
  - from:
    - ipBlock:
        cidr: 100.x.x.x/32
```

rade:
See also #3344 for a similar "fail open" issue.

brb:
Probably this issue can be closed after #3168 has been fixed (I'm planning to do it after the Egress NPC PR has been merged).

brb:
Fixed by #3367

PR #3367, opened 2018-08-03
Add ipBlock support for Ingress netpol

### LINKED ISSUE #3168: Support ipBlock for ingress rules of NetworkPolicy in weave-npc

brb:
This PR enables ipBlock in k8s Ingress netpol as described in https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#networkpolicy-v1-networking-k8s-io.

When using the feature, one should be aware that in many cases a packet source IP addr gets mangled before reaching the `filter/FORWARD` chain in which the NPC filtering takes place. So, policies using ipBlock in ingress might not be applied to the packet. See https://github.com/weaveworks/weave/blob/master/docs/k8s-src-ip.md for the cases.

However, in some cases it is possible to disable mangling by setting `externalTrafficPolicy` to `Local`. See https://kubernetes.io/docs/tutorials/services/source-ip/.

Fix #3168 #3347

### DIFF
--- a/npc/analyser.go
+++ b/npc/analyser.go
@@ -63,6 +63,7 @@ func (ns *ns) analysePolicy(policy *networkingv1.NetworkPolicy) (
 			} else {
 				for _, peer := range ingressRule.From {
 					var srcSelector *selectorSpec
+					var srcRuleHost ruleHost
 
 					// NetworkPolicyPeer describes a peer to allow traffic from.
 					// Exactly one of its fields must be specified.
@@ -72,17 +73,22 @@ func (ns *ns) analysePolicy(policy *networkingv1.NetworkPolicy) (
 							return nil, nil, nil, nil, err
 						}
 						addIfNotExist(srcSelector, podSelectors)
-
+						srcRuleHost = srcSelector
 					} else if peer.NamespaceSelector != nil {
 						srcSelector, err = newSelectorSpec(peer.NamespaceSelector, nil, "", ipset.ListSet)
 						if err != nil {
 							return nil, nil, nil, nil, err
 						}
 						nsSelectors[srcSelector.key] = srcSelector
+						srcRuleHost = srcSelector
+					} else if peer.IPBlock != nil {
+						ipBlock := newIPBlockSpec(peer.IPBlock, ns.name)
+						ipBlocks[ipBlock.key] = ipBlock
+						srcRuleHost = ipBlock
 					}
 
 					if allPorts {
-						rule := newRuleSpec(policyTypeIngress, nil, srcSelector, targetSelector, nil)
+						rule := newRuleSpec(policyTypeIngress, nil, srcRuleHost, targetSelector, nil)
 						rules[rule.key] = rule
 					} else {
 						withNormalisedProtoAndPort(ingressRule.Ports, func(proto, port string) {
--- a/site/kubernetes/kube-addon.md
+++ b/site/kubernetes/kube-addon.md
@@ -154,7 +154,7 @@ labels. For more information on configuring network policies in
 Kubernetes see the
 [walkthrough](https://kubernetes.io/docs/tasks/administer-cluster/declare-network-policy/)
 and the [NetworkPolicy API object
-definition](https://v1-7.docs.kubernetes.io/docs/api-reference/v1.7/#networkpolicy-v1-networking)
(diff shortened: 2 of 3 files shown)


murali-reddy:
LGTM

> When using the feature, one should be aware that in many cases a packet source IP addr gets mangled before reaching the filter/FORWARD chain in which the NPC filtering takes place. So, policies using ipBlock in ingress might not be applied to the packet

Good point. I think this can not happen only with `cluster IP` or when externalTrafficPolicy is Local. Are we going to document this?

brb:
Thanks for the review and for the tip to document the constraints. I've updated the docs, see my last commit.

> I think this can not happen only with cluster IP

When accessing via ClusterIP, src IP addr is set to IP addr of the `weave` bridge of a host which initiated a request.

rade:
What is left to do here?

fvigotti:
Hi, I've tested this patch contained in latest weave container available in docker hub, 
to me it doesn't work either when container are on hostport or hostnetwork, I don't have a good knowledge about the internals of weave, but isn't possible to filter ip address on those ports too ? 
this is the only way for a bare metal server to have the ip address of the client 

(see : https://github.com/kubernetes/kubernetes/issues/67202 ) 
and If cannot be filtered in weave this leave with no solution for filtering ingress ip address ( even with your patch ) 
( seems that the filter is applied only on weave interfaces , but with dnat to internal pod ( as in hostport ) it should work.. )

if you think that this patch should also work for hostport | hostnetwork and in your setup works and want more info about my setup let me know.. 


if instead It is intented to mantain this deficiency in weave design or maybe will be fixed in future , let me know too :) 


at the moment the only solution left to bare metal cluster seems to filter everything in a sidecar proxy ( one proxy with one filtering configuration in each container that need to expose somthing on public ip ( but not to all public ips ) .. but this sound like a bad pattern to me.. ) 

thank you,
Francesco

murali-reddy:
> Hi, I've tested this patch contained in latest weave container available in docker hub,
to me it doesn't work either when container are on hostport or hostnetwork,

@fvigotti thanks for trying out the ipBlock support

As per semantics of network policies 

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#networkpolicy-v1-networking-k8s-io

```
Traffic is allowed to a pod if there are no NetworkPolicies selecting the pod (and cluster policy otherwise allows the traffic), OR if the traffic source is the pod's local node ....
```
So any traffic from local node should be permitted to the pod. Expected semantics of Kubernetes for network policies leaves ambiguity with pod's running in host network. I believe this constraints was put int place for kubelet for liveliness/readiness checks. There is no reason for other pods in host network should have access to the rest of the pods. Could you please open a separate issue, perhaps it can be considered to meet the end goal (kubelet should be able access all the pods) yet restricting un-intended pods even if there are in host network or use host ports.

fvigotti:
Hi @murali-reddy  thank for your fast reply,
but maybe I've not been descriptive enough in my mesasge, what I'm trying to filter with the ingress policy is not from hosts IPs (nodes )  nor pod..
I'm trying to filter IPs that are out of the cluster on public networks, for Ingress traffic into hostport/hostnetwork  ( no LB services ( not available in bare metal ) , no NodePort (not usabe for standard services shared to public across the cluster ) )

murali-reddy:
@fvigotti sorry i have misread your description.

From what I understand you would like to apply ingress network policies with ipBlocks for the pod's running on hostnetwork or using host-ports?  You are looking to permit/deny ingress traffic from outside the cluster.

> ( seems that the filter is applied only on weave interfaces , but with dnat to internal pod ( as in hostport ) it should work.. )

You are right. Polices are applied to pod's running in pod network CIDR. So even if network policy pod & namespace selectors picks the pods in hostnetwork or using host port policies wont get applied. Network policy semantics does not specify if policies should be applied to hostnetwork/hosport pods as well. I guess can be discussed. Would you mind opening a new issue for this please.

PR #3373, opened 2018-08-10
[WIP] Workaround to fix Kernel bug related ipset entry deletion

### LINKED ISSUE #3296: failed: ipset v6.32: Element cannot be deleted from the set: it's not added
Excerpt from log, focusing on one IP address:

```
INFO: 2018/05/09 16:00:59.820416 Starting Weaveworks NPC 2.3.0; node name "container11"
INFO: 2018/05/09 16:00:59.820625 Serving /metrics on :6781
Wed May  9 16:00:59 2018 <5> ulogd.c:843 building new pluginstance stack: 'log1:NFLOG,base1:BASE,pcap1:PCAP'
DEBU: 2018/05/09 16:00:59.836833 Got list of ipsets: [weave-local-pods weave-?b%zl9GIe0AET1(QI^7NWe*fO weave-iuZcey(5DeXbzgRFs8Szo]+@p weave-a4Xwd(HKP4gdUva6MiTEIZAnp weave-z~y01unAQHA]WxHG!ALB)5]}s weave-merwHPWBXl40TI+Z;zAiZs*_y weave-6iyod:sR_fJJ!asxT3lEJWI[d weave-E.1.0W^NGSp]0_t5WwH/]gX@L weave-k?Z;25^M}|1s7P3|H9i;*;MhG weave-U2!Dhw$r^7X%1=}Rnjqffc87_ weave-$@i@JIAK[omT8D]7^N@EPQ={9 weave-cFh|zJj^65nuF3x;YP1O.tN%9 weave-^/BTzYx)2*FacLfNjGa)Wjg=4 weave-0EHD/vdN#O4]V?o4Tx7kS;APH weave-4vtqMI+kx/2]jD%_c0S%thO%V]
DEBU: 2018/05/09 16:00:59.836864 Flushing ipset 'weave-local-pods'
...
INFO: 2018/05/09 16:01:05.590549 adding entry 10.104.238.41 to weave-z~y01unAQHA]WxHG!ALB)5]}s of e639917e-5394-11e8-8e7a-005056bf0013
INFO: 2018/05/09 16:01:05.590580 added entry 10.104.238.41 to weave-z~y01unAQHA]WxHG!ALB)5]}s of e639917e-5394-11e8-8e7a-005056bf0013
INFO: 2018/05/09 16:01:05.591479 adding entry 10.104.238.41 to weave-a4Xwd(HKP4gdUva6MiTEIZAnp of e639917e-5394-11e8-8e7a-005056bf0013
INFO: 2018/05/09 16:01:05.591502 added entry 10.104.238.41 to weave-a4Xwd(HKP4gdUva6MiTEIZAnp of e639917e-5394-11e8-8e7a-005056bf0013
INFO: 2018/05/09 16:39:05.125088 deleting entry 10.104.238.41 from weave-a4Xwd(HKP4gdUva6MiTEIZAnp of e639917e-5394-11e8-8e7a-005056bf0013
INFO: 2018/05/09 16:39:05.125110 deleted entry 10.104.238.41 from weave-a4Xwd(HKP4gdUva6MiTEIZAnp of e639917e-5394-11e8-8e7a-005056bf0013
FATA: 2018/05/09 16:39:05.133155 update pod: ipset [del weave-a4Xwd(HKP4gdUva6MiTEIZAnp 10.104.238.41] failed: ipset v6.32: Element cannot be deleted from the set: it's not added
: exit status 1
```

It looks to me like it only added once and deleted once from that set.

murali-reddy:
if the kernel version is in affected range of Kernels, then resync the entries to
expected set of entries.

https://bugzilla.netfilter.org/show_bug.cgi?id=1119

```
ipset (v6.30, v6.29, v6.25.1, but not v6.21.1) hash code is sometimes evicting (or bumping) as a 
side-effect other entries in the set upon entry deletion (ipset del).  The symptom of this is that you 
get the error "ipset v6.30: Element cannot be deleted from the set: it's not added" when deleting an 
entry that has not yet been legitimately deleted. The problem happens with a large number of entries; 
in some cases I've seen it with under 700 entries but typically it takes over 1,000 entries.  I've seen 
one, two, even three or four entries evicted on a deletion.
```
Fixes #3296 failed: ipset v6.32: Element cannot be deleted from the set: it's not added

### DIFF
--- a/net/ipset/ipset.go
+++ b/net/ipset/ipset.go
@@ -6,9 +6,12 @@ import (
 	"fmt"
 	"log"
 	"os/exec"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/weaveworks/weave/common"
 	"k8s.io/apimachinery/pkg/types"
 )
 
@@ -50,6 +53,31 @@ type ipset struct {
 	// events from k8s API server might be out of order causing duplicate IPs:
 	// https://github.com/weaveworks/weave/issues/2792.
 	users map[entryKey]map[types.UID]struct{}
+}
+
+var resyncOnDelete bool
+
+func init() {
+
+	// check if Kernel version is in between 4.2 and 4.10. There is a kerenl bug
+	// due to which ipset delete sometimes ends up in deleting unintended entries
+	// https://bugzilla.netfilter.org/show_bug.cgi?id=1119
+	// if Kernel version has this issue then we need to workaround by resyncing
+	// ipset to the expected list of entries
+	kernelVersion, err := exec.Command("uname", "-r").Output()
+	if err != nil {
+		common.Log.Fatalf("Failed to get Kernel version")
+	}
+	splitVersion := strings.SplitN(string(kernelVersion[:]), ".", 3)
+	if splitVersion[0] == "4" {
+		majorVersion, err := strconv.Atoi(splitVersion[1])
+		if err != nil {
+			common.Log.Fatalf("Failed to process Kernel major version")
+		}
+		if majorVersion >= 2 && majorVersion <= 10 {
+			resyncOnDelete = true
+		}
+	}
 }
 
 func New(logger *log.Logger, maxListSize int) Interface {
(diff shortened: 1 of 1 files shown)


rade:
Kernel versions are a poor indicator as to what functionality/bugs are present, since some distros aggressively backport changes.

murali-reddy:
> Kernel versions are a poor indicator as to what functionality/bugs are present, since some distros aggressively backport changes.

Ok, i will check if there is more reliable way. Nevertheless adding this workaround will add little overhead if Kernel is already has the fix and will not cause any side-affects.

bboreham:
If I read it right, you are triggering off the error message we saw in #3296, but the bug will hit earlier.
We need to check after every delete - maybe not straight after, maybe have a timer so we batch up multiple deletes to the same set.

murali-reddy:
@brb Key logic to detect if the Kernel has the bug is not reliable. As pointed out Kernel version is not the best way to figure if it has ipset bug. Do you have any better way to reliably figure if Kernel is affected with ipset issue?

brb:
I wouldn't bother with checking the kernel version, and I would enable the safe-delete for all.

Also, as we already bookkeeping all ipset elements in NPC, wouldn't it better to compare against the ones in NPC instead of `ipset list <...>`?

murali-reddy:
> I wouldn't bother with checking the kernel version, and I would enable the safe-delete for all.

@brb agree, its little bit of overhead but I can not think of any other way to find if the kernel is affected with the bug

> Also, as we already bookkeeping all ipset elements in NPC, wouldn't it better to compare against the ones in NPC instead of ipset list <...>?

Could you point me where bookkeeping of ipset elements is done? My intent is since its helper function, we don't know who the consumers are if they do any bookkeeping etc. Ideally would like to contain the changes with in the helper function without taking any help from the consumer of this utility library.

brb (commented):
stale PR

PR #3381, opened 2018-08-20
Monitor and throw error message in the logs if Weave bridge interface is down

### LINKED ISSUE #3133: Do something more useful when the weave bridge is DOWN

murali-reddy:
Monitor and throw error message in the logs if Weave bridge interface is found to be in DOWN state

Fixes #3133 Do something more useful when the weave bridge is DOWN

On Weave restart there is already logic in place to create the bridge interface
and bring it up, this fix only monotors the weave bridge interface and logs error if
it is in DOWN state

### DIFF
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net"
 	"path/filepath"
+	"syscall"
 
 	"github.com/coreos/go-iptables/iptables"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


net/bridge.go
```diff
@@ -609,3 +614,27 @@ func reexpose(config *BridgeConfig, log *logrus.Logger) error {
...
+		return fmt.Errorf("Unable to find link %q: %s", ifaceName, err)
+	}
+
+	updatesChannel := make(chan netlink.LinkUpdate)
+	if err := netlink.LinkSubscribe(updatesChannel, nil); err != nil {
+		return errors.Wrapf(err, "error monitoring link %q for UP/DOWN notifications", ifaceName)
+	}
+
+	go func() {
+		for {
```
brb: Nit: `for update := range updatesChannel {`?

PR #3385, opened 2018-08-28
Add logic to re-try chooseBest() (which will select best forwarder)

### LINKED ISSUE #1737: [fastdp] can get stuck in 'sleeve' mode even though fastdp is possible

murali-reddy:
changes differentiate beteween fatal errors (for e.g. ipsec init)
from transient errors (heartbeast misses). In case of transient
errors OverlayForwarder is marked to be unhealty. When overlay forwarder
is unhealty its kept on hold so chooseBest() will skip selecting the forwarder

Once OverlayForwarder is considered healthy run chooseBest() again so the
forwarder can be selected if its the best forwarder

Fixes #1737

### DIFF
--- a/router/fastdp.go
+++ b/router/fastdp.go
@@ -713,12 +713,18 @@ func (fwd *fastDatapathForwarder) doHeartbeats() {
 		select {
 		case <-fwd.heartbeatTimer.C:
 			if fwd.confirmed {
+				log.Debug(fwd.logPrefix(), "sending Heartbeat to peer")
 				fwd.sendHeartbeat()
 			}
 			fwd.heartbeatTimer.Reset(fwd.heartbeatInterval)
 
 		case <-fwd.heartbeatTimeout.C:
+			log.Debug(fwd.logPrefix(), "missed Heartbeat from peer, marking fastdp forwarder as un-healthy")
+
+			// treat missed heartbeat as transient error. Indicate to overlay forwarder
+			// that Forwarder is un-healthy so it can pick next best forwarder
 			fwd.healthChan <- false
+
 			if fwd.healthy {
 				fwd.healthy = false
 			} else {
@@ -756,7 +762,6 @@ func (fwd *fastDatapathForwarder) handleError(err error) {
 
 func (fwd *fastDatapathForwarder) sendHeartbeat() {
 	fwd.lock.RLock()
-	log.Debug(fwd.logPrefix(), "sendHeartbeat")
 
 	// the heartbeat payload consists of the 64-bit connection uid
 	// followed by the 16-bit packet size.
@@ -813,7 +818,10 @@ func (fwd *fastDatapathForwarder) handleVxlanSpecialPacket(frame []byte, sender 
 
 	if !fwd.ackedHeartbeat {
 		fwd.ackedHeartbeat = true
+		log.Debug(fwd.logPrefix(), "Ack Heartbeat from peer")
 		fwd.handleError(fwd.sendControlMsg(FastDatapathHeartbeatAck, nil))
+	} else {
+		log.Debug(fwd.logPrefix(), "Got Heartbeat Ack from peer")
 	}
 
 	// we can receive a heartbeat before Confirm() has set up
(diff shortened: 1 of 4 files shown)


murali-reddy:
Some of the users tried private image for this patch (see #1737) but unfortunately issue did not occur to verify the patch works. I have tested the patch by injecting the heartbeat misses. Please review the changes and if the retry approach is fine.

router/overlay_switch.go
```diff
@@ -248,6 +256,9 @@ loop:
 			e.established = true
 			establishedChan = nil
 
+		case healthy := <-healthChan:
```
bboreham: it would be slightly more consistent to say `fwd.HealthChannel()` here

router/fastdp.go
```diff
@@ -699,19 +703,33 @@ func (fwd *fastDatapathForwarder) ErrorChannel() <-chan error {
...
+			log.Debug(fwd.logPrefix(), "missed Heartbeat from peer, marking fastdp forwarder as un-healthy")
+
+			// treat missed heartbeat as transient error. Indicate to overlay forwarder
+			// that Forwarder is un-healthy so it can pick next best forwarder
+			fwd.healthChan <- false
+
+			if fwd.healthy {
+				fwd.healthy = false
+			} else {
+				err = fmt.Errorf("timed out waiting for vxlan heartbeat")
```
bboreham: I can't see how we will ever hit this line. `healthy` starts off `true` and is reset to `true` when we reset the timer.
I can see that you don't want an error which will shut down the forwarder, but you should have some replacement log line.

And, if `err` is never set, then we don't need the surrounding logic to handle it.

router/overlay_switch.go
```diff
@@ -166,6 +170,9 @@ type subForwarderEvent struct {
 
 	// is this an error event?
 	err error
+
+	// event to indicate if forwarder is in healthy state to be considered for best forwarded
+	healthy bool
```
bboreham: I think it would be better to have this as a `*bool`, then you can tell the difference between an event sent with this set to `false` and an event where it was never set.

router/fastdp.go
```diff
@@ -699,19 +703,33 @@ func (fwd *fastDatapathForwarder) ErrorChannel() <-chan error {
...
+				log.Debug(fwd.logPrefix(), "sending Heartbeat to peer")
 				fwd.sendHeartbeat()
 			}
 			fwd.heartbeatTimer.Reset(fwd.heartbeatInterval)
 
 		case <-fwd.heartbeatTimeout.C:
-			err = fmt.Errorf("timed out waiting for vxlan heartbeat")
+			log.Debug(fwd.logPrefix(), "missed Heartbeat from peer, marking fastdp forwarder as un-healthy")
+
+			// treat missed heartbeat as transient error. Indicate to overlay forwarder
```
bboreham: I think you should switch from fast-heartbeat to slow-heartbeat here, otherwise a connection that is never acknowledged will continue retrying every 500ms.

router/fastdp.go
murali-reddy: I should have expressed the intent more clearly. What `else` block does is if a forwarder is not healthy already and we have a consecutive heartbeat timeout then that forwarder is shutdown permanently. Which is not really better approach (just waiting till 2 heartbeatTimeout before shutting down the forwarder)

So if a forwarder is not `healthy` then should we have a time bound till when weaver keep sending the hearbeats? or should we keep sending the heartbeats forever without shutting down the forwarder?

router/fastdp.go
murali-reddy: Added the switch to slow-heatbeat before session is established with peer.

router/fastdp.go
murali-reddy: I made the change so that forwarder keep trying heartbeats and never shutdowns.

murali-reddy:
Addressed review comments PTAL

bboreham:
I tried this out: my principal tool to turn off fastdp was:
```
sudo iptables -D INPUT -p udp --dport 6784 -j DROP 
sudo iptables -D OUTPUT -p udp --dport 6784 -j DROP
```

I tried various combinations and all seemed fine.  Thanks!

PR #3388, opened 2018-08-31
add --without-masquerade option to `weave expose`

leprechau:
Add the ability to skip creation of MASQUERADE rules via iptables when executing the `weave expose` command


leprechau:
First PR to the weave project ... all comments/suggestions/etc... appreciated.

weave
```diff
@@ -430,9 +430,10 @@ EOF
 }
 
 expose_ip() {
+    [ -z $WITHOUT_MASQUERADE ] || skipNAT="/true"
     ipam_cidrs allocate_no_check_alive weave:expose $CIDR_ARGS
     for CIDR in $ALL_CIDRS ; do
-        call_weave "POST" "/expose/$CIDR"
+        call_weave "POST" "/expose/$CIDR$skipNAT"
```
bboreham: I think we have tended to do optional settings as query parameters (i.e. after `?`) rather than as part of the path

.gitignore
```diff
@@ -36,6 +36,9 @@ coverage.html
 # Sed backup files
 *.bak
 
+# JetBrains/GoLand/IntelliJ project file
+.idea
```
bboreham: This could go in your personal Git settings - I don't think we have any tool-specific ignores that aren't for tools specifically used in Weave Net.

weave
leprechau: Makes sense, I'll change that.

.gitignore
leprechau: Okay, I'll remove it.

leprechau:
Updated to address comments.  Let me know if there is anything else.

brb:
@leprechau Thanks for your contribution! Mind explaining why someone would want to skip the creation of the NAT rules?

leprechau:
We’re using weave in a standard Docker environment without k8s and want external systems (not part of the weave network) to see the container IP and not the IP of the host.

In our specific case we are using a helper written in-house (https://github.com/myENA/aardvark) to publish container routes and update the default route within the container when it starts.

brb:
@leprechau Wouldn't `--no-masq-local` from #3298 be enough for your case?

leprechau:
@brb Trying now ... I see the flag in `prog/weaver/main.go` but didn't see it listed in the actual `weave` help usage.  If it works I'll send a much smaller PR adding the option to the usage docs.

leprechau:
So, the `--no-masq-local` does not seem to work for my particular case.

What I did ...

1) Reset weave via `weave reset` on all test nodes
2) Restarted weave
```
weave launch --password xxxxxx --no-detect-tls --ipalloc-range 10.156.8.0/22 --no-dns --without-dns --proxy=false --no-masq-local
```
3) Created a new docker network using the `weave` driver
```
docker network create -d weave --subnet=10.156.8.0/22 --gateway 10.156.8.1 app
```
4) Started a test container on the app network
```

leprechau:
@brb Am I missing something in how the `--no-masq-local` option is supposed to work?  The default route in my container is the weave bridge but the traffic is still NAT'd through the host unless I completely remove the MASQ rules.

brb:
@leprechau 

> Result ... traffic leaving the container 10.156.8.2 still shows up outside of the weave network as being NAT'd through the host 10.156.0.183

OK, I misread your requirement - I thought that you want external systems to access containers w/o NAT (this is what the `--no-masql-local` for). Unfortunately, `"traffic leaving the container"` (I believe = "traffic originating from the container") will be NAT'ed.

leprechau:
@brb yes, I'm wanting hosts outside of the weave network to see traffic that originates from a container on the weave network to see the weave assigned IP address of the container.

leprechau:
@bboreham @brb Awesome, thank you both as well!

PR #3391, opened 2018-09-03
add documentation to enable debug logging

murali-reddy:


site/kubernetes/kube-addon.md
```diff
@@ -203,6 +203,18 @@ $ kubectl logs -n kube-system weave-net-1jkl6 weave
 For easier viewing, pipe the output into a file , especially if
 it is long.
 
+By default log level is set to `info`. If you wish to see more detailed logs you can set the desired log level for the `-log-level` flag through the `EXTRA_ARGS` environment variable for the `weave` and `weave-npc` containers in the weave-net daemon set. Add environment variable as below.
```
brb: `weave-npc` does not use `$EXTRA_ARGS`, so you cannot change its log level via the var.

bboreham: also it is currently set to run at `debug` level

bboreham:
It would be simpler to just add `--log-level` to the command, unless you specifically wanted it in both invocations of `weaver`.

bboreham:
I guess an env var is safer, also can be added via the mechanism at https://www.weave.works/docs/net/latest/kubernetes/kube-addon/#configuration-options

bboreham:
This should really be on the 2.4 branch, so it feeds into the public website immediately.

bboreham:
I've rebased and changed the branch to 2.4

Issue #3394, opened 2018-09-05
weave-kube reclaim is not able to rm a dead peer if a running peer has the same IP addr


brb:
As seen in https://github.com/weaveworks/weave/issues/3384#issuecomment-418461201, it's possible for a new node to have an IP addr of a dead node. In this situation, the weave-kube reclaimer won't be able to rmpeer the dead node, because it doesn't identify running nodes by their peer names, but their hostname (https://github.com/weaveworks/weave/blob/v2.4.0/prog/kube-utils/main.go#L114).

Raffo:
Thanks for this issue! Does it mean that the peer name is determined in a unique way (as explained in the doc) but the deletion is just wrong cause it only uses the hostname?

bboreham:
To clarify (as noted in #3384) what is supposed to happen:

>  * Kubernetes has a list of Nodes.
>  * Weave Net maintains a list of Nodes that have been added as "peers" to the Weave Net cluster.
>
> Because we don't get a chance to clean up when a Node is being shut down, we only notice afterwards
that a node has been deleted. Specifically we spot that that a node is in the second list and not in the first list.

This bug is that the 5-6 lines of code which compute the difference of those two lists are doing it on hostname not on peer name.

To cause a problem the hostname re-use has to happen before any Weave Net pod starting up has had a chance to remove it.

bboreham:
An idea to address this:
Store the Kubernetes UID of each Node in the peer list, then we can differentiate new Nodes that have the same name.

Raffo:
Using the Kubernetes UUID sounds like a good idea as we know for sure that NodeName and Hostname are not unique.

PR #3397, opened 2018-09-06
Reclaim IPs when node name is re-used

bboreham:
If we can see that a different peer was previously running with the same node name, mark it for reclaim.

This is a straightforward change to improve #3394, but not a full fix, which would involve storing the UID and dealing with upgrading from a previous version which didn't store it.

### DIFF
--- a/prog/kube-utils/main.go
+++ b/prog/kube-utils/main.go
@@ -96,7 +96,7 @@ func checkIamInPeerList(cml *configMapAnnotations, c *kubernetes.Clientset, peer
 
 // For each of those peers that is no longer listed as a node by
 // Kubernetes, remove it from Weave IPAM
-func reclaimRemovedPeers(weave *weaveapi.Client, cml *configMapAnnotations, nodes []nodeInfo, myPeerName string) error {
+func reclaimRemovedPeers(weave *weaveapi.Client, cml *configMapAnnotations, nodes []nodeInfo, myPeerName, myNodeName string) error {
 	for loopsWhenNothingChanged := 0; loopsWhenNothingChanged < 3; loopsWhenNothingChanged++ {
 		if err := cml.Init(); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


bboreham:
> not a full fix

actually it doesn't seem that bad now: it doesn't let any _other_ peers reclaim that node which is now replaced, but as long as Weave Net does start up on that node it will do the reclaim.

prog/kube-utils/main.go
```diff
@@ -110,9 +110,19 @@ func reclaimRemovedPeers(weave *weaveapi.Client, cml *configMapAnnotations, node
 		for _, peer := range storedPeerList.Peers {
 			peerMap[peer.NodeName] = peer
 		}
+		// 1a. Check if the list held a peer with my NodeName but different PeerName
+		var deleteOldMe *peerInfo
+		if oldMe, found := peerMap[myNodeName]; found && oldMe.PeerName != myPeerName {
```
brb: `storedPeerList` might contain multiple entries with the same `NodeName`. Only one of such entries can be stored in `peerMap` (line 111), so we are risking of not removing other entries.

bboreham: Good spot.  That can happen if the problem has re-occurred over time.

bboreham:
I rewrote that part so the map is not indexed by node-name.

prog/kube-utils/main.go
```diff
@@ -106,13 +106,25 @@ func reclaimRemovedPeers(weave *weaveapi.Client, cml *configMapAnnotations, node
 		if err != nil {
 			return err
 		}
+		nodeMap := make(map[string]struct{}, len(nodes))
+		for _, node := range nodes {
+			nodeMap[node.name] = struct{}{}
```
brb: Nit: s/nodeMap/nodeSet

bboreham: changed

PR #3399, opened 2018-09-10
watch for Kubernetes node delete events and reclaim removed peers IP space on delete event

murali-reddy:
Changes to invoke `reclaimRemovedPeers` in `kube-utils` on Kubernetes node delete event


bboreham:
One thought: triggering a reclaim on all peers immediately will give a "thundering herd" effect with lots of contention over the ConfigMap we use as a lock.

Maybe have each peer sleep a random number of seconds before beginning, to avoid this?

murali-reddy:
> Maybe have each peer sleep a random number of seconds before beginning, to avoid this?

Agree. I will add random delay.

murali-reddy:
Following up on the comments from the slack conversation
https://weave-community.slack.com/archives/C9N5HME4B/p1536572237000100

> proposes to run another in the background to listen to Kubernetes events, and I wondered if it would be better to put it in a sidecar instead (i.e. a third container in the DaemonSet) Originally I was worried about defunct processes, but now I think that's ok because the shell will be the parent (in #3399).
Another downside of just firing off the process in the background is that if it hits a problem and dies this might not be noticed.

> I'm in favor of moving the functionality into `weaver`. I'm afraid that over time `weave-kube/launch.sh` will become a new `weave` (shell script).

@brb Are you suggesting to move `kube-utils` functionality in to `weaver`? or complete Kubernetes bits (launch.sh+weave-kube)? That would be significant work.  Change in this PR are very contained changes and I was wondering we should go for sidecar approach for now for immediate functional benefit?

brb:
@murali-reddy 

> Are you suggesting to move kube-utils functionality in to weaver?

Just the reclaim parts from kube-utils. It shouldn't require a lot of effort - move to a separate pkg and start the reclaiming process in a goroutine.

bboreham:
Chatting with Murali IRL yesterday, I felt the sidecar idea is bad because we will have to ask users to fetch two logs for troubleshooting.
I dislike the idea of `weaver` getting 35MB larger, too.

So, right now, I'm ok with running it as a background process from the `launch.sh` shell.

brb:
If so, then we need to ask k8s to pass `--init` to the `weave` Docker container (not sure whether it's possible), as the shell script is not a perfect init system.

bboreham:
> the shell script is not a perfect init system

Can you expand?  We already run stuff in the background in that shell script, and I hadn't noticed an issue.

brb:
The `A simple init system` section in https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/ describes it well.

bboreham:
That identifies a problem where the parent dies and leaves the child.  Since in our case the parent is the main process of the container, if it dies the entire container will go away.

Or did I miss something?

brb:
If the shell process (parent) is terminated, the child will receive `SIGKILL` which is unhandled (no chance to do a proper cleanup).

murali-reddy:
At least in the context of the function this background process does, graceful shutdown is not critical

murali-reddy:
I will add `weave forget` for the deleted node as separate PR (need to add go weave client API call for `forget`). Let me know if we have consensus on the approach to go ahead.

Testing done:

On AWS cluster, deleted couple of instances in ASG and verified on node delete IP space allocated for the node is reclaimed for all deleted nodes.

Let me know if its worth to write an integration test by doing `kubectl delete node` and verifying IP space is reclaimed.

brb:
Thanks for the PR.

>  Let me know if we have consensus on the approach to go ahead.

I still think that we should do it properly and run the reclaimer from the `weaver` process.

prog/kube-utils/main.go
```diff
@@ -183,6 +190,42 @@ func reclaimPeer(weave weaveClient, cml *configMapAnnotations, peerName string,
...
 
+// regiesters with Kubernetes API server for node delete events. Node delete event handler
+// invokes reclaimRemovedPeers to remove it from IPAM so that IP space is reclaimed
+func registerForNodeUpdates(client *kubernetes.Clientset, stopCh <-chan struct{}, nodeName, peerName string) {
+	informerFactory := informers.NewSharedInformerFactory(client, 0)
+	nodeInformer := informerFactory.Core().V1().Nodes().Informer()
+	common.Log.Debugln("registering for updates for node delete events")
+	nodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			node := obj.(*v1core.Node)
```
bboreham: I think this can be a `cache.DeletedFinalStateUnknown`, which is possible after this process has reconnected to the api-server.

prog/kube-utils/main.go
```diff
@@ -183,6 +190,42 @@ func reclaimPeer(weave weaveClient, cml *configMapAnnotations, peerName string,
...
+			config, err := rest.InClusterConfig()
+			if err != nil {
+				common.Log.Fatalf("[kube-peers] Could not get cluster config: %v", err)
+			}
+			client, err := kubernetes.NewForConfig(config)
+			if err != nil {
+				common.Log.Fatalf("[kube-peers] Could not make Kubernetes connection: %v", err)
+			}
+			cml := newConfigMapAnnotations(configMapNamespace, configMapName, client)
+			weave := weaveapi.NewClient(os.Getenv("WEAVE_HTTP_ADDR"), common.Log)
```
bboreham: seems untidy to do this differently between `justReclaim` and `runReclaimDaemon` - this line could go inside `reclaimRemovedPeers`

murali-reddy: Agree. Moved the weave client creation part to `reclaimRemovedPeers`. To keep things tidy does it make sense to add `addMyselfToPeerList` to `runReclaimDaemon` as well (that the only difference i see between `justReclaim` and `runReclaimDaemon`?

prog/kube-utils/main.go
murali-reddy: Added to logic to handle deleted objects.

prog/kube-utils/main.go
```diff
@@ -183,6 +191,51 @@ func reclaimPeer(weave weaveClient, cml *configMapAnnotations, peerName string,
...
+				if !ok {
+					common.Log.Fatalf("[kube-peers] Couldn't get node object from tombstone: %#v", obj)
+				}
+				nodeObj, ok = tombstone.Obj.(*v1core.Node)
+				if !ok {
+					common.Log.Fatalf("[kube-peers] Tombstone contained object that is not a Node: %#v", obj)
+				}
+			}
+			common.Log.Debugln("[kube-peers] Nodes deleted:", nodeObj.Name)
+			config, err := rest.InClusterConfig()
```
bboreham: this is a bit weird - surely we already have a connection to Kubernetes otherwise we wouldn't be in a callback

prog/kube-utils/main.go
```diff
@@ -183,6 +191,51 @@ func reclaimPeer(weave weaveClient, cml *configMapAnnotations, peerName string,
...
+				tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+				if !ok {
+					common.Log.Fatalf("[kube-peers] Couldn't get node object from tombstone: %#v", obj)
+				}
+				nodeObj, ok = tombstone.Obj.(*v1core.Node)
+				if !ok {
+					common.Log.Fatalf("[kube-peers] Tombstone contained object that is not a Node: %#v", obj)
+				}
+			}
+			common.Log.Debugln("[kube-peers] Nodes deleted:", nodeObj.Name)
```
bboreham: Now I look at it again, you're going through a lot of work just to print out a debug statement - the code doesn't use the name of the node that was deleted.

murali-reddy: Removed unnecessary verbose code.

murali-reddy:
I have included `weave connect --replace` as well now. Tested by deleting/adding the nodes from ASG on AWS.

On node delete now 
- `rmpeer` is done so IP addresses of deleted nodes are reclaimed and we have clean `weave status ipam`
- `connect --replace with empty peer list`: a node deleted is no longed connected and we have clean `weave status connections`

Please take a look.

Issue #3417, opened 2018-09-27
Liveness probe makes debugging very hard without apparent benefit


annismckenzie:
The Kubernetes DaemonSet uses a liveness probe to restart failed Weave pods which makes debugging issues with Weave almost impossible. We've since switched it over to a readiness probe which results in the behavior of the Weave pod not being in the ready state while an operator can then go into the Weave container and see what's wrong (`weave --local status` and all of those).

My question is: is there something that I'm missing on why a liveness probe was chosen?

## What happened?
Failing Weave containers are continuously being reaped by the Kubelet, making it impossible to debug a problem.

## Anything else we need to know?
No, this isn't a question specific to any cloud provider or hardware, just Weave on K8s using the DaemonSet.

## Versions:
```
$ weave version
2.4.1

$ docker version
Client:
 Version:      17.03.1-ce
 API version:  1.27
 Go version:   go1.7.5
 Git commit:   c6d412e32
 Built:        Fri Mar 24 00:39:57 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.03.1-ce
 API version:  1.27 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   c6d412e32
 Built:        Fri Mar 24 00:39:57 2017
 OS/Arch:      linux/amd64
 Experimental: false

$ uname -a
Linux 4.9.0-7-amd64 #1 SMP Debian 4.9.110-1 (2018-07-05) x86_64 GNU/Linux

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-07T23:17:28Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-07T23:08:19Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

murali-reddy:
`liveness` and `readiness` probes serves different purposes, so its not a question on one over other. 

>The Kubernetes DaemonSet uses a liveness probe to restart failed Weave pods which makes debugging issues with Weave almost impossible

[...]

PR #3421, opened 2018-09-28
kubernetes: Morph the livenessProbe into readinessProbe

dlespiau:
The liveness probe has the not-so-great side effect of killing the weave pods
if anything goes wrong, making debugging really difficult.
    
Not only that, we also bumped into the livenessProbe killing pods before they
had a chance to start (see below for the full story).
    
The readinessProbe is only used for Kubernetes Services: no traffic will be
routed to non-Ready pods when using the Service IP. We can use to to signal the
user something in wrong with Weave Net while not having Kubernetes delete the
pods behind our backs.
    
I removed the ininitialDelaySeconds field as Kurbernetes won't kill the pod if
not ready.
    
Notes from a debugging session:
    
Some nodes may have a vast amount of iptables rules, making each iptables
command take a few seconds (~5-10s). At startup, when the weave container
creates its chains, it runs quite a few iptables commands and the total time of
execution was > 30s.
    
It's important to note that, at this point, the weave process hasn't started as
those iptables commands are run by an initial script.
    
Kubernetes was trying to check for the container liveness, was failing, and
then proceeded to kill the weave pods before they had a chance to fully start.

Fixes: #3417

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
@@ -110,12 +110,11 @@ items:
                       fieldPath: spec.nodeName
               image: 'weaveworks/weave-kube:latest'
               imagePullPolicy: Always
-              livenessProbe:
+              readinessProbe:
                 httpGet:
                   host: 127.0.0.1
                   path: /status
                   port: 6784
-                initialDelaySeconds: 30
               resources:
                 requests:
                   cpu: 10m
--- a/prog/weave-kube/weave-daemonset-k8s-1.7.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.7.yaml
@@ -119,12 +119,11 @@ items:
                       fieldPath: spec.nodeName
               image: 'weaveworks/weave-kube:latest'
               imagePullPolicy: Always
-              livenessProbe:
+              readinessProbe:
                 httpGet:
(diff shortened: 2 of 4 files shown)


bboreham:
See #3417 wherein we argue that the liveness probe should be removed

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
```diff
@@ -110,12 +110,11 @@ items:
                       fieldPath: spec.nodeName
               image: 'weaveworks/weave-kube:latest'
               imagePullPolicy: Always
-              livenessProbe:
+              readinessProbe:
                 httpGet:
                   host: 127.0.0.1
                   path: /status
                   port: 6784
-                initialDelaySeconds: 30
```
annismckenzie: I'd still keep this. There's no need for the Kubelet to check for readiness immediately, taking pressure off that path?

brb:
We should update Launch Generator too.

dlespiau:
I have a PR updating launch-generator accordingly but didn't mention it here as it's not part of the open source project.

prog/weave-kube/weave-daemonset-k8s-1.6.yaml
bboreham: On second thoughts: "meh".  It's a small code path and the readiness probe fires every 10 seconds, so hardly any load.

PR #3423, opened 2018-10-02
Replace weave-local-pods ipset with physdev iptables rule

### LINKED ISSUE #3344: Replace weave-local-pods ipset with physdev iptables rule
We can get rid of the `weave-local-pods` ipset used by the weave-npc ingress rule `-A WEAVE-NPC -m set ! --match-set weave-local-pods dst -j ACCEPT` with `-A WEAVE-NPC -m physdev --physdev-out=vethwe-bridge -j ACCEPT`.

The former rule is insecure, as it might allow ingress traffic if the ipset is not properly maintained (e.g. due to crashing weave-npc as in https://github.com/weaveworks/weave/issues/3032#issuecomment-403397070) or if weave-npc lags behind processing updates from kube-apiserver.

murali-reddy:
by-pass ingress network policy chains and accept the packet if
destination is not one of the local pods

return the packets so the egress network policy chains are skipped
in case packet's source is not one of the local pods

Fixes #3344

### DIFF
--- a/prog/weave-npc/main.go
+++ b/prog/weave-npc/main.go
@@ -125,6 +125,12 @@ func createBaseRules(ipt *iptables.IPTables, ips ipset.Interface) error {
 		}
 	}
 
+	// If the destination address is not any of the local pods, let it through
+	if err := ipt.Append(npc.TableFilter, npc.MainChain,
+		"-m", "physdev", "--physdev-out=vethwe-bridge", "-j", "ACCEPT"); err != nil {
+		return err
+	}
+
 	if err := ipt.Append(npc.TableFilter, npc.MainChain,
 		"-m", "state", "--state", "NEW", "-j", string(npc.DefaultChain)); err != nil {
 		return err
@@ -132,21 +138,6 @@ func createBaseRules(ipt *iptables.IPTables, ips ipset.Interface) error {
 
 	if err := ipt.Append(npc.TableFilter, npc.MainChain,
 		"-m", "state", "--state", "NEW", "-j", string(npc.IngressChain)); err != nil {
-		return err
-	}
-
-	// If the destination address is not any of the local pods, let it through
-	found, err := ipsetExist(ips, npc.LocalIpset)
-	if err != nil {
-		return err
-	}
-	if !found {
-		if err := ips.Create(npc.LocalIpset, ipset.HashIP); err != nil {
-			return err
-		}
-	}
-	if err := ipt.Append(npc.TableFilter, npc.MainChain,
-		"-m", "set", "!", "--match-set", npc.LocalIpset, "dst", "-j", "ACCEPT"); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


prog/weave-npc/main.go
```diff
@@ -125,28 +125,19 @@ func createBaseRules(ipt *iptables.IPTables, ips ipset.Interface) error {
 		}
 	}
 
+	// If the destination address is not any of the local pods, let it through
 	if err := ipt.Append(npc.TableFilter, npc.MainChain,
-		"-m", "state", "--state", "NEW", "-j", string(npc.DefaultChain)); err != nil {
+		"-m", "physdev", "--physdev-out=vethwe-bridge", "-j", "ACCEPT"); err != nil {
```
bboreham: Hard-coding `vethwe-bridge` here seems wrong.  

Can this rule go in `weaver` where the rules go that drive traffic through the weave-npc chains?
Note that weave-npc is not otherwise tied to the specific implementation of Weave Net.

prog/weave-npc/main.go
```diff
@@ -179,7 +170,7 @@ func createBaseRules(ipt *iptables.IPTables, ips ipset.Interface) error {
 
 	ruleSpecs := [][]string{
 		{"-m", "state", "--state", "RELATED,ESTABLISHED", "-j", "ACCEPT"},
-		{"-m", "state", "--state", "NEW", "-m", "set", "!", "--match-set", npc.LocalIpset, "src", "-j", "RETURN"},
+		{"-m", "physdev", "--physdev-in=vethwe-bridge", "-j", "RETURN"},
```
bboreham: I'm not so sure this rule can move.  But again, hard-coding the device name is wrong.

prog/weave-npc/main.go
murali-reddy: removed hard-coded bridge name. PTAL

> Can this rule go in weaver where the rules go that drive traffic through the weave-npc chains?

No. moving it to `weaver` means this scenario will not work with kubenet etc. It fits in well in weave-npc as part of bunch of base rules that gets configured.

prog/weave-npc/main.go
```diff
@@ -127,7 +128,7 @@ func createBaseRules(ipt *iptables.IPTables, ips ipset.Interface) error {
 
 	// If the destination address is not any of the local pods, let it through
 	if err := ipt.Append(npc.TableFilter, npc.MainChain,
-		"-m", "physdev", "--physdev-out=vethwe-bridge", "-j", "ACCEPT"); err != nil {
+		"-m", "physdev", "--physdev-out="+bridgeName, "-j", "ACCEPT"); err != nil {
```
brb: I think `bridgeName` is a bit confusing here. IMO, `bridgePortName` would be better.

murali-reddy:
Addressed all comments. PTAL

prog/weave-npc/main.go
```diff
@@ -194,6 +186,17 @@ func createBaseRules(ipt *iptables.IPTables, ips ipset.Interface) error {
 		return err
 	}
 
+	// delete `weave-local-pods` ipset which is no longer used by weave-npc
+	weaveLocalPodExist, err := ipsetExist(ips, npc.LocalIpset)
+	if err != nil {
+		common.Log.Errorf("Failed to destroy ipset '%s'", npc.LocalIpset)
```
bboreham: This message is a little misleading.  It's something like "failed to look for ipset".

PR #3430, opened 2018-10-17
Bump go-odp dependency

brb:
To include the fix https://github.com/weaveworks/go-odp/pull/7.

TL;DR: without this fix, `fastdp` would fail to work on the 4.19 kernel.

Thanks @lkpdn for spotting and providing the fix.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -671,11 +671,11 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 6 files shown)


fgeorgatos:
still trying to understand it: is it to be expected that only TCP traffic reproduces it? (ie OK w. `-u`)

PR #3431, opened 2018-10-17
add checkpoint to include Kubernetes cluster information

### LINKED ISSUE #3324: Extend checkpoint to include information about Kubernetes clusters
Currently, we do not have any coarse grained information about Kubernetes clusters which are powered by Weave Net. The information would be used for prioritizing issues.

1. Kubernetes version.
2. Nodes count.
3. Whether NetworkPolicies are used (and what type - Ingress / Egress).

murali-reddy:
Fixes #3324

### DIFF
--- a/prog/weaver/checkpoint.go
+++ b/prog/weaver/checkpoint.go
@@ -1,12 +1,17 @@
 package main
 
 import (
+	"os"
+	"strconv"
 	"sync/atomic"
 	"syscall"
 	"time"
 
-	"github.com/weaveworks/go-checkpoint"
+	checkpoint "github.com/weaveworks/go-checkpoint"
 	weave "github.com/weaveworks/weave/router"
+	api "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 var checker *checkpoint.Checker
@@ -14,10 +19,12 @@ var newVersion atomic.Value
 var success atomic.Value
 
 const (
-	updateCheckPeriod = 6 * time.Hour
+	updateCheckPeriod  = 6 * time.Hour
+	configMapName      = "weave-net"
+	configMapNamespace = "kube-system"
 )
 
-func checkForUpdates(dockerVersion string, router *weave.NetworkRouter) {
+func checkForUpdates(dockerVersion string, router *weave.NetworkRouter, peers []string) {
 	newVersion.Store("")
 	success.Store(true)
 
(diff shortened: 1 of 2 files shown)


prog/weaver/checkpoint.go
```diff
@@ -48,10 +55,11 @@ func checkForUpdates(dockerVersion string, router *weave.NetworkRouter) {
 	}
 	kernelVersion := string(releaseBytes[:i])
 
-	flags := map[string]string{
-		"docker-version": dockerVersion,
-		"kernel-version": kernelVersion,
-	}
+	flags := make(map[string]string)
```
bboreham: I don't understand the need for this change.

prog/weaver/checkpoint.go
```diff
@@ -1,23 +1,30 @@
...
 
 const (
-	updateCheckPeriod = 6 * time.Hour
+	updateCheckPeriod  = 6 * time.Hour
+	configMapName      = "weave-net"
+	configMapNamespace = "kube-system"
 )
 
-func checkForUpdates(dockerVersion string, router *weave.NetworkRouter) {
+func checkForUpdates(dockerVersion string, router *weave.NetworkRouter, peers []string) {
```
bboreham: I think I would just pass in `len(peers)`, on the principle of minimum information.

murali-reddy: Modified.

prog/weaver/checkpoint.go
murali-reddy: Reverted this change. Not required.

prog/weaver/main.go
```diff
@@ -687,7 +687,7 @@ func determineQuorum(initPeerCountFlag int, router *weave.NetworkRouter) uint {
 	// that resolve to the same peer, in which case the quorum
 	// might be larger than it needs to be.  But the user can
 	// specify it explicitly if that becomes a problem.
-	clusterSize := uint(len(peers) + 1)
+	clusterSize := uint(len(peers))
```
bboreham: Why did we lose the `+1` here?  The comment suggests it is the "conservative assumption"

murali-reddy: :man_facepalming: My bad. Unintended.

Actually wanted to remove in the call checkForUpdates. For this call len(peers) should be sufficient right?
``` go
checkForUpdates(dockerVersion, router, uint(len(peers)+1)
```

bboreham: Yeah, I would leave off the `+1` there.

PR #3447, opened 2018-11-06
Add a note on how to secure the configuration.

### LINKED ISSUE #3446: [docs] improve the k8s docs on encryption

The [k8s docs](https://www.weave.works/docs/net/latest/kubernetes/kube-addon/) briefly mention the `password-secret` but provide no explanation on what role it plays. If installed in a default manner, weavenet runs unencrypted and again, the consequences of that aren't immediately clear.

I would suggest to improve the documentation to
- clearly state that the default setup is not encrypted and explaining what that means, which attack scenarios weave is exposed to
- clearly state how to secure weave net

Most people tend to use the most simple configuration without bothering with options unless absolutely necessary. Being one of the major CNI plugins weavenet could do better in terms of explaining the security landscape of the default installation.

farcaller:
Fixes #3446

### DIFF
--- a/site/kubernetes/kube-addon.md
+++ b/site/kubernetes/kube-addon.md
@@ -33,7 +33,7 @@ with a single command:
 $ kubectl apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\n')"
 ```
 
-**Important:** this configuration won't enable encryption and might be unfit for production use depending on your circumstances. Read on to see the alteranatives.
+**Important:** this configuration won't enable encryption by default. If your data plane traffic isn't secured that will allow malicios actors to access your pod network. Read on to see the alternatives.
 
 After a few seconds, a Weave Net pod should be running on each
 Node and any further pods you create will be automatically attached to the Weave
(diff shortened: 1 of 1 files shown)


bboreham:
Given that Weave Net is the only popular container network to even offer cross-host encryption, I think "unfit for production use" is a bit strong.  We'd be implying that Flannel, Calico, etc., are unfit.

How about "this configuration won't enable encryption, which you might need for production use depending on your circumstances. Read on to see the alteranatives." ?

farcaller:
It's really hard to pick the correct wording for a product that I don't develop, sorry if my CL sounded too harsh.

If I understand it right, with encryption set to off the whole data plane is insecure, thus the cross-host encryption doesn't matter as a malicious agent can manipulate the data flows. Is the assumption incorrect? I might be misunderstanding the consequences of the "encryption: off" scenario.

bboreham:
My imagination is failing me for what sort of "manipulate the data flows" attack is possible with Weave Net that is not possible with any other L2 or L3 pod network given the same access to the hosts.

If you want to argue that TCP/IP in general is insecure, feel free, but I don't think the Weave Net instructions are a suitable place to do that.

Now, we might want to draw attention to the _control plane_ being unauthenticated absent a password; this would be a slightly longer story as it should mention the `--trusted-subnets` option to avoid a performance hit.

bboreham:
Encryption is intended to stop attackers reading the traffic on the wire.
Encryption takes time, so it slows things down, which is why it is off by default.

> I can potentially convince weave into allowing a compromised node into the cluster, that wouldn't be part of it but would be able to establish connections over the pod network

This matches what I said about the control plane being unauthenticated absent a password. Authentication and encryption are both done through the same password in Weave Net.

However I would expect most people to have a firewall in place preventing connections to Weave Net from the outside world. For instance, this is the default for all cloud providers.

farcaller:
> However I would expect most people to have a firewall in place preventing connections to Weave Net from the outside world

I'd think this expectation should be clearly stated in the documentation? Running weave net on a manually deployed kubeadm cluster I'd expect the docs to be crystal clear about the expectations on the default configuration. It's common to expect the CNI to be safe and secure by default.

PR #3454, opened 2018-11-21
prevent kubernetes node connecting to self

### LINKED ISSUE #3398: Peer connection attempts reported as failed

Peer connection attempts to self not to be reported as `failed`.

## Background

We got alarmed by these.  After some research, the takeway is that it's noise.  However, also saw a lot of issues about peers not being able to join shortly after `CONN_LIMIT`.  We don't scale nodes much, so less of a concern for us, but still thought we should pay attention to this metric just in case.  Our workaround is simply to treat N failed connections for N nodes in the cluster as normal.

## Some Ideas

1. [Adding a new state](https://github.com/weaveworks/weave/pull/1318#issuecomment-131999338) just for this might be a no-go.
1. Adding special handling [in the metrics reporting](https://github.com/weaveworks/weave/blob/e2d6ee4223751363976b3c1f20f2d9c17e6a99da/prog/weaver/metrics.go#L51) to ignore when `Info ~ "cannot connect to ourself"` might be too brittle.
1. Patch something in the kubernetes add-on so that a node doesn't try to connect to itself.  The hope is that something in kubernetes land can better inform weave which peers to try.

If core devs provide a bit of guidance, we might be able to submit a PR.

## Versions:
- weave 2.4.0
- kubernetes 1.11

## Logs:
```
kubectl -n kube-system exec weave-net-fqv5t -c weave -- /home/weave/weave --local report | jq -r '.Router.Connections[] | select(.State != "established")'
{
  "Address": "10.210.1.78:6783",
  "Outbound": true,
  "State": "failed",
  "Info": "cannot connect to ourself, retry: never",
  "Attrs": null
}
```

murali-reddy:
prevent kubernetes node connecting to self by excluding the node IP from the list of the peers passed to weaver

Fixes #3398

### DIFF
--- a/prog/kube-utils/main.go
+++ b/prog/kube-utils/main.go
@@ -16,14 +16,15 @@ import (
 	"time"
 
 	"github.com/vishvananda/netlink"
-	weaveapi "github.com/weaveworks/weave/api"
-	"github.com/weaveworks/weave/common"
 	"golang.org/x/sys/unix"
 	api "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+
+	weaveapi "github.com/weaveworks/weave/api"
+	"github.com/weaveworks/weave/common"
 )
 
 type nodeInfo struct {
--- a/prog/kube-utils/main.go
+++ b/prog/kube-utils/main.go
@@ -15,14 +15,15 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


prog/kube-utils/main.go
```diff
@@ -15,14 +15,15 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/vishvananda/netlink"
+	weaveapi "github.com/weaveworks/weave/api"
+	"github.com/weaveworks/weave/common"
+	"golang.org/x/sys/unix"
```
bboreham: We have a bit of a convention where imports are split into three blocks: first Go standard library, then imports from outside the repo, then imports from inside the repo.

PR #3291, opened 2018-04-30
Remove subprocess for changing netns

bboreham:
Now that Go 1.10 has removed the issue which made us do this.

Less faffing around, more efficient.

PR #3459, opened 2018-11-23
handle gracefully add/delete of vethwedu dummy interface

### LINKED ISSUE #3414: Weave Net Daemonset fails to restart pod due to existing dummy interface
The Weave Net Daemonset, who controls the weave-net pod in each node, should be able to restart a pod in a node when it fails / stopped for any reason.

## What happened?
The weave-net pod gets Error and CrashLoopBack status and unable to function again, until I terminate that node.

## How to reproduce it?
SSH into a node, use docker command to kill the weave-net container. Of course, this is just for re-produce. In fact, when on our production cluster, we sometimes meet the situation when weave-net crashes on a node and don't know why.

The logs point out that weave-net fails to create dummy interface:

```
FATA: 2018/09/26 05:08:04.369497 creating dummy interface: file exists
```

I have a small investigation, and it looks like the bug comes from `net/bridge.go`, in function `initPrep`, at those lines:

```golang
	dummy := &netlink.Dummy{LinkAttrs: netlink.NewLinkAttrs()}
	dummy.LinkAttrs.Name = "vethwedu"
	if err = netlink.LinkAdd(dummy); err != nil {
		return errors.Wrap(err, "creating dummy interface")
	}
```

Before the weave-net starts, it creates a dummy interface object, and when my pod starts, the interface already exists, checked with `ip link | grep vethwedu` command. It looks like in the previous session of weave-net, it fails to delete this dummy interface, or it is killed **before** deleting it. When I delete the dummy manually with `ip link delete vethwedu`, the pod runs smoothly and back to normal.

Adding a small check and delete if the dummy exists before creating a new one would solve this problem. Is it a good solution for this? If that's okay, I'll open an PR.

```golang
	if existingDummy, err = netlink.LinkByName("vethwedu"); err == nil {
            if err := netlink.LinkDel(existingDummy); err != nil {
              return errors.Wrap(err, "deleting existing dummy interface")
	    }
	}
        //...
```

I run our Kubernetes cluster on AWS, using KOPS.

## Versions:
```
$ weave version: 2.4.1

$ docker version
Client:
 Version:      17.03.2-ce
 API version:  1.27
 Go version:   go1.7.5
 Git commit:   f5ec1e2
 Built:        Tue Jun 27 02:09:56 2017
 OS/Arch:      linux/amd64

$ uname -a
Linux ip-172-50-52-229 4.4.121-k8s #1 SMP Sun Mar 11 19:39:47 UTC 2018 x86_64 GNU/Linux

$ kubectl version
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.8", GitCommit:"c138b85178156011dc934c2c9f4837476876fb07", GitTreeState:"clean", BuildDate:"2018-05-21T18:53:18Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

## Logs:
```

DEBU: 2018/09/26 05:08:04.266053 [kube-peers] Checking peer "7a:1f:1c:b2:b7:7e" against list &{[{ea:43:25:46:a3:95 ip-172-50-61-167.ap-southeast-2.compute.internal} {16:ac:8c:71:55:20 ip-172-50-80-215.ap-southeast-2.compute.internal} {3e:af:c2:26:08:00 ip-172-50-110-136.ap-southeast-2.compute.internal} {2e:04:f7:c2:42:71 ip-172-50-32-17.ap-southeast-2.compute.internal} {32:4d:7c:65:31:8d ip-172-50-50-186.ap-southeast-2.compute.internal} {ae:92:06:fd:b3:e7 ip-172-50-93-2.ap-southeast-2.compute.internal} {7e:69:df:8f:8f:17 ip-172-50-41-197.ap-southeast-2.compute.internal} {62:aa:0a:e8:65:96 ip-172-50-109-73.ap-southeast-2.compute.internal} {f6:30:3e:1d:4b:8b ip-172-50-113-192.ap-southeast-2.compute.internal} {7a:1f:1c:b2:b7:7e ip-172-50-52-229.ap-southeast-2.compute.internal} {12:b7:c5:3d:f4:82 ip-172-50-67-61.ap-southeast-2.compute.internal} {aa:0d:6b:9c:56:9b ip-172-50-44-14.ap-southeast-2.compute.internal} {82:e9:f1:ce:c5:29 ip-172-50-58-155.ap-southeast-2.compute.internal} {26:a8:11:0d:76:e2 ip-172-50-33-242.ap-southeast-2.compute.internal}]}
INFO: 2018/09/26 05:08:04.297726 Command line options: map[ipalloc-range:100.96.0.0/11 port:6783 docker-api: expect-npc:true host-root:/host nickname:ip-172-50-52-229.ap-southeast-2.compute.internal conn-limit:100 db-prefix:/weavedb/weave-net http-addr:127.0.0.1:6784 metrics-addr:0.0.0.0:6782 name:7a:1f:1c:b2:b7:7e datapath:datapath ipalloc-init:consensus=14 no-dns:true]
INFO: 2018/09/26 05:08:04.297772 weave  2.3.0
FATA: 2018/09/26 05:08:04.369497 creating dummy interface: file exists
```

murali-reddy:
use `LinkAddIfNotExist` to add, dummy interface, and delete dummy interface in `defer` block

fixes #3414

### DIFF
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -55,6 +55,7 @@ datapath of old kernel versions (https://github.com/weaveworks/weave/issues/1577
 
 const (
 	WeaveBridgeName  = "weave"
+	WeaveDummyIfName = "vethwedu"
 	DatapathName     = "datapath"
 	DatapathIfName   = "vethwe-datapath"
 	BridgeIfName     = "vethwe-bridge"
@@ -328,8 +329,8 @@ func (b bridgeImpl) initPrep(config *BridgeConfig) error {
 	// instead we create a temporary interface with the desired MTU,
 	// attach that to the bridge, and then remove it again.
 	dummy := &netlink.Dummy{LinkAttrs: netlink.NewLinkAttrs()}
-	dummy.LinkAttrs.Name = "vethwedu"
-	if err = netlink.LinkAdd(dummy); err != nil {
+	dummy.LinkAttrs.Name = WeaveDummyIfName
+	if err = LinkAddIfNotExist(dummy); err != nil {
 		return errors.Wrap(err, "creating dummy interface")
 	}
 	if err := netlink.LinkSetMTU(dummy, config.MTU); err != nil {
@@ -338,11 +339,16 @@ func (b bridgeImpl) initPrep(config *BridgeConfig) error {
 	if err := netlink.LinkSetMasterByIndex(dummy, b.bridge.Attrs().Index); err != nil {
 		return errors.Wrap(err, "setting dummy interface master")
 	}
-	if err := netlink.LinkDel(dummy); err != nil {
-		return errors.Wrap(err, "deleting dummy interface")
-	}
-
-	return nil
+	defer func() {
+		var dummyIf netlink.Link
+		dummyIf, err = netlink.LinkByName(WeaveDummyIfName)
+		if err == nil {
+			if err = netlink.LinkDel(dummyIf); err != nil {
+				err = errors.Wrap(err, "deleting dummy interface")
+			}
+		}
+	}()
+	return err
 }
 
 func (b bridgeImpl) init(config *BridgeConfig) error {


net/bridge.go
```diff
@@ -338,11 +339,16 @@ func (b bridgeImpl) initPrep(config *BridgeConfig) error {
 	if err := netlink.LinkSetMasterByIndex(dummy, b.bridge.Attrs().Index); err != nil {
 		return errors.Wrap(err, "setting dummy interface master")
 	}
-	if err := netlink.LinkDel(dummy); err != nil {
-		return errors.Wrap(err, "deleting dummy interface")
-	}
-
-	return nil
+	defer func() {
```
bboreham: I'm a bit puzzled here.  The issue talks about using `defer` right after the interface is created.  But here you're two operations later.

Also you're right before `return` so not much value from `defer`.

Am I missing something?

murali-reddy: So earlier (below snippet), after link creation, failure in either `LinkSetMTU` or `LinkSetMasterByIndex` results in not deleting the interface. When Weave pod restarts, it fails on `LinkAdd` and goes in crashback loop.

```go
	if err = netlink.LinkAdd(dummy); err != nil {
		return errors.Wrap(err, "creating dummy interface")
	}
	if err := netlink.LinkSetMTU(dummy, config.MTU); err != nil {
		return errors.Wrapf(err, "setting dummy interface mtu to %d", config.MTU)
	}
	if err := netlink.LinkSetMasterByIndex(dummy, b.bridge.Attrs().Index); err != nil {
		return errors.Wrap(err, "setting dummy interface master")
	}
	if err := netlink.LinkDel(dummy); err != nil {
		return errors.Wrap(err, "deleting dummy interface")
	}
```

murali-reddy: Sorry my bad. It should be right after `LinkAdd`, will update.

bboreham:
Trying various versions reveals the veth goes away with the container when run with Weave Net 2.3.0, and does not go away with 2.4.0.

This is the result of `ip link show` with 2.3.0:

```
89: vethwepl1869@if88: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1376 qdisc noqueue master weave state UP mode DEFAULT group default 
    link/ether f2:2f:46:37:1e:55 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

and 2.4.0:

```
100: vethwepl3241@if99: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1376 qdisc noqueue master weave state UP mode DEFAULT group default 
    link/ether 0e:74:97:80:97:d7 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

Can't spot any relevant difference.

Looking at the changes in 2.4.0, I suspect #3291 - where we change namespace differently.  I see the veth does go away if I stop the `weaver` process.

net/bridge.go
```diff
@@ -328,21 +329,26 @@ func (b bridgeImpl) initPrep(config *BridgeConfig) error {
...
 	dummy := &netlink.Dummy{LinkAttrs: netlink.NewLinkAttrs()}
-	dummy.LinkAttrs.Name = "vethwedu"
-	if err = netlink.LinkAdd(dummy); err != nil {
+	dummy.LinkAttrs.Name = WeaveDummyIfName
+	if err = LinkAddIfNotExist(dummy); err != nil {
 		return errors.Wrap(err, "creating dummy interface")
 	}
+	defer func() {
+		var dummy netlink.Link
+		dummy, err = netlink.LinkByName(WeaveDummyIfName)
```
bboreham: can we just re-use the `dummy` variable that was initialized earlier, instead of creating a new one?

bboreham: I think you don't even have to assign to it - the initialization earlier leaves `dummy` in a state where you can just use it.  Unless I'm missing something?

murali-reddy: Right. I changed the logic to re-use `dummy`. PTAL

net/bridge.go
```diff
@@ -327,22 +328,24 @@ func (b bridgeImpl) initPrep(config *BridgeConfig) error {
...
-	if err = netlink.LinkAdd(dummy); err != nil {
+	var dummy netlink.Link
+	dummy = &netlink.Dummy{LinkAttrs: netlink.NewLinkAttrs()}
+	dummy.Attrs().Name = WeaveDummyIfName
+	if err = LinkAddIfNotExist(dummy); err != nil {
 		return errors.Wrap(err, "creating dummy interface")
 	}
+	defer func() {
+		if err = netlink.LinkDel(dummy); err != nil {
+			err = errors.Wrap(err, "deleting dummy interface")
```
bboreham: I think assigning to the local variable `err` at this point has no effect.
If `err` is made a named return value from `initPrep` it would work, but suppose both `LinkSetMTU` and `LinkDel` fail, we'd only see one error.
Maybe use a different `err` to record the `LinkDel()` result and use `Wrap()` to add the previous error (if any).

bboreham: Hmm, `Wrap()` can only wrap one error.  Maybe just add the text of this one to the previous one, if it exists.

PR #3582, opened 2019-01-15
Check and report the error return from getKubePeers()

### LINKED ISSUE #3581: "failed to get peers" with no explanation
#3400 broke error reporting from `kube-utils` - the error return from `getKubePeers()` is never looked at.

This was mentioned in #3420 but the underlying problem was solved before we looked at the error reporting issue.

I'm quite surprised this wasn't picked up by any of the linting tools we run.

bboreham:
Fixes #3581


murali-reddy:
LGTM

PR #3597, opened 2019-02-12
small grammatical changes to readme

dimitropoulos:
Just updating some language that might not have been looked at in a while judging by https://github.githistory.xyz/weaveworks/weave/blob/master/README.md


abuehrle:
LGTM  Thanks @dimitropoulos  good catch on all fronts.

PR #3603, opened 2019-02-27
Improve 'expecting PMTU update' message with address

bboreham:
Debugging an issue yesterday I realised this message is nearly useless because you have no idea which connection it refers to.

Also remove `EMGSIZE` because it catches the eye and looks like an error in the log file.

Maybe we should just remove this message, or move it to debug level?  I'm not sure why we need it.


murali-reddy:
LGTM

bboreham:
After discussion IRL I moved the message to debug level.

PR #3629, opened 2019-04-09
gossip broadcast updated ring on donating a range to other peer

### LINKED ISSUE #3628: allocator should broadcast updated ring after donating space

When a node donates tokens to peers, it is expected to reflect the updated ring across the nodes in the cluster.

## What happened?

It is noticed that other peers occasionally does not see updated tokens in the ring and continue to see the token as owned by previous donor owner. It is observed delay could be in the order of minutes.

During this window of time its possible a conflicting operation can be performed leaving the ring in inconsistent state.

When debugging #3444 I found the root cause to be this issue. 

## How to reproduce it?

Just keep repeating `kubectl delete` a node then rejoin the node to the cluster, you will see updated ring on the node which donated the range, but occasionally rest of the nodes would see updated ring after long delay.

murali-reddy:
Fixes #3628


murali-reddy:
rebased to 2.5 branch

PR #3635, opened 2019-04-24
bump up the version of entry by large value so that transfer of range takes precedence

murali-reddy:
bump up the version of entry by large value so that transfer of range takes precedence (if there is any conflict in merging entry) over other operations on the entry
that increments the version

Fixes #3612


murali-reddy:
rebased to 2.5 branch

Issue #2084, opened 2016-03-22
IPAM self-preservation prevents convergence


bboreham:
If, in the absence of persistence, a peer gives away some space, then dies and restarts and reconnects to a peer that has not heard about the donation, when it receives the real information later it will object and drop the connection.

Repro: in the context of `weaveworks/weave/test` (same script as #2083):

````
#! /bin/bash

. ./config.sh

start_suite "IPAM panic"

weave_on $HOST1 launch-router --no-discovery
weave_on $HOST1 expose
weave_on $HOST2 launch-router --no-discovery $HOST1
weave_on $HOST3 launch-router --no-discovery $HOST2
weave_on $HOST3 expose
weave_on $HOST1 stop-router
weave_on $HOST1 launch-router --no-discovery $HOST2
````

wait a couple of minutes, then look at HOST1's weave logs:

````
INFO: 2016/03/22 11:50:25.219476 Command line options: map[nickname:host1 port:6783 http-addr:127.0.0.1:6784 ipalloc-range:10.32.0.0/12 dns-listen-address:172.17.0.1:53 name:46:98:1c:22:6d:b7 no-discovery:true datapath:datapath dns-effective-listen-address:172.17.0.1]
INFO: 2016/03/22 11:50:25.220158 Communication between peers is unencrypted.
INFO: 2016/03/22 11:50:25.222141 Our name is 46:98:1c:22:6d:b7(host1)
INFO: 2016/03/22 11:50:25.222186 Initial set of peers: [192.168.48.12]
INFO: 2016/03/22 11:50:25.223175 Docker API on unix:///var/run/docker.sock: &[GoVersion=go1.5.3 Os=linux Arch=amd64 KernelVersion=4.2.0-34-generic BuildTime=2016-03-10T15:59:07.784447681+00:00 Version=1.10.3 ApiVersion=1.22 GitCommit=20f81dd]
INFO: 2016/03/22 11:50:25.223201 Assuming quorum size of 2
INFO: 2016/03/22 11:50:25.223241 [allocator 46:98:1c:22:6d:b7] Initialising via deferred consensus
INFO: 2016/03/22 11:50:25.224131 Listening for DNS queries on 172.17.0.1
INFO: 2016/03/22 11:50:25.224162 Sniffing traffic on datapath (via ODP)
INFO: 2016/03/22 11:50:25.225853 Listening for HTTP control messages on 127.0.0.1:6784
[...]

Issue #3627, opened 2019-04-09
debug logging for IPAM ring and entry updates


murali-reddy:
## What you expected to happen?

When users reports issues relates IPAM token conflicts for e.g. https://github.com/weaveworks/weave/issues/3600, https://github.com/weaveworks/weave/issues/3569
its hard to figure what went wrong.

We already have logs for operations like transfer of tokens, granting a range to other peers etc, but some issues are related to conflict of tokens version, and owner etc. It would be helpful to have debug logs for IPAM operations on ring and entries which update peer and version of ring entries. 

## What happened?

There is no trace of how entry in IPAM ring ends up with certain version and owner

PR #3637, opened 2019-04-30
resolve inconsistencies detected during ring merge

### LINKED ISSUE #1962: resolve inconsistencies detected during ring merge
The following are resolvable inconsistencies that currently cause a ring merge to fail, which in turn causes the connection on which the "other" ring was transmitted to be dropped...
- we receive an entry that splits one of our ranges, giving some of it away to another peer. This is in fact acceptable provided that we have no allocations in the range that got given away.
- we receive an entry with a token and version identical to an entry we have, but holding different content. We could apply a simple tie-break here, e.g. pick the entry with the highest free count and, if that is the same, lowest peer id. The only time we should error is when this tie-breaking ends up giving away a range we own and which has allocations.
- we receive an entry with a token that is equal to one of the entries we have, and has a newer version, but the entry we hold belongs to us. i.e. we've received an update to one of our own tokens, which shouldn't happen. We can, however, accept the received entry if either it belongs to a different peer and we do not have allocations in the range effectively given away, or it belongs to our own peer, in which case we should set our version to the one received plus one, effectively imposing our existing entry.

The 2nd of these is particularly important to address, at least partially, since it can cause a merge to fail on a peer for entries involving two other peers. Fixing just that would entail applying the tie-break as described and fail if that picks the received entry and the entry we hold belongs to us.

murali-reddy:
implements tie-breaker suggestions in #1962 and unit-tests to test the behaviour

Fixes #1962

### DIFF
--- a/ipam/ring/ring.go
+++ b/ipam/ring/ring.go
@@ -326,9 +326,14 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName, r *Ring) (result e
 				if mine.Peer == ourPeer {
 					// We received update to our own tokens accept the received entry
 					// if either it belongs to a different peer and we do not have allocations
-					// in the range effectively given away, or it belongs to our own peer
-					if theirs.Peer != ourPeer && !checkEntryHasAllocations(mine, es.entry(i+1)) || theirs.Peer == ourPeer {
+					// in the range effectively given away, or it belongs to our own peer, in
+					// which case we should set our version to the one received plus one,
+					// effectively imposing our existing entry.
+					if theirs.Peer != ourPeer && !checkEntryHasAllocations(mine, es.entry(i+1)) {
 						addTheirs(*theirs)
+					} else if theirs.Peer == ourPeer {
+						mine.Version = theirs.Version + 1
+						addToResult(*mine)
 					} else {
 						err = errNewerVersion(mine, theirs)
 						return
@@ -357,6 +362,16 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName, r *Ring) (result e
 			return
 		}
 		addTheirs(*theirs)
+	}
+
+	// reset the free space for entries if there is invalid free space
+	// due to accepting an unexpected update from the peers
+	for i := 0; i < len(result); i++ {
+		distance := r.distance(result.entry(i).Token, result.entry(i+1).Token)
+		if result.entry(i).Peer == r.Peer && result.entry(i).Free > distance {
+			// case that can arise when a range that we own that got split and had no allocations
+			result.entry(i).Free = distance
+		}
 	}
 
 	return
--- a/ipam/ring/ring.go
+++ b/ipam/ring/ring.go
@@ -302,7 +302,7 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName, r *Ring) (result e
 			// insert, checking that a range owned by us hasn't been split
 			if previousOwner != nil && *previousOwner == ourPeer && theirs.Peer != ourPeer {
 				// check we have no allocations in the range that got split
-				if checkEntryHasAllocations(es.entry(i-1), mine) {
+				if checkEntryHasAllocations(theirs, mine) {
 					err = errEntryInMyRange(theirs)
 					return
 				}
--- a/ipam/ring/ring.go
+++ b/ipam/ring/ring.go
@@ -292,6 +292,7 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName, r *Ring) (result e
 	var i, j int
 	for i < len(es) && j < len(other) {
 		mine, theirs = es[i], other[j]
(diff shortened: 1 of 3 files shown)


ipam/ring/ring.go
```diff
@@ -292,17 +303,38 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName) (result entries, u
...
 			case mine.Version < theirs.Version:
-				if mine.Peer == ourPeer { // We shouldn't receive updates to our own tokens
-					err = errNewerVersion(mine, theirs)
-					return
+				if mine.Peer == ourPeer {
+					// We received update to our own tokens accept the received entry
+					// if either it belongs to a different peer and we do not have allocations
+					// in the range effectively given away, or it belongs to our own peer
+					if theirs.Peer != ourPeer && !checkEntryHasAllocations(mine, es.entry(i+1)) || theirs.Peer == ourPeer {
+						addTheirs(*theirs)
```
bboreham: This is not implementing what it says in #1962 for the latter case:
>  we should set our version to the one received plus one, effectively imposing our existing entry.

murali-reddy: I tried the suggestion earlier, but was failing for this case (not sure if this in-valid test scenario)

```
	// we receive an entry that splits one of our ranges, giving some of it
	// away to another peer. accept the entry provided that we have no allocations
	// in the range that got given away.
	ring1 = NewRing(start, end, peer1name)
	ring2 = NewRing(start, end, peer2name)
	ring1.Entries = []*entry{{Token: start, Peer: peer1name, Free: 128, Version: 1}, {Token: middle, Peer: peer2name, Free: 128}}
	ring2.Entries = []*entry{{Token: start, Peer: peer1name, Free: 8, Version: 2}, {Token: dot8, Peer: peer2name, Free: 120, Version: 1}, {Token: middle, Peer: peer2name, Free: 128}}
	require.NoError(t, merge(ring1, ring2))
```

` imposing our existing entry` would mean keeping the free count as it is.

bboreham: The `Allocator` calls `reportFreeSpace()` after every action in `actorLoop()`, which should correct the free space.
I can't immediately see why that test case fails.

bboreham: After lengthy discussion, I understand the test fails because `Merge()` checks validity of everything before returning.

There is another case where accepting the incoming data is incorrect: if we have allocated some IPs in the free space before the update arrives.

So, I'm thinking that recalculating the free space when accepting an unexpected update is the way to go.

murali-reddy: > recalculating the free space when accepting an unexpected update is the way to go
`ring` does not have access to `space` to re-calculate the free space (need invasive changes across allocator, ring, space and tests to accommodate this). Given the conflicting conditions (merging of un-allocated ranges) this PR is addressing, I made change to reset the `Free` to distance between the tokens. PTAL

bboreham:
Appears to be removing the debug logging added in #3627 - is that intended?

ipam/ring/ring.go
```diff
@@ -282,29 +300,57 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName) (result entries, u
 		case mine.Token > theirs.Token:
 			// insert, checking that a range owned by us hasn't been split
 			if previousOwner != nil && *previousOwner == ourPeer && theirs.Peer != ourPeer {
-				err = errEntryInMyRange(theirs)
-				return
+				// check we have no allocations in the range that got split
+				if checkEntryHasAllocations(es.entry(i-1), mine) {
```
bboreham: This doesn't seem quite right - suppose the tokens are A ... B ... C, and `theirs.Token` is B.
We want to check if there are allocations in the range B..C, but this code checks A..C.

murali-reddy: This is the case of range that got split. Its more like `mine` sees A...C and and `theirs.Token` is B with A<B< C. The check is at the point where `mine.Token` is C and `theirs.Token` is B and `checkEntryHasAllocations(es.entry(i-1), mine)` is checking over range A-C.

murali-reddy:
> Appears to be removing the debug logging added in #3627 - is that intended?

Unintended change due to rebase. Fixed it.

ipam/ring/ring.go
bboreham: That's what I meant to say - we don't mind if there are allocations in `A..B`, since that's staying with the same ownership; we want to check in `B..C` only since that's the bit that's changing.

murali-reddy: got it. Modified the check to only to range `B..C`

ipam/ring/ring.go
```diff
@@ -282,29 +301,57 @@ func (es entries) merge(other entries, ourPeer mesh.PeerName) (result entries, u
 		case mine.Token > theirs.Token:
 			// insert, checking that a range owned by us hasn't been split
 			if previousOwner != nil && *previousOwner == ourPeer && theirs.Peer != ourPeer {
-				err = errEntryInMyRange(theirs)
-				return
+				// check we have no allocations in the range that got split
+				if checkEntryHasAllocations(theirs, mine) {
```
bboreham: This checks `theirs.Free < size`, i.e. it is checking if _they_ have any allocations in the range.
I think, to do this correctly, there needs to be some way to check the actual allocations.
Maybe pass in a function that will do that?

murali-reddy: Added a function that uses `space` to achieve `checkEntryHasAllocations` for an entry in ring.

bboreham:
Does this fix #2084 ?

PR #3638, opened 2019-05-02
return error in case Weave CNI plug-in fails to perform CmdDel 

### LINKED ISSUE #3587: weave-net hangs - leaking IP addresses?

`/status` etc to work

## What happened?

`/status` (and all other endpoints) hang

Issuance of IP addresses to pods also seemed to stop on the affected nodes

## How to reproduce it?

Unknown. Happened on multiple nodes in our cluster

## Versions:

```
$ weave --local version
weave 2.5.0
$ docker version
$ uname -a
Linux kubernetes-kubernetes-cr0-17-1547767704 4.14.67-coreos #1 SMP Mon Sep 10 23:14:26 UTC 2018 x86_64 Linux
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.10", GitCommit:"098570796b32895c38a9a1c9286425fb1ececa18", GitTreeState:"clean", BuildDate:"2018-08-02T17:19:54Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.10", GitCommit:"098570796b32895c38a9a1c9286425fb1ececa18", GitTreeState:"clean", BuildDate:"2018-08-02T17:11:51Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}

```

## Logs:
or, if using Kubernetes:
```
```
(After triggering a `kill -ABRT`)

[weave-logs-1548156065-weave-net-4mdf6.txt](https://github.com/weaveworks/weave/files/2782585/weave-logs-1548156065-weave-net-4mdf6.txt)

## Network:

```
$ ip route
$ ip -4 -o addr
$ sudo iptables-save
```

murali-reddy:
return error in case Weave CNI plug-in fails to perform CmdDel so that kubelet will retry DEL (preventing IP leak)

Also if there are no IP's associated with container iD just return (so that its idempotent) with out raising error. With out this change, if kubelet fails to perform successful ADD, it would perform DEL (as part of cleanup) with containerid. Weave IPAM will return an error as there are no IP's associated, resulting in kubelet performing the operation unsuccessfully (resulting in retry's forever).

Fixes #3587

### DIFF
--- a/plugin/net/cni.go
+++ b/plugin/net/cni.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strings"
 
 	"github.com/containernetworking/cni/pkg/ipam"
 	"github.com/containernetworking/cni/pkg/skel"
@@ -206,7 +207,13 @@ func (c *CNIPlugin) CmdDel(args *skel.CmdArgs) error {
 		err = ipam.ExecDel(conf.IPAM.Type, args.StdinData)
 	}
 	if err != nil {
+		if strings.Contains(err.Error(), fmt.Sprintf("Delete: no addresses for %s", args.ContainerID)) {
+			// ignore the error in the case where there are no IP addresses associated with container
+			fmt.Fprintln(os.Stderr, "weave-cni:", fmt.Sprintf("Delete: no addresses for %s", args.ContainerID))
+			return nil
+		}
 		logOnStderr(fmt.Errorf("unable to release IP address: %s", err))
+		return err
 	}
 	return nil
 }


ipam/allocator.go
```diff
@@ -406,7 +406,7 @@ func (alloc *Allocator) Delete(ident string) error {
 func (alloc *Allocator) delete(ident string) error {
 	cidrs := alloc.removeAllOwned(ident)
 	if len(cidrs) == 0 {
-		return fmt.Errorf("Delete: no addresses for %s", ident)
+		return nil
```
bboreham: Maybe log this to stderr in case it is of interest to someone debugging?
In the typical case stderr is inherited from kubelet, i.e. the message will show up in the kubelet logs.

murali-reddy: Added debug log.

bboreham: Ah, sorry, I misread what you'd done.  If you want to swallow the error for CNI that code should be in the CNI plugin.
Unless I'm missing the reason why you'd want to do it for every caller?

murali-reddy: Actually this change goes into `weaver` process. Either called directly or through CNI plugin, either way to me this method need to be idempotent. Hence suppressing the error here.

bboreham: Yes, I'm missing why it needs to be idempotent for every caller.

murali-reddy: I have restricted the changes to CNI plugin only by ignoring the error. PTAL

bboreham: That is remarkably ugly :stuck_out_tongue_closed_eyes:  I guess we can go with it.

Did wonder about changing ipam to return 204 for the no-addresses case, but I see it currently returns 204 in every case!

murali-reddy: > That is remarkably ugly

Indeed. Could not think more idiomatic way.  Thought of adding a custom error type but call crosses process boundaries.

murali-reddy:
rebased to 2.5

Issue #3464, opened 2018-11-29
ping fails until network is setup


damoon:
## Use case
I updated my setup from weave 2.4.1 to weave 2.5.0.
All tests but one passed.
The test is to create a kuberntes job and ping a single time.

I tested with weave 2.4.1 and weave 2.5.0
kubernetes 1.10 1.11 1.12
all the same docker version 17.03.3-ce

The problem only happens with 2.5.0 not with 2.4.1.

Because of the found workaround (wait for 1 sec) i assume 2.5.0 is delayed by some small amount of time to setup the network.
Just to be clear, waiting for 1 sec all the time is not something i plan to do :D

## What you expected to happen?

running a job to ping should ping once and succeed.
```
# kubectl logs google-dns-kw6ss 
PING 8.8.8.8 (8.8.8.8): 56 data bytes
64 bytes from 8.8.8.8: seq=0 ttl=61 time=12.615 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 12.615/12.615/12.615 ms
```

## What happened?
pinging failed
```
# kubectl logs google-dns-ftlff 
PING 8.8.8.8 (8.8.8.8): 56 data bytes

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 0 packets received, 100% packet loss
```

## How to reproduce it?
running the job creates the same result reliable.

## Anything else we need to know?
```
apiVersion: batch/v1
kind: Job
metadata:
  name: google-dns
spec:
  template:
    metadata:
      name: google-dns
    spec:
      containers:
      - name: dig
        image: azukiapp/dig:0.3.0
        command:
          - sh
          - -c
          - "ping -c 1 8.8.8.8"
      restartPolicy: Never
```
kind of workaround is to add a artificial delay before the container starts it actual work
```
          - "sleep 1 && ping -c 1 8.8.8.8"
```

## Versions:

```
working:
$ weave version
weave script 2.4.1
weave 2.4.1

broken:
$ weave version
weave script 2.5.0
weave 2.5.0


$ docker version
Client:
 Version:      17.03.3-ce
 API version:  1.27
 Go version:   go1.7.5
 Git commit:   e19b718
 Built:        Thu Aug 30 01:04:51 2018
 OS/Arch:      linux/amd64

Server:
 Version:      17.03.3-ce
 API version:  1.27 (minimum version 1.12)
 Go version:   go1.7.5
[...]

PR #3639, opened 2019-05-03
lazily add iptables rule to drop egress traffic from the pods only if network policies are applied

murali-reddy:
does not address #3464 but just avoids the issue atleast for those who
don't use network policies

### DIFF
--- a/npc/controller.go
+++ b/npc/controller.go
@@ -1,9 +1,8 @@
 package npc
 
 import (
+	"fmt"
 	"sync"
-
-	goiptables "github.com/coreos/go-iptables/iptables"
 
 	"github.com/pkg/errors"
 	coreapi "k8s.io/api/core/v1"
@@ -39,7 +38,7 @@ type controller struct {
 
 	nss               map[string]*ns // ns name -> ns struct
 	nsSelectors       *selectorSet   // selector string -> nsSelector
-	defaultEgressDrop bool
+	defaultEgressDrop bool           // flag to track if base iptable rule to drop egress traffic is added or not
 }
 
 func New(nodeName string, ipt iptables.Interface, ips ipset.Interface) NetworkPolicyController {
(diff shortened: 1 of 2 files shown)


npc/controller.go
```diff
@@ -121,11 +124,24 @@ func (npc *controller) AddNetworkPolicy(obj interface{}) error {
 	npc.Lock()
 	defer npc.Unlock()
 
+	// lazily add default rule to drop egress traffic only when network policies are applied
+	if !npc.defaultEgressDrop {
+		npc.defaultEgressDrop = true
+		ipt, err := goiptables.New()
```
bboreham: can we use `npc.ipt`?

npc/controller.go
```diff
@@ -121,11 +124,24 @@ func (npc *controller) AddNetworkPolicy(obj interface{}) error {
...
+		if err != nil {
+			common.Log.Fatalf("Failed to create iptables handler %v", err.Error())
+		}
+		if err := ipt.Append(TableFilter, EgressChain,
+			"-m", "mark", "!", "--mark", EgressMark, "-j", "DROP"); err != nil {
+			common.Log.Errorf("Failed to add iptable rule to drop egress traffic from the pods by default due to %s", err.Error())
+			npc.defaultEgressDrop = false
```
bboreham: Is there any point in continuing here?  This policy isn't going to achieve much if we don't drop packets.

npc/controller.go
```diff
@@ -35,8 +37,9 @@ type controller struct {
 	ipt iptables.Interface
 	ips ipset.Interface
 
-	nss         map[string]*ns // ns name -> ns struct
-	nsSelectors *selectorSet   // selector string -> nsSelector
+	nss               map[string]*ns // ns name -> ns struct
+	nsSelectors       *selectorSet   // selector string -> nsSelector
+	defaultEgressDrop bool
```
bboreham: could add a short comment saying what this means

npc/controller.go
murali-reddy: Agree. Since this base rule is important only for egress network policies. I made change to restrict it only for egress network policies.

murali-reddy:
Addressed review comments. PTAL

PR #3647, opened 2019-05-28
add support for combination of podSelector and namespaceSelector in network policies

### LINKED ISSUE #3312: Support both podSelector and namespaceSelector in NetworkPolicy
Upcoming Kubernetes release will allow us to select pods in selected namespaces (https://github.com/kubernetes/kubernetes/pull/60452). We should support it in weave-npc.

murali-reddy:
- enhance selectorSpec to accommodate both pod and namespace selectors
- enhance analysePolicy to handle policies with both selectors

Fixes #3312

### DIFF
--- a/npc/analyser.go
+++ b/npc/analyser.go
@@ -12,12 +12,14 @@ import (
 
 func (ns *ns) analysePolicy(policy *networkingv1.NetworkPolicy) (
 	rules map[string]*ruleSpec,
-	nsSelectors, podSelectors map[string]*selectorSpec,
+	nsSelectors, podSelectors, namespacedPodSelectors map[string]*selectorSpec,
 	ipBlocks map[string]*ipBlockSpec,
 	err error) {
 
 	nsSelectors = make(map[string]*selectorSpec)
 	podSelectors = make(map[string]*selectorSpec)
+	namespacedPodSelectors = make(map[string]*selectorSpec)
+
 	ipBlocks = make(map[string]*ipBlockSpec)
 	rules = make(map[string]*ruleSpec)
 	policyTypes := make([]policyType, 0)
@@ -33,7 +35,7 @@ func (ns *ns) analysePolicy(policy *networkingv1.NetworkPolicy) (
 	// If empty, matches all pods in a namespace
 	targetSelector, err := newSelectorSpec(&policy.Spec.PodSelector, nil, policyTypes, ns.name, ipset.HashIP)
 	if err != nil {
-		return nil, nil, nil, nil, err
+		return nil, nil, nil, nil, nil, err
 	}
 
 	// To prevent targetSelector being overwritten by a subsequent selector with
@@ -52,7 +54,7 @@ func (ns *ns) analysePolicy(policy *networkingv1.NetworkPolicy) (
 
 			if !allPorts {
 				if err = checkForNamedPorts(ingressRule.Ports); err != nil {
-					return nil, nil, nil, nil, fmt.Errorf("named ports in network policies is not supported yet. "+
+					return nil, nil, nil, nil, nil, fmt.Errorf("named ports in network policies is not supported yet. "+
 						"Rejecting network policy: %s from further processing. "+err.Error(), policy.Name)
 				}
 			}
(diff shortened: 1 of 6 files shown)


npc/analyser.go
```diff
@@ -72,16 +72,22 @@ func (ns *ns) analysePolicy(policy *networkingv1.NetworkPolicy) (
 					var srcRuleHost ruleHost
 
 					// NetworkPolicyPeer describes a peer to allow traffic from.
-					// Exactly one of its fields must be specified.
-					if peer.PodSelector != nil {
-						srcSelector, err = newSelectorSpec(peer.PodSelector, nil, ns.name, ipset.HashIP)
+					if peer.PodSelector != nil && peer.NamespaceSelector != nil {
+						srcSelector, err = newSelectorSpec(peer.PodSelector, peer.NamespaceSelector, nil, ns.name, ipset.HashIP)
```
bboreham: These six lines seem identical to the next `else` clause, apart from substituting `nil` for `peer.NamespaceSelector`.

npc/analyser.go
```diff
@@ -137,17 +143,23 @@ func (ns *ns) analysePolicy(policy *networkingv1.NetworkPolicy) (
 					var dstRuleHost ruleHost
 
 					// NetworkPolicyPeer describes a peer to allow traffic to.
-					// Exactly one of its fields must be specified.
-					if peer.PodSelector != nil {
-						dstSelector, err = newSelectorSpec(peer.PodSelector, nil, ns.name, ipset.HashIP)
+					if peer.PodSelector != nil && peer.NamespaceSelector != nil {
+						dstSelector, err = newSelectorSpec(peer.PodSelector, peer.NamespaceSelector, nil, ns.name, ipset.HashIP)
```
bboreham: These six lines seem identical to the next `else` clause, apart from substituting `nil` for `peer.NamespaceSelector`.

npc/analyser.go
bboreham: Is it correct to pass `ns.name` here?  Don't we want a pod+namespace selector to work the same from any namespace?

murali-reddy:
> I've noted some superficial issues, but I got stuck trying to figure out the bigger picture.
Why do we end up with nsSelectors and podSelectors when the latter can also have a namespace?

thanks for the review. I modified code to make it more explicit. So now there are three book keeping selector sets serving different purpose. `nsSelectors` and `podSelectors` are the existing ones with no changes to interpretation. `namespacedPodsSelectors` is the new one to represent matching pods in matching namespaces

``` go
	nsSelectors             *selectorSet // reference to global selectorSet that is shared across the `ns`. Used to represent all pods in the matching namespaces
	podSelectors            *selectorSet // used to represent the matching pods in namespace respresented by this `ns`
	namespacedPodsSelectors *selectorSet // reference to global selectorSet that is shared across the `ns`. Used to represent matching pods in matching namespace
```

npc/selector.go
```diff
@@ -61,8 +75,15 @@ type selector struct {
 	spec *selectorSpec
 }
 
-func (s *selector) matches(labelMap map[string]string) bool {
-	return s.spec.selector.Matches(labels.Set(labelMap))
+func (s *selector) matches(podsLabelMap *map[string]string, namespaceLabelMap *map[string]string) bool {
```
bboreham: The use of pointer-to-map here is a little unusual.
I think I can see what you are doing, and I can't say it's wrong, but I think it at least deserves a comment saying what the two parameters mean.

In many cases the code would be cleaner with three different `matches()` methods.

npc/selector.go
```diff
@@ -12,25 +12,39 @@ import (
 )
 
 type selectorSpec struct {
-	key         string          // string representation (for hash keying/equality comparison)
-	selector    labels.Selector // k8s Selector object (for matching)
-	policyTypes []policyType    // If non-empty, then selectorSpec is a target selector for given policyTypes.
+	key               string          // string representation (for hash keying/equality comparison)
+	podSelector       labels.Selector // k8s Selector object (for matching pods)
+	namespaceSelector labels.Selector // k8s Selector object (for matching namespaces)
```
bboreham: What is the relationship between `namespaceSelector` and `nsName` ?
I'm wondering if `nsName` can be replaced by a selector matching that one namespace.

npc/selector.go
murali-reddy: added separate matches to make it cleaner

npc/selector.go
murali-reddy: nsName is just used to generate appropriate comments for the iptable rules. No relationship between namespaceSelector and nsName

bboreham: For the record, the key point I was missing is that "a selector matching that one namespace" is not something you can expect to construct, given an arbitrary namespace.  Selectors can only match on labels, and the namespace may not have any labels.

kfox1111:
I'm a bit confused. Say I have 3 namespaces. A, B and C. only namespace B is labeled 'name=B'

I was trying to get A -> A and B -> A and not C !-> A

If I register this snippet in namespace A
```
  ingress:
  - from:
    - podSelector: {}
```

I get the first part. A -> A. B !-> A, C !-> A. Working as expected.

If I try:
```
  - from:
    - namespaceSelector:
        matchLabels:
          name: B
      podSelector: {}
```

I get A -> A, B !-> A, C !-> A. A can talk to A, but B cant. This seems backwards according to the spec?

Then I try this one:
```
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          name: B
    - podSelector: {}
```

A can talk to A, B can talk to A, and C can not talk to A. This also seems wrong. Either it should be just a rule for B -> A, or it should be the equiv of the first two snippets together?

murali-reddy:
@kfox1111 What image did you use for testing? This feature is right now only in master. I just rechecked the scenarios I see its working as expected.

kfox1111:
Oh. Didn't realize this was an unreleased feature. That would explain it. Thanks.

PR #3648, opened 2019-06-03
Fix the incorrect case #4 in k8s-src-ip.md

murali-reddy:
With cluster IP, traffic is masquraded only when --masquerade-all option is specified
or --cluster-cidr is specified and the source is not from the cluster CIDR range

https://github.com/kubernetes/kubernetes/blob/v1.14.2/pkg/proxy/iptables/proxier.go#L842


bboreham:
Is it worth being specific about those flags, where it says "default kubeadm options" ?

PR #3655, opened 2019-06-30
Add tolerations for nodes tainted with NoExecute

tanishq-dubey:
If you add nodes with the following taint:
```
kubectl taint nodes node1 key1=value1:NoExecute
```
Weave is evicted from the node, and thus, the node will not be able to start any new pods due to it not being able to resolve network endpoints. By allowing the Weave DaemonSet to tolerate a `NoExecute` taint, Weave will run on these tainted nodes marked to avoid execution, as it should.

---

This is was found because I have a set of nodes that are marked with `NoExecute` so only certain applications that tolerate the taint can run on them. Turns out one of those certain applications became Weave. Pods that were running before I applied the taint were fine, but pods that were schedule after Weave was evicted from the node were unable to start and left in a `Pending` state.

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.6.yaml
@@ -156,6 +156,8 @@ items:
           tolerations:
             - effect: NoSchedule
               operator: Exists
+            - effect: NoExecute
+              operator: Exists
           volumes:
             - name: weavedb
               hostPath:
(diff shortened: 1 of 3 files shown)


tanishq-dubey:
Just for completeness, I did not modify https://github.com/weaveworks/weave/blob/master/prog/weave-kube/weave-daemonset.yaml#L22 and was wondering if that needed to be changed to include the `NoExecute`?

murali-reddy:
> Just for completeness, I did not modify https://github.com/weaveworks/weave/blob/master/prog/weave-kube/weave-daemonset.yaml#L22 and was wondering if that needed to be changed to include the NoExecute?

@tanishq-dubey its old manifest (though naming is counter-intutuve) for kubernetes cluster < 1.6.  So no need to change.

Issue #3675, opened 2019-07-31
Weave reconnect occasionally fails because InitSARemote can get called after Destroy


hpdvanwyk:
## What you expected to happen?
Weave should always reconnect after a network failure

## What happened?
Very occasionally when reconnecting `InitSARemote` gets called after `Destroy` has been called on a connection. This causes a valid xfrm policy to be updated to an old SPI and the connection stops working.

## How to reproduce it?
This was found while testing https://github.com/weaveworks/weave/pull/3669 . I used the following script:
```
#!/bin/bash

while true; do
        # disconnect network
        qm set 102 --net1 model=virtio,bridge=vmbr1,macaddr=62:40:98:FF:02:72,link_down=1
        sleep 55
        # reconnect network
        qm set 102 --net1 model=virtio,bridge=vmbr1,macaddr=62:40:98:FF:02:72,link_down=0
        sleep 10

        # check if weave still works
        if ssh rancher@172.30.12.191 ping 10.42.128.0 -c1; then
                date
                echo pass
        else
                sleep 10
                if ssh rancher@172.30.12.191 ping 10.42.128.0 -c1; then
                        date
                        echo pass
                else
                        date
                        echo broken
                        break
                fi
        fi
done
```

This bug only happened after about 26 hours.
Without https://github.com/weaveworks/weave/pull/3669 it is unlikely to trigger since using the same method would hit https://github.com/weaveworks/weave/issues/3666 much more frequently.

## Anything else we need to know?

Exact same setup as https://github.com/weaveworks/weave/issues/3666

## Versions:

```
/home/weave # ./weave --local version
weave 2.5.2

$ docker version
Client:
 Version:           18.06.3-ce
 API version:       1.38
 Go version:        go1.10.4
 Git commit:        d7080c1
 Built:             Wed Feb 20 02:24:22 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.3-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       d7080c1
  Built:            Wed Feb 20 02:25:33 2019
  OS/Arch:          linux/amd64
  Experimental:     false

$ uname -a
Linux hostyname0 4.14.122-rancher #1 SMP Tue May 28 01:50:21 UTC 2019 x86_64 GNU/Linux
$ kubectl version
[...]

PR #3676, opened 2019-07-31
Do not do ipsec.InitSARemote on a stopped fastDatapathForwarder.

hpdvanwyk:
It is possible for ipsec.InitSARemote to be called after ipsec.Destroy.
If an xfrm policy for that source/destination pair exists it will be
updated to the SPI from the Destroyed connection causing IPSEC to stop
working.
Fixes https://github.com/weaveworks/weave/issues/3675


bboreham:
Thanks!
CI ran successfully at https://circleci.com/workflow-run/b0c0a854-fcea-426b-a42a-d3a52609b814

PR #3679, opened 2019-08-02
Answer how Weave Net works

dholbach:
Actually Bryan answered it, when a reader of our website asked.


dholbach:
Not sure why the smoke test fails...

bboreham:
Rebased to 2.5 branch, so the doc update will go out straight away

PR #3708, opened 2019-10-09
Handle range wrap-around when checking for allocations

### LINKED ISSUE #3693: IPAM ring inconsistency reconcilliation fails when range crosses zero
E.g. at https://circleci.com/gh/weaveworks/weave/12331
https://circleci.com/gh/weaveworks/weave/12345
https://circleci.com/gh/weaveworks/weave/12377

It doesn't seem to say where it's failing:

```
>>> Running /home/circleci/src/github.com/weaveworks/weave/test/191_create_bridge_2_test.sh on [test-12331-0-3.us-central1-a.weave-net test-12331-0-4.us-central1-a.weave-net]
>>> Test /home/circleci/src/github.com/weaveworks/weave/test/191_create_bridge_2_test.sh finished after 33.9 secs with error: exit status 1
Re-create bridge after restart
f66a1219fda5e35241da6aa8e2771b6f0f5876e52a8be913970e82aff7a70202
fc08a2df8e7f6a43c9b9769aec5fde74993524eb5bce3f15c8fbc27c7083b71b
ac55848b4c8b317f2179ba0a9b0364d8e9480bf04d0164df52b9a559c0347b99
d2474e5825ccd172f8ad6e2814cc87594cf725dc90ec96567f13e710b393ff5a
05234413fec5275bc5e1279bcc19d67b2b82e77f8b8da5139cd9912345edcb7b
10.32.0.2
fcf6c0c74d838a0aae215c2b70c9f241ee19c0cf6aaa324d26fb6f84a334e817
10.32.0.1
```

Looks like it happens after the `expose`.

bboreham:
Fixes #3693 

`Ring` wraps around at zero but `Space` doesn't know about this, so we have to normalise ranges coming out of `Ring`.  In turn we may need to split one range into two, across the end of the `Ring`.

The unit tests passed because they are calling `distance` inside the Ring package, which understands the wrap-around.  We may want to make them more realistic, maybe move up to `ipam` package.

Also log the event when we do accept another peer's token, since that should be a rare event, and it's important if we've made a mistake.

### DIFF
--- a/ipam/allocator.go
+++ b/ipam/allocator.go
@@ -814,8 +814,14 @@ func (alloc *Allocator) sendRingUpdate(dest mesh.PeerName) {
 	alloc.gossip.GossipUnicast(dest, msg)
 }
 
-func (alloc *Allocator) checkRangeHasAllocations(r address.Range) bool {
-	return alloc.space.NumFreeAddressesInRange(r) != r.Size()
+func (alloc *Allocator) hasAllocations(rs []address.Range) bool {
+	for _, r := range rs {
+		alloc.debugf("hasAllocations: %v %v vs %v", r, alloc.space.NumFreeAddressesInRange(r), r.Size())
+		if alloc.space.NumFreeAddressesInRange(r) != r.Size() {
+			return true
+		}
+	}
+	return false
 }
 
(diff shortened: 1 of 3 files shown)


murali-reddy:
LGTM

PR #3711, opened 2019-10-10
Reduce time test script waits for IPs to be reclaimed

bboreham:
Now we have a callback when a node is removed, there should be no reason to wait 90 seconds.

The reclaim does wait up to 5 seconds before starting, so allowing 10 seconds in the script should be sufficient.

(I'm puzzled that #3399 put the timeout _up_)


bboreham:
Included in #3716

PR #3714, opened 2019-10-11
Lint cleanup

bboreham:
I ran golangci-lint and it pointed out a few things to clean up.

This isn't all of them, but nothing in the remaining set looked either important or easy.


bboreham:
Looks like the http changes have broken some smoke-tests.

PR #3715, opened 2019-10-11
Update weaveworks/mesh to v0.3

bboreham:
This release has a number of changes to defer updates by up to one second, to reduce the cpu usage of route recalculation and reduce network bandwith for topology updates.

The effect is it behaves much better on networks of over 100 nodes.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -686,12 +686,12 @@
   revision = "6b0aa22550d9325eb8f43418185859e13dc0de1d"
 
 [[projects]]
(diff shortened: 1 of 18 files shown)


murali-reddy:
LGTM.

Tested on 200 node cluster. When scaling up the cluster to 200 nodes all weave-net pods connections gets `established` and `fastdp` is used. But on testing the worst-case scenarios of deleting all weave-net pod or delete and re-install the weave-net ds does not result in all the weave-net pod getting established fully. But that can be addressed seperatley.

PR #3720, opened 2019-10-16
Update Prometheus client lib and switch to default metrics

bboreham:
There were some changes in the API for 1.1.0, so just drop our unique set-up and use the default.

In particular this means weaver will start reporting Go metrics such as heap size.


bboreham:
Subsumed into #3838

PR #3725, opened 2019-10-25
Retry IP allocator claims

bboreham:
Test 320 has been failing consistently in CI, because an IP address claim fails on first attempt.
Probably this is because topology gossip has been slowed down.

This PR allows claims to be retried a few times before giving up.

Also add the error message why an IP claim is denied, if we have one, to make troubleshooting easier.


murali-reddy:
LGTM

PR #3729, opened 2019-11-01
Remove the need for a third host in test 320

bboreham:
This was necessary when we didn't have ipam persistence, as the comment said "to remember the IPAM CRDT when we stop host2". 

But we have persistence to hold the CRDT so we don't need host3.

Rename the test so the test runner only gives it 2 hosts.


murali-reddy:
LGTM

PR #3735, opened 2019-11-11
Add weave-daemonset-k8s-1.11.yaml to release script

bboreham:
This is the manifest for Kubernetes 1.11 and above.
It was added to release 2.6 by hand, but this change will mean it gets included automatically next time.


murali-reddy:
LGTM

PR #3740, opened 2019-11-22
Raise Kubernetes CPU requests from 10m to 50m

bboreham:
This setting is passed by Kubelet to the Linux scheduler to ask for a share of available CPU, so we should ask for a bit more than 1% of 1 CPU. Otherwise if a CPU is contended Weave Net will get throttled frequently.

Weave Net in fastdp mode shouldn't need a lot of CPU most of the time, but at startup there is quite a lot to do.

The setting is also used by the Kubernetes scheduler to decide how much will fit on a node, so setting it much higher will reserve CPU that we shouldn't need most of the time.


murali-reddy:
LGTM

PR #3747, opened 2019-12-13
support Iptables 1.8 with legacy and nf_tables support

### LINKED ISSUE #3465: Weave Net breaks when host OS uses iptables 1.8
## What happened?

I installed Weave Net on a test cluster where the host OS is Debian Buster (aka Debian testing, rolling distro with the ~latest version of everything). After installing Weave, pod<>pod and pod<>internet communication is completely broken, no traffic flows at all, even between pods on the same machine.

I root-caused it to an incompatibility in iptables versions between the weave pods and the host OS. Debian Buster now ships iptables 1.8. It has a major change, which is that the `iptables` command is now a translating facade on top of nftables, i.e. creating rules with `iptables` or `iptables-restore` actually programs nf_tables in the kernel.

OTOH, the weave pod contains iptables 1.6, the previous stable release which programs the "classic" iptables netfilter stack. So, docker on the host OS ends up programming nf_tables rules (because it uses the host iptables 1.8), and weave ends up programming legacy iptables rules (because it uses iptables 1.6). For some reason I don't fully understand, having both programmed causes packets to get dropped instead of forwarded on the host, before the packets get transmitted to the target container.

## How to reproduce it?

I filed extensive reproduction steps in a sibling bug with the Calico folks, please refer to https://github.com/projectcalico/calico/issues/2322#issuecomment-443402871 . The only changes for weave are to use Weave's pod-network-cidr in kubeadm, obviously install weave instead of calico, and then some of the output changes slightly like different interface names. Everything else, including the core failure mode and the fix, plays out the same.

Additionally, the hacky steps to verify that iptables 1.8 is the problem are at https://github.com/projectcalico/calico/issues/2322#issuecomment-443405705 - basically hackily overwrite the iptables binaries with the ones from debian stable (which still uses 1.6), reboot the machine, and Weave starts working perfectly again.

There's also a bug tracking similar problems in core k8s, at https://github.com/kubernetes/kubernetes/issues/71305 . In core k8s this mismatch breaks kube-proxy, but it's the exact same root cause, mismatched iptables versions.

## Versions:

$ weave version

Whichever the latest one is - my test harness is running calico atm, can't look it up.

$ docker version

```
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        e68fc7a
 Built:             Tue Aug 21 17:24:43 2018
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       e68fc7a
  Built:            Tue Aug 21 17:23:06 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

$ uname -a

```
Linux cluster1-controller 4.18.0-2-amd64 #1 SMP Debian 4.18.10-2 (2018-11-02) x86_64 GNU/Linux
```

$ kubectl version
```
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.3", GitCommit:"435f92c719f279a3a67808c80521ea17d5715c66", GitTreeState:"clean", BuildDate:"2018-11-26T12:57:14Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.3", GitCommit:"435f92c719f279a3a67808c80521ea17d5715c66", GitTreeState:"clean", BuildDate:"2018-11-26T12:46:57Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
```

murali-reddy:
Fixes #3465

- upgrade to alpine image (for amd64 only) with iptables 1.8.3
- updated alpine image bundles both (legacy and nftable) version iptables, iptables-save, iptables-restore
- by default  iptables* commands soft linked to iptables-legacy*
- user can explicitly specify the iptable backed
- have borrowed back-end detection logic from kube-proxy

need to test with Kubernetes 1.17 which has kube-proxy in nft mode

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -6,6 +6,34 @@ modprobe_safe() {
     modprobe $1 || echo "Ignore the error if \"$1\" is built-in in the kernel" >&2
 }
 
+# Setup iptables backend to be legacy or nftable
+setup_iptables_backend() {
+    if [ -n "${IPTABLES_BACKEND}" ]; then
+      mode=$IPTABLES_BACKEND
+    else
+      # auto-detect if iptables backend mode to use if not specified explicitly
+      num_legacy_lines=$( (iptables-legacy-save || true) 2>/dev/null | grep '^-' | wc -l)
+      num_nft_lines=$( (iptables-nft-save || true) 2>/dev/null | grep '^-' | wc -l)
+      if [ "${num_legacy_lines}" -ge 10 ]; then
+        mode="legacy"
+      else
+        if [ "${num_legacy_lines}" -ge "${num_nft_lines}" ]; then
+          mode="legacy"
+        else
+          mode="nft"
+        fi
+      fi
+    fi
+    if [ $mode = "nft" ]; then
+      rm /sbin/iptables
+      rm /sbin/iptables-save
+      rm /sbin/iptables-restore
+      ln /sbin/iptables-nft /sbin/iptables
+      ln /sbin/iptables-nft-save /sbin/iptables-save
+      ln /sbin/iptables-nft-restore /sbin/iptables-restore
+    fi
+}
+
 # Check whether xt_set actually exists
 xt_set_exists() {
     # Clean everything up in advance, in case there's leftovers
@@ -23,6 +51,8 @@ xt_set_exists() {
     ipset destroy weave-kube-test
     [ -z "$NOT_EXIST" ] || (echo "\"xt_set\" does not exist" >&2 && return 1)
 }
+
+setup_iptables_backend
 
 # Default if not supplied - same as weave net default
 IPALLOC_RANGE=${IPALLOC_RANGE:-10.32.0.0/12}
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -28,9 +28,9 @@ setup_iptables_backend() {
       rm /sbin/iptables
       rm /sbin/iptables-save
       rm /sbin/iptables-restore
-      ln /sbin/iptables-nft /sbin/iptables
-      ln /sbin/iptables-nft-save /sbin/iptables-save
-      ln /sbin/iptables-nft-restore /sbin/iptables-restore
+      ln -s /sbin/iptables-nft /sbin/iptables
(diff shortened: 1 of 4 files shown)


bboreham:
Should we do anything for the non-Kubernetes case?  I.e. the `weave` script.

murali-reddy:
> need to test with Kubernetes 1.17 which has kube-proxy in nft mode

tested the auto detection logic with Kubernetes 1.17(kube-proxy and kubelet using iptables nft mode) and Debain buster (iptables with nftable as default)

> Should we do anything for the non-Kubernetes case? I.e. the weave script.

We could ask users to use iptables-legacy on the host ([for e.g.](iptables-legacy)) for the moment so even update to the base image should work fine. Alternatively we could soft-link to appropriate binaries  in the weave container during `weave launch`

prog/weave-kube/launch.sh
```diff
@@ -6,6 +6,34 @@ modprobe_safe() {
...
+        else
+          mode="nft"
+        fi
+      fi
+    fi
+    if [ $mode = "nft" ]; then
+      rm /sbin/iptables
+      rm /sbin/iptables-save
+      rm /sbin/iptables-restore
+      ln /sbin/iptables-nft /sbin/iptables
```
bboreham: Since Alpine appears to use symbolic links (`ln -s`), I think we should follow suit, just in case it causes confusion.

prog/weave-kube/launch.sh
```diff
@@ -6,6 +6,34 @@ modprobe_safe() {
     modprobe $1 || echo "Ignore the error if \"$1\" is built-in in the kernel" >&2
 }
 
+# Setup iptables backend to be legacy or nftable
+setup_iptables_backend() {
+    if [ -n "${IPTABLES_BACKEND}" ]; then
+      mode=$IPTABLES_BACKEND
```
bboreham: please add this var to docs

murali-reddy:
Updated base branch to 2.6 and addressed review comments.

PR #3756, opened 2020-01-20
Change NO_MASQ_LOCAL option to default ON

### LINKED ISSUE #3389: Feature request: Please make NO_MASQ_LOCAL=1 the default setting
Please make NO_MASQ_LOCAL=1 the default setting.

People expect that "ExternalTrafficPolicy: Local" behaves as described by the Kubernetes documentation. This currently not the case with Weave, which causes confusion. I lost hours troubleshooting why ExternalTrafficPolicy was not working as expected, and it was only due to pure luck and some creative googling that I discovered the problem was caused by Weave.

While some might argue that NO_MASQ_LOCAL is mentioned in the release notes of 2.4.0, It feels wrong that you need extra config parameters to make Weave work in a way that is compliant with the standard.

murali-reddy:
Fixes #3389

### DIFF
--- a/prog/weave-kube/launch.sh
+++ b/prog/weave-kube/launch.sh
@@ -67,6 +67,7 @@ DB_PREFIX=${DB_PREFIX:-/weavedb/weave-net}
 
 # Default for network policy
 EXPECT_NPC=${EXPECT_NPC:-1}
+NO_MASQ_LOCAL=${NO_MASQ_LOCAL:-1}
 
 # Ensure we have the required modules for NPC
 if [ "${EXPECT_NPC}" != "0" ]; then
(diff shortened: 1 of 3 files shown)


bboreham:
I just had a thought that we could make the script default ON unless it sees `NO_MASQ_LOCAL=0`.
This would simplify configuration and tie the change to a code release, rather than config.

murali-reddy:
> This would simplify configuration and tie the change to a code release, rather than config.

updated PR to be code only change

bboreham:
I think we should change the docs https://github.com/weaveworks/weave/blob/9f00f78d3b9d5a8a31fdd90ec691095028e2690a/site/kubernetes/kube-addon.md#L406

What about test 840, which is setting the variable?

PR #3770, opened 2020-02-11
fix weave-kube manifests

murali-reddy:
- add missing dnsPolicy: ClusterFirstWithHostNet in weave-daemonset-k8s-1.11.yaml
- update apiVersion from rbac.authorization.k8s.io/v1beta1 to rbac.authorization.k8s.io/v1

`dnsPolicy: ClusterFirstWithHostNet` was added in `weave-daemonset-k8s-1.8.yaml` but missing in `weave-daemonset-k8s-1.11.yaml`

### DIFF
--- a/prog/weave-kube/weave-daemonset-k8s-1.11.yaml
+++ b/prog/weave-kube/weave-daemonset-k8s-1.11.yaml
@@ -8,7 +8,7 @@ items:
       labels:
         name: weave-net
       namespace: kube-system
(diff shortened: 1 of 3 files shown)


bboreham:
I guess it's good to be consistent but do we ever actually do any DNS lookups in the typical Kubernetes config?

murali-reddy:
> but do we ever actually do any DNS lookups in the typical Kubernetes config

No. https://github.com/weaveworks/weave/pull/3692#issuecomment-527511653

Issue #3764, opened 2020-02-04
weave-npc blocking connections with valid network policy after a period of time (2.6.0)


naemono:
## What you expected to happen?

Similar to #3761, we are seeing traffic being blocked by weave-npc, but we *are* using network policies. I would expect traffic to not be blocked by NPC with valid network policy in place

## What happened?

We have seen now, consistently (about once every 1-2 weeks) traffic gets blocked between pods inside of a namespace where traffic was working fine earlier.  After we debugged the issue, and saw the ipset's on the host to not have valid entries for the pods, we restart weave on the host, and the ipsets become populated, and traffic continues to flow.

## How to reproduce it?

I wish we had an easy way to consistently reproduce this issue, but we are beginning to see this issue nearly every week within one specific cluster.

## Anything else we need to know?
cloud provider: aws
custom built cluster using in house automation.

## Versions:
```
# ./weave version
weave script 2.6.0
# docker version
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea838
 Built:             Wed Nov 13 07:29:52 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea838
  Built:            Wed Nov 13 07:28:22 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
# uname -a
Linux ip-10-0-173-150 4.15.0-1056-aws #58-Ubuntu SMP Tue Nov 26 15:14:34 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.2", GitCommit:"f6278300bebbb750328ac16ee6dd3aa7d3549568", GitTreeState:"clean", BuildDate:"2019-08-05T16:54:35Z", GoVersion:"go1.12.7", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #3772, opened 2020-02-13
Add configurable reconciliation loop for pods, namespaces, and networ…

naemono:
re #3764 

This adds a configurable reconciliation loop option, which defaults for the existing behavior of 0s (no reconciliation loop), for pods, namespaces, and network policies.  We have been testing these changes in 7 internal non-production clusters with a defined reconciliation interval of > 0s for the past week with no noticeable negative impacts , and will be rolling out to our production clusters in the upcoming week.  We also have not had a single reoccurrence of the issue in issues 3764 since this release.

### DIFF
--- a/npc/controller.go
+++ b/npc/controller.go
@@ -129,6 +129,11 @@ func (npc *controller) UpdatePod(oldObj, newObj *coreapi.Pod) error {
 	npc.Lock()
 	defer npc.Unlock()
 
+	// if the version hasn't changed, do nothing
+	if oldObj.GetResourceVersion() == newObj.GetResourceVersion() {
+		return nil
+	}
+
 	common.Log.Debugf("EVENT UpdatePod %s %s", js(oldObj), js(newObj))
 	return npc.withNS(oldObj.ObjectMeta.Namespace, func(ns *ns) error {
 		return errors.Wrap(ns.updatePod(oldObj, newObj), "update pod")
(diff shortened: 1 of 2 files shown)


bboreham:
Thanks for the PR!  Some unit tests failed; please take a look at https://circleci.com/gh/weaveworks/weave/13224

naemono:
> Thanks for the PR! Some unit tests failed; please take a look at https://circleci.com/gh/weaveworks/weave/13224

Looking now, thanks

naemono:
```
#!/bin/bash -eo pipefail
bin/provision_test_vms.sh
Cannot run smoke tests: no secret key

Exited with code exit status 1
```
is this normal?  Do I need to do something to get the smoke tests to run?

bboreham:
The smoke-tests run with credentials on Google Cloud (accessed via the "secret key"); we don't allow PRs from other repos to see those credentials.  I've pushed your branch to the main repo so it will run the smoke-tests.

bboreham:
I would like to clarify what this PR achieves. Take the "update pod" event: if a message comes in that the pod data is updated, but the data is the same as the previous version, then weave-npc will do nothing.

So this PR will help in the case that the Kubernetes data in memory is out of sync with the api-server, but does not help in the case that the iptables rules or ipsets are out of sync with the data.

Is that your understanding?  Is there any data that #3764 is caused by the first kind of mismatch rather than the second?

naemono:
Yes, that is a good point, it certainly wouldn't appear to have an effect if the ipsets were out of sync on the host itself during a reconciliation, as it's not doing a comparison.  But it would catch one of the 2 out of sync scenarios you reference.  Would you prefer that this PR includes data comparison during reconciliation?  That seems like it would make a lot of sense.

bboreham:
I think these two things are orthogonal and I always prefer to do unrelated changes in separate PRs.
Adding a periodic reconcilliation from the state of Kubernetes objects to iptables is a good idea, as discussed at #3764.

I would not call this one "reconcilliation"; Kubernetes [calls it "resync"](https://godoc.org/k8s.io/client-go/tools/cache#NewInformer).

bboreham:
@naemono are you likely to return to this?
Note some work at #3802 for recreating iptables rules when they get wiped out.

naemono:
I will also get back to this in next Sprint next week.  Sorry for delays.

naemono:
closing this as #3792 solves the actual, underlying problem of weave-npc application continuing to run when goroutine panics.

Issue #3771, opened 2020-02-12
weave-npc should gracefully restart upon go routine panic


Quentin-M:
Forking off this issue: https://github.com/weaveworks/weave/issues/3764 as per @bboreham's request.

TL;DR >

As of right now, when one of weave-npc's controller/go-routine panics, weave-npc will simply [log](https://github.com/kubernetes/apimachinery/blob/210bcde8e5a7715fe776460ca1df1a3f2738725e/pkg/util/runtime/runtime.go#L36) the panic rather than propagating it in order to restart the go-routine, or in order to restart weave-npc as a whole (thus potentially saving it from a panic loop if the memory structures are in an unexpected state. This will leave weave-npc running in a non-functioning state.

Furthermore, when weave-npc restarts, it incurs a 10s+ downtime as weave-npc [resets](https://github.com/weaveworks/weave/blob/b05c5f63c8feab1defffbf97d53e88915ef0d296/prog/weave-npc/main.go#L253) every IPSets/Rules, then re-creates them, instead of gracefully reconciling the host / desired / current states. The trouble is that, when a bad informer sends unexpected data (as per issue above), all weave-npc containers will crash at once, hence creating a full cluster downtime - potentially lengthened by the slowdown of the API due to sheer amount of requests.

/cc @murali-reddy

bboreham:
Note I said in #3764 that it is written not to "simply log".

"it appears to me that the intention is to re-raise the panic and hence exit the whole program. I am mystified why it keeps on running"

Quentin-M:
The code snippet explicitly intercepts the panic from the crashed routine with recover(), log and move on. Only if the "ReallyCrash" global variable is set to true, or if an additional handler is globally registered with a new panic or with an exit syscall, will it do anything besides logging. Code works as expected, behavior is however wrong.

Best Regards,
Quentin Machu,
Head of DevOps, BitMEX
+1 (415) 720 1243

bboreham:
> Only if the "ReallyCrash" global variable is set to true,

`ReallyCrash` defaults to `true`. Why would it change, in this context?

gobomb:
I found `ReallyCrash` is `true` but the process didn't crash. I am confused.

gobomb:
[...]

PR #3792, opened 2020-04-03
Terminate NPC application if any of the watches panic

naemono:
Catch panics within any of the ResourceEventHandlerFuncs and terminate the full NPC application

Per discussions in  https://github.com/weaveworks/weave/issues/3764
and https://github.com/weaveworks/weave/issues/3771

This change will cause the whole NPC application to crash when a panic occurs within any of the ResourceEventHandlerFuncs, which are called when add/update/deletes happen for pods/namespaces/network policies.  Stripped down the weave main package to just this https://gist.github.com/naemono/7c47213de6bb2eaab07b66b7ad2b86b8 and tested within one of our K8s clusters, and proved that this approach will cause the whole application to exit upon an internal panic.  This approach is considerably simpler than a full reconcile on restart, which is discussed in #3771 .  Since the NPC application is in a fully broken state when an internal goroutine crashes, it seems as if this approach (allowing the orchestration system to restart the application/pod) would be better long term than allowing the application to continue running in a failed state.

Suggestions on better approach welcomed.  Unsure exactly how to unit test this; suggestions on that welcomed as well.

### DIFF
--- a/prog/weave-npc/main.go
+++ b/prog/weave-npc/main.go
@@ -11,6 +11,7 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
@@ -219,7 +220,7 @@ func ipsetExist(ips ipset.Interface, name ipset.Name) (bool, error) {
 
 func stopOnPanicRecover(stopChan chan struct{}) {
 	if r := recover(); r != nil {
-		stopChan <- struct{}{}
+		os.Exit(1)
 	}
 }
 
@@ -334,26 +335,11 @@ func root(cmd *cobra.Command, args []string) {
 	}
 	npController = makeController(client.NetworkingV1().RESTClient(), "networkpolicies", &networkingv1.NetworkPolicy{}, npHandlers)
 
+	go nsController.Run(wait.NeverStop)
+	go podController.Run(wait.NeverStop)
+	go npController.Run(wait.NeverStop)
+
 	signals := make(chan os.Signal, 1)
-	stopChan = make(chan struct{})
-
-	go func() {
-		nsController.Run(stopChan)
-		signals <- syscall.SIGINT
-	}()
-	go func() {
-		podController.Run(stopChan)
-		signals <- syscall.SIGINT
-	}()
-	go func() {
-		npController.Run(stopChan)
-		signals <- syscall.SIGINT
-	}()
-	go func() {
-		<-stopChan
-		close(stopChan)
-	}()
-
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 	common.Log.Fatalf("Exiting: %v", <-signals)
 }
--- a/prog/weave-npc/main.go
+++ b/prog/weave-npc/main.go
@@ -11,7 +11,6 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
@@ -32,6 +31,7 @@ var (
 	nodeName       string
 	maxList        int
 	bridgePortName string
+	stopChan       chan struct{}
 )
 
 func handleError(err error) { common.CheckFatal(err) }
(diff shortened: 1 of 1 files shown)


prog/weave-npc/main.go
```diff
@@ -259,9 +259,19 @@ func root(cmd *cobra.Command, args []string) {
 	nsController := makeController(client.Core().RESTClient(), "namespaces", &coreapi.Namespace{},
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
+				defer func() {
+					if r := recover(); r != nil {
```
bboreham, line 263: could this go into a (named) subroutine?

prog/weave-npc/main.go
```diff
@@ -312,16 +352,36 @@ func root(cmd *cobra.Command, args []string) {
...
-	go nsController.Run(wait.NeverStop)
-	go podController.Run(wait.NeverStop)
-	go npController.Run(wait.NeverStop)
-
 	signals := make(chan os.Signal, 1)
+	stopChan = make(chan struct{})
+
+	go func() {
+		nsController.Run(stopChan)
+		signals <- syscall.SIGINT
```
bboreham, line 370: What's the intention here?  Looks like you are faking a signal arriving from the OS?

prog/weave-npc/main.go
```diff
@@ -32,6 +31,7 @@ var (
 	nodeName       string
 	maxList        int
 	bridgePortName string
+	stopChan       chan struct{}
```
bboreham, line 34: I think this could be a local variable in `root()`

prog/weave-npc/main.go
```diff
@@ -312,16 +352,36 @@ func root(cmd *cobra.Command, args []string) {
...
+		podController.Run(stopChan)
+		signals <- syscall.SIGINT
+	}()
+	go func() {
+		npController.Run(stopChan)
+		signals <- syscall.SIGINT
+	}()
+	go func() {
+		<-stopChan
+		close(stopChan)
```
bboreham, line 382: Isn't it dangerous to close this while there are still goroutines that could send to it?

prog/weave-npc/main.go
naemono: I am, since that's what weave is waiting on to completely stop the application fully

prog/weave-npc/main.go
naemono: well, the application is ending anyways, and I was considering waitgroups, but I'm not sure the harm here, as the application is completely shutting down at this point regardless.  I'm certainly not against adding a waitgroup, and waiting until all the others stop completely.

prog/weave-npc/main.go
naemono: that would likely be cleaner, yes.  I will make this change.

prog/weave-npc/main.go
bboreham: How about `os.Exit(1)` ?

prog/weave-npc/main.go
naemono: I didn't initially notice this.  thank you.  change made.

prog/weave-npc/main.go
naemono: yeah, but if the weave team wants waitgroups in place, I'll likely have to use a channel of some sort, and this was already setup to use... I'm not against this, but I'm curious as to how the weave team feels.

naemono:
@bboreham as for the question of why the k8s code isn't handling this, I found this when I was digging into this

https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L145

Which mentions 
```
// Until loops until stop channel is closed, running f every period.
```

Which is why I added the goroutine to close the channel when a signal is sent on this channel.  I was wondering why sending a struct down that channel didn't kill everything, until I found that...

prog/weave-npc/main.go
bboreham: On line 222 we know the program has panic'd; it seems best to exit quickly and simply at that point rather than adding channels, fake signals, etc.

prog/weave-npc/main.go
bboreham: Please don't.

naemono:
I'll get back to this in the next Sprint and spend some more time here.

prog/weave-npc/main.go
naemono: I've updated this to simply `os.Exit` when one of the goroutines panics.

prog/weave-npc/main.go
naemono: this has been removed.

prog/weave-npc/main.go
```diff
@@ -218,8 +218,17 @@ func ipsetExist(ips ipset.Interface, name ipset.Name) (bool, error) {
...
+	if r := recover(); r != nil {
+		os.Exit(1)
+	}
+}
+
 func root(cmd *cobra.Command, args []string) {
-	var npController cache.Controller
+	var (
+		npController cache.Controller
+		stopChan     chan struct{}
```
bboreham, line 230: `stopChan` seems to be left-over and unneeded now.

prog/weave-npc/main.go
```diff
@@ -218,8 +218,17 @@ func ipsetExist(ips ipset.Interface, name ipset.Name) (bool, error) {
 	return false, nil
 }
 
+func stopOnPanicRecover(stopChan chan struct{}) {
+	if r := recover(); r != nil {
+		os.Exit(1)
```
bboreham, line 223: What's the user experience if this happens?  Can they understand where the real problem occurred?

gobomb:
It's maybe a bug of  client-go that  panic in event handler didn't propagated and the process didn't crash.  I created an issue to report it in client-go repo. How do you think？

https://github.com/kubernetes/client-go/issues/838

bboreham:
Brilliant work @gobomb , I agree this is very likely the root cause of the problem.
I have proposed a different fix: https://github.com/kubernetes/kubernetes/pull/93679

If @naemono isn't going to fix the nits I can do that.

bboreham:
Please see my proposed fix-up at https://github.com/weaveworks/weave/compare/master...fail-controller-if-goroutines-fail

(I should add that I consider this `stopOnPanicRecover()` as a temporary work-around until there is a Kubernetes client-go that fixes the underlying problem)

bboreham:
Replaced by #3841

PR #3801, opened 2020-05-13
Disable accept_ra on Weave Net interfaces

bboreham:
To retain control over addressing and routing.

Slight refactor to keep the `/proc` path out of the weeds, and plumb through the setting to the Docker plugin rather than second-guessing it there.

### DIFF
--- a/net/veth.go
+++ b/net/veth.go
@@ -49,6 +49,13 @@ func CreateAndAttachVeth(procPath, name, peerName, bridgeName string, mtu int, k
 	if err := bridgeType.attach(veth); err != nil {
 		return cleanup("attaching veth %q to %q: %s", name, bridgeName, err)
 	}
+	// No ipv6 router advertisments please
+	if err := sysctl(procPath, "net/ipv6/conf/"+name+"/accept_ra", "0"); err != nil {
+		return cleanup("setting accept_ra to 0: %s", err)
+	}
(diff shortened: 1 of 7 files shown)


net/veth.go
```diff
@@ -49,6 +49,13 @@ func CreateAndAttachVeth(name, peerName, bridgeName string, mtu int, keepTXOn bo
 	if err := bridgeType.attach(veth); err != nil {
 		return cleanup("attaching veth %q to %q: %s", name, bridgeName, err)
 	}
+	// No ipv6 router advertisments please
+	if err := sysctl(procPath, "net/ipv6/conf/"+name+"/accept_ra", "0"); err != nil {
```
saada, line 53: I'm unsure whether `sysctl()` adds the leading `/` or not. Assuming procPath=`/proc`, i'm assuming sysctl uses filepath.join().

bboreham: You can see it here: https://github.com/weaveworks/weave/blob/6ad3e69faf9c88425db559cc8bf51c0a9f83c5e1/net/arp.go#L23

saada: i validated that this is the right path: `/proc/sys/net/ipv6/conf/INTERFACENAME/accept_ra`
from [here](https://sysctl-explorer.net/net/ipv6/accept_ra/)

Issue #3586, opened 2019-01-18
Weave Net iptables rules disappear on firewalld reload


JoostvdB94:
enhancement
## What you expected to happen?
Persistent firewall settings, even after firewalld reload

## What happened?
Weave chains & rules in IPTABLES (on a specific node) disappear when reloading the firewall on that node.
Kubernetes and docker rules are coming back, although they seem to disappear too after a reload of firewalld.

## How to reproduce it?
- Rollout (or use a existing) kubernetes cluster (mine is on-premise, using `kubeadm init`)
- (Start `firewalld` if not already running)
- run `systemctl reload firewalld` (or use another tool to do this e.g. `firewall-cmd --reload`)
- Check IPTABLES filter rules `iptables -L -v -n -t filter --line-numbers`

See that `WEAVE-NPC-EGRESS` `WEAVE-NPC` and all linked chains have disappeared.

## Anything else we need to know?
Using Weave as CNI for kubernetes, running version 2.5.0

## Versions:

```
$ weave version 2.5.0
$ docker version 17.03
$ uname -a 
Linux ---- 3.10.0-693.21.1.el7.x86_64 #1 SMP Wed Mar 7 19:03:37 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
$ kubectl version 
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.2", GitCommit:"17c77c7898218073f14c8d573582e8d2313dc740", GitTreeState:"clean", BuildDate:"2018-10-24T06:54:59Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
```

bboreham:
Related:  #2208, #3106

bboreham:
I changed the title because I think it's better to have it describe the problem rather than a potential solution.
Docker has [code](https://github.com/moby/moby/pull/9397/files) to react to firewalld reload.  This might be easier to implement than a full reconcile (or let us poll more slowly if we do do a reconcile).

bboreham:
#3802 has addressed the iptables rules used by the "router" part, but the "npc" (Network Policy Controller) remains to do.
A possible work-around is to configure Weave Net to run with no network policies.

PR #3802, opened 2020-05-13
Reload router iptables rules if they get cleared

foot:
Adds a `CANARY` iptable chain and reloads all the weavenet iptable chains if the `CANARY` is removed by something (usually `firewalld`).

Partially addresses https://github.com/weaveworks/weave/issues/3586

### Details

- This exposes `ConfigureIPTables` and tries to make it safe to call again and again.
-  The destructive resets are pulled out into `ResetIPTables` and are called at launch so iptable state should remain mostly the same as prior to this PR.

### DIFF
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -429,7 +429,7 @@ func (f fastdpImpl) attach(veth *netlink.Veth) error {
 	return odp.AddDatapathInterfaceIfNotExist(f.datapathName, veth.Attrs().Name)
 }
 
-// ResetIPTables resets IPTables if they're in a strange state from a previous run.
+// ResetIPTables resets IPTables in case they're in a strange state from a previous run.
 func ResetIPTables(config *BridgeConfig, ips ipset.Interface) error {
 	ipt, err := iptables.New()
 	if err != nil {
@@ -568,23 +568,13 @@ func ConfigureIPTables(config *BridgeConfig, ips ipset.Interface) error {
 	}
 
 	// For the cases where the weave bridge is the default gateway for
-	// containers (e.g. Kubernetes): create the ipset to store CIDRs allocated
-	// by IPAM for local containers. In the case of Kubernetes, external traffic
-	// sent to these CIDRs avoids SNAT'ing so that NodePort with
-	// `"externalTrafficPolicy":"Local"` would receive packets with correct
-	// src IP addr.
+	// containers (e.g. Kubernetes): In `ResetIPTables` (which we assume
+	// to have been called at this point) we create an ipset to store CIDRs
+	// allocated by IPAM for local containers.
+	// In the case of Kubernetes, external traffic sent to these CIDRs
+	// avoids SNAT'ing so that NodePort with `"externalTrafficPolicy":"Local"`
+	// would receive packets with correct src IP addr.
 	if config.NoMasqLocal {
-		ips := ipset.New(common.LogLogger(), 0)
-
-		noMasqLocalExists, err := ips.Exists(NoMasqLocalIpset)
-		if err != nil {
-			common.Log.Errorf("Failed to look if ipset '%s' exists: %s", NoMasqLocalIpset, err)
-		} else if !noMasqLocalExists {
-			if err := ips.Create(NoMasqLocalIpset, ipset.HashNet); err != nil {
-				return err
-			}
-		}
-
 		if err := ipt.AppendUnique("nat", "WEAVE",
 			"-m", "set", "--match-set", string(NoMasqLocalIpset), "dst",
 			"-m", "comment", "--comment", "Prevent SNAT to locally running containers",
--- a/net/bridge.go
+++ b/net/bridge.go
@@ -262,7 +262,7 @@ func EnsureBridge(procPath string, config *BridgeConfig, log *logrus.Logger, ips
 		break
 	}
 
-	if err := configureIPTables(config, ips); err != nil {
+	if err := ConfigureIPTables(config, ips); err != nil {
 		return bridgeType, errors.Wrap(err, "configuring iptables")
 	}
 
@@ -297,9 +297,7 @@ func EnsureBridge(procPath string, config *BridgeConfig, log *logrus.Logger, ips
 		return bridgeType, errors.Wrapf(err, "configuring ARP cache on bridge %q", config.WeaveBridgeName)
 	}
 
-	// NB: No concurrent call to Expose is possible, as EnsureBridge is called
-	// before any service has been started.
-	if err := reexpose(config, log); err != nil {
+	if err := Reexpose(config, log); err != nil {
 		return bridgeType, err
 	}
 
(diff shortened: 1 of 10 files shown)


prog/weaver/main.go
```diff
@@ -506,6 +508,26 @@ func main() {
...
+					Log.Print("Re-configuring iptables")
+					err := weavenet.ConfigureIPTables(&bridgeConfig, ips)
+					if err != nil {
+						Log.Errorf("Error configuring iptables: %s", err)
+					}
+				}
+			}
+		}()
+	} else {
+		Log.Printf("Not refreshing iptables (interval: %v)", iptablesRefresh)
```
foot, line 528: Log.Debug wasn't working w/ `LOG_LEVEL=debug` but I might have been doing something wrong?

net/bridge.go
```diff
@@ -426,7 +426,24 @@ func (f fastdpImpl) attach(veth *netlink.Veth) error {
 	return odp.AddDatapathInterfaceIfNotExist(f.datapathName, veth.Attrs().Name)
 }
 
-func configureIPTables(config *BridgeConfig, ips ipset.Interface) error {
+// Dummy way to check whether a given ipset exists.
+// TODO(brb) Use "ipset -exist create <..>" for our purpose instead (for some reasons
+// creating an ipset with -exist fails).
+func ipsetExist(ips ipset.Interface, name ipset.Name) (bool, error) {
```
foot, line 432: This is a handy fn a copy pasted from somewhere else. Where do you think is a good home?

prog/weaver/main.go
bboreham: I don't think there is an env var - you can do `weave launch --log-level=debug`
Feel free to add the env var in a separate PR.

net/bridge.go
bboreham: I think it can go in `weaveworks/weave/net/ipset/ipset.go`, though there is already a function `Exist()` that does something slightly different.  It should probably be renamed `ExistEntry`.

prog/weaver/main.go
```diff
@@ -216,6 +217,7 @@ func main() {
 	mflag.StringVar(&discoveryEndpoint, []string{"-peer-discovery-url"}, "https://cloud.weave.works/api/net", "url for peer discovery")
 	mflag.StringVar(&token, []string{"-token"}, "", "token for peer discovery")
 	mflag.StringVar(&advertiseAddress, []string{"-advertise-address"}, "", "address to advertise for peer discovery")
+	mflag.DurationVar(&iptablesRefresh, []string{"-iptables-refresh-interval"}, 0*time.Second, "Interval to reapply iptables. 0 to only apply on launch.")
```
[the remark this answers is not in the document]
foot: Lets take it for a spin

net/ipset/ipset.go
```diff
@@ -121,10 +122,26 @@ func (i *ipset) DelEntry(user types.UID, ipsetName Name, entry string) error {
 	return doExec("del", string(ipsetName), entry)
 }
 
-func (i *ipset) Exist(user types.UID, ipsetName Name, entry string) bool {
+func (i *ipset) EntryExists(user types.UID, ipsetName Name, entry string) bool {
 	return i.existUser(user, ipsetName, entry)
 }
 
+// Dummy way to check whether a given ipset exists.
+// TODO(brb) Use "ipset -exist create <..>" for our purpose instead (for some reasons
```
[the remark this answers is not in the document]
foot: Yes, https://github.com/weaveworks/weave/commit/18a6237ec7713161f6fde5712d8d83e26e6a829a 

`// delete `weave-local-pods` ipset which is no longer used by weave-npc`

There is no longer a create I'll drop the comment.

net/bridge.go
```diff
@@ -530,11 +560,17 @@ func configureIPTables(config *BridgeConfig, ips ipset.Interface) error {
...
+		if err != nil {
+			common.Log.Errorf("Failed to look if ipset '%s' exists: %s", NoMasqLocalIpset, err)
+		} else if !noMasqLocalExists {
+			if err := ips.Create(NoMasqLocalIpset, ipset.HashNet); err != nil {
+				return err
+			}
 		}
-		if err := ipt.Insert("nat", "WEAVE", 1,
+
+		if err := ipt.AppendUnique("nat", "WEAVE",
```
bboreham, line 573: Mostly a note to myself: we were only adding one rule to the `WEAVE` chain so it doesn't matter whether we `Insert` or `Append` it.

net/bridge.go
```diff
@@ -292,9 +292,7 @@ func EnsureBridge(procPath string, config *BridgeConfig, log *logrus.Logger, ips
 		return bridgeType, errors.Wrapf(err, "configuring ARP cache on bridge %q", config.WeaveBridgeName)
 	}
 
-	// NB: No concurrent call to Expose is possible, as EnsureBridge is called
-	// before any service has been started.
```
foot, line 296: This is a bit of a question.

But `Expose` can already by hit concurrently as its an http endpoint too?

There are some comments in the `Expose` helpers like
> [// The IP addr might have been already set by a concurrent request, just ignore then](https://github.com/weaveworks/weave/blob/master/net/expose.go#L61)

Which suggest it should be okay? We probably don't need to wrap Expose in a Lock or anything?

bboreham: Your point seems valid.  Maybe the thinking was that an address could not be added while we were traversing the list.
Maybe it relates to some earlier state of the code where concurrency was an issue.

foot:
Manual testing:
- I've loaded this up in a 3 node k8s cluster on GCP / RHEL7
- Disabled the npc container
- Restart firewalld 
- weave-net recovers all its iptable rules!

net/bridge.go
```diff
@@ -426,10 +424,40 @@ func (f fastdpImpl) attach(veth *netlink.Veth) error {
 	return odp.AddDatapathInterfaceIfNotExist(f.datapathName, veth.Attrs().Name)
 }
 
-func configureIPTables(config *BridgeConfig, ips ipset.Interface) error {
+// ResetIPTables resets IPTables if they're in a strange state from a previous run.
```
bboreham, line 427: s/if/in case/

net/bridge.go
```diff
@@ -530,11 +558,17 @@ func configureIPTables(config *BridgeConfig, ips ipset.Interface) error {
...
-		if err := ips.Create(NoMasqLocalIpset, ipset.HashNet); err != nil {
-			return err
+
+		noMasqLocalExists, err := ips.Exists(NoMasqLocalIpset)
+		if err != nil {
+			common.Log.Errorf("Failed to look if ipset '%s' exists: %s", NoMasqLocalIpset, err)
+		} else if !noMasqLocalExists {
+			if err := ips.Create(NoMasqLocalIpset, ipset.HashNet); err != nil {
+				return err
+			}
```
bboreham, line 568: On reflection I think this change is wrong: we don't make the tracker restart from scratch so if we do recreate the ipset mid-flight it won't have the right stuff in it.  I think it is ok just to move the destroy/recreate to `ResetIPTables()` and assume it exists here.

net/expose.go
```diff
@@ -103,13 +103,8 @@ func addNatRule(ipt *iptables.IPTables, rulespec ...string) error {
...
 	for {
 		if err := ipt.AppendUnique("nat", "WEAVE", rulespec...); err != nil {
-			if ierr, ok := err.(*iptables.Error); ok {
-				if status, ok := ierr.ExitError.Sys().(syscall.WaitStatus); ok {
-					// (magic exit code 4 found in iptables source code; undocumented)
-					if status.ExitStatus() == 4 {
-						continue
-					}
-				}
+			if isResourceError(err) {
```
bboreham, line 106: good spot!

prog/weaver/main.go
```diff
@@ -506,6 +512,22 @@ func main() {
...
+		applyIPTables := func() {
+			Log.Info("Re-configuring iptables")
+			err := weavenet.ConfigureIPTables(&bridgeConfig, ips)
+			if err != nil {
+				Log.Errorf("Error configuring iptables: %s", err)
+			}
+
+			weavenet.Reexpose(&bridgeConfig, Log)
+		}
+		go weavenet.Monitor(Log, "WEAVE-CANARY", []string{"mangle", "nat", "filter"}, applyIPTables, iptablesRefresh, wait.NeverStop)
```
bboreham, line 526: We do shut down other goroutines started here, generally with some kind of `defer Stop()`, so creating a `chan` and `defer close()`-ing it would be preferable to `NeverStop`.

foot:
@bboreham PTAL at the last 2 commits when you have a moment!

prog/weaver/main.go
```diff
@@ -216,6 +217,7 @@ func main() {
 	mflag.StringVar(&discoveryEndpoint, []string{"-peer-discovery-url"}, "https://cloud.weave.works/api/net", "url for peer discovery")
 	mflag.StringVar(&token, []string{"-token"}, "", "token for peer discovery")
 	mflag.StringVar(&advertiseAddress, []string{"-advertise-address"}, "", "address to advertise for peer discovery")
+	mflag.DurationVar(&iptablesRefresh, []string{"-iptables-refresh-interval"}, 10*time.Second, "Interval to reapply iptables. 0 to only apply on launch.")
```
bboreham, line 220: This description is now out of date.  In fact since we only use it to poll the canary chain, I suspect we don't need to make it configurable.

prog/weaver/main.go
```diff
@@ -512,6 +517,24 @@ func main() {
 		go exposeForAWSVPC(allocator, defaultSubnet, bridgeConfig.WeaveBridgeName, waitReady.Add())
 	}
 
+	if iptablesRefresh > 0 {
+		Log.Printf("Entering iptable refresh loop (interval %v)", iptablesRefresh)
```
bboreham, line 521: Probably don't need this.

net/iptables.go
```diff
@@ -104,3 +108,117 @@ func ensureRulesAtTop(table, chain string, rulespecs [][]string, ipt *iptables.I
...
+}
+
+const (
+	// Max time we wait for an iptables flush to complete after we notice it has started
+	iptablesFlushTimeout = 5 * time.Second
+	// How often we poll while waiting for an iptables flush to complete
+	iptablesFlushPollTime = 100 * time.Millisecond
+)
+
+// Monitor is part of Interface
```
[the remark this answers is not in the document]
bboreham: Reminder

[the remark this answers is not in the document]
bboreham: @foot did you consider changing this?

Issue #3828, opened 2020-07-05
Another app is currently holding the xtables lock. Perhaps you want to use the -w option?


Congelli501:
## What you expected to happen?
Weave starts normally, on the first try

## What happened?
On a busy cluster (pod waiting to be created on a node) weave takes a lot of tries to start (fail at start, goes into CrashLoopBackOff status).
The "weave" container of the weave pod of the node shows the following log:
```
Another app is currently holding the xtables lock. Perhaps you want to use the -w option?
```

After killing the weave pod a lot of time (to retry after a CrashLoopBackOff), weave end up starting up, and then run normally.
On one of my node, it took me 20 minutes to make weave start.

## How to reproduce it?
Setup a busy kubernetes node (100 pods) and reboot it. The node will start with a lot of assigned pod waiting to start and won't be able to start weave.

## Anything else we need to know?
This bug occurred after a cluster upgrade (reinstall).
I didn't have this issue on my previous cluster (Ubuntu 18.04 amd64, Kubernetes v1.13.11, Weave 2.5.2)

Cloud provider: bare metal

## Versions:

```
$ weave version
2.6.5
$ docker version
Client:
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.13.8
 Git commit:        afacb8b7f0
 Built:             Tue Jun 23 22:26:12 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.8
  Git commit:       afacb8b7f0
  Built:            Thu Jun 18 08:26:54 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.3.3-0ubuntu2
  GitCommit:        
 runc:
  Version:          spec: 1.0.1-dev
  GitCommit:        
 docker-init:
  Version:          0.18.0
  GitCommit:        

$ uname -a
Linux master0 5.4.0-40-generic #44-Ubuntu SMP Tue Jun 23 00:01:04 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

$ kubectl version
[...]

PR #3835, opened 2020-07-28
wait added to iptables commands

nesc58:
Hi weave-team,
I have found some code lines where the `-w` flag is missing for the `iptables` command.

What do you think about it?

It may could solve the issue: https://github.com/weaveworks/weave/issues/3828

The commands runs before the weave go application runs. Within the application all iptables commands are wrapped with the `go-iptables`-package which uses always the `--wait` flag internally


bboreham:
Thanks; I've re-pushed to the weaveworks repo so the smoke-tests will run.

Issue #3847, opened 2020-08-20
ipset v7.2: Set cannot be destroyed: it is in use by a kernel component - proposed solution


KevDBG:
## What you expected to happen?

weaveworks/weave-kube:2.6.4 start

## What happened?
Related to the issue https://github.com/weaveworks/weave/issues/3828, i have the same problem in other context. 
We using rancher rke 2.4.5 and k8s K8S cluster v1.18-6-rancher1-1. When rancher try to start weaveworks/weave-kube:2.6.4, is stuck with the log "ipset v7.2: Set cannot be destroyed: it is in use by a kernel component".

## How to reproduce it?
Indeed, i reproduce the problem directly in the container, i tried to launch all these commands in one block (at the same time):

```
/ #     iptables -w -F WEAVE-KUBE-TEST
/ #     iptables -w -X WEAVE-KUBE-TEST
/ #     ipset destroy weave-kube-test
/ #     iptables -w -F WEAVE-KUBE-TEST 2>/dev/null || true
/ #     iptables -w -X WEAVE-KUBE-TEST 2>/dev/null || true
/ #     ipset destroy weave-kube-test 2>/dev/null || true
/ # 
/ #     ipset create weave-kube-test hash:ip
/ #     iptables -w -t filter -N WEAVE-KUBE-TEST
/ #     if ! iptables -w -A WEAVE-KUBE-TEST -m set --match-set weave-kube-test src -j DROP; then
>         NOT_EXIST=1
>     fi
/ #     iptables -w -F WEAVE-KUBE-TEST
/ #     iptables -w -X WEAVE-KUBE-TEST
/ #     ipset destroy weave-kube-test
ipset v7.2: Set cannot be destroyed: it is in use by a kernel component

```
In this case we have the related issue (even if we use the -w). I tried an other thing related to this comment -> https://github.com/weaveworks/weave/issues/3816#issuecomment-640474405

```
/ #     iptables -w -F WEAVE-KUBE-TEST
/ #     iptables -w -X WEAVE-KUBE-TEST
/ #     ipset destroy weave-kube-test
/ #     iptables -w -F WEAVE-KUBE-TEST 2>/dev/null || true
/ #     iptables -w -X WEAVE-KUBE-TEST 2>/dev/null || true
/ #     ipset destroy weave-kube-test 2>/dev/null || true
/ # 
/ #     ipset create weave-kube-test hash:ip
/ #     iptables -w -t filter -N WEAVE-KUBE-TEST
[...]

PR #3851, opened 2020-08-24
Pause and retry ipset deletion if it fails the first time

NeonSludge:
There is an issue (#3847) with using both `iptables` in the `nft` mode and `ipset` on modern kernels (for example, the ones used by CentOS 8 or Ubuntu 20.04). If an ipset is being destroyed right after deleting a referencing rule then there is a good chance that the `ipset destroy` command will fail with a `"Set cannot be destroyed: it is in use by a kernel component"` error.

The more permanent solution would be retiring `ipset` and `iptables-nft` altogether, fully migrating to nftables and leveraging its native data types support instead of the IP sets.
This might be a lengthy process, so this PR offers a temporary solution until then: a second try after a 100ms pause that gets triggered if `ipset destroy` fails.

### DIFF
--- a/net/ipset/ipset.go
+++ b/net/ipset/ipset.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


bboreham:
Thanks!  Agreed this is not ideal but will hopefully improve symptoms in the short term.

Issue #3829, opened 2020-07-06
Weave net randomly blocking connections


bcollard:
## What you expected to happen?
No connection randomly blocked, especially when there are no netpol (network policies) or when netpols target different ports and pods. See logs at the end of this issue.
Same issue than https://github.com/weaveworks/weave/issues/3761

## What happened?
Looking at the logs of my weave-net pods, I can see multiple connections blocked by weave-npc. Looks like it's random.

## How to reproduce it?
Installed Weave CNI plugin with this method: https://www.weave.works/docs/net/latest/kubernetes/kube-addon/#-installation

## Anything else we need to know?
On-prem DC. K8s installed with kubeadm. Here is my kubeadm `ClusterConfiguration`: https://gist.github.com/bcollard/b55108e4355b4edab169a025e02723e8
I first installed Weave-net without specifying the pod subnet. So pods spawn with IP address within the 10.32.0.0/12 range. Then, I redeployed the weave daemonset with the IPALLOC_RANGE set to 10.244.0.0/16, which is the subnet configured for pods in kubeadm.
That's why you'll see pods with these 2 kinds of IP; and also a bit more IP routes and rules.

## Versions:
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:07:13Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```

## Logs:
```
$ kubectl logs -n kube-system <weave-net-pod> weave
```
https://gist.github.com/bcollard/bda85abdcb2ee9c5779dc38512b494eb

```
$ ks exec -it weave-net-npt9n -c weave -- /home/weave/weave --local status

        Version: 2.6.5 (failed to check latest version - see logs; next check at 2020/07/07 02:06:37)

        Service: router
       Protocol: weave 1..2
           Name: 46:b1:bd:d0:2d:c5(fr0-viaas-271)
     Encryption: disabled
[...]

PR #3852, opened 2020-08-26
Move egress logging rule beside drop rule

bboreham:
Otherwise we might log a blocked connection even though the packet was not dropped.

Noticed while investigating #3829


bboreham:
Yes, instead of setting it true before attempting then setting it back to false on error, we just wait till both commands have succeeded before setting it true.

The whole function has a lock on `npc` so we can't get another thread trying to start the operation at the same time.

PR #3877, opened 2020-12-17
Tighten Kubernetes volume mounts

bboreham:
When the Weave Net DaemonSet first installs on a node, it writes out CNI binaries and configuration into the host filesystem. 
Mount those directories closer to where they are needed so that the pod does not have write access to so much of the host disk.


bboreham:
Needs rework after #3880

PR #3996, opened 2023-08-02
Reweave - an effort to create a modern build system for weave net

rajch:
This is an effort to create a modern build environment for weave net, and simultaneously upgrade dependencies and the go compiler version.

This should solve some known problems, including:
* Incompatibility with some versions of CNI and containerd
* Vulnerabilities in the current weave images (2.8.1)
* Mismatch between the Docker Hub hosted weave:latest and weave:2.8.1 images
* Problems with multi-architecture images


kingdonb:
Thank you for doing all this @rajch ! You are a pillar of the Weave Net community, I'm pleased to take a look at this today.

kingdonb:
This is a mega-pr! 🎉  Thanks for pushing all this work, I will find time to review it soon.

I've been busy working on Flux v2.1 release, so we can get the docs in order and release them without any internal inconsistency. Sorry for the delays in merging this.

rajch:
Some more CVEs had turned up. Resolved by upgrading dependencies.

